PR #21799, opened 2024-04-22
Implement wgpuAdapterGetLimits, add test that limit requests are applied

### LINKED ISSUE #21798: Missing field maxColorAttachmentBytesPerSample in requiredLimits for wgpuAdapterRequestDevice
When requesting a webGPU device, a device descriptor is passed. The device descriptor may specify certain limits, the device must be capable of supporting ("required limits"). Currently, the limit `maxColorAttachmentBytesPerSample` is always ignored.

The C function for requesting a webGPU device is `wgpuAdapterRequestDevice` and its binding is in `library_webgpu.js`: 

https://github.com/emscripten-core/emscripten/blob/11bd2d995584a21e6756c44a64cf2eaf8c988077/src/library_webgpu.js#L2550

The descriptor passed is a C struct, which is used to fill a javascript object `requiredLimits`. The fields are explicitly copied using `setLimitU32IfDefined`/`setLimitU64IfDefined`, but the field `maxColorAttachmentBytesPerSample` is missing. This means, specifying this limits as part of the required limits for the device won't have any effect and the device will always support the default value for `maxColorAttachmentBytesPerSample`.

https://github.com/emscripten-core/emscripten/blob/11bd2d995584a21e6756c44a64cf2eaf8c988077/src/library_webgpu.js#L2562-L2613

pkomon-tgm:
- Implement the wgpuAdapterGetLimits entry point.
- Add a test which verifies that limit requests are correctly passed through to requestDevice(), by requesting all of the adapter's max limits, then ensuring that the device has the same limits. This verifies that wgpuAdapterGetLimits and wgpuDeviceGetLimits work, and that limit requests are applied (though it can only verify limits for which the current system supports a value better than the default).

Fixes #21798


kainino0x (dismissed):
Thanks!

kainino0x (commented):
WebGPU isn't available on the test VMs so you'll need to exit cleanly if WebGPU isn't available, similar to the neighboring test:
https://github.com/emscripten-core/emscripten/blob/e57881a98f10dc10672eadbb32e7a7720d3795f6/test/webgpu_basic_rendering.cpp#L28

pkomon-tgm:
Ah, right, I missed that webGPU might not be available - fixed it :)

test/webgpu_required_limits.cpp
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  const adapter = await navigator.gpu.requestAdapter();
+  if (!adapter) {
+    throw Error("Couldn't request WebGPU adapter.");
+  }
+
+  console.log("adapter limits", adapter.limits);
+
+  return adapter.limits.maxColorAttachmentBytesPerSample;
```
sbc100, line 20: It looks kind of like we are peaking past the C/C++ API here and directly accessing the webgpu JS API.. is that purely so that we can verify that the C/C++ API reporting the same thing as the underlying JS API?   Is that OK/normal/expected for webGPU testing?

test/webgpu_required_limits.cpp
```diff
@@ -0,0 +1,98 @@
...
+  assert(status == WGPURequestDeviceStatus::WGPURequestDeviceStatus_Success);
+
+  WGPUSupportedLimits device_supported_limits{};
+  wgpuDeviceGetLimits(device, &device_supported_limits);
+
+  std::cout << "required maxColorAttachmentBytesPerSample="
+            << adapter_limit_maxColorAttachmentBytesPerSample << std::endl;
+  std::cout << "supported maxColorAttachmentBytesPerSample="
+            << device_supported_limits.limits.maxColorAttachmentBytesPerSample
+            << std::endl;
```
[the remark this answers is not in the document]
pkomon-tgm: No, you are right, the test doesn't depend on C++ at all; changed it.

test/webgpu_required_limits.cpp
pkomon-tgm: So we are trying to request a webGPU device with limits higher than the default limits and if we get a device, we want the device limits to fulfill the requested limits. For I set only `maxColorAttachmentBytesPerSample`, because this was previously ignored (i.e. requesting a device with `maxColorAttachmentBytesPerSample` did not fail but returned a device that only supported the default value for `maxColorAttachmentBytesPerSample`).

But we can only test requesting higher-than-default limits, if the physical device (webGPU adapter) actually supports higher-than-default limits. So, I first check if the supported limit is higher than the default one (which is currently 32 according to the [webGPU spec](https://www.w3.org/TR/webgpu/#dom-supported-limits-maxcolorattachmentbytespersample)) and if not, I let the test pass. The reason for using the JS API directly here is that the C API bindings are not implemented yet: 

https://github.com/emscripten-core/emscripten/blob/6c23c37f2575e7249164911e10cbc06df752a4dd/src/library_webgpu.js#L2539-L2542

I hope letting the test pass when we can't request the higher limit is an alright way to do it (although the test case passing isn't really meaningful then).

PS: Its my first time writing a test-case that is depending on HW limits, please be patient with me here 🥲

sbc100: Sounds reasonable to me, I just wanted to check my assumptions.

@kainino0x is the authority on this stuff so I'll leave lgtm them.

pkomon-tgm: Now that I think about it, implementing `wgpuAdapterGetLimits` is trivial (and the implementation would be pretty much identical to `wgpuDeviceGetLimits`. Maybe encapsulate filling the WGPULimit struct in a new function in `LibraryWebGPU` (if is this the right place to add such a function?).

I think it would make this test case slightly more useful (and eliminate the need for accessing the JS API as well).

Would I have to open a new PR or can I just add it as part of this one?

test/webgpu_required_limits.cpp
```diff
@@ -0,0 +1,98 @@
...
+         adapter_limit_maxColorAttachmentBytesPerSample);
+}
+
+void on_adapter_request_ended(WGPURequestAdapterStatus status,
+                              WGPUAdapter adapter,
+                              char const* message,
+                              void* userdata) {
+  if (status == WGPURequestAdapterStatus_Unavailable) {
+    std::cout << "WebGPU unavailable; exiting cleanly" << std::endl;
+    exit(0);
```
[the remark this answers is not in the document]
kainino0x: This would be great. Do we need to be able to detect whether WebGPU is going to be available?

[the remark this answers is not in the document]
sbc100: The way we do this is that we default to assuming everything is available and then ask folks to opt out any setting environment variables.  e.g. `EMTEST_LACKS_GRAPHICS_HARDWARE`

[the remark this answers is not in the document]
kainino0x: Easy, that would be perfect.

test/webgpu_required_limits.cpp
kainino0x: Sure, let's go ahead and do that. It would resolve one small concern I have that this test currently relies on comparing two different adapter objects which aren't guaranteed to be the same.

test/webgpu_required_limits.c
```diff
@@ -0,0 +1,97 @@
...
+
+  printf("required maxColorAttachmentBytesPerSample=%d\n",
+         adapter_limit_maxColorAttachmentBytesPerSample);
+  printf("supported maxColorAttachmentBytesPerSample=%d\n",
+         device_supported_limits.limits.maxColorAttachmentBytesPerSample);
+
+  // device supported limit MUST be equal or larger than what was requested,
+  // otherwise device acquisition should have failed
+  assert(device_supported_limits.limits.maxColorAttachmentBytesPerSample >=
+         adapter_limit_maxColorAttachmentBytesPerSample);
```
kainino0x, line 40: The device limits should be always exactly what was requested, so this can test `==`.

test/webgpu_required_limits.c
```diff
@@ -0,0 +1,97 @@
...
+#include <webgpu/webgpu.h>
+
+int adapter_limit_maxColorAttachmentBytesPerSample = -1;
+
+EM_ASYNC_JS(int, get_limit, (), {
+  if (!navigator.gpu) {
+    throw Error("WebGPU not supported.");
+  }
+
+  const adapter = await navigator.gpu.requestAdapter();
```
kainino0x, line 13: (As mentioned this adapter object could technically be a different hardware adapter than the other one which could cause the test to fail. Using GetLimits would avoid this. This is incredibly unlikely to actually cause problems though.)

test/webgpu_required_limits.c
```diff
@@ -0,0 +1,97 @@
...
+  assert(status == WGPURequestAdapterStatus_Success);
+
+  // retrieving limits supported by adapter - currently unsupported!
+  //WGPUSupportedLimits adapter_supported_limits = {0,};
+  //wgpuAdapterGetLimits(adapter, &adapter_supported_limits);
+  //adapter_limit_maxColorAttachmentBytesPerSample = adapter_supported_limits.limits.maxColorAttachmentBytesPerSample;
+
+  // use js callout instead
+  adapter_limit_maxColorAttachmentBytesPerSample = get_limit();
+  assert(adapter_limit_maxColorAttachmentBytesPerSample > 0);
```
kainino0x, line 61: nit: You can assert that it's `>= 32`

pkomon-tgm:
Figured this should fix all comments, so I replied here rather than indiviually.

Implemented binding for `wgpuAdapterGetLimits`, I moved the code for filling up the WGPULimits structs into a function in the `WebGPU` object, hope this is the correct place to add it. Also changed the test case to test whether the requested device supports all requested limits rather than just `maxColorAttachmentBytesPerSample`. This way I found that another field was not set when requesting a device, which was `maxBufferSize`. So I also added that.

test/webgpu_required_limits.cpp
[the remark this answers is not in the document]
kainino0x: I have opened https://github.com/emscripten-core/emscripten/pull/21870 for that and I'll complete it after this PR lands.

test/webgpu_required_limits.c
```diff
@@ -0,0 +1,93 @@
...
+  ASSERT_LIMITS_COMPATIBLE(maxInterStageShaderVariables);
+  ASSERT_LIMITS_COMPATIBLE(maxColorAttachments);
+  ASSERT_LIMITS_COMPATIBLE(maxColorAttachmentBytesPerSample);
+  ASSERT_LIMITS_COMPATIBLE(maxComputeWorkgroupStorageSize);
+  ASSERT_LIMITS_COMPATIBLE(maxComputeInvocationsPerWorkgroup);
+  ASSERT_LIMITS_COMPATIBLE(maxComputeWorkgroupSizeX);
+  ASSERT_LIMITS_COMPATIBLE(maxComputeWorkgroupSizeY);
+  ASSERT_LIMITS_COMPATIBLE(maxComputeWorkgroupSizeZ);
+  ASSERT_LIMITS_COMPATIBLE(maxComputeWorkgroupsPerDimension);
+#undef assertLimitCompatible
```
kainino0x, line 46:
```suggestion
#undef ASSERT_LIMITS_COMPATIBLE
```

test/webgpu_required_limits.c
```diff
@@ -0,0 +1,93 @@
...
+  wgpuAdapterRequestDevice(adapter, &device_desc, on_device_request_ended, NULL);
+}
+
+int main() {
+  const WGPUInstance instance = wgpuCreateInstance(NULL);
+
+  WGPURequestAdapterOptions adapter_options = {0,};
+  wgpuInstanceRequestAdapter(instance, &adapter_options, on_adapter_request_ended, NULL);
+
+  return 0;
```
kainino0x, line 92:
```suggestion
  // This code is returned when the runtime exits unless something else sets
  // it, like exit(0).
  return 99;
```

src/library_webgpu.js
```diff
@@ -339,6 +339,54 @@ var LibraryWebGPU = {
...
+      var limitsPtr = {{{ C_STRUCTS.WGPUSupportedLimits.limits }}};
+
+      function setLimitValueU32(name, limitOffset) {
+        var limitValue = limits[name];
+        {{{ makeSetValue('limitsOutPtr', 'limitsPtr + limitOffset', 'limitValue', 'i32') }}};
+      }
+      function setLimitValueU64(name, limitOffset) {
+        var limitValue = limits[name];
+        {{{ makeSetValue('limitsOutPtr', 'limitsPtr + limitOffset', 'limitValue', 'i64') }}};
+      }
```
kainino0x, line 352: existing nit, but this could be clearer:

```suggestion
    fillLimitStruct: (limits, supportedLimitsOutPtr) => {
      var limitsOutPtr = supportedLimitsOutPtr + {{{ C_STRUCTS.WGPUSupportedLimits.limits }}};

      function setLimitValueU32(name, limitOffset) {
        var limitValue = limits[name];
        {{{ makeSetValue('limitsOutPtr', 'limitOffset', 'limitValue', 'i32') }}};
      }
      function setLimitValueU64(name, limitOffset) {
        var limitValue = limits[name];
        {{{ makeSetValue('limitsOutPtr', 'limitOffset', 'limitValue', 'i64') }}};
      }
```

kainino0x (approved):
Looks great, thank you! And nice catch on `maxBufferSize`.

kainino0x:
@sbc100 the remaining failure is happening on `main` too, could you merge this after that's fixed, or force merge this?

sbc100:
Happy to force merge there.   Could you prepare to good commit message in the PR description.  The current on I get if I merge now would be:

```
* Add test case for field maxColorAttachmentBytesPerSample when requesting webGPU device

* Add missing field maxColorAttachmentBytesPerSample when requesting webGPU device

* Fix failing test for webGPU required limits when webGPU unavailable

* Use only c style interface in webGPU required limits test

* Fix C style for webgpu required limits test

* Add impl for wgpuAdapterGetLimits, add missing field maxBufferSize to required limits

- move code filling WGPULimits structs into func
- use from wgpuAdapterGetLimits and wgpuDeviceGetLimits
- rewrite test case for required limit to check if supported limits
  match requested limits
- add missing field maxBufferSize when requesting device

* Update test/webgpu_required_limits.c

* Apply suggestions from code review
```

I'm not sure which parts are relevent.

kainino0x:
@sbc100 Yes, good idea. I've rewritten the PR title and summary, please use those.

```
Implement wgpuAdapterGetLimits, add test that limit requests are applied
```
```
- Implement the wgpuAdapterGetLimits entry point
- Add a test which verifies that limit requests are correctly passed through to requestDevice(), by requesting all of the adapter's max limits, then ensuring that the device has the same limits. This verifies that wgpuAdapterGetLimits and wgpuDeviceGetLimits work, and that limit requests are applied (though it can only verify limits for which the current system supports a value better than the default).

Fixes #21798
```

PR #3277, opened 2015-03-20
readlink should normalize against parent, not against cwd

janjongboom:
Currently running the test suite, the tests I could find that utilize symlinks (unistd_links / stat) run ok now.

I verified the behavior specified in the modified test against gcc, previous definition was not correct.

Issue #3222, opened 2015-03-03
NODEFS cannot read files from sparse disk image when mounted with absolute path under OSX


janjongboom:
I have a sparse disk image (f.e. to have a case-sensitive file system) on OS/X, which is mapped from `/Volumes/repos` to `/Users/janjongboom/repos`. Now when I mount my full file system through NODEFS:

``` js
FS.mount(NODEFS, { root: '/' }, '/test');
```

And then trying to read a file from the disk image through C++:

``` cpp
FILE* fd = fopen("/test/Users/janjongboom/repos/janos/README.md", "r");
```

It fails, because NODEFS readlink() returns `/Volumes/repos`. Which is incorrect. It should take the mountpoint of the filesystem into account, because this folder is mapped on `/test/Volumes/repos`.

Patch in #3223

sbc100:
I now think the fix for this issue was wrong.     If you have links to `/Volumes/repos` in your filesystem, but `/Volumes/repos` is not part of the emscripten VFS then they should not be readable.  i.e. they are effectivly dangling symlinks from the POV of the emscripten filesystem, even though they point a valid location in the external filesystem.

For this kind of thing to would you would either need to mount `/Volumes/repos` inside the VFS, or use NODERAWFS.

sbc100:
We are looking at effectively reverting this change in #21805

Issue #21653, opened 2024-03-29
lto2.test_embind_i64_val is failing


sbc100:
Looks like an issue with LTO combined with JS stub library:

```
/usr/local/google/home/sbc/dev/wasm/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libembind-rtti.a(bind.o): lazy definition of __getTypeName
/tmp/tmp4nrdaxqolibemscripten_js_symbols.so: exported __getTypeName due to import of _emval_as_int64
wasm-ld: error: /usr/local/google/home/sbc/dev/wasm/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libembind-rtti.a(bind.o): attempt to add bitcode file after LTO (__getTypeName)
/tmp/tmp4nrdaxqolibemscripten_js_symbols.so: exported __getTypeName due to import of _emval_as_uint64
/tmp/tmp4nrdaxqolibemscripten_js_symbols.so: exported __getTypeName due to import of _emval_take_value
em++: error: '/usr/local/google/home/sbc/dev/wasm/llvm-build/bin/wasm-ld -o test_i64_val.wasm /tmp/emscripten_temp_sc7s8kis/test_i64_val_0.o -lembind-rtti --trace-symbol=__getTypeName -L/usr/local/google/home/sbc/dev/wasm/emscripten/cache/sysroot/lib/wasm32-emscripten/lto -lGL-getprocaddr -lal -lhtml5 -lbulkmemory -lstubs -lnoexit -lc -ldlmalloc -lcompiler_rt -lc++-noexcept -lc++abi-noexcept -lsockets -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr /tmp/tmp4nrdaxqolibemscripten_js_symbols.so --strip-debug -u__cxa_atexit --export=_emscripten_stack_alloc --export=__wasm_call_ctors --export=emscripten_stack_get_current --export=_emscripten_stack_restore --export-if-defined=__start_em_asm --export-if-defined=__stop_em_asm --export-if-defined=__start_em_lib_deps --export-if-defined=__stop_em_lib_deps --export-if-defined=__start_em_js --export-if-defined=__stop_em_js --export-if-defined=main --export-if-defined=__main_argc_argv --export-table -z stack-size=65536 --no-growable-memory --initial-heap=16777216 --no-entry --table-base=1 --global-base=1024' failed (returned 1)
```

PR #21805, opened 2024-04-23
[nodefs] Don't try to resolve absolute symlinks outside of the VFS

mho22:
This PR reverts the behaviour that was added in #3277 and suggested in #3222.


src/library_fs.js
```diff
@@ -904,7 +904,8 @@ FS.staticInit();` +
       if (!link.node_ops.readlink) {
         throw new FS.ErrnoError({{{ cDefs.EINVAL }}});
       }
-      return PATH_FS.resolve(FS.getPath(link.parent), link.node_ops.readlink(link));
+      var parent = PATH_FS.resolve(FS.getPath(link.parent), link.node_ops.readlink(link)) === path ? link.parent.parent : link.parent;
```
sbc100, line 907: Can you construct a test case that shows why this is needed?

Perhaps modify `test/unistd/links.c`.

Also can you explain in more detail why this is needed?  I'm strugglying the understanding why sometimes we resolve relative to one path and sometimes relative to another

sbc100: Thank you for the context, https://github.com/WordPress/wordpress-playground/issues/1283 looks like it has more info

mho22: @sbc100 The value of the `public/store` symlink when I run `ls -la` in the `public` directory returns an absolute path `/Users/.../test/store` while my test folder name is `test`. 

I used the php method [symlink()](https://www.php.net/manual/en/function.symlink.php) to create the symlink.

I hope this helps.

sbc100: What arguments did you use to create the symlink?   i.e. is it a link to `../store` perhaps?  Or maybe and absolute path?

mho22: I used this script : 

`test/script.php` 

```
<?php

symlink( __DIR__.'/store', __DIR__.'/public/store' );
```

and ran the script in the `test` folder : `php script.php`

sbc100: I see.. and presumably both of those are absolute paths?  

That means that target of the symbol is absolute (`/full/path/to/public/store`) and not relative (`../public/store`)?   Is that what you want?   Does the problem go away if you use a relative symlink?

mho22: You're right! It works perfectly with relative paths, but unfortunately, I'm dealing with absolute paths in my current situation. A situation where I am not the one who created the symlinks. Do you think it's still worth pursuing?

sbc100: Yes, this is worth fixing, I'm was just trying to figure out the set of circumstances in which this occurs.

mho22: I should try different use cases then. My suggestion probably works if the symlink path leads to a file in parent directories. But what if the symlink path leads to children directories within a parent directory. I should investigate.

mho22:
@sbc100 I ended up removing the suggestion I made. Because it didn't work for other use cases. But I could dig into code to find out where it could be problematic and tried to correct these lines with multiple use cases.

Here are the use cases I based my PR on : 

```
// USE CASE 0 RELATIVE PATH : a parent's neighbour directory in a child directory
symlink( '../relative', './public/relative' );

// USE CASE 1 RELATIVE PATH : A child directory in a parent directory
symlink( './directory/subdirectory', './public/subsubrelative' );

// USE CASE 2 RELATIVE PATH : a parent's child directory in a child's subdirectory
symlink( '../../relative/subrelative', './public/directory/subrelative' );

// USE CASE 3 ABSOLUTE PATH : a parent's neighbour directory in a child directory
symlink( __DIR__ . '/absolute', __DIR__ . '/public/absolute' );

// USE CASE 4 ABSOLUTE PATH : A child directory in a parent directory
symlink( __DIR__ . '/public/directory/subdirectory', __DIR__ . '/public/subsubabsolute' );

// USE CASE 5 ABSOLUTE PATH : a parent's child directory in a child's subdirectory
symlink( __DIR__ . '/absolute/subabsolute', __DIR__ . '/public/directory/subabsolute' );
```

My use cases are using emscripten directly within `@php-wasm/node` so those lines are `php`. I have no clue how to implement this in a test for `emscripten`.

sbc100:
> @sbc100 I ended up removing the suggestion I made. Because it didn't work for other use cases. But I dug into code to find out where it could be problematic and tried to correct these lines with multiple use cases in mind.
> 
> Here are the use cases I based my PR on :
> 
> ```
> // USE CASE 0 RELATIVE PATH : a parent's neighbour directory in a child directory
> symlink( '../relative', './public/relative' );
> 
> // USE CASE 1 RELATIVE PATH : A child directory in a parent directory
> symlink( './directory/subdirectory', './public/subsubrelative' );
> 
> // USE CASE 2 RELATIVE PATH : a parent's child directory in a child's subdirectory
> symlink( '../../relative/subrelative', './public/directory/subrelative' );
> 
> // USE CASE 3 ABSOLUTE PATH : a parent's neighbour directory in a child directory
> symlink( __DIR__ . '/absolute', __DIR__ . '/public/absolute' );
> 
> // USE CASE 4 ABSOLUTE PATH : A child directory in a parent directory
> symlink( __DIR__ . '/public/directory/subdirectory', __DIR__ . '/public/subsubabsolute' );
> 
> // USE CASE 5 ABSOLUTE PATH : a parent's child directory in a child's subdirectory
> symlink( __DIR__ . '/absolute/subabsolute', __DIR__ . '/public/directory/subabsolute' );
> ```
> 
> My use cases are using emscripten directly within `@php-wasm/node` so those lines are `php`. I have no clue how to implement this in a test for `emscripten`.

Do you know which of these 6 cases fail currently with emscripten and which succeed?

mho22:
@sbc100 without the modifications I made in this PR, the three last don't work. Because these are absolute paths. But with the modifications I added with my last commit, all of the uses cases are working.

mho22:
@sbc100 I reduced the code modification to one simple line in `src/library_nodefs.js` on line `238`: 

```diff
- path = nodePath.relative(nodePath.resolve(node.mount.opts.root), path);
+ path = nodePath.relative( PATH.isAbs(path) ? NODEFS.realPath(node.parent) : nodePath.resolve(node.mount.opts.root), path );
```

This will check if the `symlink` path is absolute or relative. If it is relative, it runs the already working code, if it is absolute, it takes the parent path instead of the root path and returns its new `relative` path.

Here is an example of this behavior with a relative path based on the structure I gave above [ Assuming this is in a `test` directory at the root of the machine ] : 

```
- test
    - public
        - directory
        - store ~symlink~
    - store
```

If `store` symlink is a relative path, it should be : `../store`

So the line above will be : 

`path = nodePath.relative( '/test', '../store' );` returning `../store`. This is actually working fine.

If `store` symlink is an absolute path, it should be : `/test/store`

This is the current behavior : 

`path = nodePath.relative( '/test', '/test/store' );` returning `store`. 

This is not working. Because it then concatenates `store` with the parent path being `/test/public` resulting in `/test/public/store` and calling again `readlink` leading to an almost indefinite loop.

So the modification above will bring this : 

`path = nodePath.relative( '/test/public', '/test/store' );` returning `../store`. 

This transforms the absolute path in a relative path that we know works completely fine.



I believe this is correct, but for some reason, a lot of tests are failing. Feel free to share any insights if any.

mho22:
@sbc100 Now it seems only this test is failing : 

```
test_embind_i64_val
```

Probably linked with the issue [#21653](https://github.com/emscripten-core/emscripten/issues/21653) you opened a month ago.

mho22:
@sbc100 Hi ! I merged the 'main' branch into the PR and every test passed. What do you think I should do next ?

src/library_nodefs.js
```diff
@@ -235,7 +235,7 @@ addToLibrary({
         var path = NODEFS.realPath(node);
         try {
           path = fs.readlinkSync(path);
-          path = nodePath.relative(nodePath.resolve(node.mount.opts.root), path);
+          path = nodePath.relative( PATH.isAbs(path) ? NODEFS.realPath(node.parent) : nodePath.resolve(node.mount.opts.root), path );
```
sbc100, line 238: I wonder if we should even be calling `nodePath.relative` at all?  What happens if you just remove this line?

sbc100:
Can you take the use cases you created above and add them to `test/unistd/links.c`.  You can then run that test with `./test/runner.py other.test_unistd_links_nodefs`

src/library_nodefs.js
mho22: Without this line, the `resolve` method which works correctly with relative paths will return wrong paths when using absolute paths. The idea with this line is to return a relative path no matter what.

sbc100: But is that how readlink is supposed to work?    If the link as absolute then shouldn't readlink return an absolute path?

mho22: @sbc100 You're correct. Removing this line resolved the issues I was experiencing.

kleisauke: It looks like issue #3222 was the reason for this line.

mho22:
@sbc100 When I run the command you gave me I get this error : 

```
⚡  emscripten [patch-1] ⚡ ./test/runner.py other.test_unistd_links_nodefs
ERROR: could not find the following tests: other.test_unistd_links_nodefs
```

I tried without the `other` term : 

```
⚡ emscripten [patch-1] ⚡ ./test/runner.py test_unistd_links_nodefs
Test suites:
['test_core']
Running test_core: (1 tests)
test_unistd_links_nodefs (test_core.core0.test_unistd_links_nodefs) ... emcc: error: emscripten setup is not complete ("npm packages" is out-of-date). Run `bootstrap` to update
None
None
FAIL

...

AssertionError: subprocess exited with non-zero return code(1): `emscripten/test/unistd/links.c -o links.js -sNO_DEFAULT_TO_CXX -Wclosure -Werror -Wno-limited-postlink-optimizations -O0 -DNODEFS -lnodefs.js
```

When I run `bootstrap` I get this :

```
⚡ emscripten [patch-1] ⚡ bootstrap
Up-to-date: npm packages
Up-to-date: create entry points
Up-to-date: git submodules
```

mho22:
@kleisauke and @sbc100 I pushed the code without that line and indeed the `test_unistd_symlink_on_nodefs` didn't pass. 

The line `21` in file `test/unistd/symlink_on_nodefs.c` : 
```
fs.symlinkSync(fs.realpathSync('./new-directory'), './symlink');
```

Inserts an absolute path in symlink. Therefore it couldn't be recognized except if you transform that absolute path in a relative path based on the mounted root in file `src/library_nodefs.js`: 

``` 
path = nodePath.relative(nodePath.resolve(node.mount.opts.root), path);
```


The `test/unistd/symlink_on_nodefs.c` test pass if the line is replaced : 

```diff
- fs.symlinkSync(fs.realpathSync('./new-directory'), './symlink');
+ fs.symlinkSync('./new-directory', './symlink');
```

Obviously. 

So the absolute path issue still remains. Except if I replace : 

```diff
- path = nodePath.relative(nodePath.resolve(node.mount.opts.root), path);
+ path = nodePath.relative( PATH.isAbs(path) ? NODEFS.realPath(node.parent) : nodePath.resolve(node.mount.opts.root), path );
```

this `readlink` method should return a relative path to make symlinks with absolute paths work. Maybe should we rename `library_nodefs` `readlink` function name to be more explicit ?

sbc100:
Having looked at `test_unistd_symlink_on_nodefs` I think the test is invalid.   Absolute symlinks to locations outside of the VFS tree should not be supported and should not be magically translated into relative symlinks.

src/library_nodefs.js
sbc100: Hmm.. from my reading of #3222 and the fix that landed with the test_unistd_symlink_on_nodefs test, i don't think it is correct. See `https://github.com/emscripten-core/emscripten/pull/21805#issuecomment-2094926239` below

mho22:
@sbc100 Ok then, I'll modify the test `test_unistd_symlink_on_nodefs` like indicated above with an absolute path from the VFS tree and remove the `path` line in file `src/library_nodefs.js` . I think I should add the tests you asked in the `test_unistd_symlink_on_nodefs` test file since it has a lot of similarities with it.

test/unistd/symlink_on_nodefs.c
```diff
@@ -16,19 +16,41 @@
...
     FS.mkdir('working');
     FS.mount(NODEFS, { root: '.' }, 'working');
 
-    FS.mkdir('direct-link');
-    FS.mount(NODEFS, { root: './symlink' }, 'direct-link');
+    FS.mkdir('direct-inside-link');
+    FS.mount(NODEFS, { root: './inside-symlink' }, 'direct-inside-link');
+
+    FS.mkdir('direct-outside-link');
+    FS.mount(NODEFS, { root: './outside-symlink' }, 'direct-outside-link');
```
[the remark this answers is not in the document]
mho22: I will copy your changes.

sbc100:
Can you update the PR title to be something like "[nodefs] Don't try to resolve absolute symlinks are outside of the VFS".

And can you mention in the PR title that this effectively reverts the behaviour that was added in #3277.

test/unistd/links.c
```diff
@@ -31,10 +31,11 @@ void setup() {
     FS.mkdir('folder');
   );
 #else
+  int fd;
   mkdir("working", 0777);
   chdir("working");
   symlink("../test/../there!", "link");
-  int fd = open("file", O_RDWR);
+  fd = open("file", O_RDWR);
```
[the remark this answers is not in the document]
mho22: @sbc100 My bad. I'm on it.

mho22:
@sbc100 I made the correction you asked, maybe did I miss something. Don't hesitate to give me other insights to improve the code readability.

test/unistd/symlink_on_nodefs.c
```diff
@@ -16,19 +16,41 @@
...
+    fs.symlinkSync("/working/absolute/subabsolute/file", "directory/subdirectory/subabsolute");
+    fs.symlinkSync("/working/directory/subdirectory/file", "subdirectoryabsolute");
 
     FS.mkdir('working');
     FS.mount(NODEFS, { root: '.' }, 'working');
 
-    FS.mkdir('direct-link');
-    FS.mount(NODEFS, { root: './symlink' }, 'direct-link');
+    FS.mkdir('direct-inside-link');
+    FS.mount(NODEFS, { root: './inside-symlink' }, 'direct-inside-link');
```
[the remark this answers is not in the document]
mho22: Corrected.

test/unistd/links.c
```diff
@@ -31,10 +31,11 @@ void setup() {
     FS.mkdir('folder');
```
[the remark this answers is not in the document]
sbc100: I would like to see some tests or absolute symlinks added to `test/unistd/links.c`.

`test/unistd/links.c` it where we run as many of the symlink tests as possible.  `symlink_on_nodefs` I think should be reserved to tests that we can't run outside of nodefs (i.e. tests that involve using non-trivial mounting setups that require nodefs to pass).

One reason that we should use `links.c` where possible is that its actually possible to run `links.c` on native desktop systems to confirm they have the same behaviour as emscripten (i.e. to check emscripten is doing the right thing). 

Perhaps `symlink_on_nodefs.c` should be renamed to something better.. but I can't think of anything.    Basically its for tests that can't run outside of nodefs.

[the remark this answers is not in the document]
mho22: @sbc100 Ok, I understand. I should probably duplicate the tests : 

```
test_relative_path_symlinks();
test_absolute_path_symlinks();
```
 from `symlink_on_nodefs.c` based on the symlink system from `links.c` in `setup()` function.
 
 Would you like me to refactor the code readability in `links.c` as well like you indicated below : 
 
 ```
 int main() {
   setup();
   test_foo_bar();
   test_baz();
   return 0;
}
```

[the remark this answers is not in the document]
Comment (sbc100) [reply test/unistd/links.c]: I think so yes, but maybe wait for #21890 to land to avoid too many conflicts.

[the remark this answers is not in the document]
Comment (sbc100) [reply test/unistd/links.c]: Actually I wonder if we even need `symlink_on_nodefs.c` at all anymore?  Since it won't really be testing anything will it?   Other than "symlinks to outside the VFS don't work" which is basically the same as "symlinks to non-existant files don't work"?  Or is it will worth keeping it around do you think?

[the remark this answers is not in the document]
Comment (mho22) [reply test/unistd/links.c]: My understanding is that this use case and this test could probably be an indicator that this is not possible and that it shouldn't be possible. So we could imagine that the only test that the `symlink_on_nodefs.c` could have are : 

```
  test_inside_symlink();
  test_outside_symlink();
```

The others will be moved to `links.c` ? 

What I find mostly important here is the proof that this has been decided.

**Comment by mho22** [comment]
@sbc100 I merged and brang the corrections you asked and I kept the `symlink_on_nodefs.c` test file with the remaining three tests : 

```
test_inside_symlink
test_outside_symlink
test_mount_link
```

What do you think about this ?

File: test/unistd/links.c [hunk 7]
```diff
@@ -59,56 +100,131 @@ int main() {
...
+    assert(fd);
+    char buffer[13] = {0};
+    rtn = fread(buffer, 1, 13, fd);
+    assert(rtn <= 13);
+    printf("buffer: '%s'\n\n", buffer);
+    fclose(fd);
+  }
+}
+
+void test_absolute_path_symlinks()
```
sbc100, line 198: Can you put the opening curly braces on the same line as the function decl?

test/unistd/symlink_on_nodefs.c
```diff
@@ -17,37 +17,57 @@
...
-  printf("buffer is '%s'\n", buffer);
+  printf("buffer: '%s'\n", buffer);
   fclose(fd);
+}
 
-  // This should fail, since it resolves to ../new-directory which is not
-  // mounted
-  fd = fopen("/direct-link/test", "r");
+void test_outside_symlink()
+{
```
sbc100, line 48: Maybe add a comment here.

```
// outside-symlink is link to an absolute path which is not part of the emscripten VFS
// and so we should be able to open it.
```

test/unistd/symlink_on_nodefs.c
```diff
@@ -17,37 +17,57 @@
...
+  FILE* fd = fopen("/working/outside-symlink/test", "r");
   assert(fd == NULL);
```
sbc100, line 50: `assert(errno == ENOENT)`

test/unistd/links.c
```diff
@@ -31,25 +31,66 @@ void setup() {
   rtn = write(fd, "test", 5);
   assert(rtn == 5);
   close(fd);
-  rtn = mkdir("folder", 0777);
+  rtn = mkdir("directory", 0777);
```
sbc100, line 34: Make we can use `makedir(name)` and `makefile(name, content)` helpers to make this setup more readable?

sbc100: (also makelink so that we can add the `assert(rtn == 0)` for those symlink calls)

sbc100 (commented):
This is looking like a really awesome change now!  Thanks for working on this.

test/unistd/links.c
mho22: Sorry, on it.

test/unistd/symlink_on_nodefs.c
mho22: On it.

test/unistd/links.c
mho22: Should I replace every `mkdir` function for a `makedir` one, `write` function for a `makefile` and a `makelink` for any `symlink` function ?

sbc100 (commented):
Could you also add an entry to to the ChangeLog, mentioning that this essentially reverts a previous change in behaviour?

mho22:
> Could you also add an entry to to the ChangeLog, mentioning that this essentially reverts a previous change in behaviour?

@sbc100 Is this changelog ok for you ? : 

```
3.1.61 - 05/08/24
-----------------
- Reverts the behaviour producing generic relative paths on symlinks.
  See ##3277 or ##3222. ()
  ```

File: test/unistd/links.c (hunk 10 above)
Comment (sbc100) [reply test/unistd/links.c:34]: Yes, that is what I was thinking.

Then it would read something like:
```
makedir("test"):
makefile("test/file", "content");
makelink("file1", "file2");
```

**Comment by sbc100** [comment]
> > Could you also add an entry to to the ChangeLog, mentioning that this essentially reverts a previous change in behaviour?
> 
> @sbc100 Is this changelog ok for you ? :
> 
> ```
> 3.1.61 - 05/08/24
> -----------------
> - Reverts the behaviour producing generic relative paths on symlinks.
>   See (#3277) or (#3222).
> ```

Can you mention that this is nodejs behaviour?

**Comment by sbc100** [comment]
> > > Could you also add an entry to to the ChangeLog, mentioning that this essentially reverts a previous change in behaviour?
> > 
> > 
> > @sbc100 Is this changelog ok for you ? :
> > ```
> > 3.1.61 - 05/08/24
> > -----------------
> > - Reverts the behaviour producing generic relative paths on symlinks.
> >   See (#3277) or (#3222).
> > ```
> 
> Can you mention that this is nodejs behaviour?

How about:

```
- Under nodefs, symbolic links to files outside of mount locations no longer work.
  This reverts the previous behaviour added in #3277. (#21805)
```

**Comment by mho22** [comment]
@sbc100 I added `changedir` to make the `setup()` method a little bit clearer. I hope this change meets your expectations.

File: ChangeLog.md [hunk 11]
```diff
@@ -524,7 +529,7 @@ See docs/process.md for more on how version tagging works.
   (#18861)
 - The `emscripten_proxy_async_with_callback` API was replaced with a simpler
   `emscripten_proxy_callback` API that takes a second callback to be called if
-  the worker thread dies before completing the proxied work.  
+  the worker thread dies before completing the proxied work.
```
sbc100, line 532: Can you revert all these unralated changes to the ChangeLog where newlines are removed?

BTW, the way I setup my editor is to (1) show trailing whitespace and (2) have a shortcut to strip it all, but not to have it do this automatically.. otherwise you end up with changes like this.

sbc100 (approved):

test/unistd/links.c
```diff
@@ -17,39 +17,70 @@
...
-#endif
-  rtn = chdir("working");
+void makedir(char *dir) {
+  int rtn = mkdir(dir, 0777);
   assert(rtn == 0);
-  symlink("../test/../there!", "link");
-  int fd = open("file", O_RDWR | O_CREAT, 0777);
+}
+
+void makefile(char *file, char *content) {
```
sbc100, line 25: Use `const char *` for both args here and args below.

sbc100 (approved):
LGTM.. just two ultra minor nits.

test/unistd/links.c
mho22: On it!

ChangeLog.md
mho22: Thanks for the insights and sorry for the inconvenience.

mho22:
@sbc100 Minor changes merged. This was a long conversation 🙂. Thank you for your time!

PR #21809, opened 2024-04-23
Don't pass `-fvisibility=default` unless building with `-fPIC`. NFC

sbc100:
This means that, by default, we will get `-fvisibility=hidden` since that is the default in clang when building for wasm.


emcc.py
```diff
@@ -386,7 +386,7 @@ def get_clang_flags(user_args):
     # enabled so that it can be added to the attributes in LLVM IR.
     if settings.SUPPORT_LONGJMP == 'wasm':
       flags.append('-mexception-handling')
-
+    #flags.append('-fwhole-program-vtables')
```
kripken, line 389: Experimental code?

emcc.py
```diff
@@ -369,12 +369,12 @@ def get_clang_flags(user_args):
   if settings.RELOCATABLE and '-fPIC' not in user_args:
     flags.append('-fPIC')
 
-  # We use default visiibilty=default in emscripten even though the upstream
-  # backend defaults visibility=hidden.  This matched the expectations of C/C++
-  # code in the wild which expects undecorated symbols to be exported to other
-  # DSO's by default.
-  if not any(a.startswith('-fvisibility') for a in user_args):
-    flags.append('-fvisibility=default')
+    # We use default visiibilty=default in emscripten even though the upstream
```
kripken, line 372:
```suggestion
    # For relocatable code we default to visibility=default in emscripten even though the upstream
```

emcc.py
```diff
@@ -369,12 +369,12 @@ def get_clang_flags(user_args):
...
+    # backend defaults visibility=hidden.  This matched the expectations of C/C++
```
kripken, line 373:
```suggestion
    # backend defaults visibility=hidden.  This matches the expectations of C/C++
```

kripken (approved):

emcc.py
```diff
@@ -369,12 +369,12 @@ def get_clang_flags(user_args):
...
-  # code in the wild which expects undecorated symbols to be exported to other
-  # DSO's by default.
-  if not any(a.startswith('-fvisibility') for a in user_args):
-    flags.append('-fvisibility=default')
+    # For relocatable code we default to visibility=default in emscripten even
+    # though the upstream backend defaults visibility=hidden.  This matches the
+    # expectations of C/C++ code in the wild which expects undecorated symbols
+    # to be exported to other DSO's by default.
+    if not any(a.startswith('-fvisibility') for a in user_args):
+      flags.append('-fvisibility=default')
```
kripken, line 377: Re-reading this now, I realize we don't get here if `-fPIC` was passed, as the condition on line 369 checks that (and not just `RELOCATABLE`). Is that ok?

sbc100: Oops, yes this should be in `if settings.RELOCATABLE`

sbc100: Done

kripken (approved):

PR #21810, opened 2024-04-23
Define unimplemented syscalls as weak symbols in stubs

squk:
With the current implementation of stubs, users can provide overrides for syscalls with existing implementations. However overriding unimplemented stubs is currently impossible. This fixes that.


sbc100 (approved):

squk:
Thanks @sbc100 !

PR #21814, opened 2024-04-23
Generate accurate import names in emitDCEGraph

sbc100:
This avoid the receiving code having to attempt to demangle them by removing leading underscore.

The other advantage of this method is that allows for JS and native names to differ in other ways (not only the leading underscore).  For example, `wasmMemory` vs `memory` and `wasmTable` vs `__indirect_function_table`.


kripken (approved):
Nice!

Issue #21081, opened 2024-01-15
JSPI asynchronous calls into WASM are much slower than Asyncify


rhashimoto:
I think this is more likely to be a Chromium bug and I have [filed a bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1518374&q=&can=4) there, but I don't understand the details well enough to know for sure.

Asynchronous calls from JavaScript into WASM are 2+ orders of magnitude slower with JSPI over Asyncify. I'm using Chrome 122.0.6248.0 (Official Build) canary (arm64) on an M2 Mac with chrome://flags/#enable-experimental-webassembly-stack-switching enabled.

I have a [test page here](https://rhashimoto.github.io/browser-test-cases/jspi/) ([source](https://github.com/rhashimoto/browser-test-cases/tree/master/jspi)) that times 10,000 calls to a C function that calls an asynchronous JavaScript function (use Chrome Canary as some older versions just crash the page). It loops over the calls once from C and once from Javascript, for both Asyncify and JavaScript builds. The output of the page looks like this:

```
asyncify 10000 iterations in C 0.01 seconds
asyncify 10000 iterations in JS 0.01 seconds
jspi 10000 iterations in C 0.003 seconds
jspi 10000 iterations in JS 3.554 seconds
```

This shows that with Asyncify, calling the JS function from C and calling it from JavaScript to C to JavaScript, is the same 0.01 seconds. JSPI is faster than Asyncify at calling the JS function from C, but is 350 times slower at calling from JS.

**Version of emscripten/emsdk:**
$ emcc -v
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.47 (431685f05c67f0424c11473cc16798b9587bb536)
clang version 18.0.0 (https://github.com/llvm/llvm-project 21030b9ab4487d845e29792063f5666d8c4b8e09)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /home/dev/emsdk/upstream/bin

**Full link command and output with `-v` appended:**
$ make (both Asyncify and JSPI builds shown here)
[...]

PR #21815, opened 2024-04-24
[jspi] Switch to new JSPI API.

brendandahl:
Differences with the new API:
 - The JS engine now takes care of the suspender, so we don't need to modify the wasm file with binaryen.
 - Imports and exports and are now marked as async with WebAssembly.Suspending and WebAssembly.promising wrappers.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 3.1.61 (in development)
 -----------------------
+- The JSPI feature now uses the updated browser API for JSPI (available in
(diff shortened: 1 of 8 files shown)


brendandahl:
We should wait for chrome unstable to include the new API (v126.0.6430) before we land this.

sbc100 (approved):

msqr1:
This would make asyncify useless, right?

brendandahl:
This PR also depends on https://github.com/WebAssembly/binaryen/pull/6546

msqr1:
It seems that we have to solve #21081 before completely going over.

PR #21820, opened 2024-04-24
[EH] Rename with_both_* decorators to with_all_*

aheejin:
Given that we are about to add a third option (new Wasm EH adopted in Oct 2023) to `with_both_eh_sjlj` and `with_both_sjlj`, this renames them to `with_all_eh_sjlj` and `with_all_sjlj`. Also this changes the suffix of `with_all_sjlj`'s Wasm tests from `wasm_sjlj` to just `wasm`.


aheejin:
I'm not very sure, but probably in the order of several months.

PR #21823, opened 2024-04-24
[test] Improve composability of test parameterization. NFC

sbc100:
Also, introduce new `@with_dylink_reversed` which triggere the need
for better composability.
    
For composabilty to work wrapper function need to be correctly marked
with `@wraps` and a helper need to be used when modifying
`_parameterized` such that existing entries are honored.


sbc100:
If you prefer I can split out the new `with_dylink_reversed` stuff

test/common.py
```diff
@@ -651,6 +647,23 @@ def find_browser_test_file(filename):
...
+
+  This function create or adds to the `_parameterize` property of a function
+  which is then expanded by the RunnerMeta metaclass into multiple separate
+  test functions.
+  """
+  prev = getattr(func, '_parameterize', None)
+  if prev:
+    # If we're parameterizing 2nd time, construct a cartesian product for various combinations.
+    func._parameterize = {
+      '_'.join(filter(None, [k1, k2])): v1 + v2 for (k1, v1), (k2, v2) in itertools.product(prev.items(), parameters.items())
```
kripken, line 661: Am I reading this right that `v1 + v2` would, in a situations with params
```
'wasm': (True,)
```
and
```
'wasm': (True,)
```
lead to a result of `(True,) + (True,)` - ? I assume not, but I'm not sure how else to read this.

sbc100: This code is only being moved, I didn't write it, or try to understand it fully as part of this PR.

I do know that it is working in the expected way though.

sbc100: `(True,) + (True,)` makes sense to me. That would produce `(True, True)` as you might expect.   Those would then be passed as arguments to the outer wrapper, which would iterpret the first arguement and then pass the rest into the inner wrapper, which would then interpret the second argument and pass the empty list of argument to final/real test function.

kripken: Thanks, it does produce `(True, True)`... well, I never knew how tuple math works in Python I guess :smile: But why does that make sense? I'd expect lane-by-lane addition, so
```
(True,)  + (True,) = (True + True,) = (2,)
```
(I verified that `True + True = 2` in Python... which I also did not expect, to be honest!)

kripken (approved):

Issue #21827, opened 2024-04-25
3.1.58 breaks usage in Node.js worker


toyobayashi:
minimum repro:

https://github.com/toyobayashi/emscripten-node-worker-3158-repro
https://github.com/toyobayashi/emscripten-node-worker-3158-repro/actions/runs/8827891977

issue 1: Cannot find correct `.wasm` path when run emscripten generated JavaScript in a Node.js Worker.
issue 2: The empty `*.worker.js` is written in `cwd`, not the same directory where `.js` exists.
issue 3: Pthread worker will read incorrect `.wasm` file again even if I specified `locateFile` when `-sSTRICT=1`

sbc100:
Is this perhaps a duplicate of https://github.com/emscripten-core/emscripten/issues/21826?

The core misunderstand there seems to be that pthreads need to be able to locate the wasm file.  In fact pthreads recieve the wasm module via postMessage so they should never be trying to locate or read the wasm file themselves.

I believe the issue that caused some confusion in #21826 was that locateFile is still being called on worker startup, even though the worker is never using the result of that call.    I think I will update the code so workers never actaully call locateFile to avoid this confusion.

toyobayashi:
not sure whether issue 3 is duplicated, but I confirm 1 and 2 are not related to #21826.

I mean run emscripten js in my own Node.js worker (the worker is emscripten's main thread), please take a look the repro code and CI test, it does break something.

toyobayashi:
Issue 1 has nothing to do with locateFile, after I experienced issue 1, I tried to use locateFile to solve it then experienced issue 3.

sbc100:
So we have a fix for (2) in #21830 and (3) in #21829.

That just leaves (1) which still needs some investigation.

sbc100:
Final fix is in #21832.  I will push for a new release sooner rather than later to get these fixes out.

Issue #21826, opened 2024-04-24
In `v3.1.58`, why pthreads don't inherit the `moduleArg`?


DenizUgur:
I'm not sure if this was an oversight but with the latest version, pthreads don't use the same `moduleArg` that was given to the main module. I'm relying on this argument to supply hashed version of my wasm file. Is there any other suggested way to achieve this?

https://github.com/emscripten-core/emscripten/blob/e5f1e343f827b477058e462e840b00b53478c1a8/tools/link.py#L2458

sbc100:
Can you share more details of your use case?  In particular I'm not sure how the `moduleArg` would be transmitted to the worker.  

How were you populating the `moduleArg`?  Do you know how it was visible to the workers?

sbc100:
`mainScriptUrlOrBlob` should still be supported.    See: https://github.com/emscripten-core/emscripten/blob/7c31c12cfc03f1130a63e379f6858f82e4337ca8/src/library_pthread.js#L458-L466

So there must be some way this usage is different and not working for your use case.   Hopefully we can figure it and add a test for that technique.

DenizUgur:
What would be the suggested way of setting wasm binary name during runtime? Also how do you enable `mainScriptUrlOrBlob` because I don't see it in my build output.

DenizUgur:
I think I'll use a combination of `sessionStorage` and `locateFile`. Not sure what would be the most elegant way to do this natively with Emscripten but if you believe this is a valid use-case then I think it's worth looking into.

sbc100:
Pthreads don't need to locate the wasm file because they all get the wasm module delivered via postMessage on startup.  i.e. the module is shared from the main thread.

`mainScriptUrlOrBlob` tells the system is used to the control the name/content for the main JavaScript script itself.

If `mainScriptUrlOrBlob` is not showing up in your output it could be you are overriding `INCOMING_MODULE_JS_API`?  Or maybe you are using `-sSTRICT` with makes `INCOMING_MODULE_JS_API` empty by default?

sbc100:
I don't understand this comment: https://github.com/ffmpegwasm/ffmpeg.wasm/blob/ae1cdac7db79c5315f9a1b716fcbd9bcfe27c902/src/bind/ffmpeg/bind.js#L93-L96,  It seems to suggest that each worker needs to locate the wasm file.. but IIUC workers have always recieved the wasm module via postMessage so have never needed to know its location.

[...]

PR #21829, opened 2024-04-25
Skip wasmBinaryFile initialization on pthread workers

sbc100:
The fact that we were calling `locateFile` in an attempt to find the wasm binary on each worker was confusing at best.  The wasm file is always received via postMessage when running on the worker.

See #21826 and #21827


sbc100:
Moving this code does seems to have some tiny effects on code size.. some positive but mostly minutely negative.   Since this code is simply moved from top level to the inside of a function, with zero changes, I can only assume that is slightly confusing closure compiler.

An alternative to move this code would be to instead wrap it in `if (!ENVIRONMENT_IS_PTHREAD && !ENVIRONMETN_IS_WORKER)`

sbc100:
I tried looking at the generated code (after closure compiler was applied) both before and after this change but I'm struggling to come with any rationale for why it might regress.

I took one same regression which is `other.test_metadce_mem_O3_standalone_narg` which regressed from 4554 to 4580 (24 bytes or 0.5%).

I've attached the result of clang-formatting the resulting JS files.

Can you see where the differences are coming from.
[compare.zip](https://github.com/emscripten-core/emscripten/files/15113195/compare.zip)

sbc100:
I did some refactoring, and put this code into a function that can do early returns instead of branches.  The code size effects are now almost all wins (especially with gz).

kripken:
Thanks, after adding some whitespace it is easier to flip back and forth between the files to see the differences,

https://gist.github.com/kripken/286edabc866138d05716c2efbbd1cad4

The main issue (aside from code moving around, which does not affect size) is that in the new code the locateFile usage remains in a function, where it has a `return`, and there is a call to it. Before it was inlined, saving those bytes.

I see no reason for it not to be inlined in the new code. Perhaps closure is more careful about inlining into nested scopes. Or maybe it is because of the condition right before it - a little hard to see what that was before closure, but maybe you have some idea there.

If not this is probably ok, but it's worth trying a bit to avoid a regression here I think.

kripken:
Ah, our last two comments raced. Your new code lgtm, and if gzip size is mostly a win then that seems fine.

sbc100:
Perhaps because createWasm is already quote a large function closure chose not to inline.   Or perhaps the inlining rules for top level code are differen to function-scope code?

Anyway, after extracting the function the code size effects seem much better.

PR #21840, opened 2024-04-26
Ensure all unused elements are removed during applyDCEGraphRemovals. NFC

sbc100:
There were bugs in `applyDCEGraphRemovals` that were preventing it from finding certain unused exports in certain cases.

This change cleanup up the pass and adds assertions that all unused imports and exports are actually located and removed, thus preventing these types of bugs from sneaking in again.

The test we had for `applyDCEGraphRemovals` were worked, but they were writing in a slightly different style to the current emscripten output. In particular the export name and JS name were matching, even though the actual compiler output produces JS names that contain a leading underscore.  I also updates the tests to match this style for consistency.

I believe the effects of this bug were not captured by our code size tests because they all run closure compiler afterwards which was removing the unused exports itself.


tools/acorn-optimizer.mjs
```diff
@@ -978,67 +979,84 @@ function emitDCEGraph(ast) {
 // way (and we leave further DCE on the JS and wasm sides to their respective
 // optimizers, closure compiler and binaryen).
 function applyDCEGraphRemovals(ast) {
-  const unused = new Set(extraInfo.unused);
+  const unused_exports = new Set(extraInfo.unused_exports);
```
kripken, line 982:
```suggestion
  const unusedExports = new Set(extraInfo.unusedExports);
```
I think this would be more consistent, though I don't feel strongly. I can see an argument for the JSON style being different than JS.

tools/acorn-optimizer.mjs
```diff
@@ -978,67 +979,84 @@ function emitDCEGraph(ast) {
...
         }
-      }
+        return true;
+      });
     } else if (node.type === 'VariableDeclaration') {
-      // Handle the case we declare a variable but don't assign to the module:
-      //   var x = wasmExports['x'];
-      // and
-      //   var x = () => (x = wasmExports['x']).apply(...);
+      // Handle the various ways in which extract wasmExports:
```
kripken, line 1002:
```suggestion
      // Handle the various ways in which we extract wasmExports:
```

tools/extract_metadata.py
```diff
@@ -241,13 +241,10 @@ def get_main_reads_params(module, export_map):
...
 def get_global_exports(module, exports):
   global_exports = {}
-  internal_start_stop_symbols = set(['__start_em_asm', '__stop_em_asm',
-                                     '__start_em_lib_deps', '__stop_em_lib_deps',
-                                     '__em_lib_deps'])
   internal_prefixes = ('__em_js__', '__em_lib_deps')
   for export in exports:
     if export.kind == webassembly.ExternType.GLOBAL:
-      if export.name in internal_start_stop_symbols or any(export.name.startswith(p) for p in internal_prefixes):
+      if any(export.name.startswith(p) for p in internal_prefixes):
```
kripken, line 247: How is this part of the change related?

sbc100: This is needed since we use the set of global exports here:

https://github.com/emscripten-core/emscripten/blob/7f369e1e9f27af3385471bb00314a73d467d72d6/tools/building.py#L751-L752

Without this the list of unused exports we pass also includes some of these globals, which would then cause the assertion to fire.

Since this is the only place the `WASM_GLOBAL_EXPORTS` is used I think its fine.

sbc100: I factored out `is_internal_global` here which I think makes things more clear.

kripken (approved):

PR #21853, opened 2024-04-29
Note newly enabled features in LLVM in ChangeLog

aheejin:
https://github.com/llvm/llvm-project/pull/80923 enabled multivalue and reference-types by default. Even though this is an LLVM change, given that we release emsdk with all LLVM/Binaryen/Emscripten versions, it could be worth mentioning in the ChangeLog.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -24,6 +24,9 @@ See docs/process.md for more on how version tagging works.
   of pthread builds so that is generated alongside the main JavaScript file.
   See #21701. ()
 - `-sASYNCIFY=2` is setting now deprecated, use `-sJSPI` instead.
+- LLVM now enables multivalue and reference-types features by default. Enabling
+  multivalue feature only enables the language feature and does not turn on the
+  multivalue ABI. (https://github.com/llvm/llvm-project/pull/80923)
 
 3.1.58 - 04/23/24
 -----------------
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -24,9 +24,10 @@ See docs/process.md for more on how version tagging works.
   of pthread builds so that is generated alongside the main JavaScript file.
   See #21701. ()
 - `-sASYNCIFY=2` is setting now deprecated, use `-sJSPI` instead.
-- LLVM now enables multivalue and reference-types features by default. Enabling
-  multivalue feature only enables the language feature and does not turn on the
-  multivalue ABI. (https://github.com/llvm/llvm-project/pull/80923)
+- Due to the upstream LLVM change
+  (https://github.com/llvm/llvm-project/pull/80923), multivalue and
+  reference-types features are now enabled by default in Emscripten. This only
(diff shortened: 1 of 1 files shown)


ChangeLog.md
```diff
@@ -24,6 +24,9 @@ See docs/process.md for more on how version tagging works.
   of pthread builds so that is generated alongside the main JavaScript file.
   See #21701. ()
 - `-sASYNCIFY=2` is setting now deprecated, use `-sJSPI` instead.
+- LLVM now enables multivalue and reference-types features by default. Enabling
```
sbc100, line 27: Perhaps phrase this as something that emscripten now supports?   Maybe "Due to and upstream llvm change, the x and y features are not enabled by default in emscripten"?

aheejin: Tried doing that.

kripken (approved):

kleisauke:
Should we also update `MIN_NODE_VERSION` to `170200` since reference-types requires that version?
https://github.com/emscripten-core/emscripten/blob/63c648fa17be49fa9640d4a40c8ba2f7a9b2d444/src/settings.js#L1884

sbc100:
> Should we also update `MIN_NODE_VERSION` to `170200` since reference-types requires that version?
> 
> https://github.com/emscripten-core/emscripten/blob/63c648fa17be49fa9640d4a40c8ba2f7a9b2d444/src/settings.js#L1884

Good point.  Maybe we should just go to 18 since that is LTS.

Do all our min browser versions support reference types?

kleisauke:
> Do all our min browser versions support reference types?

Ah, it looks like `MIN_CHROME_VERSION` and `MIN_SAFARI_VERSION` should also be adjusted to `96` and `150000`, respectively.
https://github.com/emscripten-core/emscripten/blob/63c648fa17be49fa9640d4a40c8ba2f7a9b2d444/src/settings.js#L1877
https://github.com/emscripten-core/emscripten/blob/63c648fa17be49fa9640d4a40c8ba2f7a9b2d444/src/settings.js#L1867

See: https://caniuse.com/wasm-reference-types

sbc100:
> > Do all our min browser versions support reference types?
> 
> Ah, it looks like `MIN_CHROME_VERSION` and `MIN_SAFARI_VERSION` should also be adjusted to `96` and `150000`, respectively.
> 
> https://github.com/emscripten-core/emscripten/blob/63c648fa17be49fa9640d4a40c8ba2f7a9b2d444/src/settings.js#L1877
> 
> 
> https://github.com/emscripten-core/emscripten/blob/63c648fa17be49fa9640d4a40c8ba2f7a9b2d444/src/settings.js#L1867
> 
> See: https://caniuse.com/wasm-reference-types

Yeah, I'm not totally sure we need/want to do this.  Especially since we are not going to actually be generating any reference types by default.

aheejin:
I'll then land this for now.

PR #21857, opened 2024-04-30
Add openal alext.h header

### LINKED ISSUE #21723: Include `alext.h` header for AL
The [docs claim support for some OpenAL extensions](https://emscripten.org/docs/porting/Audio.html#supported-openal-extensions), however I was a bit confused to find the `alext.h` missing when trying to utilize the `ALC_HRTF_SOFT` extension. Based on [this test](https://github.com/emscripten-core/emscripten/blob/main/test/openal_playback.cpp#L123), it looks as if the define value must be used directly in order to use the extension, which is a little difficult to figure out, to me.

Is there a reason to not add the `alext.h` file found in the [openal-soft include folder](https://github.com/kcat/openal-soft/blob/master/include/AL/alext.h) to the "emscripten/system/include/AL" folder? I believe this would also alleviate issues like #9851 and would render #9852 obsolete.

As of writing I see that the discussion in #9852 suggests doing exactly this, but I figured I'd make this issue just to make the potential solution even clearer for anyone willing to revive the effort to remedy this (I may do so myself, if I find the time)

gabeklavans:
I couldn't find mention of the version of openal-soft that al.h and alc.h were pulled from, so I somewhat arbitrarily pulled alext.h from openal-soft v1.19.1. Let me know if those versions are recorded somewhere.
 
resolves #21723

### DIFF
--- a/system/include/AL/alext.h
+++ b/system/include/AL/alext.h
@@ -1,0 +1,477 @@
+/**
+ * OpenAL cross platform audio library
+ * Copyright (C) 2008 by authors.
+ * This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Library General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *  Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU Library General Public
+ *  License along with this library; if not, write to the
+ *  Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ * Or go to http://www.gnu.org/copyleft/lgpl.html
+ */
+
+#ifndef AL_ALEXT_H
+#define AL_ALEXT_H
+
+#include <stddef.h>
+/* Define int64_t and uint64_t types */
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
+#include <inttypes.h>
+#elif defined(_WIN32) && defined(__GNUC__)
+#include <stdint.h>
+#elif defined(_WIN32)
+typedef __int64 int64_t;
+typedef unsigned __int64 uint64_t;
+#else
+/* Fallback if nothing above works */
+#include <inttypes.h>
+#endif
+
+#include "alc.h"
+#include "al.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef AL_LOKI_IMA_ADPCM_format
+#define AL_LOKI_IMA_ADPCM_format 1
+#define AL_FORMAT_IMA_ADPCM_MONO16_EXT           0x10000
+#define AL_FORMAT_IMA_ADPCM_STEREO16_EXT         0x10001
+#endif
+
+#ifndef AL_LOKI_WAVE_format
+#define AL_LOKI_WAVE_format 1
+#define AL_FORMAT_WAVE_EXT                       0x10002
+#endif
+
+#ifndef AL_EXT_vorbis
+#define AL_EXT_vorbis 1
+#define AL_FORMAT_VORBIS_EXT                     0x10003
+#endif
+
+#ifndef AL_LOKI_quadriphonic
+#define AL_LOKI_quadriphonic 1
+#define AL_FORMAT_QUAD8_LOKI                     0x10004
+#define AL_FORMAT_QUAD16_LOKI                    0x10005
+#endif
+
+#ifndef AL_EXT_float32
+#define AL_EXT_float32 1
+#define AL_FORMAT_MONO_FLOAT32                   0x10010
+#define AL_FORMAT_STEREO_FLOAT32                 0x10011
+#endif
+
+#ifndef AL_EXT_double
+#define AL_EXT_double 1
+#define AL_FORMAT_MONO_DOUBLE_EXT                0x10012
+#define AL_FORMAT_STEREO_DOUBLE_EXT              0x10013
+#endif
+
+#ifndef AL_EXT_MULAW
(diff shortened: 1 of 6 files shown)


system/include/AL/alext.h
```diff
@@ -0,0 +1,477 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *  Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU Library General Public
+ *  License along with this library; if not, write to the
+ *  Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ * Or go to http://www.gnu.org/copyleft/lgpl.html
+ */
```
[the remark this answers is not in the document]
sbc100: Yes, you should leave this here.

What modifications did you make exactly?

test/openal_extensions.c
```diff
@@ -0,0 +1,71 @@
+// So far just does some sanity checks for the available extensions.
```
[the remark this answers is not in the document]
sbc100: Yup, you can copy the one from any of the other test files.

test/openal_extensions.c
```diff
@@ -0,0 +1,71 @@
...
+static void check_al_extension(const ALchar *extension) {
+    if (alIsExtensionPresent(extension) != ALC_TRUE) {
+        fprintf(stderr, "Extension %s was not present.", extension);
+        result = EXIT_FAILURE;
+    }
+}
+
+int main() {
+
+    for(int i = 0; i < NUM_ALC_EXTENSIONS; i++) {
```
sbc100, line 62: We use 2 spaces for indentation and space after keywords like `for`

test/test_browser.py
```diff
@@ -2312,6 +2312,9 @@ def test_openal_error(self, args):
   def test_openal_capture_sanity(self):
     self.btest('openal_capture_sanity.c', expected='0')
 
+  def test_openal_extensions(self):
+    self.btest('openal_extensions.c', expected='0')
```
sbc100, line 2316: Can you use `best_exit`?  This means you can remove the `REPORT_RESULT` stuff from the test and remove the expected argument here (since the default is zero).

test/openal_extensions.c
```diff
@@ -0,0 +1,71 @@
+// So far just does some sanity checks for the available extensions.
+//
+// In the future, could also try to enable different extensions and test for
+// output correctness.
+
+#include <stdio.h>
+#include <stdlib.h>
+#ifdef __EMSCRIPTEN__
+#include <emscripten.h>
```
sbc100, line 9: Is this include needed?

test/openal_extensions.c
```diff
@@ -0,0 +1,71 @@
...
+
+    for(int i = 0; i < NUM_ALC_EXTENSIONS; i++) {
+        check_alc_extension(alc_extensions[i]);
+    }
+
+    for(int i = 0; i < NUM_AL_EXTENSIONS; i++) {
+        check_al_extension(al_extensions[i]);
+    }
+
+    end_test();
```
sbc100, line 70: You can just `return 0` here if you use `btest_exti`

test/openal_extensions.c
```diff
@@ -0,0 +1,71 @@
...
+    if (alcIsExtensionPresent(device, extension) != ALC_TRUE) {
+        fprintf(stderr, "Extension %s was not present.", extension);
+        result = EXIT_FAILURE;
+    }
+}
+
+static void check_al_extension(const ALchar *extension) {
+    if (alIsExtensionPresent(extension) != ALC_TRUE) {
+        fprintf(stderr, "Extension %s was not present.", extension);
+        result = EXIT_FAILURE;
```
sbc100, line 56: Can you turn all these into calls to `assert` instead?

sbc100 (commented):
Thanks!

gabeklavans:
@sbc100 
> What modifications did you make exactly?

I removed a bunch of typedefs that were causing errors, as well as removed the import of efx.h

system/include/AL/alext.h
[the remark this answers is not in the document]
gabeklavans: I removed a bunch of typedefs that were causing errors, as well as removed the import of efx.h

test/openal_extensions.c
gabeklavans: Unsure, copied from openal_capture_sanity.c, since I saw it in a couple of files. I can remove and see if the test passes.

gabeklavans:
@sbc100 it looks like I'm failing some tests. I'm not exactly sure how to parse the output, but it seems it might have something to do with size? Is the alext.h perhaps too large?

sbc100:
It looks like maybe your branch is not up-to-date with the current main branch?  Can you try rebasing or merging?

sbc100 (approved):

sbc100:
The test failure you are seeing here is unrelated and should be fixed in #21885

sbc100:
It was pointed out that this header is licences under the BSD licence and so far all our runtime code is MIT licenced.   So I think we will have to revert this and create our own version of this header outselves.   @gabeklavans would you be up for doing this?   If you can do it in the next few days we can maybe avoid reverting this PR

gabeklavans:
> It was pointed out that this header is licences under the BSD licence and so far all our runtime code is MIT licenced.   So I think we will have to revert this and create our own version of this header outselves.   @gabeklavans would you be up for doing this?   If you can do it in the next few days we can maybe avoid reverting this PR   

How's this sound: if I don't get it done by Friday 11:00 PM EST you can go ahead and revert. I'm not sure if I'll be able to finish it up but I can give it a shot.

Additionally, what would constitute our own version? I'd want to keep all the API-side defines the same as what is documented for OpenAL soft. So I suppose if I just start the header from scratch and only define what we need, would that suffice?

sbc100:
Yes, something like this: 38a151b96c640d81bdeea41428fa509ebebea003

gabeklavans:
@sbc100 just for a thorough notification, I have https://github.com/emscripten-core/emscripten/pull/22349 up ready for review.

PR #21858, opened 2024-04-30
Update library_glfw.js - to receive backspace and tab inputs

kunyoungparkk:
in web application, i couldn't receive backspace input because of this code block.

### DIFF
--- a/src/library_glfw.js
+++ b/src/library_glfw.js
@@ -418,9 +418,10 @@ var LibraryGLFW = {
       // This logic comes directly from the sdl implementation. We cannot
       // call preventDefault on all keydown events otherwise onKeyPress will
       // not get called
-      if (event.keyCode === 8 /* backspace */ || event.keyCode === 9 /* tab */) {
-        event.preventDefault();
-      }
+      // However, this code block prevents web applications from receiving backspace inputs.
+      //if (event.keyCode === 8 /* backspace */ || event.keyCode === 9 /* tab */) {
+      //  event.preventDefault();
+      //}
     },
 
     onKeyup: (event) => {


src/library_glfw.js
```diff
@@ -418,9 +418,10 @@ var LibraryGLFW = {
       // This logic comes directly from the sdl implementation. We cannot
       // call preventDefault on all keydown events otherwise onKeyPress will
       // not get called
-      if (event.keyCode === 8 /* backspace */ || event.keyCode === 9 /* tab */) {
-        event.preventDefault();
-      }
+      // However, this code block prevents web applications from receiving backspace inputs.
```
sbc100, line 421: But isn't that idea in fact that the key events are being received by the GLFW code?    Do you want both GLFW and some other listener (e.g. the containing JS code?) receive the same key events?

kunyoungparkk: Yes, that's right. We're creating a program that displays the results of the rendering work with glfw and controls it even in the input tag on the web page. And even if the design was based on the assumption that input was received only on one side, it's hard to understand that only backspace and tabs were blocked from bubbling.

kunyoungparkk: Thank you for letting me know about this issue. I think the person who designed the code wanted to have that key input applied only in the part that was ported with sdl or glfw. But I think this should be controlled in the JavaScript part. It's true that these are libraries that have a role in handling events, but I don't think it's right for them to monopolize a particular event on the webpage. If we keep it this way, even if there's no focus on the ported part (probably canvas), it blocks the event. It may not be a problem with a correct answer, but anyway there's a problem where the input tag is not available at all as of now. And most of all, I'm not sure why exactly this part is necessary. Wouldn't it be better to have scalability?

sbc100: Thats intersting.. i would have expected this capturing of the event to only occur when the canvas is focused.  Doing otherwise certainly seems odd.

kripken: I think it might make sense to make this more controllable. Ignoring backspace and tab is probably the best behavior for games, say in Doom where you hold down tab to show the map, then if it also moved between DOM items on the page you'd lose focus on the canvas etc. But in an application that also has some DOM text processing on the page then obviously backspace might be needed.

We probably don't want to change the default here, as that could break people, but an option for this seems reasonable to me. However, we should unify this across all the relevant code, which is at least SDL1, SDL2, and GLFW (and maybe others, we should check). Once the code is refactored to a single shared place then an option to control it sounds good to me.

PR #21863, opened 2024-04-30
Ensure compatibility with older versions of Node.js

### LINKED ISSUE #21118: NODEFS and NODERAWFS broken since v3.1.45 with node < v18.3.0
Hi,

In https://github.com/emscripten-core/emscripten/pull/20077 the functions fs.readSync and fs.writeSync have been updated with signatures that did not exist in node v10.19.0 (for instance fs.writeSync with options was only added in node v18.3.0), which breaks running with any earlier version of node.

The four occurrences should really be rewritten as follows for them to work as expected as from node v10.19.0:

```js
return fs.readSync(stream.nfd, new Int8Array(buffer.buffer, offset, length), 0, length, position);
return fs.writeSync(stream.nfd, new Int8Array(buffer.buffer, offset, length), 0, length, position);
var bytesRead = fs.readSync(stream.nfd, new Int8Array(buffer.buffer, offset, length), 0, length, position);
var bytesWritten = fs.writeSync(stream.nfd, new Int8Array(buffer.buffer, offset, length), 0, length, position);
```

Thanks a lot for the fix!

Kind regards,
Maurice

kleisauke:
The previous `fs.writeSync` API was not available in Node.js 17.x, see:
https://nodejs.org/api/fs.html#fswritesyncfd-buffer-options

Also, do the same for `fs.readSync` to ensure compat with Node.js < v13.13.0.

Resolves #21118.

---

Noticed this while trying to update wasm-vips minimum required Node.js version to 17.2.0 (commit https://github.com/kleisauke/wasm-vips/commit/9fc633883c84bbb0a75dee28fef4b298be05faa7 - in preparation for https://github.com/llvm/llvm-project/pull/80923).


src/library_nodefs.js
```diff
@@ -276,14 +276,22 @@ addToLibrary({
         // Node.js < 6 compatibility: node errors on 0 length reads
         if (length === 0) return 0;
         try {
-          return fs.readSync(stream.nfd, new Int8Array(buffer.buffer, offset, length), { position: position });
+#if MIN_NODE_VERSION < 131300
+          return fs.readSync(stream.nfd, new Int8Array(buffer.buffer, offset, length), 0, length, position);
```
sbc100, line 280: Do newer versions of node also support this old pattern?  If so, can we not just use this older pattern everwhere?  Is there any downside?  It looks like it creates less gabage since you don't need to create a new option object each call.

If newer versions don't support this pattern then I suppose we would want a runtime check rather than a compile time one.

kleisauke: Newer versions of Node.js ought to support this too. I guarded this for code size reasons, but perhaps that is negligible.

kleisauke: Commit 65a6ff350972343da0cf6b2a0d72e1e2af6707e4 prefers the old API instead.

sbc100 (approved):

sbc100:
Can we add some test to verify that fixes the issue.   (See `test-node-compat` in `.circleci/config.yml` where we run some tests under node 10.19.0)

sbc100:
> > Can we add some test to verify that fixes the issue. (See `test-node-compat` in `.circleci/config.yml` where we run some tests under node 10.19.0)
> 
> I tried to verify this locally with Node.js v10.19.0 but it failed on the optional chaining (`?.`) operator.
> 
> ```
> $ ./test/runner core0.test_fs_js_api core0.test_fs_write
> Test suites:
> ['test_core']
> Running test_core: (2 tests)
> Using 2 parallel test processes
> -- begin program output --
> /tmp/emtest_qrar9kcz/emscripten_test_core0_26cak6rw/test_fs_js_api.js:649
>   Module['monitorRunDependencies']?.(runDependencies);
>                                    ^
> 
> SyntaxError: Unexpected token .
>     at Module._compile (internal/modules/cjs/loader.js:723:23)
>     at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
>     at Module.load (internal/modules/cjs/loader.js:653:32)
>     at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
>     at Function.Module._load (internal/modules/cjs/loader.js:585:3)
>     at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
>     at startup (internal/bootstrap/node.js:283:19)
>     at bootstrapNodeJSCore (internal/bootstrap/node.js:623:3)
> -- end program output --
> test_fs_js_api (test_core.core0.test_fs_js_api) ... FAIL
> -- begin program output --
> /tmp/emtest_mmgpbjvl/emscripten_test_core0_8rucs0x_/test_write.js:575
>   Module['monitorRunDependencies']?.(runDependencies);
>                                    ^
> 
> SyntaxError: Unexpected token .
>     at Module._compile (internal/modules/cjs/loader.js:723:23)
>     at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
>     at Module.load (internal/modules/cjs/loader.js:653:32)
>     at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
>     at Function.Module._load (internal/modules/cjs/loader.js:585:3)
>     at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
>     at startup (internal/bootstrap/node.js:283:19)
>     at bootstrapNodeJSCore (internal/bootstrap/node.js:623:3)
> -- end program output --
> test_fs_write (test_core.core0.test_fs_write) ... FAIL
> 
> ...
> ```
> 
> FWIW, I could also reproduce this with the `core2.test_hello_world` test.
> 
> https://github.com/emscripten-core/emscripten/blob/63c648fa17be49fa9640d4a40c8ba2f7a9b2d444/.circleci/config.yml#L764

The test framework should automatically inject `-sMIN_NODE_VERSION=..` in this case which should cause tranpilation (and removal of the optional chaining).   See `get_emcc_node_flags` in `building.py`

kleisauke:
Commit 99b0518cce1fd1f97495047b1fb2b4bcdff282d9 adds a test case for this. I had to modify the unit test to prevent an `Invalid typed array length`-error, which appeared because the `Uint8Array` did not account for the null byte in its size calculation.

<details>
  <summary>Details</summary>

```console
$ grep NODE_JS_TEST .emscripten 
NODE_JS_TEST = [os.path.expanduser('~/Downloads/node-v10.19.0-linux-x64/bin/node')]
JS_ENGINES = [NODE_JS_TEST, V8_ENGINE]
$ ./test/runner core0.test_fs_write_rawfs 
Test suites:
['test_core']
Running test_core: (1 tests)
test_fs_write_rawfs (test_core.core0.test_fs_write_rawfs) ... (test did not pass in JS engine: ['/home/kleisauke/Downloads/node-v10.19.0-linux-x64/bin/node'])
FAIL

FAIL: test_fs_write_rawfs (test_core.core0.test_fs_write_rawfs)
  File "/home/kleisauke/emscripten/test/common.py", line 710, in resulting_test
    return func(self, *args)
           ^^^^^^^^^^^^^^^^^
  File "/home/kleisauke/emscripten/test/test_core.py", line 160, in metafunc
    func(self)
  File "/home/kleisauke/emscripten/test/test_core.py", line 5829, in test_fs_write
    self.do_run_in_out_file_test('fs/test_write.cpp')
  File "/home/kleisauke/emscripten/test/common.py", line 1692, in do_run_in_out_file_test
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kleisauke/emscripten/test/common.py", line 1743, in _build_and_run
    self.assertContained(o, js_output, regex=regex)
  File "/home/kleisauke/emscripten/test/common.py", line 1416, in assertContained
AssertionError: Expected to find 'read Hello! 1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890
' in 'read 
', diff:

--- expected
+++ actual
@@ -1 +1,2 @@
-read Hello! 1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890
+read
+
$ git checkout old-nodejs-compat
M	.circleci/config.yml
M	test/fs/test_write.cpp
M	test/test_core.py
Switched to branch 'old-nodejs-compat'
$ ./test/runner core0.test_fs_write_rawfs 
Test suites:
['test_core']
Running test_core: (1 tests)
test_fs_write_rawfs (test_core.core0.test_fs_write_rawfs) ... ok

Ran 1 test in 1.125s

OK
$ git restore test/fs/test_write.cpp
$ ./test/runner core0.test_fs_write_rawfs 
Test suites:
['test_core']
Running test_core: (1 tests)
test_fs_write_rawfs (test_core.core0.test_fs_write_rawfs) ... -- begin program output --
/home/kleisauke/emscripten/out/test/test_write.js:113
      throw ex;
      ^

RangeError: Invalid typed array length: 101
    at new Int8Array (<anonymous>)
    at write (/home/kleisauke/emscripten/out/test/test_write.js:2405:49)
    at Object.write (/home/kleisauke/emscripten/out/test/test_write.js:4985:16)
    at Object.79964 (/home/kleisauke/emscripten/out/test/test_write.js:840:8)
    at runEmAsmFunction (/home/kleisauke/emscripten/out/test/test_write.js:4465:26)
    at _emscripten_asm_const_int (/home/kleisauke/emscripten/out/test/test_write.js:4468:10)
    at wasm-function[16]:73
    at wasm-function[100]:3
    at /home/kleisauke/emscripten/out/test/test_write.js:578:12
    at callMain (/home/kleisauke/emscripten/out/test/test_write.js:5075:15)
    at doRun (/home/kleisauke/emscripten/out/test/test_write.js:5113:23)
    at run (/home/kleisauke/emscripten/out/test/test_write.js:5125:5)
    at runCaller (/home/kleisauke/emscripten/out/test/test_write.js:5065:19)
    at removeRunDependency (/home/kleisauke/emscripten/out/test/test_write.js:515:7)
    at receiveInstance (/home/kleisauke/emscripten/out/test/test_write.js:694:5)
    at receiveInstantiationResult (/home/kleisauke/emscripten/out/test/test_write.js:712:5)
-- end program output --
FAIL
...
```
</details>

test/fs/test_write.cpp
```diff
@@ -24,7 +24,7 @@ int main()
         var pos = lengthBytesUTF8(str);
 
         str = '1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890';
-        data = new Uint8Array(100);
+        data = new Uint8Array(101);
```
sbc100, line 27: Can you do `str.length` there instead maybe?

sbc100 (approved):

test/fs/test_write.cpp
kleisauke: 1da12e4f7acf8a885ce199d98b136cbc3002e823

sbc100 (approved):

curiousdannii:
Node 17 is EOL. We shouldn't go out of our way to support it. But, these changes are very simple, so I guess why not.

sbc100:
> Node 17 is EOL (and has been for almost 2 years). We shouldn't go out of our way to support it. But, these changes are very simple, so I guess why not.

We support versions of node much older than 17.   Our support goes as far back as far as 10.19.0!

PR #21867, opened 2024-04-30
Allow EXPORTED_FUNCTIONS to include and export JS library symbols

sbc100:
Prior to this change one would need to add JS library symbols to both `DEFAULT_LIBRARY_FUNCS_TO_INCLUDE` and `EXPORTED_FUNCTIONS` to achieve this (confusingly with different name mangling for each).

This is a step on the way to #8380.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 3.1.60 (in development)
 -----------------------
+- The `EXPORTED_FUNCTIONS` list can now include JS library symbols even if they
+  have not been otherwise included (e.g. via `DEFAULT_LIBRARY_FUNCS_TO_INCLUDE`).
+  (#21867)
 
 3.1.59 - 04/30/24
 -----------------
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -7234,7 +7234,6 @@ int main(int argc, char** argv) {
     self.run_process([EMCC, 'main.c',
                       '--js-library=lib.js',
                       '-sMAIN_MODULE=2',
-                      '-sDEFAULT_LIBRARY_FUNCS_TO_INCLUDE=foo,bar',
                       '-sEXPORTED_FUNCTIONS=_main,_foo,_bar'])
 
     # Fist test the successful use of a JS function with dlsym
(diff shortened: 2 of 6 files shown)


ChangeLog.md
```diff
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 3.1.60 (in development)
 -----------------------
+- The `EXPORTED_FUNCTIONS` list can now include JS library symbols even if they
+  have not been otherwise include (e.g. via `DEFAULT_LIBRARY_FUNCS_TO_INCLUDE`).
```
kripken, line 24:
```suggestion
  have not been otherwise included (e.g. via `DEFAULT_LIBRARY_FUNCS_TO_INCLUDE`).
```

test/test_other.py
```diff
@@ -13533,7 +13540,7 @@ def test_stdint_limits(self):
 
   def test_legacy_runtime(self):
     self.set_setting('EXPORTED_FUNCTIONS', ['_malloc', '_main'])
-    self.set_setting('DEFAULT_LIBRARY_FUNCS_TO_INCLUDE', ['$intArrayFromString', '$ALLOC_NORMAL'])
+    self.set_setting('DEFAULT_LIBRARY_FUNCS_TO_INCLUDE', ['$intArrayFromString'])
```
kripken, line 13543: This was testing that we can include either a function or a constant, I think - why remove it?

tools/link.py
```diff
@@ -3103,6 +3102,8 @@ def add_js_deps(sym):
           add_js_deps(sym)
         for sym in settings.EXPORTED_RUNTIME_METHODS:
           add_js_deps(shared.demangle_c_symbol_name(sym))
+        for sym in settings.EXPORTED_FUNCTIONS:
+          add_js_deps(shared.demangle_c_symbol_name(sym))
```
kripken, line 3106: Do the native methods not cause any problems here?

sbc100: No, `add_js_deps` here is really `maybe_add_js_deps` since it only does anyting if the symbol is actually found in the js_syms map.

test/test_other.py
sbc100: I was thinking because its listed below in `EXPORTED_RUNTIME_METHODS` its no longer needed here (i.e. its duplicated here).     But I can revert this line.

tools/link.py
kripken: I see. I guess we never have overlap between the two sets of names since native ones are prefixed, so that seems safe for now. But if we remove the prefixing later maybe it would be an issue?

kripken (approved):

tools/link.py
sbc100: It should still be fine.  This pass here is just adding any native deps for JS symbols we might want.  Its conservative, so the worst case would be that don't end up need the extra exports (i.e. if a native symbol satisfies the requirement the the JS symbol is not actually needed)  and metadce would remove them.

kripken: Sounds good.

PR #21870, opened 2024-05-01
Add @requires_webgpu, make WebGPU tests fail if unavailable

kainino0x:
**WIP, needs rebase over #21799**

This makes these tests safer so they won't just silently pass if you happen to run against a browser/system where WebGPU isn't available (currently including Chrome on Linux and Firefox/Safari).

Verified these 5 tests fail, using:
`EMTEST_BROWSER='/Applications/Google\ Chrome\ Canary.app/Contents/MacOS/Google\ Chrome\ Canary --disable-gpu' test/runner.py 'browser.test_webgpu*'`
(`--disable-gpu` has the side effect of disabling WebGPU since we don't ship a software fallback in Chrome right now.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -368,6 +368,7 @@ commands:
             # TODO: Do GL testing when https://bugzil.la/1375585 (lack of WebGL
             #       support in headless mode) resolves
             EMTEST_LACKS_GRAPHICS_HARDWARE: "1"
+            EMTEST_LACKS_WEBGPU: "1"
             EMTEST_LACKS_SOUND_HARDWARE: "1"
             # OffscreenCanvas support is not yet done in Firefox.
             EMTEST_LACKS_OFFSCREEN_CANVAS: "1"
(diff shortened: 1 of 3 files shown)


sbc100 (approved):

.circleci/config.yml
```diff
@@ -798,6 +798,8 @@ jobs:
           test_targets: "other skip:other.test_native_link_error_message"
   test-browser-chrome:
     executor: bionic
+    environment:
+      EMTEST_LACKS_WEBGPU: "1"
```
sbc100, line 802: Does this mean that some tests that previously ran under chrome will no longer run?  Or are all such tests already ones that didn't actually test anything?

i.e. are there some webgpu browser tests that don't actually requires a real GPU to run?

kainino0x: `test_webgpu_basic_rendering*` and `test_webgpu_required_limits` previously skipped themselves if a WebGPU adapter was not available. Now they fail in that case.

`test_webgpu_get_device*` do not require a GPU, it actually doesn't touch the WebGPU API at all, it only tests some Emscripten code.

sbc100: In that case I wonder if we should move `test_webgpu_get_device` to `test_other.py` so it doesn't depend on the browser?    Otherwise won't we loose that coverage?

kainino0x: I didn't add `@requires_webgpu` to those tests, so they should still run.

If we moved them to `test_other.py` how would we run them, under Node?

sbc100: Sounds good!

(yes other tests run under either node or d8)

PR #21875, opened 2024-05-01
[test] Parameterize some more tests. NFC

sbc100:
I also removed the test_opt_levels as it didn't seem to be testing anything unique.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -955,22 +955,22 @@ f.close()
   # Tests that it's possible to pass C++11 or GNU++11 build modes to CMake by building code that
   # needs C++11 (embind)
   @requires_ninja
-  def test_cmake_with_embind_cpp11_mode(self):
-    for args in [[], ['-DNO_GNU_EXTENSIONS=1']]:
-      self.clear()
-      # Use ninja generator here since we assume its always installed on our build/test machines.
-      configure = [EMCMAKE, 'cmake', '-GNinja', test_file('cmake/cmake_with_emval')] + args
-      print(str(configure))
-      self.run_process(configure)
-      build = ['cmake', '--build', '.']
-      print(str(build))
-      self.run_process(build)
+  @parameterized({
+    '': [[]],
+    '_no_gnu': [['-DNO_GNU_EXTENSIONS=1']],
+  })
+  def test_cmake_with_embind_cpp11_mode(self, args):
+    # Use ninja generator here since we assume its always installed on our build/test machines.
+    configure = [EMCMAKE, 'cmake', '-GNinja', test_file('cmake/cmake_with_emval')] + args
+    self.run_process(configure)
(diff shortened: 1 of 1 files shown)


test/test_other.py
```diff
@@ -4771,18 +4779,13 @@ def test(cxx, no_exit, assertions, flush=0, keepalive=0, filesystem=1):
...
-      print(args)
-      self.run_process([EMXX, test_file('fs_after_main.cpp')])
-      self.assertContained('Test passed.', self.run_js('a.out.js'))
-
-  def test_opt_levels(self):
-    for opt in ['-O1', '-O2', '-Os', '-Oz', '-O3', '-Og', '-Ofast']:
-      print(opt)
-      proc = self.run_process([EMCC, '-v', test_file('hello_world.c'), opt], stderr=PIPE)
-      self.assertContained(opt, proc.stderr)
-      self.assertContained('hello, world!', self.run_js('a.out.js'))
```
kripken, line 4785: Looks like this test was removed entirely? We may lose coverage for `-Ofast` due to that (`-Og` appears in another test outside of this diff).

test/test_other.py
```diff
@@ -8779,37 +8783,41 @@ def test_side_module_folder_deps(self):
     self.run_process([EMCC, test_file('hello_world.c'), '-sMAIN_MODULE', '-o', 'main.js', '-L', 'subdir', '-lside2'])
 
   @is_slow_test
-  def test_lto(self):
+  @parameterized({
+    '01': (['-O1'],),
```
kripken, line 8787: Looks like we lost the no-arguments case? Is LTO without opts not worth testing I guess?

test/test_other.py
sbc100: Yeah, I figured we have lots of testing for opt levels elsewhere already, but perhaps you are right that `-Ofast` and `-Og` are not well tested otherwise.

test/test_other.py
```diff
@@ -4779,6 +4779,13 @@ def test(cxx, no_exit, assertions, flush=0, keepalive=0, filesystem=1):
           for flush in [0, 1]:
             test(cxx, no_exit, assertions, flush)
 
+  def test_opt_levels(self):
+    for opt in ['-O1', '-O2', '-Os', '-Oz', '-O3', '-Og', '-Ofast']:
```
kripken, line 4783:
```suggestion
    for opt in ['-Og', '-Ofast']:
```

kripken (approved):
lgtm % suggestion

test/test_other.py
kripken: Hmm, perhaps this test could be renamed something like `test_rare_opt_levels` to just give coverage of the rare ones? (that don't have lots of other coverage) I think then it would be self-consistent.

kripken (approved):

Issue #21881, opened 2024-05-02
Compilation result is not minified with -O3 and LEGACY_VM_SUPPORT


ekharkunov:
When use `-O3` along with `LEGACY_VM_SUPPORT` or any of `MIN_FIREFOX_VERSION`/`MIN_SAFARI_VERSION`/etc. options to enable support of ols browsers - result javascript file isn't minified.
If compile with `-O3` option - everything is ok. Result is minified.

**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.55 (165133b1cc977f0b9a277e42ef809b823157189c)
clang version 19.0.0git (https:/github.com/llvm/llvm-project 6c7805d5d186a6d1263f90b8033ad85e2d2633d7)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /Users/yauheni/work/defold/tmp/dynamo_home/ext/SDKs/emsdk-3.1.55/upstream/bin
```

**Failing command line in full:**
```
yauheni@workstation test-emscripten % ~/work/defold/tmp/dynamo_home/ext/SDKs/emsdk-3.1.55/upstream/emscripten/emcc -v -O3 -sLEGACY_VM_SUPPORT=1 main.c
 /Users/yauheni/work/defold/tmp/dynamo_home/ext/SDKs/emsdk-3.1.55/upstream/bin/clang -target wasm32-unknown-emscripten -fignore-exceptions -fvisibility=default -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr --sysroot=/Users/yauheni/work/defold/tmp/dynamo_home/ext/SDKs/emsdk-3.1.55/upstream/emscripten/cache/sysroot -DEMSCRIPTEN -Werror=implicit-function-declaration -Xclang -iwithsysroot/include/fakesdl -Xclang -iwithsysroot/include/compat -v -O3 main.c -c -o /var/folders/ym/b9r0ldbx2_7c94bph_0rpw8m0000gn/T/emscripten_temp_woqn2y23/main_0.o
clang version 19.0.0git (https:/github.com/llvm/llvm-project 6c7805d5d186a6d1263f90b8033ad85e2d2633d7)
Target: wasm32-unknown-emscripten
Thread model: posix
[...]

PR #21882, opened 2024-05-02
Ensure we minify whitespace, even when transpiling using babel

sbc100:
We could instead rely try to configure babel do this, but it seems more consistent to simply run our existing minifier again.

Fixes: #21881

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -2228,11 +2228,15 @@ def phase_binaryen(target, options, wasm_target):
     if options.use_closure_compiler:
       with ToolchainProfiler.profile_block('closure_compile'):
         final_js = building.closure_compiler(final_js, extra_closure_args=options.closure_args)
-      save_intermediate_with_wasm('closure', wasm_target)
+      save_intermediate('closure')
+
     if settings.TRANSPILE:
       with ToolchainProfiler.profile_block('transpile'):
         final_js = building.transpile(final_js)
-      save_intermediate_with_wasm('traspile', wasm_target)
+      save_intermediate('transpile')
+      # Run acorn one more time to minify whitespace after babel runs
+      if settings.MINIFY_WHITESPACE:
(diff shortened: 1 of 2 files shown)


tools/link.py
```diff
@@ -2228,11 +2228,15 @@ def phase_binaryen(target, options, wasm_target):
...
         final_js = building.closure_compiler(final_js, extra_closure_args=options.closure_args)
-      save_intermediate_with_wasm('closure', wasm_target)
+      save_intermediate('closure')
+
     if settings.TRANSPILE:
       with ToolchainProfiler.profile_block('transpile'):
         final_js = building.transpile(final_js)
-      save_intermediate_with_wasm('traspile', wasm_target)
+      save_intermediate('traspile')
+      # Run acron one more time to minify whitespace after babel runs
```
kripken, line 2237:
```suggestion
      # Run acorn one more time to minify whitespace after babel runs
```

tools/link.py
```diff
@@ -2228,11 +2228,15 @@ def phase_binaryen(target, options, wasm_target):
...
       with ToolchainProfiler.profile_block('closure_compile'):
         final_js = building.closure_compiler(final_js, extra_closure_args=options.closure_args)
-      save_intermediate_with_wasm('closure', wasm_target)
+      save_intermediate('closure')
+
     if settings.TRANSPILE:
       with ToolchainProfiler.profile_block('transpile'):
         final_js = building.transpile(final_js)
-      save_intermediate_with_wasm('traspile', wasm_target)
+      save_intermediate('traspile')
```
kripken, line 2236:
```suggestion
      save_intermediate('transpile')
```

kripken (approved):
+1 for not letting Babel do it, as it may minify whitespace in some other manner that could end up confusing when comparing builds.

Issue #21883, opened 2024-05-02
[emrun] emrun replaces `_` with `-` for the arguments that should be passed to the program


orhun-sketchup:
emrun replaces `_` with `-` for the arguments after the `--`. In my case it causes gtest to fail with unrecognized command line arguments.

**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.56 (cf90417346b78455089e64eb909d71d091ecc055)
clang version 19.0.0git (https:/github.com/llvm/llvm-project 34ba90745fa55777436a2429a51a3799c83c6d4c)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: E:\src\csp\3e7dbf\1\bin\upstream\bin

**Failing command line in full:**
`emrun test.html -- --gtest_output=xml:/home/web_user/test.xml`

PR #21884, opened 2024-05-02
Fix emrun arg handling

sbc100:
Fixes: #21883


emrun.py
```diff
@@ -1569,14 +1569,13 @@ def parse_args(args):
 
   # Support legacy argument names with `_` in them (but don't
   # advertize these in the --help message.
```
kripken, line 1571:
```suggestion
  # advertize these in the --help message).
```

kripken: (might as well fix a typo while we are here)

kripken (approved):

PR #21885, opened 2024-05-02
Fix regal build after recent clang change. NFC

sbc100:
There appears to be a typo in Regal's linear.h:

```
linear.h:585:26: error: no member named 'negate' in 'Vec4<T>'; did you mean 'Negate'?
  585 |       Vec4 rv(*this); rv.negate(); return rv;
      |                          ^~~~~~
      |                          Negate
linear.h:534:11: note: 'Negate' declared here
  534 |     void  Negate() {
```

This was not an issue until https://github.com/llvm/llvm-project/pull/90152 landed.

Adding `-fdelayed-template-parsing` seems to fix the issue.

### DIFF
--- a/tools/ports/regal.py
+++ b/tools/ports/regal.py
@@ -107,7 +107,8 @@ def get(ports, settings, shared):
       '-I' + source_path_lookup3,
       '-I' + source_path_boost,
       '-Wno-deprecated-register',
-      '-Wno-unused-parameter'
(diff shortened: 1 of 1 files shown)


sbc100:
Separately I wonder if we should just delete the regal port: https://github.com/nigels-com/regal/issues/4

kripken (approved):

PR #21889, opened 2024-05-03
asyncLoad: fix empty error message

Kojoley:
Before:
```
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".] {
}
```

After:
```
[Error: ENOENT: no such file or directory, open 'A:\path\to\foo.wasm'] {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'A:\\path\\to\\foo.wasm'
}
```

### DIFF
--- a/src/library.js
+++ b/src/library.js
@@ -3125,11 +3125,11 @@ addToLibrary({
 #endif
       onload(new Uint8Array(arrayBuffer));
       if (dep) removeRunDependency(dep);
-    }, (event) => {
+    }, (err) => {
       if (onerror) {
-        onerror();
+        onerror(err);
       } else {
-        throw `Loading data file "${url}" failed.`;
+        throw err;
       }
     });
     if (dep) addRunDependency(dep);


sbc100 (approved):
Thanks!

src/library.js
```diff
@@ -3125,11 +3125,11 @@ addToLibrary({
 #endif
       onload(new Uint8Array(arrayBuffer));
       if (dep) removeRunDependency(dep);
-    }, (event) => {
+    }, (err) => {
       if (onerror) {
-        onerror();
+        onerror(err);
```
kripken, line 3130: I looked at the first few users of this and didn't see any that use the parameter - where did you see it used? Or was this for an out-of-tree usage?

src/library.js
```diff
@@ -3125,11 +3125,11 @@ addToLibrary({
...
       } else {
-        throw `Loading data file "${url}" failed.`;
+        throw err;
```
kripken, line 3132: Perhaps it makes sense to keep the text here? The context could be nice. I'm not sure if there's a way to append more text to an existing Error object, though...

src/library.js
sbc100: @Kojoley could you elaborate on what API you were using when you saw the difference above?    How where you using it exactly?

Kojoley:
```sh
mkdir link_test
echo ; > link_test.cpp
call emcc -fPIC -sSIDE_MODULE -Wall -g -c "link_test.cpp" -o "link_test.o"
call emcc -fPIC -sSIDE_MODULE "link_test.o" -o "link_test/foo.wasm"

echo int main() { return 0; } > main.cpp
call emcc -fPIC -sMAIN_MODULE -Wall -g -c "main.cpp" -o "main.o"
call emcc -fPIC -sMAIN_MODULE "main.o" "link_test/foo.wasm" -o "link_test.js"

node link_test.js
```

src/library.js
Kojoley: Look at the OP/commit message, readAsync already has the accurate error information, this does not add any additional information, quite an opposite, it has less information.

kripken: Interesting. So the text "Loading data file .. failed" was swallowed somewhere, I guess? Anyhow, sounds good that this improves things.

src/library.js
kripken: Do you know where in that output it uses the parameter? Or is it hard to figure out? (I'd hope you can add `console.log(new Error().stack)` right where the parameter is received, to show the stack trace.)

Kojoley: I'm not the person that can comfortably debug that for you, sorry. I don't write Javascript nor have a clue about emscripten internals. It took me a while already to get to this point.

sbc100: Thanks for the repro case.  I think this change is good but I will do a little more investigation here before landing so I (we) fully understand what is going on here.

sbc100: When I run the above code I do now see a useful error:

```
$ node link_test.js
node:fs:596
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, open '/usr/local/google/home/sbc/dev/wasm/emscripten/foo.wasm'
    at Object.openSync (node:fs:596:3)
    at Object.readFileSync (node:fs:464:35)
    at readAsync (/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:91:16)
    at asyncLoad (/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:1781:31)
    at loadLibData (/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:1869:18)
    at getExports (/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:1889:18)
    at loadDynamicLibrary (/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:1906:16)
    at /usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:1947:11 {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/usr/local/google/home/sbc/dev/wasm/emscripten/foo.wasm'
}

Node.js v18.20.3
```

So maybe this bug no longer exists?

sbc100:
It looks like that has been fixed by some other means.   I can reproduce this with 3.1.57:

```
$ node link_test.js
/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:128
      throw ex;
      ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".] {
}

Node.js v18.20.3
```

But not with tot:

```
$ node link_test.js
node:fs:596
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, open '/usr/local/google/home/sbc/dev/wasm/emscripten/foo.wasm'
    at Object.openSync (node:fs:596:3)
    at Object.readFileSync (node:fs:464:35)
    at readAsync (/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:91:16)
    at asyncLoad (/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:1781:31)
    at loadLibData (/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:1869:18)
    at getExports (/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:1889:18)
    at loadDynamicLibrary (/usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:1906:16)
    at /usr/local/google/home/sbc/dev/wasm/emscripten/link_test.js:1947:11 {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/usr/local/google/home/sbc/dev/wasm/emscripten/foo.wasm'
}

Node.js v18.20.3
```

src/library.js
Kojoley: I had moved on a long time ago.

PR #5847, opened 2017-11-25
Use read bit to get execute bit on Windows

saschanaz:
Currently NODEFS propagates write bit to get execute bit but this causes #5268 where `C:/Users` has permission 444. This PR uses read bit instead to fix this.

Fixes #5268 
Closes #5819

PR #21902, opened 2024-05-07
[nodefs] Fix execute permission bit on Windows when retrieving mode attribute

fluiddot:
On Windows, files and directories created within a mounted `NODEFS` instance consistently have the execution permission bit `x` disabled, even though they are actually accessible. The permissions returned consistently indicate `0666`. If this value is used to create other files or directories to match the same permissions, they will possess a different value, potentially leading to inconsistencies.

This PR applies a solution similar to https://github.com/emscripten-core/emscripten/pull/5847, albeit within the `getattr` function of `NODEFS`.


test/test_other.py
```diff
@@ -14269,6 +14269,44 @@ def test_windows_batch_file_dp0_expansion_bug(self):
...
+      assert((s.st_mode & S_IXUSR) == S_IXUSR);
+      assert((s.st_mode & S_IXGRP) == S_IXGRP);
+      assert((s.st_mode & S_IXOTH) == S_IXOTH);
+
+      err = stat("/working/new-dir/test.txt", &s);
+      assert((s.st_mode & S_IXUSR) == S_IXUSR);
+      assert((s.st_mode & S_IXGRP) == S_IXGRP);
+      assert((s.st_mode & S_IXOTH) == S_IXOTH);
+    }
+    '''.replace('{{WORKING_DIR}}', Path(self.working_dir).as_posix())
```
fluiddot, line 14306: We need to convert the format of the `working_dir` path to POSIX, otherwise, it doesn't mount NODEFS in the right location.

fluiddot: This issue is similar to what's described in Node's `path` reference [Windows vs. POSIX](https://nodejs.org/api/path.html#windows-vs-posix).

src/library_nodefs.js
```diff
@@ -127,6 +127,11 @@ addToLibrary({
         if (NODEFS.isWindows && !stat.blocks) {
           stat.blocks = (stat.size+stat.blksize-1)/stat.blksize|0;
         }
+        if (NODEFS.isWindows) {
```
sbc100, line 130: Can you combine this with the block above so we only check `isWindows` once?

test/test_other.py
```diff
@@ -14269,6 +14269,44 @@ def test_windows_batch_file_dp0_expansion_bug(self):
     create_file('build_with_quotes.bat',  f'@"emcc" {test_file("hello_world.c")}')
     self.run_process(['build_with_quotes.bat'])
 
+  @only_windows('Check that directory permissions are properly retrieved on Windows')
+  @requires_node
+  def test_windows_nodefs_execution_permission(self):
+    if self.get_setting('WASMFS'):
+      self.set_setting('FORCE_FILESYSTEM')
+    delete_dir('new-dir')
```
sbc100, line 14277: You don't need to do this here since tests are guaranteed to start with an empty dir

test/test_other.py
```diff
@@ -14269,6 +14269,44 @@ def test_windows_batch_file_dp0_expansion_bug(self):
...
+      assert((s.st_mode & S_IXOTH) == S_IXOTH);
+
+      err = stat("/working/new-dir/test.txt", &s);
+      assert((s.st_mode & S_IXUSR) == S_IXUSR);
+      assert((s.st_mode & S_IXGRP) == S_IXGRP);
+      assert((s.st_mode & S_IXOTH) == S_IXOTH);
+    }
+    '''.replace('{{WORKING_DIR}}', Path(self.working_dir).as_posix())
+    self.emcc_args += ['-lnodefs.js']
+    self.do_run(src)
```
sbc100, line 14308: You can combine these two lines into `self.do_run(src, emcc_args=['-lnodefs.js'])`

test/test_other.py
```diff
@@ -14269,6 +14269,44 @@ def test_windows_batch_file_dp0_expansion_bug(self):
...
+    src = r'''
+    #include <assert.h>
+    #include <emscripten.h>
+    #include <sys/stat.h>
+    #include <string.h>
+
+    int main(int argc, char * argv[]) {
+      EM_ASM(
+        FS.mkdir('/working');
+        FS.mount(NODEFS, { root: '{{WORKING_DIR}}' }, '/working');
```
sbc100, line 14287: I think you can just do `{ root: '.' }` here?  Unless I'm missing something?

test/test_other.py
```diff
@@ -14269,6 +14269,44 @@ def test_windows_batch_file_dp0_expansion_bug(self):
...
+        FS.mkdir('/working/new-dir');
+        FS.writeFile('/working/new-dir/test.txt', 'test');
+      );
```
sbc100, line 14290: Can you put this `EM_ASM` block into a separate `setup()` function?

test/test_other.py
```diff
@@ -14269,6 +14269,44 @@ def test_windows_batch_file_dp0_expansion_bug(self):
     create_file('build_with_quotes.bat',  f'@"emcc" {test_file("hello_world.c")}')
     self.run_process(['build_with_quotes.bat'])
 
+  @only_windows('Check that directory permissions are properly retrieved on Windows')
+  @requires_node
+  def test_windows_nodefs_execution_permission(self):
+    if self.get_setting('WASMFS'):
```
sbc100, line 14275: I think this will always be false unless you add `@also_with_wasmfs`.  I'm not sure if you want to do that here though.

test/test_other.py
```diff
@@ -14269,6 +14269,44 @@ def test_windows_batch_file_dp0_expansion_bug(self):
...
+      self.set_setting('FORCE_FILESYSTEM')
```
sbc100, line 14276: Is this needed?  Maybe just remove these two lines?

test/test_other.py
```diff
@@ -14269,6 +14269,44 @@ def test_windows_batch_file_dp0_expansion_bug(self):
...
+      err = stat("/working/new-dir", &s);
+      assert(S_ISDIR(s.st_mode));
+      assert((s.st_mode & S_IXUSR) == S_IXUSR);
+      assert((s.st_mode & S_IXGRP) == S_IXGRP);
+      assert((s.st_mode & S_IXOTH) == S_IXOTH);
+
+      err = stat("/working/new-dir/test.txt", &s);
+      assert((s.st_mode & S_IXUSR) == S_IXUSR);
+      assert((s.st_mode & S_IXGRP) == S_IXGRP);
+      assert((s.st_mode & S_IXOTH) == S_IXOTH);
```
sbc100, line 14304: I think you can just do `assert(s.st_mode & S_IXOTH)` since each of those defines is just a single bit

src/library_nodefs.js
```diff
@@ -127,6 +127,11 @@ addToLibrary({
...
+          // Node.js on Windows never represents permission bit 'x', so
+          // propagate read bits to execute bits.
+          stat.mode = stat.mode | (stat.mode & 292) >> 2;
```
sbc100, line 133: Can you put brackets around the `(stat.mode & 292) >> 2` so its clear that the shift only applies the part of this expression?

sbc100 (commented):
lgtm % some nits and tests updates

test/test_other.py
fluiddot: I added this part based on what I saw in other tests ([example](https://github.com/fluiddot/emscripten/blob/fa80cf25b8641974d60605b8a24cb7150fd63701/test/test_core.py#L5788-L5795)). Not exactly sure if it's needed but we could remove it if it's not 👍.

test/test_other.py
fluiddot: As shared [here](https://github.com/emscripten-core/emscripten/pull/21902#discussion_r1593703949), probably not. I'll remove it.

src/library_nodefs.js
fluiddot: I noticed you updated the code where this is based in https://github.com/emscripten-core/emscripten/pull/21904. I'll apply the same here. Thanks 🙇 !

test/test_other.py
fluiddot: Ah, true. I think I introduced this early on when I was running the test without using the test runner. Now using the test runner, I confirmed that using `{ root: '.' }` works as expected. I'll update it.

src/library_nodefs.js
fluiddot: Done in https://github.com/emscripten-core/emscripten/pull/21902/commits/75abe47bf4c8a12f0fb3d8ed7d1bf70518bb9b21.

fluiddot:
@sbc100 thank you so much for reviewing the PR so quickly 🙇 ! I've checked the feedback and pushed changes accordingly. Let me know if you could take another look, thanks!

sbc100 (approved):
Thanks for working on this!

sbc100 (approved):

PR #20704, opened 2023-11-14
emmalloc: Fix allocations of high alignment

kripken:
Take the alignment into account when requesting new memory from the OS. If the
alignment is high then we need to ask for enough to ensure we end up aligned, or
else we can end up allocating double the memory we need (see #20645).

This only changes the amount we allocate from the OS if the alignment is non-
standard, that is, this is NFC for normal calls to `malloc`.

Also remove an assertion that limited the maximum alignment. mimalloc wants
4 MB alignment.

Fixes #20654

PR #21905, opened 2024-05-07
[mimalloc] Remove no-longer-needed limit on emmalloc alignment

kripken:
#20704 removed the maximum alignment limit from emmalloc, so we no longer
need it in the mimalloc port. Without it, we can properly use the 4GB-aligned pages
that mimalloc requests.

### DIFF
--- a/system/lib/mimalloc/src/prim/emscripten/prim.c
+++ b/system/lib/mimalloc/src/prim/emscripten/prim.c
@@ -1,5 +1,5 @@
 /* ----------------------------------------------------------------------------
-Copyright (c) 2018-2023, Microsoft Research, Daan Leijen
+Copyright (c) 2018-2023, Microsoft Research, Daan Leijen, Alon Zakai
 This is free software; you can redistribute it and/or modify it under the
 terms of the MIT license. A copy of the license can be found in the file
 "LICENSE" at the root of this distribution.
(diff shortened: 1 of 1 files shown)


system/lib/mimalloc/src/prim/emscripten/prim.c
```diff
@@ -78,17 +78,10 @@ int _mi_prim_alloc(size_t size, size_t try_alignment, bool commit, bool allow_la
...
   *is_zero = false;
-  // emmalloc has some limitations on alignment size.
-  // TODO: Why does mimalloc ask for an align of 4MB? that ends up allocating
-  //       8, which wastes quite a lot for us in wasm. If that is unavoidable,
-  //       we may want to improve emmalloc to support such alignment. See also
-  //       https://github.com/emscripten-core/emscripten/issues/20645
+  // emmalloc has a minimum alignment size.
   #define MIN_EMMALLOC_ALIGN           8
-  #define MAX_EMMALLOC_ALIGN (1024*1024)
   if (try_alignment < MIN_EMMALLOC_ALIGN) {
```
sbc100, line 83: Can try_alignment ever be lower than 8 in mimalloc?

sbc100 (approved):

system/lib/mimalloc/src/prim/emscripten/prim.c
kripken: I think no, since it makes that assumption internally. But we could probably make it do the clamping itself silently, I suppose.

kripken: Oh sorry, you said mimalloc and I read that as emmalloc. I'm not sure. The initial allocations are definitely 4MB-sized, but there may be other incidental allocations in theory in the codebase somewhere.

PR #21922, opened 2024-05-09
Factor out try/cartch duplication in nodefs code. NFC

sbc100:


src/library_wasmfs_node.js
```diff
@@ -54,39 +64,35 @@ addToLibrary({
...
-        type = 2;
-      } else if (entry.isSymbolicLink()) {
-        type = 3;
-      } else {
-        type = 0;
-      }
-      __wasmfs_node_record_dirent(vec, name, type);
-      stackRestore(sp);
+    return wasmfsTry(() => {
+      let entries = fs.readdirSync(path, { withFileTypes: true });
```
kripken, line 76: Should the try be only on this line? It looks like now it includes the lines after as well.

kripken: Is that not worth it? In general I think we've preferred such scopes to be as limited as possible, to avoid the risk of unrelated errors being caught.

sbc100: The handler here just rethrow stuff that doesn't have a `code` so it wouldn't be a big issue I think,

kripken: Fair enough, I don't feel strongly.

kripken (approved):

PR #21926, opened 2024-05-09
Test reading from stdin in shell environments

sbc100:
Also refactor FS_stdin_getChar so that it only includes the shell
code when needed.

I verified that this test runs under spidermonkey and v8.


kripken (approved):

sbc100:
Some codesize wins here due to shell code not getting includes anymore by default!

PR #21932, opened 2024-05-13
JSPI - JSPI_<EXPORTS/IMPORTS> setting and docs update.

brendandahl:
- Replace `ASYNCIFY_IMPORTS` and `ASYNCIFY_EXPORTS` with `JSPI_IMPORTS` and `JSPI_EXPORTS`.
- Better document JSPI support.

### DIFF
--- a/site/source/docs/porting/asyncify.rst
+++ b/site/source/docs/porting/asyncify.rst
@@ -1,23 +1,31 @@
 .. _asyncify section:
 
-========
-Asyncify
-========
+=================
+Asynchronous Code
+=================
 
-Asyncify lets **synchronous** C or C++ code interact with **asynchronous**
-JavaScript. This allows things like:
+Emscripten supports two ways (Asyncify and JSPI) that let **synchronous** C or
+C++ code interact with **asynchronous** JavaScript. This allows things like:
 
  * A synchronous call in C that yields to the event loop, which
    allows browser events to be handled.
  * A synchronous call in C that waits for an asynchronous operation in JS to
    complete.
 
-Asyncify automatically transforms your compiled code into a form that can be
-paused and resumed, and handles pausing and resuming for you, so that it is
-asynchronous (hence the name "Asyncify") even though you wrote it in a normal
-synchronous way.
 
-See the
+In general the two options are very similar, but rely on different underlying
+mechanisms to work.
+
+  * `Asyncify` - Asyncify automatically transforms your compiled code into a
+    form that can be paused and resumed, and handles pausing and resuming for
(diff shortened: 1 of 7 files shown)


test/test_other.py
```diff
@@ -3204,7 +3204,7 @@ def test_embind_return_value_policy(self):
 
   def test_jspi_wildcard(self):
     self.require_jspi()
-    self.emcc_args += ['-sASYNCIFY_EXPORTS=async*']
+    self.emcc_args += ['-sJSPI_EXPORTS=async*']
```
sbc100, line 3207: Is this the only place we use this in the whole test suite?

sbc100 (approved):

test/test_other.py
brendandahl: There are three others. I've left them though so we still test that option until we remove it. I suppose I could just leave one, if that's preferred.

sbc100: Yes, for deprecated settings I'd prefer that all tests use the new thing.. and we have a single test for the deprecated one (can be version of an existing test).

sbc100:
Can you add settings that are actually deprecated to `DEPRECATED_SETTINGS` in `settings.py`?

site/source/docs/porting/asyncify.rst
```diff
@@ -332,8 +366,25 @@ In this example, a function "func" is called which returns a Number.
...
+
+Differences Between Asyncify and JSPI
+#####################################
+
+Besides using different underlying mechanisms, Asyncify and JSPI also handle
+async imports and exports differently. Asyncify will automatically determine
+what exports will become async based on what could potentially call an
+an async import (``ASYNCIFY_IMPORTS``). However, with JSPI, the async imports
+and exports must be explicitly set using ``JSPI_IMPORTS`` and ``JSPI_EXPORTS``
+settings.
```
kripken, line 378: Is it not possible to compute exports in the JSPI case so that users don't need to specify them explicitly?

brendandahl: It's possible and we started down that path in this [PR](https://github.com/WebAssembly/binaryen/pull/5211). I think the big difference with JSPI, is all "JSPI'ed" exports return a promise even if they don't ever suspend. If we do the auto approach that could lead to exports returning a promise where it's not expected. I think this worked okay with asyncify since often the export didn't actually suspend, so the library code worked as expected.

We could do some hybrid approach, where we check to see what exports we think will suspend and if they're not in the list error out or give the user a list of suspending exports.

kripken: I see, thanks, makes sense.

PR #6402, opened 2018-03-30
[WIP] enable accessing member variables without copying

froydnj:
Accessing member variables by copying, especially for non-trivial member
variable types, is inefficient and results in peculiar behavior when
using such members from JavaScript.  Let's add an alternative property
definition method, `ref_property`, that can be used to return certain
fields "by reference".

You lose the ability to assign *from* members that you declare with `ref_property`--at least as far as idiomatic JS goes, you might be able to use your knowledge that the member actually comes out as a pointer and fish around in Emscripten internals to dereference it.  I'm not sure if assigning from such members is an important feature, or whether it should simply be prominently warned against in the documentation.

Fixes #3551.

Issue #17573, opened 2022-08-04
Embind fails to update the property of an object if the object is a property of another object


sebastiendu:
Modifying the value of a C++ class instance property via Embind fails silently when this instance is the property of another class instance:

```C++
class A {
public:
    int i = 42;
};

class B {
public:
    A a;
};

EMSCRIPTEN_BINDINGS(test) {
    using namespace emscripten;
    class_<A>("A").constructor().property("i", &A::i);
    class_<B>("B").constructor().property("a", &B::a);
}

void workingReferenceCFunction() {
    B b;
    b.a.i = 2;
    printf("b.a.i = %d\n", b.a.i); // 2, correct
}
```

```javascript
function failingJavascriptTestFunction() {
    const b = new test.B();
    b.a.i = 2;
    console.log("b.a.i = %d", b.a.i); // 42, wrong, should be 2, b.a.i was not modified
}
```

**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.18 (d5ca9bba6513763d5bdddbd0efff759332bd85d7)
clang version 16.0.0 (https://github.com/llvm/llvm-project 48129cf0ed5053ed3fdf4f928180635e84892614)
Target: wasm32-unknown-emscripten
Thread model: posix

**Full link command:**
em++ -sEXPORT_NAME=TestEmbind -sMODULARIZE=1 -sEXPORT_ES6=1 -lembind -o testEmbind.js --no-entry main.cpp

brendandahl:
This is the same as #3551. It also looks like there was a PR created to address this in #6402, but it was never merged.

It does seem like odd behavior.

PR #21935, opened 2024-05-14
[embind] Return value policy support for properties.

brendandahl:
All the return polices now also work with `property` bindings. Using `return_value_policy::reference()` is especially helpful for binding child objects of a class so they don't always create copies and potentially leak.

Fixes #6402, #17573


test/embind/test_return_value_policy.cpp
```diff
@@ -264,7 +331,27 @@ EMSCRIPTEN_BINDINGS(xxx) {
...
+    // Check that compiling with getter and setter methods works as expected.
+    // These are only tested for compilation since they use the same underlying
+    // mechanisms as the other property bindings.
+    // Getter only.
+    .property("getterValue", &ValueHolder::get, return_value_policy::reference())
+    // Getter and setter.
+    .property("getterSetterValue", &ValueHolder::get, &ValueHolder::set, return_value_policy::reference())
+    // std::function getter.
+    .property("getterPtrWithStdFunctionReference", std::function<const Value*(const ValueHolder&)>(&value_holder), return_value_policy::reference())
     ;
```
kripken, line 356: I think the use case I had in mind for this feature was with raw pointers - is that supported as well? If so can it be tested here?

brendandahl: It's supported and tested above. https://github.com/emscripten-core/emscripten/pull/21935/files#diff-4fe14676abdaa75c05abe1f28bf3cd5c2dbe96d8abc99f7c893fa883b89e06e5R343-R345

kripken: Thanks, I see.

Though I was expecting to read `allow_raw_pointers()`. Is that not needed in these cases? I can't tell from the docs.

brendandahl: I'm hoping to deprecate `allow_raw_pointers()` in favor of return policies. Using a return value policy implies that raw pointers are allowed. There's a little note on this [here](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#raw-pointers). I should probably update that to be more generic now that RVP is allowed on properties too.

kripken: I see, thanks!

test/embind/test_return_value_policy.cpp
```diff
@@ -78,26 +78,44 @@ void assertAllCountsZero() {
...
+  void set(Value v) {
+    value = v;
+  }
+  Value get() const {
+    return value;
+  }
+  void initializePtr() {
+    valuePtr = new Value();
+  }
+  Value* getNewPtr() const {
```
kripken, line 102: I see this function moved and became `() const {` - was that just a cleanup or is it necessary?

test/embind/test_return_value_policy.cpp
```diff
@@ -78,26 +78,44 @@ void assertAllCountsZero() {
...
   static Value* staticGetPtr() {
     static Value v;
     return &v;
   }
 };
 
-Value* value_holder(ValueHolder& target) {
-  return target.getNewPtr();
+const Value* value_holder(const ValueHolder& target) {
+  return target.getNewPtrConst();
```
kripken, line 118: Same question here.

test/embind/test_return_value_policy.cpp
brendandahl: Embind requires getter function's to be const. We could have a const and non-const version for testing functions/properties, but doesn't really seem worth it.

kripken: Makes sense. Is that a new change with this PR, though? That's what I'm trying to understand, as the function seems to have worked ok as non-const before.

brendandahl: `const` is not a new requirement for getters. I added the `const` here since I now use `value_holder` to also test a getter property.  `value_holder` calls `ValueHolder.getNewPtr` so it must also be const.

kripken: I see, thanks.

kripken (approved):
I can't say I have a full understanding of the C++ template magic but it looks reasonable. Docs and tests look good.

PR #21938, opened 2024-05-14
Add mount option { autoPersist: true } to IDBFS mount.

juj:
Add new setting -sIDBFS_AUTO_PERSIST which changes the semantics of an IndedexDB mount to automatically persist the VFS to IndexedDB after closing any file that has been written to. This enables users to avoid needing to call FS.syncfs() or fsync() to persist files to the filesystem.

### DIFF
--- a/src/library_idbfs.js
+++ b/src/library_idbfs.js
@@ -55,9 +55,10 @@ addToLibrary({
     mount: (...args) => { 
       var mnt = MEMFS.mount(...args);
       mnt.idbPersistState = 0; // IndexedDB sync starts in idle state
+      var memfs_node_ops = mnt.node_ops;
       mnt.node_ops = Object.assign({}, mnt.node_ops); // Clone node_ops to inject write tracking
       mnt.node_ops.mknod = (parent, name, mode, dev) => {
-        var node = MEMFS.createNode(parent, name, mode, dev);
+        var node = memfs_node_ops.mknod(parent, name, mode, dev);
         node.idbfs_mount = mnt.mount; // Remember for each IDBFS node which IDBFS mount point they came from so we know which mount to persist on modification.
         node.memfs_stream_ops = node.stream_ops; // Remember original MEMFS stream_ops for this node
         node.stream_ops = Object.assign({}, node.stream_ops); // Clone stream_ops to inject write tracking
@@ -80,6 +81,12 @@ addToLibrary({
 
         return node;
       };
+      // Also kick off persisting the filesystem on other operations that modify the filesystem.
+      mnt.node_ops.mkdir   = (...args) => { var ret = memfs_node_ops.mkdir(...args);   IDBFS.queuePersist(mnt.mount); return ret; };
+      mnt.node_ops.rmdir   = (...args) => { var ret = memfs_node_ops.rmdir(...args);   IDBFS.queuePersist(mnt.mount); return ret; };
+      mnt.node_ops.symlink = (...args) => { var ret = memfs_node_ops.symlink(...args); IDBFS.queuePersist(mnt.mount); return ret; };
+      mnt.node_ops.unlink  = (...args) => { var ret = memfs_node_ops.unlink(...args);  IDBFS.queuePersist(mnt.mount); return ret; };
+      mnt.node_ops.rename  = (...args) => { var ret = memfs_node_ops.rename(...args);  IDBFS.queuePersist(mnt.mount); return ret; };
       return mnt;
     },
 #else
--- a/src/library_idbfs.js
+++ b/src/library_idbfs.js
@@ -23,8 +23,70 @@ addToLibrary({
     },
     DB_VERSION: 21,
     DB_STORE_NAME: 'FILE_DATA',
+
+#if IDBFS_AUTO_PERSIST
+    // Queues a new VFS -> IDBFS synchronization operation
+    queuePersist: (mount) => {
+      function onPersistComplete() {
+        if (mount.idbPersistState === 'again') startPersist(); // If a new sync request has appeared in between, kick off a new sync
+        else mount.idbPersistState = 0; // Otherwise reset sync state back to idle to wait for a new sync later
+      }
+      function startPersist() {
+        mount.idbPersistState = 'idb'; // Mark that we are currently running a sync operation
+        IDBFS.syncfs(mount, /*populate:*/false, onPersistComplete);
+      }
+
+      if (!mount.idbPersistState) {
+        // Programs typically write/copy/move multiple files in the in-memory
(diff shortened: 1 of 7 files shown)


curiousdannii:
Great idea! I was just adding this to my project (overwriting the default library function), but your way should be much more reliable.

```
    // Sync IDBFS after saving a savefile
    fd_close(fd) {
        try {
            const stream = SYSCALLS.getStreamFromFD(fd)
            const flags = stream.flags
            const path = stream.path
            FS.close(stream)
            // Don't run if we're currently syncing (because syncing will run this code)
            if (FS.syncFSRequests === 0) {
                // Then check that we're saving a savefile
                if (path.startsWith('/saves/') && flags & 3) {
                    FS.syncfs(err => {
                        if (err) {
                            console.log(err)
                        }
                    })
                }
            }
            return 0
        } catch (e) {
            if (typeof FS == 'undefined' || !(e.name === 'ErrnoError')) throw e
            return e.errno
        }
    },
```

src/settings.js
```diff
@@ -1045,6 +1045,11 @@ var FORCE_FILESYSTEM = false;
 // [link]
 var NODERAWFS = false;
 
+// Enables IDBFS backend to automatically persist filesystem to IDBFS after
+// any filesystem operation that modifies the filesystem.
+// This avoids needing to manually call FS.syncfs(cb) after FS operations.
+var IDBFS_AUTO_PERSIST = false;
```
curiousdannii, line 1051: Add 
```
// [link]
```

sbc100:
What do you think about enabling this by default?

juj:
> What do you think about enabling this by default?

Personally that would be ok for me, although I do worry that there might be some user projects that would be broken by such a change, in case they might have custom semantics for when the IndexedDB database would be synchronized. So maybe err to side of caution for now?

sbc100:
> > What do you think about enabling this by default?
> 
> Personally that would be ok for me, although I do worry that there might be some user projects that would be broken by such a change, in case they might have custom semantics for when the IndexedDB database would be synchronized. So maybe err to side of caution for now?

sgtm.  We can always consider enabling by default later.

juj:
Any more reviews, or a ✅ ?

src/library_idbfs.js
```diff
@@ -23,8 +23,78 @@ addToLibrary({
...
+      } else if (mount.idbPersistState === 'idb') {
+        // There is an active IndexedDB sync operation in-flight, but we now
+        // have accumulated more files to sync. We should therefore queue up
+        // a new sync after the current one finishes so that all writes
+        // will be properly persisted.
+        mount.idbPersistState = 'again';
+      }
+    },
+
+    mount: (...args) => {
```
sbc100, line 55: What do you think about making this a mount option instead of a settings?   (That would limit the scope of this chance to just this library file).   Just an idea.   I guess the downside would be a minor code size increase for idbfs users who don't need this feature, but I imagine it would be marginal.. since the FS code is so heavy.

src/library_idbfs.js
```diff
@@ -23,8 +23,78 @@ addToLibrary({
...
+
+    mount: (...args) => { 
+      var mnt = MEMFS.mount(...args);
+      mnt.idbPersistState = 0; // IndexedDB sync starts in idle state
+      var memfs_node_ops = mnt.node_ops;
+      mnt.node_ops = Object.assign({}, mnt.node_ops); // Clone node_ops to inject write tracking
+      mnt.node_ops.mknod = (parent, name, mode, dev) => {
+        var node = memfs_node_ops.mknod(parent, name, mode, dev);
+        node.node_ops = mnt.node_ops; // Propagate injected node_ops to the newly created child node
+        node.idbfs_mount = mnt.mount; // Remember for each IDBFS node which IDBFS mount point they came from so we know which mount to persist on modification.
```
sbc100, line 63: Can you put these comments on the line above the code, to avoid the long lines?

src/library_idbfs.js
juj: Yeah, the FS code is quite heavy already, so I wouldn't include it in the aggressive minimal code size optimization domain.

I suppose it could be an option, I'll see about refactoring it to that form.

src/library_idbfs.js
```diff
@@ -23,8 +23,78 @@ addToLibrary({
...
+            IDBFS.queuePersist(n.idbfs_mount);
+            n.isModified = false;
+          }
+          if (n.memfs_stream_ops.close) return n.memfs_stream_ops.close(stream);
+        };
+
+        return node;
+      };
+      // Also kick off persisting the filesystem on other operations that modify the filesystem.
+      mnt.node_ops.mkdir   = (...args) => { var ret = memfs_node_ops.mkdir(...args);   IDBFS.queuePersist(mnt.mount); return ret; };
```
sbc100, line 86: Can you write these as comma expressions to avoid the return statement and curly braces?

```
 mnt.node_ops.mkdir   = (...args) =>  IDBFS.queuePersist(mnt.mount), memfs_node_ops.mkdir(...args);
```

tlively (commented):
~Would we be able to make this the default behavior rather than adding another setting?~

Edit: already discussed.

src/library_idbfs.js
juj: Hmm, does that parse?

```js
function f() { return 1; }
function g() { return 2; }
var foo = (...args) => f(), g();
```
gives a syntax error, though
```js
function f() { return 1; }
function g() { return 2; }
var foo = (...args) => { return f(), g() };
```
works.

juj:
Updated code to make this a runtime setting.

src/library_idbfs.js
sbc100: Oh interesting.  Looks like simple comma expression work: `var foo = (...args) => f, g;`.

More complex ones require braces I guess: `var foo = (...args) => (f(), g());`

sbc100 (approved):
lgtm.  Looks like the PR title needs updating.

sbc100 (commented):
Might be worth a ChangeLog entry?

I do think we should consider making this the default at some point.  Perhaps we should reach out to existing users via the mailing list to see if this makes sense?

One possible way to make this the default and have folks opt out would be to have the public `IDBFS.syncfs()` assert if you its called when autoPersist is enabled?

curiousdannii:
Just tried this out. While it does persist any writes, it doesn't initially read in from IDB, which I was expecting/hoping for. That's not hard to do yourself (`FS.syncfs(true, () => {})`), but it would be nice if it was no longer needed. But maybe there's a reason it can't be (timing related perhaps, seeing as it's not an async function?)

sbc100:
That seems like a logical expectation.

PR #21943, opened 2024-05-14
[Docs] Add default values to settings doc page

kripken:
This adds `Default value: ..` for each setting. Example output:

settings_default

Context: Recent discussion on finding the default values: https://groups.google.com/g/emscripten-discuss/c/SjoruKoQ2SM/m/R_HYP1ELAwAJ

### DIFF
--- a/tools/maint/update_settings_docs.py
+++ b/tools/maint/update_settings_docs.py
@@ -56,7 +56,7 @@ all_tags = {
 output_file = path_from_root('site/source/docs/tools_reference/settings_reference.rst')
 
 
-def write_setting(f, setting_name, comment, tags):
+def write_setting(f, setting_name, setting_default, comment, tags):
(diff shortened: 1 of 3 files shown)


tools/maint/update_settings_docs.py
```diff
@@ -92,9 +93,13 @@ def write_file(f):
           continue
       current_comment.append(line)
     elif line.startswith('var'):
-      setting_name = line.split()[1]
+      # Format:
+      #   var NAME = DEFAULT;
+      # Split it and strip the final ';'.
+      _, setting_name, _, setting_default = line.split()
```
sbc100, line 99: `line.strip(';').split()`?

sbc100 (approved):

kripken:
After applying the feedback I noticed that this does not handle one setting, which has a multi-line value. I added a TODO for it.

PR #15651, opened 2021-11-29
Copy of issue/14745 code changes

CDLlo:
Making a separate PR for the changes in the [prior PR by ArthurG0](https://github.com/emscripten-core/emscripten/pull/14795). Unfortunately the branch being pulled from is not available for me to update/rebase and ArthurG0 left our company so it is not expected of them to follow up on the necessary changes for the stale PR to be accepted. I added them to the AUTHORS file in this branch to keep record of their contributions.

To restate the original PR, the original issue was that the WebIDL Binder was incorrectly handling the const char* return type as it treated it as an `int` rather than a `char*`. The fix allows the basic type to be cast into the correct type in the generated glue code.

PR #14795, opened 2021-08-02
issue/14745 code

ArthurG0:

Issue #14745, opened 2021-07-23
Bug in generated C++ wrapper for overriding C++ classes in JS when using WebIDL_Binder


ArthurG0:
Hello,

Noticed a potential bug in webidl_binder. When generating the glue code for two abstract classes and returning a `const char *` (`[Const] DOMString` in WebIDL ) variable, the glue code treats it as an `int` - return type and doesn't type cast it to a `char *` , which leads to a compilation error. Below is a small example to illustrate the issue

Files provided:
`Foo.h` - provides struct definitions for Foo and Bar, the two abstract structs used. 
`Full.idl` - WebIDL Interfaces of Foo and Bar
`Glue.cpp, Glue.js` - files generated by the WebIDL Binder (using emcc version 2.0.21)
`Wrapper.cpp` - a Wrapper file needed for the compilation step

Commands used:
```
- tools/webidl_binder Full.idl Glue
- emcc Wrapper.cpp --post-js Glue.js -o EMCCOutput.js
```

The error is:

cannot initialize return object of type '`char *`' with an rvalue of type '`int`' , and refers to this piece of code in Glue.cpp:

```
char* getString() const  {
    return EM_ASM_INT({
      var self = Module['getCache'](Module['JSFoo'])[$0];
      if (!self.hasOwnProperty('getString')) throw 'a JSImplementation must implement all functions, you forgot JSFoo::getString.';
      return self['getString']();
    }, (int)this);
  }
```
As you can see, the return type is an integer that is not being type cast into a char pointer, causing a compilation error.

ArthurG0:
Link to the archive with referenced Files:

https://drive.google.com/file/d/1qYS8-GxdCfWzUA9owx19XItT6uGOC9aV/view?usp=sharing

ArthurG0:
**Found a possible solution that enables the code to compile.**

**Needs further testing to make sure it doesn't break anything else before submitting a PR.**

**In `tools/webidl_binder.py`,**


js_impl_methods.append(r'''  %s %s(%s) %s {
    %sEM_ASM_%s({
      var self = Module['getCache'](Module['%s'])[$0];
      if (!self.hasOwnProperty('%s')) throw 'a JSImplementation must implement all functions, you forgot %s::%s.';
      %sself['%s'](%s)%s;
    }, (int)this%s);
  }''' % (c_return_type, func_name, dec_args, maybe_const,
          basic_return, 'INT' if c_return_type not in C_FLOATS else 'DOUBLE',
[...]

PR #21944, opened 2024-05-14
[WebIDL] Generate correct code when a pointer is returned from an interface method

agnickolov:
This fixes the problem reported under issue https://github.com/emscripten-core/emscripten/issues/14745. The previous two PRs were never completed:
https://github.com/emscripten-core/emscripten/pull/15651
https://github.com/emscripten-core/emscripten/pull/14795

Those developers are no longer involved with the project this issue is blocking and it fell to me to get it through to completion. I cleaned up the test code a bit from the earlier PRs as well. Please look at the previous PRs for details.

### DIFF
--- a/test/webidl/test.h
+++ b/test/webidl/test.h
@@ -194,3 +194,38 @@ typedef struct LongLongTypes {
   unsigned long long* lluArray;
   long long ll;
 } LongLongTypes;
+
+// Returning child objects in a hierarchy
+
+struct ISmallObject {
+  virtual int getID(int number) = 0;
+};
+
+struct IObjectProvider {
+  virtual ISmallObject* getObject() = 0;
+};
+
+class SmallObject : public ISmallObject {
+public:
+  int getID(int number) {
+    return number;
+  }
+};
+
+class ObjectProvider : public IObjectProvider {
+public:
+  ISmallObject* getObject() {
+    return &m_smallObject;
+  }
+private:
+  SmallObject m_smallObject;
+};
+
+class ObjectFactory {
+public:
+  IObjectProvider* getProvider() {
+    return &m_ObjectProvider;
+  }
+private:
+  ObjectProvider m_ObjectProvider;
+};
--- a/tools/webidl_binder.py
+++ b/tools/webidl_binder.py
@@ -647,15 +647,16 @@ def render_function(class_name, func_name, sigs, return_type, non_pointer,
       if i == max_args:
         dec_args = ', '.join([type_to_cdec(raw[j]) + ' ' + args[j] for j in range(i)])
         js_call_args = ', '.join(['%s%s' % (('(ptrdiff_t)' if sig[j] in interfaces else '') + take_addr_if_nonpointer(raw[j]), args[j]) for j in range(i)])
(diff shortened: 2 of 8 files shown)


tools/webidl_binder.py
```diff
@@ -649,13 +649,13 @@ def make_call_args(i):
...
         js_impl_methods.append(r'''  %s %s(%s) %s {
-    %sEM_ASM_%s({
+    %s (%s) EM_ASM_%s({
       var self = Module['getCache'](Module['%s'])[$0];
       if (!self.hasOwnProperty('%s')) throw 'a JSImplementation must implement all functions, you forgot %s::%s.';
       %sself['%s'](%s)%s;
     }, (ptrdiff_t)this%s);
   }''' % (c_return_type, func_name, dec_args, maybe_const,
-          basic_return, 'INT' if c_return_type not in C_FLOATS else 'DOUBLE',
+          basic_return, c_return_type, 'INT' if c_return_type not in C_FLOATS else 'DOUBLE',
```
sbc100, line 658: We have `EM_ASM_PTR` now that we should probably use for pointer returns.

I wonder if use that then we don't need the above cast?

Maybe create a new `em_asm_macro` variable here that can hold the name of the macro to use?

agnickolov: This is a 2.5 years old PR. I can investigate with your hint and perhaps offer an improved version. I know very little about the project internals, however.

agnickolov: I'm concerned that I'm getting test failures with the fixed code. Assuming they worked before, that likely means there are a bunch of tests that validate the existing wrong behavior...

agnickolov:
I'm concerned that I'm getting test failures with the fixed code. Assuming they worked before, that likely means there are a bunch of tests that validate the existing wrong behavior.

EDIT: At least one of the failures seems to be due to a configuration issue. The following unexpected line appears in the test output:
`+Warning: Enlarging memory arrays, this is not fast! 16908288,25296896
`
I'm not sure what should be done about it, however.

tools/webidl_binder.py
agnickolov: I refactored it with your suggestion, but the type cast cannot be avoided - the C++ compiler still raises an error for incompatible types. I incorporated the use of EM_ASM_PTR instead of EM_ASM_INT and extracted into a variable as you suggested.

agnickolov:
All tests appear to be passing, but an archive check seems to have stalled...

test/webidl/test.h
```diff
@@ -194,3 +194,38 @@ typedef struct LongLongTypes {
...
+  SmallObject m_smallObject;
+};
+
+class ObjectFactory {
+public:
+  IObjectProvider* getProvider() {
+    return &m_ObjectProvider;
+  }
+private:
+  ObjectProvider m_ObjectProvider;
```
sbc100, line 230: Does it really take this complex set of interwoven classes to reproduce this issue?  i.e. do you need both a provider and a factory here?

sbc100 (approved):
lgtm with one last comment

test/webidl/test.h
agnickolov: Unfortunately, yes. A single level of indirection does not expose the issue.

PR #21948, opened 2024-05-15
fix friend declaration in val.h

walkingeyerobot:
The code as it was trying to declare a template instantiation as a friend, and then specialize it later, which caused a build break. This fix just declares the whole template as a friend.


sbc100:
I supposed this is superseded by #21949?

brendandahl:
Yeah.

PR #21957, opened 2024-05-16
Fix for register_vector and optional class_ collision.

wffurr:
Fixes `BindingError: Cannot register type 'emscripten::val' twice` error. This happened with pre-existing binding code that was migrated to Emscripten 3.1.52 which introduces the register_optional function, used internally by register_vector, in https://github.com/emscripten-core/emscripten/pull/21076.

The fix ignores duplicate registrations of emscripten::val, and doesn't rely on the boolean already-registered flag local to the register_optional function.

### DIFF
--- a/system/include/emscripten/bind.h
+++ b/system/include/emscripten/bind.h
@@ -1919,14 +1919,8 @@ public:
 #if __cplusplus >= 201703L
 template<typename T>
 void register_optional() {
-    // Optional types are automatically registered for some internal types so
-    // only run the register method once so we don't conflict with a user's
-    // bindings if they also register the optional type.
-    thread_local bool hasRun;
-    if (hasRun) {
-        return;
-    }
-    hasRun = true;
+    // Optional types are automatically registered for some internal types; the
+    // JS implementation will ignore duplicate registrations.
     internal::_embind_register_optional(
         internal::TypeID<std::optional<T>>::get(),
         internal::TypeID<typename std::remove_pointer<T>::type>::get());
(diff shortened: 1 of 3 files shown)


wffurr:
Tests pass except for the binary size, +7 bytes.  Please take a look and let me know if this change is a good one and if so I'll fix that test.

sbc100:
> Tests pass except for the binary size, +7 bytes. Please take a look and let me know if this change is a good one and if so I'll fix that test.

Rebasing/merging should fix those failure.  They are related I believe.

system/include/emscripten/bind.h
```diff
@@ -1919,14 +1919,8 @@ class class_ {
...
 void register_optional() {
-    // Optional types are automatically registered for some internal types so
-    // only run the register method once so we don't conflict with a user's
-    // bindings if they also register the optional type.
-    thread_local bool hasRun;
-    if (hasRun) {
-        return;
-    }
-    hasRun = true;
+    // Optional types are automatically registered for some internal types; the
```
brendandahl, line 1922: Allowing duplicate registrations will lead to a slightly strange behavior where the order matters. e.g.

The following will fail:
```c++
  register_optional<Foo*>();
  class_<std::optional<Foo*>>("OptionalFoo");
  ```
but swapping them will work:
```c++
  class_<std::optional<Foo*>>("OptionalFoo");
  register_optional<Foo*>();
  ```

I guess if a user is creating their own binding for std::optional and they want to override the behavior that could be fine.

wffurr:
This PR also doesn't fully address backwards incompatibility between `optional<T>` as a `class_` binding and `optional<T>` as `undefined|T`.  Namely, the `class_` version of the optional needs to be deleted from JS or it leaks heap memory, whereas the `undefined|T` version does not, unless `T` itself needs to be deleted.

Issue #21908, opened 2024-05-08
AUDIO_WORKLET + MODULARIZE broken in 3.1.59


kichikuou:
**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.59 (0e4c5994eb5b8defd38367a416d0703fd506ad81)
clang version 19.0.0git (https:/github.com/llvm/llvm-project df762a1643bb5b0b3c907611d118c82d4b68a39d)
Target: wasm32-unknown-emscripten
Thread model: posix
```

**Failing command line in full:**
Compile `audioworklet.c` in the Emscripten webaudio test directory:
```
cd emscripten/test/webaudio
emcc audioworklet.c -o audioworklet.html -sAUDIO_WORKLET -sWASM_WORKERS -sMODULARIZE -sEXPORT_ES6
```
And open `audioworklet.html` in the browser.

**Error:**
```
audioworklet.js:635 Aborted(Access to module property ('stackAlloc') is no longer possible via the module constructor argument; Instead, use the result of the module constructor.)
audioworklet.js:654 Uncaught RuntimeError: Aborted(Access to module property ('stackAlloc') is no longer possible via the module constructor argument; Instead, use the result of the module constructor.)
    at abort (audioworklet.js:654:11)
    at Object.get (audioworklet.js:1947:9)
    at new WasmAudioWorkletProcessor (audioworklet.aw.js:24:37)
```

It looks like a regression by #21775.

sbc100:
Interesting.  I wonder why the test we have that (`browser.test_audio_worklet_es6`) is not failing..

sbc100:
The reason we don't see this fail is because `browser.test_audio_worklet_es6` doesn't keep the runtime alive long enough do anything with the audio worklet, it just exits when main returns.   Also, we don't currently have any way to actually run audio tests during CI so even without this crash the test isn't actually playing any audio when run in CI.  Instead we must rely mostly on interactive tests today to test audio stuff.

PR #21775, opened 2024-04-17
In MODULARIZE mode avoid modifying the incoming moduleArg. NFC

sbc100:
This avoids leaking of the partially constructed module and means that only way to get access the module instance is via waiting on the promise.

Previously we were attaching all our module properties to the incoming object, but not, as far as I can tell, for any good reason.

In case anybody was actually depending on this, inject thunks into the moduleArg that abort on access with an actionable error.

PR #21958, opened 2024-05-16
Fix AUDIO_WORKLET + MODULARIZE

sbc100:
Fixes: #21908

### DIFF
--- a/src/audio_worklet.js
+++ b/src/audio_worklet.js
@@ -150,7 +150,7 @@ class BootstrapMessages extends AudioWorkletProcessor {
     // scope to create the real AudioWorkletProcessors that call out to Wasm to
     // do audio processing.
     let p = globalThis['messagePort'] = this.port;
-    p.onmessage = (msg) => {
+    p.onmessage = async (msg) => {
       let d = msg.data;
       if (d['_wpn']) {
         // '_wpn' is short for 'Worklet Processor Node', using an identifier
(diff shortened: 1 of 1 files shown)


sbc100:
Verified fix using interactive tests

kripken:
Can you please explain why renaming this makes a difference? I can't figure it out.

MuTsunTsai:
@kripken Looks like the key is adding `await` in this line:

https://github.com/emscripten-core/emscripten/blob/a36f43c1c3357442da300d43db7b363c74985e80/src/audio_worklet.js#L164

juj:
LGTM, but has there been a semantics change to how MODULARIZE works? The `await` there is surprising to me.

sbc100:
> LGTM, but has there been a semantics change to how MODULARIZE works? The `await` there is surprising to me.

With `MODULARIZE` the constructor functions has always returned a Promise of a module (IIUC, thats the main difference between `MODULARIZE` and now removed `MODULARIZE_INSTANCE`).   

There was a semantic change back in #21775 where the argument passed into the module constructor was not modified in-place.. this avoids leaking the module before the promise resolves.

sbc100:
> Can you please explain why renaming this makes a difference? I can't figure it out.

Sorry the rename is unrelated.  I will revert that.

sbc100:
Hopefully all these code will be removed soon anyway once I figure out how to inline the `ww.js` and `aw.js` files like I did for the pthread `worker.js`

sbc100:
Simplified this PR by removing the rename.  PTAL.

kripken:
Thanks @MuTsunTsai , I missed that part among the rest of the diff...

kripken (approved):
Can we add a test for this?

sbc100:
> Can we add a test for this?

This failure was already showing up but only in the interactive tests... so it was already tests, just not by CI tests (that same goes for pretty much all our audio tests).

PR #18692, opened 2023-02-09
[EH/SjLj] Make Wasm EH force Wasm SjLj

aheejin:
Currently `SUPPORT_LONGJMP` defaults to `emscripten` unless explicitly otherwise. So when a user specified `-fwasm-exceptions`, it resulted in Wasm EH + Emscripten SjLj, which has many restrictions (See https://github.com/emscripten-core/emscripten/discussions/17526#discussioncomment-3265159 for details) and is the combination we do not intend to support for the long term.

This PR makes `SUPPORT_LONGJMP` default value depend on the exception mode. If Wasm EH (`-fwasm-exception`) is used, it defaults to `wasm`, and if Emscripten EH (`-sDISABLE_EXCEPTION_CATCHING=0`) is used or no exception support is used, it defaults to `emscripten`. This way users get to use the same kind of EH + SjLj, which is easier to maintain and has less restrictions.

PR #21971, opened 2024-05-21
[SjLj] __USING_WASM_SJLJ__ -> __WASM_SJLJ__

aheejin:
In line with #21970, this changes `__USING_WASM_SJLJ__` to `__WASM_SJLJ__`. We don't define this in Clang now (which we can maybe consider, but given that Clang doesn't have dedicated flags for SjLj and there don't seem to be other platforms doing similar things), but given that the affected files currently are only in Emscripten, I think we can safely change this.

### DIFF
--- a/system/lib/compiler-rt/emscripten_setjmp.c
+++ b/system/lib/compiler-rt/emscripten_setjmp.c
@@ -13,7 +13,7 @@
 
 #include "emscripten_internal.h"
 
-#ifdef __USING_WASM_SJLJ__
+#ifdef __WASM_SJLJ__
 struct __WasmLongjmpArgs {
   void *env;
   int val;
@@ -25,7 +25,7 @@ struct __WasmLongjmpArgs {
 struct jmp_buf_impl {
   void* func_invocation_id;
   uint32_t label;
-#ifdef __USING_WASM_SJLJ__
+#ifdef __WASM_SJLJ__
   struct __WasmLongjmpArgs arg;
 #endif
 };
@@ -48,7 +48,7 @@ uint32_t __wasm_setjmp_test(void* env, void* func_invocation_id) {
   return 0;
 }
 
-#ifdef __USING_WASM_SJLJ__
+#ifdef __WASM_SJLJ__
 // llvm uses `1` for the __c_longjmp tag.
 // See https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/CodeGen/WasmEHFuncInfo.h
 #define C_LONGJMP 1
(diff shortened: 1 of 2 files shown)


sbc100 (commented):
Do we need this macro in practice?   Can we perhaps merge it with __WASM_EXCEPTIONS__?

aheejin:
That's a good question... I think we separated it mainly because there was time we supported Wasm EH but not Wasm SjLj so we let people use Wasm EH with Emscripten SjLj. This still works but not the recommended mode of using it anymore; we now use Wasm SjLj by default if `-fwasm-exceptions` is given. But you can still manually override it by `-fwasm-exceptions -sSUPPORT_LONGJMP=emscripten` or something.

Merging the two preprocessor means we disallow using Wasm EH with Emscripten SjLj. Would that be OK? (At this point I can't think of a reason why someone wants to use Wasm EH with Emscripten SjLj though)

sbc100:
> That's a good question... I think we separated it mainly because there was time we supported Wasm EH but not Wasm SjLj so we let people use Wasm EH with Emscripten SjLj. This still works but not the recommended mode of using it anymore; we now use Wasm SjLj by default if `-fwasm-exceptions` is given. But you can still manually override it by `-fwasm-exceptions -sSUPPORT_LONGJMP=emscripten` or something.
> 
> Merging the two preprocessor means we disallow using Wasm EH with Emscripten SjLj. Would that be OK? (At this point I can't think of a reason why someone wants to use Wasm EH with Emscripten SjLj though)

The more configurations we can disallow the better IMO!

aheejin:
I we can probably land this PR as is. What I commented was actually incorrect. We actually already have been forcing Wasm SjLj when Wasm EH is used (#18692), which predates this PR. And while replacing the three cases of `__WASM_SJLJ__` here with `__WASM_EXCEPTIONS__` doesn't seem to break anything, it doesn't seem to simplify anything either, because the only place we [use](https://github.com/emscripten-core/emscripten/blob/8046a49bf15c643e3ee4394468995d12e4e5952c/tools/system_libs.py#L855) `__WASM_SJLJ__` is [`SjLjLibrary`](https://github.com/emscripten-core/emscripten/blob/8046a49bf15c643e3ee4394468995d12e4e5952c/tools/system_libs.py#L842-L873) and the only subclass of `SjLjLibrary` is [`libcompiler_rt`](https://github.com/emscripten-core/emscripten/blob/8046a49bf15c643e3ee4394468995d12e4e5952c/tools/system_libs.py#L918-L964). So even if we replace `__WASM_SJLJ__` with `__WASM_EXCEPTIONS__` in libcompiler_rt these usages are separate from the usage in libcxx/libcxxabi/libunwind, and it looks more confusing because libcompiler_rt doesn't seem to be related to exceptions.

sbc100 (approved):

PR #21982, opened 2024-05-22
Re-enable test after LLVM roll [ci skip]

kripken:
These new expectations will work once LLVM rolls.

Looks like code size worsened but perhaps LLVM sees benefit in inlining differently...


kripken:
Builds should be ready so this should be good to go now, if tests pass.

sbc100 (approved):

PR #21985, opened 2024-05-23
Make RTLD_LOCAL work correctly for preloaded DSOs

msorvig:
Follow-up to commit c9a5e63c, for the FS.createPreloadedFile() case.

Make sure loadWebAssemblyModule() and loadDynamicLibrary() are called with a valid 'localScope' object when invoked from the .so file type preload handler.


sbc100:
Thanks for working on this!  Do you think we can add a test for this?

msorvig:
Yep, having a test makes sense. I'll see what can be done.

msorvig:
Added a test for the preload case.

test/test_core.py
```diff
@@ -3958,6 +3958,48 @@ def test_dlfcn_rtld_local(self):
...
+      #include <assert.h>
+      #include <dlfcn.h>
+      #include <stdio.h>
+      #include <sys/stat.h>
+
+      int main() {
+        // Check the file exists in the VFS
+        struct stat statbuf;
+        assert(stat("/libB.so", &statbuf) == 0);
+        void *lib_handle = dlopen("/libB.so", RTLD_NOW);
```
sbc100, line 3990: Perhaps add `RTLD_LOCAL` explicitly since that is what we are testing here isn't it?

test/test_core.py
```diff
@@ -3958,6 +3958,48 @@ def test_dlfcn_rtld_local(self):
 
     self.do_runf('main.c', 'main\nfunc_a\nfunc_sub\nfunc_b\nfunc_sub\ndone\n')
 
+  def test_dlfcn_with_preload(self):
+    # Create chain of dependencies and load the first libary with preload plugin.
+    # main -> libB.so -> libA.sp
```
sbc100, line 3963: libA.sp -> libA.so

test/test_core.py
```diff
@@ -3958,6 +3958,48 @@ def test_dlfcn_rtld_local(self):
...
+      #include <sys/stat.h>
+
+      int main() {
+        // Check the file exists in the VFS
+        struct stat statbuf;
+        assert(stat("/libB.so", &statbuf) == 0);
+        void *lib_handle = dlopen("/libB.so", RTLD_NOW);
+        assert(lib_handle);
+        typedef int (*intfunc)();
+        intfunc x = (intfunc)dlsym(lib_handle, "libB_fun");
```
sbc100, line 3993: How does this test fail without this change?

msorvig: It fails with this error:

`Assertion failed: undefined symbol 'libA_fun'. perhaps a side module was not linked in? if this global was expected to arrive from a system library, try to build the MAIN_MODULE with EMCC_FORCE_STDLIBS=1 in the environment`

sbc100 (approved):

test/test_core.py
```diff
@@ -3958,6 +3958,48 @@ def test_dlfcn_rtld_local(self):
...
+    self.run_process([EMCC, 'libA.c', '-sSIDE_MODULE', '-o', 'libA.so'])
+    create_file('libB.c', r'''
+      #include <stdio.h>
+      int libA_fun();
+
+      int libB_fun() {
+        return libA_fun()*2;
+      }
+    ''')
+    self.run_process([EMCC, 'libB.c', '-sSIDE_MODULE', '-o', 'libB.so', 'libA.so'])
```
sbc100, line 3979: Rather than calling `run_process` directly, I think you need to either:

1. Use `self.build_dlfcn_lib`
2. Use `self.emcc`
3. Keep using self.run_process by add `self.get_emcc_args()` yourself.

msorvig: I changed it to use `build_dlfcn_lib` and `self.prep_dlfcn_main` similar to other tests.

msorvig:
`test_minimal_runtime_code_size_hello_webgl_wasm2js`
`AssertionError: 22052 != 22037`

How do we update the expected code size here if there is a change? Should it be done as a part of this PR?

sbc100:
> `test_minimal_runtime_code_size_hello_webgl_wasm2js` `AssertionError: 22052 != 22037`
> 
> How do we update the expected code size here if there is a change? Should it be done as a part of this PR?

That codesize change looks unrelated.  I well see if a rebase is needed.

sbc100:
It looks like maybe you need to rebase or merge with main branch

sbc100 (approved):

test/test_core.py
```diff
@@ -3958,6 +3958,51 @@ def test_dlfcn_rtld_local(self):
 
     self.do_runf('main.c', 'main\nfunc_a\nfunc_sub\nfunc_b\nfunc_sub\ndone\n')
 
+  def test_dlfcn_preload(self):
```
sbc100, line 3961: Adding the `@needs_dylink` decorator should fix the wasm2js test failure.

msorvig:
Test failure is "WASM2JS is not compatible with relocatable output" - I'm guessing the test function needs a "@needs_dylink" decorator so I'll add that.

test/test_core.py
msorvig: Yep, done.

PR #21990, opened 2024-05-23
Make use of rest operator in library_dylink.js. NFC

sbc100:


sbc100 (commented):
View with whitespace ignored to see what is really changing here.

kripken (approved):

PR #21991, opened 2024-05-23
Make sure dummy worker.js is never actually run. NFC

sbc100:


kripken (commented):
Hmm, I worry this will break users that run custom minification after `emcc`. Until this PR that minification left the empty file empty but now it would error?

sbc100:
> Hmm, I worry this will break users that run custom minification after `emcc`. Until this PR that minification left the empty file empty but now it would error?

Hmm.. ok maybe I'll leave it be.

sbc100:
OK I made it into a runtime error instead.

kripken (approved):

Issue #21989, opened 2024-05-23
pthread workers fail to run when ENVIRONMENT=worker only


steveanton:
isPthread is returning false for pthread workers in this scenario.

isPthread check inside of the pthread workers relies on the 'name' property being set on the Worker constructor from the main Emscripten JS. However, that 'name' property is only set in two conditions (https://github.com/emscripten-core/emscripten/blob/main/src/library_pthread.js#L427): when ENVIRONMENT_MAY_BE_NODE or ENVIRONMENT_MAY_BE_WEB. Notably ENVIRONMENT_MAY_BE_WEB only includes ENVIRONMENT=web and not ENVIRONMENT=worker.

jozefchutka:
Interestingly it fails to use `tot`. Should I create a separate issue?

```
emsdk install tot
Fetching latest changes to the branch/tag 'main' for '/ffmpeg-wasm/modules/emsdk/releases'...
refs/heads/main
Already up to date.
Successfully updated and checked out branch/tag 'main' on repository '/ffmpeg-wasm/modules/emsdk/releases'
Current repository version: "Tue, 28 May 2024 23:35:42 +0000 e43446a5bd1842cdefe6e5c1a4031aee23541b65"
error: failed to find build of any recent emsdk revision
```

jozefchutka:
It is a very different issue - `emsdk install tot` command fails

sbc100:
It looks like maybe you are not using emsdk from git?

sbc100:
Is your emsdk checkout up-to-date?

jozefchutka:
I believe I am on the very latest emsdk ([4205979](https://github.com/emscripten-core/emsdk/commit/4205979) today 2024-05-29) and releases are pulled against [61b6a2eb](https://chromium.googlesource.com/emscripten-releases/+/61b6a2eba1dc0392a1309a2cd5dbfa82113dfd41) which is from 2024-05-29 (today). Am I doint anything wrong?

```
root@c72f63f99f92:/ffmpeg-wasm/modules/emsdk# git status
HEAD detached at 4205979
nothing to commit, working tree clean
root@c72f63f99f92:/ffmpeg-wasm/modules/emsdk# emsdk install tot
Fetching latest changes to the branch/tag 'main' for '/ffmpeg-wasm/modules/emsdk/releases'...
refs/heads/main
Updating e43446a5..61b6a2eb
Fast-forward
 DEPS | 6 +++---
[...]

PR #21993, opened 2024-05-23
Fix for starting pthreads when `-sENVIRONMENT=worker`

sbc100:
Fixes: #21989


kripken (commented):
I don't see a fix here, only a test?

sbc100:
> I don't see a fix here, only a test?

Sorry, updated.

dschuff (approved):

PR #21994, opened 2024-05-23
Revert "In MODULARIZE mode avoid modifying the incoming moduleArg. NFC (#21775)"

sbc100:
This change effectively reverts 4dac6d6d76.

It turns out that there is at least one use case where accessing the partially constructed module prior to the promise resolution is necessary.  Specifically when using `--preload-files` the file packager output expects to be able to use the `Module.FS` functions before the module is loaded.

### DIFF
--- a/src/shell.js
+++ b/src/shell.js
@@ -21,7 +21,7 @@
 // before the code. Then that object will be used in the code, and you
 // can continue to use Module afterwards as well.
 #if MODULARIZE
-var Module = Object.assign({}, moduleArg);
+var Module = moduleArg;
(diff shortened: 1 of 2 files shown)


sbc100:
ping..

kripken (approved):
With WasmFS we may not need this, so in the future perhaps we can revisit this. Or maybe there is a better way to do file preloading.

sbc100:
> With WasmFS we may not need this, so in the future perhaps we can revisit this. Or maybe there is a better way to do file preloading.

Exactly, yes

Issue #21930, opened 2024-05-13
Randomly occurring libpng bug in png_read_IDAT_data


curiousdannii:
**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.59 (0e4c5994eb5b8defd38367a416d0703fd506ad81)
clang version 19.0.0git (https:/github.com/llvm/llvm-project df762a1643bb5b0b3c907611d118c82d4b68a39d)
```

I've found a randomly occurring bug. I uploaded a good build to https://curiousdannii.github.io/infocom-frotz/sfrotz.html?story=shogun, which successfully shows the title graphic. But most of the time when I build my app, I get errors in `png_read_IDAT_data`. To be clear, the randomness is at compilation time - once compiled it either always works or never does.

My code is at https://github.com/curiousdannii/infocom-frotz
Run `./build.sh` to build it, or `rm frotz/sfrotz.js && ./build.sh` to relink it.
It actually appears that the randomness is happening at the linking stage, as simply running the linker like that is enough to switch between working and broken builds.

I was at my wit's end, even trying to rebuild libpng from source, unless I noticed that just occasionally it does work. Then I was at least a little relieved that if the compiler/linker isn't being consistent it's probably not my fault.

I'll attach two zips of the working and broken output, in case inspecting the .wasm would help.

[dist-broken.zip](https://github.com/emscripten-core/emscripten/files/15288799/dist-broken.zip)
[dist-working.zip](https://github.com/emscripten-core/emscripten/files/15288802/dist-working.zip)

sbc100:
O, that sounds like a nasty issue.  Do you have any idea what the error are?  Are they related to memory layout?  Or static data?  
 Miscompilation?  What kind of errors are they?

Can you share the full set of link flags here in the bug?

curiousdannii:
The errors depend on what optimisation level I try.

O0: completely crashes/exits:
```
program exited (with status: 0), but keepRuntimeAlive() is set (counter=0) due to an async operation, so halting execution but not exiting the runtime or preventing further async execution (you can use emscripten_force_exit, if you want to force a true shutdown)
	exitJS	@	sfrotz.js:4399
	imports.<computed>	@	sfrotz.js:9144
	$inflate.1	@	sfrotz.wasm:0x30421b
	$png_read_IDAT_data	@	sfrotz.wasm:0x32dc3f
[...]

PR #21995, opened 2024-05-24
Make the order of linking of ports deterministic

sbc100:
In python `dict()` ordering is deterministic, but `set()` ordering is not.  Annoying.

This change adds a simple `OrderedSet` and uses that for managing ports.

Fixes: #21930

### DIFF
--- a/tools/ports/__init__.py
+++ b/tools/ports/__init__.py
@@ -369,6 +369,37 @@ class Ports:
     utils.write_file(filename, contents)
 
 
+class OrderedSet:
+  """Partial implementation of OrderedSet.  Just enough for what we need here."""
+  def __init__(self, items):
+    self.dict = dict()
+    for i in items:
+      self.dict[i] = True
+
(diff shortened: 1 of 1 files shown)


tools/ports/__init__.py
```diff
@@ -369,14 +369,45 @@ def write_file(filename, contents):
     utils.write_file(filename, contents)
 
 
+class OrderedSet:
+  """Partial implementation of OrderedSet.  Just enough for what we need here."""
+  def __init__(self, items):
+    self.dict = dict()
+    for i in items:
+      self.dict[i] = True
```
tlively, line 377: This could be shortened:
```suggestion
    self.dict = {item: True for item in items}
```

tlively (approved):
Wow, that is annoying.

tools/ports/__init__.py
sbc100: Oops, will followup.

PR #21997, opened 2024-05-24
Fix test_print_resource_dir on windows. NFC

sbc100:


sbc100:
Landing TBR to get the rollers green again.

PR #22002, opened 2024-05-26
Use `fetch()` for loading packages & shared libraries

seanmorris:
This PR switches AJAX requests for file packages & shared libraries to be loaded with `fetch()` rather than `XmlHttpRequest()`.  This allows service workers to make AJAX requests, since `XmlHttpRequest` is not available in service workers.

Previously, `--embed-file` was necessary to load file packages, which embeds the files into the binary and precludes caching and re-use of the package. Loading shared libraries is not possible in a service worker without this patch.

Fixes https://github.com/emscripten-core/emscripten/issues/22003


src/web_or_worker_shell_read.js
```diff
@@ -20,7 +20,19 @@
       return new Uint8Array(/** @type{!ArrayBuffer} */(xhr.response));
     };
   }
-
+#if USE_FETCH
+  readAsync = (url, onload, onerror) => {
+    fetch(url)
+    .then(response => {
+      if(response.ok) {
+        return response.arrayBuffer();
```
sbc100, line 28: Why not call `onload` here?

src/web_or_worker_shell_read.js
```diff
@@ -20,7 +20,19 @@
...
+      }
+      throw new Error(response.statusText + ' : ' + response.url);
```
sbc100, line 30: Should this not call `onerror` instead?

sbc100 (commented):
I wonder if we need a new option at all here or if we should just do this in all cases?

src/web_or_worker_shell_read.js
seanmorris: `response.arrayBuffer();` returns a promise that resolves to a buffer, not a literal buffer.

src/web_or_worker_shell_read.js
seanmorris: Let me check the make sure the error has the correct stack if its not actually thrown, and then I can change it.

seanmorris:
> I wonder if we need a new option at all here or if we should just do this in all cases?

I'd like this to be enabled by default, but I don't want to introduce any surprises in cases where people are polyfilling/overriding `XmlHttpRequest`. Perhaps we can enable it by default in the next minor release, and eventually deprecate the use of `XmlHttpRequest` gracefully.

sbc100:
> > I wonder if we need a new option at all here or if we should just do this in all cases?
> 
> I'd like this to be enabled by default, but I don't want to introduce any surprises in cases where people are polyfilling/overriding `XmlHttpRequest`. Perhaps we can enable it by default in the next minor release, and eventually deprecate the use of `XmlHttpRequest` gracefully.

We will also likely need to write a minimal polyfil for cases where we target older browser. (i.e. `-sLEGACY_BROWSER_SUPPORT`).   See the existing polyfills in `src/polyfill`.

sbc100:
> > I wonder if we need a new option at all here or if we should just do this in all cases?
> 
> I'd like this to be enabled by default, but I don't want to introduce any surprises in cases where people are polyfilling/overriding `XmlHttpRequest`. Perhaps we can enable it by default in the next minor release, and eventually deprecate the use of `XmlHttpRequest` gracefully.

Polyfilling/overriding XMLHttpRequest is not something that I know of anyone doing and we don't document emscripten as supporting this.   I think its reasonable to simply make the switch and add a ChangeLog entry for this.. if somebody really is overriding XMLHttpRequest it should be easy enough to also override fetch.   That was we can avoid adding a new setting (each time we do this it increases our testing burden).

seanmorris:
> > > I wonder if we need a new option at all here or if we should just do this in all cases?
> > 
> > 
> > I'd like this to be enabled by default, but I don't want to introduce any surprises in cases where people are polyfilling/overriding `XmlHttpRequest`. Perhaps we can enable it by default in the next minor release, and eventually deprecate the use of `XmlHttpRequest` gracefully.
> 
> Polyfilling/overriding XMLHttpRequest is not something that I know of anyone doing and we don't document emscripten as supporting this. I think its reasonable to simply make the switch and add a ChangeLog entry for this.. if somebody really is overriding XMLHttpRequest it should be easy enough to also override fetch. That was we can avoid adding a new setting (each time we do this it increases our testing burden).

I'll go ahead and make those changes.

src/web_or_worker_shell_read.js
seanmorris: I remembered why I used this pattern as I was editing: The `catch()` at the end of the function will catch errors coming from the `fetch()` itself, but `fetch()` won't throw errors by itself if the request itself fails, i.e. in a 404 or 500 error. For that we have to throw our own error. When we do that, we skip the `then()` on line 32 and head right to `catch()`.

tools/file_packager.py
```diff
@@ -965,59 +965,68 @@ def generate_js(data_target, data_files, metadata):
...
-              Module.dataFileDownloads[url] = {
-                loaded: event.loaded,
-                total: size
-              };
-            } else {
-              Module.dataFileDownloads[url].loaded = event.loaded;
+        Module.dataFileDownloads = Module.dataFileDownloads || {};
+        const url = packageName;
+        fetch(url)
+        .catch(error => { throw new Error(error + ' : ' + url) }) // Do this first so we don't catch errors from then()
```
sbc100, line 975: Instead of `.catch` here is it the same to write the error handler as the second argument to `then()`?     Is one way preferable over the other?

sbc100, line 975: I think our style uses braces around the single argument to arrow functions

tools/file_packager.py
```diff
@@ -1113,15 +1122,14 @@ def generate_js(data_target, data_files, metadata):
...
-    }
-    xhr.open('GET', REMOTE_METADATA_NAME, true);
-    xhr.overrideMimeType('application/json');
-    xhr.send(null);
+    fetch(REMOTE_METADATA_NAME)
+    .then(response => {
+      if(response.ok) {
+        return response.json();
+      }
+      throw new Error(response.statusText + ' : ' + response.url);
```
sbc100, line 1130: Would `return Promise.reject(new Error(response.statusText + ' : ' + response.url))` instead of throwing?

src/web_or_worker_shell_read.js
sbc100: How about `return new Promise.reject(..)`

Then I think maybe you can do `.then(onload, onerror)` without the extra catch?   Not sure which is more idomatic.

tools/file_packager.py
seanmorris: In this case, we need to `catch()` before `then()` so we only handle fetch errors here. If we putt this at the bottom, or in the second param of  `this(),` then it will also act on the network errors generated by `then(),` and I'd need to add logic to skip certain errors.

seanmorris: `throw` isn't a function, so it needs to have braces or it counts as a syntax error. I'm switching to `Promise.reject` which **_is_** a function.

seanmorris:
> If possible, I'd like to see the `fetchAsync` change land first and the file_packager change as followup.
> 
> The `fetchAsync` is going to be a lot more risky, since it effects basically all users whereas the file packager change is something that only some users will be effected by.

No problem! https://github.com/emscripten-core/emscripten/pull/22015

seanmorris:
> If possible, I'd like to see the `fetchAsync` change land first and the file_packager change as followup.
> 
> The `fetchAsync` is going to be a lot more risky, since it effects basically all users whereas the file packager change is something that only some users will be effected by.

https://github.com/emscripten-core/emscripten/pull/22015
https://github.com/emscripten-core/emscripten/pull/22016

Let me know how those look.

sbc100:
Now that #22015 and #22016 contains the rest of this PR I think we can close this one?

Issue #22001, opened 2024-05-25
--emit-tsd does not output function argument names


jdarpinian:
When using --emit-tsd, all of the function arguments in the resulting .d.ts file are named _0, _1, etc. They have correct types, but the real argument names are important to me. Am I missing a compiler flag that would enable outputting argument names, or is this not yet implemented?

jdarpinian:
I guess I see the problem. Embind gets the argument types by template metaprogramming, but getting the argument names would require C++ reflection which is not standardized yet. Besides waiting for that (hopefully within our lifetimes?) I guess other options would be writing a clang plugin, or parsing DWARF information (when present).

jdarpinian:
I see that Embind actually has an undocumented feature for this! If you include a parenthesized argument list in the name string passed to embind::function, it will be used to supply argument names in the .d.ts file. However it doesn't handle types in the argument list, it has to be a list of names without types. I'll try making a PR to accept types in this list too, and check that the types match the real types.

PR #22009, opened 2024-05-28
[tsgen] Check parameter types match

jdarpinian:
When using embind::function() with --emit-tsd, parameter names for TypeScript definitions cannot be detected automatically and must be supplied manually in string form in the function name. This is error-prone as it duplicates the function prototype but in a form that the compiler can't type-check. This change adds some macros which allow specifying the parameter names and types in a form that the compiler can type-check. This will reduce the likelihood of parameters in embind::function() declarations getting out of sync with the functions they expose.

This change also reduces the burden of maintaining the parameter names, as one can simply copy and paste directly from the function prototype without modification. Previously you were required to edit out the types from the function prototype manually.

Addresses #22001

### DIFF
--- a/src/embind/embind_shared.js
+++ b/src/embind/embind_shared.js
@@ -145,7 +145,7 @@ var LibraryEmbindShared = {
 #if ASSERTIONS
       assert(signature[signature.length - 1] == ")", "Parentheses for argument names should match.");
 #endif
-      return signature.substr(argsIndex, signature.length - argsIndex - 1).replaceAll(" ", "").split(",").filter(n => n.length);
+      return signature.substr(argsIndex, signature.length - argsIndex - 1).match(/\w+(?=\s*(,|$))/g) || [];
     } else {
       return [];
     }
--- a/system/include/emscripten/bind.h
+++ b/system/include/emscripten/bind.h
@@ -593,7 +593,57 @@ void function(const char* name, ReturnType (*fn)(Args...), Policies...) {
         isAsync<Policies...>::value);
 }
 
+#define EMSCRIPTEN_FUNCTION(name, fn, args) \
+    ([]() { \
+        auto lambda = +[]args{}; \
+        static_assert(::emscripten::internal::are_function_pointer_args_same<decltype(fn), decltype(lambda)>::value, "Argument list does not match function signature"); \
+       ::emscripten::function(name#args, fn); \
+    })()
+
+#define EMSCRIPTEN_FUNCTION_INLINE(return_type, name, args, body) \
+    ::emscripten::function(name#args, +[]args -> return_type body);
+
+#define EMSCRIPTEN_MEMBER_FUNCTION(name, fn, args) \
+    function((([]() { \
+        auto lambda = +[]args{}; \
+        static_assert(::emscripten::internal::are_function_pointer_args_same<decltype(fn), decltype(lambda)>::value, "Argument list does not match function signature"); \
+    })(), name#args), fn)
+
+#define EMSCRIPTEN_CLASS_FUNCTION(name, fn, args) \
+    class_function((([]() { \
+        auto lambda = +[]args{}; \
+        static_assert(::emscripten::internal::are_function_pointer_args_same<decltype(fn), decltype(lambda)>::value, "Argument list does not match function signature"); \
+    })(), name#args), fn)
+
 namespace internal {
+
+template<typename T>
+struct function_traits;
+
+template<typename Ret, typename... Args>
+struct function_traits<Ret(*)(Args...)> {
+    using return_type = Ret;
+    using argument_types = std::tuple<Args...>;
+};
+
+template<typename Ret, typename ClassType, typename... Args>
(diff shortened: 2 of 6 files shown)


system/include/emscripten/bind.h
```diff
@@ -593,8 +593,58 @@ void function(const char* name, ReturnType (*fn)(Args...), Policies...) {
...
+
+template<typename Ret, typename ClassType, typename... Args>
+struct function_traits<Ret(ClassType::*)(Args...)> {
+    using return_type = Ret;
+    using argument_types = std::tuple<Args...>;
+};
+
+// Helper type to compare function argument types
+template<typename Func1, typename Func2>
+struct are_function_pointer_args_same {
```
sbc100, line 643: Is there a reason for using `function_pointer_args` here rather than just `function_args`?  Are we only looking at pointer args?

src/embind/embind_shared.js
```diff
@@ -145,7 +145,7 @@ var LibraryEmbindShared = {
 #if ASSERTIONS
       assert(signature[signature.length - 1] == ")", "Parentheses for argument names should match.");
 #endif
-      return signature.substr(argsIndex, signature.length - argsIndex - 1).replaceAll(" ", "").split(",").filter(n => n.length);
+      return signature.substr(argsIndex, signature.length - argsIndex - 1).match(/\w+(?=\s*(,|$))/g) || [];
```
sbc100, line 148: If you we use `.slice` here we can use `-1` as the second argument.

sbc100, line 148: Can you explain what this regex is doing? Perhaps in a comment on the line above?

Do we need to worry about complex cases where the itself contains a comma (e.g. `std::pair<int, int>`)?   This kind of case seems extremely tricky.

For this reason I wonder if we should reconsider and just accept a list of names, without types?   (especially since we don't actually need the types themselves here do we?  (i.e. the types comes via a different mechanism, right?)

system/include/emscripten/bind.h
jdarpinian: Yes, this actually compares only function pointers. I will update the comments.

src/embind/embind_shared.js
sbc100: If we drop the types then we can also drop the type checking. 

That only leaves the minor issue or not being able to copy and paste directly, which I don't think is such a large advantage to warrant this much churn and a potentially breaking change (since we might already have users of these argument names, no?)

jdarpinian: It's not a breaking change as omitting the types is still allowed when using emscripten::function directly. The existing test cases all still pass, unmodified. Also I did a search on GitHub and didn't find a single case of anyone using this feature, probably because it's undocumented.

The type checking is one of my main motivations for this change. I am maintaining bindings for a C++ library. When I update the bindings to a new version of the library I want compile errors when existing functions change, so that I can be sure to update the bindings. Otherwise they are sure to get out of date.

sbc100: To be clear the types themselves will automatically updated right?   The part that needs to be manually updated is that names for the arguments?     Would it be enough to perhaps just check the number of arguments?     That might be enough to catch a lot of common cases where the names need updating?

sbc100: > Also I did a search on GitHub and didn't find a single case of anyone using this feature, probably because it's undocumented.

I see.. but to be clear if anyone was using this feature then this change would break their build?  Is that right?

jdarpinian: > if anyone was using this feature then this change would break their build? Is that right?

No, that is incorrect. Again, the existing tests pass unmodified.

jdarpinian: > Would it be enough to perhaps just check the number of arguments? That might be enough to catch a lot of common cases where the names need updating?

Yes it will catch some cases. Type checking will catch more.

Obviously ideally this would check the actual argument names from the function's own implementation. Unfortunately this would require C++ reflection and it is not clear when or if that will ever be available.

brendandahl:
I was hoping we could find a way to work around this issue and avoid having to manually duplicate all the argument names. Some potential ideas to do this in an automated fashion:

- use the dwarf debug info
- use a clang plugin

I'd like to do a bit more research to see if either of those options (or something else) are viable before we add a whole new duplicated API to handle this. I do think both are going to be challenging to go from a function pointer to finding the function name though.

jdarpinian:
> * use the dwarf debug info
> * use a clang plugin

Yes, those are the other possibilities I mentioned in issue #22001. I did look into dwarf a bit. I agree that it would be pretty complex to map a function pointer to a parameter list, and of course it wouldn't work when dwarf isn't generated, and there may be other gotchas. I didn't find a good JS dwarf library either so it would probably need a custom tool built in some other language.

I think it's quite likely a clang plugin could do the job with fewer caveats. I didn't look further into it for several reasons. Currently I'm thinking I will probably build a single-purpose libclang-based tool to generate my bindings rather than try to upstream improvements of this type to emscripten.

Issue #22003, opened 2024-05-26
Can't preload files/shared libs in service worker due to lack of XHR [PR OPEN]


seanmorris:
Service workers no longer allow XHR-based requests (not sure if they ever did) so we can't load file packages or shared libraries without using `fetch()`. **The following PR allows the user to compile with `-sUSE_FETCH` to correct this:**

# https://github.com/emscripten-core/emscripten/pull/22002

PR #22016, opened 2024-05-30
Use fetch() for loading file packages.

seanmorris:
This PR switches AJAX requests for file packages to be loaded with fetch() rather than XmlHttpRequest(). This allows service workers to make AJAX requests, since XmlHttpRequest is not available in service workers.

Previously, --embed-file was necessary to load file packages, which embeds the files into the binary and precludes caching and re-use of the package.

Partially fixes https://github.com/emscripten-core/emscripten/issues/22003

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -2270,7 +2270,7 @@ class BrowserCore(RunnerCore):
       if reporting == Reporting.FULL:
         # If C reporting (i.e. the REPORT_RESULT macro) is required we
         # also include report_result.c and force-include report_result.h
-        self.run_process([EMCC, '-c', '-I' + TEST_ROOT,
+        self.run_process([EMCC, '-c', '-fPIC', '-I' + TEST_ROOT,
                           '-DEMTEST_PORT_NUMBER=%d' % self.port,
                           test_file('report_result.c')] + self.get_emcc_args(compile_only=True))
         args += ['report_result.o', '-include', test_file('report_result.h')]
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -699,7 +699,8 @@ If manually bisecting:
           <center><canvas id='canvas' width='256' height='256'></canvas></center>
           <hr><div id='output'></div><hr>
           <script type='text/javascript'>
-            window.onerror = (error) => {
+            window.addEventListener('unhandledrejection', event => {
+              const error = String(event.reason);
               window.disableErrorReporting = true;
               window.onerror = null;
               var result = error.includes("test.data") ? 1 : 0;
@@ -707,7 +708,7 @@ If manually bisecting:
               xhr.open('GET', 'http://localhost:8888/report_result?' + result, true);
               xhr.send();
               setTimeout(function() { window.close() }, 1000);
-            }
+            });
             var Module = {
               locateFile: function (path, prefix) {if (path.endsWith(".wasm")) {return prefix + path;} else {return "''' + assetLocalization + r'''" + path;}},
               print: (function() {
@@ -5548,6 +5549,133 @@ Module["preRun"] = () => {
       self.run_process(shared.get_npm_cmd('webpack') + ['--mode=development', '--no-devtool'])
     shutil.copyfile('webpack/src/hello.wasm', 'webpack/dist/hello.wasm')
     self.run_browser('webpack/dist/index.html', '/report_result?exit:0')
+
+  def test_fetch_polyfill_preload(self):
+    path = 'hello-world.txt'
+    create_file(path, 'hello, world!')
+    create_file('main.cpp', r'''
+      #include <stdio.h>
+      #include <string.h>
+      #include <emscripten.h>
+      int main() {
+        FILE *f = fopen("%s", "r");
+        char buf[100];
+        fread(buf, 1, 20, f);
+        buf[20] = 0;
+        fclose(f);
+        printf("|%%s|\n", buf);
+        return 0;
+      }
+      ''' % path)
+    create_file('on_window_error_shell.html', r'''
+      <html>
+          <center><canvas id='canvas' width='256' height='256'></canvas></center>
+          <hr><div id='output'></div><hr>
+          <script type='text/javascript'>
+            window.addEventListener('error', event => {
+              const error = String(event.message);
+              console.log({error});
+              window.disableErrorReporting = true;
+              window.onerror = null;
+              var xhr = new XMLHttpRequest();
+              xhr.open('GET', 'http://localhost:8888/report_result?' + error, true);
+              xhr.send();
+              setTimeout(function() { window.close() }, 1000);
+            });
+            var Module = {
+              print: (function() {
+                var element = document.getElementById('output');
+                return function(text) {
+                  console.log({text});
+                  if(window.disableErrorReporting) return;
+                  element.innerHTML += text.replace('\n', '<br>', 'g') + '<br>';
+                  var xhr = new XMLHttpRequest();
+                  xhr.open('GET', 'http://localhost:8888/report_result?' + (text === '|hello, world!|' ? 1 : 0), true);
+                  xhr.send();
+                };
+              })(),
+              canvas: document.getElementById('canvas')
+            };
+          </script>
+          {{{ SCRIPT }}}
+        </body>
+      </html>''')
+
+    def test(args, expect_fail):
+      self.compile_btest('main.cpp', args + ['--preload-file', path, '--shell-file', 'on_window_error_shell.html', '-o', 'a.out.html'])
+      js = read_file('a.out.js')
+      if expect_fail:
+        create_file('a.out.js', 'fetch = undefined;\n' + js)
+        return self.run_browser('a.out.html', '/report_result?TypeError: fetch is not a function')
+      else:
+         return self.run_browser('a.out.html', '/report_result?1')
+
+    test([], expect_fail=True)
+    test(['-sLEGACY_VM_SUPPORT'], expect_fail=False)
+    test(['-sLEGACY_VM_SUPPORT', '-sNO_POLYFILL'], expect_fail=True)
+
+  def test_fetch_polyfill_shared_lib(self):
+    create_file('library.c', r'''
+      #include <stdio.h>
+      int library_func() {
+        return 42;
+      }
+    ''')
+    create_file('main.c', r'''
+      #include <assert.h>
+      #include <dlfcn.h>
+      #include <stdio.h>
+      #include <emscripten.h>
+      int main() {
+        int found = EM_ASM_INT(
+          return preloadedWasm['/library.so'] !== undefined;
+        );
+        void *lib_handle = dlopen("/library.so", RTLD_NOW);
+        typedef int (*voidfunc)();
+        voidfunc x = (voidfunc)dlsym(lib_handle, "library_func");
+        printf("Got val: %d\n", x());
+        assert(x() == 42);
+        return 0;
+      }
+    ''')
+    create_file('on_window_error_shell.html', r'''
+      <html>
+          <center><canvas id='canvas' width='256' height='256'></canvas></center>
+          <hr><div id='output'></div><hr>
+          <script type='text/javascript'>
+            var Module = {
+              print: (function() {
+                var element = document.getElementById('output');
+                return function(text) {
+                  if(window.disableErrorReporting) return;
+                  element.innerHTML += text.replace('\n', '<br>', 'g') + '<br>';
+                  var result = text.includes("42") ? 1 : 0;
+                  var xhr = new XMLHttpRequest();
+                  xhr.open('GET', 'http://localhost:8888/report_result?' + result, true);
+                  xhr.send();
+                };
+              })(),
+              canvas: document.getElementById('canvas')
+            };
+          </script>
+          {{{ SCRIPT }}}
+        </body>
+      </html>''')
+
+    self.run_process([EMCC, 'library.c', '-sSIDE_MODULE', '-O2', '-o', 'library.so'])
+
+    def test(args, expect_fail):
+      self.compile_btest('main.c', ['library.so', '-sMAIN_MODULE', '--shell-file', 'on_window_error_shell.html', '-o', 'a.out.html'])
+      js = read_file('a.out.js')
+      if expect_fail:
+        create_file('a.out.js', 'fetch = undefined;\n' + js)
+        return self.run_browser('a.out.html', '/report_result?abort:TypeError')
+      else:
+         return self.run_browser('a.out.html', '/report_result?1')
+
+    test([], expect_fail=True)
+    test(['-sLEGACY_VM_SUPPORT'], expect_fail=False)
+    test(['-sLEGACY_VM_SUPPORT', '-sNO_POLYFILL'], expect_fail=True)
 
 
 class emrun(RunnerCore):
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -359,7 +359,7 @@ def generate_object_file(data_files):
 
 def main():
   if len(sys.argv) == 1:
-    err('''Usage: file_packager TARGET [--preload A [B..]] [--embed C [D..]] [--exclude E [F..]]] [--js-output=OUTPUT.js] [--no-force] [--use-preload-cache] [--indexedDB-name=EM_PRELOAD_CACHE] [--separate-metadata] [--lz4] [--use-preload-plugins]
+    err('''Usage: file_packager TARGET [--preload A [B..]] [--embed C [D..]] [--exclude E [F..]]] [--js-output=OUTPUT.js] [--no-force] [--use-preload-cache] [--indexedDB-name=EM_PRELOAD_CACHE] [--separate-metadata] [--lz4] [--use-preload-plugins] [--no-node]
   See the source for more details.''')
     return 1
 
@@ -775,7 +775,7 @@ def generate_js(data_target, data_files, metadata):
       }
       var REMOTE_PACKAGE_NAME = Module['locateFile'] ? Module['locateFile'](REMOTE_PACKAGE_BASE, '') : REMOTE_PACKAGE_BASE;\n''' % (js_manipulation.escape_for_js_string(data_target), js_manipulation.escape_for_js_string(remote_package_name))
     metadata['remote_package_size'] = remote_package_size
-    ret += '''var REMOTE_PACKAGE_SIZE = metadata['remote_package_size'];\n'''
+    ret += '''      var REMOTE_PACKAGE_SIZE = metadata['remote_package_size'];\n'''
 
     if options.use_preload_cache:
       # Set the id to a hash of the preloaded data, so that caches survive over multiple builds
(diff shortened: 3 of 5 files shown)


test/common.py
```diff
@@ -2266,7 +2266,7 @@ def compile_btest(self, filename, args, reporting=Reporting.FULL):
       if reporting == Reporting.FULL:
         # If C reporting (i.e. the REPORT_RESULT macro) is required we
         # also include report_result.c and force-include report_result.h
-        self.run_process([EMCC, '-c', '-I' + TEST_ROOT,
+        self.run_process([EMCC, '-c', '-fPIC', '-I' + TEST_ROOT,
```
sbc100, line 2269: This can be reverted now

test/test_browser.py
```diff
@@ -5541,6 +5541,66 @@ def test_webpack(self):
...
+      </html>''')
+
+    def test(args, expect_fail):
+      self.compile_btest('main.cpp', args + ['--preload-file', path, '--shell-file', 'on_window_error_shell.html', '-o', 'a.out.html'])
+      js = read_file('a.out.js')
+      if expect_fail:
+        create_file('a.out.js', 'fetch = undefined;\n' + js)
+        return self.run_browser('a.out.html', '/report_result?1')
+      else:
+         return self.run_browser('a.out.html', '/report_result?1')
```
sbc100, line 5602: Indentation looks wrong here.

test/test_browser.py
```diff
@@ -5541,6 +5541,66 @@ def test_webpack(self):
...
+              canvas: document.getElementById('canvas')
+            };
+          </script>
+          {{{ SCRIPT }}}
+        </body>
+      </html>''')
+
+    def test(args, expect_fail):
+      self.compile_btest('main.cpp', args + ['--preload-file', path, '--shell-file', 'on_window_error_shell.html', '-o', 'a.out.html'])
+      js = read_file('a.out.js')
```
sbc100, line 5597: Move this line down inside the `if` block

tools/file_packager.py
```diff
@@ -965,59 +965,82 @@ def generate_js(data_target, data_files, metadata):
...
-        }
-        xhr.onload = function(event) {
-          if (xhr.status == 200 || xhr.status == 304 || xhr.status == 206 || (xhr.status == 0 && xhr.response)) { // file URLs can return 0
-            var packageData = xhr.response;
-            callback(packageData);
-          } else {
-            throw new Error(xhr.statusText + " : " + xhr.responseURL);
+
+          // If we're using the polyfill, readers won't be available...
+          if(!response.body && response.arrayBuffer) {
```
sbc100, line 985: Space after `if`

sbc100: Would it be enough to just check `!response.body` here?

tools/file_packager.py
```diff
@@ -965,59 +965,82 @@ def generate_js(data_target, data_files, metadata):
...
-              num++;
-            }
-            total = Math.ceil(total * Module.expectedDataFileDownloads/num);
-            if (Module['setStatus']) Module['setStatus'](`Downloading data... (${loaded}/${total})`);
-          } else if (!Module.dataFileDownloads) {
-            if (Module['setStatus']) Module['setStatus']('Downloading data...');
+        Module.dataFileDownloads = Module.dataFileDownloads || {};
+        const url = packageName;
+        fetch(url)
+        .catch(cause => Promise.reject(new Error('Network Error : ' + url, {cause}))) // If fetch fails, rewrite the error to include the failing URL & the cause.
```
sbc100, line 975: Is this a common way to format `.catch` and `.then` method that follow a call like `fetch` (i.e. not indented at all)

tools/file_packager.py
```diff
@@ -965,59 +965,82 @@ def generate_js(data_target, data_files, metadata):
...
+
+          const iterate = () => reader.read().then(handleChunk).catch(cause => {
+            return Promise.reject(new Error('Unexpected error while handling : ' + response.url + ' ' + cause, {cause}));
+          });
+
+          const handleChunk = ({done, value}) => {
+            if (!done) {
+              chunks.push(value);
+              loaded += value.length;
+              Module.dataFileDownloads[url] = Module.dataFileDownloads[url] ?? {};
```
sbc100, line 1004: How about `Module.dataFileDownloads[url] |= {}`?

tools/file_packager.py
```diff
@@ -965,59 +965,82 @@ def generate_js(data_target, data_files, metadata):
...
-      };
 
+          const reader = response.body.getReader();
+          const headers = response.headers;
+
+          const size = headers.get('Content-Length') ?? packageSize;
+          const chunks = [];
+
+          const iterate = () => reader.read().then(handleChunk).catch(cause => {
+            return Promise.reject(new Error('Unexpected error while handling : ' + response.url + ' ' + cause, {cause}));
```
sbc100, line 997: Use JS string formatting here and above?  (using backticks)

tools/file_packager.py
```diff
@@ -1113,15 +1136,14 @@ def generate_js(data_target, data_files, metadata):
...
-    xhr.onreadystatechange = function() {
-     if (xhr.readyState === 4 && xhr.status === 200) {
-       loadPackage(JSON.parse(xhr.responseText));
-     }
-    }
-    xhr.open('GET', REMOTE_METADATA_NAME, true);
-    xhr.overrideMimeType('application/json');
-    xhr.send(null);
+    fetch(REMOTE_METADATA_NAME)
+    .then(response => {
```
sbc100, line 1140: Space after `if`

tools/file_packager.py
```diff
@@ -1113,15 +1136,14 @@ def generate_js(data_target, data_files, metadata):
...
+      if(response.ok) {
+        return response.json();
+      }
+      return Promise.reject(new Error(response.statusText + ' : ' + response.url));
```
sbc100, line 1144: I notices when debugging that chrome doesn't always set `statusText`.  Can we just use status instead?

test/test_browser.py
```diff
@@ -5541,6 +5541,66 @@ def test_webpack(self):
...
+                  var xhr = new XMLHttpRequest();
+                  xhr.open('GET', 'http://localhost:8888/report_result?' + (text === '|hello, world!|' ? 1 : 0), true);
+                  xhr.send();
+                };
+              })(),
+              canvas: document.getElementById('canvas')
+            };
+          </script>
+          {{{ SCRIPT }}}
+        </body>
```
sbc100, line 5592: Rather than require special shell can you just rely on the fact that default reporting will try bot `fetch` and `origFetch`: https://github.com/emscripten-core/emscripten/blob/a6e1f5c0792393059510430831d908f74e13e1ce/test/browser_reporting.js#L16

i.e. when you disable `fetch` you can cache the original fetch so that that browser reporting doesn't break.

sbc100 (commented):
lgtm modulo a few comments

sbc100:
@seanmorris do you have some time up update this PR?  I'd love to see it landing if possible.

test/common.py
sbc100: I think this should be good to land if you revert this line?

tools/file_packager.py
sbc100: It seems that prettier lines to indent the `.catch` and `.then`.  Can you do the same here?

e.g.

```
fetch(url)
  .catch(...)
  .then(...)
```

sbc100, line 975: It seems like this `catch` line here is just for debugging .. is it really needed in production do you think?

seanmorris:
> @seanmorris do you have some time up update this PR? I'd love to see it landing if possible.

@sbc100 I've been super busy over the last couple months but let me know what you think!

test/test_browser.py
```diff
@@ -5382,6 +5382,58 @@ def test_webpack(self, es6):
     shutil.copyfile('webpack/src/hello.wasm', 'webpack/dist/hello.wasm')
     self.run_browser('webpack/dist/index.html', '/report_result?exit:0')
 
+  def test_fetch_polyfill_preload(self):
+    path = 'hello-world.txt'
+    create_file(path, 'hello, world!')
+    create_file('main.cpp', r'''
```
sbc100, line 5388: Can you use `main.c` instead of `main.cpp` here?

sbc100 (commented):
lgtm % a few more comments.  Thanks for working on this!

test/test_browser.py
seanmorris: @sbc100 I'm not able to get the exception out of the browser without the custom shell. It just hangs with the exception listed as uncaught in the browser's console.

tools/file_packager.py
```diff
@@ -952,54 +952,75 @@ def generate_js(data_target, data_files, metadata):
...
+                }
+
+                if (totalSize) {
+                  if (Module['setStatus']) Module['setStatus'](`Downloading data... (${totalLoaded}/${totalSize})`);
+                }
+                else {
+                  if (Module['setStatus']) Module['setStatus']('Downloading data...');
+                }
+                return iterate();
+              }
```
sbc100, line 1009: `else` on the same line as the closing brace.

tools/file_packager.py
```diff
@@ -952,54 +952,75 @@ def generate_js(data_target, data_files, metadata):
...
+
+                for (const dowload of Object.values(Module.dataFileDownloads)) {
+                  totalLoaded += dowload.loaded;
+                  totalSize += dowload.total;
+                }
+
+                if (totalSize) {
+                  if (Module['setStatus']) Module['setStatus'](`Downloading data... (${totalLoaded}/${totalSize})`);
+                }
+                else {
```
sbc100, line 1005: `else` on the same line as the closing brace.

tools/file_packager.py
```diff
@@ -952,54 +952,75 @@ def generate_js(data_target, data_files, metadata):
...
+                  if (Module['setStatus']) Module['setStatus']('Downloading data...');
+                }
+                return iterate();
+              }
+              else {
+                const size = chunks.map((c) => c.length).reduce((a, b) => a + b, 0);
+                let index = 0;
+                const packageData = new Uint8Array(size);
+                for(const chunk of chunks) {
```
sbc100, line 1014: Space after `for` keyword.

tools/file_packager.py
```diff
@@ -952,54 +952,75 @@ def generate_js(data_target, data_files, metadata):
...
+                  if (Module['setStatus']) Module['setStatus'](`Downloading data... (${totalLoaded}/${totalSize})`);
+                }
+                else {
+                  if (Module['setStatus']) Module['setStatus']('Downloading data...');
+                }
+                return iterate();
+              }
+              else {
+                const size = chunks.map((c) => c.length).reduce((a, b) => a + b, 0);
+                let index = 0;
```
sbc100, line 1012: Maybe call this `offset`?

tools/file_packager.py
```diff
@@ -952,54 +952,75 @@ def generate_js(data_target, data_files, metadata):
...
+                if (totalSize) {
+                  if (Module['setStatus']) Module['setStatus'](`Downloading data... (${totalLoaded}/${totalSize})`);
+                }
+                else {
+                  if (Module['setStatus']) Module['setStatus']('Downloading data...');
+                }
+                return iterate();
+              }
+              else {
+                const size = chunks.map((c) => c.length).reduce((a, b) => a + b, 0);
```
sbc100, line 1011: Maybe call this `totalSize`?   Shouldn't this always be the same as `packageSize`?

tools/file_packager.py
```diff
@@ -952,54 +952,75 @@ def generate_js(data_target, data_files, metadata):
...
-              Module['dataFileDownloads'][url].loaded = event.loaded;
+        Module.dataFileDownloads ??= {};
+        fetch(packageName)
+          .catch((cause) => Promise.reject(new Error(`Network Error: ${packageName}`, {cause}))) // If fetch fails, rewrite the error to include the failing URL & the cause.
+          .then((response) => {
+
+            let loaded = 0;
+
+            if (!response.ok) {
+              return Promise.reject(new Error(`${response.status}: ${response.url}`));
```
sbc100, line 967: I wonder if we should be calling `errback` instead here?   Although it looks like the existing code doesn't so maybe we can just leave it?

tools/file_packager.py
```diff
@@ -952,54 +952,72 @@ def generate_js(data_target, data_files, metadata):
...
+
+                for (const dowload of Object.values(Module.dataFileDownloads)) {
+                  totalLoaded += dowload.loaded;
+                  totalSize += dowload.total;
+                }
+
+                if (totalSize) {
+                  if (Module['setStatus']) Module['setStatus'](`Downloading data... (${totalLoaded}/${totalSize})`);
+                } else {
+                  if (Module['setStatus']) Module['setStatus']('Downloading data...');
```
sbc100, line 1005: I wonder if we have tests for this `setStatus` callback.. can you confirm it gets called with roughly the same frequency as before?

tools/file_packager.py
```diff
@@ -952,54 +952,72 @@ def generate_js(data_target, data_files, metadata):
...
+                let totalLoaded = 0;
+                let totalSize = 0;
+
+                for (const dowload of Object.values(Module.dataFileDownloads)) {
+                  totalLoaded += dowload.loaded;
+                  totalSize += dowload.total;
+                }
+
+                if (totalSize) {
+                  if (Module['setStatus']) Module['setStatus'](`Downloading data... (${totalLoaded}/${totalSize})`);
```
sbc100, line 1003: Since we are modernizing this JS I think you can write this as `Module['setStatus'].?(...)`

sbc100: Sorry `?.()`

sbc100:
I think this change LGTM now, pending confirmation that the progress callbacks still work as expected.

(Can you mark as resolves the discussions that are resolved above?)

tools/file_packager.py
```diff
@@ -952,54 +952,72 @@ def generate_js(data_target, data_files, metadata):
...
-        };
-        xhr.send(null);
+
+            const reader = response.body.getReader();
+            const headers = response.headers;
+
+            const size = headers.get('Content-Length') ?? packageSize;
+            const chunks = [];
+
+            const iterate = () => reader.read().then(handleChunk).catch((cause) => {
```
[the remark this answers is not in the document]
seanmorris: If I recall correctly the browser will call this roughly every few kilobytes. I think it maps to the size of the chunks after the `Transfer-encoding: chunked` header from the server but I need to double check.

tools/file_packager.py
```diff
@@ -952,54 +952,73 @@ def generate_js(data_target, data_files, metadata):
...
+              return Promise.reject(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}));
+            });
+
+            const handleChunk = ({done, value}) => {
+              if (!done) {
+                chunks.push(value);
+                loaded += value.length;
+                Module.dataFileDownloads[packageName] = Module.dataFileDownloads[packageName] ?? {};
+                Module.dataFileDownloads[packageName].loaded = loaded;
+                Module.dataFileDownloads[packageName].total = size;
```
sbc100, line 992: Can't you just do `Module.dataFileDownloads[packageName] = {loaded, total: size}` here?    There are not other fields are there?

sbc100: Also I think `dataFileDownloads` needs to be in quotes here.

tools/file_packager.py
```diff
@@ -952,54 +952,73 @@ def generate_js(data_target, data_files, metadata):
...
+
+                for (const dowload of Object.values(Module.dataFileDownloads)) {
+                  totalLoaded += dowload.loaded;
+                  totalSize += dowload.total;
+                }
+
+                if (totalSize) {
+                  Module['setStatus']?.(`Downloading data... (${totalLoaded}/${totalSize})`);
+                } else {
+                  Module['setStatus']?.('Downloading data...');
```
sbc100, line 1005: I think this initial `setStatus` can happen  down on line 1019, then you don't need the if/else here.

tools/file_packager.py
seanmorris: @sbc100 I'd rather just create the object once and then re-use the fields, instead of re-creating it and garbage-collecting the same two fields over and over again.

tools/file_packager.py
```diff
@@ -952,54 +952,70 @@ def generate_js(data_target, data_files, metadata):
...
+
+            const iterate = () => reader.read().then(handleChunk).catch((cause) => {
+              return Promise.reject(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}));
+            });
+
+            const handleChunk = ({done, value}) => {
+              if (!done) {
+                chunks.push(value);
+                loaded += value.length;
+                Module['dataFileDownloads'][packageName] = Module['dataFileDownloads'][packageName] ?? {};
```
sbc100, line 990: I think you can use `?? = {}` here

tools/file_packager.py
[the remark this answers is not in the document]
sbc100: Is this true for the XHR onprogress callback too?   i.e. we are going to regress users who expect progress callback but don't use `Transfer-encoding: chunked`?

tools/file_packager.py
```diff
@@ -952,54 +952,70 @@ def generate_js(data_target, data_files, metadata):
...
+                Module['dataFileDownloads'][packageName].loaded = loaded;
+                Module['dataFileDownloads'][packageName].total = size;
+
+                let totalLoaded = 0;
+                let totalSize = 0;
+
+                for (const dowload of Object.values(Module['dataFileDownloads'])) {
+                  totalLoaded += dowload.loaded;
+                  totalSize += dowload.total;
```
sbc100, line 999: Is this supposed to be `download`?

tools/file_packager.py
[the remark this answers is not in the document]
seanmorris: For a file of 13631488 bytes **without** the `Transfer-Encoding` header, I get the following chunk sizes (bytes):

2097152
65536
393216
720896
917504
1310720
786432
1310720
196608
65536
65536
458752
2097152
851968
1245184
1048576

tools/file_packager.py
seanmorris: @sbc100 I get roughly 16 calls for a 13 megabyte file, how does that compare to baseline?

sbc100: I don't know I'm afraid.

tools/file_packager.py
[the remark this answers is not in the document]
seanmorris: @sbc100  I checked out the `main` branch and hacked the `onprogress` event to tell me its chunk sizes. For the same 13_631_488 byte file,  I get two events,  with 1_048_576 and 12_582_912 bytes each.

tools/file_packager.py
seanmorris: Looks like this solution is _more_ granular than the existing one. Is that acceptable?

sbc100: Sounds good to me!

sbc100: Could you add that information the PR description?

tools/file_packager.py
```diff
@@ -952,54 +952,71 @@ def generate_js(data_target, data_files, metadata):
...
+            });
+
+            const handleChunk = ({done, value}) => {
+              if (!done) {
+                chunks.push(value);
+                loaded += value.length;
+
+                Module['dataFileDownloads'][packageName] ??= {};
+                Module['dataFileDownloads'][packageName].loaded = loaded;
+                Module['dataFileDownloads'][packageName].total = size;
```
sbc100, line 993: I still think `Module['dataFileDownloads'][packageName] = { loaded, total: size }` is smaller and more readable these 3 lines.   I don't think we need to worry to much about a tiny amount of JS garbage here during the file download phase.

seanmorris: @sbc100 Is there anything that could grab a reference to the objects in `Module['dataFileDownloads']` that could come out-of-sync if we start swapping the objects?

tools/file_packager.py
```diff
@@ -952,54 +952,62 @@ def generate_js(data_target, data_files, metadata):
...
+
+                for (const download of Object.values(Module['dataFileDownloads'])) {
+                  totalLoaded += download.loaded;
+                  totalSize += download.total;
+                }
+
+                Module['setStatus']?.(`Downloading data... (${totalLoaded}/${totalSize})`);
+                return iterate();
+              } else {
+                const packageData = new Uint8Array(chunks.reduce((a, b) => a.concat(b), []));
```
seanmorris, line 1003: @sbc100 Thanks for having me revisit this section. I eliminated the second pass.

sbc100 (approved):
Thanks for all the back and forth on this!  I think this is ready to land now!

tools/file_packager.py
seanmorris: NICE!!!

seanmorris:
> Thanks for all the back and forth on this! I think this is ready to land now!

NICE!!!

PR #22021, opened 2024-05-30
Simplify a couple of interactive browser tests. NFC

sbc100:
These tests don't need to do a separate compile and run step when they can both just use the btest helper.


sbc100:
These were the only two uses of `compile_test` outside of `test_browser.py`.

dschuff (approved):

PR #22015, opened 2024-05-30
Use `fetch` rather than `XMLHttpRequest` for downloading Wasm files

seanmorris:
This PR switches requests for Wasm files to be loaded with `fetch()` rather than `XMLHttpRequest()`.

This fixes running in service workers where `XMLHttpRequest` is not available.

Partially fixes https://github.com/emscripten-core/emscripten/issues/22003, only applies to Wasm files.

PR #22026, opened 2024-05-30
Re-enable XHR as fallback for fetch on webview

sbc100:
In #22015 we switched `readAsync` to be based on fetch rather than XHR.  However that broke the assumption in `getBinaryenPromise` that was assuming `readAsync` was XHR-based.

This change reinstates the XHR fallback in the case of webviews where loading from `file://` URLs is not only possible but (apparently) common.

Sadly we don't have any webview tests so I have no way to actually test this within our current frameworks.


kripken (commented):
Can we instead enable the Fetch polyfill for webview?

Unfortunately we support webview by default in ENVIRONMENT so this would bloat default builds, but perhaps that can be changed first?

sbc100:
> Can we instead enable the Fetch polyfill for webview?
> 
> Unfortunately we support webview by default in ENVIRONMENT so this would bloat default builds, but perhaps that can be changed first?

The thing is that fetch is supported in webviews, its just the fetch doesn't support file URLs (even in webviews).   I don't think it would be good idea idea to clobber the existing webview.

Also use the fetch polyfill would be a bigger bloat that what is this PR proposes I believe.

sbc100:
> Can we instead enable the Fetch polyfill for webview?
> 
> Unfortunately we support webview by default in ENVIRONMENT so this would bloat default builds, but perhaps that can be changed first?

This current change is actually a codesize win even for default ENVIRONMENT that includes webiew:

```
$ ./emcc ~/test/hello.c -O3 --closure=1 -o new.js
$ wc -c old.js new.js 
 6051 old.js
 6023 new.js
```

sbc100:
For some of our smallest codesize tests this change is actually at 20% saving!

seanmorris:
We could probably implement WebView tests with NodeGTK & WebKit2 GTK: https://www.npmjs.com/package/node-gtk https://webkitgtk.org/

I used it to implement a browser using Vrzno as a bridge from Js to php-wasm in node: https://github.com/seanmorris/php-gtk

sbc100:
> We could probably implement WebView tests with NodeGTK & WebKit2 GTK: https://www.npmjs.com/package/node-gtk https://webkitgtk.org/

Thanks!  That could be very useful.   I think the problem we have is that we really don't know how many users are targetting vebview in this way.  Its hard to know if its worth investing a lot of time developing a test harness :(


> 
> I used it to implement a browser using Vrzno as a bridge from Js to php-wasm in node: https://github.com/seanmorris/php-gtk

dschuff (approved):

seanmorris:
> > We could probably implement WebView tests with NodeGTK & WebKit2 GTK: https://www.npmjs.com/package/node-gtk https://webkitgtk.org/
> 
> Thanks! That could be very useful. I think the problem we have is that we really don't know how many users are targetting vebview in this way. Its hard to know if its worth investing a lot of time developing a test harness :(
> 
> > I used it to implement a browser using Vrzno as a bridge from Js to php-wasm in node: https://github.com/seanmorris/php-gtk

https://github.com/emscripten-core/emscripten/pull/22042

PR #21638, opened 2024-03-28
Update libcxx and libcxxabi to LLVM 18.1.2

aheejin:
This updates libcxx and libcxxabi to LLVM 18.1.2. Due to https://github.com/llvm/llvm-project/pull/65534 we need to update both at the same time.

Things to note:
- Disable hardening mode: https://github.com/emscripten-core/emscripten/pull/21638/commits/e508f3d51821d24ca91f2e1ad1deb514d3797912
   LLVM 18.1.2 adds support for "Hardening Modes" to libcxx (https://libcxx.llvm.org/Hardening.html). There are four modes: none, fast, extensive and debug. Different hardening modes make different trade-offs between the amount of checking and runtime performance. We for now disable it (i.e., set it to 'none') so that we don't have any effects on the performance. We can consider enabling it when we ever get to enable the debug version of libcxx.

- Disable C++20 time zone support: https://github.com/emscripten-core/emscripten/pull/21638/commits/df9af64eb3f36b56133abacb6a0687253bd9563c
   LLVM 18.1.2 adds C++20 time zone support (https://libcxx.llvm.org/DesignDocs/TimeZone.html) to libcxx, which requires access IANA Time Zone Database. Currently it seems it only supports Linux:
   https://github.com/llvm/llvm-project/blob/26a1d6601d727a96f4301d0d8647b5a42760ae0c/libcxx/src/tz.cpp#L45-L49
   So this excludes the two source files from build (which is done via `CMakeLists.txt` in the upstream LLVM) and sets `_LIBCPP_HAS_NO_TIME_ZONE_DATABASE` macro in `__config_site`. In future maybe we can consider implementing this in JS.

- `__cxa_init_primary_exception` support: https://github.com/emscripten-core/emscripten/pull/21638/commits/6a0909e75ab975a95c84a14bead86e2f3cadf60c and https://github.com/emscripten-core/emscripten/pull/21638/commits/a4ec944dc5cd15b3c5263355189ecca496f5650a
   https://github.com/llvm/llvm-project/pull/65534 introduces `__cxa_init_primary_exception` and uses this in libcxx. Like several other methods like the below in `cxa_exception.cpp`,
   https://github.com/emscripten-core/emscripten/blob/fbdd9249e939660cb0d20f00d6bc1897c2f3905e/system/lib/libcxxabi/src/cxa_exception.cpp#L264-L269
[...]

PR #22027, opened 2024-05-31
[EH] Add __cxa_init_primary_exception to cxa_noexception.cpp

aheejin:
tl;dr:

This adds `__cxa_init_primary_exception` to `cxa_noexception.cpp`.

Other targets build `cxa_exception.cpp` in `-fignore-exceptions` mode and build `cxa_noexception.cpp` only in `-fno-exeptions` mode. But we build `cxa_noexception.cpp` in `-fignore-exceptions` mode, which means it needs the definition for `__cxa_init_primary_exception` for linking to succeed when no EH argument is given (which means `-fignore-exceptions`).

---

Long version (Feel free to skip):

Background:

After #21638, `__cxa_init_primary_exception` was added in libcxxabi: https://github.com/emscripten-core/emscripten/blob/28c10a1c5e2862edadd68ab627478204ae96d134/system/lib/libcxxabi/src/cxa_exception.cpp#L209-L226 https://github.com/emscripten-core/emscripten/blob/28c10a1c5e2862edadd68ab627478204ae96d134/system/lib/libcxxabi/src/cxa_exception_emscripten.cpp#L155-L162

Currently the files containing `__cxa_init_primary_exception`, `cxa_exception.cpp` and `cxa_exception_emscripten.cpp`, are only compiled when any of the EH mode is specified. `cxa_exception.cpp` is compiled when Wasm EH is selected, and `cxa_exception_emscripten.cpp` is compiled when Emscripten EH is selected:
https://github.com/emscripten-core/emscripten/blob/28c10a1c5e2862edadd68ab627478204ae96d134/tools/system_libs.py#L1599-L1608

and this function is called from `make_exception_ptr` in libcxx: https://github.com/emscripten-core/emscripten/blob/28c10a1c5e2862edadd68ab627478204ae96d134/system/lib/libcxx/include/__exception/exception_ptr.h#L87-L99

And `make_exception_ptr` is called from `std::promise`'s destructor: https://github.com/emscripten-core/emscripten/blob/28c10a1c5e2862edadd68ab627478204ae96d134/system/lib/libcxx/include/future#L1161-L1168

---

Bug:

Currently any program that calls `std::promise`'s destructor without specifying any exception-related arguments fails, saying `undefined symbol: __cxa_init_primary_exception`. Not specifying any exception arguments, meaning not specifying any among `-fno-exceptions`, `-fwasm-exceptions`, `-fexceptions`, or
`-sDISABLE_EXCEPTION_CATCHING=0`, defaults to `-fignore-exceptions`, which allows throwing but not catching.

---

Analysis:

The callsite of `__cxa_init_primary_exception` in `make_exception_ptr` is guarded with `#ifndef _LIBCPP_HAS_NO_EXCEPTIONS`, so it seems it is supposed to be included only when exceptions are enabled. This `_LIBCPP_HAS_NO_EXCEPTIONS` is defined when `__cpp_exceptions` is defined:
https://github.com/emscripten-core/emscripten/blob/28c10a1c5e2862edadd68ab627478204ae96d134/system/lib/libcxx/include/__config#L644-L646

And that `__cpp_exceptions` is defined in clang, when `-fcxx-exceptions` is given:
https://github.com/llvm/llvm-project/blob/be566d2eacdaed972b90d2eeb1e66d732c9fe7c1/clang/lib/Frontend/InitPreprocessor.cpp#L638-L639

And that `-fcxx-exceptions` can be specified in command line, but it is also programmatically added here:
https://github.com/llvm/llvm-project/blob/be566d2eacdaed972b90d2eeb1e66d732c9fe7c1/clang/lib/Driver/ToolChains/Clang.cpp#L371-L388

You can see it is added by default unless the arch is XCore or PS4/PS5, which means for Wasm it has been always added so far, unless `-fno-exceptions` is explicitly specified.

So I tried checking the arguments there in Clang and adding `-fcxx-exceptions` only if either of `-fwasm-exceptions` or `-enable-emscripten-cxx-exceptions` is given. But this fails when none of EH is selected (which means `-fignore-exceptions`), because if `-fcxx-exceptions` is not added, we can't use keywords like `throw`.

So basically the problem is, other targets build `cxa_exception.cpp` in `-fignore-exceptions` mode and build `cxa_noexception.cpp` only in `-fno-exeptions` mode. But we build `cxa_noexception.cpp` in `-fignore-exceptions` mode, which means it needs the definition for `__cxa_init_primary_exception` for linking to succeed, because `_LIBCPP_HAS_NO_EXCEPTIONS` cannot be defined in `-fignore-exceptions` (because we couldn't disable `-fcxx-exceptions` above in Clang to use `throw`)

So this adds `__cxa_init_primary_exception` to `cxa_noexception.cpp`.


sbc100 (approved):
I assume this change is probably suitable for upstreaming?

aheejin:
Not sure. I've been upstreaming the changes that are likely necessary for upstream CMake users, and unless they use our `system_libs.py`, this is not gonna be necessary.

Also, `cxa_noexceptions.cpp` contains other emscripten-only functions too. Do you think we should upstream all these?
https://github.com/emscripten-core/emscripten/blob/28c10a1c5e2862edadd68ab627478204ae96d134/system/lib/libcxxabi/src/cxa_noexception.cpp#L52-L77

mzukovec:
Thanks for looking into this!

sbc100:
> Not sure. I've been upstreaming the changes that are likely necessary for upstream CMake users, and unless they use our `system_libs.py`, this is not gonna be necessary.
> 
> Also, `cxa_noexceptions.cpp` contains other emscripten-only functions too. Do you think we should upstream all these?
> 
> https://github.com/emscripten-core/emscripten/blob/28c10a1c5e2862edadd68ab627478204ae96d134/system/lib/libcxxabi/src/cxa_noexception.cpp#L52-L77

I guess my question is, is this `__cxa_init_primary_exception` (or the need for it) emscripten-specific?   How come calls to it are being generated by the wasm backend but not other backends?

aheejin:
> I guess my question is, is this `__cxa_init_primary_exception` (or the need for it) emscripten-specific? How come calls to it are being generated by the wasm backend but not other backends?

Calls to `__cxa_init_primary_exception` are generated in other backends too. But they use `cxa_exceptions.cpp` in case of `-fignore-exceptions` so it's not a problem. They use `cxa_noexceptions.cpp` only when `-fno-exceptions` is given, in which case `_LIBCPP_HAS_NO_EXCEPTIONS` is defined and the calls to `__cxa_init_primary_exception` is not compiled:
https://github.com/emscripten-core/emscripten/blob/28c10a1c5e2862edadd68ab627478204ae96d134/system/lib/libcxx/include/__exception/exception_ptr.h#L87-L99

But we use `cxa_noexception.cpp` in case of `-fignore-exceptions`, which is our default. That's why I added the function there.

sbc100:
> > I guess my question is, is this `__cxa_init_primary_exception` (or the need for it) emscripten-specific? How come calls to it are being generated by the wasm backend but not other backends?
> 
> Calls to `__cxa_init_primary_exception` are generated in other backends too. But they use `cxa_exceptions.cpp` in case of `-fignore-exceptions` so it's not a problem. They use `cxa_noexceptions.cpp` only when `-fno-exceptions` is given, in which case `_LIBCPP_HAS_NO_EXCEPTIONS` is defined and the calls to `__cxa_init_primary_exception` is not compiled:
> 
> https://github.com/emscripten-core/emscripten/blob/28c10a1c5e2862edadd68ab627478204ae96d134/system/lib/libcxx/include/__exception/exception_ptr.h#L87-L99
> 
> But we use `cxa_noexception.cpp` in case of `-fignore-exceptions`, which is our default. That's why I added the function there.

I see.. so the issue stems from the different/special way that we handle `-fignore-exceptions`.  Might be worth adding a comment here to that effect.

aheejin:
> I see.. so the issue stems from the different/special way that we handle `-fignore-exceptions`. Might be worth adding a comment here to that effect.

#22034

PR #22031, opened 2024-05-31
WebGPU: Implement wgpuAdapterGetInfo on GPUAdapter.info

beaufortfrancois:
This PR adds support for the upcoming GPUAdapter info attribute in Chromium. You can try the WIP CL at https://chromium-review.googlesource.com/c/chromium/src/+/5458182

image


beaufortfrancois:
@kainino0x Can you review when you have time? 

**FIXME:** I don't know how to properly call `_free()` on string pointers in `wgpuAdapterInfoFreeMembers`

system/include/webgpu/webgpu_cpp.h
```diff
@@ -1214,6 +1216,24 @@ namespace wgpu {
...
+        ~AdapterInfo();
+        AdapterInfo(const AdapterInfo&) = delete;
+        AdapterInfo& operator=(const AdapterInfo&) = delete;
+        AdapterInfo(AdapterInfo&&);
+        AdapterInfo& operator=(AdapterInfo&&);
+        ChainedStructOut  * nextInChain = nullptr;
+        char const * const vendor = nullptr;
+        char const * const architecture = nullptr;
+        char const * const device = nullptr;
+        char const * const description = nullptr;
```
sbc100, line 1230: Would it make sense to have fixed length strings here?  That way you could use some kind of static buffer on the C/C++ side and JS could fill in the values without needing to call malloc?

sbc100: Alternatively if you want to keep this mechanism, then a common way to do this would be to have JS-reused the string buffer to consecutive calls (e.g. use `realloc` of a global pointer), so that no `free` is ever needed (since successive calls don't allocate).

beaufortfrancois: Thanks for the review @sbc100!

1. I'm not sure about the fixed length strings here. This info can't change for the life the adapter but it's not clear to me how to implement this suggestion.

2. I didn't find any `_realloc()` call in Emscripten ;(
Where would you store the global pointer? Something like `LibraryWebGPU.$WebGPU.vendorPtr` where `malloc` would be called in the first place and then realloc?

sbc100: Re fixed length I was thinking to use arrays rather than pointers. e.g:

`char vendor[MAX_INFO_STR];`

That way no separate allocations are needed.

Regarding `_realloc`, you would just need to add it to the `__deps` of the JS function in order to use it.    Yes you could use something like `vendorPtr`.   You could also try to put all the strings into one single buffer with a single allocation.   Sounds like fixed arrays would be simpler though.

beaufortfrancois: @kainino0x Any preference?

kainino0x: In standardization for webgpu.h, we chose to use allocated strings instead of fixed-length char arrays (IIRC - it seems I forgot to take notes on the topic in https://github.com/webgpu-native/webgpu-headers/issues/143). We could pick a max length and use char arrays for vendor/architecture/device, but it's hard to pick a length for `description` since it can be pretty long - for example in GL, layered implementations like ANGLE typically wrap their own description around the underlying implementation's description, so the string gets longer with each layer. It's a little unfortunate for emscripten because JS will almost never expose a description at all in WebGPU, but in the general case it makes sense. We also have other entry points that will use the same pattern to return arrays of enums, etc., that we probably don't want to have to stabilize fixed max sizes for, so it'll be a pattern.

kainino0x: We also didn't want to have pointers invalidated subtly by other calls, as would happen if we made them live until next overwrite. For that, plus reasons relating to implementing webgpu.h in rust, we also didn't want to require the strings to be owned by the parent object either (even if they would be immutable).

beaufortfrancois: Thank you @kainino0x for the background.

I've updated this PR with `realloc`. Let me know if that looks good to you

system/lib/webgpu/webgpu.cpp
```diff
@@ -81,3 +81,12 @@ void wgpuSurfaceGetCapabilities(WGPUSurface surface,
...
   // wgpuSurfaceCapabilities doesn't currently allocate anything.
 }
+
+// WGPUAdapterInfo
+
+void wgpuAdapterInfoFreeMembers(WGPUAdapterInfo value) {
+  delete[] value.vendor;
+  delete[] value.architecture;
+  delete[] value.device;
+  delete[] value.description;
```
kainino0x, line 91: In C++, `malloc` must be used with `free` and `new` must be used with `delete`. So to match the use of `_malloc` from JS, these must use `free` instead of `delete`.

src/library_webgpu.js
```diff
@@ -2536,6 +2538,39 @@ var LibraryWebGPU = {
...
+  wgpuAdapterGetInfo: (adapterId, info) => {
+    var adapter = WebGPU.mgrAdapter.get(adapterId);
+    {{{ gpu.makeCheckDescriptor('info') }}}
+
+    function allocateUTF8String(stringPtr, stringValue) {
+      var stringSize = lengthBytesUTF8(stringValue) + 1;
+      if (!stringPtr) {
+        stringPtr = _malloc(stringSize);
+      } else {
+        _realloc(stringPtr, stringSize)
```
kainino0x, line 2551: The `wgpuAdapterGetInfo` function isn't supposed to do anything with incoming pointers, only overwrite them with new pointers. Using realloc will free the old pointers, which isn't supposed to happen.

Removing realloc I think makes this helper equivalent to `stringToNewUTF8`.

kainino0x: Oh sorry, I see the incoming pointers are coming from `adapterWrapper` not from `info`.

Same point though - we're never supposed to implicitly free these allocations, the caller is responsible for doing that by calling `wgpuAdapterInfoFreeMembers`.

We should not need `infoPtrs` at all.

beaufortfrancois: Oh I didn't know `stringToNewUTF8` existed. It will simplify code for sure. Thanks!

system/lib/webgpu/webgpu.cpp
beaufortfrancois: I'm not sure how to call `free` in cpp with `value.vendor` and so on.

system/include/webgpu/webgpu_cpp.h
juj: If it might be of help, in juj/wasm_webgpu I implemented `wgpu_adapter_get_info()` via no-allocation fields like this:

https://github.com/juj/wasm_webgpu/blob/ddf33db3b8c625061b97f89a43a7bf23b2fa9826/lib/lib_webgpu.js#L860-L875

beaufortfrancois: Thanks @juj for jumping into this thread. I don't see `wgpuAdapterInfoFreeMembers` in juj/wasm_webgpu. Is that intended?

juj: Yeah, since the memory is caller allocated, there is no need to call a separate function to free up memory.

kainino0x: I can raise the idea of using arrays instead of having a Free step with other webgpu.h folks. For now let's stick with what we have written down.

system/lib/webgpu/webgpu.cpp
kainino0x: Should just be `free(value.vendor);`
where `free` comes from `#include <cstdlib>`

src/library_webgpu.js
kainino0x: I just searched for the file `stringToUTF8` was defined in and found this, wasn't sure if you'd seen it because it is so similar to what you did :)

system/lib/webgpu/webgpu.cpp
beaufortfrancois: That's what I tried but this resulted in the following error:

```
../../../system/lib/webgpu/webgpu.cpp:89:3: error: no matching function for call to 'free'
   89 |   free(value.vendor);
      |   ^~~~
/Users/fbeaufort/github/emscripten/cache/sysroot/include/sched.h:85:6: note: candidate function not viable: no known conversion from 'const char *' to 'void *' for 1st argument
   85 | void free(void *);
```

sbc100: I think its safe to `const_cast` that away.

kainino0x: Ah... yes, you'll need to use `const_cast` to make the pointer non-const.

beaufortfrancois: Thanks!

src/library_webgpu.js
beaufortfrancois: And we're good now.

beaufortfrancois:
@kainino0x I believe we're in good shape now.

kainino0x (approved):
LGTM!

kainino0x:
Summary I wrote into the squash-and-merge commit:
> Implements `wgpuAdapterGetInfo()`, the C equivalent of `GPUAdapter.info`.
Uses `GPUAdapter.info` (not `GPUAdapter.requestAdapterInfo()`), which is
new in the WebGPU spec, so this requires an updated browser.
> 
> `wgpuAdapterGetProperties()` still exists and still doesn't return any real info.

PR #22035, opened 2024-05-31
Convert readAsync to return a promise. NFC

sbc100:
Followup to #22026

Since all the call sites are mostly promise based it makes logical sense to return a promise here.


dschuff (approved):

src/preamble.js
```diff
@@ -648,15 +648,12 @@ function getBinaryPromise(binaryFile) {
...
-          try { resolve(getBinarySync(binaryFile)); }
-          catch (e) { reject(e); }
-        });
-    });
+    // Fetch the binary using readAsync
+    return readAsync(binaryFile).then(
+      (response) => new Uint8Array(/** @type{!ArrayBuffer} */(response)),
+      // Fall back to getBinarySync if readAsync fails
+      () => getBinarySync(binaryFile)
+    ).catch(() => getBinarySync(binaryFile));
```
kripken, line 656: Why does `getBinarySync` appear twice here?

sbc100:
here you can see another example of the macOS test running timing out

PR #22042, opened 2024-06-02
Webview tests

seanmorris:


.circleci/config.yml
```diff
@@ -796,6 +839,17 @@ jobs:
...
+  test-browser-webview:
+    executor: bionic
+    environment:
+      EMTEST_LACKS_WEBGPU: "1"
+    steps:
+      - run-tests-webview:
+          title: "browser"
+          # Skip test_4gb_fail as it OOMs on the current bot
+          test_targets: "
+            browser skip:browser.test_4gb_fail
```
sbc100, line 851: Do we want to be running all browser tests in this mode or should we just run a select few?

sbc100 (commented):
I wonder if we can write a specific test for loading from file:// URLs since that is the only different that I currently know of between webview and non-webview browsers.  i.e. browser cannot fetch load wasm files from file:// URls but apparently electroc apps can?

.circleci/config.yml
```diff
@@ -796,6 +833,23 @@ jobs:
           # some native-dependent tests fail because of the lack of native
           # headers on emsdk-bundled clang
           test_targets: "other skip:other.test_native_link_error_message"
+  test-browser-webview:
+    # We don't use `bionic` here since its tool old to run recent node versions:
```
kripken, line 837:
```suggestion
    # We don't use `bionic` here since it's too old to run recent node versions:
```

.circleci/config.yml
kripken: As this entire suite is large (and as a result slow) I would support picking a select few to run.

Issue #22040, opened 2024-06-02
`::dup2(fd, -1)` does not return -1 or set `errno`


hly2019:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
3.1.54

**Failing command line in full:**
```shell
# wasm:
emcc dup2.cpp -o main.js 
node main.js

# native:
g++ dup2.cpp -o main
./main
```

Hi, it seems the return value of `int dup2(int oldfd, int newfd)` is not correct and the `errno` is not set when the `newfd` is -1, as shown in the code and results below.

According to https://man7.org/linux/man-pages/man2/dup.2.html#RETURN_VALUE, `dup2` should return -1 and set `errno` on error, and so it does with native execution. 
But in wasm execution, the return value is not -1, and the `errno` is not set when the second argument, `new_fd` is -1. This issue may lead to problems with error checking. 

I'd appreciate it so much if you could check it. Thanks!

#### Code:
Here shows the case I faced. 
I also tried the case when the `fd` is got from calling `::open` with a regular file (compiled with `--preload-file`), and the same problem also happens.
```c++
#include <unistd.h>
#include <string.h>
#include <stdio.h>
#include <errno.h>
#include <fcntl.h>
int main() {
    int fds[2];
    pipe(fds);
    int fd = fds[0];
    // int fd = ::open("temp.txt", O_RDWR); // compile with --preload-file temp.txt
    int ret = ::dup2(fd, -1); // the problem remains if here the fd is from ::open with regular file
    printf("ret of dup2: %d, error num: %d, error info: %s\n", ret, errno, strerror(errno));
    return 0;
}
```

#### Results:
In wasm, the `errno` is not set and the return value is also not -1, which is not as expected.

```shell
# in wasm:
ret of dup2: 5, error num: 0, error info: No error information

# in native:
ret of dup2: -1, error num: 9, error info: Bad file descriptor
```

PR #22048, opened 2024-06-03
Fix dup2 when newfd is invalid

sbc100:
Fixes: #22040

### DIFF
--- a/src/library_syscall.js
+++ b/src/library_syscall.js
@@ -1009,6 +1009,8 @@ var SyscallsLibrary = {
     assert(!flags);
 #endif
     if (old.fd === newfd) return -{{{ cDefs.EINVAL }}};
+    // Check newfd is within range of valid open file descriptors.
+    if (newfd < 0 || newfd >= FS.MAX_OPEN_FDS) return -{{{ cDefs.EBADF }}};
     var existing = FS.getStream(newfd);
     if (existing) FS.close(existing);
     return FS.dupStream(old, newfd).fd;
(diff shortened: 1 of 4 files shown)


src/library_syscall.js
```diff
@@ -1009,6 +1009,8 @@ var SyscallsLibrary = {
     assert(!flags);
 #endif
     if (old.fd === newfd) return -{{{ cDefs.EINVAL }}};
+    // Check newfd is within range of valid open file descriptors.
+    if (newfd < 0 || newfd > 1024) return -{{{ cDefs.EBADF }}};
```
kripken, line 1013: Is 1024 a hardcoded limit in musl? Can we perhaps use a `cDefine` for it rather than hardcode it here?

src/library_syscall.js
```diff
@@ -1009,6 +1009,8 @@ var SyscallsLibrary = {
...
     var existing = FS.getStream(newfd);
```
kripken, line 1014: Actually, shouldn't this line error if `newfd` is invalid? That would be a more precise check than the maximum id of an fd.

sbc100: `FS.getStream()` returns undefined is the stream is not currently open its not supposed to throw an error.    For that we have `FS.getStreamChecked()` but we don't want to use that here since we want to allows for the case where newfs is not open.

I think this is a reasonable place to do that check since this is the only place where the user can supply "proposed" file descriptor that is not already open (AFAIK).

src/library_syscall.js
sbc100: Looks like already had FS.MAX_OPEN_FDS

kripken (approved):

PR #21785, opened 2024-04-18
Avoid creating JS symbols for symbols only used in dynamic linking

sbc100:
Symbols that are exported using EMSCRIPTEN_KEEPALIVE are supposed           
to be exported to the outside world (i.e. on the Module object) and         
also be available to call JS within the module.                          
                                                                         
Symbols exports for the purposed of dynamic linking so not need to be       
exported on the Module and are added (at runtime) to `wasmImports` which 
acts as a kind of global symbol table for the program.                   
                                                                         
In in the case of `-sMAIN_MODULE=1` we export *all* symbols from         
all libraries, and prior to this change it was not possible to           
distingish between all the exported generated because of                 
`--export-dynamic`, and the exports generated due to                     
`EMSCRIPTEN_KEEPALIVE`.                                                  
                                                                         
This change allows us to differentiate by running `wasm-ld` twice: once  
without `--export-dynamic` (to get the smaller list of                   
`EMSCRIPTEN_KEEPALIVE`) and then once with `--export-dynamic` to         
produce the actual wasm that we output.                                  
                                                                         
This takes the list of exports that we turn in to JS globals from 7993   
to 28, massively reducing the overhead of `-sMAIN_MODULE=1`.

PR #22053, opened 2024-06-04
FIX Fill in GOT entries when libs are loaded with allowUndefined

### LINKED ISSUE #22052: Emscripten calls null GOT entry when side module loaded with `allowUndefined: true`
The following code example looks up `GOT.__cxa_throw` which is `0` and then calls it which throws `TypeError: getWasmTableEntry(...) is not a function`. Calling `reportUndefinedSymbols` or removing `allowUndefined: true` resolves the crash.

### module.cpp
```cpp
#include <stdexcept>

extern "C"
int cpp_func() { 
    try { 
        throw std::runtime_error(""); // crashes trying to throw this
    } catch (const std::exception &e) {
        return 0;
    }
    return 1;
} 
```

### main.c
```C
#include <dlfcn.h>
#include "stdio.h"
#include "emscripten.h"

typedef int (*InitFunc)(void);

int main(void) {
    EM_ASM({
        loadDynamicLibrary("module.so", { allowUndefined: true });
    });

    void* handle = dlopen("module.so", RTLD_NOW);
    printf("handle: %p\n", handle);
    if (!handle) {
        return 1;
    }
    InitFunc init_func = (InitFunc)dlsym(handle, "cpp_func");
    printf("init_func: %p\n", init_func);
    if (!init_func) {
        return 1;
    }
    int res = init_func(); // Crashes in init_func
    printf("res: %d\n", res);
    return 0;
}
```

### build.sh

```sh
em++ -c module.cpp -fPIC -fexceptions -g2 -O2 
em++ module.o -o module.so \
    -sWASM_BIGINT -sSIDE_MODULE=1 -fexceptions -g2 -O2

emcc -c main.c -fPIC -fexceptions -g2 -O2 
emcc main.o -o main.js \
    -sWASM_BIGINT -sMAIN_MODULE=1 -fexceptions -g2 -O2
```

Then run with `node main.js`.

Checked on 3.1.52, 3.1.58 and tip of tree:
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.62-git (0a0fa668c63cdaf14cfccb2ce62bc5afa46b4170)
clang version 19.0.0git (https:/github.com/llvm/llvm-project 0e62d5cf55479981da5e05e406bbca4afb3cdc4f)
Thread model: posix
InstalledDir: /home/rchatham/Documents/programming/emsdk/upstream/bin
Build config: +assertions
```

hoodmane:
Resolves #22052.


src/library_dylink.js
```diff
@@ -172,6 +172,16 @@ var LibraryDylink = {
...
+      var value = resolveGlobalSymbol(symName, true).sym;
+      if (!value) {
+        return rtn;
+      }
+      if (typeof value == 'function') {
+        /** @suppress {checkTypes} */
+        rtn.value = addFunction(value, value.sig);
+      } else if (typeof value == {{{ POINTER_JS_TYPE }}}) {
+        rtn.value = value;
+      }
```
sbc100, line 184: It looks like this code is very similar to existing code in both `reportUndefinedSymbols` and `updateGOT`.  I wonder if we can factor this better?

Also, I think there is probably some reason we delay updated the GOT at GOTHandler time and instead do it later on.    Perhaps we could instead modify the existing `updateGOT` function?

hoodmane: Sounds good I'll try factoring out the shared logic and moving the change to updateGot.

hoodmane:
@sbc100 I think ideally `reportUndefinedSymbols()` shouldn't adjust the GOT because the name makes it sound like it's for error detection/debugging. The fact that it can be necessary to keep things from breaking is unfortunate given the name.

sbc100:
> @sbc100 I think ideally `reportUndefinedSymbols()` shouldn't adjust the GOT because the name makes it sound like it's for error detection/debugging. The fact that it can be necessary to keep things from breaking is unfortunate given the name.

I agree.. we should probably split the `reportUndefinedSymbols` logic

src/library_dylink.js
hoodmane: > Is __cxa_throw effected because its a JS symbol?

The problem specifically seems to be that `updateGOT` fills in the GOT table for all wasm exports from the loaded library. So it misses JavaScript symbols exported from the main module. Perhaps it would fix it to add GOT entries for all of the JS-implemented library funcs in the main module?

sbc100: How about if we rename `reportUndefinedSymbols` to `handleUndefinedSymbols` and have it take an `reportErrors` flags?

hoodmane: Well it seems to me that the only reason we need this GOT update behavior in `reportUndefinedSymbols` is because of JS symbols from the main module which are not handled by `updateGOT`. So they only get added to the GOT if first we have a library that needs them. Why not just stick all JS symbols in the GOT at startup and not worry about having something like `handleUndefinedSymbols`?

sbc100: I don't quite remember the reason for the status quo,  but one reason might be that we want native symbols to take presence over JS symbols.  Another might be that we only want to populate the GOT with symbols that we actually need, not all JS symbols.    I'm just guessing though

hoodmane: There's not really any comments in the code about this, or any documentation about when and why it might be necessary to call `handleUndefinedSymbols`. Do we have a test for it? If not, we should make one.

But how do weak linkage symbols work? If I have `libone.so` and it exports a symbol with weak linkage and then `libtwo.so` exports the same symbol again, it doesn't seem like it gets replaced in the GOT. But [Wikipedia says](https://en.wikipedia.org/wiki/Weak_symbol):

> Even if two platforms support the same or similar syntax for marking symbols as weak, the semantics may differ in subtle points, e.g. whether weak symbols during dynamic linking at runtime lose their semantics or not.

So maybe this is compliant. How about just having a set of JS symbols and changing in `GOTHandler` if the current symbol is a JS symbol and if so filling it in.

sbc100: With weak symbols the first library that defines the symbol should win the race, and so I wouldn't expect any replacement in that that.

hoodmane: > During linking, a strong symbol can override a weak symbol of the same name. In contrast, in the presence of two strong symbols by the same name, the linker resolves the symbol in favor of the first one found.

So if you see first a weak definition and then later a strong one, shouldn't the second strong one be used over the weak one? They have some examples on the wikipedia page too.
https://en.wikipedia.org/wiki/Weak_symbol#Examples

sbc100: Oh sorry, yes, strong symbols should always win, although I'm not sure how important that edge case is.

hoodmane: But you think the case where a JS symbol is overridden by a native symbol is more important? Are the JS symbols automatically counted as weak?

sbc100: No, not necessarily.   I was just trying to imagine/remember the reason for factoring the code this way.

I do think we have good test coverage here,  so I think its reasonably safe to try refactoring this code and if the tests pass it should be fine.

sbc100: There is also a non-zero cost to adding JS functions to the wasm table.. see the `addFunction` helper.

hoodmane: Right, in the future where `WebAssembly.Function` is available it will be better. Is it possible we could just import and reexport all these JS functions from the main module so that they would all get added by `updateGOT` in the normal way?

hoodmane:
@sbc100 Could this be a regression from #21785? It seems like a regression from Pyodide 0.25 to 0.26 and we upgraded from Emscripten 3.1.52 to 3.1.58. In the changelog in that range, #21785 seems like the most likely culprit.

sbc100:
> @sbc100 Could this be a regression from #21785? It seems like a regression from Pyodide 0.25 to 0.26 and we upgraded from Emscripten 3.1.52 to 3.1.58. In the changelog in that range, #21785 seems like the most likely culprit.


I assume that by "this" you are referring to #22052?  Its hard to say right now since I don't fully understand the exact issue.  However I will say that #21785 pertains the native symbols and which of them get exported on the `Module`, whereas `#22052` appears to pertain to JS symbols, is that right?

sbc100:
> It seems like a regression from Pyodide 0.25 to 0.26 and we upgraded from Emscripten 3.1.52 to 3.1.58

Is it possible to confirm that for sure by upgrading just emscripten, and keeping all other things constant?  Even better could you bisect to find out exactly which emscripten change caused the regression?

hoodmane:
Okay it reproduces on 3.1.57. So maybe not a regression.

hoodmane:
Does not reproduce in 3.1.52, so indeed a regression since then. Will bisect.

hoodmane:
Regression was between 3.1.56 and 3.1.57.

hoodmane:
The regression was #21638. Reverting this on top of the main branch has no conflicts and makes the problem go away.

sbc100:
Did this PR get superseded or do we still want to work on landing it?

hoodmane:
I will try to check in a couple days but I think it's still a problem and it would be nice to work on landing a fix.

PR #22055, opened 2024-06-04
support to locateFile on initializing audioworklet

visla-jiaozebo:
as using audio worklet, the worker module (.aw.js file) needs to be loaded from the root directory. Perhaps it should be loaded like the wasm or worker, i.e. use locateFile to find its current path?


sbc100:
I think the problem here with the test failure is that `locateFile` is not available under `MINIMAL_RUNTIME`.

visla-jiaozebo:
> I think the problem here with the test failure is that `locateFile` is not available under `MINIMAL_RUNTIME`.

oh, we the outside developers could not see the test results.  sure the locateFile could be missing on some conditions which I don't know.

Issue #22050, opened 2024-06-04
std::time_put does not properly handle "%Ec" format


hly2019:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
3.1.54

**Failing command line in full:**
```shell
# emcc:
emcc locale_Ec.cpp  -o main.js
node main.js

# g++:
g++ locale_Ec.cpp  -o main
./main
```

Hi, it seems the `std::time_put` doesn't handle properly with the formatting `%Ec` in Emscripten. For example, the code shown below produces a result `"%c"`, which seems incorrect and meaningless. 

I'd appreciate it if you could please check this issue. Thank you!

#### Code
```c++
#include <sstream>
#include <locale>
#include <chrono>
#include <iostream>
#include <string.h>
int main() {
    std::wstring format = L"%Ec";
    auto t_now = std::chrono::time_point_cast<std::chrono::seconds>(std::chrono::system_clock::now());
    auto t = std::chrono::system_clock::to_time_t(t_now);
    std::tm tm = *std::gmtime(&t);
    auto loc = std::locale::classic();
    const std::time_put<wchar_t>& timeput = std::use_facet<std::time_put<wchar_t>>(loc);
    std::wostringstream os;
    os.imbue(loc);
    timeput.put(os, os, L' ', &tm, format.c_str(), format.c_str() + format.size());
    std::wcout << "error?: " << strerror(errno) << std::endl;
    std::wcout << "value of os.str: "<< os.str() << std::endl;
    return 0;
}
```

#### Results:

```shell
# in wasm:
error?: No error information
value of os.str: %c

# in native:
error?: Success
value of os.str: Tue Jun  4 03:56:58 2024
```

PR #22057, opened 2024-06-04
Fix support for %Ec in strftime

sbc100:
Fixes: #22050


sbc100:
I also fixed some bad indentation so please review with ignore-whitespace.

dschuff (approved):

PR #22067, opened 2024-06-06
Replace `.concat` with push and/or spread operator. NFC

sbc100:


src/library_openal.js
```diff
@@ -842,7 +842,7 @@ var LibraryOpenAL = {
       case {{{ cDefs.AL_VELOCITY }}}:
         return AL.currentCtx.listener.velocity;
       case {{{ cDefs.AL_ORIENTATION }}}:
-        return AL.currentCtx.listener.direction.concat(AL.currentCtx.listener.up);
+        return AL.currentCtx.listener.direction + AL.currentCtx.listener.up;
```
kripken, line 845: These are arrays so this changes the result, doesn't it? E.g.

```
> [0, 0, 0] + [1, 2, 3]
'0,0,01,2,3'
> [0, 0, 0].concat([1, 2, 3])
[ 0, 0, 0, 1, 2, 3 ]
```

sbc100: oops, reverted.

test/other/metadce/test_metadce_mem_O3_standalone_narg_flto.jssize
```diff
@@ -1 +1 @@
-4475
+4553
```
kripken, line 1: Odd that this increases code size?

sbc100: Interesting yes... maybe I'll split out the internal stuff and land that on its own.

PR #22068, opened 2024-06-06
Replace XMLHttpRequest with fetch in library_trace.js. NFC

sbc100:


sbc100:
I think the existing code here might actually be broken/untested since the loadWorkerViaXHR call it's ready callback with 2 arguments but the callsite passed a function that expects a single argument.

kripken (approved):

PR #22073, opened 2024-06-08
Important bugfixes to OpenAL implementation.

geneotech:
- The ``audiobuf`` in ``offset -= src.bufQueue[src.bufsProcessed].audiobuf.duration;`` is misspelled - ``audioBuf`` was likely intended. I found this because I was getting a ``TypeError: src.bufQueue[src.bufsProcessed] is undefined``, although I'm not 100% sure it was caused by the typo.
- ``switch`` cases for ``AL_SOURCE_DISTANCE_MODEL`` were checking against the string ``'AL_SOURCE_DISTANCE_MODEL'`` instead of the integer value ``0x200``, causing ``AL_INVALID_ENUM`` on ``alIsEnabled``, ``alEnable`` and ``alDisable`` with this enum.
- ``updateSourceTime`` sometimes leaves the src.bufOffset with a ``NaN`` value, although sadly I'm not sophisticated enough to understand why. I have added a harmless check that just sets it to 0 in case it turns ``NaN``, which arguably shouldn't break anything, and it completely fixes the following crash:
	```
	TypeError: AudioBufferSourceNode.start: Argument 2 is not a finite floating-point value.
		scheduleSourceAudio http://localhost:6931/Hypersomnia.js:8343
		setSourceState http://localhost:6931/Hypersomnia.js:8498
		sourceSeek http://localhost:6931/Hypersomnia.js:8853
		setSourceParam http://localhost:6931/Hypersomnia.js:9397
		_alSourcef http://localhost:6931/Hypersomnia.js:10089
		__emscripten_receive_on_main_thread_js http://localhost:6931/Hypersomnia.js:7102
	```
	I have already battle-tested this one ``if`` clause in [my browser game](https://github.com/TeamHypersomnia/Hypersomnia) and it seems to work.

### DIFF
--- a/test/openal_playback.cpp
+++ b/test/openal_playback.cpp
@@ -257,6 +257,11 @@ int main() {
 #elif defined(TEST_ANIMATED_LOOPED_DOPPLER_PLAYBACK)
   printf("You should hear a continuously looping clip of the 1902 piano song \"The Entertainer\" played back at a dynamic playback rate that smoothly varies its pitch according to a sine wave doppler shift. Press OK when confirmed.\n");
 #elif defined(TEST_ANIMATED_LOOPED_PANNED_PLAYBACK)
+  assert(!alIsEnabled(AL_SOURCE_DISTANCE_MODEL));
+  alDisable(AL_SOURCE_DISTANCE_MODEL);
+  assert(!alIsEnabled(AL_SOURCE_DISTANCE_MODEL));
+  alEnable(AL_SOURCE_DISTANCE_MODEL);
+  assert(alIsEnabled(AL_SOURCE_DISTANCE_MODEL));
   printf("You should hear a continuously looping clip of the 1902 piano song \"The Entertainer\" smoothly panning around the listener. Press OK when confirmed.\n");
 #elif defined(TEST_ANIMATED_LOOPED_RELATIVE_PLAYBACK)
   alSourcei(sources[0], AL_SOURCE_RELATIVE, AL_TRUE);
--- a/src/library_openal.js
+++ b/src/library_openal.js
@@ -726,7 +726,7 @@ var LibraryOpenAL = {
       if (src.bufQueue[src.bufsProcessed].audioBuf !== null) {
         src.bufsProcessed = 0;
         while (offset > src.bufQueue[src.bufsProcessed].audioBuf.duration) {
-          offset -= src.bufQueue[src.bufsProcessed].audiobuf.duration;
+          offset -= src.bufQueue[src.bufsProcessed].audioBuf.duration;
           src.bufsProcessed++;
         }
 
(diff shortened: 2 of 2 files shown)


sbc100 (commented):
Would it be possible at add a test for this or update one of the existing tests in `test/test_interactive.py`?

geneotech:
Hello! Do you mean for ``AL_SOURCE_DISTANCE_MODEL``? I'm afraid the other issues are too complex to be easily reproduced in a simple test.

sbc100:
> Hello! Do you mean for `AL_SOURCE_DISTANCE_MODEL`? I'm afraid the other issues are too complex to be easily reproduced in a simple test.

Whatever can be reasonably tested should be yes.   Perhaps the first 2 parts of this change at least should be fairly straight forward to test?

sbc100:
In fact, I wonder if we can split out the `updateSourceTime` into its own PR, since it seems a little more complex, and worthy of some deeper analysis, whereas the other two fixes look trivial.

For the `updateSourceTime` issue in particalar I'd be interested to know which of the assignments to `src.bufOffset` causes is used when its is NaN.   Are you able to catch it in devtools?  Can you see which of the three branches in `updateSourceTime` that assigned to `bufOutput` were taken?

sbc100 (approved):
LGTM.   If you are able to modify one of the existing tests to cover this code that would be great.    Either way I'm happy to see this change land.

geneotech:
Thank you! This test to ``alDisable``/``alEnable``/``alIsEnabled`` should cover it, the panning still seems to work. I'll make the separate PR with updateSourceTime only once I find the real cause.

test/openal_playback.cpp
```diff
@@ -257,6 +257,11 @@ int main() {
 #elif defined(TEST_ANIMATED_LOOPED_DOPPLER_PLAYBACK)
   printf("You should hear a continuously looping clip of the 1902 piano song \"The Entertainer\" played back at a dynamic playback rate that smoothly varies its pitch according to a sine wave doppler shift. Press OK when confirmed.\n");
 #elif defined(TEST_ANIMATED_LOOPED_PANNED_PLAYBACK)
+  assert(!alIsEnabled(AL_SOURCE_DISTANCE_MODEL));
+  alDisable(AL_SOURCE_DISTANCE_MODEL);
+  assert(!alIsEnabled(AL_SOURCE_DISTANCE_MODEL));
+  alEnable(AL_SOURCE_DISTANCE_MODEL);
+  assert(alIsEnabled(AL_SOURCE_DISTANCE_MODEL));
```
sbc100, line 264: Is there a reason you put this in the `TEST_ANIMATED_LOOPED_PANNED_PLAYBACK` block?  (Rather than up on line 247?)

geneotech: It seems that the block around 247 is a common one and I thought I was supposed to 

> update one of the existing tests

:)
Additionally I thought it might be a good idea to test if the option being enabled does not mess up the sound spatialization, while keeping the default for all the other tests.

sbc100: Fair enough.  LGTM.

PR #22083, opened 2024-06-11
Use spread operator over concat in internal JS code. NFC

sbc100:


kripken (approved):

sbc100:
> Isn't the spread operator super slow when the input size is large? Or we will practically never have big number of elements?

I think all the uses of the spread operator here is with arrays that are known to be very small.

PR #22084, opened 2024-06-11
[library_openal] Simplify string operations. NFC

sbc100:
Split out from #22067


kripken (approved):
lgtm but please run the interactive OpenAL tests locally to verify.

sbc100:
> lgtm but please run the interactive OpenAL tests locally to verify.

Done.  I added some test code too to get more coverage here.

Issue #22085, opened 2024-06-11
Since PR #21775, additional file packages error on load in modularize build when registered for preRun


past-due:
**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.58-git

**Description**:

Since PR #21775 (and Emscripten 3.1.58), the incoming `Module` is not modified in modularize builds.

Code that previously loaded additional file packages (by loading their .js) prior to instantiation of the compiled emscripten module - and worked in Emscripten < 3.1.58 - now fails. This path leads to the file package's `runWithFS()` being registered for `preRun`, but `runWithFS()` then apparently does not access the "actual" Module object which has `FS_createPath` (etc) added.

However, there appears to be a very simple fix:
`callRuntimeCallbacks` already passes the actual `Module`, so by using it in `runWithFS()`, the issue is resolved.

**Proposed PR: #22086**

EDIT: 
@sbc100 While I just noticed that PR #21775 was reverted more recently, the above PR seems to fix the issue for this use-case, and it certainly seems to me like runWithFS() should be using the passed-in Module argument in any case?

PR #22086, opened 2024-06-11
file_packager.py: Fix runWithFS() to use passed-in Module (issue since PR #21775)

past-due:
Fixes: #22085

`callRuntimeCallbacks` passes the actual `Module`, so use it

### DIFF
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -1094,12 +1094,12 @@ def generate_js(data_target, data_files, metadata):
       }\n'''
 
   ret += '''
-    function runWithFS() {\n'''
+    function runWithFS(Module) {\n'''
   ret += code
   ret += '''
     }
     if (Module['calledRun']) {
-      runWithFS();
+      runWithFS(Module);
     } else {
       if (!Module['preRun']) Module['preRun'] = [];
       Module["preRun"].push(runWithFS); // FS is not initialized yet, wait for it


tools/file_packager.py
```diff
@@ -1094,12 +1094,12 @@ def generate_js(data_target, data_files, metadata):
...
+    function runWithFS(Module) {\n'''
   ret += code
   ret += '''
     }
     if (Module['calledRun']) {
-      runWithFS();
+      runWithFS(Module);
     } else {
       if (!Module['preRun']) Module['preRun'] = [];
       Module["preRun"].push(runWithFS); // FS is not initialized yet, wait for it
```
sbc100, line 1105: What about this case?  Wouldn't we need to pass in the module here too?    Maybe we can make a closure here to capture the internal `Module` ?

past-due: `callRuntimeCallbacks` already passes the internal `Module` in to `preRun` (etc) callbacks, so we should not need to do anything else here?

Reference: https://github.com/emscripten-core/emscripten/blob/404fde04e762fbe12e90e9ca56463bdbb9730ab1/src/library.js#L2832-L2837

sbc100: Great!

sbc100 (approved):
Do you want to mark this a non-draft?

past-due:
> Do you want to mark this a non-draft?

Done 👍

PR #22087, opened 2024-06-11
Use JS optional chaining in more places. NFC

sbc100:
.


kripken (approved):
Looks like `??=` was added in exactly the Chrome/Firefox/Safari minimum versions we support by default, convenient...

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

(and `?.` was even earlier apparently)

Interesting it hurts gzip size in some cases but that seems like just noise.

RReverser:
> Looks like `??=` was added in exactly the Chrome/Firefox/Safari minimum versions we support by default, convenient...

It's because we bumped versions we support specifically to support that operator during the first back-refactor PR 😅

RReverser (approved):

kripken:
Oh cool :laughing: no coincidence then...

Issue #21072, opened 2024-01-12
WebGPU: Return useful information in wgpuAdapterGetProperties


austinEng:
Currently, the implementation returns empty strings in all of the vendorName, architecture, device, name, etc.. fields.
It should be implemented by forwarding the fields from [GPUAdapterInfo](https://gpuweb.github.io/gpuweb/#gpuadapterinfo)

@kainino0x

beaufortfrancois:
I'd be happy to implement this.

FYI I naively thought it would be as simple as something like this:

```diff
   wgpuAdapterGetProperties: (adapterId, properties) => {
-    {{{ gpu.makeCheckDescriptor('properties') }}}
-    {{{ makeSetValue('properties', C_STRUCTS.WGPUAdapterProperties.vendorID, '0', 'i32') }}};
-    {{{ makeSetValue('properties', C_STRUCTS.WGPUAdapterProperties.vendorName, '0', 'i32') }}};
-    {{{ makeSetValue('properties', C_STRUCTS.WGPUAdapterProperties.architecture, '0', 'i32') }}};
-    {{{ makeSetValue('properties', C_STRUCTS.WGPUAdapterProperties.deviceID, '0', 'i32') }}};
-    {{{ makeSetValue('properties', C_STRUCTS.WGPUAdapterProperties.name, '0', 'i32') }}};
-    {{{ makeSetValue('properties', C_STRUCTS.WGPUAdapterProperties.driverDescription, '0', 'i32') }}};
-    {{{ makeSetValue('properties', C_STRUCTS.WGPUAdapterProperties.adapterType, gpu.AdapterType.Unknown, 'i32') }}};
-    {{{ makeSetValue('properties', C_STRUCTS.WGPUAdapterProperties.backendType, gpu.BackendType.WebGPU, 'i32') }}};
-    {{{ makeSetValue('properties', C_STRUCTS.WGPUAdapterProperties.compatibilityMode, '0', 'i32') }}};
+    var adapter = WebGPU.mgrAdapter.get(adapterId);
+    {{{ runtimeKeepalivePush() }}}
+    adapter["requestAdapterInfo"]().then((info) => {
+      {{{ runtimeKeepalivePop() }}}
+      {{{ gpu.makeCheckDescriptor('properties') }}}
+      {{{ makeSetValue('properties', C_STRUCTS.WGPUAdapterProperties.vendorID, '0', 'i32') }}};
+
[...]

PR #22088, opened 2024-06-12
WebGPU: Add deprecation warning to wgpuAdapterGetProperties

beaufortfrancois:
As raised in https://github.com/emscripten-core/emscripten/issues/21072#issuecomment-2161823488, this PR adds a deprecation warning to wgpuAdapterGetProperties.


src/library_webgpu.js
```diff
@@ -2557,7 +2557,10 @@ var LibraryWebGPU = {
     {{{ makeSetValue('info', C_STRUCTS.WGPUAdapterInfo.deviceID, '0', 'i32') }}};
   },
 
+  wgpuAdapterGetProperties__deps: ['$warnOnce'],
   wgpuAdapterGetProperties: (adapterId, properties) => {
+    warnOnce('getProperties is deprecated, use getInfo instead');
```
kainino0x, line 2562:
```suggestion
    warnOnce('wgpuAdapterGetProperties is deprecated, use wgpuAdapterGetInfo instead');
```

kainino0x (approved):

src/library_webgpu.js
beaufortfrancois: FYI I was following convention of using short names as in https://github.com/emscripten-core/emscripten/blob/c0aa4038c73ed8f64007767638c2a1be1319d330/src/library_webgpu.js#L1987

kainino0x: Ah, we're definitely not consistent, but better to aim for clearer messages anyway.
(In the case of getMappedRange at least the name is quite specific. "getProperties" sounds very generic even though only one such thing exists.)

PR #22092, opened 2024-06-13
Fix asan.test_dlfcn_preload. NFC

sbc100:
This was a new test added in #21985.

This changes means that any symbols that libb.so needs are exported/preserved when building the main module.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -3980,7 +3980,7 @@ ok
     ''')
     self.build_dlfcn_lib('libb.c', outfile='libb.so', emcc_args=['liba.so'])
 
-    self.prep_dlfcn_main(['--preload-file', 'libb.so', '--use-preload-plugins'])
+    self.prep_dlfcn_main(['--preload-file', 'libb.so', '--use-preload-plugins', '-L.', '-sAUTOLOAD_DYLIBS=0', 'libb.so'])
     create_file('main.c', r'''
       #include <assert.h>
       #include <dlfcn.h>


kripken:
From the comment on `AUTOLOAD_DYLIBS` I only understand that it affects the runtime dynamic linking process. Is there more to it then? The docs say
```
// For MAIN_MODULE builds, automatically load any dynamic library dependencies
// on startup, before loading the main module.
```

sbc100:
> From the comment on `AUTOLOAD_DYLIBS` I only understand that it affects the runtime dynamic linking process. Is there more to it then? The docs say
> 
> ```
> // For MAIN_MODULE builds, automatically load any dynamic library dependencies
> // on startup, before loading the main module.
> ```

When building with `-sMAIN_MODULE=2` then the linker will keep alive any symbols needed by side modules.  So the documention for this is really under `-sMAIN_MODULE=2` :

https://github.com/emscripten-core/emscripten/blob/0c504193efb3d0b51d30c07895544b29cbad1950/src/settings.js#L1135-L1137

kripken (approved):

kripken:
Fix lgtm but I still don't quite understand the docs. Are you saying that `AUTOLOAD_DYLIBS=0` tells it to not load dynamic libraries automatically, and from that it infers that it does not need to keep alive the symbols those libraries need? That seems a little odd if so (if the dynamic libraries were provided on the commandline then I'd expect `autoload=0` to just affect how they are loaded, not whether they are taken into account otherwise).

sbc100:
> Fix lgtm but I still don't quite understand the docs. Are you saying that `AUTOLOAD_DYLIBS=0` tells it to not load dynamic libraries automatically, and from that it infers that it does not need to keep alive the symbols those libraries need? That seems a little odd if so (if the dynamic libraries were provided on the commandline then I'd expect `autoload=0` to just affect how they are loaded, not whether they are taken into account otherwise).

When you add shared libraries to the command line that normally does two things:

1. It tells the dynamic loader to load those libraries on the command line
2. It tells the linker to keep any needed symbols from the main module alive.

The `AUTOLOAD_DYLIBS=0` setting is a way to explicitly opt out of (1) while still allowing (2).

kripken:
I see, thanks. That all makes perfect sense.

I was still confused so I re-read the code and I think I misread
```
['--preload-file', 'libb.so', '--use-preload-plugins', '-L.', '-sAUTOLOAD_DYLIBS=0', 'libb.so']
```
`libb.so` appears twice, once as a preload and once so the compiler knows its symbols, and the second one is added - I think that is what I've been missing.

PR #18418, opened 2022-12-22
Add test for TypeIDs diverging between shared library and executable

torarnv:
When the shared library is built with symbol visibility hidden as the default, the type infos for any type is not available to anything that links to the library, and the library or executable will have its own copy of the type info, which is not merged.

The results in emscripten::internal::TypeID having different TYPEID values depending on which module does the lookup, which causes trouble for embind's type registry as that's indexed on the TYPEIDs.

Context: #16711

Issue #16711, opened 2022-04-13
Dynamic linking emscripten::val(..) fails in SIDE_MODULE


lpotter:
emscripten 3.1.7

Error message: BindingError: emval::as has unknown type N10emscripten3valE

Using emscripten::val("some text") in a SIDE_MODULE fails under some circumstances. Other uses of emscripten::val work, just not the c'tor.

Fails:
   emscripten::val rawPlatform = emscripten::val::global("navigator")["platform"];
    if (rawPlatform.call<bool>("includes", emscripten::val("Mac")))

Succeeds:
   emscripten::val rawPlatform = emscripten::val::global("navigator")["platform"];
    if (rawPlatform.call<bool>("includes", std::string("Mac")))

Both work fine in non dynamic linking build, or if invoked from MAIN_MODULE.

I tried a simple example app, but it was not failing there. But fails in a big project - Qt.

Also, we use  MODULARIZE and EXPORT_NAME

Any comments, hints, or suggestions welcome.

sbc100:
Would you mind posting a full example that fails?   I imagine this is because the static constructor generated by EMSCRIPTEN_BINDINGS is not being run when side module is loaded (for some reason).

lpotter:
The codebase in which it fails is rather large - Qt's qtbase. I have not been able to create a smaller example in which it fails. If you just want the failing binaries somewhere to run, I can upload them somewhere.

lpotter:
Here are the console messages with LIBRARY_DEBUG=1
https://pastebin.com/pUJGnG76

Last line being:
11:16:58.791 [library call:__emval_as: 0x8 (8),0x91c4e8 (9553128),0x565270 (5657200)] qtloader.js:396:25

and it is failing on .as<emscripten::val>();

kripken:
@lpotter 

I hope we get someone with embind expertise to look at more embind bugs soon, but for now, if this is important for you then I think a minimal testcase could make this much more accessible even for non-embind experts (like @sbc100 and myself). Specifically, it would be good to have a small testcase that can be built from source, that clearly shows the issue. Extracting the relevant parts out of your entire app into a small testcase would be what I recommend. (The best thing possible would be to extend an existing testcase in our test suite to show the problem.)

Separately, it may be worth trying some general things like [building with sanitizers](https://emscripten.org/docs/debugging/Sanitizers.html) - those can catch bugs both in your application and in emscripten system code.

lpotter:
[...]

PR #22095, opened 2024-06-14
Make embind work with -fvisibility=hidden

msorvig:
Export emscripten::val and emscripten::memory_view in order to prevent embind errors:

  BindingError: _emval_take_value has unknown type N10emscripten11memory_viewIhEE

Embind generates a numerical type id from the address of the std::type_info object resulting from evaluating a typeid expressiion (e.g. 'void *id = &typeid(T)').

However, C++ does not guarantee that this address is identical for all evaluations of the typeid expression. In practice it is when using static linking, but not when using dynamic linking when the libraries are
built with the '-fvisibility=hidden' compiler option.

The non-identical id's then cause embind to decide that types have not been registered when used from a library, since they have been registered with a different id by the main wasm module.

Exporting the types in question makes typeid addresses identical again, and fixes/works around the issue.

### DIFF
--- a/system/include/emscripten/val.h
+++ b/system/include/emscripten/val.h
@@ -285,7 +285,7 @@ private:
 static const char name##_symbol[] = #name;                          \
 static const ::emscripten::internal::symbol_registrar<name##_symbol> name##_registrar
 
-class val {
+class EMBIND_VISIBILITY_DEFAULT val {
 public:
   // missing operators:
   // * ~ - + ++ --
--- a/system/include/emscripten/wire.h
+++ b/system/include/emscripten/wire.h
@@ -24,6 +24,7 @@
 #include <string>
 
 #define EMSCRIPTEN_ALWAYS_INLINE __attribute__((always_inline))
+#define EMBIND_VISIBILITY_DEFAULT __attribute__((visibility("default")))
 
 #ifndef EMSCRIPTEN_HAS_UNBOUND_TYPE_NAMES
 #define EMSCRIPTEN_HAS_UNBOUND_TYPE_NAMES 1
@@ -430,7 +431,7 @@ constexpr bool typeSupportsMemoryView() {
 } // namespace internal
 
 template<typename ElementType>
-struct memory_view {
+struct EMBIND_VISIBILITY_DEFAULT memory_view {
     memory_view() = delete;
     explicit memory_view(size_t size, const ElementType* data)
         : size(size)
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -7718,6 +7718,36 @@ void* operator new(size_t size) {
     else:
       self.assertTrue(seen_lines.issuperset([6, 7, 11, 12]), seen_lines)
 
+  @needs_dylink
+  def test_embind_dylink_visibility_hidden(self):
+    # Check that embind is usable from a library built with "-fvisibility=hidden"
+
+    create_file('liblib.cpp', r'''
+      #include <emscripten/val.h>
+      #define EXPORT __attribute__((visibility("default")))
+      using namespace emscripten;
+      EXPORT void liba_fun() {
+        unsigned char buffer[1];
+        val view(typed_memory_view(1, buffer));
+      }
+    ''')
+    self.build_dlfcn_lib('liblib.cpp', emcc_args=['-fvisibility=hidden'])
+
+    self.prep_dlfcn_main()
+    self.clear_setting('NO_AUTOLOAD_DYLIBS')
+    create_file('main.cpp', r'''
+      #include <stdio.h>
+      #include <emscripten/val.h>
+      using namespace emscripten;
+      void liba_fun();
+      int main() {
+        liba_fun();
+        printf("done\n");
+        return 0;
+      }
+    ''')
+    self.do_runf('main.cpp', 'done\n', emcc_args=['--bind'])
+
   @no_wasm2js('TODO: source maps in wasm2js')
   def test_dwarf(self):
     self.emcc_args.append('-g')


msorvig:
This looks to be sufficient for Qt's use case. More types can be exported if needed.

There's a bit more to it: ideally embind should work with non-exported ("hidden")  types as well, and it looks like C++ API like  std::type_info::hash_code() and std::type_index may not work correctly for non-exported types.

Looking at at system/lib/libcxx/include/typeinfo, there are two options from implementing typeinfo:

1. _LIBCPP_TYPEINFO_COMPARISON_IMPLEMENTATION = 1: (default): "This implementation of type_info assumes a unique copy of the RTTI for a given type inside a program"
2. _LIBCPP_TYPEINFO_COMPARISON_IMPLEMENTATION = 2: "This implementation of type_info does not assume there is always a unique  copy of the RTTI for a given type inside a program"

Emscripten currently uses the default option 1, but does not generate unique RTTI for non-exported types in libraries/side modules.

brendandahl:
@sbc100 Do you think there's some way we could properly have unique rtti for the program?

msorvig:
Note previous discussion on https://github.com/emscripten-core/emscripten/issues/16711

test/test_other.py
```diff
@@ -3346,6 +3357,54 @@ def test_embind_jsgen_method_pointer_stability(self):
...
+    if emcc_args:
+      cmd += emcc_args
+    self.run_process(cmd)
+
+  def prep_dylink_main(self):
+    self.clear_setting('SIDE_MODULE')
+    self.set_setting('MAIN_MODULE', 1)
+
+  @requires_dylink
+  def test_embind_dylink_visibility_hidden(self):
```
sbc100, line 3373: This looks similar to what `test_dylink_rtti` is doing in `test_core.py`, but with some embind specifics.  As such, perhaps it belongs alongside that test?

Rather than duplicating the three helper functions from test_core.py why not just put this test in test_core.py itself?

sbc100: It also might make more sense alongside `test_embind_no_rtti`

sbc100:
> @sbc100 Do you think there's some way we could properly have unique rtti for the program?

I think the way to achieve that is make sure the types are marked as `default`and not `hidden` visibility.

This fix seems correct to me for the core embind types.  They always want to be shared I think.

We also have `-DEMSCRIPTEN_HAS_UNBOUND_TYPE_NAMES=0` which looks like it tells embind not to use RTTI info at all... although I don't really understand the naming of that macro.

test/test_other.py
msorvig: Yes, tst_core could make sense as well (I had it there first). Is it primarily a dylink or embind test? test_embind_no_rtti sounds related, I'll check what that does.

(Note I have vacation time is coming up, so I'll get back to this in a couple of week's time.)

brendandahl:
@msorvig would you mind updating this branch to re-run tests.

brendandahl:
@laminowany Do you have a small reproducer we could test with?

laminowany:
@brendandahl  unfortunately I don't really have a "small" case. Basically first I'm building the Qt framework from dev branch, then I'm building this https://github.com/mitchcurtis/slate using previously built Qt.
Are there some intermediate build files that could give you some insights? Or do you need some specific information that I could extract?

brendandahl:
> @brendandahl unfortunately I don't really have a "small" case. Basically first I'm building the Qt framework from dev branch, then I'm building this https://github.com/mitchcurtis/slate using previously built Qt. Are there some intermediate build files that could give you some insights? Or do you need some specific information that I could extract?

The output from `cmake --verbose` could be helpful or relalvent emcc flags you're using.

laminowany:
Here are flags I'm using:
`em++ -g -s MODULARIZE=1 -s EXPORT_NAME=app_entry -s EXPORTED_RUNTIME_METHODS=UTF16ToString,stringToUTF16,JSEvents,specialHTMLTargets,FS,callMain -s INITIAL_MEMORY=50MB -s MAXIMUM_MEMORY=4GB -s MAX_WEBGL_VERSION=2 -s WASM_BIGINT=1 -s STACK_SIZE=5MB -s ALLOW_MEMORY_GROWTH  -sMAIN_MODULE=1 -sFETCH  -lembind`

laminowany:
This line seems to cause an issue:
`   emscripten::val rawPlatform = emscripten::val::global("navigator")["platform"];`

laminowany:
Seems like `emscripten::val` does not have a stable `rawType` id across libraries? It works when applying patch from https://github.com/emscripten-core/emscripten/pull/18418

laminowany:
I'm new to Emscripten project so I dont know a lot about codebase, feel free to correct me!

So basically seems to me that types exported by embind: https://github.com/emscripten-core/emscripten/blob/1ddd29400e4aadc6f87314e76afd413805173ad4/system/lib/embind/bind.cpp#L122-L136

have different typeIDs, if they come from different object files (i.e. side module vs main module).

This is understandable given then typeID is generated using RTTI: 
https://github.com/emscripten-core/emscripten/blob/1ddd29400e4aadc6f87314e76afd413805173ad4/system/include/emscripten/wire.h#L64-L69

(to be fair I'm not sure what is the lifetime of typeid object in Emscripten and if is safe to take its address)

There is also other mechanism to generate typeIDs without RTTI which can be triggered by setting `EMSCRIPTEN_HAS_UNBOUND_TYPE_NAMES=0`, but it seems to be bugged on the program I tested.

Could you point me @brendandahl into some direction? 
I see two potential approaches:
1. Come up with new way to generate typeID which would be stable? Maybe `typeid(T).name()` is a better candidate since type names remain the same across object files? Then, it can be stored by name (or hash of it) in a dictionary like `unordered_map`.
2. Having a second mechanism of checking if type is registered in `embind_shared,js`.  Currently its done here:
https://github.com/emscripten-core/emscripten/blob/1ddd29400e4aadc6f87314e76afd413805173ad4/src/lib/libembind_shared.js#L157-L163
where its being checked in `registeredTypes` by raw type. Maybe having another register like `registeredNames`  and then checking as a fallback by name could be a fix? This is more or less what https://github.com/emscripten-core/emscripten/pull/18418 is doing. Or is it too much of workaround?

brendandahl:
Option 1 sounds interesting. I'm not that familiar with `typeid(T).name()` inner workings, but it appears to be what we'd want.  If you could get it to work with an integer hash of the name, then all the JS may "just work" as is.

sbc100:
IIRC typeid in general should be stable across DLLs.  I'm pretty sure we do have tests for it.

The problem only occurs when you do `-fvisibility=hidden` then types are not shared at all between DLLs.  I wonder if we can add some explict visibility decorators to the types in question to force them not to be hidden?

laminowany:
After further investigation it seems like this patch is actually solving my problem! :smile: 
I was building locally different version of Emscripten in the past, and when I cherry-picked this change I did not have clean build. Running `emcc --clear-cache` did the trick for me and now it seems like everything is working :tada: 
Sorry for the confusion, that was my mistake.

sbc100:
@msorvig are you still interested in getting this landed?  I think it just requires a test update and it should be good to go.

msorvig:
Hello, sorry for being a bit unresponsive - Yes, let's see if we can get this merged.

I'm not 100% sure if this fixes completely the issue, but it should help. Emscripten should implement RTTI correctly also for unexported / -fvisibility=hidden types, and then embind can use type_info::hash_code() for its ID.

type_info::name() was suggested above and is also an option, but it looks like the reference documentation at https://en.cppreference.com/w/cpp/types/type_info states that the name is not guaranteed to be unique or stable.

laminowany:
What is the status of it? Could we get it merged @sbc100 ?

system/include/emscripten/val.h
```diff
@@ -24,6 +24,7 @@
 #include <variant>
 #endif
 
+#define EMSCRIPTEN_EXPORT __attribute__((visibility("default")))
```
brendandahl, line 27: val.h includes wire.h so this shouldn't be needed.

system/include/emscripten/wire.h
```diff
@@ -24,6 +24,7 @@
 #include <string>
 
 #define EMSCRIPTEN_ALWAYS_INLINE __attribute__((always_inline))
+#define EMSCRIPTEN_EXPORT __attribute__((visibility("default")))
```
brendandahl, line 27: I'd prefer not to use `EMSCRIPTEN_EXPORT` since we may use that in the future for wasm exports. How about just `EMSCRIPTEN_VISIBILITY_DEFAULT`

brendandahl (changes requested):

system/include/emscripten/wire.h
sbc100: Also, how about `EMBIND_` as the prefix here?   (We can change EMSCRIPTEN_ALWAYS_INLINE too)

msorvig: Changed to EMBIND_VISIBILITY_DEFAULT, did not touch EMSCRIPTEN_ALWAYS_INLINE (yet)

test/test_core.py
```diff
@@ -7722,6 +7722,41 @@ def test_source_map(self, args):
     else:
       self.assertTrue(seen_lines.issuperset([6, 7, 11, 12]), seen_lines)
 
+  @no_asan("asan does not work with dynamic linking")
```
sbc100, line 7725: is that true?   I don't see any other dylink tests marked like this.

test/test_core.py
```diff
@@ -7722,6 +7722,41 @@ def test_source_map(self, args):
...
+    ''')
+
+    # side module
+    self.build_dlfcn_lib('liba.cpp', outfile='liba.so', emcc_args=['-std=c++11', '-fvisibility=hidden'])
+
+    # main module
+    self.set_setting('MAIN_MODULE', 1)
+    self.clear_setting('SIDE_MODULE')
+
+    self.emcc_args = ['liba.so', '-L.', '--bind']
```
sbc100, line 7746: Can you move this to `emcc_args=` in the call to do_runf?

test/test_core.py
```diff
@@ -7722,6 +7722,41 @@ def test_source_map(self, args):
...
+      #define EXPORT __attribute__((visibility("default")))
+      using namespace emscripten;
+      EXPORT void liba_fun() {
+        unsigned char buffer[1];
+        val view(typed_memory_view(1, buffer));
+      }
+    ''')
+
+    # side module
+    self.build_dlfcn_lib('liba.cpp', outfile='liba.so', emcc_args=['-std=c++11', '-fvisibility=hidden'])
```
sbc100, line 7740: Most other tests that call this function use `self.prep_dlfcn_main()` after.  Is there some reason not to do that here?

test/test_core.py
msorvig: The error message may be inaccurate here. I'll see if I can reproduce the test failure again.

test/test_core.py
msorvig: Yep, done,

test/test_core.py
msorvig: It sets NO_AUTOLOAD_DYLIBS, which causes an assertion "Assertion failed: undefined symbol '_ZTIN10emscripten11memory_viewIhEE'. "

Other than that it's fine, so we can for instance use it and then clear NO_AUTOLOAD_DYLIBS.

sbc100 (approved):

test/test_core.py
msorvig: Looks like @no_asan is no longer required.

sbc100:
@brendandahl is this good with you now?

brendandahl (approved):

laminowany:
Thanks! :heart:

PR #22097, opened 2024-06-14
Add a benchmarker that stores sizes in a skia perf compatible file.

brendandahl:
This will be run by emscripten releases to start tracking size benchmarks on our skia perf instance. I've intentially only added two benchmarks to start for testing.

The output format was verified using perf-tool. More info on the format at https://skia.googlesource.com/buildbot/+/refs/heads/main/perf/FORMAT.md

### DIFF
--- a/test/test_benchmark.py
+++ b/test/test_benchmark.py
@@ -57,7 +57,7 @@ EMTEST_BENCHMARKERS = os.getenv('EMTEST_BENCHMARKERS', 'clang,v8,v8-lto,v8-ctors
 
 
 class Benchmarker():
-  # whether to record statistics. set by SizeBenchmarker.
+  # Whether to record statistics. Set by SizeBenchmarker.
   record_stats = False
 
   # called when we init the object, which is during startup, even if we are
@@ -269,12 +269,16 @@ class EmscriptenBenchmarker(Benchmarker):
     return ret
 
 
+# This benchmarker will make a test benchmark build with Emscripten and record
+# the file output sizes in out/test/stats.json. The file format is specified at
+# https://skia.googlesource.com/buildbot/+/refs/heads/main/perf/FORMAT.md
+# Running the benchmark will be skipped.
 class SizeBenchmarker(EmscriptenBenchmarker):
   record_stats = True
 
   def __init__(self, name):
     # do not set an engine, as we will not run the code
-    super().__init__(name, engine=None, extra_args=[])
+    super().__init__(name, engine=None)
 
   def run(self, args):
     return
--- a/test/test_benchmark.py
+++ b/test/test_benchmark.py
@@ -454,8 +454,8 @@ class benchmark(common.RunnerCore):
     for b in benchmarkers:
       if skip_native and isinstance(b, NativeBenchmarker):
         continue
-      if isinstance(b, SizeBenchmarker):
-        # The size benchmarker doesn't need to do repetitions.
+      if not b.run:
+        # If we won't run the benchmark, we don't need repetitions.
         reps = 0
       baseline = b
       print('Running benchmarker: %s: %s' % (b.__class__.__name__, b.name))
--- a/test/test_benchmark.py
+++ b/test/test_benchmark.py
@@ -10,6 +10,7 @@ import shutil
 import sys
 import time
 import unittest
+import json
 import zlib
 from pathlib import Path
 from typing import List
@@ -56,6 +57,7 @@ EMTEST_BENCHMARKERS = os.getenv('EMTEST_BENCHMARKERS', 'clang,v8,v8-lto,v8-ctors
 
 
 class Benchmarker():
+  record_stats = False
   # called when we init the object, which is during startup, even if we are
   # not running benchmarks
   def __init__(self, name):
@@ -112,13 +114,38 @@ class Benchmarker():
 
     # size
 
-    size = sum(os.path.getsize(f) for f in self.get_output_files())
-    gzip_size = sum(len(zlib.compress(read_binary(f))) for f in self.get_output_files())
+    recorded_stats = []
 
-    print('        size: %8s, compressed: %8s' % (size, gzip_size), end=' ')
+    def add_stat(name, size, gzip_size):
+      recorded_stats.append({
+        'value': name,
+        'measurement': size,
+      })
+      recorded_stats.append({
+        'value': name + ' (gzipped)',
+        'measurement': gzip_size,
+      })
+
+    total_size = 0
+    total_gzip_size = 0
+
+    for file in self.get_output_files():
+      size = os.path.getsize(file)
+      gzip_size = len(zlib.compress(read_binary(file)))
+      if self.record_stats:
+        add_stat(utils.removeprefix(os.path.basename(file), 'size_'), size, gzip_size)
+      total_size += size
+      total_gzip_size += gzip_size
+
+    if self.record_stats:
+      add_stat('total', total_size, total_gzip_size)
+
+    print('        size: %8s, compressed: %8s' % (total_size, total_gzip_size), end=' ')
     if self.get_size_text():
       print('  (' + self.get_size_text() + ')', end=' ')
     print()
+
+    return recorded_stats
 
   def get_size_text(self):
     return ''
@@ -240,6 +267,16 @@ class EmscriptenBenchmarker(Benchmarker):
     return ret
 
 
+class SizeBenchmarker(EmscriptenBenchmarker):
+  record_stats = True
+
+  def __init__(self, name):
+    super().__init__(name, None, [])
+
+  def run(self, args):
+    return
+
+
 CHEERP_BIN = '/opt/cheerp/bin/'
 
 
@@ -317,6 +354,7 @@ aot_v8 = (config.V8_ENGINE if config.V8_ENGINE else []) + ['--no-liftoff']
 named_benchmarkers = {
   'clang': NativeBenchmarker('clang', [CLANG_CC], [CLANG_CXX]),
   'gcc': NativeBenchmarker('gcc',   ['gcc', '-no-pie'],  ['g++', '-no-pie']),
+  'size': SizeBenchmarker('size'),
   'v8': EmscriptenBenchmarker('v8', aot_v8),
   'v8-lto': EmscriptenBenchmarker('v8-lto', aot_v8, ['-flto']),
   'v8-ctors': EmscriptenBenchmarker('v8-ctors', aot_v8, ['-sEVAL_CTORS']),
@@ -337,6 +375,7 @@ for name in EMTEST_BENCHMARKERS.split(','):
 
 class benchmark(common.RunnerCore):
   save_dir = True
+  stats = [] # type: ignore
 
   @classmethod
   def setUpClass(cls):
@@ -357,6 +396,17 @@ class benchmark(common.RunnerCore):
       pass
     fingerprint.append('llvm: ' + config.LLVM_ROOT)
     print('Running Emscripten benchmarks... [ %s ]' % ' | '.join(fingerprint))
+
+  @classmethod
+  def tearDownClass(cls):
+    super().tearDownClass()
+    if cls.stats:
+      output = {
+        'version': 1,
+        'git_hash': '',
+        'results': cls.stats
+      }
+      utils.write_file('stats.json', json.dumps(output, indent=2) + '\n')
 
   # avoid depending on argument reception from the commandline
   def hardcode_arguments(self, code):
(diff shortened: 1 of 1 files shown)


brendandahl:
Better to view without whitespace changes https://github.com/emscripten-core/emscripten/pull/22097/files?w=1

test/test_benchmark.py
```diff
@@ -921,6 +955,36 @@ def output_parser(output):
       return float(re.search(r'Total elapsed: ([\d\.]+)', output).group(1))
     self.do_benchmark('matrix_multiply', read_file(test_file('matrix_multiply.cpp')), 'Total elapsed:', output_parser=output_parser, shared_args=['-I' + test_file('benchmark')])
 
+  @non_core
+  def test_sizes(self):
+    assert common.EMTEST_BENCHMARK_SIZE_FILE, 'Filename for measurements should be set.'
```
kripken, line 960: Rather than an env var, how about always writing this to the same place under `out/test/` (or maybe `out/bench/`)?

Then the user would run `./test/runner.py benchmark.test_sizes` and then copy from that directory.

That seems a little more consistent with how we do other things, and avoids a new env var, which seems useful. But I don't feel strongly.

brendandahl: I'm also currently using `EMTEST_BENCHMARK_SIZE_FILE` to only run the `EmscriptenBenchmarker`. I suppose I could modify the benchmarks to only include that one benchmarker within the test. Does that seem fine or did you have something else in mind?

kripken: Oh, I was thinking `./test/runner.py benchmark.test_sizes` was enough to just run that one benchmark, but I guess we also need to only run it in the one benchmarker. I don't remember if benchmarks are given the benchmarker, but that should be available somehow, and then we could do something like
```python
  if not self.benchmarker.is_emscripten():
    self.skipTest('benchmarking only runs on emscripten')
```

test/test_benchmark.py
```diff
@@ -305,52 +333,58 @@ def get_output_files(self):
...
-      # CheerpBenchmarker('cheerp-v8-wasm', aot_v8),
-    ]
-
-if config.SPIDERMONKEY_ENGINE and config.SPIDERMONKEY_ENGINE in config.JS_ENGINES:
-  # TODO: ensure no baseline compiler is used, see v8
+if common.EMTEST_BENCHMARK_SIZE_FILE:
+  TEST_REPS = 0
   benchmarkers += [
-    # EmscriptenBenchmarker('sm', SPIDERMONKEY_ENGINE),
+    EmscriptenBenchmarker('size', None, []),
```
sbc100, line 339: I'm not sure we want to keep adding environment variables as a way to configure which benchmakers we run.    I had the same objection when `EMTEST_FORCE64` was added I think.
  
  We should probably come up with some kind of command line scheme for configuring the benchmarkers.   I think most ppl just tend to edit this file as they need before the run the tests but that doesn't seem great either.

sbc100: Perhaps we could give each of these benchmarkers a unique string name and then have something like `EMTEST_BENCHMARKERS=...`  ?

test/test_benchmark.py
brendandahl: I changed it around a bit per your recommendation and Sam's. I write the file to the same spot always and now there's a new environment variable to explicitly say which benchmarkers you want to run.

test/common.py
```diff
@@ -63,7 +63,9 @@
 EMTEST_LACKS_NATIVE_CLANG = None
 EMTEST_VERBOSE = None
 EMTEST_REBASELINE = None
-EMTEST_FORCE64 = None
+# A comma separated list of benchmarkers to run during test_benchmark tests. See
+# `named_benchmarkers` in `test_benchmark.py` for what is available.
+EMTEST_BENCHMARKERS = 'clang,v8,v8-lto,v8-ctors'
```
kripken, line 68: Could this only be in `test_benchmark.py`?

test/runner.py
```diff
@@ -406,7 +406,7 @@ def configure():
   common.EMTEST_LACKS_NATIVE_CLANG = int(os.getenv('EMTEST_LACKS_NATIVE_CLANG', '0'))
   common.EMTEST_REBASELINE = int(os.getenv('EMTEST_REBASELINE', '0'))
   common.EMTEST_VERBOSE = int(os.getenv('EMTEST_VERBOSE', '0')) or shared.DEBUG
-  common.EMTEST_FORCE64 = int(os.getenv('EMTEST_FORCE64', '0'))
+  common.EMTEST_BENCHMARKERS = os.getenv('EMTEST_BENCHMARKERS')
```
kripken, line 409: This too, I hope?

test/test_benchmark.py
```diff
@@ -921,6 +928,36 @@ def output_parser(output):
       return float(re.search(r'Total elapsed: ([\d\.]+)', output).group(1))
     self.do_benchmark('matrix_multiply', read_file(test_file('matrix_multiply.cpp')), 'Total elapsed:', output_parser=output_parser, shared_args=['-I' + test_file('benchmark')])
 
+  @non_core
+  def test_sizes(self):
+    if len(benchmarkers) != 1 or not isinstance(benchmarkers[0], EmscriptenBenchmarker):
+      self.skipTest('size benchmark only runs on emscripten')
+    benchmarkers[0].skip_bench = True
```
kripken, line 935: Can we maybe replace the new `skip_bench` mechanism with a `skipTest()` that is called at the **end** of this function, after the work was done?

site/source/docs/getting_started/test-suite.rst
```diff
@@ -181,11 +181,19 @@ As with all the test suites, you can also run a specific benchmark:
...
+benchmarkers (names can be found in `named_benchmarkers` in
+`test/test_benchmark.py`).
+
+.. code-block:: bash
+
+  # Run one specific benchmark and with clang and v8.
+  EMTEST_BENCHMARKERS=clang,v8 test/runner benchmark.test_skinning
+
+To further customize how the benchmarks are run, you will want to edit the file
+`test/test_benchmark.py`. Some of the options include:
```
sbc100, line 195: Sorry to be a pain but can we split this change into its own PR?

Could we also remove `EMTEST_FORCE64` at the same time?  (since I believe this feature superseded that).

brendandahl: Split out into https://github.com/emscripten-core/emscripten/pull/22139

test/test_benchmark.py
brendandahl: I may be misunderstanding, but I think in that case `b.bench` in `do_benchmark` will still try to run and do multiple rounds of the benchmark.

test/test_benchmark.py
```diff
@@ -112,10 +113,33 @@ def display(self, baseline=None):
...
+        'measurement': size,
+      })
+      self.size_measurements.append({
+        'value': name + ' (gzipped)',
+        'measurement': gzip_size,
+      })
+
+    total_size = 0
+    total_gzip_size = 0
+    save_measurements = hasattr(self, 'size_measurements') and self.size_measurements is not None
```
sbc100, line 128: You can just use `getattr(self, 'size_measurements', None)` here I think

test/test_benchmark.py
```diff
@@ -112,10 +113,33 @@ def display(self, baseline=None):
...
+
+    for file in self.get_output_files():
+      size = os.path.getsize(file)
+      gzip_size = len(zlib.compress(read_binary(file)))
+      if save_measurements:
+        # Remove the 'size_' prefix added by the benchmark runner.
```
sbc100, line 134: I don't understand... why does every file start with `size_`?

test/test_benchmark.py
```diff
@@ -112,10 +113,33 @@ def display(self, baseline=None):
...
+        add_measurement(os.path.basename(file)[5:], size, gzip_size)
```
sbc100, line 135: You can use `utils.removeprefix(..)` here to be more explictl

test/test_benchmark.py
```diff
@@ -903,6 +931,36 @@ def output_parser(output):
...
+          'stats': []
+        }
+      }
+      # The benchmark will add to this attribute.
+      benchmarkers[0].size_measurements = result['measurements']['stats']
+      testFn()
+      output['results'].append(result)
+
+    run_sub_test('primes', self.test_primes)
+    run_sub_test('primes_nocheck', self.test_primes_nocheck)
```
sbc100, line 961: This running of a test method from another test method seems odd to me.

Is there not perhaps some way we could use a decorator which we add to `primes` and `primes_nocheck`?   Something like `@track_size`?  Then the track_size decorator could take care of doing all of this?   

Then when the suite is run in "track sizes" mode we could skip any tests that don't have the decorator?

brendandahl: From emscripten-releases I want to run one command that will run all the sizes tests so we don't have to update emscripten-releases each time we add another size benchmark. Do you have any thoughts on how to achieve that with a decorator? I suppose I could add a parameterized version of each test, so then I could run `test/runner benchmark.test_*_sizes` or something.

test/test_benchmark.py
kripken: Hmm, I think you're right. This does need to work with the internals there. So I understand why this is needed to prevent `test_sizes` from being handled normally. But how do we prevent the sub-functions it calls (`test_primes` etc.) from also being treated in that way? That's what I find confusing I guess.

Separately, I think it would be nice if we could more clearly separate this function, which is a "meta-benchmark function" in a sense, from the normal benchmark functions. Perhaps instead of `@non_core` we could have a decorator `@meta` (or some better name) which implies both non-core (not run by default) and also the decorator sets `skip_bench`? The decorator would be a clearer indication of the special-ness of this function, rather than a line inside the function that sets `skip_bench`, is all I am thinking here.

test/test_benchmark.py
sbc100: I was thinking that you would just run the entire branchmark suite.

In sizes mode we would just skip any benchmarks that are not annotated.. and then adding new tests would be a simple as adding the annotation.

test/test_benchmark.py
brendandahl: When each benchmarker builds it prefixes all the output files with the name of the benchmarker.

test/test_benchmark.py
brendandahl: I think I've come up with something better. Now you can use the 'size' benchmarker and it will record stats for any test you run and save them on tear down. This will require us to specify which benchmarks we're recording in emscripten-releases, but I think that's fine since it gets rid of the weird meta test here and I don't imagine the list would change much.

test/test_benchmark.py
```diff
@@ -60,6 +61,7 @@ class Benchmarker():
   # not running benchmarks
   def __init__(self, name):
     self.name = name
+    self.record_stats = False
```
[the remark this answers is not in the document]
brendandahl: You're right. I previously didn't have a separate class for SizeBenchmarker so it had to be an instance property.

test/test_benchmark.py
```diff
@@ -56,6 +57,7 @@
 
 
 class Benchmarker():
+  record_stats = False
```
kripken, line 60:
```suggestion
  # whether to record statistics. set by SizeBenchmarker.
  record_stats = False

```

test/test_benchmark.py
```diff
@@ -240,6 +267,16 @@ def get_output_files(self):
     return ret
 
 
+class SizeBenchmarker(EmscriptenBenchmarker):
+  record_stats = True
+
+  def __init__(self, name):
+    super().__init__(name, None, [])
+
```
kripken, line 275:
```suggestion
  def __init__(self, name):
    # do not set an engine, as we will not run the code
    super().__init__(name, engine=None, extra_args=[])

```
(why must `extra_args` be set?)

test/test_benchmark.py
```diff
@@ -240,6 +267,16 @@ def get_output_files(self):
     return ret
 
 
+class SizeBenchmarker(EmscriptenBenchmarker):
```
kripken, line 270: Maybe add a comment here about what this does, what the format it saves it in is, etc.? (or is that documented somewhere else?)

test/test_benchmark.py
```diff
@@ -240,6 +269,21 @@ def get_output_files(self):
...
+# Running the benchmark will be skipped.
+class SizeBenchmarker(EmscriptenBenchmarker):
+  record_stats = True
+
+  def __init__(self, name):
+    # do not set an engine, as we will not run the code
+    super().__init__(name, engine=None)
+
+  def run(self, args):
+    return
```
kripken, line 284: Can this
```suggestion
  # we will not actually run the benchmarks
  run = None
```
work together with the next comment?

test/test_benchmark.py
```diff
@@ -397,11 +454,27 @@ def do_benchmark(self, name, src, expected_output='FAIL', args=None,
     for b in benchmarkers:
       if skip_native and isinstance(b, NativeBenchmarker):
         continue
+      if isinstance(b, SizeBenchmarker):
+        # The size benchmarker doesn't need to do repetitions.
+        reps = 0
```
kripken, line 459:
```suggestion
      if not b.run:
        # If we won't run the benchmark, we don't need repetitions.
        reps = 0
```

kripken (commented):
lgtm % this pair of suggestions. If they work they avoid hardcoding the class name in unrelated code, which I think is nice. But lgtm either way.

kripken (approved):

PR #22099, opened 2024-06-16
[contrib.glfw3] new version

ypujante:
New version of [underlying project](https://github.com/pongasoft/emscripten-glfw/releases/tag/v3.4.0.20240616) 

Release notes:

-   Implemented `glfwGetClipboardString`. Note that due to the async (and restrictive) nature of the
`navigator.clipboard.readText` call, this synchronous API returns whatever was set via a previous call
to `glfwSetClipboardString` and ignores the external clipboard entirely.

### DIFF
--- a/tools/ports/contrib/glfw3.py
+++ b/tools/ports/contrib/glfw3.py
@@ -6,8 +6,8 @@
 import os
 from typing import Dict
 
-TAG = '3.4.0.20240601'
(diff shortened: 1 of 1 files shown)


ypujante:
Somebody reported an issue (unrelated to this PR) and I will do only 1 PR to bring both changes once the issue is fixed.

PR #22106, opened 2024-06-18
Correct some wrong English grammar

shlomif:
Not tested (given my computer at home has only 8 GB of RAM)

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1249,7 +1249,7 @@ See docs/process.md for more on how version tagging works.
   their own secondary sysroot may be able to simplify their build system by
   removing this completely and relying on the new default.
 - Reinstated the warning on linker-only `-s` settings passed when not linking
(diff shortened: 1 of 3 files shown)


shlomif:
Just for the record: I detected another issue of using "Where where" instead of "Where" in two of the same files

sbc100:
> Just for the record: I detected another issue of using "Where where" instead of "Where" in two of the same files

Do you want to add that fix to this PR?  (I.e. should I wait on landing this?)

sbc100 (approved):

shlomif:
> > Just for the record: I detected another issue of using "Where where" instead of "Where" in two of the same files
> 
> Do you want to add that fix to this PR? (I.e. should I wait on landing this?)

I'd rather the change will be merged as-is, and I will prepare a new pull-request (to make it a single-purpose commit).

PR #22109, opened 2024-06-18
Cleanup fetch tests. NFC

sbc100:


test/fetch/test_fetch_waitable.c
```diff
@@ -8,27 +8,31 @@
...
-  while(ret == EMSCRIPTEN_RESULT_TIMED_OUT)
-  {
-    /* possibly do some other work; */
-    ret = emscripten_fetch_wait(fetch, 0/*milliseconds to wait. 0 to just poll, INFINITY=wait until completion*/);
+  while (ret == EMSCRIPTEN_RESULT_TIMED_OUT) {
+    // possibly do some other work
+    // milliseconds to wait. 0 to just poll, INFINITY=wait until completion
+    ret = emscripten_fetch_wait(fetch, 1);
+  }
+  if (ret != EMSCRIPTEN_RESULT_SUCCESS) {
```
kripken, line 24: This looks like an additional check, not a cleanup, or am I misreading it?

sbc100: Yeah, I was testing to see if `EMSCRIPTEN_FETCH_WAITABLE` worked at all using this example.  Turns out it doesn't work and hasn't worked since we removed the fetch worker.  But this test was lacking this check and was silently succeeding

sbc100: Or at least it was reporting `status` == 0 rather than the actual error code here.

test/fetch/test_fetch_stream_async.c
```diff
@@ -35,15 +33,14 @@ void downloadProgress(emscripten_fetch_t *fetch)
...
+int main() {
   emscripten_fetch_attr_t attr;
   emscripten_fetch_attr_init(&attr);
   strcpy(attr.requestMethod, "GET");
   attr.attributes = EMSCRIPTEN_FETCH_STREAM_DATA;
   attr.onsuccess = downloadSucceeded;
   attr.onprogress = downloadProgress;
   attr.onerror = downloadFailed;
-  attr.timeoutMSecs = 2*60;
+  attr.timeoutMSecs = 2*60*1000;
```
kripken, line 44: Strains the definition of NFC but still true, I guess :smile:

kripken (approved):

PR #22122, opened 2024-06-20
Update npm dependencies. NFC

sbc100:
I used npm-check-update to generate this PR:

1. npm install --global npm-check-updates
2. ncu -u
4. npm install

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,23 +5,23 @@
   "packages": {
     "": {
       "dependencies": {
-        "@babel/cli": "^7.23.4",
-        "@babel/core": "^7.23.7",
-        "@babel/preset-env": "^7.23.8",
-        "acorn": "^8.11.3",
+        "@babel/cli": "^7.24.7",
+        "@babel/core": "^7.24.7",
+        "@babel/preset-env": "^7.24.7",
+        "acorn": "^8.12.0",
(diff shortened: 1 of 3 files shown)


sbc100:
I held back closure-compiler and eslint which both seemed to have some issues on upgrade.

kripken (approved):

sbc100:
@brendandahl could you take a look at the test failure here?   It looks like one of the generated typescript signatures changed:

```
--- /root/project/test/other/test_emit_tsd.d.ts
+++ /root/project/test/other/test_emit_tsd.d.ts.new
@@ -9,7 +9,7 @@
      * @param {number=} maxBytesToRead
      * @return {string}
      */
-    function UTF8ArrayToString(heapOrArray: any, idx: number, maxBytesToRead?: number): string;
+    function UTF8ArrayToString(heapOrArray: any, idx: number, maxBytesToRead?: number | undefined): string;
     let wasmTable: WebAssembly.Table;
     let HEAPF32: any;
     let HEAPF64: any;
```

I can't tell if that is for the better or worse though?  Isn't `| undefined` redundant in the face of `?`

brendandahl:
I would think those would be the same. I can do a bisect and see when it changed, but I don't think you need to block this PR on it.

PR #22124, opened 2024-06-21
NFC Factor out common code between updateGOT and reportUndefinedSymbols

hoodmane:
Suggested in discussion on #22053


hoodmane:
@sbc100 I think the failure is a flake.

kripken:
I re-ran tests here but the error recurred, though it did look like a flake. I tried to update the branch to fix that but now more tests fail, so there may be a larger issue here.

src/library_dylink.js
```diff
@@ -201,9 +201,30 @@ var LibraryDylink = {
 #endif
     ;
   },
+  $setGOTEntry__deps: ['$addFunction', '$getFunctionAddress'],
+  $setGOTEntry: (symName, value) => {
+    const entry = GOT[symName];
```
sbc100, line 206: It looks like both calls sites already have looked up the entry so perhaps we can pass the entry here too to avoid this extra lookup?

PR #22125, opened 2024-06-21
Fix single line comment preservation in acorn-optimizer

sbc100:
Terser has `comment1` and `comment2` types but there is no such thing as just `comment`.


sbc100:
Actually maybe we should just drop these comments, since its hard to put them in the correct place in the final output.

kripken (approved):
Seems worthwhile to me even if the comments might move slightly. The change is like a linter would do.

PR #21573, opened 2024-03-20
Fix for zero count in glUniform family of functions

sbc100:
The previous fix for this was in #16837, but I looks like that only covered the new "garbage-free" webgl2 path.  When old webgl1 path was still using the zero count value.

As part of this change I resurrected                                     
`test_webgl_draw_triangle_with_uniform_color.c` which has not actually   
be used since #20925.  

Fixes: #21567

PR #22130, opened 2024-06-22
Fix a crash in miniTempWebGL*Buffers optimization

kainino0x:
library_webgl.js attempts to use a pool of temp float or int buffers for uniform uploads up to 288 elements, however it would only allocate 288 temp buffers, so 0 through 287 would be fine but 288 would crash (miniTempWebGLFloatBuffers[288] would be undefined).

Confirmed manually that any one of the six "Just at the optimization limit" calls would crash before, and with the fix does not.

Fix for the change in PR #21573

### DIFF
--- a/src/library_webgl.js
+++ b/src/library_webgl.js
@@ -28,13 +28,14 @@ var LibraryGL = {
 
   $miniTempWebGLFloatBuffers: [],
   $miniTempWebGLFloatBuffers__postset: `var miniTempWebGLFloatBuffersStorage = new Float32Array({{{ GL_POOL_TEMP_BUFFERS_SIZE }}});
-for (/**@suppress{duplicate}*/var i = 0; i < {{{ GL_POOL_TEMP_BUFFERS_SIZE }}}; ++i) {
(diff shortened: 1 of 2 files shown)


src/library_webgl.js
```diff
@@ -28,13 +28,15 @@ var LibraryGL = {
 
   $miniTempWebGLFloatBuffers: [],
   $miniTempWebGLFloatBuffers__postset: `var miniTempWebGLFloatBuffersStorage = new Float32Array({{{ GL_POOL_TEMP_BUFFERS_SIZE }}});
-for (/**@suppress{duplicate}*/var i = 0; i < {{{ GL_POOL_TEMP_BUFFERS_SIZE }}}; ++i) {
+// Create GL_POOL_TEMP_BUFFERS_SIZE+1 temporary buffers, for uploads of size 0 through GL_POOL_TEMP_BUFFERS_SIZE inclusive
+for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}}; ++i) {
```
sbc100, line 32: I guess I was the one who broke this in #21573?

Could we instead update all the other places where `GL_POOL_TEMP_BUFFERS_SIZE` to use less than?  e.g. all the places there we currently do `if (count <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}})`

sbc100 (approved):

PR #22136, opened 2024-06-24
Improve comment preservation in acorn-optimizer. NFC

sbc100:
Prior to this change we would drop any comment that was more than 20 characters from JS symbol, based on the end of the commend and that start of the symbol.    This mean that all but the last single-line comments in a sequence would be dropped.

With this change we map all comments to the start of the next symbol. This means that we can have many comments in row that all map the same location and we decide to either keep all of them or discard them.

### DIFF
--- a/tools/acorn-optimizer.mjs
+++ b/tools/acorn-optimizer.mjs
@@ -1932,7 +1932,7 @@ function minifyGlobals(ast) {
 
 // Utilities
 
-function reattachComments(ast, comments) {
+function reattachComments(ast, commentsMap) {
   const symbols = [];
 
   // Collect all code symbols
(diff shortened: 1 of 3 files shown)


tools/acorn-optimizer.mjs
```diff
@@ -2029,16 +2033,22 @@ let extraInfo = null;
...
+// outputted code if --closureFriendly was requested.
+const sourceComments = {};
+const currentComments = [];
 let ast;
 try {
   ast = acorn.parse(input, {
     // Keep in sync with --language_in that we pass to closure in building.py
     ecmaVersion: 2021,
     preserveParens: closureFriendly,
-    onComment: closureFriendly ? sourceComments : undefined,
```
kripken, line 2041: Looks like before we avoided collecting info here, if we would not use it later. Is that not worth still doing, for efficiency?

tools/acorn-optimizer.mjs
```diff
@@ -1949,33 +1949,37 @@ function reattachComments(ast, comments) {
...
+    while (j < symbols.length && symbols[j].start.pos < pos) {
       ++j;
     }
     if (j >= symbols.length) {
-      trace(`dropping comment: no symbol comes after it (${comments[i].value.slice(0, 30)})`);
+      trace('dropping comments: no symbol comes after them');
       break;
     }
-    if (symbols[j].start.pos - comments[i].end > 20) {
+    if (symbols[j].start.pos - pos > 20) {
```
kripken, line 1961: This was `comments[i].end`, but now as the key in that map it was set as `token.start` on line 2048. Why is that ok?

tools/acorn-optimizer.mjs
sbc100: Oh yes I can revert that part.

tools/acorn-optimizer.mjs
sbc100: Yes this is kind of the point of this PR.   We associate each comment not with its own end position, but instead with the start position of the next token.

This means if you have 100 comments in a row they will be associated with the same `pos` here.

kripken: Thanks, now I see... I was thinking that part was NFC, which confused me :smile:

tools/acorn-optimizer.mjs
kripken: Same think in `onToken`, I think? We don't need to build up `sourceComments/currentComments` if not `closureFriendly`.

sbc100: Refactored.

tools/acorn-optimizer.mjs
```diff
@@ -2029,19 +2033,30 @@ let extraInfo = null;
...
+  ecmaVersion: 2021,
+  sourceType: exportES6 ? 'module' : 'script',
+  allowAwaitOutsideFunction: true,
+};
+if (closureFriendly) {
+  const currentComments = [];
+  Object.assign(params, {
+    preserveParens: true,
+    onToken: (token) => {
+      // Associate comments the start position of the next token.
```
kripken, line 2050:
```suggestion
      // Associate comments with the start position of the next token.
```

kripken (approved):

PR #22150, opened 2024-06-26
Cleanup parse_benchmark_output. NFC

sbc100:
View with ignore whitespace.


kripken (approved):
lgtm, but I believe this isn't tested, so please verify manually.

PR #22153, opened 2024-06-27
changed LLVM root path to lower case in sanity.txt output for Windows

xanzarx:
Visual Studio Code [CMakeTools](https://github.com/microsoft/vscode-cmake-tools) extension cleans cache when the editor opens in Windows. Possible root cause is different LLVM root paths string case in santiy.txt: 
- For startup it produces path starting with  'c:\' e.g  `c:\projects\emscripten\upstream\bin`
- For regular cmake configuration or build it produces path starting with 'C:\' e.g `C:\projects\emscripten\upstream\bin` 
This leads to rebuilding cache every time the editor is launched, which is a bit frustrating 
Link to similar [issue](https://github.com/microsoft/vscode-cpptools/issues/9770)

For windows path to folders is case independent anyways, so possible fix is just put lowercase string to sainty.txt for Windows.

### DIFF
--- a/tools/shared.py
+++ b/tools/shared.py
@@ -433,7 +433,10 @@ def set_version_globals():
 
 
 def generate_sanity():
-  return f'{EMSCRIPTEN_VERSION}|{config.LLVM_ROOT}\n'
+  llvm_root = config.LLVM_ROOT
+  if WINDOWS:
+    llvm_root = config.LLVM_ROOT.lower()
+  return f'{EMSCRIPTEN_VERSION}|{llvm_root}\n'
 
 
 @memoize


sbc100:
I'd like to understand this issue a little more before landing this.  How/where/why is VSCode settings the llvm root?

sbc100:
Is `EM_CONFIG` being set somewhere perhaps?

tools/shared.py
```diff
@@ -433,7 +433,10 @@ def set_version_globals():
 
 
 def generate_sanity():
-  return f'{EMSCRIPTEN_VERSION}|{config.LLVM_ROOT}\n'
+  llvm_root = config.LLVM_ROOT
```
sbc100, line 436: can you use do `llvm_root = os.path.normcase(config.LLVM_ROOT)` here?

xanzarx:
As it turns out running commands with drive letter in different case in windows terminal application will lead change sanity.txt (and cache rebuild) 

`C:\Projects\test>C:\projects\emscripten\upstream\emscripten\em++ main.cpp`
sanity.txt: 
```
3.1.59|C:/Projects/emscripten/upstream/bin
```

`C:\Projects\test>c:\projects\emscripten\upstream\emscripten\em++ main.cpp`
sanity.txt:
```
3.1.59|c:/Projects/emscripten/upstream/bin
```

For not it seems the issue was not caused (at least directly) by interaction with VS Code or CMake tools plugin  for VSCode

tools/shared.py
xanzarx: Yes, thanks for pointing out

PR #22157, opened 2024-06-27
Use C/C++ bool type to implement EM_BOOL

sbc100:
This reduces the size of several structs and can result in code size savings in some cases.  The reason
the code size savings don't show up in trivial examples is (I believe) because this change also increases
the use of HEAP8 (where previously some examples only depended on HEAP32).

This change is split of from a larger change I have planned to remove the use of EM_BOOL completely: #22155.


sbc100:
I think we should probably do a release containing this change before we land part 2 so that we can separate any possible fallout.

kripken (approved):

sbc100:
I had to update a bunch of places where we were accessing these fields (u8 vs u32 access).  PTAL.

sbc100:
I used the following commands to fine and fix all locations in the JS library code that was treating these values as 32 bit:

```
// Find all EM_BOOL struct members
$ git grep "^\s*EM_BOOL[^(]*$" system/include/ | awk '{ print $3 }' | sed "s/;//" | sort | uniq > bools.txt
// Find all places where these members are used in HEAP32/HEAPU32 accesses
$ for bool in `cat bools.txt`; do git grep "HEAPU\?32.*$bool"; done
// Find all the places they are used in `i32`/`u32` `makeGetValue`/`makeSetValue`
$ for bool in `cat bools.txt`; do git grep "make[GS]etValue.*\.$bool\>.*32"; done
```

src/library_websocket.js
```diff
@@ -294,7 +294,7 @@ var LibraryWebSocket = {
     var url = UTF8ToString({{{ makeGetValue('createAttributes', 0, '*') }}});
     var protocols = {{{ makeGetValue('createAttributes', C_STRUCTS.EmscriptenWebSocketCreateAttributes.protocols, '*') }}}
     // TODO: Add support for createOnMainThread==false; currently all WebSocket connections are created on the main thread.
-    // var createOnMainThread = HEAP32[createAttrs+2];
+    // var createOnMainThread = HEAP8[createAttributes+2];
```
kripken, line 297: This looks wrong. That's a pointer.

test/code_size/hello_webgl_wasm2js.json
```diff
@@ -1,8 +1,8 @@
 {
   "a.html": 346,
   "a.html.gz": 262,
-  "a.js": 21690,
-  "a.js.gz": 11405,
-  "total": 22036,
-  "total_gz": 11667
+  "a.js": 21720,
```
kripken, line 4: slightly odd size increases here. do you know why?

sbc100: I guess because the program previously only used HEAP32 but now also uses HEAP8?

src/library_websocket.js
sbc100: It looks right to me.   createAttributes is a pointer a struct.  Two bytes into that struct is a boolean called `createOnMainThread`: https://github.com/emscripten-core/emscripten/blob/b8c6e231ee05cbadef1abd9e41dde5cd19e5a733/system/include/emscripten/websocket.h#L96

kripken: Ah, I see, thanks.

kripken: (But it was wrong before, then: the HEAP32 was missing the `>> 2`. Anyhow, it's in a comment...)

Issue #21024, opened 2024-01-06
Behavioral Inconsistency of `strptime` between emscripten and glibc/musl


gudzpoz:
In short, `strptime("2012/2/30", "%Y/%m/%d", &date)` yields `2012/3/1` in emscripten while `gcc a.c` and `musl-gcc a.c` both give `2012/2/30`. (Possibly UB though.)

**Version of emscripten/emsdk:**
```console
$ emcc -v       
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.51-git (c0c2ca1314672a25699846b4663701bcb6f69cca)
clang version 18.0.0git (/startdir/llvm-project f2464ca317bfeeedddb7cbdea3c2c8ec487890bb)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /opt/emscripten-llvm/bin
```

**Failing command line in full:**
```c
// a.c
#include <stdio.h>
#include <time.h>

int main() {
  struct tm date;
  strptime("2012/2/30", "%Y/%m/%d", &date);
  printf("%d/%d/%d\n", date.tm_year + 1900, date.tm_mon + 1, date.tm_mday);
  return 0;
}
```
```console
$ emcc a.c
$ node a.out.js
2012/3/1
```
In comparison, `gcc a.c` and `musl-gcc a.c` both give `2012/2/30`.

**Spec:**
The [POSIX spec](https://pubs.opengroup.org/onlinepubs/9699919799/functions/strptime.html) does not seem to specify what to do when it encounters an invalid number that does not fit into the range, so this is possibly not a but but an undefined behavior instead? (But even if it is a UB, it really helps if emscripten follows the behavior of glibc/musl.)

PR #22158, opened 2024-06-28
Use musl's strptime implementation

sbc100:
This is a slight regression in code size compared the JS version, but makes
the wasm file more standalone and portable.   I measured the combined size of  
`core2.test_strptime_days` (with closure compiler enabled) and got the   
following results, which is a 2% regression:                                                    
                                                                         
- before: 16777 bytes                                                    
- after: 17235 bytes                                                     
                                                                         
As part of this I updated musl's implementation to support several          
features that our current tests depend on:                         
                                                                         
- Add support for '%z' for parsing timezone offsets.                     
- Add support for '%U' and '%W' (week of the year)                       
- Derive and set tm_yday/tm_mday/tm_yday fields where possible.          
                                                                         
Regarding '%U' and '%W' it seems this is missing from both musl and         
bionic:                                                                  
                                                                         
- https://github.com/aosp-mirror/platform_bionic/blob/4cfb2bde90fcf0a1f835e2b670b00e9f226c1c7c/libc/tzcode/strptime.c#L364-L375
- https://github.com/bminor/musl/blob/f314e133929b6379eccc632bef32eaebb66a7335/src/time/strptime.c#L126-L132
                                                                         
However, I implemented here (based on code from FreeBSD) in order to        
avoid any regressions (since we do have test coverage for these flags).

Fixes: #21024


kripken:
> This seems to save on code size compared the JS version [..] I measured the combined size of core2.test_strptime_days (with closure compiler enabled) and got the following results:
>
>    before: 16777 bytes
>    after: 17235 bytes

That looks like a regression, not a savings, or are the two sentences talking about different things?

sbc100:
> > This seems to save on code size compared the JS version [..] I measured the combined size of core2.test_strptime_days (with closure compiler enabled) and got the following results:
> > before: 16777 bytes
> > after: 17235 bytes
> 
> That looks like a regression, not a savings, or are the two sentences talking about different things?

Yeah sorry, it was a saving, but then once I added all that new code (the stuff that musl doesn't implement) it grew by a little and now its a slight regression., but its only 2%.

system/lib/libc/musl/src/time/strptime.c
```diff
@@ -1,3 +1,67 @@
...
+ * TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE FOUNDATION OR CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
```
kripken, line 64: These licenses look like 2-part BSD, which is different from musl and Emscripten's MIT license. We'd need to mention that in our main `LICENSE` file (which so far has only had to talk about MIT).

This isn't a big deal, I think, but given this also increases code size I am a little unsure of the benefit here. The code being more standalone is helpful, but maybe there are other ways to achieve that?

(Does WASI not have APIs to get the day of the week and stuff like that, like the Web does? If it does then perhaps we can farm out to Web/WASI APIs rather than bundle this code.)

sbc100: Another option is to remove all these additions, which would:

(1) + remove the license stuff
(2) + make this change a code size win
(3) - regress out support for certain formatting chars (to the same level as musl).

Regarding WASI APIs, no I don't think that would be needed since there is nothing system dependent going on here, we are just translating data from one format to another.

sbc100: Regarding the license, yes that is annoying I forgot we don't already have BSD stuff in out stdlib..

PR #22166, opened 2024-06-29
Perform cDefs/cStructs macro substitution using acorn pass

sbc100:
Rather than depending on the JS pre-processor we can achieve better results using and acorn pass.  Some upsides of doing it this way are:

1. Better syntax highlighing since JS library files are closer to normal JS.
2. Less noise in input files.

In debug builds include the cDefs/cStructs mappings in the generated code.  In release builds we use acorn to replace all occurrences of cDefs/cStructs.


sbc100:
This is just a draft for now.

sbc100:
To see what it looks like when we use this new technique in our JS library code 3ea9b6a15

tools/acorn-optimizer.mjs
```diff
@@ -1930,6 +1934,53 @@ function minifyGlobals(ast) {
...
+    : 'struct_info_generated.json';
+  const structInfoURL = new URL('../src/' + structInfoFile, import.meta.url);
+  // Load struct and define information.
+  const structInfo = JSON.parse(read(structInfoURL));
+  const structs = structInfo.structs;
+  const defines = structInfo.defines;
+
+  fullWalk(ast, (node) => {
+    if (node.type == 'MemberExpression') {
+      if (node.object.name == 'cDefs') {
```
kripken, line 1949: We'd probably want scoping analysis here too, to avoid issues in user code with the short name `cDefs` (which might occur as the result of bundling minified code, say).

kripken (commented):
Overall this seems like a potentially promising direction. However I agree with @dschuff 's past point that we should probably have a full plan for how to handle functions and everything else before starting down this path, so we don't end up stuck in the middle.

tools/acorn-optimizer.mjs
brendandahl: I was going to propose some common prefix or object so it's easy to see what will be replaced as we add more. e.g. `Preprocessor.cDefs`

sbc100: Good idea.  However I also want to keep these things short if possible.  I was thinking maybe `EM.EINVAL` ?  and putting everything directly on the `EM` object?

brendandahl: `EM` sounds good, but if they're going to be replaced, why keep them as short as possible?

sbc100: For readability of the source code.   These things are used a lot, and they are also often logically or'd together so repeated long prefixed just look noisy to me.. but maybe that just me?

Issue #22161, opened 2024-06-29
`_emscripten_memcpy_js` runtime crash with MAIN/SIDE_MODULE + WASM_BIGINT


Faless:
**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.61 (67fa4c16496b157a7fc3377afd69ee0445e8a6e3)
clang version 19.0.0git (https:/github.com/llvm/llvm-project 7cfffe74eeb68fbb3fb9706ac7071f8caeeb6520)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /media/Storage/emsdk/upstream/bin
```

**Failing command line in full:**

I've created a small test showing the issue (patch: [test.diff.txt](https://github.com/user-attachments/files/16039500/test.diff.txt) ):

```python
  @with_env_modify({'EMCC_FORCE_STDLIBS': '1'})
  def test_memops_bulk_memory_side(self):
    create_file('pre.js', '''
    Module["dynamicLibraries"] = ["libside.wasm"];
    ''')
    create_file('side.c', '''
#include <assert.h>
#include <string.h>
#include <emscripten/console.h>
#include <emscripten/emscripten.h>

void EMSCRIPTEN_KEEPALIVE side_main() {
  char buffer[1024];
  char out[1024];
  memset(buffer, 'a', 1024);
  buffer[sizeof(buffer) - 1] = '\\0';
  memcpy(out, buffer, sizeof(buffer));
  assert(strcmp(buffer, out) == 0);
  emscripten_console_log(buffer);
}
    ''')
    create_file('main.c', '''
#include <assert.h>
#include <string.h>

extern void side_main();

int main() {
  side_main();
  return 0;
}
    ''')
    common_args = [] # OK
    common_args += ['-sMIN_SAFARI_VERSION=150000'] # Breaks
    #common_args += ['-sWASM_BIGINT=1'] # Also breaks (because it forces MIN_SAFARI_VERSION to 150000)
    #common_args += ['-mbulk-memory'] # Fixes the build, and can also be only applied to the SIDE MODULE
    side_args = ['-sSIDE_MODULE=2'] + common_args
    main_args = ['-sMAIN_MODULE=1', '-sEXPORT_ALL=1', '-sWARN_ON_UNDEFINED_SYMBOLS=0', '--pre-js=pre.js'] + common_args
[...]

PR #22171, opened 2024-07-01
Preserve system library ordering even under EMCC_FORCE_STDLIBS=1

sbc100:
In particular `EMCC_FORCE_STDLIBS=1` would previously place libc before libbulkmemory which don't work.  libbulkmemory always needs to come first on the link line.

Fixes: #22161

### DIFF
--- a/tools/system_libs.py
+++ b/tools/system_libs.py
@@ -2292,12 +2292,17 @@ def get_libs_to_link(args):
   if settings.SIDE_MODULE:
     return libs_to_link
 
-  for forced in force_include:
-    if forced not in system_libs_map:
-      shared.exit_with_error('invalid forced library: %s', forced)
-    add_library(forced)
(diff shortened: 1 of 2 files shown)


Faless:
Thank you so much :blue_heart:  , I tested the PR, and confirmed it fixes Godot nothreads+dlink builds.

tools/system_libs.py
```diff
@@ -2390,6 +2393,10 @@ def add_sanitizer_libs():
     add_library('libwasmfs')
 
   add_sanitizer_libs()
+
+  # Add forced stdlibs at the end so that the order of system libraries is
+  # still preserved above, even when EMCC_FORCE_STDLIBS=1 is used.
```
kripken, line 2398: Could this comment go on line 2295? It seems to apply to the other calls to this function as well. Though, the call on 2328 is not actually last - could it be?

kripken (approved):

PR #22178, opened 2024-07-02
Disallow EMTEST_SAVE_DIR, which has not worked in mode 2

kripken:
At some point mode 2 there broke. To avoid confusion, remove the env var entirely,
and provide a clear error message pointing people to the proper options.

### DIFF
--- a/test/runner.py
+++ b/test/runner.py
@@ -303,7 +303,8 @@ def load_test_suites(args, modules, start_at, repeat):
   if not found_start:
     utils.exit_with_error(f'unable to find --start-at test: {start_at}')
   if total_tests == 1 or parallel_testsuite.num_cores() == 1:
-    common.EMTEST_SAVE_DIR = True
+    # TODO: perhaps leave it at 2 if it was 2 before?
+    common.EMTEST_SAVE_DIR = 1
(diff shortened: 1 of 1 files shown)


kripken:
(I think mode 2 broke because `configure()` is called twice, and we read and wrote the env var for internal reasons, so we trampled the state. To avoid that, this PR makes us not use the env var as internal state at all.)

sbc100 (approved):

aheejin:
Do we have any comment on what these values (0/1/2) for `EMTEST_SAVE_DIR` mean? I just empirically used to use `--save-dir --no-clean` because it worked, but I'm not sure what these values mean. Perhaps we can add some comments on its definition? (If this is a user-facing option, adding an entry in the doc would be better, but I guess it's mostly used by us?)

kripken:
Those values are not user-facing after this PR, and maybe I should refactor the internals to avoid them..? I wasn't sure how to rename them though in a way that is consistent with the ones around them.

But, they just mean 1 = save-dir, 2 = no-clean (which also implies save-dir, so note that you don't need both).

aheejin (approved):

PR #22183, opened 2024-07-03
Remove stray $ character from logging message. NFC

sbc100:
Here I use the JS syntax by mistake.


kripken (approved):
Worked in bash too :wink:

PR #22184, opened 2024-07-03
[embind] Generate better TS def for pointers.

### LINKED ISSUE #22182: --emit-tsd to generate nullable types for pointer types
When binding functions that return pointers (either raw or smart pointers) to a type or accept them as a argument, those values can be `nullptr` on the c++ side. We think that the TS bindings generated through `--emit-tsd` should respect this. For instance:

```cpp
struct Foo {};
Foo* foo1() { return new Foo(); };
void foo2(Foo* foo) {};

EMSCRIPTEN_BINDINGS(Bar)
{
	em::class_<Foo>("Foo");
	em::function("foo1", &foo1, em::allow_raw_pointers());
	em::function("foo2", &foo2, em::allow_raw_pointers());
}
```
will emit the following TS bindings:
```ts
export interface Foo {
  delete(): void;
}
interface EmbindModule {
  Foo: {};
  foo1(): Foo;
  foo2(_0: Foo): void;
}
```
however I would have expected 
```ts
  foo1(): Foo | null;
  foo2(_0: Foo | null): void;
```
since C++'s `nullptr` is a valid input/output for foo1/2() and therefore the same counts for the JavaScript side.

Am I missing something? 

Note that using `std::optional<Foo>` is not really an option because `Foo` is not necessarily a value type that can be copied around/easily wrapped into an optional.

```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.62 (34c1aa36052b1882058f22aa1916437ba0872690)
clang version 19.0.0git (https:/github.com/llvm/llvm-project c00ada070207979f092be9046a02fcfff8b9f9ce)
Thread model: posix
```

brendandahl:
Pointers should be allowed to be null in the TS definitions.

Fixes #22182

### DIFF
--- a/src/embind/embind_gen.js
+++ b/src/embind/embind_gen.js
@@ -44,12 +44,13 @@ var LibraryEmbind = {
   },
   $FunctionDefinition__deps: ['$createJsInvoker', '$createJsInvokerSignature', '$emittedFunctions'],
   $FunctionDefinition: class {
-    constructor(name, returnType, argumentTypes, functionIndex, thisType = null, isAsync = false) {
+    constructor(name, returnType, argumentTypes, functionIndex, thisType = null, isConstructor = false, isAsync = false) {
       this.name = name;
       this.returnType = returnType;
       this.argumentTypes = argumentTypes;
       this.functionIndex = functionIndex;
       this.thisType = thisType;
+      this.isConstructor = isConstructor;
       this.isAsync = isAsync;
     }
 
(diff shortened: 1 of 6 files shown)


src/embind/embind_gen.js
```diff
@@ -575,7 +583,7 @@ var LibraryEmbind = {
           context,
           isPureVirtual,
           isAsync) {
-    createFunctionDefinition(methodName, argCount, rawArgTypesAddr, context, true, isAsync, (funcDef) => {
+    createFunctionDefinition(methodName, argCount, rawArgTypesAddr, context, true, false, isAsync, (funcDef) => {
```
sbc100, line 586: I wonder if its time to make this function take a property bag instead of all those arguments?  (As a followup?)

sbc100 (approved):

src/embind/embind_gen.js
brendandahl: Yeah, it was probably time a few arguments ago.

PR #22187, opened 2024-07-04
Fix WebGPU limits.maxBufferSize, these are 64-bit iso 32-bit

croes:
Fix implementation of WebGPU limit queries, `maxBufferSize` is 64-bit iso 32-bit.

There was already a test for this (`browser.test_webgpu_required_limits`), but that one was failing on my end (I do have >4GB VRAM).
The WebGPU headers also declare this as 64-bit, so I think it was a small oversight in the implementation.


kainino0x (approved):
Oops, LGTM!

@lokokung could you pull this into dawn's fork? (I don't think there have been any other bugfixes since I forked it)

PR #22190, opened 2024-07-05
Copy source files into source map file

### LINKED ISSUE #22189: Embed source files in source map file
I want to generate a source map when compiling my C code that includes all the source files in the source map file itself instead of having to serve all the source code files alongside the source map file. This can be done by adding a `sourcesContent` key to the source map file. Otherwise I would have to upload 93 C files to npm with my JavaScript library.

wasm-sourcemap.py includes support for doing this already with the `--sources` argument

https://github.com/emscripten-core/emscripten/blob/b3c25673f53974ff1f30df43701dc63af7cbbdc4/tools/wasm-sourcemap.py#L37

but there's no way to pass that argument from the `emcc` command

https://github.com/emscripten-core/emscripten/blob/b3c25673f53974ff1f30df43701dc63af7cbbdc4/tools/building.py#L1107-L1111

This could be a `-ginline-source-map` option or `-gsource-map -s INLINE_SOURCE_MAP`, which would match [TypeScript's `--inlineSourceMap`](https://www.typescriptlang.org/tsconfig/#inlineSourceMap). Though I expected this to be the default behavior. Serving C files over HTTP is weird.

My use case is that I have an npm package compiled with Emscripten and my source map requires files from ../../../../../../../opt/homebrew/Cellar/emscripten/3.1.61/libexec/system/lib/ and I have no idea how I could put that path in my npm module https://github.com/tree-sitter/tree-sitter/pull/3381

verhovsky:
Closes #22189

### DIFF
--- a/tools/building.py
+++ b/tools/building.py
@@ -1127,6 +1127,7 @@ def emit_wasm_source_map(wasm_file, map_file, final_wasm):
                    wasm_file,
                    '--dwarfdump=' + LLVM_DWARFDUMP,
                    '-o',  map_file,
+                   '--sources',
                    '--basepath=' + base_path]
   check_call(sourcemap_cmd)
 
--- a/tools/wasm-sourcemap.py
+++ b/tools/wasm-sourcemap.py
@@ -61,6 +61,7 @@ class Prefixes:
     if name in self.cache:
       return self.cache[name]
 
+    result = name
     for p in self.prefixes:
       if name.startswith(p['prefix']):
         if p['replacement'] is None:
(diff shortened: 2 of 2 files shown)


verhovsky:
To test this on M1 macOS, I had to

```sh
brew install llvm
git checkout f2568054e3ab2251647e87c8bd651373e1b003b7
```

make this edit

```diff
diff --git a/tools/building.py b/tools/building.py
index 900918b44..d2fc19d60 100644
--- a/tools/building.py
+++ b/tools/building.py
@@ -39,7 +39,7 @@ logger = logging.getLogger('building')
 
 #  Building
 binaryen_checked = False
-EXPECTED_BINARYEN_VERSION = 115
+EXPECTED_BINARYEN_VERSION = 117
 
 _is_ar_cache: Dict[str, bool] = {}
 # the exports the user requested
```

Then I did 

```sh
cd ..
git clone https://github.com/tree-sitter/tree-sitter.git
cd tree-sitter
git checkout 3abb1047
```

made this edit there

```diff
 script/build-wasm | 1 +
 1 file changed, 1 insertion(+)

diff --git a/script/build-wasm b/script/build-wasm
index d22b5875..c1a37414 100755
--- a/script/build-wasm
+++ b/script/build-wasm
@@ -117,6 +117,7 @@ exported_functions=$(
 
 # Use emscripten to generate `tree-sitter.js` and `tree-sitter.wasm`
 # in the `target/scratch` directory
+emcc=../emscripten/emcc
 $emcc                                            \
   -s WASM=1                                      \
   -s INITIAL_MEMORY=33554432                     \
```

then ran

```
./scripts/build-wasm
```

and it turns out that `sourceContents` is removed by some other step (binaryen?) after if it's written with `--sources`.

PR #22204, opened 2024-07-09
Factor out html5 event faking code used in browser tests. NFC

sbc100:


test/test_browser.py
```diff
@@ -1772,6 +1674,7 @@ def test_glgears_long(self, args):
     'full_944': ('hello_world_gles_full_944.c',),
   })
   def test_glgears_animation(self, filename):
+    shutil.copyfile(test_file('browser/fake_events.js'), 'fake_events.js')
```
kripken, line 1677: Why does this copy the file instead of include it like other tests?

sbc100: Because here the code is used directly in the html via a script tag and not via `--pre-js`, so that file needs to be directly serveable over http.

kripken: I see, thanks.

kripken (approved):

PR #22217, opened 2024-07-10
Make emscripten_resize_heap compatible with PURE_WASI

### LINKED ISSUE #22211: ALLOW_MEMORY_GROWTH with PURE_WASI doesn't seems to work

```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.62-git
clang version 19.0.0git
Thread model: posix

```

```
emcc -sALLOW_MEMORY_GROWTH=1 -sSTANDALONE_WASM=1 -sPURE_WASI=1 -O3 emb.ll -o emb.wasm  
```

When using `PURE_WASI` with `ALLOW_MEMORY_GROWTH` the `memory.grow` instruction(the WASM one) seems to be removed, and prevent the memory to grow. Removing the `PURE_WASI` re-introduce the `memory.grow` instruction, but also requires the host to export `emscripten_notify_memory_growth` function.

inkeliz:
Previosly, using PURE_WASI was incompatible with
EMSCRIPTEN_MEMORY_GROWTH flag, and was not documented.

Now, using EMSCRIPTEN_MEMORY_GROWTH with PURE_WASI is possible. It will not require the host to export
emscripten_notify_memory_growth and will grow the memory as expected from EMSCRIPTEN_MEMORY_GROWTH.

Fixes #22211

### DIFF
--- a/system/lib/standalone/standalone.c
+++ b/system/lib/standalone/standalone.c
@@ -152,14 +152,16 @@ size_t emscripten_get_heap_max() {
 }
 
 int emscripten_resize_heap(size_t size) {
-#if defined(EMSCRIPTEN_MEMORY_GROWTH) && !defined(EMSCRIPTEN_PURE_WASI)
+#if defined(EMSCRIPTEN_MEMORY_GROWTH)
   size_t old_size = __builtin_wasm_memory_size(0) * WASM_PAGE_SIZE;
   assert(old_size < size);
   ssize_t diff = (size - old_size + WASM_PAGE_SIZE - 1) / WASM_PAGE_SIZE;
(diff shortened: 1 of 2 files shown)


sbc100 (approved):
lgtm!

I wonder if we can write a test for this?

sbc100:
Can you marge/rebase changes from the main branch?  It should address the current test failures.

sbc100:
It looks like this change fixes the `test_memorygrowth_MAXIMUM_MEMORY_standalone` test so that it is no longer impure.

Can you remove the `impure=True` from `@also_with_standalone_wasm` on line 2090 of test_core.py?

PR #22220, opened 2024-07-10
[NFC] Strip the names section using llvm-objcopy in the metadce tests

kripken:
This is more precise than using `wasm-opt`, as it literally just removes the names
section, and avoids any Binaryen roundtripping artifacts.

The improvements here are just the effects of removing those artifacts.


sbc100 (approved):
Awesome!

Should be a lot fast too.

PR #22225, opened 2024-07-11
Fix racy bevaviour in Ports.install_header_dir. NFC

sbc100:
This change reduces to usage of install_header_dir to just the ports that need it, and updates it to use `maybe_copy` which avoids touching files that are already up-to-date.

Fixes: #22224


dschuff (approved):
🎉

PR #22230, opened 2024-07-12
Ensure emscripten root is at the start of python path

sbc100:
This avoids issues where something called e.g. `tools` already exists in the python path.

See https://github.com/emscripten-core/emsdk/issues/1420#issuecomment-2226178120


kleisauke:
I confirm that this resolves the issue for me, thanks!

sbc100:
If you could approve then this change will auto-merge

kleisauke (approved):

PR #22241, opened 2024-07-14
Add target features section to bulk memory asm files

sbc100:
Without this the feature might not be enabled when linked into the final binary.

This fixes the lto2.test_dylink_syslibs_all which was broken by #22231 which enabled bigint, whicn in turn enabled bulk memory (by bumping the min browser version).


sbc100:
Landing TBR to make the CI green again

tlively:
LGTM!

Issue #22226, opened 2024-07-12
emsdk/upstream/emscripten/cache/sysroot/include/compat/emmintrin.h is not lib/clang/19/include/**/emmintrin.h


gblikas:
I'm looking to narrow down a build issue. I am compiling opencv4 via 

```bash
vcpkg install opencv4:wasm32-emscripten --editable
```

and have modified a bad header to `#include <emmintrin.h>` as it is a missed dependency. I did this because the documentation for [Using SIMD with WebAssembly](https://emscripten.org/docs/porting/simd.html#using-simd-with-webassembly) indicates that `_mm_setr_epi64` should be available via `#include <emmintrin.h>`. Judging by my error's output: 

```bash
/Users/gblikas/projects/vcpkg/buildtrees/opencv4/src/4.8.0-2bf495557d/modules/core/include/opencv2/core/hal/intrin_sse.hpp:252:15: error: use of undeclared identifier '_mm_setr_epi64'; did you mean '_mm_set_epi64'?
  252 |         val = _mm_setr_epi64((__m64)v0, (__m64)v1);
      |               ^~~~~~~~~~~~~~
      |               _mm_set_epi64
/Users/gblikas/projects/emsdk/upstream/emscripten/cache/sysroot/include/compat/emmintrin.h:1133:1: note: '_mm_set_epi64' declared here
 1133 | _mm_set_epi64(long long q1, long long q0)
      | ^
```

and 

```bash
$ cat /Users/gblikas/projects/emsdk/upstream/emscripten/cache/sysroot/include/compat/emmintrin.h  | grep "_setr"
_mm_setr_pd(double __c0, double __c1)
_mm_setr_epi32(int i0, int i1, int i2, int i3)
_mm_setr_epi16(short w0, short w1, short w2, short w3, short w4, short w5, short w6, short w7)
_mm_setr_epi8(char b0, char b1, char b2, char b3, char b4, char b5, char b6, char b7, char b8, char b9, char b10, char b11, char b12, char b13, char b14, char b15)
```

it appears as if the included header `emmintrin.h` from `**/upstream/emscripten/cache/sysroot/include/compat/emmintrin.h` is not the same as any of the `lib/clang/19/include/**/emmintrin.h` headers. 

I'm not sure what `emscripten/cache/sysroot` is, so this issue is most-likley user error. However, 

```bash
$ grep -lR "_mm_setr_epi64" emsdk
[...]

PR #22243, opened 2024-07-14
Update emmintrin.h

gblikas:
> PR for #22226 

I don't believe this PR is not complete, as I'm still awaiting feedback on the `setr_epi64` methods.


tlively:
It would be good to add tests for these new functions in test/sse

gblikas:
> It would be good to add tests for these new functions in test/sse

Will do

system/include/compat/emmintrin.h
```diff
@@ -1177,6 +1177,18 @@ _mm_set1_epi8(char __b)
...
+_mm_setr_epi64(__m64 q0, __m64 q1)
+{
+  return (__m128i){ (long long)q0, (long long)q1 };
+}
+
+static __inline__ __m128i __attribute__((__always_inline__, __nodebug__))
+_mm_setr_epi64(int64_t q0, int64_t q1)
+{
+  return (__m128i)wasm_i64x2_make(q0, q1);
+}
```
sbc100, line 1190: This looks like it defines the same function twice.. which I guess means the second one will be ignored?   Should we remove one of them?

Also, can we add some kind of test for this?

gblikas: @sbc100 , see https://github.com/emscripten-core/emscripten/pull/22243#issuecomment-2253229600 - in process

allsey87:
@gblikas do you need some help finishing off this PR?

PR #22247, opened 2024-07-15
[NFC] split up gen_struct_info.py's inspect_headers function

walkingeyerobot:
This helps when trying to generate struct info via another build system (i.e. bazel). I'm now able to call into these functions individually.

I think breaking up the large function into smaller and more focused functions also helps readability.

### DIFF
--- a/tools/maint/gen_struct_info.py
+++ b/tools/maint/gen_struct_info.py
@@ -102,6 +102,7 @@ DEFAULT_JSON_FILES = [
     utils.path_from_root('src/struct_info_webgpu.json'),
 ]
 
+
 def show(msg):
   if shared.DEBUG or not QUIET:
     sys.stderr.write('gen_struct_info: %s\n' % msg)
--- a/tools/maint/gen_struct_info.py
+++ b/tools/maint/gen_struct_info.py
@@ -77,6 +77,30 @@ from tools.settings import settings
 QUIET = (__name__ != '__main__')
 DEBUG = False
 
+CFLAGS = [
+    # Avoid parsing problems due to gcc specifc syntax.
+    '-D_GNU_SOURCE',
+]
+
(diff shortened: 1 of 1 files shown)


tools/maint/gen_struct_info.py
```diff
@@ -77,6 +77,31 @@
 QUIET = (__name__ != '__main__')
 DEBUG = False
 
+CFLAGS = [
+    # Avoid parsing problems due to gcc specifc syntax.
```
sbc100, line 81: Typo in specifc (I know its in the original,  and its almost certainly mine, but still worth fixing).

tools/maint/gen_struct_info.py
```diff
@@ -272,6 +279,35 @@ def inspect_headers(headers, cflags):
...
+
+
+def inspect_headers(headers, cflags):
+  # Write the source code to a temporary file.
+  src_file = tempfile.mkstemp('.c', text=True)
+  show('Generating C code... ' + src_file[1])
+  code = generate_c_code(headers)
+  os.write(src_file[0], '\n'.join(code).encode())
+  os.close(src_file[0])
+  src_file_path = src_file[1]
```
sbc100, line 292: How about doing `src_file_fd, src_file_path = tempfile.mkstemp('.c', text=True)` above instead of adding this line?

sbc100 (approved):

tools/maint/gen_struct_info.py
walkingeyerobot: Done.

tools/maint/gen_struct_info.py
walkingeyerobot: Done. Also made the temp js file match.

sbc100 (approved):

PR #22251, opened 2024-07-17
Expose `loadStructInfo` JS macro function add additional struct/macro definitions.

sbc100:
Eventually it would be nice to enable external use of the gen_struct_info.py script to generate these things, but this is a step in that direction.


walkingeyerobot (approved):
thanks so much!

src/modules.mjs
```diff
@@ -285,15 +285,23 @@ function addToLibrary(obj, options = null) {
 let structs = {};
 let defines = {};
 
+/**
+ * Read JSON file containing struct and macro/define infomation
```
kripken, line 289:
```suggestion
 * Read JSON file containing struct and macro/define information
```

kripken (approved):

Issue #17228, opened 2022-06-15
entire source of module printed when nodejs exception thrown


dvtate:
This has been an issue that I've experienced with various packages built with emscripten since 2019 and honestly surprised it's still there. It's one of the biggest painpoints of working with emscripten + node.js for me personally

```js
import binaryen from 'binaryen';

let s = "testing";

// Desparately trying to debug to figure out what problem is
// But this message gets burried in undesired garbage
console.log("s = ", s); 

// Fatal error which makes emscripten print massive unreadable textwall for no discernable reason
// filling console and making logs unreadable
s(444);
```
running the above code via `node index.js` results in the following message (which takes several seconds to print) and appears to be generated by emscripten
![image](https://user-images.githubusercontent.com/15305612/173884459-1337fd4b-1075-4da6-ba0a-f408d36aeb51.png)

I've experienced this exact problem with every emscripten package I've used including internal ones at previous company so I don't think this is the fault of the package maintainers.

kripken:
Trying to reproduce this, the following works:
* In a new directory, make an empty `package.json` file with just `{ }`.
* `npm install binaryen`
* Save the JS in the top comment as `a.mjs` - the `mjs` suffix is crucial.
* `node a.mjs` in at least node 16.

I have no idea why node would print the contents of an imported package like that... the error isn't even in the same file! The stack trace even shows that. How bizarre... removing the module import line, the behavior is as expected. Just importing that module, not even using it, is enough to make node print it on an unrelated exception.

I'm not sure how to even debug that. Advice from any node experts reading this would be very welcome...

kripken:
I can't reproduce this when generating the imported module. For example,
```
emcc tests/hello_world.c -s EXPORT_ES6 -o index.js -O3 --closure 1 -s SINGLE_FILE -s MODULARIZE
```
That emits an index.js that, when imported here, does not cause the bug. I tried lots of variations on those flags (which emit something in the same general shape as the binaryen build from npm).

If so then perhaps this is specific to `binaryen.js` somehow? cc @dcodeIO

dvtate:
Did you try calling the default export/making it actually load the wasm? I think it getting interrupted while instantiating the wasm module could be the problem. For whatever reason I'm getting an error with the `__dirname` variable (maybe Node v18 problem idk). But after hardcoding that to the correct value I was able to reproduce the issue with the hello_world.c test file.

[...]

PR #22257, opened 2024-07-18
Disable NODEJS_CATCH_EXIT by default. NFC

sbc100:
It turns out that this setting is only needed in very specific
circumstances.  Specifically its only needed if native code calls
`exit` a callback that was not wrapped in callUserCallback.

This change removes the need for this setting from all our unit tests
and disables the setting by default.

Fixes: #17228

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 3.1.64 (in development)
 -----------------------
+- The `NODEJS_CATCH_EXIT` setting is now disabled by default.  This setting
+  is only useful under very specific circumstances, and has some downsides, so
+  disabling it by default makes sense. (#22257)
 - Updated the SCons tool to not require the `EMSCRIPTEN_ROOT` environment
   variable, in which case it will assume that SCons will find the binaries in
   (its) `PATH`.
--- a/src/library_browser.js
+++ b/src/library_browser.js
@@ -770,13 +770,11 @@ var LibraryBrowser = {
   emscripten_async_load_script__deps: ['$UTF8ToString'],
   emscripten_async_load_script: (url, onload, onerror) => {
     url = UTF8ToString(url);
-    onload = {{{ makeDynCall('v', 'onload') }}};
-    onerror = {{{ makeDynCall('v', 'onerror') }}};
-
 #if PTHREADS
     if (ENVIRONMENT_IS_PTHREAD) {
       err(`emscripten_async_load_script("${url}") failed, emscripten_async_load_script is currently not available in pthreads!`);
-      return onerror ? onerror() : undefined;
+      onerror && {{{ makeDynCall('v', 'onerror') }}}();
+      return;
     }
 #endif
 #if ASSERTIONS
@@ -787,17 +785,20 @@ var LibraryBrowser = {
     var loadDone = () => {
       {{{ runtimeKeepalivePop() }}}
       if (onload) {
+        var onloadCallback = () => callUserCallback({{{ makeDynCall('v', 'onload') }}});
         if (runDependencies > 0) {
-          dependenciesFulfilled = onload;
+          dependenciesFulfilled = onloadCallback;
         } else {
-          onload();
+          onloadCallback();
         }
       }
     }
 
     var loadError = () => {
       {{{ runtimeKeepalivePop() }}}
-      onerror?.();
+      if (onerror) {
+        callUserCallback({{{ makeDynCall('v', 'onerror') }}});
+      }
     };
 
 #if ENVIRONMENT_MAY_BE_NODE && DYNAMIC_EXECUTION
--- a/test/common.py
+++ b/test/common.py
@@ -1288,10 +1288,7 @@ class RunnerCore(unittest.TestCase, metaclass=RunnerMeta):
     long_lines = []
 
     def cleanup(line):
-      if len(line) > 2048:
-        # Sanity check that this is really the emscripten program/module on
-        # a single line.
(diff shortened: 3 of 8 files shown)


RReverser (approved):
I don't fully know the implications of removing it for those who need it (in the mentioned circumstances), but personally I always had to disable it manually so 👍.

.circleci/config.yml
```diff
@@ -885,8 +880,10 @@ jobs:
       - upload-test-results
 
   test-mac-arm64:
-    executor: mac-arm64
+    executor: python/default
+    resource_class: macos.m1.medium.gen1
     environment:
+      EMSDK_NOTTY: "1"
```
kripken, line 886: What are these changes?

src/library_browser.js
```diff
@@ -770,13 +770,10 @@ var LibraryBrowser = {
...
   emscripten_async_load_script: (url, onload, onerror) => {
     url = UTF8ToString(url);
-    onload = {{{ makeDynCall('v', 'onload') }}};
-    onerror = {{{ makeDynCall('v', 'onerror') }}};
-
 #if PTHREADS
     if (ENVIRONMENT_IS_PTHREAD) {
       err(`emscripten_async_load_script("${url}") failed, emscripten_async_load_script is currently not available in pthreads!`);
-      return onerror ? onerror() : undefined;
+      return onerror ? {{{ makeDynCall('v', 'onerror') }}}() : undefined;
```
kripken, line 776: Is this change related?

test/common.py
```diff
@@ -1288,10 +1288,7 @@ def clean_js_output(self, output):
     long_lines = []
 
     def cleanup(line):
-      if len(line) > 2048:
-        # Sanity check that this is really the emscripten program/module on
-        # a single line.
-        assert line.startswith('var Module=typeof Module!="undefined"')
+      if len(line) > 2048 and line.startswith('var Module=typeof Module!="undefined"'):
```
kripken, line 1291: What new pattern of size > 2048 is now being accepted here, aside from the program/module on a single line as the old comment used to say?

test/test_fs_after_main.c
```diff
@@ -60,10 +62,10 @@ int main() {
       counter++;
       if (counter < 5) {
         out('js queueing');
-        setTimeout(looper, 1);
+        safeSetTimeout(looper, 1);
       } else {
         out('js finishing');
-        setTimeout(Module['_finish'], 1);
+        safeSetTimeout(Module['_finish'], 1);
```
kripken, line 68: Can you explain why this change is needed? A comment might be good in the code too.

test/common.py
sbc100: We had one test where we have line over 2kb but it was coming from a long line which was the minified wasm2js output, not the overall emscripten output, so it didn't start with `var Module..`

So now, such tests would not get this "REPLACED ENTIRE PROGRAM ON SINGLE LINE" hack.. only when we is clearly the entire emscripten program minified do we do this trick?  We could come up with a pattern to detect wasm2js output but I don't think its worth it.

sbc100: BTW the `len(line) > 2048` condition was always there, I just most the assertion up into a condition becuase it was firing.

src/library_browser.js
sbc100: Yes, this is part of the change.  It makes it possible to safely call `exit` from one of the `emscripten_async_load_script` callbacks by wrapping them in `callUserCallback` (rather then directly calling the).

.circleci/config.yml
sbc100: This fix landed as #22260.  Reverted here.

src/library_browser.js
kripken: How does this make that safe? This change looks NFC to me, like a cleanup - `onerror` is just moved around. What am I missing? Or is this part NFC but the code later is what makes a difference?

test/common.py
kripken: Yes, I understand that the 2048 is pre-existing, but sorry, I'm still not following what changed here. If the long content is minified wasm2js output, why is it printed out after this PR, but not before?

src/library_browser.js
sbc100: When the callbacks are called asyncronously below they are now called via callUserCallback.  Before they were not.

test/common.py
sbc100: Because one of the consequences of this change is that when node prints the line that threw an error is not prints a more accurate line, rather then printing the location of the rethrow.

Because we are no longer rethrowing at least on error is now (correctly) reported as coming from inside of wasm2js content rather than the minified emscripten-generated JS.

This code here in common.py avoids echoing the entire emscripten-genertaed module, but it doesn't (currently avoid) echoing the asmjs code (which appears on a different line).

kripken: I see now, thanks.

ChangeLog.md
```diff
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 3.1.64 (in development)
 -----------------------
+- The `NODEJS_CATCH_EXIT` setting is now disabled by default.  This setting
+  is only useful under very specific circumstances, and has some downsides, so
+  disabling it by default makes sense. (#22257)
```
```suggestion
- The `NODEJS_CATCH_EXIT` setting is now disabled by default.  This setting
  is only useful under very specific circumstances, and has some downsides, so
  disabling it by default makes sense. If you run code in Node.js and want the
  old behavior, pass `-sNODEJS_CATCH_EXIT`. (#22257)
```

kripken (approved):

PR #22260, opened 2024-07-18
[ci] Fix mac arm64 builder

sbc100:
We have been seeing the `brew install python3` step fail consistently in the last 24 hours.

I'm not really sure why this fixes the issue.


sbc100:
Will land TBR to green the tree.

PR #16895, opened 2022-05-06
[WasmFS] Allow backends to have short reads and writes

tlively:
We previously tried to prevent reads and writes beyond the ends of files by
calling `getSize` on files in the backend-independent implementations of `read`
and `write` and adjusting the length we passed to the backend accordingly. This
works fine when files have meaningful lengths, reads and writes are guaranteed
not to be shortened by the underlying APIs, and files cannot change size
concurrently with the syscall execution, but those requirements are not always
met.

Instead, change the backend API to match the `pread` and `pwrite` syscall APIs,
which allow accesses to be shorter than requested. Specifically, change the
return types to be `ssize_t` so that the accessed lengths can be returned. Error
codes are now returned as negative values, just like in the syscalls themselves.

This change will also have performance benefits, especially for the OPFS
backend, in which getting the size of a file is an asynchronous operation that
is much slower than a small read or write.

PR #22261, opened 2024-07-18
Fix partial write case when using _fd_write/doWritev/writev

### LINKED ISSUE #22258: doWritev has a race condition

Verified in current HEAD of main branch, the issue has been present since initial implementation in https://github.com/emscripten-core/emscripten/commit/fca780db9168bbd10b9a9724914a8a345eebabea

I don't have a command line that fails as my implementation of stream_ops.write exposes the issue, but I hope that I can describe the issue sufficiently here. I intend to provide a PR for this soon.

Consider a custom device for writing registered as a device as per [documentation](https://emscripten.org/docs/api_reference/Filesystem-API.html#devices), roughly:

```javascript
const dev = FS.makedev(32, 0);
FS.registerDevice(dev, io);
FS.mkdev(path, mode, dev);
```

where `io` has has a `write` method that has a 20-byte buffer.

And a C program that writes something like this device:

```c
FILE *f = fopen(.. registered device ..);
fputs("abcdefghijklmnopqrstuvwxyz", f)
fputs("\n", f);
fflush(f);
```

This C code will eventually call `__stdio_write` at which point `iovs[2]` will contain two entries, one for the `abc...` and length 26 and second entry `\n` and length 1.

`__stdio_write` calls syscall `_fd_write` and ends up in `doWritev`.

`doWritev` iterates through the `iovs` structure writing the data for each entry. However, the bug is that if a write does not write all data, the loop continues. If the output buffer that write stores to gets (partially) emptied between the two calls to write the `\n` from the second call can end up in the output buffer. This is the race condition which is why in my code I only occasionally see this. Finally, since `doWritev` will return 21 in this case, the `__stdio_write` will retry the write syscall and the data written ends up being `"abcdefghijklmnopqrst\nvwxyz\n"`

Here is the problematic code:

https://github.com/emscripten-core/emscripten/blob/cc5083ebb39660468bf3d712dfbbbe0aa1b738d8/src/library_wasi.js#L223-L227

What is needed is to abort the loop if a write does not write `len` bytes:

```javascript
      var curr = FS.write(stream, HEAP8, ptr, len, offset);
      if (curr < 0) return -1;
      ret += curr;
      if (curr < len) break; // no more space to write
      if (typeof offset != 'undefined') {
```

Note that the read equivalent `doReadv` does have this early exit for the loop:

https://github.com/emscripten-core/emscripten/blob/cc5083ebb39660468bf3d712dfbbbe0aa1b738d8/src/library_wasi.js#L209

jonahgraham:
doWritev iterates through the iovs structure writing the data for each entry. The bug fixed here is that if a write does not write all data, the loop continues. If the subsequent write also accepts data, it has accepted data out of order.

Without the fix to doWritev this new test outputs:

```
TO DEVICE: "A"
TO DEVICE: "X"
TO DEVICE: "C"
TO DEVICE: "X"
TO DEVICE: "Y"
TO DEVICE: "Z"
TO DEVICE: "a"
TO DEVICE: "\n"
TO DEVICE: "c"
TO DEVICE: "\n"
done
```

Fixes #22258

### DIFF
--- a/src/library_wasi.js
+++ b/src/library_wasi.js
@@ -223,7 +223,10 @@ var WasiLibrary = {
       var curr = FS.write(stream, HEAP8, ptr, len, offset);
       if (curr < 0) return -1;
       ret += curr;
-      if (curr < len) break; // no more space to write
+      if (curr < len) {
+        // No more space to write.
+        break;
+      }
       if (typeof offset != 'undefined') {
         offset += curr;
       }
--- a/src/library_wasi.js
+++ b/src/library_wasi.js
@@ -223,6 +223,7 @@ var WasiLibrary = {
       var curr = FS.write(stream, HEAP8, ptr, len, offset);
       if (curr < 0) return -1;
       ret += curr;
+      if (curr < len) break; // no more space to write
       if (typeof offset != 'undefined') {
         offset += curr;
       }
--- a/test/fs/test_writev_partial_write.c
+++ b/test/fs/test_writev_partial_write.c
@@ -68,7 +68,9 @@ void test_writev_direct(void) {
 // and is included here as this code most closely matches the original bug
 // report
 void test_via_stdio(void) {
-  FILE* f = fopen("/device", "w");
+  // XXX: We open in append mode because truncating JS based files is not
+  // supported yet. See #22262
+  FILE* f = fopen("/device", "a");
   assert(f);
   // Use line buffering. The bug is exposed with line buffering because with
   // line buffering two entries in __stdio_write's iovs are used.
--- a/test/fs/test_writev_partial_write.c
+++ b/test/fs/test_writev_partial_write.c
@@ -70,7 +70,11 @@ void test_writev_direct(void) {
 void test_via_stdio(void) {
   // XXX: We open in append mode because truncating JS based files is not
   // supported yet. See #22262
+#ifdef WASMFS
   FILE* f = fopen("/device", "a");
+#else
+  FILE* f = fopen("/device", "w");
+#endif
   assert(f);
   // Use line buffering. The bug is exposed with line buffering because with
   // line buffering two entries in __stdio_write's iovs are used.
--- a/test/fs/test_writev_partial_write.c
+++ b/test/fs/test_writev_partial_write.c
@@ -16,8 +16,7 @@
 
 EM_JS_DEPS(main, "$ERRNO_CODES");
 
-int main() {
-
+void setup(void) {
  EM_ASM(
     var device = FS.makedev(80, 0);
     FS.registerDevice(device, {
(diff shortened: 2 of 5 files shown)


test/test_core.py
```diff
@@ -5942,6 +5942,10 @@ def test_fs_writev(self):
     self.set_setting('FORCE_FILESYSTEM')
     self.do_runf('fs/test_writev.c', 'success')
 
+  def test_fs_writev_partial_write(self):
+    self.set_setting('FORCE_FILESYSTEM')
```
sbc100, line 5946: Is this line needed?

test/fs/test_writev_partial_write.c
```diff
@@ -0,0 +1,84 @@
...
+    FILE* f = fopen("/device", "w");
+    assert(f);
+    // Use line buffering. The bug is exposed with line buffering because with
+    // line buffering two entries in __stdio_write's iovs are used.
+    setvbuf(f, NULL, _IOLBF, 0);
+    fputs("abc", f);
+    fputs("\n", f);
+    fflush(f);
+    fclose(f);
+  }
```
sbc100, line 80: Could you split this function into 3 functions: `setup`, `test_writev_direct`, `test_via_stdio` and then call each of the from main?

sbc100 (commented):
Thanks do much for the fix and the great test case.

I wonder if its enough to put this test in test_other.py?  I..e maybe we don't need to run it all the different configurations?

test/test_core.py
jonahgraham: It isn't. I'll remove in my next commit.

test/fs/test_writev_partial_write.c
jonahgraham: Yes.

jonahgraham:
> I wonder if its enough to put this test in test_other.py? I..e maybe we don't need to run it all the different configurations?

I have moved it. I don't think it needs running in multiple configurations. I wasn't really sure where to put the test, please let me know if the new place is good.

I have added the review comments as additional commits, with the expectation that it will be squashed when merged. Let me know if you want me to squash them now.

test/test_other.py
```diff
@@ -15028,3 +15028,6 @@ def test_extra_struct_info(self):
     stderr = self.run_process([EMCC, test_file('hello_world.c'), '--js-library', test_file('other/test_extra_struct_info.js')], stderr=PIPE).stderr
     self.assertContained('(before) AF_INET=2', stderr)
     self.assertContained('(after) AF_INET=42', stderr)
+
+  def test_fs_writev_partial_write(self):
```
kripken, line 15032:
```suggestion
  @also_with_wasmfs
  def test_fs_writev_partial_write(self):
```
This will also test in WasmFS. The fix doesn't affect it, but I think it's worth verifying that we don't have this bug there (some of the WasmFS code was ported from JS).

jonahgraham: I added this, but how do I run the test with wasmfs? This is what I have been doing to run the test until now: `test/runner other.test_fs_writev_partial_write`

kripken: `test/runner other.test_fs_writev_partial_write_wasmfs` should work, that is, that annotation adds a new test like the original but with suffix `_wasmfs`.

jonahgraham: I have now figured it out: `test/runner other.test_fs_writev_partial_write_wasmfs`

However I see that the test fails with `FS.makedev is not a function` error ([link to circleci](https://app.circleci.com/pipelines/github/emscripten-core/emscripten/35922/workflows/c4064c8f-7e08-4abe-8568-7af99a0e7d75/jobs/802019?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary&invite=true#step-111-377742_39))

So I added `FORCE_FILESYSTEM` back in and it leads to a new error `UNREACHABLE executed`

I am trying to debug that now.

jonahgraham: The unreachable is here: https://github.com/emscripten-core/emscripten/blob/8fdaf16b7d84537d441a469e3ab08471c5f0265e/system/lib/wasmfs/js_impl_backend.h#L104

with this backtrace:

```
RuntimeError: unreachable
    at test_writev_partial_write.wasm.wasmfs::handle_unreachable(char const*, char const*, unsigned int) (wasm://wasm/test_writev_partial_write.wasm-00a47e2e:wasm-function[409]:0xffbe)
    at test_writev_partial_write.wasm.wasmfs::JSImplFile::setSize(long long) (wasm://wasm/test_writev_partial_write.wasm-00a47e2e:wasm-function[636]:0x1b6b0)
    at test_writev_partial_write.wasm.wasmfs::DataFile::Handle::setSize(long long) (wasm://wasm/test_writev_partial_write.wasm-00a47e2e:wasm-function[438]:0x11ff6)
    at test_writev_partial_write.wasm.doOpen(wasmfs::path::ParsedParent, int, unsigned int, wasmfs::Backend*, OpenReturnMode) (wasm://wasm/test_writev_partial_write.wasm-00a47e2e:wasm-function[433]:0x11cb1)
    at test_writev_partial_write.wasm.__syscall_openat (wasm://wasm/test_writev_partial_write.wasm-00a47e2e:wasm-function[439]:0x12095)
    at test_writev_partial_write.wasm.fopen (wasm://wasm/test_writev_partial_write.wasm-00a47e2e:wasm-function[44]:0x1a0c)
    at test_writev_partial_write.wasm.test_via_stdio (wasm://wasm/test_writev_partial_write.wasm-00a47e2e:wasm-function[26]:0x10ce)
    at test_writev_partial_write.wasm.__original_main (wasm://wasm/test_writev_partial_write.wasm-00a47e2e:wasm-function[27]:0x11c0)
    at test_writev_partial_write.wasm.main (wasm://wasm/test_writev_partial_write.wasm-00a47e2e:wasm-function[28]:0x11ee)
    at /scratch/emscripten/emscripten/test_writev_partial_write.js:639:12
```

jonahgraham: (It feels like I should be opening another issue on this, but I continue for here in the meantime)

The unreachable is because we are trying to truncate the file (I `fopen`ed it in `"w"` mode). If I open in `"a"` mode so we don't truncate it works fine.

The truncating happens here:

https://github.com/emscripten-core/emscripten/blob/d72d7226f4733af8ff993dec70198cf09a24142d/system/lib/wasmfs/syscalls.cpp#L563-L564

New/additional commit coming in a moment with this change.

jonahgraham: > `test/runner other.test_fs_writev_partial_write_wasmfs` should work, that is, that annotation adds a new test like the original but with suffix `_wasmfs`.

Somehow I blanked on seeing this comment - much appreciated and sorry for overlooking it!

test/fs/test_writev_partial_write.c
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright 2022 The Emscripten Authors.  All rights reserved.
```
kripken, line 2:
```suggestion
 * Copyright 2024 The Emscripten Authors.  All rights reserved.
```

src/library_wasi.js
```diff
@@ -223,6 +223,7 @@ var WasiLibrary = {
       var curr = FS.write(stream, HEAP8, ptr, len, offset);
       if (curr < 0) return -1;
       ret += curr;
+      if (curr < len) break; // no more space to write
```
kripken, line 226:
```suggestion
      if (curr < len) {
        // No more space to write.
        break;
      }
```

kripken (commented):
Thanks! Looks good aside from those comments.

test/test_other.py
jonahgraham: > This will also test in WasmFS. The fix doesn't affect it, but I think it's worth verifying that we don't have this bug there (some of the WasmFS code was ported from JS).

FWIW the test passes on WasmFS because this case was fixed in WasmFS as part of https://github.com/emscripten-core/emscripten/pull/16895

Specifically the same condition I added exists here:

https://github.com/emscripten-core/emscripten/blob/d72d7226f4733af8ff993dec70198cf09a24142d/system/lib/wasmfs/syscalls.cpp#L159-L164

(I have a PR in a second that fixes the copy-paste error in the comment)

src/library_wasi.js
kripken: (just a style change to match our conventions)

jonahgraham:
Thanks for the reviews and feedback. I am starting to get a hang of the test suite now. I'll revisit any more review feedback tomorrow.

kripken (approved):
Perfect, thanks!

PR #22266, opened 2024-07-19
Add --help to docs for test/runner

jonahgraham:
If you run `test/runner` with no arguments, very little is output:

```sh
$ test/runner
Test suites:
[]
```

I provide this as a simple fix to the docs, but perhaps the correct fix is to make `test/runner` behave as the docs originally explained?


kripken (approved):
Thanks! This looks good to me. We can consider separately if we want to change the behavior of calling runner with no arguments.

PR #22272, opened 2024-07-23
Support destructuring arguments in library functions.

brendandahl:


src/jsifier.mjs
```diff
@@ -598,7 +598,7 @@ function(${args}) {
         // Give the function the correct (mangled) name. Overwrite it if it's
         // already named.  This must happen after the last call to
         // modifyJSFunction which could have changed or removed the name.
-        if (contentText.match(/^\s*([^}]*)\s*=>/s)) {
+        if (contentText.match(/=>/s)) {
```
sbc100, line 601: This seems wrong..  we are trying to detect functions that are arrow functions themselves, but this seems to detect any function that contains an arrow function within it anywhere.

sbc100 (approved):

src/jsifier.mjs
brendandahl: Yeah, it will lead to some false positives. The current regex also runs into issues though if an arrow function is used within the function e.g. 

```
function abc(arg1) {
  var z = (foo)=> {};
}
```
matches.

sbc100: Are you sure?  I thought the ^ at the start means it only matches the start of the whole string?

brendandahl: Yeah try `"function abc(arg1) {\nvar z = (foo)=> {}; }".match(/^\s*([^}]*)\s*=>/s)` in the console. It does match the start of the string, but continues to the next line since the `[^}]*` eats the new line and continues on until it runs into the `}`

PR #22278, opened 2024-07-24
Prevent closure mangling of Module properties in MODULARIZE mode

kleisauke:
Avoids that Closure compiler would rename/mangle `noImageDecoding`, `noAudioDecoding`, `noWasmDecoding`, etc., in `MODULARIZE` mode.

Before:
```console
$ emcc test/hello_world.c -sMODULARIZE --closure=1 --use-preload-plugins
$ grep noImageDecoding a.out.js || echo "Closure mangled noImageDecoding"
Closure mangled noImageDecoding
```

After:
```console
$ emcc test/hello_world.c -sMODULARIZE --closure=1 --use-preload-plugins
$ grep noImageDecoding a.out.js || echo "Closure mangled noImageDecoding"
      return !f.noImageDecoding && /\.(jpg|jpeg|png|bmp)$/i.test(c);
```

Note: these type declarations need to be kept in-sync with:
https://github.com/emscripten-core/emscripten/blob/eba870222fbd2e2bc0488de7d74a232a9a590a71/src/shell.js#L27-L37

### DIFF
--- a/src/closure-externs/closure-externs.js
+++ b/src/closure-externs/closure-externs.js
@@ -235,6 +235,17 @@ var id;
  * Used in MODULARIZE mode as the name of the incoming module argument.
  * This is generated outside of the code we pass to closure so from closure's
  * POV this is "extern".
+ * @type {{
+ *   noImageDecoding: boolean,
+ *   noAudioDecoding: boolean,
+ *   noWasmDecoding: boolean,
+ *   canvas: HTMLCanvasElement,
+ *   ctx: Object,
+ *   dataFileDownloads: Object,
+ *   preloadResults: Object,
+ *   useWebGL: boolean,
+ *   expectedDataFileDownloads: number,
+ * }}
  */
 var moduleArg;
 


kripken:
Is there an advantage to declaring closure externs for these rather than quoting their use? I mean that this should also fix the issue, IIUC:
```diff
diff --git a/src/library_browser.js b/src/library_browser.js
index 99538d9d5..2332bd4b8 100644
--- a/src/library_browser.js
+++ b/src/library_browser.js
@@ -124,7 +124,7 @@ var LibraryBrowser = {
 
       var imagePlugin = {};
       imagePlugin['canHandle'] = function imagePlugin_canHandle(name) {
-        return !Module.noImageDecoding && /\.(jpg|jpeg|png|bmp)$/i.test(name);
+        return !Module['noImageDecoding'] && /\.(jpg|jpeg|png|bmp)$/i.test(name);
       };
       imagePlugin['handle'] = function imagePlugin_handle(byteArray, name, onload, onerror) {
         var b = new Blob([byteArray], { type: Browser.getMimetype(name) });
```
And for properties that are used in multiple places, we have a helper function that grabs them (quoted) to a local for later usage, which is more compact than multiple quoted names,
```diff
diff --git a/src/library_browser.js b/src/library_browser.js
index 99538d9d5..98f88b6b4 100644
--- a/src/library_browser.js
+++ b/src/library_browser.js
@@ -124,7 +124,7 @@ var LibraryBrowser = {
 
       var imagePlugin = {};
       imagePlugin['canHandle'] = function imagePlugin_canHandle(name) {
-        return !Module.noImageDecoding && /\.(jpg|jpeg|png|bmp)$/i.test(name);
+        return !noImageDecoding && /\.(jpg|jpeg|png|bmp)$/i.test(name);
       };
       imagePlugin['handle'] = function imagePlugin_handle(byteArray, name, onload, onerror) {
         var b = new Blob([byteArray], { type: Browser.getMimetype(name) });
diff --git a/src/settings.js b/src/settings.js
index 4bdc2e8f1..0036b3bcd 100644
--- a/src/settings.js
+++ b/src/settings.js
@@ -1008,7 +1008,8 @@ var INCOMING_MODULE_JS_API = [
   'preMainLoop', 'preRun',
   'preinitializedWebGLContext', 'preloadPlugins',
   'print', 'printErr', 'quit', 'setStatus', 'statusMessage', 'stderr',
-  'stdin', 'stdout', 'thisProgram', 'wasm', 'wasmBinary', 'websocket'
+  'stdin', 'stdout', 'thisProgram', 'wasm', 'wasmBinary', 'websocket',
+  'noImageDecoding',
 ];
 
 // If set to nonzero, the provided virtual filesystem if treated
diff --git a/src/shell.js b/src/shell.js
index bb50bb820..14d863315 100644
--- a/src/shell.js
+++ b/src/shell.js
@@ -478,6 +478,7 @@ checkIncomingModuleAPI();
 {{{ makeModuleReceive('arguments_', 'arguments') }}}
 {{{ makeModuleReceive('thisProgram') }}}
 {{{ makeModuleReceive('quit_', 'quit') }}}
+{{{ makeModuleReceiveWithVar('noImageDecoding') }}}
 
 // perform assertions in shell.js after we set up out() and err(), as otherwise if an assertion fails it cannot print the message
 #if ASSERTIONS
```

kleisauke:
Quoting these occurrences could be an alternative. I considered this approach, but then I noticed that `Module.canvas` and `Module.ctx` needed similar adjustments, which would require a larger changeset.

Another advantage of this approach is that it would help users who utilize `--js-library`, `--pre-js` and/or `--post-js`. For example:
```console
$ echo "console.info(Module.ctx);" > pre.js
$ emcc test/hello_world.c -sMODULARIZE --closure=1 --pre-js=pre.js
$ grep console.info a.out.js
console.info(c.ctx);
```
(previously, `Module.ctx` would have been mangled in this case)

kripken:
Good points. I'm not sure what's best here. @sbc100 what do you think?

sbc100:
For normal incoming API element I believe @kripken is correct and we should stick to the pattern of using `makeModuleReceiveWithVar` et el.

Can you make this change for `noImageDecoding` as a separate PR, since that is clearly missing from `INCOMING_MODULE_JS_API`

I think properties such as `canvas` may be special in some way but I cannot remember exactly how.   Will investigate.

sbc100:
For `Module.ctx` that seems to be an internal global that cannot be set from the outside, so it should be minified.

kleisauke:
Closing now that PR #22326 has been merged.

PR #22282, opened 2024-07-25
Add WebP decoding support in file preloading

kripken:


sbc100 (approved):
I wonder why github doesn't render the image as part of review?

PR #22283, opened 2024-07-25
Fix order of resolve/reject in File URI handling code

### LINKED ISSUE #22248: [BUG]: Incorrect Parameter Order in Promise Constructor Causes Program Errors
https://github.com/emscripten-core/emscripten/blob/5af95da75d22bb1795777c1c6c319e84f59f5114/src/web_or_worker_shell_read.js#L26-L38

The first parameter of a Promise is resolve, and the second parameter is reject. This is a consensus on MDN, and this mistake caused an error in my program.

[MDN Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#chained_promises)

```js
const myPromise = new Promise((resolve, reject) => {
  setTimeout(() => {
    resolve("foo");
  }, 300);
});

myPromise
  .then(handleFulfilledA, handleRejectedA)
  .then(handleFulfilledB, handleRejectedB)
  .then(handleFulfilledC, handleRejectedC);
```

kripken:
I'm not sure how to test this, as it only affects File URIs.

Fixes #22248


sbc100 (approved):
I think this requires testing under Cordova / Electron.  IIRC somebody was working on adding that.

PR #22294, opened 2024-07-31
Fix glDrawElements emulation

eukarpov:
This patch fixes an issue related to defining the buffer size before drawing. It introduces a slight overhead of O(N), however, it should not significantly impact performance. One option to prevent this overhead is to use glDrawRangeElements from OpenGL ES 3.0. However, the current glDrawRangeElements emulation reuses glDrawElements and should be rewritten.


src/library_webgl.js
```diff
@@ -3794,19 +3794,37 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
...
       GLctx.bufferSubData(0x8893 /*GL_ELEMENT_ARRAY_BUFFER*/,
                           0,
                           HEAPU8.subarray(indices, indices + size));
+      
+      // Calculating vertex count if shader's attribute data is on client side
+      if (count > 0) {
+        for (var i = 0; i < GL.currentContext.maxVertexAttribs; ++i) {
+          var cb = GL.currentContext.clientBuffers[i];
+          if (cb.clientside && cb.enabled) {
+            var array_classes = {
```
kripken, line 3811:
```suggestion
            var arrayClasses = {
```

src/library_webgl.js
```diff
@@ -3794,19 +3794,37 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
...
+                5121 /* GL_UNSIGNED_BYTE */: Uint8Array,
+                5123 /* GL_UNSIGNED_SHORT */: Uint16Array,
+                5125 /* GL_UNSIGNED_INT */: Uint32Array};
+            if (array_classes[type])
```
kripken, line 3815:
```suggestion
            if (array_classes[type]) {
```

src/library_webgl.js
```diff
@@ -3794,19 +3794,37 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
...
+                vertexes = Math.max.apply(null, new array_classes[type](HEAPU8.buffer, indices, count)) + 1;
```
kripken, line 3816:
```suggestion
              vertexes = Math.max.apply(null, new array_classes[type](HEAPU8.buffer, indices, count)) + 1;
```

src/library_webgl.js
```diff
@@ -3794,19 +3794,37 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
...
                           0,
                           HEAPU8.subarray(indices, indices + size));
+      
+      // Calculating vertex count if shader's attribute data is on client side
+      if (count > 0) {
+        for (var i = 0; i < GL.currentContext.maxVertexAttribs; ++i) {
+          var cb = GL.currentContext.clientBuffers[i];
+          if (cb.clientside && cb.enabled) {
+            var array_classes = {
+                5121 /* GL_UNSIGNED_BYTE */: Uint8Array,
```
kripken, line 3812: Please write out these constants like 5121 in hex, as we do elsewhere.

src/library_webgl.js
kripken, line 3815: What happens if the type is not one of the known classes? Should we not error?

kripken (commented):
I have a few comments here, mostly style.

Can we add a test for this? We do have tests for FULL_ES2, e.g.

https://github.com/emscripten-core/emscripten/blob/211716995699da51940f07360d30b189d2584bd3/test/third_party/cubegeom/cubegeom_pre_vao_es.c

which is tested by

`./test/runner.py browser.test_cubegeom_pre_vao_es`

Can that test perhaps be modified to also test this, or is a new test easy to add? Without a test I worry this might regress in the future.

src/library_webgl.js
juj: Per GL spec it would be correct for a GL_INVALID_ENUM error to be raised, so that execution can continue. Although given that this is on a rare emulation path, I think it would be ok to let this throw an exception on the next line.

juj:
The per-draw-call temp JavaScript garbage that is generated in emulation code that counts the max index is worrying, as past experience shows that browsers will quickly deteriorate to microstuttering in animation in such cases. Ideally I think the code would avoid generating temp JS objects.

However since this is for relatively rarely used emulation path, that is fine to me either way. LGTM.

kripken:
> Ideally I think the code would avoid generating temp JS objects.

Good point, +1 - let's avoid the object creation here.

src/library_webgl.js
eukarpov: Changed.

src/library_webgl.js
eukarpov: Changed.

src/library_webgl.js
eukarpov: The code has been refactored and replaced with array.reduce.

src/library_webgl.js
eukarpov: Changed.

src/library_webgl.js
eukarpov: Handling for unknown types has been added.

src/library_webgl.js
eukarpov: browser.test_cubegeom_pre_vao_es has been extended with client-side vertex array testing. The issue was reproducible without the fix. Thanks for pointing out this test, it was helpful.

eukarpov:
> > Ideally I think the code would avoid generating temp JS objects.
> 
> Good point, +1 Let's avoid the object creation here.

For now the code has been changed with array.reduce. Do you mean `new arrayClass` should be rewritten by using a loop and the buffer directly?

kripken:
By object creation, we mean the line with `const arrayClasses = { .. }`. That allocates a new object each time it is reached (maybe not each loop iteration if the JS VM is smart, but each call, at least). A sequence of ifs could avoid that, and should be fast enough here. If there were very many cases, another option would be to allocate an object once in the global scope, but I don't think we need that here.

src/library_webgl.js
```diff
@@ -3794,19 +3794,47 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
...
+                0x1403 /* GL_UNSIGNED_SHORT */: Uint16Array,
+#if FULL_ES3
+                0x1405 /* GL_UNSIGNED_INT */: Uint32Array
+#endif
+            };
+
+            const arrayClass = arrayClasses[type];
+            if (!arrayClass) {
+              GL.recordError(0x502 /* GL_INVALID_OPERATION */);
+              err('type is not supported in glDrawElements');
```
kripken, line 3822: This logging should be guarded by `#if GL_ASSERTIONS`, see e.g. line 336.

eukarpov:
> By object creation, we mean the line with `const arrayClasses = { .. }`. That allocates a new object each time it is reached (maybe not each loop iteration if the JS VM is smart, but each call, at least). A sequence of ifs could avoid that, and should be fast enough here. If there were very many cases, another option would be to allocate an object once in the global scope, but I don't think we need that here.

Thanks for the clarification. The type processing has been changed.

src/library_webgl.js
eukarpov: Changed.

src/library_webgl.js
```diff
@@ -3803,19 +3803,47 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
...
+#endif
+              default:
+                GL.recordError(0x502 /* GL_INVALID_OPERATION */);
+#if GL_ASSERTIONS
+                err('type is not supported in glDrawElements');
+#endif
+                return;
+            }
+
+            vertexes = new arrayClass(HEAPU8.buffer, indices, count).reduce((max, current) => Math.max(max, current)) + 1;
```
kripken, line 3835: We could perhaps avoid this `new arrayClass` allocation, but the code would get larger and more complex, so I think this is good as it is.

kripken (approved):
Thanks!

caiiiycuk:
Great! Thank you.

PR #22296, opened 2024-07-31
[test] Remove `output_nicerizer` test feature. NFC

sbc100:
After #22295 there are no more required uses of this feature

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -3107,8 +3107,7 @@ The current type of b is: 9
         return 0;
       }
       '''
-    self.do_run(src, 'Sort with main comparison: 5 4 3 2 1 *Sort with lib comparison: 1 2 3 4 5 *',
-                output_nicerizer=lambda x: x.replace('\n', '*'))
+    self.do_run(src, 'Sort with main comparison: 5 4 3 2 1 \nSort with lib comparison: 1 2 3 4 5 \n')
 
   @needs_dylink
   def test_dlfcn_data_and_fptr(self):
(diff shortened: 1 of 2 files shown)


test/test_core.py
```diff
@@ -3086,18 +3077,17 @@ def test_dlfcn_qsort(self):
...
-        lib_handle = dlopen("liblib.so", RTLD_NOW);
+        void* lib_handle = dlopen("liblib.so", RTLD_NOW);
         if (lib_handle == NULL) {
           printf("Could not load lib.\\n");
           return 1;
         }
-        getter_ptr = (CMP_TYPE (*)()) dlsym(lib_handle, "get_cmp");
-        if (getter_ptr == NULL) {
+        cmp_t lib_cmp = (cmp_t)dlsym(lib_handle, "lib_cmp");
+        if (lib_cmp == NULL) {
```
kripken, line 3086: These changes look separate, or am I not seeing the connection?

sbc100: I just simplified the test a little while I was trying to see what it was doing.

sbc100: Reverted these changes

test/test_core.py
```diff
@@ -5443,15 +5442,11 @@ def test_files(self):
 
     self.do_run_in_out_file_test('test_files.c')
 
-  def test_files_m(self):
-    # Test for Module.stdin etc.
-    # needs to flush stdio streams
-    self.set_setting('EXIT_RUNTIME')
```
kripken, line 5449: It sounds like this was testing this option before? Or am I misunderstanding the comment?

sbc100: Just adding a final newline is enough to flush the stdio streams.  No need for the explicit `EXIT_RUNTIME` here.  I don't think this pertains to the actual test, which is why I removed it.

kripken (approved):

Issue #22188, opened 2024-07-04
In release mode, the "minification" of shell.html remove spaces that should not be removed 


ypujante:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
3.1.61

I have a custom `shell.html` that I provide to emcc via the `--shell-file=` option.

When I do a build in release mode (with optimizations), `shell.html` gets minified.

Here is an excerpt:

```html
<p>This tool has no tracking, ads or server-side component: it is running 100% in your browser. It uses<a href=https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage>localStorage</a>to store information across browser sessions.</p><p>You can provide feedback or report issues on<a href=https://github.com/pongasoft/webgpu-shader-toy>GitHub</a>.</p>
```

This is the original:

```html
<p>This tool has no tracking, ads or server-side component: it is running 100% in your browser. It uses <a href="https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage">localStorage</a> to store information across browser sessions. </p>
<p>You can provide feedback or report issues on <a href="https://github.com/pongasoft/webgpu-shader-toy">GitHub</a>.</p>
```

As you can see all the spaces BEFORE and AFTER all anchor tags are removed which makes the rendering wrong and TBH quite unreadable...

<img width="1393" alt="Screenshot 2024-07-04 at 11 46 26" src="https://github.com/emscripten-core/emscripten/assets/357116/053dcd61-36cc-4e06-8f30-fda362508a97">

kripken:
The tool we use to minify HTML is `html-minifier-terser`, called from here:

https://github.com/emscripten-core/emscripten/blob/217010a223375e6e9251669187d406ef2ddf266e/tools/link.py#L2565

Perhaps adjusting the flags there can fix this? Or perhaps this is a bug in that project?

sbc100:
@ypujante I've been wondering if `--shell-file` is actually a very useful feature.  Would it not be simpler to just maintain your html file outside of emscripten?  Is there any advantage you get from having emscripten process/use your html?

IIUC most folks who use emscripten just have it generate the js + wasm and take a care of maintain there own html on the side.  Would that work for you?

ypujante:
@sbc100 I found another issue with the minification process:

original:
```html
<input type="text" value="-" disabled>
```

minified:
```html
<input value=- disabled>
```

[...]

PR #22297, opened 2024-07-31
Remove support for MINIFY_HTML

sbc100:
This setting has been causing some problems (see #22188) so I think we should just remove it.

My understanding is that most users of emscripten use it to generate JS and not html.  Going forward I'm tempted to also remove the `--shell-file` command line flag since I'm not sure there any many compelling use cases for a custom shell html.  While html generation is certainly useful for quick testing and examples, its not really something web sites are likely to use in production since they will likely have their own existing flow for build html.

Fixes: #22188

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -24,6 +24,9 @@ See docs/process.md for more on how version tagging works.
   is only useful under very specific circumstances, and has some downsides, so
   disabling it by default makes sense. (#22257)
 - Add WebP (`.webp`) decoding support in file preloading. (#22282)
+- The `MINIFY_HTML` settings, along with the internal use of terser to minify
+  html output was removed.  Folks who want to run terser of their html can
+  continue to do so via an explict separate build step. (#22297)
 
 3.1.64 - 07/22/24
 -----------------------
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,7 @@
         "@babel/core": "^7.24.7",
         "@babel/preset-env": "^7.24.7",
         "acorn": "^8.12.0",
-        "google-closure-compiler": "20230802.0.0",
-        "html-minifier-terser": "7.2.0"
+        "google-closure-compiler": "20230802.0.0"
       },
       "devDependencies": {
         "es-check": "^7.2.1",
@@ -1852,6 +1851,7 @@
       "version": "0.3.5",
       "resolved": "https://registry.npmjs.org/@jridgewell/source-map/-/source-map-0.3.5.tgz",
       "integrity": "sha512-UTYAUj/wviwdsMfzoSJspJxbkH5o1snzwX0//0ENX1u/55kkZZkcTZP6u9bwKGkv+dkk9at4m1Cpt0uY80kcpQ==",
+      "dev": true,
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.3.0",
         "@jridgewell/trace-mapping": "^0.3.9"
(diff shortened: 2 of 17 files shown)


kripken:
If the problem in #22188 is unfixable then I agree, we should remove this feature as it is broken. But if it is fixable then I think we'd just be pushing the problem forward to users that want this minification? In that case I think we should try to fix it here once in upstream.

Or, to put it another way, I think HTML minification is reasonable for us to do, the same as JS minification. It's a good out of the box experience. But if there aren't tools that easily work for us, then I agree we should give up on it.

Akaricchi:
I still use a hacky custom shell to deploy the web builds of Taisei, though I'm not too happy about it. I'd like to properly modularize it eventually, but I have other priorities (as a very much not-a-web-dev, I find every excuse to not touch that stuff).

I've also encountered #22188, and my "solution" was to replace the mis-minified spaces with `&nbsp;` tokens.

sbc100:
> If the problem in #22188 is unfixable then I agree, we should remove this feature as it is broken. But if it is fixable then I think we'd just be pushing the problem forward to users that want this minification? In that case I think we should try to fix it here once in upstream.

I think fixing it would involved digging deep into how terser html minifiction works and understanding all the edge cases.  Not something I have much of an appetite for, especially if its not a feature anyone actually needed/uses.
> 


> Or, to put it another way, I think HTML minification is reasonable for us to do, the same as JS minification. It's a good out of the box experience. But if there aren't tools that easily work for us, then I agree we should give up on it.

I disagree.  Almost all of our users rely on out generated JS and depend on it being small.  Almost none of our user, I imagine, are depending on emscripten to build their html file.   Why would they?   This feature is great so small demos and hello world examples and testing, but I don't see any use of it in production.  In production surely you want to author your own HTML from scratch.

kripken:
@sbc100 sounds reasonable, but I'm not actually sure we know how many users use our HTML generation and optimization. I assume @juj does, but I could be wrong?

Might also be worth asking on the mailing list. If we can't find a significant amount of users then I am ok with removing this.

juj:
> This setting has been causing some problems (see #22188) so I think we should just remove it.

I'm not sure if this is the best approach. Why drop useful features just because a user runs into a bug?

I use this html-minifier based approach in my engine builds, and it presents a full solution from CMake to final output, especially for SINGLE_FILE cases. It is very convenient.

Just because there are many users that are building to .js does not mean that building to .html with minification could not be useful to other sets of users?

I've posted a PR https://github.com/emscripten-core/emscripten/pull/22346 to fix the root issue in the minification, and an upstream ticket https://github.com/terser/html-minifier-terser/issues/179 to convey the unexpected behavior to the developers of the minifier.

sbc100:
Thanks for the info @juj and for filing the bug.


Can I ask about "I use this html-minifier based approach in my engine builds, and it presents a full solution from CMake to final output, "?   Are you using this feature in combination with `--shell-file` to produce production builds?   Or are you using the emscripten-default html for debugging/testing?

Akaricchi:
> Are you using this feature in combination with --shell-file to produce production builds?

Not the person you're asking, but I am currently using it like that.

sbc100:
> > Are you using this feature in combination with --shell-file to produce production builds?
> 
> Not the person you're asking, but I am currently using it like that.

Is there some reason you using emscripten to produce the html file?  I'm trying to understand the advantage to using this feature.  Why not just ship the html that you have already written?

Akaricchi:
> Is there some reason you using emscripten to produce the html file? I'm trying to understand the advantage to using this feature. Why not just ship the html that you have already written?

I suppose I could do that. It's just nice to have html minification without getting any web-specific tools involved, as @juj said. It won't be a huge problem for me if this is removed, but it sure is annoying having to unbreak my build system every time I update Emscripten. I build for a variety of platforms and this toolchain is the only one that has this problem.

juj:
> Are you using this feature in combination with `--shell-file` to produce production builds?

Yes I am. For examples, [Bat Chase](http://clb.confined.space/batchase/) and [Spider Solitaire](https://oummg.com/spider/) both build to HTML. First one uses default minimal runtime shell, the second one uses a custom --shell-file directive for minimal runtime.

juj:
> Is there some reason you using emscripten to produce the html file? I'm trying to understand the advantage to using this feature.

This allows one to integrate the generation of the page via the build system.

> Why not just ship the html that you have already written?

I don't have a separate html page. I write the shell html file to be the page.

sbc100:
Fair enough.  I was kind of hoping that folks could use external tools or build systems to produce the html they need.   It seems like something that a separate tool could do, but I sounds like MINIMAL_RUNTIME in particular has more deep integration with the HTML output.

As usual, I'm trying to reduce the number of things emcc itself does to try to bring down the complexity and testing surface.   I'll abandon this effort for not.

PR #22298, opened 2024-07-31
Avoid defining `Module` in `shell_minimal.js`. NFC

sbc100:
Instead we can just always use the same technique to define `Module` in `shell_minimal.js`, regardless of the target environment.

This simplifies the code and reduces code size.

### DIFF
--- a/src/shell_minimal.js
+++ b/src/shell_minimal.js
@@ -16,10 +16,9 @@ if (!Module) /** @suppress{checkTypes}*/Module =
 #endif
   {"__EMSCRIPTEN_PRIVATE_MODULE_EXPORT_NAME_SUBSTITUTION__":1};
 
-#elif ENVIRONMENT_MAY_BE_NODE || ENVIRONMENT_MAY_BE_SHELL
-
-// When running on the web we expect Module to be defined externally, in the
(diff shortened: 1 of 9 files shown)


kripken (approved):

src/shell_minimal.js
```diff
@@ -28,9 +27,6 @@ var Module =
   // Otherwise do a good old typeof check.
   typeof {{{ EXPORT_NAME }}} != 'undefined' ? {{{ EXPORT_NAME }}} : {};
 #endif
-
-#else
-var Module = {{{ EXPORT_NAME }}};
```
juj, line 33: It looks like this reduces code size only for users who do not need to define `Module` in .html file at all. (if user does need to define Module in .html, then the original `var Module = {{{ EXPORT_NAME }}};` definition would be shorter.

Although I think that is ok.

It may be simpler to further condense the code to

```js
#elif SUPPORTS_GLOBALTHIS
var Module = globalThis.{{{ EXPORT_NAME }}} || {};
#else
var Module = typeof {{{ EXPORT_NAME }}} != 'undefined' ? {{{ EXPORT_NAME }}} : {};
#endif
```

juj (approved):

Issue #22300, opened 2024-07-31
WebGPU: Wrong values returned from wgpuBufferGetMapState


pkomon-tgm:
`wgpuBufferGetMapState` returns values that don't align with the corresponding enum `WGPUBufferMapState` in `webgpu.h`.

`wgpuBufferGetMapState` is implemented here 

https://github.com/emscripten-core/emscripten/blob/87ef02c43974ef645e9fe0232fab6a6baf086071/src/library_webgpu.js#L2007-L2010

and uses `Int_BufferMapState` to map the states from strings to ints

https://github.com/emscripten-core/emscripten/blob/87ef02c43974ef645e9fe0232fab6a6baf086071/src/library_webgpu.js#L399-L403

and returns those. However, the values for the corresponding C enum are

https://github.com/emscripten-core/emscripten/blob/87ef02c43974ef645e9fe0232fab6a6baf086071/system/include/webgpu/webgpu.h#L270-L275

This means, calls to `wgpuBufferGetMapState` return `0` for `'umapped'`, `1` for `'pending'`, `2` for `'mapped'` but the corresponding enum values in `webgpu.h` (thus expected return values) are `1` for `'unmapped'`, `2` for `'pending'` and `3` for `'mapped'`. We noticed the issue because we got `0`, which we didn't expect, since its not even a value of the enum.

PR #22301, opened 2024-07-31
Fix wrong return values for wgpuBufferGetMapState

pkomon-tgm:
Fixes https://github.com/emscripten-core/emscripten/issues/22300.

### DIFF
--- a/src/library_webgpu.js
+++ b/src/library_webgpu.js
@@ -397,9 +397,9 @@ var LibraryWebGPU = {
 
     // Map from enum string back to enum number, for callbacks.
     Int_BufferMapState: {
-      'unmapped': 0,
-      'pending': 1,
-      'mapped': 2,
+      'unmapped': 1,
+      'pending': 2,
+      'mapped': 3,
     },
     Int_CompilationMessageType : {
       'error': 0,


gozetor:
Hello!
Can this PR be merged? It really blocks our WebGPU adoption.
Thank you!

kripken (approved):
This looks trivial enough that I think I can review it myself. Looks correct!

sbc100:
@kainino0x would it be reasonable to have a test for this?

kainino0x:
In principle it would be good to do that (not just for these tables but also for the codegen-time tables), however our focus is currently on Dawn's fork of these bindings, and we've already obsoleted this problem there by [autogenerating](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/generator/templates/emdawnwebgpu/library_webgpu_enum_tables.js) all of these tables so they can't get out of sync.

On the Emscripten side, things should hopefully not break too quickly since we're not actively developing Emscripten's version right now. When we unfork and port the work from Dawn back into Emscripten, we'll definitely try to avoid reintroducing the manually-maintained tables (though I don't know precisely how we'll do it yet).

src/library_webgpu.js
```diff
@@ -397,9 +397,9 @@ var LibraryWebGPU = {
 
     // Map from enum string back to enum number, for callbacks.
     Int_BufferMapState: {
-      'unmapped': 0,
-      'pending': 1,
-      'mapped': 2,
+      'unmapped': 1,
+      'pending': 2,
+      'mapped': 3,
```
sbc100, line 402: It looks like `Int_CompilationMessageType` might also be off by one:

```
system/include/webgpu/webgpu.h:    WGPUCompilationMessageType_Error = 0x00000001,
system/include/webgpu/webgpu.h:    WGPUCompilationMessageType_Warning = 0x00000002,
system/include/webgpu/webgpu.h:    WGPUCompilationMessageType_Info = 0x00000003,
```

How did this stuff ever work?  Or is it just not tested?

kainino0x: A lot of these are pretty niche and most applications don't use them. It probably broke at some point when we updated the header (which shifted the numbers) but no one was using it.

PR #22302, opened 2024-07-31
Fix --emit-tsd when worker environment is required

m4burns:
This PR fixes the following issue when combining `--emit-tsd` with `-sWASM_WORKERS` or `-sPROXY_TO_WORKER`:

```
$ echo 'int main(){}' > foo.c
$ emcc -sWASM_WORKERS -lembind --emit-tsd foo.d.ts foo.c
emcc: error: when building with multithreading enabled and a "-sENVIRONMENT=" directive is specified, it must include "worker" as a target! (Try e.g. -sENVIRONMENT=web,worker)
```

When `run_embind_gen` in `link.py` sets up a separate linking pass to generate the temporary executable that will emit bindings, it doesn't include the `worker` environment required when using `-sWASM_WORKERS` or `-sPROXY_TO_WORKER`. This PR simply adds the `worker` environment when either of these flags is present.

Note: It's not sufficient to set `settings.WASM_WORKERS = 0` in `run_embind_gen`, as linking the binding-emitting executable will fail if the code depends on any WASM_WORKERS features.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -3353,6 +3353,14 @@ More info: https://emscripten.org
     self.emcc(test_file('other/embind_tsgen.cpp'), extra_args)
     self.assertFileContents(test_file('other/embind_tsgen_ignore_3.d.ts'), read_file('embind_tsgen.d.ts'))
 
+  def test_embind_tsgen_worker_env(self):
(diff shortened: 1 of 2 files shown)


brendandahl (approved):
Looks good thanks.

PR #22303, opened 2024-07-31
[contrib.glfw3] new project version 3.4.0.20240731

ypujante:
This contains 2 new versions with the following release notes for the underlying project:

#### 3.4.0.20240731 - 2024-07-31

- Added `emscripten_glfw_get_clipboard_string` the C version of  `emscripten::glfw3::GetClipboardString` to
  retrieve the clipboard asynchronously
- Added a helper class `emscripten::glfw3::FutureClipboardString` to greatly simplify the more frequent use-cases
- `GetClipboardString::value()` now returns the internal clipboard in case of error, instead of throwing exception
- Added `optimizationLevel` option to the emscripten port

#### 3.4.0.20240727 - 2024-07-27

- Introduced C++ API (namespace `emscripten::glfw3`) included with `GLFW3/emscripten_glfw3.h`:
  - provides a more correct API with sensible defaults (ex: `std::string_view` / `std::optional<std::string_view>` 
    vs `char const *` which may or may not be `nullptr`)
  - allow for C++ only API (ex: `std::future`)
  - the C API is still available if you would rather stick to it
- Implemented  `emscripten::glfw3::GetClipboardString` which provides a way of fetching the global
  clipboard in a browser environment (`glfwGetClipboardString` is not the right API due to the asynchronous nature 
  of the underlying platform API).
- The cursor position is no longer clamped to the window size, and as a result, can have negative values or values
  greater than the window size.
  Note that GLFW implements a similar behavior on the macOS desktop platform.
- Implemented `glfwSetWindowPosCallback`
- Added support for GLFW Window Attribute `GLFW_HOVERED`
- Fixed [#6](https://github.com/pongasoft/emscripten-glfw/issues/6): _`emscripten_glfw_make_canvas_resizable` does not clean up properly_.
- Fixed an issue with opacity: when using opacity, the handle is not working unless its z-index is higher than the 
  canvas z-index

### DIFF
--- a/tools/ports/contrib/glfw3.py
+++ b/tools/ports/contrib/glfw3.py
@@ -38,6 +38,7 @@ opts: Dict[str, Union[str, bool]] = {
 
 port_name = 'contrib.glfw3'
 
+
 def get_lib_name(settings):
   return (f'lib_{port_name}-O{opts["optimizationLevel"]}' +
           ('-nw' if opts['disableWarning'] else '') +
--- a/tools/ports/contrib/glfw3.py
+++ b/tools/ports/contrib/glfw3.py
@@ -4,32 +4,42 @@
 # found in the LICENSE file.
 
 import os
-from typing import Dict
+from typing import Union, Dict
 
-TAG = '3.4.0.20240627'
-HASH = '6598834deece7087fd5dda14ec6d410ae651c39b9955eb050dd736a7d3eb650075fc69cf70340f4f0514bef63723a5bbcc315397ec44ba7cab46e59fa137e27f'
+TAG = '3.4.0.20240731'
+HASH = '1d348f2a6423def537bc11ba5a67347d23696f623e0155e315711d0e23e9b4e6f623019c24c699b6dd5b727322f7093af804d58fc48488f37888ba17300c3aa8'
 
 # contrib port information (required)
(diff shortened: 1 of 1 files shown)


ypujante:
@sbc100 Hi Sam. All tests are passing. Any chance my PR gets in before the next emscripten release? I have some changes (async clipboard access) that are pretty important for a problem I am trying to solve in ImGui.

I also wanted to let you know a few things:

* `contrib.glfw3` is now officially supported in ImGui since the latest release ([1.91.0](https://github.com/ocornut/imgui/releases/tag/v1.91.0)) and that is really cool :). As far as I know this is the first project with official support.
* I released an actual application/tool that uses `contrib.glfw3` and ImGui: [WebGPU Shader Toy](https://pongasoft.github.io/webgpu-shader-toy/)
* For the purpose of this tool, I actually turned ImGui into an [emscripten port](https://github.com/pongasoft/webgpu-shader-toy/blob/master/external/emscripten-ports/imgui.py) which is by far the most complex port I wrote so far. But it is really cool because it makes using ImGui trivial: `--use-port=imgui.py:backend=glfw:renderer=wgpu`

sbc100 (approved):

ypujante:
Thank you @sbc100!

PR #22304, opened 2024-07-31
Fix CheckTypeSize when flags imply shared memory

m4burns:
This PR fixes emscripten's patched `CheckTypeSize` CMake module to work when compiler flags imply shared memory use.

Before this change, trying to configure a CMake project that uses `check_type_size` when compiler flags implied shared memory (e.g. `-sWASM_WORKERS`) caused an error:

```
usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_ae175.dir/link.txt --verbose=1
/.../emcc  -O3 -msimd128 -sWASM_WORKERS  @CMakeFiles/cmTC_ae175.dir/objects1 -o cmTC_ae175.js @CMakeFiles/cmTC_ae175.dir/linkLibs.rsp
error: library_pthread_stub.js: library_pthread_stub.js:11: #error "STANDALONE_WASM does not support shared memories yet"
```

This PR resolves the issue by changing `oformat` to `bare`, which skips the post-link phase. Type size information can still be extracted from the output.

### DIFF
--- a/cmake/Modules/CheckTypeSize.cmake
+++ b/cmake/Modules/CheckTypeSize.cmake
@@ -113,7 +113,7 @@ function(__check_type_size_impl type var map builtin language)
   configure_file(${__check_type_size_dir}/CheckTypeSize.c.in ${src} @ONLY)
   try_compile(HAVE_${var} ${CMAKE_BINARY_DIR} ${src}
     COMPILE_DEFINITIONS ${CMAKE_REQUIRED_DEFINITIONS}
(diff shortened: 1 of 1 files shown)


kripken:
I see we have `test_cmake_threads` in `test/test_other.py`. That enables shared memory here:

https://github.com/emscripten-core/emscripten/blob/ea35c8cde53d4691504436300493d60b49ff3744/test/cmake/threads/CMakeLists.txt#L5

Is that not enough to show the problem this PR fixes? (If not, can we add to that test, or add a new one?)

m4burns:
I'm guessing nothing in that test calls `check_type_size` - I will try to add a test for this.

Issue #22265, opened 2024-07-19
Argument order passed to the em_dlopen_callback does not match the declaration.


davits:
**Version of emscripten/emsdk:**
`emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.44 (bec42dac7873903d09d713963e34020c22a8bd2d)`

The code below is copied from main branch, so the problem most probably is there as well.

emscripten.h declares `emscripten_dlopen()` function and success callbacks as follows:
```cpp
// Asynchronous version of dlopen.  Since WebAssembly module loading in general
// is asynchronous the normal dlopen function can't be used in all situations.
typedef void (*em_dlopen_callback)(void* handle, void* user_data);
void emscripten_dlopen(const char *filename, int flags, void* user_data, em_dlopen_callback onsuccess, em_arg_callback_func onerror);
```
But in my local tests, user_data is passed as first argument and the handle is second.
Also async dlopen test in emscripten lists success callback arguments in (user_data, handle) order:
https://github.com/emscripten-core/emscripten/blob/a1fe3902bf73a3802eae0357d273d0e37ea79898/test/other/test_dlopen_async.c#L9

kripken:
Thanks @davits , that does look wrong.

It seems to have been wrong from the first landing in #14503, though, so maybe we are reading that wrong..? There are tests in that PR, so it seems odd. @sbc100 what do you think?

PR #22306, opened 2024-07-31
Fix argument names in em_dlopen_callback declaration. NFC

sbc100:
This matches the order in which the current implementation and the current tests expect that argument to be passed.  Ideally we would pass the dso handler first, but that would be breaking change.

Fixes: #22265


kripken (approved):
The current order sort of makes sense, in that both onsuccess and onerror have the user arg first? But onerror has no other arg, so that's only in a trivial sense, and yeah, onsuccess does seem slightly odd.

PR #22308, opened 2024-08-01
Disable tests to let binaryen#6793 roll in

kripken:
https://github.com/WebAssembly/binaryen/pull/6793 alters names of functions, which two metadce
tests turn out to be sensitive to. (It changes the suffixes we use to deduplicate
identical names.)


kripken:
> (You can use @disabled(...) instead of self.skipTest.. not that it really matters)

Ah, thanks, I'll do that next time (to avoid re-running tests here).

Issue #21450, opened 2024-02-28
--embind-emit-tsd fails when linking dynamic libraries


dchengTSC:
`emcc` fails when generating typescript definitions when a dynamic library is being linked.

Here's a zip file that contains some sample code that reproduces the issue. Run `build.sh` at the root of the zip file to get the error.

[ErrorWithDynamicLinkingAndEmbindAndTSDGen.zip](https://github.com/emscripten-core/emscripten/files/14437793/ErrorWithDynamicLinkingAndEmbindAndTSDGen.zip)

I think the problem is that the tsd generation requires the dynamic library(`.so`) file to be in the Emscripten temp directory in order to succeed. I discovered this by setting `EMCC_DEBUG=2`, which preserves the contents of the Emscripten temp directory. Then, I copy the `.so` file over to the temp directory and re-run the node command (eg: `node /var/folders/_6/dll3kgx93qz05hvnymcr5jzh0000gp/T/emscripten_temp_u26tl33o/tsgen_a.out.js`). It properly emits the tsd after I do this.

**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.54 (a95c44ee924d02fa1498f846595485d27c31daa8)
clang version 19.0.0git (https://github.com/llvm/llvm-project e769fb8699e3fa8e40623764f7713bfc783b0330)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /Users/d.cheng/projects/emsdk/upstream/bin
```
**Failing command line in full:**
```
➜  ~/projects/DynamicLinkingAndEmbindWithTSDGeneration git:(main) ✗ emcc main.cpp MainLibrary.cpp SomeLibrary/SomeLibrary.so --bind --embind-emit-tsd=MainLibrary.d.ts -sMAIN_MODULE=2 -o main.html             
/private/var/folders/_6/dll3kgx93qz05hvnymcr5jzh0000gp/T/emscripten_temp_f2ow26jt/tsgen_a.out.js:136
      throw ex;
      ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
[...]

PR #22312, opened 2024-08-01
[embind] Support TypeScript generation when using dynamic linking.

### LINKED ISSUE #21859: --embind-emit-tsd / --emit-tsd do not work with dynamic linkage
In case a WASM binary is linked dynamically the option to emit typescript interface does not work.
The main cause seems to be that the generated wasm and js file are copied to a temp directory during running emitting the typescript interfaces while all of the possible dependencies are not.

As a workaround I changed link.py #https://github.com/emscripten-core/emscripten/blob/63c648fa17be49fa9640d4a40c8ba2f7a9b2d444/tools/link.py#L1963 to the following content:

```
outfile_wasm = in_temp('tsgen_a.out.wasm')
import pathlib, glob, shutil
root_dir =  str(pathlib.Path(wasm_target).parent.resolve())
out_temp = str(pathlib.Path(outfile_js).parent.resolve())
for file in glob.glob(root_dir + r'/*.so*'):
    logger.info('Copying: ' + file)
    shutil.copy(src=file, dst=out_temp)"
```
The example workaround above simply copies all .so files existent in the same location as the wasm/js file to the same location in the temp directory. With the workaround the dynamic lib is properly loaded and the typescript interfaces are emitted. There is most likely a much more sophisticated solution to that problem.

3.1.56

The issue occurs during linking when option --emit-tsd or --embind-emit-tsd is performed

As soon as the generated WASM binary contains the following dynamic link section:
`dylink.0 .so NAMEOFLIB.so` the linking step fails within: function run_embind_gen (https://github.com/emscripten-core/emscripten/blob/63c648fa17be49fa9640d4a40c8ba2f7a9b2d444/tools/link.py#L1924C5-L1924C19) due to node execution failure while trying to load any dynamically linked dependencies in:
`out = shared.run_js_tool(outfile_js, [], node_args, stdout=PIPE)` 

**Possible solution**
During linkage all dependencies are already propagated to emscripten for the actual linking step, ensure these dependencies are not only put into the dynlink.0 section but also properly copied/symlinked from their origin to the temporary folder used for generating tsd utilizing node when using --embind-emit-tsd / --emit-tsd.

brendandahl:
Copy the library files to the temp directory so that the TS generation can find them when running.

Fixes #21859, #21450

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -1872,7 +1872,7 @@ def phase_link(linker_arguments, wasm_target, js_syms):
 
 
 @ToolchainProfiler.profile_block('post link')
-def phase_post_link(options, state, in_wasm, wasm_target, target, js_syms, base_metadata=None):
+def phase_post_link(options, state, in_wasm, wasm_target, target, js_syms, base_metadata=None, linker_inputs=None):
   global final_js
 
   target_basename = unsuffixed_basename(target)
@@ -1892,10 +1892,10 @@ def phase_post_link(options, state, in_wasm, wasm_target, target, js_syms, base_
   metadata = phase_emscript(in_wasm, wasm_target, js_syms, base_metadata)
 
   if settings.EMBIND_AOT:
-    phase_embind_aot(wasm_target, js_syms)
+    phase_embind_aot(wasm_target, js_syms, linker_inputs)
 
   if options.emit_tsd:
-    phase_emit_tsd(options, wasm_target, state.js_target, js_syms, metadata)
+    phase_emit_tsd(options, wasm_target, state.js_target, js_syms, metadata, linker_inputs)
 
   if options.js_transform:
     phase_source_transforms(options)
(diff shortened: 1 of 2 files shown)


brendandahl:
Alternatively, I was thinking we could change `locateFile` to look in multiple places, but this copying seems slightly less hacky.

tools/link.py
```diff
@@ -1924,11 +1924,17 @@ def phase_emscript(in_wasm, wasm_target, js_syms, base_metadata):
...
-def run_embind_gen(wasm_target, js_syms, extra_settings):
+def run_embind_gen(wasm_target, js_syms, extra_settings, linker_inputs):
   # Save settings so they can be restored after TS generation.
   original_settings = settings.backup()
   settings.attrs.update(extra_settings)
 
+  if settings.MAIN_MODULE and linker_inputs:
+    # Copy libraries to the temp directory so they can be used when running
+    # in node.
+    for _i, linker_input in linker_inputs:
```
brendandahl, line 1935: Should we maybe filter by file name here so we don't copy object files? Looking at how the linker inputs work though, a side module can have any file ending.

sbc100: You can use `building.is_wasm_dylib()` I think

tools/link.py
```diff
@@ -1872,7 +1872,7 @@ def phase_link(linker_arguments, wasm_target, js_syms):
 
 
 @ToolchainProfiler.profile_block('post link')
-def phase_post_link(options, state, in_wasm, wasm_target, target, js_syms, base_metadata=None):
+def phase_post_link(options, state, in_wasm, wasm_target, target, js_syms, base_metadata=None, linker_inputs=None):
```
sbc100, line 1875: Is this function ever called without the final argument here?

brendandahl: Yeah, from `run_post_link`

sbc100 (approved):

PR #22315, opened 2024-08-02
Only set CMAKE_EXECUTABLE_SUFFIX in toolchain if not defined already

### LINKED ISSUE #18860: Using Emscripten in CMake doesn't output html
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.25 (febd44b21ecaca86e2cb2a25ef3ed4a0a2076365)
clang version 16.0.0 (https://github.com/llvm/llvm-project effd75bda4b1a9b26e554c1cda3e3b4c72fa0aa8)
Thread model: posix
InstalledDir: D:\sdk\emsdk\upstream\bin

I am trying to add emscripten support to my project, but when I specify `-o output.html` in `target_link_options`, it doesn't output a html file, but a js and a wasm file

Using `set(CMAKE_EXECUTABLE_SUFFIX ".html")` also doesn't work

The CMakeLists.txt snippet
```
set(CMAKE_TOOLCHAIN_FILE $ENV{EMSDK_HOME}/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake)
set(CMAKE_EXECUTABLE_SUFFIX ".html")
set(LINK_OPTIONS ${LINK_OPTIONS} -o output.html -sMIN_WEBGL_VERSION=2 -sMAX_WEBGL_VERSION=2 -sUSE_GLFW=3 -sWASM=1)
set(LINK_LIBRARIES ${LINK_LIBRARIES} -lglfw3 -lGLESv2 -lEGL)
```

starball5:
Fixes #18860


cmake/Modules/Platform/Emscripten.cmake
```diff
@@ -270,7 +270,9 @@ if (EMSCRIPTEN_GENERATE_BITCODE_STATIC_LIBRARIES)
   message(FATAL_ERROR "EMSCRIPTEN_GENERATE_BITCODE_STATIC_LIBRARIES is not compatible with the llvm backend")
 endif()
 
-set(CMAKE_EXECUTABLE_SUFFIX ".js")
+if(NOT DEFINED CMAKE_EXECUTABLE_SUFFIX)
+  set(CMAKE_EXECUTABLE_SUFFIX ".js")
+endif()
```
sbc100, line 275: It looks like we do have tests that set CMAKE_EXECUTABLE_SUFFIX  (e.g. `test/cmake/target_html/CMakeLists.txt`) but maybe they work because they don't set it on the command line.

If that is correct then it would be good to add a command line test.

sbc100: Below we use the pattern: `if ("${CMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES}" STREQUAL "")` .. is there some reason to prefer one over the other?

starball5: > but maybe they work because they don't set it on the command line.

@sbc100 I'm pretty sure those were working because they were set as non-cache variables _after_ the call to `project()` (IIRC, toolchain file is read upon several events, one of which is calling `project()`). I.e. call to `project()`, toolchain is read, toolchain sets `CMAKE_EXEUCTABLE_SUFFIX` as non-cache variable, project's CMakeLists.txt sets `CMAKE_EXEUCTABLE_SUFFIX`, overwriting the value set by the toolchain (I have not verified this, but it's my semi-educated guess). What I have written here with `if(NOT DEFINED CMAKE_EXECUTABLE_SUFFIX)` will do the right thing for the case where the user sets `CMAKE_EXECUTABLE_SUFFIX` as a cache variable (Ex. via `-D` in the configure command).

starball5: > is there some reason to prefer one over the other?

@sbc100 There's a behavioural difference. using if(DEFINED) checks if the variable is _defined_. A variable can be defined as an empty string.

```cmake
set(empty "")
if(DEFINED empty)
    message("var empty is defined")
endif()
unset(empty)
if(DEFINED empty)
    message("I will not print")
endif()
```

sbc100: Presumably we should be consistent withing this file though.  Checking for defined in one case and empty string in another seems inconsistent.

starball5: where there is a semantic difference, insisting upon "consistency" for appearances seems like the wrong mindset to me. I don't know what the blocks with `if ("${CMAKE_<LANG>_IMPLICIT_INCLUDE_DIRECTORIES}" STREQUAL "")` are for, so I can't speak on what should be done there. If the point is to just not interfere if anyone else has defined a value (including defining it to an empty string), then you should use `if(DEFINED)`.

sbc100: OK, we can followup with a conversion of the existing usages of `STREQUAL ""`.  I suspect they should be using `if(DEFINED)` too.

sbc100 (approved):
Change LGTM, but can you add a test for this (i.e. update one of the existing cmake tests so that it will fail without this change).

sbc100:
Looks like the test-other failures are geniune.

aviallon:
Is that still needed, or should this be closed?

sbc100:
@starball5 would you have time to rebase and add a test?  Or should someone else take a look?

sbc100:
I tried to pick up this PR but I noticed that `-DCMAKE_EXECUTABLE_SUFFIX=xx` does not work from the command line ,even after this PR.  It also doesn't work from the command line for native / host projects.

However, `-DCMAKE_EXECUTABLE_SUFFIX_<LANG>=xx` , *does* work.   So I created a PR to test this: https://github.com/emscripten-core/emscripten/pull/26244.

Its unfortunate that `-DCMAKE_EXECUTABLE_SUFFIX` does not work but it seems to be a limitation of cmake and not emscripten.

@bradking maybe can you explain this seemingly behavior?

bradking:
`CMAKE_EXECUTABLE_SUFFIX` was not designed to be set by end users or in the cache.  It's normally set by the platform modules or toolchain files.  If it needs to be changed for some use case, that needs to be modeled with other options such that the platform module knows to set the needed value.

cmake/Modules/Platform/Emscripten.cmake
```diff
@@ -266,7 +266,9 @@ if (EMSCRIPTEN_GENERATE_BITCODE_STATIC_LIBRARIES)
   message(FATAL_ERROR "EMSCRIPTEN_GENERATE_BITCODE_STATIC_LIBRARIES is not compatible with the llvm backend")
 endif()
 
-set(CMAKE_EXECUTABLE_SUFFIX ".js")
+if(NOT DEFINED CMAKE_EXECUTABLE_SUFFIX)
+  set(CMAKE_EXECUTABLE_SUFFIX ".js")
```
bradking, line 270: `if(NOT DEFINED CMAKE_EXECUTABLE_SUFFIX)` may never be true because CMake's `Modules/CMakeGenericSystem.cmake` always sets a default empty value earlier.  Therefore this change may regress the existing default.

sbc100:
Thanks @bradking.  I suppose emscripten is a little strange in that we support different output file suffixes and each one comes with different output behaviour so users often want to set the output suffix themselves.  

I think this PR can be closed than.

I might still try to land #26244.

PR #21379, opened 2024-02-20
Use musl implementation of strftime rather than custom JS code. NFC

sbc100:
This leads to some code size savings in several tests, reduces the differences between standalone mode and non-standalone mode and reduces the amount of custom JS code we need to maitain.

In simple microbenchmark that does nothing but export `strftime` we do code size win at `-O0` but a regression in codesize at `-O2`:

```
$ wc -c old_O0.*
74026 old.js
12124 old.wasm
86150 total
$ wc -c new_O0.*
62120 new.js
22852 new.wasm
84972 total
```

```
$ wc -c old_O2.*
8492 old.js
 194 old.wasm
8686 total
$ wc -c new_O2.*
 5040 new.js
14291 new.wasm
19331 total
```

This regression comes from the fact that `snprintf` is used in the implementation of `strftime` and its hard to optimize away the cost of that function.

In practice I suspect that any application large enough to call `strftime` is likely already directly or indirectly pulling in `snprintf` so I would hope that this will be codesize win in practice since it removes about 3k of JS.

PR #22316, opened 2024-08-02
Fix gmtime etc in STANDALONE_WASM mode

sbc100:
This was broken in #21379.

Replaces: #22276


walkingeyerobot (approved):
Tested and everything looks great! I'll do some further testing next week and let you know if anything pops up, but I don't expect it to. Thanks a ton!

PR #22323, opened 2024-08-06
Fix runMetaWithFS in nodejs environment

### LINKED ISSUE #22322: runMetaWithFS always use XHR and fails in nodejs environment

Thread model: posix
InstalledDir: /home/[myname]/Workspace/github.com/emscripten-core/emsdk/upstream/bin

I was compiling libreoffice to wasm and emscripten + libreoffice already did a great job. but when i used the generated output, i found that it works in browser but not in nodejs. it says XMLHttpRequest is not defined. i tracked down the whole build process and found that libreoffice is using [--separate-metadata ](https://github.com/LibreOffice/core/blob/dd47e4b30cb7dab30588d6c79c651f218165e3c5/static/CustomTarget_emscripten_fs_image.mk#L1520-L1524) and it generated a function called runMetaWithFS 

https://github.com/emscripten-core/emscripten/blob/54c570094fa5bf48151e4a79d78a50a2b3c612c2/tools/file_packager.py#L1113-L1118

i temporarily work around it by modify locally installed emscripten like this:
```
  1113   function runMetaWithFS() {
  1114     Module['addRunDependency']('%(metadata_file)s');
  1115     var REMOTE_METADATA_NAME = Module['locateFile'] ? Module['locateFile']('%(metadata_file)s', '') : '%(metadata_file)s';
  1116     if (                
  1117       typeof process === "object" &&
  1118       typeof process.versions === "object" &&
  1119       typeof process.versions.node === "string"
  1120     ) {                 
  1121       require("fs").readFile(REMOTE_METADATA_NAME, function (err, contents) {
  1122         if (err) {      
  1123           console.log(err)
  1124         } else {        
  1125           loadPackage(JSON.parse(contents.toString()))
  1126         }               
  1127       });               
  1128       return;           
  1129     }  
```

jeffguorg:
Fixes #22322

### DIFF
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -1113,6 +1113,20 @@ def generate_js(data_target, data_files, metadata):
   function runMetaWithFS() {
     Module['addRunDependency']('%(metadata_file)s');
     var REMOTE_METADATA_NAME = Module['locateFile'] ? Module['locateFile']('%(metadata_file)s', '') : '%(metadata_file)s';
+    if (
+      typeof process === "object" &&
+      typeof process.versions === "object" &&
+      typeof process.versions.node === "string"
+    ) {
+      require("fs").readFile(REMOTE_METADATA_NAME, function (err, contents) {
+        if (err) {
+          console.log(err)
+        } else {
+          loadPackage(JSON.parse(contents.toString()))
+        }
+      });
+      return;
+    }
     var xhr = new XMLHttpRequest();
     xhr.onreadystatechange = function() {
      if (xhr.readyState === 4 && xhr.status === 200) {


tools/file_packager.py
```diff
@@ -1113,6 +1113,20 @@ def generate_js(data_target, data_files, metadata):
   function runMetaWithFS() {
     Module['addRunDependency']('%(metadata_file)s');
     var REMOTE_METADATA_NAME = Module['locateFile'] ? Module['locateFile']('%(metadata_file)s', '') : '%(metadata_file)s';
+    if (
```
sbc100, line 1116: Can you make this conditional on `options.support_node`?

tools/file_packager.py
```diff
@@ -1113,6 +1113,20 @@ def generate_js(data_target, data_files, metadata):
...
+      typeof process === "object" &&
+      typeof process.versions === "object" &&
+      typeof process.versions.node === "string"
+    ) {
```
sbc100, line 1120: Can we share the node detection code that already exists for fetch in the package data?

sbc100: How about we create a new `isNode` variable like the existing `isPthread` and `isWasmWorker` above?

Then we can also also great a local `var fs = ..` so we only need to require it once?

PR #22327, opened 2024-08-06
Simplify gen_struct_info. NFC

sbc100:
In the past we had this script setup to run from emcc directly.  Now that it only runs as a standalone tool a lot of the complexity can be removed.

### DIFF
--- a/tools/maint/gen_struct_info.py
+++ b/tools/maint/gen_struct_info.py
@@ -72,7 +72,6 @@ from tools import config
 from tools import shared
 from tools import system_libs
 from tools import utils
-from tools.settings import settings
 
 QUIET = (__name__ != '__main__')
 DEBUG = False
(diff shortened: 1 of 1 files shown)


tools/maint/gen_struct_info.py
```diff
@@ -423,7 +401,9 @@ def main(args):
   extra_cflags = []
 
   if args.wasm64:
-    settings.MEMORY64 = 2
+    # Always use =2 here so that we don't generate binary that actually requires
```
kripken, line 404:
```suggestion
    # Always use =2 here so that we don't generate a binary that actually requires
```
A very minor concern is that I imagine we will remove mode 2 some day, when memory64 is so stable we don't need it?

kripken (approved):

tools/maint/gen_struct_info.py
sbc100: Exactly yes.   For now, we need it.   I'm not changing that, by the way.  This is just the status quo.

PR #22332, opened 2024-08-07
Use exported memory when building a MAIN_MODULE. NFC

sbc100:
Side modules import the memory but the main module should export it.


src/library_dylink.js
```diff
@@ -315,8 +312,6 @@ var LibraryDylink = {
         } else if (typeof value == 'bigint') {
           entry.value = value;
 #endif
-        } else {
-          throw new Error(`bad export type for '${symName}': ${typeof value}`);
```
kripken, line 319: I am guessing these changes are to do with handling a Memory export. But I see only removals. Where is the code that handles it?

sbc100:
Marking as draft since I think this might be more complex that i thought.

I'm think as long as we are buiding the main module as PIC the linker cannot create the memory since it doesn't know how much space it needs to use.

src/library_dylink.js
sbc100: This should should just ignore exports such as memories and tables.  Its only interesting in handling globals and functions.

PR #22337, opened 2024-08-08
Use FS.initialized. NFC

sbc100:
The FS object has an initialized member, which seem like it was intended for this purpose, but we were not using it.


sbc100:
It looks like this discrepancy existed since this code was first written: a9fb60ebd640597ec710217c69aad0e5c87884b3

kripken (approved):
Funny that we were doing this wrong...

PR #22346, opened 2024-08-09
Remove the use of --collapse-inline-tag-whitespace pass

juj:
Remove the use of --collapse-inline-tag-whitespace pass, which results in too eager whitespace minification on minified HTML pages. Fixes https://github.com/emscripten-core/emscripten/issues/22188. See also https://github.com/terser/html-minifier-terser/issues/179.


kripken (approved):

juj:
I think the comment should be enough to make sure that we don't accidentally change the behavior.

PR #22349, opened 2024-08-09
replace copied OpenAL extensions header `alext.h` with our own version

gabeklavans:
addressing https://github.com/emscripten-core/emscripten/pull/21857#issuecomment-2272181506

### DIFF
--- a/system/include/AL/alext.h
+++ b/system/include/AL/alext.h
@@ -1,39 +1,5 @@
-/**
- * OpenAL cross platform audio library
- * Copyright (C) 2008 by authors.
- * This library is free software; you can redistribute it and/or
- *  modify it under the terms of the GNU Library General Public
- *  License as published by the Free Software Foundation; either
- *  version 2 of the License, or (at your option) any later version.
- *
- * This library is distributed in the hope that it will be useful,
- *  but WITHOUT ANY WARRANTY; without even the implied warranty of
- *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
- *  Library General Public License for more details.
- *
- * You should have received a copy of the GNU Library General Public
- *  License along with this library; if not, write to the
- *  Free Software Foundation, Inc.,
- *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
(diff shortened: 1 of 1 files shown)


system/include/AL/alext.h
```diff
@@ -31,7 +10,6 @@
 typedef __int64 int64_t;
 typedef unsigned __int64 uint64_t;
 #else
-/* Fallback if nothing above works */
 #include <inttypes.h>
 #endif
```
sbc100, line 14: Can you get rid of all of this and just do `#include <inttypes.h>`

system/include/AL/alext.h
```diff
@@ -1,39 +1,8 @@
...
-#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
-#include <inttypes.h>
-#elif defined(_WIN32) && defined(__GNUC__)
-#include <stdint.h>
-#elif defined(_WIN32)
-typedef __int64 int64_t;
-typedef unsigned __int64 uint64_t;
-#else
-/* Fallback if nothing above works */
 #include <inttypes.h>
```
sbc100, line 5: Actually do we even need either these includes anymore?   Can we just remove them?

system/include/AL/alext.h
```diff
@@ -42,436 +11,54 @@ typedef unsigned __int64 uint64_t;
 extern "C" {
 #endif
```
sbc100, line 12: Do we need the extern "C" anymore since we are not longer declaring any functions?

gabeklavans: It's probably a good idea to leave that in, in case anyone wants to bring in any extensions functions and they don't remember to add it back. It's found in the other headers, also, so it's uniform.

system/include/AL/alext.h
gabeklavans: True, those can probably go away. It'll be obvious if we ever need them back.

gabeklavans:
Sorry for rushing my pushes and spawning all these builds (I actually am a few minutes from being AFK for a day or two)

sbc100 (approved):

gabeklavans:
Not sure if the failing test relates to this change

PR #22351, opened 2024-08-09
Fix test_async_mainloop WRT to new clang warning regarding "main" decl. NFC

sbc100:
See #22336


dschuff (approved):
do you know why this was extern C in the first place? were we trying to call main from JS?

sbc100:
> do you know why this was extern C in the first place? were we trying to call main from JS?

No idea no.   In fact, I've removed the unneccary EMSCRIPTEN_KEEPALIVE too.

PR #22352, opened 2024-08-09
Fix for pathnames with trailing slash

sbc100:
Such paths should be rejected by open / stat / realpath but we were silently ignoring the tailing slashed.

Fixes: #22335


test/other/test_realpath_2.c
```diff
@@ -20,6 +20,8 @@ int main(int argc, char **argv) {
   testrealpath("testfile.txt");
   testrealpath("Folder/testfile.txt");
   testrealpath("testnonexistentfile.txt");
+  // Filenames with trailing slash should fail.
+  testrealpath("Folder/testfile.txt/");
   // folders
   testrealpath("Folder");
   testrealpath("/Folder");
```
tlively, line 27:
```suggestion
  testrealpath("/Folder");
  testrealpath("Folder/");
```

tlively (approved):

test/other/metadce/test_metadce_files_wasmfs.funcs
```diff
@@ -31,7 +31,9 @@ $__wasm_call_ctors
...
 $dummy
```
kripken, line 37: WasmFS code is only removed in this PR - why is this growing?

PR #22359, opened 2024-08-12
Add closing bracket

MaxHearnden:
Fixes Emscripten when using ASYNCIFY_DEBUG >= 2


sbc100 (commented):
I guess we are missing test for this.   Would we add this to one of the existing asyncify tests?

PR #22362, opened 2024-08-12
Round the number of pages passed to Memory.grow. NFC

sbc100:
Without this the trace line below prints (somewhat confusingly):

```
growMemory: 80740352 (+1048576 bytes / 16.999984741210938 pages)
```


src/library.js
```diff
@@ -190,7 +190,7 @@ addToLibrary({
   // it. Returns 1 on success, 0 on error.
   $growMemory: (size) => {
     var b = wasmMemory.buffer;
-    var pages = (size - b.byteLength + {{{ WASM_PAGE_SIZE - 1 }}}) / {{{ WASM_PAGE_SIZE }}};
+    var pages = ((size - b.byteLength + {{{ WASM_PAGE_SIZE - 1 }}}) / {{{ WASM_PAGE_SIZE }}}) | 0;
```
kripken, line 193: Not a big deal to add two bytes here I guess, but the only noticeable impact is to the debug logging below. How about adding the `|0` inside there?

sbc100: We could do, but I also think its better/cleaner if that value we pass at runtime is actually an integer.   I imagine it could be confusing for folks debugging if they see attempts to grow by a some fractional number of pages.

sbc100: I ended up reusing the existing `alignMemory` function.. and optimizing that function at the same time.  So now anyone using `alignMemory` will see a code size saving, but memory grow itself has a slight code size regression due to reusing this function.

kripken (approved):

sbc100:
This change landed as part of #22666

PR #17369, opened 2022-07-04
Make EXPORTED_RUNTIME_METHODS work for library functions

sbc100:
This helps when moving function from runtime to JS library since it
means that users who were previously using `EXPORTED_RUNTIME_METHODS` to
export such symbols will not be effected by the transition.

Previously it was also necessary to add such function to
`DEFAULT_LIBRARY_FUNCS_TO_INCLUDE` (with a different mangling!).

This is step towards unifying the way exports work.  See #8380

PR #22363, opened 2024-08-12
add PThread to runtime elements

dumbmoron:
Since 37e1e68d516, PThread is no longer included in EXPORTED_RUNTIME_METHODS, since it's not necessary for emscripten.

Currently, exporting it by manually including it in EXPORTED_RUNTIME_METHODS when configuring
yields a warning: `warning: invalid item in EXPORTED_RUNTIME_METHODS: PThread`

However, nothing breaks and PThread can be used as normal. Thus, I think it would make sense to include 
it in runtimeElements to suppress this warning.

### DIFF
--- a/src/modules.mjs
+++ b/src/modules.mjs
@@ -418,17 +418,21 @@ function exportRuntime() {
     'HEAPU64',
   ];
 
-  if (PTHREADS && ALLOW_MEMORY_GROWTH) {
-    runtimeElements.push(
-      'GROWABLE_HEAP_I8',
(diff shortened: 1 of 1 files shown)


sbc100:
`PThread` is a library symbols and these days library symbols can be included in `EXPORTED_RUNTIME_METHODS` without any warning (I think this changed a couple of years ago in #17369)

I don't see any warning when I run:

```
$ emcc -pthread ~/test/hello.c -sEXPORTED_RUNTIME_METHODS=PThread 
```

Do you see a warning when you run this command?

PR #22367, opened 2024-08-13
Cleanup main loop tests and run them under node. NFC

sbc100:


test/test_emscripten_main_loop_setimmediate.c
```diff
@@ -12,7 +12,8 @@
 int numFrames = 0;
 
 void looper() {
-  static double frame0 = emscripten_get_now();
+  static double frame0 = 0;
+  if (!frame0) frame0 = emscripten_get_now();
```
kripken, line 16: Is the old code not possible in C?

test/test_other.py
```diff
@@ -13317,6 +13317,22 @@ def test_emscripten_set_immediate(self):
...
+    self.do_runf('test_emscripten_main_loop.c', emcc_args=args)
+
+  def test_emscripten_main_loop_and_blocker(self):
+    self.do_runf('test_emscripten_main_loop_and_blocker.c')
+
+  def test_emscripten_main_loop_settimeout(self):
+    self.do_runf('test_emscripten_main_loop_settimeout.c')
+
+  def test_emscripten_main_loop_setimmediate(self):
+    self.do_runf('test_emscripten_main_loop_setimmediate.c')
```
kripken, line 13334: Maybe we don't need them all - do you have a sense that these are important to test in node?

kripken (approved):

test/test_other.py
sbc100: Personally I would like to have all tests that can run under node, run under node.   I don't think these will have much time, especially since they are in other rather than code.  Browser-only tests are much hard for me (and I would assume others) to debug.

test/test_emscripten_main_loop_setimmediate.c
sbc100: Exactly.. this a C++-ism

Issue #22373, opened 2024-08-14
Unreachable code warnings from Closure pass when using `-s MIN_WEBGL_VERSION=2`


floooh:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
3.1.64

When replacing the deprecated `-s USE_WEBGL2=1` with `-s MIN_WEBGL_VERSION=2 -s MAX_WEBGL_VERSION=2` I'm getting Closure warnings about unreachable code in a handful of places in the `library_webgl.js` shim, looking for instance like this:

```
/var/folders/dz/g9ydwg8973z9nn5bvffcwf3h0000gn/T/emscripten_temp_bf3wnx3n/clear-sapp-ui.jso4.js:2570:2: WARNING - [JSC_UNREACHABLE_CODE] unreachable code
  2570|   GLctx.bufferSubData(target, offset, HEAPU8.subarray(data, data + size));
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The warnings go away when setting the min-version to 1:

``-s MIN_WEBGL_VERSION=1 -s MAX_WEBGL_VERSION=2`

I think this is caused by 'static ifs' which always resolve to true, for instance here:

https://github.com/emscripten-core/emscripten/blob/cd097fbb69632ada93d9d63e57c0cd63c242c77a/src/library_webgl.js#L1874-L1882

Closure notes that the if-branch is always taken and returns early, which makes the remainder of the function unreachable.

(cc @sbc100)

PR #22375, opened 2024-08-14
Fix closure warnings with MIN_WEBGL_VERSION=2. NFC

sbc100:
Fixes: #22373

### DIFF
--- a/src/library_webgl.js
+++ b/src/library_webgl.js
@@ -1279,15 +1279,12 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
 
 #if GL_EMULATE_GLES_VERSION_STRING_FORMAT
         case 0x1F02 /* GL_VERSION */:
-          var glVersion = GLctx.getParameter(0x1F02 /*GL_VERSION*/);
+          var webGLVersion = GLctx.getParameter(0x1F02 /*GL_VERSION*/);
           // return GLES version string corresponding to the version of the WebGL context
+          var glVersion = `OpenGL ES 2.0 (${webGLVersion})`;
 #if MAX_WEBGL_VERSION >= 2
-          if ({{{ isCurrentContextWebGL2() }}}) glVersion = `OpenGL ES 3.0 (${glVersion})`;
-          else
+          if ({{{ isCurrentContextWebGL2() }}}) glVersion = `OpenGL ES 3.0 (${webGLVersion})`;
 #endif
-          {
-            glVersion = `OpenGL ES 2.0 (${glVersion})`;
-          }
           ret = stringToNewUTF8(glVersion);
           break;
         case 0x8B8C /* GL_SHADING_LANGUAGE_VERSION */:
@@ -1535,7 +1532,9 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
 #endif
     }
 #endif
+#if INCLUDE_WEBGL1_FALLBACK
     GLctx.compressedTexImage2D(target, level, internalFormat, width, height, border, data ? {{{ makeHEAPView('U8', 'data', 'data+imageSize') }}} : null);
+#endif
   },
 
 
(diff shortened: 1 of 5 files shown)


floooh:
Giving this a whirl, one sec...

test/test_other.py
```diff
@@ -9479,6 +9480,7 @@ def test_closure_full_js_library(self, args):
       '--minify=0',
       '-Werror=closure',
       '-sINCLUDE_FULL_LIBRARY',
+      '-sOFFSCREEN_FRAMEBUFFER',
```
floooh, line 9483: Is this change intended?

floooh:
I can confirm that the PR fixes the Closure warnings I'm seeing, and spot checking my sokol-samples (https://floooh.github.io/sokol-html5/) I'm also not seeing any runtime problems.

Only question is about the `-sOFFSCREEN_FRAMEBUFFER` change, I'm just wondering if this is an unintended change (see my comment above).

floooh (approved):

sbc100:
> I can confirm that the PR fixes the Closure warnings I'm seeing, and spot checking my sokol-samples (https://floooh.github.io/sokol-html5/) I'm also not seeing any runtime problems.
> 
> Only question is about the `-sOFFSCREEN_FRAMEBUFFER` change, I'm just wondering if this is an unintended change (see my comment above).

That is just an artifact of what the test is test for here.   The test is checking for an non-minified `.deleteVertexArray` which (when GL=2) is only part of the output in `LEGACY_GL_EMULATION` mode.  Basically this setting is just increasing the amount of code that is generated to include more things.

floooh:
Thanks for the quick fix btw @sbc100 :)

src/library_webgl.js
```diff
@@ -1864,11 +1867,13 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
       return;
     }
 #endif
+#if INCLUDE_WEBGL1_FALLBACK
```
kripken, line 1870: It looks like this is reached even without the condition on `settings.MIN_WEBGL_VERSION`, unlike all those before it, so this if is not actually for the complement condition, IIUC?

sbc100: WEBGL_USE_GARBAGE_FREE_APIS itself is conditional on MIN_WEBGL_VERSION.

The idea is that INCLUDE_WEBGL1_FALLBACK will be set only in cases where `WEBGL_USE_GARBAGE_FREE_APIS` is set and also we know that we have only webgl2 support.  In this case that means that above return on line 1867 is guaranteed to run.

sbc100: The exact same pattern repeats below on line 1886

sbc100:
Looks like we have some code size savings as a result of this too

sbc100:
> Looks like we have some code size savings as a result of this too

Sorry .. not true.

tools/link.py
```diff
@@ -1659,6 +1659,8 @@ def phase_linker_setup(options, state, newargs):
     # For chrome see: https://crbug.com/324992397
     if settings.MIN_CHROME_VERSION != feature_matrix.UNSUPPORTED and settings.MEMORY64 and settings.MAXIMUM_MEMORY > 2 ** 32:
       settings.WEBGL_USE_GARBAGE_FREE_APIS = 0
+    if settings.WEBGL_USE_GARBAGE_FREE_APIS and settings.MIN_WEBGL_VERSION >= 2:
```
kripken, line 1662: Can this check be simplified, given that `WEBGL_USE_GARBAGE_FREE_APIS` is only set when `MIN_WEBGL_VERSION >= 2` anyhow?

kripken:
```suggestion
    if settings.WEBGL_USE_GARBAGE_FREE_APIS:
      assert(settings.MIN_WEBGL_VERSION >= 2)
```

src/library_webgl.js
kripken: I see, thanks, I was confused by the condition in link.py. I wrote a suggestion to simplify it below.

kripken (commented):
lgtm % suggestion

tools/link.py
sbc100: Are you confusing `MAX_WEBGL_VERSION` (which is the setting which `WEBGL_USE_GARBAGE_FREE_APIS` depends on) with `MIN_WEBGL_VERSION` which is used here?

kripken: Ah, yes, that was exactly my confusion, thanks!

PR #22377, opened 2024-08-14
Use `@wraps` on all test decorators

sbc100:
Without this the decorated function doesn't mirror properties of the underlying tests, which was breaking things like `@crossplatform`.

See #18163

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -247,6 +247,7 @@ def requires_native_clang(func):
 def requires_node(func):
   assert callable(func)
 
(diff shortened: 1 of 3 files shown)


kleisauke (approved):
Great! It looks like this was indeed the reason why `test_module_stdin_rawfs` and `test_module_stdout_stderr_rawfs` weren't executed on the Windows CircleCI runner.
https://circleci.com/gh/emscripten-core/emscripten/806908

PR #22378, opened 2024-08-14
Simplify stdin/stdout/stderr handling in library_fs.js. NFC

sbc100:
There will be code size saving for any program that uses the filesystem and also uses a reduced `INCOMING_MODULE_JS_API`.    Sadly we don't have any code size tests that currently show this because we have two broad catagories of code size tests:

1. metadce tests: These don't use `STRICT` or reduced `INCOMING_MODULE_JS_API`
2. minimal runtime codesize tests: These use empty `INCOMING_MODULE_JS_API` but don't use the FS.


sbc100:
There will be code size saving for any program that uses the filesystem and also uses a reduced `INCOMING_MODULE_JS_API`.    Sadly we don't have any code size tests that currently show this because we have two broad catagories of code size tests:

1. metadce tests: These don't use `STRICT` or reduced `INCOMING_MODULE_JS_API`
2. minimal runtime codesize tests: These use empty `INCOMING_MODULE_JS_API` but don't use the FS.

sbc100:
Ping..

kripken (approved):

Issue #22290, opened 2024-07-29
Node V14 with CommonJS


drwwoob:
Hello, I’m using emscripten to compile c++ code into js code. Our environment is node v14, and we are trying to come up with commonjs code.
These are the tags I came up with, written in CMake

```CMAKE
 SUFFIX ".js"
    LINK_FLAGS " \
    --bind                          \
    -s MODULARIZE=1                 \
    -s ASSERTIONS=2                 \
    -s MIN_NODE_VERSION=141703      \
    -o project.js                \
    -03                             \
    -lpthread"                                 
```

which when I call `Module` from the generated `.js` file, will give me this error

``` error
if (typeof __filename != 'undefined') _scriptName ||= __filename;
                                                    ^^^
SyntaxError: Unexpected token '||='
```

I figured this is an ES12 syntax, which node 14 does not fully support. But if I am to use `-s EXPORT_ES6=1` I could not get a CommonJS code.

I found a solution that by changing

``` JS
if (typeof __filename != 'undefined') _scriptName ||= __filename;
```
in the generated `.js` file into 

```JS
if (typeof __filename != 'undefined') _scriptName = _scriptName || __filename;
```

then everything can compile. 

I’m wondering if I have to change this every time generating the c++ project, or if there are other tags that could work this around. 

**Version of emscripten/emsdk:**
3.1.61

brendandahl:
I don't immediately see why it's not happening, but babel transpilation should remove that. You could try enabling debug with the environment variable `EMCC_DEBUG=1` and seeing if babel is running.

drwwoob:
i’ve enabled `EMCC_DEBUG=1`, and here’s part of the outcome
```command
shared:DEBUG: successfully executed /Users/drwwoob/emsdk/node/18.20.3_64bit/bin/node /Users/drwwoob/emsdk/upstream/emscripten/node_modules/.bin/babel /var/folders/1t/bb241rj928n_d97qwc_7pp8h0000gp/T/emscripten_temp/robot_math.js -o /var/folders/1t/bb241rj928n_d97qwc_7pp8h0000gp/T/emscripten_temp/tmpqgzouygqbabel.js --presets @babel/preset-env --config-file /var/folders/1t/bb241rj928n_d97qwc_7pp8h0000gp/T/emscripten_temp/tmpr9kkz1fjbabel_config.json
profiler:DEBUG: block "transpile" took 0.609 seconds
[...]

PR #22385, opened 2024-08-15
Avoid using ES6 syntax outside of core module code

drwwoob:
This makes the output compatible with at least Node v10

Fixes: #22290


kripken:
Is this needed if we go with #22390? That should then transpile this code, I hope?

drwwoob:
> Is this needed if we go with #22390? That should then transpile this code, I hope?

yes, this is just an alternative solution

sbc100:
I'm fine with landing this change now since I think #22390 would be a little harder to land. 

Can you update the PR description to say something like "Avoid using ES6 syntax outside of core module code".

sbc100 (approved):

drwwoob:
> I'm fine with landing this change now since I think #22390 would be a little harder to land.
> 
> Can you update the PR description to say something like "Avoid using ES6 syntax outside of core module code".

sure, I will change do changes on this PR and close #22390.

sbc100:
> > I'm fine with landing this change now since I think #22390 would be a little harder to land.
> > Can you update the PR description to say something like "Avoid using ES6 syntax outside of core module code".
> 
> sure, I will change do changes on this PR and close #22390.

I still think the goal of #22390 is a good one.. its just a hard change to make.  Feel free to close it or leave it open aspirationally.

drwwoob:
> > > I'm fine with landing this change now since I think #22390 would be a little harder to land.
> > > Can you update the PR description to say something like "Avoid using ES6 syntax outside of core module code".
> > 
> > 
> > sure, I will change do changes on this PR and close #22390.
> 
> I still think the goal of #22390 is a good one.. its just a hard change to make. Feel free to close it or leave it open aspirationally.

Hmmm... For solution #22390, do you think it is possible to specify the portion of code that needs to be transpile, so the transile step skips through wasm2js and other needless parts?

sbc100:
> > > > I'm fine with landing this change now since I think #22390 would be a little harder to land.
> > > > Can you update the PR description to say something like "Avoid using ES6 syntax outside of core module code".
> > > 
> > > 
> > > sure, I will change do changes on this PR and close #22390.
> > 
> > 
> > I still think the goal of #22390 is a good one.. its just a hard change to make. Feel free to close it or leave it open aspirationally.
> 
> Hmmm... For solution #22390, do you think it is possible to specify the portion of code that needs to be transpile, so the transile step skips through wasm2js and other needless parts?

Why wouldn't you also want to transpile the wasm2js output?

drwwoob:
> Why wouldn't you also want to transpile the wasm2js output?

so in #22390, it is mentioned that

>It looks like moving this to later will cause it to run after wasm2js. That might not be harmful, but it is potentially a lot more work, to process all the wasm2js output and not do anything on it.

I thought this comment is saying that the wasm2js part does not need to be transpile and for efficiency should not be transpiled.

sbc100:
> > Why wouldn't you also want to transpile the wasm2js output?
> 
> so in #22390, it is mentioned that
> 
> > It looks like moving this to later will cause it to run after wasm2js. That might not be harmful, but it is potentially a lot more work, to process all the wasm2js output and not do anything on it.
> 
> I thought this comment is saying that the wasm2js part does not need to be transpile and for efficiency should not be transpiled.

Fair enough.. although most users won't be transpiling, especially in debug builds, so I personally don't think its a big deal if we slow down the release builds a little.

PR #22390, opened 2024-08-16
calling transpile method after modularising

drwwoob:
from issue [#22290](https://github.com/emscripten-core/emscripten/issues/22290)


kripken (commented):
It looks like moving this to later will cause it to run after `wasm2js`. That might not be harmful, but it is potentially a lot more work, to process all the wasm2js output and not do anything on it.

Can we perhaps reorder the other way, that is, move the modularization forward?

drwwoob:
closed with the solution provided in #22385.

sbc100:
I'm not really sure its worth fixing this, given that #22385 has landed.   Sorry to be flip flopping on this.  Unless you can see of a clear advantage?  

I guess it depend how many modern JS features we might want to use the modularization code?  If its just one conditional assignment maybe its not worth it?

drwwoob:
> I'm not really sure its worth fixing this, given that #22385 has landed. Sorry to be flip flopping on this. Unless you can see of a clear advantage?
> 
> I guess it depend how many modern JS features we might want to use the modularization code? If its just one conditional assignment maybe its not worth it?

So far the only modern JS issue I've encountered is the one that's being fixed in #22385, I would say that the fix is a good-enough solution for me.

Issue #12387, opened 2020-09-30
Clarifying endianness dependency of emscripten


miladfarca:
Hello,

It seems like Big Endian machines are not able to use emscripten, as assertions are added in this commit:
https://github.com/emscripten-core/emscripten/commit/f178541b91624d0f9d02da8fc9884b38684f3d39

I understand that there are problems when it comes to "Typed Array" in Javascript since they are endianness dependent,
however, WASM is little endian enforced. Runtime compilers such as `Google V8` are compatible with running WASM (LE) binary code on BE machines. 

Is the above runtime assertion for endianness only done when generating regular `js` code (such as `asm.js`) or it 
also checks when compiling to WebAssembly? (I'm assuming assertions are only done during compilation to JS or to WASM and emscripten isn't concerned with runtime).

miladfarca:
Hi @kripken, would you have any suggestions on this issue?

kripken:
@miladfarca Correct. Any time there is any JS - and there is by default, as wasm can't access Web APIs itself - then this is an issue.

The only solutions are either

1. Use [standalone mode](https://v8.dev/blog/emscripten-standalone-wasm) (no JS, or you write the JS yourself - not recommended if you use any non-trivial amount of JS APIs from the wasm), or
2. Add support for big endian code in emscripten, using DataView.

Option 2 wouldn't be a large amount of work. Is this important for your project?

miladfarca:
@kripken Thanks for clarifying. It's not a specific project we are trying to support, rather we are trying to support BE architectures in general. Nodejs and V8 support big endian platforms and we are seeing endianness problems with modules similar to `eslint-plugin-jest`. 

In terms of adding support using DataView, is this something we could contribute or emscripten would be able to add the changes and we can test it on BE machines?

brion:
I think some web APIs like WebGL in particular would be more difficult as they'd have to convert data back and forth, knowing which buffers contain what kind of data. However this may be moot, if BE arches are mostly a mainframe/server thing that run node, not browsers.

miladfarca:
@brion correct, at this point we are only concerned with Node.js and server related processes and as you mentioned WebGL should not be an issue.

kripken:
> In terms of adding support using DataView, is this something we could contribute or emscripten would be able to add the changes and we can test it on BE machines?

[...]

PR #22391, opened 2024-08-16
Keep the endianness check when assertions are off for node

bvibber:
Node.js has a big-endian s390 version which occasionally turns up; silent failure of an emscripten module used in a JavaScript library can be hard to debug.

Disabling node output and assertions will keep the check off. If the check sees a big-endian system, it will throw an error during initialization and this will be visible in debug logging, letting the affected downstream users know what's wrong and how they might fix or work around it.

Helps with: https://github.com/emscripten-core/emscripten/issues/12387

### DIFF
--- a/src/runtime_debug.js
+++ b/src/runtime_debug.js
@@ -4,10 +4,10 @@
  * SPDX-License-Identifier: MIT
  */
 
-#if ASSERTIONS
 
+#if (ASSERTIONS || ENVIRONMENT_MAY_BE_NODE) && !SUPPORT_BIG_ENDIAN
 // Endianness check
-#if !SUPPORT_BIG_ENDIAN
+// Note that s390 node *occasionally* pops up and is big-endian.
 (function() {
   var h16 = new Int16Array(1);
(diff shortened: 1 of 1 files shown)


src/runtime_debug.js
```diff
@@ -4,10 +4,10 @@
  * SPDX-License-Identifier: MIT
  */
 
-#if ASSERTIONS
 
+#if (ASSERTIONS || ENVIRONMENT_MAY_BE_NODE) && !SUPPORT_BIG_ENDIAN
 // Endianness check
-#if !SUPPORT_BIG_ENDIAN
+// Note that s390 node *occasionally* pops up and is big-endian.
```
kripken, line 10: How about this?
```suggestion
// Note that s390 node *occasionally* pops up and is big-endian, so we check for
// endianness when node is enabled, even when assertions are off (no browser
// platform is big-endian AFAWCT, so we don't do that on web-only builds, to
// save size there).
```

sbc100 (approved):
lgmt % kripken comment

Issue #22264, opened 2024-07-19
Linking against `noderawfs.js` causes `ioctl_tiocgwinsz` to fail


gudzpoz:
**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.61-git (67fa4c16496b157a7fc3377afd69ee0445e8a6e3)
clang version 19.0.0git (/startdir/llvm-project 7cfffe74eeb68fbb3fb9706ac7071f8caeeb6520)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /opt/emscripten-llvm/bin
```

**Reproducing steps**

1. Source file content:

   ```c
   #include <stdio.h>
   #include <stdlib.h>
   #include <sys/ioctl.h>
   #include <err.h>
   #include <fcntl.h>
   #include <unistd.h>
   #include <termios.h>
   
   void ensure_noderawfs_init() {
       int c;
       FILE *file;
       file = fopen("/etc/hostname", "r");
       if (file) {
           while ((c = getc(file)) != EOF)
               putchar(c);
           fclose(file);
       }
   }
   
   void print_tty_size() {
       struct winsize ws;
       if(ioctl(STDIN_FILENO, TIOCGWINSZ, &ws) < 0) { // <--------------------
           err(EXIT_FAILURE, "/dev/stdin not a tty");
       }
       printf("TTY Size: %d x %d\n", ws.ws_row, ws.ws_col);
   }
   
   int main() {
       ensure_noderawfs_init();
       print_tty_size();
       return 0;
   }
   ```
2. Compile the code above and run it with node:
   ```console
   $ emcc ioctl_test.c -lnodefs.js -lnoderawfs.js
   $ node ./a.out.js                      
   my-host-name      
   .../ledger/build/a.out.js:128
         throw ex;
         ^
   TypeError: Cannot read properties of undefined (reading 'ioctl_tiocgwinsz')
       at ___syscall_ioctl (.../ledger/build/a.out.js:4396:30)
      ...
   ```

**Failing command line in full:**
```console
$ node ./a.out.js
my-host-name
.../ledger/build/a.out.js:128
      throw ex;
      ^

TypeError: Cannot read properties of undefined (reading 'ioctl_tiocgwinsz')
    at ___syscall_ioctl (.../ledger/build/a.out.js:4396:30)
    at wasm://wasm/0001a2fa:wasm-function[48]:0x13c2
[...]

PR #22393, opened 2024-08-17
[NODERAWFS] Revert #18163

kleisauke:
This reverts commit c614fa57ea8048a34341891dbd6f99fb78b6869a.

---

Here's a standalone reproducer:

Before:
```console
$ curl -LO https://github.com/llvm/llvm-test-suite/raw/main/SingleSource/Benchmarks/Misc-C++/Large/ray.cpp
$ em++ -O3 -sEXIT_RUNTIME -sNODERAWFS ray.cpp -o ray.js
$ node ray.js > x.ppm
$ sha256sum x.ppm
acaa1db08ee30dafda03b41c77466b1435fe8ffdd4d88e5a5876e87dff126f58  x.ppm
$ vips copy x.ppm x.png
VipsImage: memory area too small --- should be 262144 bytes, you passed 79485
```

After:
```console
$ em++ -O3 -sEXIT_RUNTIME -sNODERAWFS ray.cpp -o ray.js
$ node ray.js > x.ppm
$ sha256sum x.ppm
cd6fa6fa75b4ba29077f7e4e531b96f17ff0b82f3c70df526c36f748d6a38c5c  x.ppm
$ vips copy x.ppm x.png
```
<details>
  <summary>x.png</summary>

x
</details>


test/other/test_fflush_fs_exit.out
```diff
@@ -1,2 +1,2 @@
 Hey1
-Hey2
+Hey2
```
sbc100, line 2: Is this change needed?

sbc100:
I'm inclined to perhaps revisit #18163.   It seems to me that if one uses `NODERAWFS` then overriding `stdout`, `stderr`, `stdin`, `print`, `printErr` should make just not effect the program stdout/stderr/stdin since the whole point of `NODERAWFS` is to directly map to node file I/O.

test/other/test_fflush_fs_exit.out
kleisauke: This is needed to accommodate `other.test_fflush_fs_exit` under `NODERAWFS` since that can flush streams without newlines.

sbc100: I see that makes sense.

kleisauke:
> I'm inclined to perhaps revisit #18163. It seems to me that if one uses `NODERAWFS` then overriding `stdout`, `stderr`, `stdin`, `print`, `printErr` should make just not effect the program stdout/stderr/stdin since the whole point of `NODERAWFS` is to directly map to node file I/O.

I agree, let's revert #18163. In retrospect, it seems that `library_tty.js` is very tied to the browser, as it skips null terminators and always flushes with a newline.

The latest revision of this PR reverts #18163 in a trivial way.

test/test_other.py
```diff
@@ -9713,7 +9733,7 @@ def test_ioctl(self):
     # ioctl requires filesystem
     self.do_other_test('test_ioctl.c', emcc_args=['-sFORCE_FILESYSTEM'])
 
-  @also_with_noderawfs
+  # @also_with_noderawfs # NODERAWFS doesn't use library_tty.js for stdio streams
```
sbc100, line 9736: It seems that this should probably work under NODERAWFS.   Perhaps open a bug an refer it that bug here?

test/test_other.py
```diff
@@ -1687,18 +1687,38 @@ def test_module_stdin(self):
...
+    assert(stderr.length === 0, 'stderr should be empty. \\n' +
+        'stderr: \\n' + stderr);
+    assert(UTF8ArrayToString(stdout, 0).startsWith('hello, world!'), 'stdout should start with the famous greeting. \\n' +
+        'stdout: \\n' + stdout);
+}
+''')
+    self.emcc_args += ['--pre-js', 'pre.js']
+    self.do_runf('hello_world.c')
+
+  # @also_with_noderawfs # NODERAWFS doesn't use library_tty.js for stdio streams
```
sbc100, line 1710: Just remove this line?

src/library_noderawfs.js
```diff
@@ -55,6 +54,12 @@ addToLibrary({
       var mode = NODEFS.getMode(path);
       return { path, node: { id: st.ino, mode, node_ops: NODERAWFS, path }};
     },
+    createStandardStreams() {
+      FS.createStream({ nfd: 0, position: 0, path: '', flags: 0, tty: false, seekable: false }, 0);
```
kleisauke, line 58: Note that I set `tty: false` here to ensure that issue #22264 remains resolved.
```console
$ emcc ioctl_test.c -lnodefs.js -lnoderawfs.js -sEXIT_RUNTIME
$ node ./a.out.js
pc-kaw
a.out.js: /dev/stdin not a tty: Not a tty
```

sbc100 (approved):

src/library_noderawfs.js
kleisauke: (I also set it for stdout and stderr, btw)

sbc100: Seems reasonable for now, but we should probably add a TODO there, and we should really be setting `tty` based on the underlying node `isatty` call for each descriptor.

kleisauke: Ah, I just noticed that `isatty` for stdio streams will return `0` and set `errno` to `ENOTTY` with this change, which is probably not what we want. I just pushed commit 88a5475f9e68be66d24d77ed974ecc65fbfa6947 instead. Let's reopen #22264 after this change.

test/test_other.py
kleisauke: Changed to: `# @also_with_noderawfs # NODERAWFS reports ENOTTY, see issue #22264.` with commit faac3bc62820b63205890b91ac513b0c035d71b1 instead.

test/test_other.py
kleisauke: Removed with commit faac3bc62820b63205890b91ac513b0c035d71b1.

test/test_other.py
```diff
@@ -1687,18 +1687,38 @@ def test_module_stdin(self):
     self.emcc_args += ['--pre-js', 'pre.js']
     self.do_runf('module/test_stdin.c', 'hello, world!')
 
-  @also_with_noderawfs
+  # @also_with_noderawfs # NODERAWFS doesn't use library_tty.js for stdio streams
```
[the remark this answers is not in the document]
kleisauke: Removed with commit faac3bc62820b63205890b91ac513b0c035d71b1.

kleisauke:
> Can you update the PR title to refer to the change you are reverting?

Sure, updated to `[NODERAWFS] Revert #18163`.

PR #22404, opened 2024-08-19
[webview] prevent potential double resolution in XMLHttpRequest handling

0xlau:
: )
Although the JavaScript engine can handle this situation, attempting to handle both `resolve` and `reject` for the same `Promise` in the code logic indicates that the code structure is not sufficiently rigorous and may have design flaws. This can obscure the intent of the code and potentially confuse other developers.


sbc100 (approved):
Makes me wish we had more testing for ENVIRONMENT_MAY_BE_WEBVIEW.

Are you using this setting (i.e. are you effected by this issue), or did you just see this bug while reading the code?

0xlau:
> Makes me wish we had more testing for ENVIRONMENT_MAY_BE_WEBVIEW.
> 
> Are you using this setting (i.e. are you effected by this issue), or did you just see this bug while reading the code?

Well, Actually I got an error while using k2-fas after packaged in electron environment. And I found that @kripken has already solved this problem. But in order to make the code more rigorous, I submitted this PR.

https://github.com/k2-fsa/sherpa-onnx/issues/1262

PR #22415, opened 2024-08-20
[tsgen] Emit getters/setters for properties when the types are differ…

### LINKED ISSUE #22387: Annotate `EmbindString` as a real string in Typescript
When using Embind, is there a way to annotate an `std::string` as a `string` type when exporting the Typescript types? 

It's a bit unconvenient that `std::string` is exported as:
```ts
type EmbindString = ArrayBuffer|Uint8Array|Uint8ClampedArray|Int8Array|string;
````

I understand the reasoning behind it, but maybe offering a way to annotate a field as being an actual string, could be very helpful. Otherwise, when using the type in Typescript, we're forced to always cast to a real string with `as string`.

I imagine something like this could work:

```cpp
  class_<Person>("Person")
      .property("name", &Person::name, /* TS Type */ "string");
```

Not sure how feasible, maybe there are better ways, but something like that would make things way easier.

brendandahl:
…ent.

Emitting a better type was already done for functions that returned strings, but now works with properties as well.

Fixes #22387

### DIFF
--- a/test/other/embind_tsgen.d.ts
+++ b/test/other/embind_tsgen.d.ts
@@ -20,6 +20,7 @@ type EmbindString = ArrayBuffer|Uint8Array|Uint8ClampedArray|Int8Array|string;
 export interface Test {
   x: number;
   readonly y: number;
+  get stringProperty(): string;set stringProperty(value: EmbindString);
   functionOne(_0: number, _1: number): number;
   functionTwo(_0: number, _1: number): number;
   functionFour(_0: boolean): number;
@@ -102,7 +103,7 @@ export interface DerivedClass extends BaseClass {
 export type ValArr = [ number, number, number ];
 
 interface EmbindModule {
-  Test: {staticFunction(_0: number): number; staticFunctionWithParam(x: number): number; staticProperty: number};
+  Test: {staticFunction(_0: number): number; staticFunctionWithParam(x: number): number; staticProperty: number; get staticStringProperty(): string;set staticStringProperty(value: EmbindString)};
   class_returning_fn(): Test;
   class_unique_ptr_returning_fn(): Test;
   Obj: {};
--- a/test/other/embind_tsgen.d.ts
+++ b/test/other/embind_tsgen.d.ts
@@ -104,7 +104,13 @@ export interface DerivedClass extends BaseClass {
 export type ValArr = [ number, number, number ];
 
 interface EmbindModule {
-  Test: {staticFunction(_0: number): number; staticFunctionWithParam(x: number): number; staticProperty: number; get staticStringProperty(): string; set staticStringProperty(value: EmbindString)};
+  Test: {
+    staticFunction(_0: number): number;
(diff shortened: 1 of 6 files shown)


test/other/embind_tsgen.d.ts
```diff
@@ -20,6 +20,7 @@ type EmbindString = ArrayBuffer|Uint8Array|Uint8ClampedArray|Int8Array|string;
 export interface Test {
   x: number;
   readonly y: number;
+  get stringProperty(): string;set stringProperty(value: EmbindString);
```
sbc100, line 23: missing newline?

test/other/embind_tsgen_ignore_2.d.ts
```diff
@@ -6,6 +6,7 @@ type EmbindString = ArrayBuffer|Uint8Array|Uint8ClampedArray|Int8Array|string;
 export interface Test {
   x: number;
   readonly y: number;
+  get stringProperty(): string;set stringProperty(value: EmbindString);
```
sbc100, line 9: Can you remind me what EmbindString is?

sbc100 (approved):

test/other/embind_tsgen_ignore_2.d.ts
brendandahl: It's alias for all the types that embind can convert to std::strings
https://github.com/emscripten-core/emscripten/blob/7f509d5320528ff31731450d87b9f560dbaf799d/test/other/embind_tsgen.d.ts#L19

sbc100 (approved):

test/other/embind_tsgen_ignore_2.d.ts
```diff
@@ -88,7 +90,7 @@ export interface DerivedClass extends BaseClass {
 export type ValArr = [ number, number, number ];
 
 interface EmbindModule {
-  Test: {staticFunction(_0: number): number; staticFunctionWithParam(x: number): number; staticProperty: number};
+  Test: {staticFunction(_0: number): number; staticFunctionWithParam(x: number): number; staticProperty: number; get staticStringProperty(): string; set staticStringProperty(value: EmbindString)};
```
sbc100, line 93: Oops, these look like they still lack newlines.

brendandahl: I fixed class properties new lines, but not the static properties. I forget which one it was, but this is the way that  TSC auto formats it or maybe it was to match typescripten. It definitely is getting unruly though, so I think I'll break from them and put it on a new line.

sbc100: Oh.. I hadn't seen that this was all inside the same curly braces.   Seems odd that TSC wouldn't use line breaks..

brendandahl: I updated it to add new lines.

sbc100 (approved):

PR #22416, opened 2024-08-20
[embind] Cleanup createJsInvoker. NFC

sbc100:
I couple of minor cleanups:

1. I noticed that argCount was always followed by - 2.
2. I noticed that argsListWired was always being used `fn` as the first argument
3. The `argsLists` were being constructs as comma separated strings, which `join` is really good at.


sbc100:
I guess we don't have any code size tests that include this function?

brendandahl:
> I guess we don't have any code size tests that include this function?

Yeah, it looks like embind_val_hello_world.cpp only tests embind's `val`. We should expand the test or add another test that includes some `EMSCRIPTEN_BINDINGS`.

sbc100:
With the new test from #22418 we can now see the code size improvment.

sbc100:
ptal

brendandahl (approved):

PR #22418, opened 2024-08-20
Add embind code size test. NFC

sbc100:
In order to make this new test even compile I moved the declaration of `dynCalls` from being hand coded in python to being decoratively included in library.js, which makes more sense anyway.

See #22416

### DIFF
--- a/src/library.js
+++ b/src/library.js
@@ -1856,14 +1856,21 @@ addToLibrary({
   },
 
 #if DYNCALLS || !WASM_BIGINT
-#if MAIN_MODULE == 1
-  $dynCallLegacy__deps: ['$createDyncallWrapper'],
+#if MINIMAL_RUNTIME
+  $dynCalls: '[]',
 #endif
+  $dynCallLegacy__deps: [
+#if MAIN_MODULE == 1
+    '$createDyncallWrapper'
+#endif
+#if MINIMAL_RUNTIME
+    '$dynCalls',
+#endif
+  ],
   $dynCallLegacy: (sig, ptr, args) => {
     sig = sig.replace(/p/g, {{{ MEMORY64 ? "'j'" : "'i'" }}})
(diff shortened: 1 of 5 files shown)


tools/emscripten.py
```diff
@@ -61,7 +61,7 @@ def compute_minimal_runtime_initializer_and_exports(post, exports, receiving):
   exports = [asmjs_mangle(x) for x in exports if x != building.WASM_CALL_CTORS]
 
   # Decide whether we should generate the global dynCalls dictionary for the dynCall() function?
-  if settings.DYNCALLS and '$dynCall' in settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE and len([x for x in exports if x.startswith('dynCall_')]) > 0:
+  if settings.DYNCALLS or not settings.WASM_BIGINT or '$dynCall' in settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE:
```
brendandahl, line 64: Is this change suppose to be in here?

sbc100: Yeah, without this the simple test doesn't compile due to closure warning about `dynCalls` symbol being missing.

sbc100: Actually I found a better way to express this dependency precisely in library.js where it belongs.

brendandahl (approved):

src/library.js
```diff
@@ -1856,14 +1856,21 @@ addToLibrary({
   },
 
 #if DYNCALLS || !WASM_BIGINT
+#if MINIMAL_RUNTIME
+  $dynCalls: '[]',
```
kripken, line 1860: This was `{}` before, and it does seem to be used more like a map than an array - why was this changed?

sbc100: Oops you are right... it should be {}

PR #22380, opened 2024-08-14
Make JS exception catching code conditional. NFC

sbc100:

PR #22426, opened 2024-08-21
Revert "Make JS exception catching code conditional. NFC (#22380)"

sbc100:
This reverts commit eb4d36d47f451d0040a0c796606eaae9aa530de3.

The change caused a regression when building some parts of the llvm testsuite.

In particular, when code is linked that was build with -fexceptions, but -fexceptions was not given at link time. 

In addition to the revered change I also had to move `__cxa_get_exception_ptr` to `private_typeinfo.cpp`.  This was converted to native function in #22421, but, when combined with this change/revert it also needs to be available in the `-noexcept` version of libc++abi.   See the comment in `private_typeinfo.cpp`


kripken (approved):
I assume this is because of the code size issue? Please add a link to that discussion.

sbc100:
> I assume this is because of the code size issue? Please add a link to that discussion.

The revert is because it causes a regression when building some parts of the llvm testsuite.

In particular when code is linked that was build with `-fexceptions` but `-fexceptions` was not given at link time.

sbc100:
I updated the PR description and included info on why this change also moves `__cxa_get_exception_ptr`.

dschuff (approved):

PR #22429, opened 2024-08-21
Actually do metadce in at least one variant of test_metadce_cxx. NFC

sbc100:
With `-O2` we were not actually running metadce.


sbc100:
I will rebase this on top of #22428 so that import export names are not lost.

test/test_other.py
```diff
@@ -8656,21 +8675,21 @@ def test_metadce_minimal_pthreads(self):
...
     # eval_ctors 1 can partially optimize, but runs into getenv() for locale
     # code. mode 2 ignores those and fully optimizes out the ctors
-    'ctors1':    (['-O2', '-sEVAL_CTORS'],   [], ['waka']),
-    'ctors2':    (['-O2', '-sEVAL_CTORS=2'], [], ['waka']),
-    'wasmfs':    (['-O2', '-sWASMFS'],       [], ['waka']),
-    'lto':       (['-O2', '-flto'],       [], ['waka']),
+    'ctors1':    (['-Oz', '-sEVAL_CTORS'],   [], ['waka']),
+    'ctors2':    (['-Oz', '-sEVAL_CTORS=2'], [], ['waka']),
+    'wasmfs':    (['-Oz', '-sWASMFS'],       [], ['waka']),
+    'lto':       (['-Oz', '-flto'],       [], ['waka']),
```
kripken, line 8692: I think this might be excessive. While the tests are called "metadce" they are really code size tests. Code size of metadce is important as higher opt levels do it, but there is also value to testing medium levels like `-O2`. Also, `-O2` is likely more common than `-Oz`.

sbc100: OK, I'll maybe only do this for the `-flto` one then?

Also, we should rename this infra to reflect that these are not really just metadce tests.

sbc100:
Limited this change to just one variant.

sbc100:
Rebased on top of main.  Should be good to go now.

kripken (approved):

PR #7672, opened 2018-12-14
full_posix_sockets

juj:
This adds full POSIX sockets API support via a `-s PROXY_POSIX_SOCKETS=1` option. This is implemented by proxying out the sockets API calls to a separate proxy bridge process running on localhost. (that one needs to separately spawn)

This can be useful for special occassions, e.g. when building unit test harnesses or similar, where one just wants to compile existing POSIX sockets code and "have it work" locally. A web browser page can then be simulated to act as a TCP or UDP server, for example.

The feature builds on top of #7670, and can only land after #7670 lands.

Will need splitting out libc to two .bc library parts, one for proxy sockets variant, and another one with the musl stubs.

PR #22432, opened 2024-08-22
getsockopt mismatch in websocket_to_posix_proxy vs. _socket

stbergmann:
It appears that ever since 31309a8c12aaf142ec483dbe3edc0aa195c1bbca "full_posix_sockets (#7672)", the _proxy side sent back a result that included an explicit option_len while the _socket side computes that value from the overall result size.


sbc100:
@juj WDYT?

PR #22435, opened 2024-08-22
Fix link error if `-lhtml5` is used without ASYNCIFY

hoodmane:
`emscripten_wget` requires asyncify but it isn't used by other functions in libhtml5, so we should drop it if asyncify is not in use.

### DIFF
--- a/system/lib/html5/emscripten_wget.c
+++ b/system/lib/html5/emscripten_wget.c
@@ -1,9 +1,12 @@
+#if ASYNCIFY // emscripten_wget requires asyncify
+
 #include <emscripten.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <string.h>
 #include <sys/stat.h>
 #include <unistd.h>
+
 
 // Creates all ancestor directories of a given file, if they do not already
 // exist. Returns 0 on success or 1 on error.
@@ -53,3 +56,5 @@ int emscripten_wget(const char* url, const char* file) {
   free(buffer);
   return fd < 0;
 }
+
+#endif


system/lib/html5/emscripten_wget.c
```diff
@@ -1,10 +1,13 @@
+#if ASYNCIFY // emscripten_wget requires asyncify
+
```
sbc100, line 2: I don't think this macro is ever defined when building this library so I don't think this can work.

sbc100:
Can you share the link error you are seeing?

hoodmane:
```
error: undefined symbol: emscripten_wget_data (referenced by root reference (e.g. compiled C/C++ code))
warning: _emscripten_wget_data may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
```

hoodmane:
Maybe `emscripten_wget_data` could be defined to always trap if `-sASYNCIFY` is not present?

sbc100:
> Maybe `emscripten_wget_data` could be defined to always trap if `-sASYNCIFY` is not present?

I wonder why the object file is getting linked in at all?  Are you perhaps build with `-sMAIN_MODULE=1` (which forces all object files to be included)?

hoodmane:
Yeah that's right.

sbc100:
Ok, I think maybe we should provide a stub JS function in that case.

sbc100:
Oh, it looks like we do already have such stubs: https://github.com/emscripten-core/emscripten/blob/b5750f565383ee1ea7a921a1f907111ab36de36f/src/library_async.js#L623-L637

I wonder why they aren't getting included in your case?  .. maybe because of `STRICT`/`AUTO_JS_LIBRARIES`?

sbc100:
Fix is in https://github.com/emscripten-core/emscripten/pull/22446

Issue #21844, opened 2024-04-27
`-sMAIN_MODULE=2` with `-pthread` misses exports `__emscripten_thread_crashed` and `__embind_initialize_bindings`


8051Enthusiast:
I'm using (a custom build of) Qt 6.7 with emscripten 3.1.58, with both `-pthread` and `-sMAIN_MODULE=2`.
When trying to run the build in a webbrowser, I'm getting an exception because `__emscripten_thread_crashed` is undefined, which is caused by an exception because `__embind_initialize_bindings` is undefined.
Setting `-sEXPORTED_FUNCTIONS=_main,__emscripten_thread_crashed,__embind_initialize_bindings` works around this (but this wasn't needed on 3.1.55 I think).
I'm guessing DCE deletes these because they're only called from javascript?

emcc --version:
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.58 (a41843e0860e52c948c1fce20307933c6631c800)
Copyright (C) 2014 the Emscripten authors (see AUTHORS.txt)
This is free and open source software under the MIT license.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
Linker command:
```
/opt/emsdk/upstream/emscripten/em++ -O3 -DNDEBUG -sMAIN_MODULE=2 -sSTACK_SIZE=5MB -pthread -s PTHREAD_POOL_SIZE=4 -s INITIAL_MEMORY=50MB -s MAXIMUM_MEMORY=4GB -s EXPORTED_RUNTIME_METHODS=UTF16ToString,stringToUTF16,JSEvents,specialHTMLTargets,FS,callMain,ENV -s EXPORT_NAME=yphbt_entry -s MAX_WEBGL_VERSION=2 -s FETCH=1 -s WASM_BIGINT=1 -s STACK_SIZE=5MB -s MODULARIZE=1 -s DISABLE_EXCEPTION_CATCHING=1 -pthread -s ALLOW_MEMORY_GROWTH -sASYNCIFY_IMPORTS=qt_asyncify_suspend_js,qt_asyncify_resume_js -s ERROR_ON_UNDEFINED_SYMBOLS=1 @CMakeFiles/yphbt.dir/objects1 -o yphbt.js @CMakeFiles/yphbt.dir/linkLibs.rsp
```

sbc100:
[...]

PR #22436, opened 2024-08-22
Fix pthreads in `-sSTRICT` mode

sbc100:
`STRICT` has the effect of making `INCOMING_MODULE_JS_API` empty by default, but the pthread runtime relies on overriding `instantiateWasm`.

See #21844

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -490,6 +490,10 @@ def setup_pthreads():
 
   default_setting('DEFAULT_PTHREAD_STACK_SIZE', settings.STACK_SIZE)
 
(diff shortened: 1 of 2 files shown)


tools/link.py
```diff
@@ -490,6 +490,10 @@ def setup_pthreads():
 
   default_setting('DEFAULT_PTHREAD_STACK_SIZE', settings.STACK_SIZE)
 
+  if settings.MINIMAL_RUNTIME and 'instantiateWasm' not in settings.INCOMING_MODULE_JS_API:
+    # pthreads runtime depend on overriding instantiateWasm
```
kripken, line 494:
```suggestion
    # pthreads runtime depends on overriding instantiateWasm
```

PR #22439, opened 2024-08-22
[tsgen] Handle different get/set types for value_objects too.

brendandahl:
Mentioned in #22387

### DIFF
--- a/src/embind/embind_gen.js
+++ b/src/embind/embind_gen.js
@@ -243,6 +243,18 @@ var LibraryEmbind = {
     }
 
   },
+  $printProperty: (prop, nameMap, out) => {
+    const setType = nameMap(prop.type, false);
+    const getType = nameMap(prop.type, true);
+    if (prop.readonly || setType === getType) {
+      out.push(`${prop.readonly ? 'readonly ' : ''}${prop.name}: ${getType}`);
+      return;
+    }
+    // The getter/setter types don't match, so generate each get/set definition.
+    out.push(`get ${prop.name}(): ${getType}`);
+    out.push(`set ${prop.name}(value: ${setType})`);
+  },
+  $ClassProperty__deps: ['$printProperty'],
   $ClassProperty: class {
     constructor(type, name, readonly) {
       this.type = type;
@@ -251,15 +263,7 @@ var LibraryEmbind = {
     }
 
     print(nameMap, out) {
-      const setType = nameMap(this.type, false);
-      const getType = nameMap(this.type, true);
-      if (this.readonly || setType === getType) {
-        out.push(`${this.readonly ? 'readonly ' : ''}${this.name}: ${getType}`);
-        return;
-      }
-      // The getter/setter types don't match, so generate each get/set definition.
-      out.push(`get ${this.name}(): ${getType}`);
-      out.push(`set ${this.name}(value: ${setType})`);
+      printProperty(this, nameMap, out);
     }
   },
   $ConstantDefinition: class {
@@ -325,6 +329,7 @@ var LibraryEmbind = {
       out.push(' ];\n\n');
     }
   },
+  $ValueObjectDefinition__deps: ['$printProperty'],
   $ValueObjectDefinition: class {
     constructor(typeId, name) {
       this.typeId = typeId;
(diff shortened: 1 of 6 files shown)


sbc100 (approved):

sbc100:
I wonder if we should start writing tests that actually use these `.ts` definitions to verify they are correct and working?

brendandahl:
> I wonder if we should start writing tests that actually use these `.ts` definitions to verify they are correct and working?

We do at least [run them through](https://github.com/emscripten-core/emscripten/blob/9a1a67a48668ba70bf58986ed2b2417e43929146/test/test_other.py#L3355-L3356) the `tsc` compiler to ensure they compile.

sbc100:
> > I wonder if we should start writing tests that actually use these `.ts` definitions to verify they are correct and working?
> 
> We do at least [run them through](https://github.com/emscripten-core/emscripten/blob/9a1a67a48668ba70bf58986ed2b2417e43929146/test/test_other.py#L3355-L3356) the `tsc` compiler to ensure they compile.

Right, but it would be pretty cool to actually wrong some typescript to use the module via the generated interface maybe?

sbc100:
> > > I wonder if we should start writing tests that actually use these `.ts` definitions to verify they are correct and working?
> > 
> > 
> > We do at least [run them through](https://github.com/emscripten-core/emscripten/blob/9a1a67a48668ba70bf58986ed2b2417e43929146/test/test_other.py#L3355-L3356) the `tsc` compiler to ensure they compile.
> 
> Right, but it would be pretty cool to actually wrong some typescript to use the module via the generated interface maybe?

Maybe we should even start wriing embind tests in TypeScript instead of JS?

brendandahl:
> > > > I wonder if we should start writing tests that actually use these `.ts` definitions to verify they are correct and working?
> > > 
> > > 
> > > We do at least [run them through](https://github.com/emscripten-core/emscripten/blob/9a1a67a48668ba70bf58986ed2b2417e43929146/test/test_other.py#L3355-L3356) the `tsc` compiler to ensure they compile.
> > 
> > 
> > Right, but it would be pretty cool to actually wrong some typescript to use the module via the generated interface maybe?
> 
> Maybe we should even start wriing embind tests in TypeScript instead of JS?

Yeah, I don't think it would be too hard to write some TS tests. What we also probably want is a way to make sure the TS definitions that we generate match what the JS actually provides at runtime. I'm not aware of a great way to do that, beside maybe compiling all the JS with the TS definitions.

sbc100:
> > > > > I wonder if we should start writing tests that actually use these `.ts` definitions to verify they are correct and working?
> > > > 
> > > > 
> > > > We do at least [run them through](https://github.com/emscripten-core/emscripten/blob/9a1a67a48668ba70bf58986ed2b2417e43929146/test/test_other.py#L3355-L3356) the `tsc` compiler to ensure they compile.
> > > 
> > > 
> > > Right, but it would be pretty cool to actually wrong some typescript to use the module via the generated interface maybe?
> > 
> > 
> > Maybe we should even start wriing embind tests in TypeScript instead of JS?
> 
> Yeah, I don't think it would be too hard to write some TS tests. What we also probably want is a way to make sure the TS definitions that we generate match what the JS actually provides at runtime. I'm not aware of a great way to do that, beside maybe compiling all the JS with the TS definitions.

Won't manually exercising the API from the outside do this?   I kind of like that idea of writing `.ts` tests that run against the ts defintions that the actually run the underlying JS that in turn runs that actual wasm.  So many layers!

brendandahl:
> > > > > > I wonder if we should start writing tests that actually use these `.ts` definitions to verify they are correct and working?
> > > > > 
> > > > > 
> > > > > We do at least [run them through](https://github.com/emscripten-core/emscripten/blob/9a1a67a48668ba70bf58986ed2b2417e43929146/test/test_other.py#L3355-L3356) the `tsc` compiler to ensure they compile.
> > > > 
> > > > 
> > > > Right, but it would be pretty cool to actually wrong some typescript to use the module via the generated interface maybe?
> > > 
> > > 
> > > Maybe we should even start wriing embind tests in TypeScript instead of JS?
> > 
> > 
> > Yeah, I don't think it would be too hard to write some TS tests. What we also probably want is a way to make sure the TS definitions that we generate match what the JS actually provides at runtime. I'm not aware of a great way to do that, beside maybe compiling all the JS with the TS definitions.
> 
> Won't manually exercising the API from the outside do this? I kind of like that idea of writing `.ts` tests that run against the ts defintions that the actually run the underlying JS that in turn runs that actual wasm. So many layers!

Yeah, that should work. I was thinking of trying something a little more automated that would auto check the "shape" of JS objects against their TS definition at runtime.

brendandahl:
Filed #22444

PR #22440, opened 2024-08-22
Mark 3.1.65 as released

sbc100:


kleisauke:
It looks like the 3.1.65 tag is missing from https://github.com/emscripten-core/emscripten/tags.

sbc100:
Ah, sorry, I keep running `git push <tagname>` .. which pushes it my `origin` branch by default instead of `git push upstream 3.1.65`

Hopefully this will all be automated soon thank to the efforts of @dschuff !

Issue #22442, opened 2024-08-23
Deadlock with -sPROXY_TO_PTHREAD and  _emscripten_set_resize_callback_on_thread


stbergmann:
As mentioned in the mail thread at <https://groups.google.com/g/emscripten-discuss/c/Nnr0ENExjoE> "_emscripten_run_callback_on_thread causing deadlocks":  After switching from emsdk 3.1.46 to recent Emscripten trunk (`emcc -v`: `emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.65-git (23197a2d369b8e47e7ea1ba7c498309da835b718)`), my application started to deadlock on start.  The application is recent LibreOffice trunk (see <https://git.libreoffice.org/core/+/refs/heads/master/static/README.wasm.md>), using `-sPROXY_TO_PTHREAD` on top of a slightly modified Qt5 (to make that support `-sPROXY_TO_PTHREAD`).

The browser's main thread is still serving an earlier asynchronous request sent from the application's main pthread via `emscripten_async_run_in_main_runtime_thread`,
```
a_cas_p (soffice.wasm:0x8139063)
futex_wait_main_browser_thread (soffice.wasm:0x8138f94)
emscripten_futex_wait (emscripten_futex_wait.c:130)
__timedwait_cp (__timedwait.c:94)
__pthread_cond_timedwait (pthread_cond_timedwait.c:117)
pthread_cond_wait (pthread_cond_wait.c:5)
emscripten_proxy_sync_with_ctx (proxying.c:385)
emscripten_proxy_sync (proxying.c:405)
_emscripten_run_callback_on_thread (callback.c:39)
(anonymous) (VM9:1236)
focusEventHandlerFunc (VM9:14188)
eventHandler.eventListenerFunc (VM9:11156)
methodCaller<() => void> (VM54:4)
__emval_call_method (VM9:9810)
void emscripten::val::call<void>(char const*) const::'lambda'(emscripten::internal::_EM_METHOD_CALLER*, emscripten::_EM_VAL*, emscripten::internal::_EM_DESTRUCTORS**, void const*)::operator()(emscripten::internal::_EM_METHOD_CALLER*, emscripten::_EM_VAL*, emscripten::internal::_EM_DESTRUCTORS**, void const*) const (val.h:532)
[...]

PR #20370, opened 2023-10-02
Remove use of legacy proxy API from library_html5.js. NFC

sbc100:
AFAICT this is the final usage of the legacy proxy API within emscripten.

PR #22443, opened 2024-08-23
Make _emscripten_run_callback_on_thread asynchronous

stbergmann:
...fixing a regression introduced in c41f6590b1f9fcda12b2adb2d32cd352f163e717 "Remove use of legacy proxy API from library_html5.js. NFC (#20370)".

Fixes: #22442


kripken:
@sbc100 this refers to your PR https://github.com/emscripten-core/emscripten/pull/20370 - has this issue already been resolved?

sbc100:
I'm not totally sure if this is the best solution.   I still need to do more investigation into #22442

sbc100:
Looking at the old code it does look like it was using async dispatch, so this change LGTM.

sbc100:
Its a shame we can't avoid the dynamic allocation here ..

sbc100 (approved):

Issue #19300, opened 2023-05-05
glCompressedTexImage2D error


konstantinegorov:
Hi!
I got the error "Uncaught TypeError TypeError: Failed to execute 'compressedTexImage2D' on 'WebGLRenderingContext': parameter 7 is not of type 'ArrayBufferView'", when I called glCompressedTexImage2D and passed null to the "data" parameter. How to avoid it? Is there any workaround?

konstantinegorov:
Is there any update?

sbc100:
Its hard to say what this might be without more details.   Can you share a full backtrace?   And the line of code at the call site?   If possible a complete program would be even better.

There are several tests in the `test` directory can use `glCompressedTexImage2D`.  It might be useful to take a look at `test/browser/test_anisotropic.c`, for example, to see if there is anything that you are doing differently.

matusfedorko:
We started experiencing this problem too after upgrading from emscripten version 3.1.39 to version 3.1.64. Here is a minimal repro based on the anisotropic test:

```
#include "SDL/SDL.h"
#include "SDL/SDL_image.h"
#include "SDL/SDL_opengl.h"

#include <stdio.h>

int main(int argc, char *argv[])
{
    // Slightly different SDL initialization
    if ( SDL_Init(SDL_INIT_VIDEO) != 0 ) {
        printf("Unable to initialize SDL: %s\n", SDL_GetError());
        return 1;
    }

    SDL_GL_SetAttribute( SDL_GL_DOUBLEBUFFER, 1 );

    SDL_Surface *screen = SDL_SetVideoMode( 600, 600, 16, SDL_OPENGL );
    if ( !screen ) {
        printf("Unable to set video mode: %s\n", SDL_GetError());
        return 1;
    }

    // Clear the screen before drawing
    glClear( GL_COLOR_BUFFER_BIT );

    glViewport( 0, 0, 600, 600 );

    GLuint texture;
    glGenTextures( 1, &texture );
    glBindTexture( GL_TEXTURE_2D, texture );
    glCompressedTexImage2D(GL_TEXTURE_2D, 0, GL_COMPRESSED_RGB_S3TC_DXT1_EXT, 4, 4, 0, 8, NULL);

    SDL_GL_SwapBuffers();

#ifndef __EMSCRIPTEN__
    // Wait for 3 seconds to give us a chance to see the image
    SDL_Delay(2000);
#endif

    // Now we can delete the OpenGL texture and close down SDL
[...]

PR #21445, opened 2024-02-28
Avoid garbage-free WebGL APIs when memory size is over 2gb.

sbc100:
Both chrome and firefox see have some issues with passing 2gb+ and 4gb+ offsets to these APIs.

Once the browser issues are addressed we can lift these restrictions over time.

Fixes: #20533

PR #22453, opened 2024-08-26
Fix calls to glCcompressedTexImage2D with null pixel data

sbc100:
I believe this has always been an issue with emscripten's WebGL1 code which only recently became more prevalent with #21445.

Fixes: #19300

### DIFF
--- a/src/library_webgl.js
+++ b/src/library_webgl.js
@@ -1520,6 +1520,11 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
   },
 
   glCompressedTexImage2D: (target, level, internalFormat, width, height, border, imageSize, data) => {
+    // `data` may be null here, which means "allocate uniniitalized space but
+    // don't upload" in GLES parlance, but `compressedTexImage2D` requires the
+    // final data parameter, so we simply pass a heap view starting at zero
+    // effectively uploading whatever happens to be near address zero.  See
+    // https://github.com/emscripten-core/emscripten/issues/19300.
 #if MAX_WEBGL_VERSION >= 2
     if ({{{ isCurrentContextWebGL2() }}}) {
       if (GLctx.currentPixelUnpackBufferBinding || !imageSize) {
@@ -1533,7 +1538,7 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
     }
 #endif
 #if INCLUDE_WEBGL1_FALLBACK
-    GLctx.compressedTexImage2D(target, level, internalFormat, width, height, border, data ? {{{ makeHEAPView('U8', 'data', 'data+imageSize') }}} : null);
+    GLctx.compressedTexImage2D(target, level, internalFormat, width, height, border, {{{ makeHEAPView('U8', 'data', 'data+imageSize') }}});
 #endif
   },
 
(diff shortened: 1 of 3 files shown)


test/browser/test_anisotropic.c
```diff
@@ -115,7 +115,12 @@ int main(int argc, char *argv[])
       while (level < 5) {
         printf("uploading level %d: %d, %d\n", level, w, h);
         assert(!glGetError());
+#if TEST_TEXSUBIMAGE
+        glCompressedTexImage2D(GL_TEXTURE_2D, level, GL_COMPRESSED_RGBA_S3TC_DXT5_EXT, w, h, 0, w*h, NULL);
```
kainino0x, line 119: > https://community.khronos.org/t/glcompressedteximage2d-and-null-data/41505

IIUC this link is saying you can call `glTexImage2D` with a null pointer to allocate the texture, then <code>gl*Compressed*TexSubImage2D</code> to initialize it.

src/library_webgl.js
```diff
@@ -1520,6 +1520,11 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
   },
 
   glCompressedTexImage2D: (target, level, internalFormat, width, height, border, imageSize, data) => {
+    // `data` may be null here, which means "allocate space but don't upload" in
```
kainino0x, line 1523: My understanding of the conversation on the issue was that if `data` is null, that's literally just telling GL to read from the null pointer, it has no actual meaning. (glTexImage2D special-cases null, but glCompressedTexImage2D does not.) So seems to me like we should have a debug assert that `data` isn't null.

test/browser/test_anisotropic.c
sbc100: According https://github.com/emscripten-core/emscripten/issues/19300, GLES users expect to be able to call `glCompressedTexImage2D` with the null final argument.

So I think the test code here should use what the existing GLES code expects to be able to do.

src/library_webgl.js
sbc100: IIUC, `data` is allowed to be null here (i.e. in the GLES C API), and the (in GLES) that means "just allocate some space".  It certainly doesn't mean "read from address zero". since that would instantly crash, right?

I think its just an oversight in the GLES man page that it doesn't document the null special case.  See https://github.com/emscripten-core/emscripten/issues/19300#issuecomment-2305744279: "If the data argument of CompressedTexImage1D, CompressedTexImage2D, or CompressedTexImage3D is NULL, and the pixel unpack buffer object is zero, a texture image with unspecified image contents is created, just as when a NULL pointer is passed to TexImage1D, TexImage2D, or TexImage3D."

test/browser/test_anisotropic.c
juj: > IIUC this link is saying you can call `glTexImage2D` with a null pointer to allocate the texture, then `gl_Compressed_TexSubImage2D` to initialize it.

I am not sure if that is how the spec is supposed to work. Calling `TexSubImage` will not change a type of the texture that was previously created with a call to `TexImage`?

I tried with the following test code:

```html
<!DOCTYPE html><html><body><script>
var gl = document.createElement("canvas").getContext('webgl');
gl.bindTexture(gl.TEXTURE_2D, gl.createTexture());
var dxt1 = gl.getExtension('WEBGL_compressed_texture_s3tc').COMPRESSED_RGB_S3TC_DXT1_EXT;
var w = 1024;
gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, w, w, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
var foo = new Uint8Array(w*w/2);
gl.compressedTexSubImage2D(gl.TEXTURE_2D, 0, 0, 0, w, w, dxt1, foo);
</script></body></html>
```
which gives a warning in Firefox:

image

that suggests the same.

src/library_webgl.js
juj: In GLES3, data is allowed to be null. In GLES2 the specification missed clarifying whether data is allowed to be null or not. https://github.com/KhronosGroup/WebGL/issues/3686

juj (approved):
Looks great to me!

sbc100:
@kenrussell, can you confirm this is the solution you recommend?

src/library_webgl.js
kainino0x: Ahh, I didn't read far enough in the GLES3 spec because it was buried way at the bottom under the table!

kainino0x: a little clarification:
```suggestion
    // `data` may be null here, which means "allocate uninitialized space but don't upload" in
```

kainino0x (approved):

PR #22459, opened 2024-08-27
Support UTIME_OMIT/UTIME_NOW in futimens

sbc100:
Fixes: #22348

### DIFF
--- a/src/library_syscall.js
+++ b/src/library_syscall.js
@@ -1003,6 +1003,9 @@ var SyscallsLibrary = {
         mtime = (seconds*1000) + (nanoseconds/(1000*1000));
       }
     }
+    // -1 here means UTIME_OMIT was passed.  FS.utime tables the max of these
+    // two values and sets the timestamp to that single value.  If both were
+    // set to UTIME_OMIT then we can skip the call completely.
     if (mtime != -1 || atime != -1) {
       FS.utime(path, atime, mtime);
     }
--- a/src/library_syscall.js
+++ b/src/library_syscall.js
@@ -56,11 +56,11 @@ var SyscallsLibrary = {
       var mtime = stat.mtime.getTime();
       var ctime = stat.ctime.getTime();
       {{{ makeSetValue('buf', C_STRUCTS.stat.st_atim.tv_sec, 'Math.floor(atime / 1000)', 'i64') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.stat.st_atim.tv_nsec, '(atime % 1000) * 1000', SIZE_TYPE) }}};
+      {{{ makeSetValue('buf', C_STRUCTS.stat.st_atim.tv_nsec, '(atime % 1000) * 1000 * 1000', SIZE_TYPE) }}};
       {{{ makeSetValue('buf', C_STRUCTS.stat.st_mtim.tv_sec, 'Math.floor(mtime / 1000)', 'i64') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.stat.st_mtim.tv_nsec, '(mtime % 1000) * 1000', SIZE_TYPE) }}};
+      {{{ makeSetValue('buf', C_STRUCTS.stat.st_mtim.tv_nsec, '(mtime % 1000) * 1000 * 1000', SIZE_TYPE) }}};
       {{{ makeSetValue('buf', C_STRUCTS.stat.st_ctim.tv_sec, 'Math.floor(ctime / 1000)', 'i64') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.stat.st_ctim.tv_nsec, '(ctime % 1000) * 1000', SIZE_TYPE) }}};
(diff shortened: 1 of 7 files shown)


sbc100:
Updated to include wasmfs support

src/library_syscall.js
```diff
@@ -56,11 +56,11 @@ var SyscallsLibrary = {
...
       var ctime = stat.ctime.getTime();
       {{{ makeSetValue('buf', C_STRUCTS.stat.st_atim.tv_sec, 'Math.floor(atime / 1000)', 'i64') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.stat.st_atim.tv_nsec, '(atime % 1000) * 1000', SIZE_TYPE) }}};
+      {{{ makeSetValue('buf', C_STRUCTS.stat.st_atim.tv_nsec, '(atime % 1000) * 1000 * 1000', SIZE_TYPE) }}};
       {{{ makeSetValue('buf', C_STRUCTS.stat.st_mtim.tv_sec, 'Math.floor(mtime / 1000)', 'i64') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.stat.st_mtim.tv_nsec, '(mtime % 1000) * 1000', SIZE_TYPE) }}};
+      {{{ makeSetValue('buf', C_STRUCTS.stat.st_mtim.tv_nsec, '(mtime % 1000) * 1000 * 1000', SIZE_TYPE) }}};
       {{{ makeSetValue('buf', C_STRUCTS.stat.st_ctim.tv_sec, 'Math.floor(ctime / 1000)', 'i64') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.stat.st_ctim.tv_nsec, '(ctime % 1000) * 1000', SIZE_TYPE) }}};
+      {{{ makeSetValue('buf', C_STRUCTS.stat.st_ctim.tv_nsec, '(ctime % 1000) * 1000 * 1000', SIZE_TYPE) }}};
```
kripken, line 63: Are these separate fixes, or related? (they do look correct)

src/library_syscall.js
```diff
@@ -978,19 +978,34 @@ var SyscallsLibrary = {
...
+      atime = now;
+      mtime = now;
     } else {
       var seconds = {{{ makeGetValue('times', C_STRUCTS.timespec.tv_sec, 'i53') }}};
       var nanoseconds = {{{ makeGetValue('times', C_STRUCTS.timespec.tv_nsec, 'i32') }}};
-      atime = (seconds*1000) + (nanoseconds/(1000*1000));
+      if (nanoseconds == {{{ cDefs.UTIME_NOW }}}) {
+        atime = now;
+      } else if (nanoseconds == {{{ cDefs.UTIME_OMIT }}}) {
+        atime = -1;
```
kripken, line 991: Is `-1`'s role obvious here? I gather from the code that `FS.utime` just takes the max of the times, so `-1` being negative is enough to make the other value be chosen, but I had to go read `FS.utime` for that - maybe a comment here would be good?

system/lib/wasmfs/syscalls.cpp
```diff
@@ -1123,6 +1119,16 @@ int __syscall_readlinkat(int dirfd,
   return bytes;
 }
 
+static double timespec_to_ms(timespec ts) {
+  if (ts.tv_nsec == UTIME_OMIT) {
+    return INFINITY;
```
kripken, line 1124: Where is INFINITY defined?

test/utime/test_futimens.c
```diff
@@ -31,6 +32,23 @@ void setup() {
   symlink("file", "folder/file-link");
 }
 
+void check_times(int fd, struct timespec* expected, int tollerance) {
```
kripken, line 35:
```suggestion
void check_times(int fd, struct timespec* expected, int tolerance) {
```

test/utime/test_futimens.c
```diff
@@ -54,13 +72,67 @@ void test() {
...
+  printf("check single UTIME_OMIT...\n");
+  newtimes[0].tv_sec = 41;
+  newtimes[0].tv_nsec = UTIME_OMIT;
+  newtimes[1].tv_sec = 42;
+  newtimes[1].tv_nsec = 88;
+  err = futimens(fd, newtimes);
+  assert(!err);
+
+#if defined(__EMSCRIPTEN__) && !defined(WASMFS)
+  // The trandiation emscripten FS only supports single timestamp so both
```
kripken, line 109:
```suggestion
  // The original emscripten FS (in JS) only supports a single timestamp so both
```

src/library_syscall.js
sbc100: I noticed these values were off by 1000 while working on the test case.  I guess its a separate fix but I think its fine to include it here.

system/lib/wasmfs/syscalls.cpp
sbc100: [<math.h>: ](https://en.cppreference.com/w/c/numeric/math/INFINITY)

kripken: Ah, thanks, what I was confused about is that `tv_nsec` is a `long`, and I missed that we return a different type here, `double`.

kripken (approved):

PR #22463, opened 2024-08-28
Move and rename test_custom_js_settings. NFC

sbc100:
- Move test to `other` suite
- Group with other `js_lib` tests
- Name test file to match test name


kripken (approved):
Nice, this will avoid some unneeded testing in `core`.

Issue #22274, opened 2024-07-23
pthread_kill signal not caught by signal handler


hly2019:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
3.1.54

**Failing command line in full:**
Hi, it seems when I use `pthread_kill` to send signals (`SIGINT`, for example) to threads, the signal I sent can't be caught by the signal handler. It looks like directly terminating the thread. 

For example, in the following code, the compiled Wasm program just terminated and aborted, but I'd expect the `SIGINT` signal could be caught by the signal handler.

I found a related [issue 14872](https://github.com/emscripten-core/emscripten/issues/14872) discussing the `pthread_kill` implementation, saying it's terminating threads instead of sending signals. Is this problem still remaining?

I'd appreciate it so much if you could take a look at it. Thank you very much!

#### Code
```cpp
#include <thread>
#include <signal.h>
#include <stdio.h>
void setupSignalHandlers() {
    struct sigaction act;
    sigemptyset(&act.sa_mask);
    act.sa_flags = SA_SIGINFO;
    act.sa_sigaction = [](int sig, siginfo_t *, void *) {
        printf("signal: %d\n", sig);
    };
    sigaction(SIGINT, &act, nullptr);
}
int main() {
    bool thread_done = false;
    auto thread = std::thread([&]() {
        setupSignalHandlers();
        thread_done = true; // make sure setupSignalHandlers is called
        std::this_thread::sleep_for(std::chrono::milliseconds(100));
    });
    while(!thread_done) { // make sure setupSignalHandlers() is called
        std::this_thread::sleep_for(std::chrono::milliseconds(10));
    }
    pthread_kill(thread.native_handle(), SIGINT); // seems not caught by the signal handler
    thread.join();
    return 0;
}

```
#### Compilation
```shell
# with emcc
$ emcc test_thread.cpp -o main.js -pthread -sNO_DISABLE_EXCEPTION_CATCHING -g; node main.js

# with native
$ g++ test_thread.cpp -o main; ./main
```

#### Results
```shell
# with wasm
Aborted(native code called abort())
/data/main.js:147
      throw ex;
      ^

RuntimeError: Aborted(native code called abort())
    at abort (/data/main.js:741:11)
    at _abort (/data/main.js:2130:7)
    at invoke_v (/data/main.js:2456:29)
    at main.wasm.std::__terminate(void (*)()) (wasm://wasm/main.wasm-008381ea:wasm-function[713]:0xf8ee)
[...]

PR #22467, opened 2024-08-28
Fix pthread_kill behaviour

sbc100:
All `pthread_kill` really does is ensure that the signal handler for a given signal occurs on a specific thread.

The slight exception to this is `SIGCANCEL` which is how `pthread_cancel` is implemented internally.

This allows us to completely remove the `cancelThread` and `killThread` JS functions.

Fixes: #22274


test/pthread/test_pthread_kill.c
```diff
@@ -13,52 +13,58 @@
...
-void BusySleep(long msecs) {
+
+void sleepms(long msecs) {
   usleep(msecs * 1000);
 }
 
+void *thread_start(void *arg) {
+  // As long as this thread is running, keep the shared variable latched to nonzero value.
+  while (!gotTermSignal) {
+    sleepms(1);
```
kripken, line 43: Why was this sleep added to the test?

system/lib/pthread/pthread_kill.c
```diff
@@ -22,5 +39,9 @@ int pthread_kill(pthread_t t, int sig) {
     return ESRCH;
   }
   if (sig == 0) return 0; // signal == 0 is a no-op.
-  return __pthread_kill_js(t, sig);
+
+  // The job of pthread_kill is basically the run the (process-wide) signal
```
kripken, line 43:
```suggestion
  // The job of pthread_kill is basically to run the (process-wide) signal
```

system/lib/pthread/pthread_kill.c
```diff
@@ -5,11 +5,28 @@
...
+#include <emscripten/console.h>
+#include <emscripten_internal.h>
 
 #include "pthread_impl.h"
 #include "lock.h"
 
+void do_raise(void* arg) {
+  int sig = (intptr_t)arg;
+  if (sig == SIGCANCEL) {
+    // For SIGCANCEL is simply enough to run the proxied function since we call
```
kripken, line 19:
```suggestion
    // For SIGCANCEL it is simply enough to run the proxied function since we call
```

system/lib/pthread/pthread_kill.c
```diff
@@ -5,11 +5,28 @@
...
+    // pthread_testcancel at the end of _emscripten_check_mailbox.  We
+    // can't/don't call pthread_testcancel or pthread_exit here because the
+    // proxying system itself is on the stack and we want to exit in clean
```
kripken, line 22:
```suggestion
    // proxying system itself is on the stack and we want to exit with clean
```

system/lib/pthread/pthread_kill.c
```diff
@@ -5,11 +5,28 @@
...
+    // state. (e.g. without holding any locks).
+    _emscripten_runtime_keepalive_clear();
```
kripken, line 24: Does "run the proxied function" refer to `_emscripten_runtime_keepalive_clear`? In `action_terminate()` in `raise.c` I see we also do `_Exit(128 + sig);`

sbc100: `SIGCANCEL` doesn't want to terminate the whole process, only the current thread.    Its a special internal signal that effects just the current thread.

system/lib/pthread/pthread_kill.c
sbc100: "the proxied function" refers to the signal handler setup by sigaction that raise will call.

system/lib/pthread/pthread_kill.c
sbc100: I updated this comment.

system/lib/pthread/pthread_kill.c
sbc100: I updated this comment.

test/pthread/test_pthread_kill.c
sbc100: In order to cancel a running thread the threads needs to reach some cancellation point, in this case the cancellation point is the mailbox check where this thread runs the signal handler code.

kripken: How did it work before? If it failed before, does that mean users might notice this change?

sbc100: Ah yes, it is possible that this change is observable.  Previously we used `worker.terminate()` to cancel the running thread, but now we require the thread to cooperate somewhat.  It would take a very uncoperative thread indeed to observe this though.  Basically a busy loop like this.

If you prefer I can seperate this change in to just the `pthread_kill` part, and keep the `pthread_cancel` behaviour the same (i.e. keep using `worker.terminate()`).

sbc100: Having said that this test is simply wrong since it was using `pthread_kill` to try to kill a thread.. since is not what that API does.   The public `pthread_kill` API will always being down the entire process.  So this test simply has the wrong expectation of how the API works.

sbc100: Actually I'm wrong.. pthread_kill was implement via `worker.terminate()` but this was always wrong.  `pthread_cancel` has always requires cooperation from the target there.

Thus the only folks who could observe this change are ones (like this test) that were wrongly assuming that pthread_kill could be used to terminate single thread.

Thus, I think this change is good as is.

kripken: Change sounds good, then. But maybe add a changelog mention?

kripken (approved):
Overall this lgtm but I am not deeply familiar with the pthreads API or this implementation code, so it might be good to wait for another review here.

sbc100:
OK added a changlog entry.

The only part of really worried about is putting the pthread_testcancel in at the end of mailbox check.   This seem correct to me and @tlively can confirm... its not safe to unwind/exit in the middle of processing and event?

Otherwise I'm pretty confident that this change in the right direction and will land before EOD.

system/lib/pthread/pthread_kill.c
```diff
@@ -5,11 +5,35 @@
...
+#include <emscripten_internal.h>
 
 #include "pthread_impl.h"
 #include "lock.h"
 
+void do_raise(void* arg) {
+  int sig = (intptr_t)arg;
+  if (sig == SIGCANCEL) {
+    // For `SIGCANCEL` there is no need to actually call raise to run the
+    // handler function.  The calling then (the one calling `pthread_cancel`)
```
tlively, line 20: `The calling then` => `The calling thread`

system/lib/pthread/pthread_kill.c
```diff
@@ -5,11 +5,35 @@
...
+    // will already have marked us as being cancelled.  All we need to do is
+    // ensure that `pthread_testcancel` is eventually called and that will cause
+    // this thread to exit.  We can't call `pthread_testcancel` here (since we
+    // are being called from the proxy queue process and we don't want to leave
+    // that in a bad state by unwinding).  Instead, we rely on
+    // `pthread_testcancel` at the end of `_emscripten_check_mailbox`.  Before
+    // we return though we do want to make sure we clear the keepalive state so
```
tlively, line 27: `we return though` => `we return, though`

test/pthread/test_pthread_kill.c
```diff
@@ -13,52 +13,58 @@
...
-  assert(s == 0);
-
-  // Wait until we see the shared variable stop incrementing. (This is a bit heuristic and hacky)
-  for(;;) {
-    int val = sharedVar;
-    BusySleep(100);
-    int val2 = sharedVar;
-    if (val == val2) break;
+  while (sharedVar == 0) {
+    sleepms(10);
```
tlively, line 59: Can we use condition variables instead of sleeps to make the test run as fast as possible?

tlively (commented):
Looks good overall 👍

tlively:
> The only part of really worried about is putting the pthread_testcancel in at the end of mailbox check. This seem correct to me and @tlively can confirm... its not safe to unwind/exit in the middle of processing and event?

Right. I think unwinding in the middle of processing a mailbox item would be asking for trouble, although I don't have enough of the system paged into my memory to say exactly what would go wrong.

test/pthread/test_pthread_kill.c
sbc100: Done: https://github.com/emscripten-core/emscripten/pull/22479

juj:
pthread_cancel() should definitely follow well-orderedness of execution. It'll prove in the wild whether it's correct.

Before this change, the pthread code didn't really support signal API at all. Now that it does, it might be worth looking into running some of the openposixtestsuite signal API tests.

PR #22472, opened 2024-08-29
Fix Closure breaking AUDIO_WORKLET

cwoffenden:
If building with `-sAUDIO_WORKLET` and `--closure=1` the generated glue JS will minimise the property names passed to the worklet's message handler:
https://github.com/emscripten-core/emscripten/blob/9445d6a1c3e3872d6e909565301a47c9c4fbd35b/src/audio_worklet.js#L179
E.g. these are received as:
```
A: 3,
B: 1,
F: 0,
s: "noise-generator"
```
No sound will be generated as a result, since calls to `emscripten_create_wasm_audio_worklet_processor_async()` will never call the passed function.

The quick fix being to add the names to the literal notation for the message (as per the commit). It might be worthwhile renaming the property names to something shorter though whilst making the changes.

Compiling any of the [webaudio](//github.com/emscripten-core/emscripten/tree/5402fc921374cf11f19600e5c3f23eeae76fba5c/test/webaudio) tests with `--closure` will reproduce the error:
```
emcc -O2 --closure=1 -sAUDIO_WORKLET=1 -sWASM_WORKERS=1 -o index.html tone_generator.c
```
Links here for precompiled [before](//wip.numfum.com/cw/2024-08-29/before.html) and [after](//wip.numfum.com/cw/2024-08-29/after.html). On the _before_ test the 'Toggle playback' button doesn't get added because [AudioWorkletProcessorCreated](//github.com/emscripten-core/emscripten/blob/5402fc921374cf11f19600e5c3f23eeae76fba5c/test/webaudio/tone_generator.c#L65) never gets called.

### DIFF
--- a/src/audio_worklet.js
+++ b/src/audio_worklet.js
@@ -168,15 +168,19 @@ class BootstrapMessages extends AudioWorkletProcessor {
         }
 #endif
         // Register a real AudioWorkletProcessor that will actually do audio processing.
-        registerProcessor(d['_wpn'], createWasmAudioWorkletProcessor(d['audioParams']));
+        // 'ap' being the audio params
+        registerProcessor(d['_wpn'], createWasmAudioWorkletProcessor(d['ap']));
 #if WEBAUDIO_DEBUG
-        console.log(`Registered a new WasmAudioWorkletProcessor "${d['_wpn']}" with AudioParams: ${d['audioParams']}`);
+        console.log(`Registered a new WasmAudioWorkletProcessor "${d['_wpn']}" with AudioParams: ${d['ap']}`);
 #endif
         // Post a Wasm Call message back telling that we have now registered the
         // AudioWorkletProcessor class, and should trigger the user onSuccess
         // callback of the
         // emscripten_create_wasm_audio_worklet_processor_async() call.
-        p.postMessage({'_wsc': d['callback'], 'x': [d['contextHandle'], 1/*EM_TRUE*/, d['userData']] }); // "WaSm Call"
+        // 'cb' is the callback function
+        // 'ch' the context handle
+        // 'ud' the passed user data
+        p.postMessage({'_wsc': d['cb'], 'x': [d['ch'], 1/*EM_TRUE*/, d['ud']] }); // "WaSm Call"
       } else if (d['_wsc']) {
         // '_wsc' is short for 'wasm call', using an identifier that will never
         // conflict with user messages
--- a/src/audio_worklet.js
+++ b/src/audio_worklet.js
@@ -174,16 +174,16 @@ class BootstrapMessages extends AudioWorkletProcessor {
         console.log(`Registered a new WasmAudioWorkletProcessor "${d['_wpn']}" with AudioParams: ${d['ap']}`);
 #endif
         // Post a Wasm Call message back telling that we have now registered the
-        // AudioWorkletProcessor class, and should trigger the user onSuccess
-        // callback of the
-        // emscripten_create_wasm_audio_worklet_processor_async() call.
-        // 'cb' is the callback function
-        // 'ch' the context handle
-        // 'ud' the passed user data
-        p.postMessage({'_wsc': d['cb'], 'x': [d['ch'], 1/*EM_TRUE*/, d['ud']] }); // "WaSm Call"
-      } else if (d['_wsc']) {
+        // AudioWorkletProcessor, and should trigger the user onSuccess callback
+        // of the emscripten_create_wasm_audio_worklet_processor_async() call.
+        //
         // '_wsc' is short for 'wasm call', using an identifier that will never
         // conflict with user messages
+        // 'cb' the callback function
+        // 'ch' the context handle
+        // 'ud' the passed user data
+        p.postMessage({'_wsc': d['cb'], 'x': [d['ch'], 1/*EM_TRUE*/, d['ud']] });
+      } else if (d['_wsc']) {
         Module['wasmTable'].get(d['_wsc'])(...d['x']);
       };
     }
--- a/src/library_webaudio.js
+++ b/src/library_webaudio.js
@@ -249,14 +249,15 @@ let LibraryWebAudio = {
 #endif
 
     EmAudio[contextHandle].audioWorklet.bootstrapMessage.port.postMessage({
-      // '_wpn' == 'Worklet Processor Name', use a deliberately mangled name so
-      // that this field won't accidentally be mixed with user submitted
-      // messages.
+      // Deliberately mangled and short names used here ('_wpn', the 'Worklet
+      // Processor Name', to not get accidentally mixed with user submitted
+      // messages, the remainder for space saving reasons, abbreviated from
+      // their variable names).
       '_wpn': UTF8ToString(HEAPU32[options]),
-      'audioParams': audioParams,
-      'contextHandle': contextHandle,
-      'callback': callback,
-      'userData': userData
+      'ap': audioParams,
+      'ch': contextHandle,
+      'cb': callback,
+      'ud': userData
     });
   },
 
--- a/src/library_webaudio.js
+++ b/src/library_webaudio.js
@@ -253,9 +253,9 @@ let LibraryWebAudio = {
 
     EmAudio[contextHandle].audioWorklet.bootstrapMessage.port.postMessage({
       // Deliberately mangled and short names used here ('_wpn', the 'Worklet
-      // Processor Name', to not get accidentally mixed with user submitted
-      // messages, the remainder for space saving reasons, abbreviated from
-      // their variable names).
+      // Processor Name' used as a 'key' to verify the message type so as to
+      // not get accidentally mixed with user submitted messages, the remainder
+      // for space saving reasons, abbreviated from their variable names).
       '_wpn': UTF8ToString(HEAPU32[options]),
       'ap': audioParams,
       'ch': contextHandle,
--- a/src/library_webaudio.js
+++ b/src/library_webaudio.js
@@ -252,11 +252,11 @@ let LibraryWebAudio = {
       // '_wpn' == 'Worklet Processor Name', use a deliberately mangled name so
       // that this field won't accidentally be mixed with user submitted
       // messages.
-      _wpn: UTF8ToString(HEAPU32[options]),
-      audioParams,
-      contextHandle,
-      callback,
-      userData
+      '_wpn': UTF8ToString(HEAPU32[options]),
+      'audioParams': audioParams,
+      'contextHandle': contextHandle,
+      'callback': callback,
+      'userData': userData
     });
   },
 
--- a/src/library_webaudio.js
+++ b/src/library_webaudio.js
@@ -122,9 +122,12 @@ let LibraryWebAudio = {
   },
 
 #if AUDIO_WORKLET
+  // emscripten_start_wasm_audio_worklet_thread_async() doesn't use stackAlloc,
+  // etc., but the created worklet does.
   emscripten_start_wasm_audio_worklet_thread_async__deps: [
     '$_wasmWorkersID',
-    '$_EmAudioDispatchProcessorCallback'],
+    '$_EmAudioDispatchProcessorCallback',
+    '$stackAlloc', '$stackRestore', '$stackSave'],
   emscripten_start_wasm_audio_worklet_thread_async: (contextHandle, stackLowestAddress, stackSize, callback, userData) => {
 
 #if ASSERTIONS
--- a/src/runtime_shared.js
+++ b/src/runtime_shared.js
@@ -13,7 +13,7 @@
     if (!shouldExport) {
       if (MODULARIZE && EXPORT_ALL) {
         shouldExport = true;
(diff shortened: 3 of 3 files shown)


juj:
Looks great. Something that seems odd is that the browser test suite does have several AudioWorklet tests that do enable Closure, e.g. https://github.com/emscripten-core/emscripten/blob/a19f7df00c5404437412fb037ddded67b03d13f8/test/test_browser.py#L5313-L5350

Any chance you'd be able to double check why those tests didn't catch this issue? And/or add/modify the tests there so that they would start failing before this fix?

The fix itself looks solid.

juj (approved):

cwoffenden:
> Any chance you'd be able to double check why those tests didn't catch this issue?

I'll take a look tomorrow (at a guess it silently fails, which is what the basic tests do).

cwoffenden:
@juj I took a look at the tests you linked to and some of the individual parameterised tests fail to initialise the audio, e.g.:
```
test/runner  browser.test_audio_worklet_closure
test/runner  browser.test_audio_worklet_pthreads_and_closure
test/runner  browser.test_audio_worklet_minimal_runtime_pthreads_and_closure
```
And other tests fail for other reasons, e.g., missing `stackAlloc`:
```
test/runner browser.test_audio_worklet_minimal_runtime
```
But running them with the test runner none seem to flag as failures. Though if I look at a complete run output I see that the worklet doesn't appear to be loaded for any of the tests:
```
  test_audio_worklet_closure (test_browser.browser) ... cache:INFO: generating system asset: symbol_lists/ed0c6582c55ebab55869b56f918759080dad520a.json... (this will be cached in "/root/cache/symbol_lists/ed0c6582c55ebab55869b56f918759080dad520a.json" for subsequent builds)
cache:INFO:  - ok
127.0.0.1 - - [29/Aug/2024 15:40:47] code 404, message File not found: /root/project/test.aw.js
```
In the end I ran each test individually, passed `--save-dir` and used `em_run` to look at each.

sbc100:
> In the end I ran each test individually, passed `--save-dir` to the test runner and used `emrun` to look at each.

FYI, when you run just a single test, the output is always saved in `out/test`, so you don't need `--save-dir` in this case.

sbc100:
> @juj I took a look at the tests you linked to and some of the individual parameterised tests fail to initialise the audio _without_ this fix, e.g.:

This seems bad, I guess we need to improve the robustness of these test.  Its hard I suppose because I'm not sure that we can actually run any actual, audio tests in CI (we run the browser tests with `EMTEST_LACKS_SOUND_HARDWARE` in CI).

src/library_webaudio.js
```diff
@@ -252,11 +252,11 @@ let LibraryWebAudio = {
...
-      _wpn: UTF8ToString(HEAPU32[options]),
-      audioParams,
-      contextHandle,
-      callback,
-      userData
+      '_wpn': UTF8ToString(HEAPU32[options]),
+      'audioParams': audioParams,
+      'contextHandle': contextHandle,
+      'callback': callback,
+      'userData': userData
```
sbc100, line 259: Is there some way we can annotate this object such that closure won't mangle these fields?

sbc100: I looked into using `/** dict */` but that didn't seem to work.  Sadly, I can't see any other way to try.

juj:
> some of the individual parameterised tests fail to initialise

> none seems to flag as a failure

Yikes, something has definitely gone wrong then. I'll try to look at this tomorrow on how to make this more robust. Was this Closure change enough to get those tests to pass for you?

cwoffenden:
> Was this Closure change enough to get those tests to pass for you?

The minimal runtime ones also fail, I’ll try to look at that tomorrow and roll it into this PR.

src/library_webaudio.js
cwoffenden: I was looking into the annotations but didn’t see anything either that would carry the naming across files.

sbc100: Can we add a comment here at least.  Something like:

```
Sadly we cannot use the shorthand object notiation here since closure would then minify the field names.
```

juj:
I see there are multiple regressions to Audio Worklet usage from different PRs in the past months. I've got them fixed now in local tree, with interactive tests passing... I'll see what a good way to land these fixes will be.

src/library_webaudio.js
cwoffenden: I shortened the names and commented, which allows the generated minified code to look like:
```
_wpn: X(G[b]),
ap: f,
ch: a,
cb: e,
ud: c
```

src/library_webaudio.js
```diff
@@ -249,14 +249,15 @@ let LibraryWebAudio = {
...
-      userData
+      // Deliberately mangled and short names used here ('_wpn', the 'Worklet
+      // Processor Name', to not get accidentally mixed with user submitted
+      // messages, the remainder for space saving reasons, abbreviated from
+      // their variable names).
+      '_wpn': UTF8ToString(HEAPU32[options]),
+      'ap': audioParams,
+      'ch': contextHandle,
+      'cb': callback,
+      'ud': userData
```
sbc100, line 260: Makes me a bit sad that we have to do this in the source code rather than have the minifier tool do it... but I guess we don't have a better way.

Do we have any codesize test for audioworklets that can means that code size improvement here?  If not maybe we should add one.


I wonder if we should do this in the pthreads code too?  @juj what do you think of this technique?

cwoffenden:
@juj Fixes also added for `MINIMAL_RUNTIME` (dependencies plus missing float heap view).

Tested manually but this works with combinations of Closure and minimal runtime.

src/audio_worklet.js
```diff
@@ -168,15 +168,19 @@ class BootstrapMessages extends AudioWorkletProcessor {
...
 #endif
         // Post a Wasm Call message back telling that we have now registered the
         // AudioWorkletProcessor class, and should trigger the user onSuccess
         // callback of the
         // emscripten_create_wasm_audio_worklet_processor_async() call.
-        p.postMessage({'_wsc': d['callback'], 'x': [d['contextHandle'], 1/*EM_TRUE*/, d['userData']] }); // "WaSm Call"
+        // 'cb' is the callback function
+        // 'ch' the context handle
+        // 'ud' the passed user data
+        p.postMessage({'_wsc': d['cb'], 'x': [d['ch'], 1/*EM_TRUE*/, d['ud']] }); // "WaSm Call"
```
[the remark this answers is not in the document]
cwoffenden: Isn't `_wpn` part of the message name so could clash? Same as `_wpn`, the worklet processor name.

There was already an existing `x` in there so I didn't add underscores.

I did tidy the comments though because they repeat the same thing.

src/audio_worklet.js
```diff
@@ -168,18 +168,22 @@ class BootstrapMessages extends AudioWorkletProcessor {
         }
 #endif
         // Register a real AudioWorkletProcessor that will actually do audio processing.
-        registerProcessor(d['_wpn'], createWasmAudioWorkletProcessor(d['audioParams']));
+        // 'ap' being the audio params
```
sbc100, line 171: Add comment about `_wpn` here too.

sbc100: Oh I see its already documented above.

src/library_webaudio.js
juj: This looks fine to me.

In general end users should never interact/snoop these postMessage()s, so their format should be strictly internal to the implementation. The only thing here is to make sure we generate messages that are not easy to confuse with user-sent messages - which I think manually mangled names make pretty clear it won't be anything that a user might have written.

sbc100: So, IIUC, there is no difference between these mangled names and the existing `_wsc` and `_wpn` mangled names.   Can we be consistent and use underscores either for all of them, or none?

juj: One difference was that the `_wsc` and `_wpn` fields were the "keys" of the data object, used to identify what the postMessaged struct was. These other fields are the satellite data, that are assumed to be present when that one key field matches. That is why I initially only mangled and underscored that one field to make it clear that "that's the key, others are satellite" disposition.

src/audio_worklet.js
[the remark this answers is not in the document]
juj: > @juj do you remember why you add the underscore?

Commented in https://github.com/emscripten-core/emscripten/pull/22472#discussion_r1752319062

src/library_webaudio.js
sbc100: Fair enough, lgtm

sbc100: Actually I'm still a little confused.  It looks like `_wsc` and `_wpn` are both fields the same object that `ap` and `ch` are part of.   

For example I see ` registerProcessor(d['_wpn'], createWasmAudioWorkletProcessor(d['ap']));` where they are both fields of `d`.

cwoffenden: I think the importance of `_wpn` is that no other message is going to have such a property. There's a test, equivalent to
```
if (msg.data['_wpn']) {
```
I can imagine minified properties being called `ap`, which might trigger a false positive, but it's unlikely there'd be a `_wpn`.

So my take, I can see reasoning for the underscore here, but for anything else it's just extra characters.

sbc100:
This change actually inspired me to do exactly the opposite here for pthreads: https://github.com/emscripten-core/emscripten/pull/22555

If we even manage to implement the single-file approach like #21701 for wasm workers then we can do that same here too.

cwoffenden:
> This change actually inspired me to do exactly the opposite here for pthreads: #22555

A better solution all round, sure.

sbc100 (approved):
I'd love to see a test for this in `browser_tests.py`.   Can we wait on this change until @juj has updated the audio_worklet tests to be more robust/accurate.  IIUC they are currently passing erroneously despite some recent breakages.

lgtm though, @juj if you want to land this first thats fine with me to.

src/runtime_shared.js
```diff
@@ -13,7 +13,7 @@
     if (!shouldExport) {
       if (MODULARIZE && EXPORT_ALL) {
         shouldExport = true;
-      } else if (AUDIO_WORKLET && (x == 'HEAP32' || x == 'HEAPU32')) {
+      } else if (AUDIO_WORKLET && (x == 'HEAP32' || x == 'HEAPU32' || x == 'HEAPF32')) {
```
sbc100, line 16: Does this mean that if the code in the audio worklet accesses any other heaps that these 3 it would fail?

juj: > Does this mean that if the code in the audio worklet accesses any other heaps that these 3 it would fail?

I don't think so - it is only for the scope of audio_worklet.js wrapper, which needs these explicitly exported. Other code works like usual, accessing the heaps via the usual JS code inside the module.

sbc100: So I'm not sure why this change is needed then?  (since there is no corresponding change in `audio_worklet.js`).  What am I missing?

cwoffenden: Without this change the worklet fails to build with minimal runtime. The tests weren’t catching it (same as the Closure tests weren’t running).

juj: > So I'm not sure why this change is needed then?

To make e.g. this line work: https://github.com/emscripten-core/emscripten/blob/d9c260d877fab10f34241325ec4c4c2b072d899c/src/audio_worklet.js#L70

sbc100: I see.  The `globalThis.HEAPF32 = Module['HEAPF32'];` line already existed but the heap wasn't always exported.  Shouldn't the list of heaps here exactly match the list of heaps set of `globalObject` in `audio_worklet.js`?

i.e. can we remove `HEAP32` from the list here?

cwoffenden: > i.e. can we remove `HEAP32` from the list here?

Isn’t it required by `stackAlloc`?

juj:
This GitHub/CircleCI keeps on giving.. 

there was a test failure on build-test and test-browser-chrome due to the single offset converter test failure, [which we know to be flaky](https://github.com/emscripten-core/emscripten/pull/22480#issuecomment-2324362708).

So I opened the test page, and clicked on "Re-run failed tests" to rerun the test-browser-chrome suite again.

However that caused the test runner to re-run tons of other suites, *except* that failing test-browser-chrome-suite that I wanted to rerun.

image

And now other suites are failing with HTTP 500 No Fun for You.

image

To me it seems that all relevant tests are passing, I'd be happy to land this PR.

src/runtime_shared.js
juj: > i.e. can we remove HEAP32 from the list here?

Leave for a separate PR.

cwoffenden: > > i.e. can we remove HEAP32 from the list here?
> 
> Plausible. Leave for a separate PR.

Since I needed to re-run the (erroneous) failed tests I took the excuse to remove `HEAP32`. Testing manually beforehand its removal runs fine with `MINIMAL_RUNTIME`.

cwoffenden:
> This GitHub/CircleCI keeps on giving..

Run it enough times and it finally all passed!

sbc100 (approved):

PR #22476, opened 2024-08-29
Fix flakiness in test_futimens_rawfs

sbc100:
This newly added test was flaky because noderawfs wasn't handling the value of -1 correctly here.

### DIFF
--- a/src/library_noderawfs.js
+++ b/src/library_noderawfs.js
@@ -93,8 +93,8 @@ addToLibrary({
     },
     utime(path, atime, mtime) {
       // -1 here for atime or mtime means UTIME_OMIT was passed.  Since node
-      // doean't support this concept we need first find the existing timestamps
-      // in order to preserve them.
+      // doesn't support this concept we need to first find the existing
+      // timestamps in order to preserve them.
       if (atime == -1 || mtime == -1) {
         var st = fs.statSync(path);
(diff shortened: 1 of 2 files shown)


src/library_noderawfs.js
```diff
@@ -99,7 +99,17 @@ addToLibrary({
       var stream = FS.getStreamChecked(fd);
       fs.ftruncateSync(stream.nfd, len);
     },
-    utime(path, atime, mtime) { fs.utimesSync(path, atime/1000, mtime/1000); },
+    utime(path, atime, mtime) {
+      // -1 here for atime or mtime means UTIME_OMIT was passed.  Since node
+      // doean't support this concept we need first find the existing timestamps
```
kripken, line 104:
```suggestion
      // doesn't support this concept we need to first find the existing timestamps
```

kripken (approved):

brendandahl:
I've noticed this test failing quite a bit lately. Can we merge this in?

sbc100:
IIRC there was some reason I didn't land this already.. perhaps that it didn't actually fix the flakeyness.. I'm not sure.

Fart10 (commented):
😜

sbc100:
> 😜

Thanks for the feedback!

PR #22488, opened 2024-09-03
Let leakWarning carry the string representation, not the stack itself

stbergmann:
At least with recent Chrome, I ran into a situation where an object of mine apparently didn't get garbage collected because it was retained via such a leakWarning stack trace.  (I'm not sure about the internal workings of that, though; but the object started to get garbage collected once I did this change.)

I have no idea if there are any other up-/downsides to carrying the stack vs. just the string here.  And my explanation of what exactly it fixed for me is probably a bit vague.  But wanted to bring it up nevertheless---maybe it is considered a good change anyway.


sbc100:
@RReverser @brendandahl

src/embind/embind.js
```diff
@@ -1526,13 +1526,14 @@ var LibraryEmbind = {
...
-        info.leakWarning = new Error(`Embind found a leaked C++ instance ${cls.name} <${ptrToString($$.ptr)}>.\n` +
+        var err = new Error(`Embind found a leaked C++ instance ${cls.name} <${ptrToString($$.ptr)}>.\n` +
         "We'll free it automatically in this case, but this functionality is not reliable across various environments.\n" +
         "Make sure to invoke .delete() manually once you're done with the instance instead.\n" +
         "Originally allocated"); // `.stack` will add "at ..." after this sentence
         if ('captureStackTrace' in Error) {
-          Error.captureStackTrace(info.leakWarning, RegisteredPointer_fromWireType);
+          Error.captureStackTrace(err, RegisteredPointer_fromWireType);
         }
+        info.leakWarning = err.stack.replace(/^Error: /, '');
```
RReverser, line 1536: This is not ideal because stack stringification is relatively expensive, so initially we tried to delay it only until we actually need said stack... but since it's under `ASSERTIONS`, maybe that's fine.

brendandahl (approved):
Since this is only happening with assertions I think it's fine. If it becomes an issue we can add a separate flag for it.

PR #22489, opened 2024-09-03
Check if cmake is on the PATH in emcmake.py

Peter0x44:
Currently, users who run emcmake without cmake on the PATH are greeted with confusing errors like:

```
emcmake cmake ..
configure: cmake .. -DCMAKE_TOOLCHAIN_FILE=D:\src\emsdk\upstream\emscripten\cmake\Modules\Platform\Emscripten.cmake -DCMAKE_CROSSCOMPILING_EMULATOR=D:/src/emsdk/node/18.20.3_64bit/bin/node.exe -G Ninja emcmake: error: 'cmake .. -DCMAKE_TOOLCHAIN_FILE=D:\src\emsdk\upstream\emscripten\cmake\Modules\Platform\Emscripten.cmake -DCMAKE_CROSSCOMPILING_EMULATOR=D:/src/emsdk/node/18.20.3_64bit/bin/node.exe -G Ninja' failed: [WinError 2] The system cannot find the file specified
```

### DIFF
--- a/emcmake.py
+++ b/emcmake.py
@@ -45,6 +45,11 @@ variables so that emcc etc. are used. Typical usage:
     # See https://github.com/emscripten-core/emscripten/issues/15522
     args.append(f'-DCMAKE_CROSSCOMPILING_EMULATOR={node_js}')
 
+  # Print a better error if we have no CMake executable on the PATH
+  if not os.path.isabs(args[0]) and not shutil.which(args[0]):
+    print(f'emcmake: cmake executable not found on PATH: `{args[0]}`')
+    return 1
+
   # On Windows specify MinGW Makefiles or ninja if we have them and no other
   # toolchain was specified, to keep CMake from pulling in a native Visual
   # Studio, or Unix Makefiles.
--- a/emcmake.py
+++ b/emcmake.py
@@ -46,7 +46,7 @@ variables so that emcc etc. are used. Typical usage:
     args.append(f'-DCMAKE_CROSSCOMPILING_EMULATOR={node_js}')
 
   # Print a better error if we have no CMake executable on the PATH
-  if not os.path.isabs(args[0]) and not shutil.which(args[0]):
+  if not os.path.dirname(args[0]) and not shutil.which(args[0]):
     print(f'emcmake: cmake executable not found on PATH: `{args[0]}`')
     return 1
 
--- a/emcmake.py
+++ b/emcmake.py
@@ -47,7 +47,7 @@ variables so that emcc etc. are used. Typical usage:
 
   # Print a better error if we have no CMake executable on the PATH
   if not os.path.dirname(args[0]) and not shutil.which(args[0]):
-    print(f'emcmake: cmake executable not found on PATH: `{args[0]}`')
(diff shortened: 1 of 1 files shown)


Peter0x44:
I'm not sure if users have any reason *not* to have CMake on their PATH, maybe the error could be adjusted to only happen if argv[0] == "cmake" or so.

emcmake.py
```diff
@@ -44,6 +44,10 @@ def has_substr(args, substr):
     # See https://github.com/emscripten-core/emscripten/issues/15522
     args.append(f'-DCMAKE_CROSSCOMPILING_EMULATOR={node_js}')
 
+  if not shutil.which('cmake'):
```
sbc100, line 47: How about:
```
if not os.path.isabs(args[0]) and not shutil.which(args[0]):
   print(f'emcmake: cmake executable not found in PATH: `{arg}`')
```

Peter0x44: Better. Good idea.

sbc100 (approved):

emcmake.py
```diff
@@ -44,6 +44,11 @@ def has_substr(args, substr):
     # See https://github.com/emscripten-core/emscripten/issues/15522
     args.append(f'-DCMAKE_CROSSCOMPILING_EMULATOR={node_js}')
 
+  # Print a better error if we have no CMake executable on the PATH
+  if not os.path.isabs(args[0]) and not shutil.which(args[0])
```
sbc100, line 48: Missing colon here

Peter0x44: Sorry. Nice catch.

emcmake.py
```diff
@@ -45,6 +45,11 @@ def has_substr(args, substr):
     # See https://github.com/emscripten-core/emscripten/issues/15522
     args.append(f'-DCMAKE_CROSSCOMPILING_EMULATOR={node_js}')
 
+  # Print a better error if we have no CMake executable on the PATH
+  if not os.path.isabs(args[0]) and not shutil.which(args[0]):
```
sbc100: I recently learned that this should probably be `if not os.path.dirname(args[0])` here instead of `isabs` since you could imagine doing something like this `emcmake ./cmake/bin/cmake ..`.  In this example you don't want to call which but that path is not absolute, its relative.

emcmake.py
```diff
@@ -45,6 +45,11 @@ def has_substr(args, substr):
...
+    print(f'emcmake: cmake executable not found on PATH: `{args[0]}`')
```
sbc100: Add `file=sys.stderr` here.

sbc100 (approved):

Peter0x44:
I think some flaky install for a test failed and stopped this from getting merged.
```
Failures
 - ninja (exited 1) - ninja not installed. An error occurred during installation:
 Error downloading 'ninja.1.13.1' from 'https://community.chocolatey.org/api/v2/package/ninja/1.13.1'.
 - pkgconfiglite (exited 1) - pkgconfiglite not installed. An error occurred during installation:
 Error downloading 'pkgconfiglite.0.28.0' from 'https://community.chocolatey.org/api/v2/package/pkgconfiglite/0.28'.
 ```

PR #19375, opened 2023-05-16
Fix test\runner sanity.test_aaa_normal to work on Windows

juj:

PR #22492, opened 2024-09-03
[test] Allow `--crossplatform-only` to include sanity tests

sbc100:
See #19375

### DIFF
--- a/test/runner.py
+++ b/test/runner.py
@@ -135,7 +135,7 @@ def get_all_tests(modules):
 
 
 def get_crossplatform_tests(modules):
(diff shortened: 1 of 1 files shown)


juj:
Did the @crossplatform directive I added get ignored without this? LGTM

juj (approved):

sbc100:
> Did the `@crossplatform` directive I added get ignored without this? LGTM

yes, the `--crossplatform-only` flag (and thus the attribute that controls it) was previously only used for `core0` and `other` tests.

PR #22494, opened 2024-09-03
WebGPU: support Premultiplied canvas output

### LINKED ISSUE #22491: [webgpu] Support WGPUCompositeAlphaMode_Premultiplied
Browsers already supports Premultiplied alpha mode in surface configuration if use JS way to create webgpu context handles.
But emsdk still have it as TODO option.
Please, make it works.

```c++
WGPUSurfaceConfiguration surfaceConfiguration {
    .device = device,
    .format = preferredFormat,
    .usage = WGPUTextureUsage_RenderAttachment,
    .alphaMode = WGPUCompositeAlphaMode_Premultiplied, // <-- does not supported
    .width = w,
    .height = h,
    .presentMode = WGPUPresentMode_Fifo,
};
wgpuSurfaceConfigure(surface, &surfaceConfiguration);
```

kainino0x:
(Note the default HTML template page's canvas has a black background, so the difference is not visible there.)

Fixes #22491

### DIFF
--- a/src/library_webgpu.js
+++ b/src/library_webgpu.js
@@ -109,6 +109,7 @@ wgpu${type}Release: (id) => WebGPU.mgr${type}.release(id),`;
     CompositeAlphaMode: {
       Auto: 0,
       Opaque: 1,
+      Premultiplied: 2,
     },
     CreatePipelineAsyncStatus: {
       Success: 0,
@@ -436,6 +437,11 @@ var LibraryWebGPU = {
       'clamp-to-edge',
       'repeat',
       'mirror-repeat',
+    ],
+    AlphaMode: [
+      undefined, // "Auto" uses the default (which is always opaque according to the spec's IDL)
+      'opaque',
+      'premultiplied',
     ],
     BlendFactor: [
       undefined,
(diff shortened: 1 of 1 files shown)


kainino0x:
@lokokung

src/library_webgpu.js
```diff
@@ -2779,7 +2781,8 @@ var LibraryWebGPU = {
       "format": WebGPU.TextureFormat[
         {{{ gpu.makeGetU32('config', C_STRUCTS.WGPUSurfaceConfiguration.format) }}}],
       "usage": {{{ gpu.makeGetU32('config', C_STRUCTS.WGPUSurfaceConfiguration.usage) }}},
-      "alphaMode": "opaque",
+      "alphaMode": WebGPU.AlphaMode[
```
sbc100, line 2784: We tend to prefer single quotes for strings in the JS library.

sbc100 (approved):

src/library_webgpu.js
kainino0x: This would make it inconsistent with the rest of the file, so I'd like to leave it for now if that's OK. I've filed #22496 since we now have a small backlog of assorted fixes we should do.

sbc100: sgtm

PR #22498, opened 2024-09-04
Use object literal shorthand in js_optimizer. NFC

sbc100:
Without this change running code trough acorn-optimizer will expand these object literals, removing the use of the shorthand.

### DIFF
--- a/test/js_optimizer/test-object-literals.js
+++ b/test/js_optimizer/test-object-literals.js
@@ -1,0 +1,14 @@
+var foo = 1;
+var baz = {};
+
+// Verify that shorthand object literals are preserved.
+var bar = {
+  foo,
+  baz,
+};
+
+// Objects that don't use shorthand should also be preserved, as-is
+var bar2 = {
+  foo: foo,
+  baz: baz,
+};
--- a/test/js_optimizer/test-object-literals.js
+++ b/test/js_optimizer/test-object-literals.js
@@ -7,8 +7,9 @@ var bar = {
   baz,
 };
 
-// Objects that don't use shorthand should also be preserved, as-is
+// Objects that could use shorthand are also converted where possible.
 var bar2 = {
   foo: foo,
-  baz: baz,
+  "baz": baz,
+  bar: foo,
 };
--- a/tools/acorn-optimizer.mjs
+++ b/tools/acorn-optimizer.mjs
@@ -2107,6 +2107,7 @@ if (!noPrint) {
     keep_quoted_props: closureFriendly, // for closure
     wrap_func_args: false, // don't add extra braces
     comments: true, // for closure as well
+    shorthand: true, // Use object literal shorthand notation
   });
 
   output += '\n';
--- a/test/js_optimizer/JSDCE-defaultArg-output.js
+++ b/test/js_optimizer/JSDCE-defaultArg-output.js
@@ -2,7 +2,7 @@ var usedAsDefaultArg = 42;
 
 var usedAsDefaultArg2 = [ 1, 2 ];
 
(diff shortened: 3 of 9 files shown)


test/js_optimizer/test-object-literals.js
```diff
@@ -0,0 +1,14 @@
+var foo = 1;
+var baz = {};
+
+// Verify that shorthand object literals are preserved.
+var bar = {
+  foo,
+  baz,
+};
+
+// Objects that don't use shorthand should also be preserved, as-is
```
kripken, line 10: This looks modified in the output above? `foo: foo,` was turned into just `foo,`.

But that seems ok - is there a reason to avoid that?

sbc100: Oh.. I thought it was able to distinguish.  I guess not.  I'll update the comment.

test/js_optimizer/test-object-literals.js
```diff
@@ -0,0 +1,14 @@
...
+// Verify that shorthand object literals are preserved.
+var bar = {
+  foo,
+  baz,
+};
+
+// Objects that could use shorthand are also converted where possible.
+var bar2 = {
+  foo: foo,
+  baz: baz,
```
kripken, line 13: Maybe add one that can't be converted?
```suggestion
  baz: baz,
  other1: other2,
```

kripken (approved):

tools/acorn-optimizer.mjs
```diff
@@ -2107,6 +2107,7 @@ if (!noPrint) {
     keep_quoted_props: closureFriendly, // for closure
     wrap_func_args: false, // don't add extra braces
     comments: true, // for closure as well
+    shorthand: true, // Use object literal shorthand notation
```
juj, line 2110: Won't this wreak havoc with Closure Compiler, like happened with that Web Audio bug?

sbc100: Thats a good question.  I don't think it will do this if string keys are used.  I'll add a test for that.

sbc100: It looks like it does create new shorthand objects, even when the keys are strings.

But even without this change it looks like acorn is removing the string quotes from the keys, which I think would also break closure wouldn't it?   Maybe we just don't run any acorn passes before we run closure?

sbc100: I confirmed that running the acorn_optmizer will remove the quotes from string key today.

I also confirmed the if the quotes a removed from string keys then closure can/will minify them, which will break, for example, pthreads which wants to post message an object with specific string keys.

This means that if we were running closure after acorn today things would already be broken.  Therefore I think this change does not make things worse WRT closure.

sbc100: Oh I see now, we have `keep_quoted_props: closureFriendly, // for closure` in acorn-optimizer.mjs which causes those quotes not to be removed the keys.   Phew, that was confusing.   Mystery solved.   This change does effect the postMessaged objects with quoted keys.

juj: > This change does effect the postMessaged objects with quoted keys.

What kind of construct should one use to avoid their postMessage()s from breaking under Closure like this?

sbc100: IIUC the way to stop closure from minifying object field names is to use quotes.    I believe that was true before, and is still true after this change.   i.e. this change doesn't change make things worse for closure users.

sbc100: Yes, you are correct, it will opportunistically apply the shorthand, but we explicit set `keep_quoted_props` when we know the output has to be closure friendly, so it should not effect those quoted properties.

PR #22501, opened 2024-09-04
Add JSPI support to MINIMAL_RUNTIME build mode.

juj:


src/preamble_minimal.js
```diff
@@ -23,8 +23,15 @@ function assert(condition, text) {
 }
 #endif
 
+#if ASYNCIFY == 1 // ASYNCIFY-mode requires checking ABORT variable to avoid operating if code has aborted during an unwind
```
sbc100, line 26: Is this a debug feature or do we need to add this new variable even in release builds?

I worry that this makes the feature matrix even more complicated and there could be other places in the codebase that deal with `ABORT` that need to be changes now that `ABORT` does exist undef `MINIMAL_RUNTIME` but only with `ASYNCIFY` is enabled.

sbc100 (approved):

src/preamble_minimal.js
```diff
@@ -23,8 +23,15 @@ function assert(condition, text) {
...
+var ABORT = 0;
```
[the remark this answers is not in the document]
juj: Err, yes. Updated the title to clarify that this PR addresses both ASYNCIFY and JSPI.

src/preamble_minimal.js
juj: I added that variable to help places like [this](https://github.com/emscripten-core/emscripten/blob/main/src/library_async.js#L311) and [this](https://github.com/emscripten-core/emscripten/blob/8969fd212f5c551f46a8e8c713994ac97993d466/src/library_async.js#L138) to compile.

There seem to be places that call to it in library_async.js: https://github.com/emscripten-core/emscripten/blob/main/src/library_async.js#L18 so looks like in ASYNCIFY==1 mode we cannot avoid adding ABORT in the build. In ASYNCIFY==2 mode, maybe we could.

Overall, I think the support would be ok to be expanded incrementally via proof by test cases.

sbc100 (approved):

sbc100 (commented):
Can we land this?

sbc100:
The esm_integration test failure is something is broken on maim.  My fault: https://github.com/emscripten-core/emscripten/pull/24946.

Feel free to to land without waiting for that to get fixed.

PR #22502, opened 2024-09-04
When building with -sENVIRONMENT=web, do not produce WASI'isms in the MINIMAL_RUNTIME build.

juj:


src/postamble_minimal.js
```diff
@@ -106,7 +106,13 @@ var imports = {
   'a': wasmImports,
 #else // MINIFY_WASM_IMPORTED_MODULES
   'env': wasmImports,
+
+// There does not currently exist a ENVIRONMENT_MAY_BE_WASI, so liken it to the shell environments.
+// (e.g. anyone building with -sENVIRONMENT=web won't be running in WASI)
+#if ENVIRONMENT_MAY_BE_NODE || ENVIRONMENT_MAY_BE_SHELL
   '{{{ WASI_MODULE_NAME }}}': wasmImports,
```
sbc100, line 113: I believe there are parts of the C library that depend on WASI-named syscalls.

We could perhaps see if any of the imports come from the WASI namespace and make this line conditional (for both minimal and regular runtime).

sbc100: Also, won't those that care about code size already fall into `MINIFY_WASM_IMPORTED_MODULES` above, avoiding this code block completely?

juj: > Also, won't those that care about code size already fall into `MINIFY_WASM_IMPORTED_MODULES` above, avoiding this code block completely?

This only happens if METADCE is being run, which it isn't if ASYNCIFY is enabled.

src/postamble_minimal.js
```diff
@@ -106,7 +106,13 @@ var imports = {
   'a': wasmImports,
 #else // MINIFY_WASM_IMPORTED_MODULES
   'env': wasmImports,
+
+// There does not currently exist a ENVIRONMENT_MAY_BE_WASI, so liken it to the shell environments.
+// (e.g. anyone building with -sENVIRONMENT=web won't be running in WASI)
+#if ENVIRONMENT_MAY_BE_NODE || ENVIRONMENT_MAY_BE_SHELL
```
sbc100, line 112: This comment seems to imply a misunderstnading about how we use the WASI APIs in emscripten.    The WASI APIs are like the regular `__syscall_` functions that we implement in JS.     They are just JS library functions like any other and can be used in any envrionment.

For example the `__proc_exit` function which is used to implement `exit` is a WASI API.

juj: That's a good point. I understand that in many cases the functions overlap. Although from the code size creep perspective that is not working out today, i.e. using that `__proc_exit` as an example, I am getting that in a build even when I build with `-sEXIT_RUNTIME=0`, so functionally that would warrant for a `#if ENVIRONMENT_IS_WASI` construct if there is no other way to avoid leaking in the unused stuff.

juj:
Closing old stale PR.

Issue #22376, opened 2024-08-14
EmscriptenWebSocketMessageEvent->socket is same for two different socket


dpoizl-atomontage:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.64 (a1fe3902bf73a3802eae0357d273d0e37ea79898)
clang version 19.0.0git (https:/github.com/llvm/llvm-project 4d8e42ea6a89c73f90941fd1b6e899912e31dd34)
Target: wasm32-unknown-emscripten
Thread model: posix

**Failing command line in full:**
em++ main.cpp -o test.html -sALLOW_MEMORY_GROWTH=1 -lwebsocket

```
#include <stdio.h>
#include <memory.h>
#include <stdlib.h>
#include <cmath>
#include <iostream>
#include <emscripten.h>
#include <emscripten/html5.h>
#include <emscripten/websocket.h>

class TcpSocket
{
	EMSCRIPTEN_WEBSOCKET_T socket;
public:
TcpSocket(int port)
{
	EmscriptenWebSocketCreateAttributes attributes;
	emscripten_websocket_init_create_attributes(&attributes);
	char url[200];
	sprintf(url, "ws://localhost:%d", port);
	attributes.url = url;
	attributes.protocols = "binary";
	attributes.createOnMainThread = EM_TRUE;
	socket = emscripten_websocket_new(&attributes);
	bool registrationSuccessful = true;
	registrationSuccessful &= (emscripten_websocket_set_onopen_callback(socket, this, OnOpen) == 0);
	registrationSuccessful &= (emscripten_websocket_set_onclose_callback(socket, this, OnClose) == 0);
	registrationSuccessful &= (emscripten_websocket_set_onerror_callback(socket, this, OnError) == 0);
	registrationSuccessful &= (emscripten_websocket_set_onmessage_callback(socket, this, OnMessage) == 0);
	printf("registration %s\n", registrationSuccessful ? "successfull" : "failed");
	
	printf("socket open %d %s\n", socket, url);
}
static EM_BOOL OnMessage(int eventType, const EmscriptenWebSocketMessageEvent *messageEvent, void *network)
{
	printf("OnMessage %d socket %d numbytes %d %p\n", eventType, messageEvent->socket, messageEvent->numBytes, network);
	return 0;
}
[...]

PR #22505, opened 2024-09-04
Fix socket ID used in socket events when multiple sockets are used. NFC

sbc100:
We were lacking tests that used multiple sockets at the same time.

Fixes: #22376

### DIFF
--- a/test/websocket/test_websocket_send.c
+++ b/test/websocket/test_websocket_send.c
@@ -3,8 +3,14 @@
 #include <emscripten/websocket.h>
 #include <assert.h>
 
+// This test performs that same server communications using two different
+// sockets. This verifies that multiple sockets are supported at the
+// simultaneously.
+EMSCRIPTEN_WEBSOCKET_T sock1;
+EMSCRIPTEN_WEBSOCKET_T sock2;
+
 EM_BOOL WebSocketOpen(int eventType, const EmscriptenWebSocketOpenEvent *e, void *userData) {
-  printf("open(eventType=%d, userData=%p)\n", eventType, userData);
+  printf("open(socket=%d, eventType=%d, userData=%p)\n", e->socket, eventType, userData);
 
   emscripten_websocket_send_utf8_text(e->socket, "hello on the other side");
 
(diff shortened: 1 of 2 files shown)


test/websocket/test_websocket_send.c
```diff
@@ -3,8 +3,14 @@
 #include <emscripten/websocket.h>
 #include <assert.h>
 
+// This test performs that same server communications using two different
+// sockets. This verifies that multiple sockets are supported at the
```
kripken, line 7:
```suggestion
// sockets. This verifies that multiple sockets are supported
```

test/websocket/test_websocket_send.c
```diff
@@ -15,43 +21,50 @@ EM_BOOL WebSocketOpen(int eventType, const EmscriptenWebSocketOpenEvent *e, void
...
-
-  // We expect to receive the text message beofre the binary one
-  assert(text_received);
-  printf("binary data:");
-  for (int i = 0; i < e->numBytes; ++i) {
-    printf(" %02X", e->data[i]);
-    assert(e->data[i] == i);
+    text_received++;
+  } else {
+    // We expect to receive the text message beofre the binary one
```
kripken, line 43:
```suggestion
    // We expect to receive the text message before the binary one
```

test/websocket/test_websocket_send.c
```diff
@@ -15,43 +21,50 @@ EM_BOOL WebSocketOpen(int eventType, const EmscriptenWebSocketOpenEvent *e, void
...
+    printf("binary data:");
```
kripken, line 44: But is that comment right? The assertion that verifies it was removed.

test/websocket/test_websocket_send.c
```diff
@@ -15,43 +21,50 @@ EM_BOOL WebSocketOpen(int eventType, const EmscriptenWebSocketOpenEvent *e, void
...
+  assert(e->socket == sock1 || e->socket == sock2);
   if (e->isText) {
     printf("text data: \"%s\"\n", e->data);
     assert(strcmp((const char*)e->data, "hello on the other side") == 0);
-    text_received = 1;
-    return 0;
-  }
-
-  // We expect to receive the text message beofre the binary one
-  assert(text_received);
```
kripken, line 38: With this assertion removed, this variable is no longer needed AFAICT? Nor is `binary_received`.

test/websocket/test_websocket_send.c
```diff
@@ -15,43 +21,50 @@ EM_BOOL WebSocketOpen(int eventType, const EmscriptenWebSocketOpenEvent *e, void
...
+    emscripten_websocket_delete(e->socket);
+    if (e->socket == sock1) {
+      sock1 = 0;
+    } else {
+      sock2 = 0;
+    }
+    // Once both sockets are closed we are done.
+    if (!sock1 && !sock2) {
+      emscripten_force_exit(0);
+    }
```
kripken, line 61: This logic should be outside of the else, no? That is the only way that we can accept both orders, I think, and not depend on binary being last.

test/websocket/test_websocket_send.c
sbc100: I put the assertion back.  It still holds.

test/websocket/test_websocket_send.c
sbc100: Whatever interleaving happens the binary message will always be last.   Since for each socket the binary message comes second.

We only close the socket once we get the binary message from it, so the only place that sock1 or sock2 could be made to be null is here inside this else branch.

kripken: In that case we can put a return after line 39, and avoid indenting all the rest in the `else`, the way it was before? Or am I missing a reason to put this code in the indented position?

kripken: lgtm with that change + the typo in the other comment.

sbc100: Why do you prefer adding an early return and putting the binary case outside the `if/else`?   The `if/else` seems quite natural to be since it one case is `text` and other other is `binary`?

sbc100: I guess it makes the diff smaller?   Done.

kripken (approved):

test/websocket/test_websocket_send.c
kripken: Yeah, it is more symmetrical the other way, but in general I think an early return is nice when one arm is much longer, as it allows that arm to be unindented. But I was also unsure if the indentation had another reason here, that's why I was asking (if we allowed more orders than before). Anyhow, lgtm either way.

PR #22509, opened 2024-09-05
Fix automatic run of interactive web audio tests.

juj:
Audio contexts cannot start until human interacts with the page, e.g. by clicking a button.

Therefore interactive audioworklet tests require user to press a button.

However, after clicking a button, all later interactive tests will have the audio context access already unlocked, so user does not need to click a button on each test when running multiple audio worklet tests, only the first one.

In such scenario, there was a bug that nothing played back from speakers on later tests, because the audio worklet was not connected to output on those later tests until user clicks a button. But since this button click was not needed, the subsequent interactive tests would surprisingly pass without actually playing back any audio.

### DIFF
--- a/test/webaudio/audioworklet_emscripten_futex_wake.cpp
+++ b/test/webaudio/audioworklet_emscripten_futex_wake.cpp
@@ -29,12 +29,12 @@ EM_BOOL ProcessAudio(int numInputs, const AudioSampleFrame *inputs, int numOutpu
 EM_JS(void, InitHtmlUi, (EMSCRIPTEN_WEBAUDIO_T audioContext, EMSCRIPTEN_AUDIO_WORKLET_NODE_T audioWorkletNode), {
   audioContext = emscriptenGetAudioObject(audioContext);
   audioWorkletNode = emscriptenGetAudioObject(audioWorkletNode);
+  audioWorkletNode.connect(audioContext.destination);
   let startButton = document.createElement('button');
   startButton.innerHTML = 'Start playback';
   document.body.appendChild(startButton);
 
   startButton.onclick = () => {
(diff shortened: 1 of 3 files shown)


test/webaudio/audioworklet_emscripten_futex_wake.cpp
```diff
@@ -29,12 +29,12 @@ EM_BOOL ProcessAudio(int numInputs, const AudioSampleFrame *inputs, int numOutpu
 EM_JS(void, InitHtmlUi, (EMSCRIPTEN_WEBAUDIO_T audioContext, EMSCRIPTEN_AUDIO_WORKLET_NODE_T audioWorkletNode), {
   audioContext = emscriptenGetAudioObject(audioContext);
   audioWorkletNode = emscriptenGetAudioObject(audioWorkletNode);
+  audioWorkletNode.connect(audioContext.destination);
```
sbc100, line 32: I'm not sure I understand what happens if they tests in run in isolation?  Doesn't previous test need to have run in order for it to work?

Or will `.connect` work, but the audio won't actually start playing until the `onclick`?

juj: If one runs
```
test/runner interactive.test_audio_worklet*
```
then the first test asks for user to press a button. Then audio plays.
Then all subsequent tests skip asking for the user to press a button, since due to the way that harness uses iframes, audio unlock carries over to subsequent tests. So audio is played out automatically, but because we only connected the audio to output when user would press a button (which no longer needs to happen with this unlock in effect), all these subsequent tests would pass without the user hearing any audio - making it seem confusing how the tests are passing but no audio plays.

So to fix, we can unconditionally connect all the audio to the output (not sure why I didn't do that in the first place)

PR #22522, opened 2024-09-05
Avoid reading in logo data if we don't need it.

lokokung:
When using a custom shell (potentially without the availability of the logo image), only try to read the logo if necessary.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -2573,9 +2573,10 @@ def generate_traditional_runtime_html(target, options, js_target, target_basenam
 
   shell = do_replace(shell, '{{{ SCRIPT }}}', script.replacement())
   shell = shell.replace('{{{ SHELL_CSS }}}', utils.read_file(utils.path_from_root('src/shell.css')))
-  logo_filename = utils.path_from_root('media/powered_by_logo_shell.png')
-  logo_b64 = base64_encode(logo_filename)
-  shell = shell.replace('{{{ SHELL_LOGO }}}', f'<img id="emscripten_logo" src="data:image/png;base64,{logo_b64}">')
+  if '{{{ SHELL_LOGO }}}' in shell:
+    logo_filename = utils.path_from_root('media/powered_by_logo_shell.png')
+    logo_b64 = base64_encode(logo_filename)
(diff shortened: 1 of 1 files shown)


tools/link.py
```diff
@@ -2573,9 +2573,10 @@ def generate_traditional_runtime_html(target, options, js_target, target_basenam
 
   shell = do_replace(shell, '{{{ SCRIPT }}}', script.replacement())
   shell = shell.replace('{{{ SHELL_CSS }}}', utils.read_file(utils.path_from_root('src/shell.css')))
-  logo_filename = utils.path_from_root('media/powered_by_logo_shell.png')
-  logo_b64 = base64_encode(logo_filename)
-  shell = shell.replace('{{{ SHELL_LOGO }}}', f'<img id="emscripten_logo" src="data:image/png;base64,{logo_b64}">')
+  if '{{{ SHELL_LOGO }}}' in shell:
+    logo_filename = utils.path_from_root('media/powered_by_logo_shell.png')
+    logo_b64 = base64_encode(logo_filename)
+    shell = shell.replace('{{{ SHELL_LOGO }}}', f'<img id="emscripten_logo" src="data:image/png;base64,{logo_b64}">')
```
sbc100, line 2579: Is this worth it?   If so, perhaps we should do the same for `SHELL_CSS` above?

lokokung: Actually, since this was changed in the other PR that landed earlier today to avoid having to read the file in, it might be fine now? Before, the Python script would fail since it couldn't read the file so it was necessary, but I think now if it's just a path it should work?

As for the CSS though, since that's being read in, I did add that one as well.

sbc100: The change I made today doesn't stop the file being read in.. it just modes the call to read into `base64_encode`.

The `logo_filename` should always exist in emscripten so I don't think it should ever fail to read that file.

lokokung:
Oh, hmm that's reasonable. I think in our internal copy of emscripten, we actually removed the media files (hence the logo file), and it was causing some errors. Let me see if we could just bring back the media folder on our side if it would better to leave this as is.

PR #22533, opened 2024-09-06
Update npm dependencies. NFC

sbc100:
I used npm-check-update to generate this PR:

1. npm install --global npm-check-updates
2. ncu -u
3. npm install
 
After I did this I found that eslint config file needed to be updated which forced me to run `npx @eslint/migrate-config .eslintrc.yml` to create the new (and less readable IMHO) config file.


dschuff (approved):
LGTM but fix the title so it doesn't include 2 different PR numbers

sbc100:
I had to update the eslint config file .. see description.

PR #22537, opened 2024-09-06
Use consistent helper for tests that use -sMODULARIZE

sbc100:
As part of this I also extended the `#preprocess` directive so it can be used with `--extern-pre-js` and `--extern-post-js` as well as the existing supprot for `--pre-js` and `--post-js`.

### DIFF
--- a/test/modularize_post_js.js
+++ b/test/modularize_post_js.js
@@ -1,0 +1,10 @@
+#preprocess
+
+// This file gets included via `--extern-post-js` in order to run tests
+// that are built with `-sMODULARIZE`.
+
+#if PTHREADS
+// Avoid instantiating the module on pthreads.
+if (!isPthread)
+#endif
+{{{ EXPORT_NAME }}}();
--- a/test/runner.py
+++ b/test/runner.py
@@ -114,7 +114,7 @@ def check_js_engines():
 
 def get_and_import_modules():
   modules = []
-  for filename in glob.glob(os.path.join(os.path.dirname(__file__), 'test*.py')):
+  for filename in glob.glob(os.path.join(common.TEST_ROOT, 'test*.py')):
     module_dir, module_file = os.path.split(filename)
     module_name, module_ext = os.path.splitext(module_file)
     __import__(module_name)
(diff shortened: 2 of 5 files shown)


test/modularize_post_js.js
```diff
@@ -0,0 +1,10 @@
+#preprocess
+
+// This file get included via `--extern-post-js` in order to run tests
```
kripken, line 3:
```suggestion
// This file gets included via `--extern-post-js` in order to run tests
```

test/modularize_post_js.js
```diff
@@ -0,0 +1,10 @@
...
+// that are built with `-sMODULARIZE`.
+
+#if PTHREADS
+// Avoid instantiating the module on pthreads.
+if (!isPthread)
+#endif
+{{{ EXPORT_NAME }}}();
```
kripken, line 10: This seems like a general thing that all users will want, but I don't have a good idea for how to provide that. We could make `Module()` do nothing at all on pthreads, in theory, but that seems risky, even if I can't think of a use for it.

test/modularize_post_js.js
```diff
@@ -0,0 +1,10 @@
+#preprocess
+
+// This file get included via `--extern-post-js` in order to run tests
+// that are built with `-sMODULARIZE`.
+
+#if PTHREADS
+// Avoid instantiating the module on pthreads.
+if (!isPthread)
```
kripken, line 8: `isPthread` is only defined inside the emscripten JS output, isn't it? How is it usable from outside it, in `--extern-post-js`?

(The code this replaces does things like check `typeof importScripts != "function"` which will work anywhere.)

sbc100: Actually `isPthread` is specifically available outside of the module, when MODULARIZE is used.  See https://github.com/emscripten-core/emscripten/blob/3d58456a7d37a8fb269295189ed00c554cb3f26d/tools/link.py#L2446-L2460
 
Inside the module we use `ENVIRONMENT_IS_PTHREAD`.

test/modularize_post_js.js
sbc100: Yes, I thank that might make sense, but doesn't have to be part of this change.  We can consider that as followup.

test/modularize_post_js.js
kripken: Thanks, I wasn't aware we added that to the outside scope, interesting.

kripken (approved):

PR #22539, opened 2024-09-07
[contrib.glfw3] new version 3.4.0.20240907 (bug fix)

ypujante:
Release notes:

- Fixed Meta/Super key workaround to also handle the rare (but possible) use-case of having other keys down before pressing the Meta/Super key


ypujante:
There is only 1 test failure:

```text
ERROR [61.427s]: test_offset_converter (test_browser.browser)
```

which I doubt is related to my changes

sbc100 (approved):

Issue #20636, opened 2023-11-07
INCOMING_MODULE_JS_API+Pthread+Modularize don't work


emaxx-google:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.48 (e967e20b4727956a30592165a3c1cde5c67fa0a8)
clang version 18.0.0 (https://github.com/llvm/llvm-project a54545ba6514802178cf7cf1c1dd9f7efbf3cde7)
Target: wasm32-unknown-emscripten
Thread model: posix
```

**Full link command and output with `-v` appended:**
```
$ em++ main.cc -pthread -s MODULARIZE=1 -s EXPORT_NAME=loadModule -s STRICT=1 -s INCOMING_MODULE_JS_API=[print,printErr] -o module.js -v
 /home/foo/env/emsdk/upstream/bin/clang --version
 "/home/foo/env/emsdk/upstream/bin/clang++" -target wasm32-unknown-emscripten -fignore-exceptions -fvisibility=default -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr -D__EMSCRIPTEN_SHARED_MEMORY__=1 --sysroot=/home/foo/env/emsdk/upstream/emscripten/cache/sysroot -Xclang -iwithsysroot/include/fakesdl -Xclang -iwithsysroot/include/compat -pthread -v -matomics -mbulk-memory main.cc -c -o /tmp/emscripten_temp_fkjbjtux/main_0.o
clang version 18.0.0 (https://github.com/llvm/llvm-project a54545ba6514802178cf7cf1c1dd9f7efbf3cde7)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /home/foo/env/emsdk/upstream/bin
 (in-process)
[...]

PR #19073, opened 2023-03-26
Make Node.js pthreads behavior in line with native

kleisauke:
By marking workers as weakly referenced. This corresponds to the ideal scenario as mentioned in commit a9cbf47267b41d0cd624ac83393b42f2261cfe69.

(originally based on commit e6f86a8d165cd054c2ac94e5e1bf32711ac06fd0 and 4b4faf9d2cbcdfa696da4534e4b1ab61f6b82972)

---

~~This passes the test suite locally, except for `test_pthread_abort_interrupt`.~~ Fixed with commit ca5f97c364baa60b7a4699f47916a2e3ed2afe5e.

/cc @RReverser

PR #22542, opened 2024-09-09
Fix keepRuntimeAlive in the case where EXIT_RUNTIME=0 and noExitRuntime is not referenced

sbc100:
Essentially the `keepRuntimeAlive` was relying on the `noExitRuntime` variable being set based on `EXIT_RUNTIME` but when `noExitRuntime` was absent the `EXIT_RUNTIME` settings was being ignored and the runtime was exiting even though `EXIT_RUNTIME=0` was set (the default).

Fixes: #20636


sbc100 (commented):
I updated the test to avoid the sleep.  PTAL

src/library_pthread.js
```diff
@@ -188,7 +188,7 @@ var LibraryPThread = {
       // worker pool as an unused worker.
       worker.pthread_ptr = 0;
 
-#if ENVIRONMENT_MAY_BE_NODE && PROXY_TO_PTHREAD
```
kleisauke, line 191: Could you clarify why this change is necessary? AFAIK, all threads should already be weakly referenced by this point, this should only be required for the "main" thread in `PROXY_TO_PTHREAD` mode, see:
https://github.com/emscripten-core/emscripten/pull/19073#discussion_r1172276555

test/other/test_pthread_strict.c
```diff
@@ -0,0 +1,39 @@
...
+  pthread_detach(t);
+  emscripten_outf("thread created, returning from main");
+
+  // Make sure the thread output doesn't appear until after the above message
+  pthread_mutex_lock(&ready_lock);
+  ready = true;
+  pthread_cond_signal(&ready_cond);
+  pthread_mutex_unlock(&ready_lock);
+
+  // This needed under node to make sure the process doesn't exit before the
```
kripken, line 35:
```suggestion
  // This is needed under node to make sure the process doesn't exit before the
```

test/other/test_pthread_strict.c
```diff
@@ -0,0 +1,39 @@
...
+  // thread is done.
+  _emscripten_thread_set_strongref(t);
```
kripken, line 37: Do we expect users to need to do this?

sbc100: No, this is just do that node doesn't exit while the background thread is running.  I could instead use a setTimeout loop instead I guess?

kripken: I guess I was unaware that this was an issue. If the main thread exits and a background thread remains, then we need to keep the main thread alive in a special way, I guess, that is the underlying issue?

sbc100: Right, its a little confusing.   Under normal circumstance, node will exit when the main function exits.  Running threads are not enough to keep node running.   The runtime has not exited in this case, the module is still available to be used, but node needs a reason to stay alive otherwise will exit the process.   

There is no equilient to this one the web, and this program will basically run forever on the web even without `_emscripten_thread_set_strongref`.. because there is no matching concept of "event loop is empty so exit" on the web.

I agree this is quite ugly here, so i might try to find a better way to keep node alive here.

kleisauke: I think the canonical way is to remove the `pthread_detach()` call and replace `_emscripten_thread_set_strongref()` with `pthread_join()`, unless I'm missing something (perhaps this test case should not block the main thread?).

sbc100: Hmm..there must be a reason I didn't do it that way... i'll need to go figure out why :)

sbc100: Ah yes, the point of this test was to verify that when main returns it doesn't kill all the running threads.  If the thread was already joined there would be no point.

I re-wrote the test without threads at all PTAL

test/other/test_no_exit_runtime_strict.c
```diff
@@ -0,0 +1,22 @@
...
+EM_JS(void, timeout_func, (), {
+  console.log("timeout_func: keepRuntimeAlive() ->", keepRuntimeAlive());
+  // once this timeout is done the node process should exit with 0
+});
+
+int main() {
+  int keep_alive = EM_ASM_INT({
+    setTimeout(timeout_func);
+    return keepRuntimeAlive();
+  }, emscripten_force_exit);
```
kripken, line 18: Why is a parameter passed in here?

test/other/test_no_exit_runtime_strict.c
```diff
@@ -0,0 +1,22 @@
+#include <assert.h>
+#include <emscripten/emscripten.h>
+#include <emscripten/em_js.h>
+#include <stdio.h>
+
+// Verify that `keepRuntimeAlive()` always returns true by default (i.e. when
+// EXIT_RUNTIME=0).
+
+EM_JS(void, timeout_func, (), {
+  console.log("timeout_func: keepRuntimeAlive() ->", keepRuntimeAlive());
```
kripken, line 10: Do you not need to ensure that keepRuntimeAlive is available in JS?

src/library_pthread.js
sbc100: Reverted this part

test/other/test_no_exit_runtime_strict.c
sbc100: Hmm.. I can't remember now.  Reverted.

kripken (approved):

PR #22543, opened 2024-09-09
Reset `noExitRuntime` every time a new pthread starts

sbc100:
Previously we were only setting this value when a worker was first created.  This means that changes to this value could persist between pthreads.


src/library_pthread.js
```diff
@@ -1066,6 +1051,16 @@ var LibraryPThread = {
     // `runtimeKeepaliveCounter` to zero. Reset it now so the new thread won't
     // be affected.
     runtimeKeepaliveCounter = 0;
+
+#if isSymbolNeeded('$noExitRuntime')
+    // Same for noExitRuntime.  The default for pthreads should always be false
+    // otherwise pthreads would never complete and attempts to pthread_join to
+    // them would block forever.
+    // pthreads can still choose to set `noExitRuntime` explicitly, or
```
kripken, line 1059:
```suggestion
    // Pthreads can still choose to set `noExitRuntime` explicitly, or
```
Unless there is a convention to never capitalize `pthread`?

kripken (approved):
Is this easily testable perhaps?

sbc100:
> Is this easily testable perhaps?

When trying to think about how to test this figured out that it might not really be the cause of any issues.  This is because its very hard to actually exit a thread with `noExitRuntime` set to true.  This is because `emscripten_force_exit` will reset this global.

In any case I still think this is nice cleanup as it removed a method and groups the noExit state setting into a single place.

src/library_pthread.js
sbc100: I tend not a capitalize it myself.

kripken (approved):
Agreed to both of the last comments.

PR #22546, opened 2024-09-09
[test] Cleanup gcc pthread operations tests

sbc100:
- Tabs to spaces
- And some print status

### DIFF
--- a/test/pthread/test_pthread_gcc_64bit_atomic_fetch_and_op.c
+++ b/test/pthread/test_pthread_gcc_64bit_atomic_fetch_and_op.c
@@ -16,161 +16,154 @@
 
 #define T uint64_t
 
-void *thread_fetch_and_add(void *arg)
-{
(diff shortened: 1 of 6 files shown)


sbc100:
Because I did a bunch of formatting as well as re-name git doesn't do a good job of showing what actually changed.  If you would prefer I can split this into 2 PRs.  The first a pure rename, and the second the formatting changes.

sbc100:
I split into 2 commits so git should show the changes better now.

kripken (approved):
Thanks for the separate commits, much easier to read that way!

sbc100:
I think I will land this manually rather then squashing, so that both commits are preserved.

kripken:
Squashing is ok here, I think. In the rare event we end up bisecting to this, the original commits are in the PR. But, either way.

sbc100:
I landed them separately.

PR #22547, opened 2024-09-10
Enable SINGLE_FILE + PTHREADS html output

sbc100:
In SINGLE_FILE mode, the JS is inlined directly into the generated HTML.  However for pthreads builds we need to be able to create workers that use the same inlined JS. To enable this we encode the JS file into a data URL that can be used for the script `src` attribute.  The pthread worker can then use this `src`.

Prior to this change using SINGLE_FILE + PTHREADS + html output would simply fail.

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -3693,11 +3693,17 @@ Module["preRun"] = () => {
     '': ([],),
     'O3': (['-O3'],),
     'minimal_runtime': (['-sMINIMAL_RUNTIME'],),
+    'single_file': (['-sSINGLE_FILE'],),
   })
   def test_pthread_create(self, args):
     self.btest_exit('pthread/test_pthread_create.c',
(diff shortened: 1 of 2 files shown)


juj:
My god, what a monstrosity. LGTM.

juj (approved):

sbc100:
> My god, what a monstrosity. LGTM.

The power of the web!  :)    I also find this kind of horrifying by I guess we have users that really like shipping just a single file?   Do you guys use this feature?

juj:
I use SINGLE_FILE in my personal projects and often when bundling test cases to bug trackers. At Unity we did use SINGLE_FILE in a past project, but my understanding is that it is no longer active. The main Unity engine in general does not use SINGLE_FILE, though we do get repeated requests to add support for it, due to ads ecosystem MRAID architecture, which is a "single file" deployment.

kripken (approved):
Nice hack!

PR #22556, opened 2024-09-10
Remove unused argument from emscripten_notify_mailbox_postmessage. NFC

sbc100:
This is an internal function so changing its signature should not be a user-visible change.


system/lib/pthread/thread_mailbox.h
```diff
@@ -37,5 +37,4 @@ void _emscripten_thread_mailbox_shutdown(pthread_t thread);
 // mailbox when it returns to its event loop. Pass in the current thread and
 // main thread ids to minimize calls back into Wasm.
 void _emscripten_notify_mailbox_postmessage(pthread_t target_thread,
-                                            pthread_t curr_thread,
-                                            pthread_t main_thread);
+                                            pthread_t curr_thread);
```
tlively, line 40: Looks like the comment above needs to be updated as well.

tlively (approved):

Issue #22553, opened 2024-09-10
lib noderawfs now requires FS.init()


walkingeyerobot:
Previously I was able to call `FS.chdir()` when using noderawfs without first calling `FS.init()`, but now I have to call `FS.init()` first. I think that just means the docs need to be updated here because `FS.init()` does more than just set up standard I/O devices in this case: https://emscripten.org/docs/api_reference/Filesystem-API.html?highlight=filesystem#setting-up-standard-i-o-devices

sbc100:
IIUC `FS.init()` should automatically be called when the module loads.   Did that somehow change maybe?   Can you share more details and perhaps we can construct a test case.

walkingeyerobot:
It's entirely possible (even likely) that I'm holding it wrong. I'm running some JS before the emscripten JS that looks approximately like this:
```js
var Module = typeof Module !== 'undefined' ? Module : {
  arguments: [],
  preRun: [function() {
    ENV = typeof ENV !== "undefined" ? ENV : {};
    ENV.FOO = 'const'; // lots of lines like this
    FS.chdir('/usr/local/bar');
  }]
};
```

This was working fine under 3.1.63 (and for a loooong time before that) and has broken under 3.1.65. If I run `FS.init();` before I call `FS.chdir` it works as expected.

I'm not going through `--pre-js`; maybe that's the problem?

sbc100:
That looks like something that we should continue to support.  I'll see if I can repro and fix.

sbc100:
What error message do you see when you don't call `FS.init()`.    Can't seem to replicate this using a trivial example:

```
$ cat test.js 
Module.preRun = [function() {
    FS.mkdir('/foo');
    FS.chdir('/foo');
}];
$ emcc --pre-js test.js ~/test/hello.c -sFORCE_FILESYSTEM
$ node ./a.out.js 
Hello, world!
```

kleisauke:
This looks like a oversight on my part. Commit c614fa57ea8048a34341891dbd6f99fb78b6869a introduced the regression, which was later reverted by commit 659f01dd3c0362806a5cffed3a5173862c7ef024, while keeping the filesystem wrapping in `FS.init()` to minimize the diff.

I just opened PR #22559 for this.

walkingeyerobot:
I'll also add that I'm on node v18.17.0 and I'm unable to change the version of node I'm on.

sbc100:
I think #22559 will fix this issue.

PR #22559, opened 2024-09-11
[NODERAWFS] Move filesystem wrapping to `__postset`

kleisauke:
Resolves: #22553.

### DIFF
--- a/src/library_noderawfs.js
+++ b/src/library_noderawfs.js
@@ -12,33 +12,26 @@ addToLibrary({
     }
     // Use this to reference our in-memory filesystem
     var VFS = Object.assign({}, FS);
-    // Override the init function with our own
-    FS.init = NODERAWFS.init;`,
-  $NODERAWFS: {
-    init() {
-      var _wrapNodeError = function(func) {
-        return function(...args) {
-          try {
-            return func(...args)
-          } catch (e) {
-            if (e.code) {
-              throw new FS.ErrnoError(ERRNO_CODES[e.code]);
-            }
-            throw e;
+    var _wrapNodeError = function(func) {
+      return function(...args) {
+        try {
+          return func(...args)
+        } catch (e) {
+          if (e.code) {
+            throw new FS.ErrnoError(ERRNO_CODES[e.code]);
           }
+          throw e;
         }
-      };
-
-      // Wrap the whole in-memory filesystem API with
-      // our Node.js based functions
-      for (var _key in NODERAWFS) {
-        /** @suppress {partialAlias} */
-        FS[_key] = _wrapNodeError(NODERAWFS[_key]);
       }
+    };
 
-      // Setup the stdin, stdout and stderr devices
-      FS.createStandardStreams();
-    },
+    // Wrap the whole in-memory filesystem API with
+    // our Node.js based functions
+    for (var _key in NODERAWFS) {
+      /** @suppress {partialAlias} */
+      FS[_key] = _wrapNodeError(NODERAWFS[_key]);
+    }`,
+  $NODERAWFS: {
     lookup(parent, name) {
 #if ASSERTIONS
       assert(parent)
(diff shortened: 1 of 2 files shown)


kleisauke:
Diff is best viewable with the `Hide whitespace changes` option.

test/test_other.py
```diff
@@ -9730,6 +9730,16 @@ def test_noderawfs_access_abspath(self):
     self.run_process([EMCC, 'access.c', '-sNODERAWFS'])
     self.run_js('a.out.js', args=[os.path.abspath('foo')])
 
+  def test_noderawfs_readfile_prerun(self):
+    create_file('foo', 'bar')
+    create_file('main.c', r'''
+      int main() { return 0; }
+    ''')
```
sbc100, line 9737: Maybe just use `test_file('hello_world.c'`)` instead of creating a new empty main.c here?

src/library_noderawfs.js
```diff
@@ -10,35 +10,27 @@ addToLibrary({
...
-      var _wrapNodeError = function(func) {
-        return function(...args) {
-          try {
-            return func(...args)
-          } catch (e) {
-            if (e.code) {
-              throw new FS.ErrnoError(ERRNO_CODES[e.code]);
-            }
-            throw e;
+    var _wrapNodeError = function(func) {
```
sbc100, line 13: It strange, the github UI won't let me expand the context to see the surrounding line.   Is there something unusual about the way your created this PR?

sbc100 (approved):

src/library_noderawfs.js
sbc100: Oh!   That feature looks like it doesn't work in "ignore whitespace" mode.

sbc100: Umm now its working.   Ignore me.

test/test_other.py
kleisauke: Ah, good point! I initially thought the `expected_output` argument of `do_runf()` always did an identical assert, which could conflict with the `printf()` in that test file.  However, that only happens if you explicitly pass `assert_identical=True`.

Fixed with commit b5801e362949a5c5f5e16c7506942819c435d92e.

test/test_other.py
```diff
@@ -9730,6 +9730,13 @@ def test_noderawfs_access_abspath(self):
     self.run_process([EMCC, 'access.c', '-sNODERAWFS'])
     self.run_js('a.out.js', args=[os.path.abspath('foo')])
 
+  def test_noderawfs_readfile_prerun(self):
+    create_file('foo', 'bar')
+    create_file('pre.js', f'''
+      Module.preRun = () => console.log(FS.readFile('{os.path.abspath('foo')}', {{ encoding: 'utf8' }}));
```
sbc100, line 9736: Do you need `{ encoding: 'utf8' }` here?

Also can you just use `FS.readFile('foo')`?

sbc100 (approved):

test/test_other.py
kleisauke: `{ encoding: 'utf8' }` is needed, otherwise it would print: `Uint8Array(3) [ 98, 97, 114 ]`.

`FS.readFile('foo')` is done with commit 57b3cc096e0a13f0edd779a7fcf12c8a8629d1c4,

sbc100 (approved):

test/test_other.py
```diff
@@ -9730,6 +9730,13 @@ def test_noderawfs_access_abspath(self):
     self.run_process([EMCC, 'access.c', '-sNODERAWFS'])
     self.run_js('a.out.js', args=[os.path.abspath('foo')])
 
+  def test_noderawfs_readfile_prerun(self):
+    create_file('foo', 'bar')
+    create_file('pre.js', r'''
+      Module.preRun = () => console.log(FS.readFile('foo', { encoding: 'utf8' }));
+    ''')
```
sbc100, line 9737: Oh, we also have utility for this: `self.add_pre_run("console.log(FS.readFile('foo', { encoding: 'utf8' })")`

That will also add `'--pre-js=pre.js'` for you.

sbc100 (approved):

PR #22557, opened 2024-09-10
Add some sanity checks when using `__i53abi` decorator

sbc100:
See #21335

PR #22561, opened 2024-09-11
[WasmFS] Adds support for large (>4GB) files in OPFS using __i53abi

goldwaving:
Currently wasmfs OPFS is limited to 2GB files or smaller.  These changes allow working with much larger files by using the __i53abi decorator.

Note that individual reads/writes are still limited to uint32_t block sizes.


src/library_wasmfs_opfs.js
```diff
@@ -363,6 +367,8 @@ addToLibrary({
     wasmfsOPFSProxyFinish(ctx);
   },
 
+  _wasmfs_opfs_write_access__i53abi: true,
+  _wasmfs_opfs_write_access__sig: 'iipij',
```
sbc100, line 371: None of these sigs should be needed since I believe they are already stored in src/library_sigs.js

system/lib/wasmfs/backends/opfs_backend.h
```diff
@@ -67,28 +67,27 @@ void _wasmfs_opfs_free_directory(int dir_id);
...
+                              off_t pos);
 
 // Get the size via an AccessHandle.
 void _wasmfs_opfs_get_size_access(em_proxying_ctx* ctx,
                                   int access_id,
                                   off_t* size);
 
-// TODO: return 64-byte off_t.
-uint32_t _wasmfs_opfs_get_size_blob(int blob_id);
+void _wasmfs_opfs_get_size_blob(int blob_id, off_t* size);
```
sbc100, line 90: Why can't `off_t` be used as a return type here?

sbc100 (approved):
lgtm % some comments.

Would it be feasible to add a test for this?

src/library_wasmfs_opfs.js
goldwaving: The sigs need to be updated.  Not sure if those changes have been added to this request.

system/lib/wasmfs/backends/opfs_backend.h
goldwaving: Maybe it can.  Not sure if/how  i64 values are returned, so I just used the same method as the other `get_size` function to be safe.

src/library_wasmfs_opfs.js
sbc100: Yes, changes to function signatures should be accompanied by the change in `library_sigs.js`.   I normally rebuild using `./test/runner other.test_gen_sig_info --rebase`.

system/lib/wasmfs/backends/opfs_backend.h
sbc100: I think if you mark this function as `i53abi` and `j` is the return type it should "just work".

goldwaving: Good to know.  I'll keep that in mind for future updates.

tlively (approved):
Thanks for working on this! LGTM if Sam is happy.

system/lib/wasmfs/backends/opfs_backend.h
sbc100: Can we make this part of this change?  No need to the out parameter then a simple return type will do (IMHO).

sbc100: Can we make this part of this change?  No need to the out parameter then a simple return type will do (IMHO).

goldwaving: I've change it to return the size directly.  Seems to work fine.

src/library_wasmfs_opfs.js
goldwaving: I've removed the sigs.  Just need to make sure that they are updated.

sbc100: `__sig` attributes are pre-calculated for builtin JS libraries (via `tools/maint/gen_sig_info.py`).    For external libraries you need to specifiy the `__sig` manually, alongside the function.

I made a change recently that will make it impossible to forget to add the `__sig` when using `__i53abi`: #22557

sbc100 (approved):

sbc100:
lgtm, but that do we think about testing this?

kripken:
I am not sure we can test this well. We'd need a file larger than 4GB for that, which will strain CI limits. Even creating such a file on disk and using some kind of streaming read might be challenging (and I don't think we have a backend that does such streaming reads in a guaranteed manner). So I'd be ok landing this without specific testing for 4GB+ files.

PR #22563, opened 2024-09-11
[embind] Fix TS generation when using AOT JS.

brendandahl:
Some AOT JS was ending up in the TS definition file because the setting to turn AOT JS was wrong.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -3364,11 +3364,10 @@ More info: https://emscripten.org
     'with_jsgen': [['-sEMBIND_AOT']]
   })
   def test_embind_tsgen(self, opts):
-    self.emcc_args += opts
     # Check that TypeScript generation works and that the program is runs as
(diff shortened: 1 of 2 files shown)


test/test_other.py
```diff
@@ -3359,7 +3359,12 @@ def test_jspi_add_function(self):
       '-Wno-experimental']
     self.do_runf('other/test_jspi_add_function.c', 'done')
 
-  def test_embind_tsgen(self):
+  @parameterized({
+    '': [[]],
+    'with_jsgen': [['-sEMBIND_AOT']]
+  })
+  def test_embind_tsgen(self, opts):
+    self.emcc_args += opts
```
sbc100, line 3367: How about `emcc_args=['-lembind', '--emit-tsd', 'embind_tsgen.d.ts'] ) opts` below instead of add a new line here?

PR #22565, opened 2024-09-13
Support preallocated pthread & wasm worker from blob

ifiddynine:
This amends existing `mainScriptUrlOrBlob` logic to make sure that we can preload JS for pthread and wasm workers reliably.

On the pthread side, this amends PTHREAD_POOL_SIZE warmup logic to allow waiting for a promise on top of existing logic in `PThread.allocateUnusedWorker`. Otherwise, there would be a race which would probably mean the JS file would be loaded several times anyways.

On the wasm worker side, simply fixes the ability to provide a blob - previously it would only accept a javascript string. We use WASM_WORKERS=2 to keep this working, mixing wasm workers + pthread.

See below for example of how it's used. I considered just refactoring the scriptUrlOrBlobPromise logic to automatically cache all worker startups, but figured that would be harder to get integrated since it would break backwards compatibility & tests that use it in emscripten main.

```
Module['mainScriptUrlOrBlobPromise'] = fetch('myscript.js', { priority: 'high' })
    .then(response => response.text())
    .then(workerScript => {
    Module['mainScriptUrlOrBlob'] = new Blob([workerScript], { type: 'application/javascript' });
});
```

### DIFF
--- a/src/library_pthread.js
+++ b/src/library_pthread.js
@@ -98,10 +98,20 @@ var LibraryPThread = {
     },
     initMainThread() {
 #if PTHREAD_POOL_SIZE
-      var pthreadPoolSize = {{{ PTHREAD_POOL_SIZE }}};
-      // Start loading up the Worker pool, if requested.
-      while (pthreadPoolSize--) {
-        PThread.allocateUnusedWorker();
+      preallocateWorkers = () => {
+        var pthreadPoolSize = {{{ PTHREAD_POOL_SIZE }}};
+        // Start loading up the Worker pool, if requested.
+        while (pthreadPoolSize--) {
+         PThread.allocateUnusedWorker();
+        }
+      };
+#if expectToReceiveOnModule('mainScriptUrlOrBlobPromise')
+      if (Module['mainScriptUrlOrBlobPromise']) {
+        Module['mainScriptUrlOrBlobPromise'].then(preallocateWorkers);
+      } else 
+#endif
+      {
+        preallocateWorkers();
       }
 #endif
 #if !MINIMAL_RUNTIME
--- a/src/library_wasm_worker.js
+++ b/src/library_wasm_worker.js
@@ -182,7 +182,7 @@ if (ENVIRONMENT_IS_WASM_WORKER
       'mem': wasmMemory,
 #else
       'wasm': wasmModule,
-      'js': Module['mainScriptUrlOrBlob'] || _scriptName,
+      'js': URL.createObjectURL(Module["mainScriptUrlOrBlob"]) || _scriptName,
       'wasmMemory': wasmMemory,
 #endif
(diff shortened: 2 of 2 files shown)


src/library_pthread.js
```diff
@@ -98,10 +98,20 @@ var LibraryPThread = {
...
+        }
+      };
+#if expectToReceiveOnModule('mainScriptUrlOrBlobPromise')
+      if (Module['mainScriptUrlOrBlobPromise']) {
+        Module['mainScriptUrlOrBlobPromise'].then(preallocateWorkers);
+      } else 
+#endif
+      {
+        preallocateWorkers();
       }
```
sbc100, line 115: How about instead updating `init()` and leaving `initMainThread()` unchanged.

e.g.:

```
#if expectToReceiveOnModule('mainScriptUrlOrBlobPromise')
if (Module['mainScriptUrlOrBlobPromise']) Module['mainScriptUrlOrBlobPromise'].then(PThread.initMainThread());
else
#endif
PThread.initMainThread()
```

This would avoid the new/extra closure.

src/library_wasm_worker.js
```diff
@@ -182,7 +182,7 @@ if (ENVIRONMENT_IS_WASM_WORKER
       'mem': wasmMemory,
 #else
       'wasm': wasmModule,
-      'js': Module['mainScriptUrlOrBlob'] || _scriptName,
+      'js': URL.createObjectURL(Module["mainScriptUrlOrBlob"]) || _scriptName,
```
sbc100, line 185: So this is fix for the case when `mainScriptUrlOrBlob` is a blob?   Is this because blobs cannot be serialized but object URLs can?

Can we add a test for this?  And perhaps make it a separate PR and it seems somewhat orthogonal.

src/library_pthread.js
```diff
@@ -98,10 +98,20 @@ var LibraryPThread = {
...
-      while (pthreadPoolSize--) {
-        PThread.allocateUnusedWorker();
+      preallocateWorkers = () => {
+        var pthreadPoolSize = {{{ PTHREAD_POOL_SIZE }}};
+        // Start loading up the Worker pool, if requested.
+        while (pthreadPoolSize--) {
+         PThread.allocateUnusedWorker();
+        }
+      };
+#if expectToReceiveOnModule('mainScriptUrlOrBlobPromise')
```
sbc100, line 108: What is `mainScriptUrlOrBlobPromise`?   I don't see that anywhere in the existing source code?  Is the idea that someone could add this to `-sINCOMING_JS_MODULE_API`?    If so I wonder if we should document it somewhere.

If that creation of the main thread needs to be delayed until the `mainScriptUrlOrBlob` is ready can't this be done via existing mechanisms?  For example, any of these approaches:

1. Using MODULARIZE and delaying the instantiation of the module.
2. Delaying the import of the emscripten-generated script
3. Using `addRunDependency` ?  (this one might not work since I'm not sure it blocks `initMainThread`.

PR #22567, opened 2024-09-13
[embind] Support annotating return values as nonnull.

brendandahl:
Embind's subclass `implement` methods were generated as returning `Class | null` after the changes to pointer types in #22184. This could be considered a regression as the implement method would never return null.

Previously, we had special handling so constructors were marked as nonnull so in the TS definitions we didn't add `| null`. I've generalized this approach to work for all function bindings so they can now use a `nonnull<ret_val>` policy too avoid the `| null`.


kripken (commented):
I am not quite sure how this is meant to work. Are there docs for it somewhere?

kripken:
(lgtm otherwise!)

kripken:
(also maybe mention in the changelog? but that can be with the docs - if docs are in a later PR, can wait on that)

brendandahl:
I added some docs and a changelog entry.

kripken (approved):

Issue #22571, opened 2024-09-15
Backport freetype patch to fix TYPEOF macro


jeroen:
Old versions of freetype export a macro `TYPEOF()` which [conflicts with a macro of the same name in client libraries](https://savannah.nongnu.org/bugs/?45376). 

The problem was fixed in freetype 9 years ago by [renaming TYPEOF to FT_TYPEOF](https://github.com/freetype/freetype/commit/5931268eecaeda3e05580bdc8885348fecc43fa8) and was ported to the master branch of https://github.com/emscripten-ports/FreeType/commit/5931268eec, but sadly the git tag `emscripten-ports/freetype@version_1` that the build script is pulling from, does not not include this commit. 

Would it be possible to backport [5931268.patch](https://github.com/emscripten-ports/FreeType/commit/5931268eec.patch) or just remove this macro from our headers below?  I think this macro is used internally in freetype, or perhaps it is not used at all, it should not break other libs.

https://github.com/emscripten-core/emscripten/blob/50e0b0785ad5f9adc9bd082451d5cdbf5c4c5229/tools/ports/freetype.py#L487-L494

PR #22572, opened 2024-09-15
Fix macro conflict for USE_FREETYPE=1

jeroen:
Basically backport of https://github.com/freetype/freetype/commit/5931268 / https://github.com/emscripten-ports/FreeType/commit/5931268eec

We want to avoid publicly exporting the `TYPEOF` macro which conflicts with other libraries. . IIUC the header here is the one that gets included in the compiler with `-s USE_FREETYPE=1` ? This should fix https://github.com/emscripten-core/emscripten/issues/22571

### DIFF
--- a/tools/ports/freetype.py
+++ b/tools/ports/freetype.py
@@ -488,9 +488,9 @@ FT_BEGIN_HEADER
 #if ( __GNUC__ >= 2                         || \
       defined( __IBM__TYPEOF__ )            || \
       ( __SUNPRO_C >= 0x5110 && !__STDC__ ) )
-#define TYPEOF( type )  (__typeof__ (type))
+#define FT_TYPEOF( type )  (__typeof__ (type))
 #else
-#define TYPEOF( type )  /* empty */
+#define FT_TYPEOF( type )  /* empty */
 #endif
 
 


tools/ports/freetype.py
```diff
@@ -488,9 +488,9 @@ def show():
 #if ( __GNUC__ >= 2                         || \
       defined( __IBM__TYPEOF__ )            || \
       ( __SUNPRO_C >= 0x5110 && !__STDC__ ) )
-#define TYPEOF( type )  (__typeof__ (type))
+#define FT_TYPEOF( type )  (__typeof__ (type))
 #else
-#define TYPEOF( type )  /* empty */
+#define FT_TYPEOF( type )  /* empty */
```
kripken, line 493: This changes the definition - do the uses not need to be updated too? I'm just confused how this works.

jeroen: Yes maybe you are right. I assumed this file was just the shipped header, but it is also used during the build.

Given that you own the repo, a much easier solution is to cherry-pick https://github.com/emscripten-ports/FreeType/commit/5931268eec into the git tag `version_1` that we pull in here. I am not why you ended up archiving that repo, but just getting this one commit in would be a big improvement for us (and harmless for others).

kripken: Hmm, I'm not sure why that was archived. @sbc100 did we plan to use upstream perhaps?

sbc100: I think we should look into upgrading our version of freetype.   I'll take a look.

sbc100: Should be done in #22585

PR #22577, opened 2024-09-17
Fix emrun sporadic exit hang on Windows

rghosh0:
Fixes https://github.com/emscripten-core/emscripten/issues/22578 ("Windows: emrun.py does not exit as expected")

We set the server's socket itself to non-blocking before shutdown. This way the threads can still be joined and the ordinary and cleaner sys.exit(...) exit function can be used. Here is a diff that we propose to be adopted (which makes this behaviour a program option for emrun).

### DIFF
--- a/emrun.py
+++ b/emrun.py
@@ -704,6 +704,8 @@ class HTTPHandler(SimpleHTTPRequestHandler):
         if not emrun_options.serve_after_exit:
           page_exit_code = int(data[6:])
           logv('Web page has quit with a call to exit() with return code ' + str(page_exit_code) + '. Shutting down web server. Pass --serve-after-exit to keep serving even after the page terminates with exit().')
+          if emrun_options.force_exit:
+            self.server.socket.setblocking(False)
           self.server.shutdown()
           return
       else:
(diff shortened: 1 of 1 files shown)


emrun.py
```diff
@@ -704,6 +704,8 @@ def do_POST(self):
         if not emrun_options.serve_after_exit:
           page_exit_code = int(data[6:])
           logv('Web page has quit with a call to exit() with return code ' + str(page_exit_code) + '. Shutting down web server. Pass --serve-after-exit to keep serving even after the page terminates with exit().')
+          if emrun_options.force_exit:
+            self.server.socket.setblocking(False)
```
kripken, line 708: Is there a downside to always doing this? I mean, instead of adding an option, to always set nonblocking after the page exits?

rghosh0: From our side there is no downside specific to this change.  We just thought that it might be categorically better received if it were presented as an option, but there is no downside to always doing this before shutdown.

kripken: Ok, then I think it's better to avoid adding an option actually. That way it will be fully tested.

kripken (approved):
Thanks! Looks good.

@sbc100 do you have any concerns here?

kripken:
Yeah, I don't fully understand the deadlock either, but this seems harmless to do during shutdown anyhow, so since it fixes issues in practice on windows, let's land it.

Thanks @rghosh0 !

PR #22581, opened 2024-09-18
[SDL_mixer] Fix Mix_PlayChannel audio 'leaking' and fix onended callback to correctly update audio status

caiiiycuk:
### Mix_PlayChannel audio 'leaking'
Hi, I found a bug while using builtin SDL_mixer 1.2. 
You can play many sounds on same channel in case if you do not manuall halt it before.

Reagrding to [documentation](https://wiki.libsdl.org/SDL2_mixer/Mix_PlayChannel) of SDL:
> If a specific channel was requested, and there is a chunk already playing there, that chunk will be halted and the new chunk will take its place.

So, when you call
```
Mix_PlayChannel(channel, sound, loops);
```

It shoud internally halt the channel. Current emscripten implemntation is wrong, the previously runned chunk on that channel continue to play.

You can easily reporoduce this with code:

```c++
/* 
compile with:
emcc main.cpp -o main.html --preload-file "./@"
*/   
#include <SDL/SDL_mixer.h>
#include <chrono>
#include <emscripten.h>

Mix_Chunk* chunk = nullptr;
uint64_t lastRun = 0;

void loop() {
    using namespace std::chrono;
    uint64_t now = duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count();
    if (now - lastRun > 5000) {
        Mix_PlayChannel(0, chunk, -1);
        lastRun = now;
    }
}

int main(int argc, char** argv) {
    if ((Mix_Init(MIX_INIT_OGG) & MIX_INIT_OGG) != MIX_INIT_OGG) {
      printf("ERR! Mix_Init: Failed to init with required flags support!\n");
      printf("ERR! Mix_Init: %s\n", Mix_GetError());
      return -1;
    }

    if (Mix_OpenAudio(MIX_DEFAULT_FREQUENCY, MIX_DEFAULT_FORMAT, 1, 1024) == -1) {
      printf("ERR! Mix_OpenAudio: %s\n", Mix_GetError());
      return -1;
    }

    Mix_AllocateChannels(1);
    chunk = Mix_LoadWAV("/music.ogg");
    emscripten_set_main_loop(loop, 0, false);
    return 0;
}
```

If you compile and run it then you will hear infinite looping of music.ogg, and every 5 sec new one is started.

I fixed this by calling Mix_HaltChannel if channel is playing audio.

### fix onended callback to correctly update audio status

Found another issue in SDL_mixer implementation. Current implementation does not detect playing status of sounds correctly (`Mix_Playing`, `Mix_PlayingMusic`). This happens because in on `onended` callback comparsion with `this` works only for case when audio created by cloning audio node, otherwise comparsion should be against audio.webAudioNode.


sbc100 (approved):
LGTM.

I wonder if we could add test for this?  Perhaps updating an existed test_interactive.py test?

caiiiycuk:
Thanks, will look into it.

caiiiycuk:
1. Added test for Mix_PlayChannel
2. Found another issue in SDL_mixer implementation. Current implementation does not detect playing status of sounds correctly (Mix_Playing, Mix_PlayingMusic). This happens because in on 'onended' callback comparsion with this works only for case when audio created by cloning audio node, otherwise comparsion should be again audio.webAudioNode.
3. Fixed (2), added test for (2)
4. Description updated

@sbc100 please review, thanks!

src/library_sdl.js
```diff
@@ -2860,9 +2861,12 @@ var LibrarySDL = {
       audio.frequency = info.audio.frequency;
     }
     audio['onended'] = function() { // TODO: cache these
-      if (channelInfo.audio == this) { channelInfo.audio.paused = true; channelInfo.audio = null; }
+      if (channelInfo.audio === this || channelInfo.audio.webAudioNode === this) { channelInfo.audio.paused = true; channelInfo.audio = null; }
```
sbc100, line 2864: Can you  split the body over multiple lines here?

src/library_sdl.js
```diff
@@ -2942,7 +2946,7 @@ var LibrarySDL = {
     } else if (info.audio) { // Play via the <audio> element
       audio = info.audio;
     }
-    audio['onended'] = function() { if (SDL.music.audio == this) _Mix_HaltMusic(); } // will send callback
+    audio['onended'] = function() { if (SDL.music.audio === this || SDL.music.audio.webAudioNode === this) _Mix_HaltMusic(); } // will send callback
```
sbc100, line 2949: Same here

test/browser/test_sdl_audio_mix_channels_halt.cpp
```diff
@@ -0,0 +1,44 @@
...
+    chunk = Mix_LoadWAV("/the_entertainer.ogg");
+    if (!chunk) {
+      printf("ERR! Mix_LoadWAV: %s\n", Mix_GetError());
+      return -1;
+    }
+
+    printf("The sound should reset to start every 300ms, if you hear overapplied music then the test has failed.\n");
+    emscripten_set_main_loop(loop, 0, 0);
+    return 0;
+}
```
sbc100, line 44: Missing newline here

test/browser/test_sdl_audio_mix_playing.cpp
```diff
@@ -0,0 +1,45 @@
...
+    }
+    Mix_PlayChannel(0, chunk, 0);
+    if (!Mix_Playing(0)) {
+      printf("ERR! Channel is not playing\n");
+      return -1;
+    }
+    started = duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count();
+    emscripten_set_main_loop(loop, 0, 0);
+    return 0;
+}
```
sbc100, line 45: And here

test/test_interactive.py
```diff
@@ -82,6 +82,24 @@ def test_sdl_audio_mix_channels(self, args):
 
     self.btest_exit('test_sdl_audio_mix_channels.c', args=['-O2', '--minify=0', '--preload-file', 'sound.ogg'] + args)
 
+  @parameterized({
+    '': ([],),
+    'wasmfs': (['-sWASMFS'],),
+  })
+  def test_sdl_audio_mix_channels_halt(self, args):
```
sbc100, line 89: I'm not sure we need these new test to run in wasmfs mode.

test/test_interactive.py
```diff
@@ -82,6 +82,24 @@ def test_sdl_audio_mix_channels(self, args):
...
+    shutil.copyfile(test_file('sounds/the_entertainer.ogg'), self.in_dir('the_entertainer.ogg'))
```
sbc100, line 90: I think you can drop the `self.in_dir` here an below.

test/browser/test_sdl_audio_mix_channels_halt.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+void loop() {
+    uint64_t now = duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count();
+    if (now - lastRun > 300) {
+        Mix_PlayChannel(0, chunk, -1);
+        lastRun = now;
+    }
+    if (now - started > 3000) {
+        emscripten_force_exit(0);
+    }
```
sbc100, line 19: Can you use 2 spaces for indentation in these new test files.

src/library_sdl.js
sbc100: Do you think its a bug there `this` can be `channelInfo.audio` in once case but `channelInfo.audio.webAudioNode` in another case?   Do you know why we have both of those modes?

caiiiycuk:
Updated, squashed.

> Do you think its a bug there this can be channelInfo.audio in once case but channelInfo.audio.webAudioNode in another case? Do you know why we have both of those modes?

I don't think it's a bug. As far as I understand, the implementation has 2 ways of loading sounds, first is [preloaded audio](https://emscripten.org/docs/porting/files/packaging_files.html#preloading-files) it preloads sounds into audio elements, second option is using webAudio. If webAudio is not available, the implementation will fallback to audio elements anyway. The implementation tries to mimic webAudio in an audio-like way. So it's valid to have 2 cases there. Honestly, I never used 1st case (audio elements).

--
Do you know why `check_clean.py` is failing and what it is?

>Check for clean checkout.  This is run after tests during CI to ensure we are not polluting the source checkout.

Should I do some extra steps to clean workdir?

test/test_interactive.py
```diff
@@ -82,6 +82,22 @@ def test_sdl_audio_mix_channels(self, args):
 
     self.btest_exit('test_sdl_audio_mix_channels.c', args=['-O2', '--minify=0', '--preload-file', 'sound.ogg'] + args)
 
+  @parameterized({
+    '': ([],),
+  })
+  def test_sdl_audio_mix_channels_halt(self, args):
+    shutil.copyfile(test_file('sounds/the_entertainer.ogg'), 'the_entertainer.ogg')
```
sbc100, line 89: Can you replace this line with `shutil.copy(test_file('sounds/the_entertainer.ogg'), '.')`?

test/test_interactive.py
```diff
@@ -82,6 +82,22 @@ def test_sdl_audio_mix_channels(self, args):
...
+
+    self.btest_exit('test_sdl_audio_mix_channels_halt.c', args=['-O2', '--minify=0', '--preload-file', 'the_entertainer.ogg'] + args)
+
+  @parameterized({
+    '': ([],),
+  })
+  def test_sdl_audio_mix_playing(self, args):
+    shutil.copyfile(test_file('sounds/noise.ogg'), 'sound.ogg')
```
sbc100, line 97: Same here

test/test_interactive.py
```diff
@@ -82,6 +82,22 @@ def test_sdl_audio_mix_channels(self, args):
...
+  })
+  def test_sdl_audio_mix_channels_halt(self, args):
+    shutil.copyfile(test_file('sounds/the_entertainer.ogg'), 'the_entertainer.ogg')
+
+    self.btest_exit('test_sdl_audio_mix_channels_halt.c', args=['-O2', '--minify=0', '--preload-file', 'the_entertainer.ogg'] + args)
+
+  @parameterized({
+    '': ([],),
+  })
+  def test_sdl_audio_mix_playing(self, args):
```
sbc100, line 96: You can drop the `@parameterized` here and the `args` argument

test/test_interactive.py
caiiiycuk: Does not work on windows:
```
Running the interactive tests. Make sure the browser allows popups from localhost.

test_sdl_audio_mix_playing (test_interactive.interactive.test_sdl_audio_mix_playing) ... ERROR

ERROR: test_sdl_audio_mix_playing (test_interactive.interactive.test_sdl_audio_mix_playing)
  File "C:\caiiiycuk\sdk\emscripten\test\test_interactive.py", line 91, in test_sdl_audio_mix_playing
    shutil.copyfile(test_file('sounds/noise.ogg'), '.')
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.2544.0_x64__qbz5n2kfra8p0\Lib\shutil.py", line 258, in copyfile
    with open(dst, 'wb') as fdst:
         ^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: '.'

Ran 1 test in 0.291s

FAILED (errors=1)
```

sbc100: Change `copyfile` to just `copy`

test/browser/test_sdl_audio_mix_playing.c
```diff
@@ -0,0 +1,42 @@
...
+  Mix_AllocateChannels(1);
+  chunk = Mix_LoadWAV("/sound.ogg");
+  if (!chunk) {
+    printf("ERR! Mix_LoadWAV: %s\n", Mix_GetError());
+    return -1;
+  }
+  Mix_PlayChannel(0, chunk, 0);
+  if (!Mix_Playing(0)) {
+    printf("ERR! Channel is not playing\n");
+    return -1;
```
sbc100, line 37: Can you use just `1` for failure here and in `emscripten_force_exit`?  We don't normally return negative values from `main`.

sbc100 (approved):

PR #22582, opened 2024-09-18
[tsgen] Expand TS generation test to compile and run a TS file.

brendandahl:
In the new TS file we can add tests for things where just checking if the definition file compiles is not enough. This will help catch issues like #22569.

### DIFF
--- a/test/other/embind_tsgen.cpp
+++ b/test/other/embind_tsgen.cpp
@@ -58,11 +58,18 @@ struct ValArr {
 EMSCRIPTEN_DECLARE_VAL_TYPE(CallbackType);
 
 struct ValObj {
-  Foo foo;
   Bar bar;
+  std::string string;
   CallbackType callback;
   ValObj() : callback(val::undefined()) {}
 };
+
+ValObj getValObj() {
+  ValObj o;
+  return o;
+}
+
+void setValObj(ValObj v) {}
 
 class ClassWithConstructor {
  public:
@@ -190,9 +197,11 @@ EMSCRIPTEN_BINDINGS(Test) {
     .element(emscripten::index<3>());
 
   value_object<ValObj>("ValObj")
-      .field("foo", &ValObj::foo)
+      .field("string", &ValObj::string)
       .field("bar", &ValObj::bar)
       .field("callback", &ValObj::callback);
+  function("getValObj", &getValObj);
+  function("setValObj", &setValObj);
 
   register_vector<int>("IntVec");
 
--- a/test/other/embind_tsgen_package.json
+++ b/test/other/embind_tsgen_package.json
@@ -1,0 +1,3 @@
+{
+  "type": "module"
+}
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -3396,10 +3396,10 @@ More info: https://emscripten.org
     shutil.copyfile(test_file('other/embind_tsgen_package.json'), 'package.json')
     cmd = shared.get_npm_cmd('tsc') + ['embind_tsgen.d.ts', 'main.ts', '--module', 'NodeNext', '--moduleResolution', 'nodenext']
     shared.check_call(cmd)
-    self.assertContained('main ran\nts ran', self.run_js('main.js'))
-
     actual = read_file('embind_tsgen.d.ts')
     self.assertFileContents(test_file('other/embind_tsgen_module.d.ts'), actual)
+    self.assertContained('main ran\nts ran', self.run_js('main.js'))
+
 
   def test_embind_tsgen_ignore(self):
     create_file('fail.js', 'assert(false);')
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -3391,8 +3391,8 @@ More info: https://emscripten.org
 
     # Test that the output compiles with a TS file that uses the defintions.
     shutil.copyfile(test_file('other/embind_tsgen_main.ts'), 'main.ts')
-    # A package file with type=module is needed to tell the TSC that we're
-    # using modules.
+    # A package file with type=module is needed to enabled ES modules in TSC and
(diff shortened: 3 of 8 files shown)


test/other/embind_tsgen.cpp
```diff
@@ -58,12 +58,19 @@ struct ValArr {
 EMSCRIPTEN_DECLARE_VAL_TYPE(CallbackType);
 
 struct ValObj {
-  Foo foo;
   Bar bar;
+  std::string string;
```
brendandahl, line 62: I changed this to be a more realistic use of a ValueObject.

test/test_other.py
```diff
@@ -3365,12 +3365,16 @@ def test_jspi_add_function(self):
...
     # expected.
-    self.do_runf('other/embind_tsgen.cpp', 'main ran',
-                 emcc_args=['-lembind', '--emit-tsd', 'embind_tsgen.d.ts'] + opts)
+    self.emcc(test_file('other/embind_tsgen.cpp'),
+              ['-o', 'embind_tsgen.mjs', '-lembind', '--emit-tsd', 'embind_tsgen.d.ts'] + opts)
 
     # Test that the output compiles with a TS file that uses the defintions.
-    cmd = shared.get_npm_cmd('tsc') + ['embind_tsgen.d.ts', '--noEmit']
+    shutil.copyfile(test_file('other/embind_tsgen.ts'), 'main.ts')
+    # A package file with type=module is needed to allow top level await in TS.
```
sbc100, line 3373: Can you instead use the `.mjs` extension?

test/other/embind_tsgen_package.json
```diff
@@ -0,0 +1,13 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "main.js",
+  "type": "module",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": ""
```
sbc100, line 12: Are all the other fields needed or just `type`?

test/test_other.py
```diff
@@ -3365,12 +3365,16 @@ def test_jspi_add_function(self):
...
     # Check that TypeScript generation works and that the program is runs as
     # expected.
-    self.do_runf('other/embind_tsgen.cpp', 'main ran',
-                 emcc_args=['-lembind', '--emit-tsd', 'embind_tsgen.d.ts'] + opts)
+    self.emcc(test_file('other/embind_tsgen.cpp'),
+              ['-o', 'embind_tsgen.mjs', '-lembind', '--emit-tsd', 'embind_tsgen.d.ts'] + opts)
 
     # Test that the output compiles with a TS file that uses the defintions.
-    cmd = shared.get_npm_cmd('tsc') + ['embind_tsgen.d.ts', '--noEmit']
+    shutil.copyfile(test_file('other/embind_tsgen.ts'), 'main.ts')
```
sbc100, line 3372: Maybe we should call this file `embind_tsgen_main.ts`?  (I mean the file in source control, not the copy)

test/test_other.py
```diff
@@ -3365,12 +3365,16 @@ def test_jspi_add_function(self):
...
+    # A package file with type=module is needed to allow top level await in TS.
+    shutil.copyfile(test_file('other/embind_tsgen_package.json'), 'package.json')
+    cmd = shared.get_npm_cmd('tsc') + ['embind_tsgen.d.ts', 'main.ts', '--module', 'NodeNext', '--moduleResolution', 'nodenext']
```
sbc100, line 3375: Why doesn't tsc generate a file called `.mjs` here I wonder?

Is the package.json file here needed for running `tsc` or for running the output?

test/other/embind_tsgen.ts
```diff
@@ -0,0 +1,28 @@
+
+import moduleLoader from './embind_tsgen.mjs';
```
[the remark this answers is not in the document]
brendandahl: I'll use `moduleFactory` since that's used in another ts test.

test/test_other.py
brendandahl: > Why doesn't tsc generate a file called `.mjs` here I wonder?

They prefer the package approach. IIRC the different extension was somewhat controversial.

> Is the package.json file here needed for running `tsc` or for running the output?

It's also needed for node to run it as an ES module. I've updated the comment.

sbc100 (approved):

test/test_other.py
```diff
@@ -3386,15 +3386,20 @@ def test_jspi_add_function(self):
...
+    # A package file with type=module is needed to enabled ES modules in TSC and
+    # also run the output JS file as a module in node.
+    shutil.copyfile(test_file('other/embind_tsgen_package.json'), 'package.json')
+    cmd = shared.get_npm_cmd('tsc') + ['embind_tsgen.d.ts', 'main.ts', '--module', 'NodeNext', '--moduleResolution', 'nodenext']
     shared.check_call(cmd)
+    self.assertContained('main ran\nts ran', self.run_js('main.js'))
 
     actual = read_file('embind_tsgen.d.ts')
-    self.assertFileContents(test_file('other/embind_tsgen.d.ts'), actual)
+    self.assertFileContents(test_file('other/embind_tsgen_module.d.ts'), actual)
```
sbc100, line 3402: Why do we have both `embind_tsgen.d.ts` and `embind_tsgen_module.d.ts` checked into source control?

Should these two lines move up to 3391 right after the file is first generated?

brendandahl: `other/embind_tsgen.d.ts` is used by a two other tests, it's the version of the output built without modularize & es6. I could alternatively rename it for the other tests.

sbc100 (approved):

PR #9989, opened 2019-12-08
Migrate Closure to be located via npm install

juj:
Removes Closure from tree. Also remove test_sanity.test_closure_compiler, that is somewhat extraneous. The function `check_closure_compiler` is well enough, and is run whenever one enables closure with `--closure 1`.

PR #22584, opened 2024-09-18
Remove a no-longer-needed closure workaround. NFC

sbc100:
This workaround was added in #9989, I assume because one or more tests failed without.


sbc100:
Assuming the tests pass I think we can land this.

kripken (approved):

sbc100:
@juj, I'm assuming you wouldn't have added this workaround unless it was needed for one or more tests?   i.e. you saw that one or more tests failed while trying to convert closure to npm in #9989.

PR #22591, opened 2024-09-19
[embind] Change how the number of arguments are verified.

### LINKED ISSUE #22389: Support omitting optional arguments in js embind bindings (2)
```
Thread model: posix
InstalledDir: /mnt/d/Sources/emsdk/upstream/bin
```

**Issue:**
I thought #21776 solved this issue, but it seems not. It will greatly improve usability in JS side. I would be very grateful if you could address it.

The following simple example doesn't work as expected:
C++ (test.cpp):
```cpp
#include <emscripten.h>
#include <emscripten/bind.h>
#include <optional>
#include <string>

void FuncWithOptional(std::string arg1, std::optional<std::string> arg2)
{
    printf("FuncWithOptional(): %s %s\n", arg1.c_str(), arg2 ? arg2.value().c_str() : "");
}

EMSCRIPTEN_BINDINGS(test)
{
    using namespace emscripten;
    register_optional<std::string>();
    function("FuncWithOptional", &FuncWithOptional);
}
```

compile with ``emcc -std=c++20 -lembind test.cpp -g -o test.js``

HTML & JavaScript:
```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
  <title>tests</title>
  <script src="test.js"></script>
  <script>
    function runTests() {
      Module.FuncWithOptional("abc", "def");
      Module.FuncWithOptional("abc", undefined);
      Module.FuncWithOptional("abc");
    }
  </script>
  <body>
    <div id="app">
      <input type="button" value="test" onclick="runTests()" />
    </div>
  </body>
</html>
```

``Module.FuncWithOptional("abc")`` is expected to work same as ``Module.FuncWithOptional("abc", undefined)``, but causes BindingError.
```
test.js:2329 FuncWithOptional(): abc def
test.js:2329 FuncWithOptional(): abc 
test.js:1166 Uncaught BindingError: function FuncWithOptional called with 1 arguments, expected 2
    at throwBindingError (test.js:1166:48)
    at Object.FuncWithOptional (eval at newFunc (test.js:1431:27), <anonymous>:7:11)
    at runTests (test.html:11:14)
    at HTMLInputElement.onclick (test.html:16:64)
throwBindingError @ test.js:1166
FuncWithOptional @ VM349:7
runTests @ test.html:11
onclick @ test.html:16
```

brendandahl:
This is really two changes, but they both will affect the same code so I did them at once:
 - The number of arguments is now only verified with ASSERTIONS enabled. This will help code size and the speed of embind invoker functions.
 - Allow optional arguments to be omitted.

Fixes #22389


test/embind/embind.test.js
```diff
@@ -1294,6 +1294,20 @@ module({
             value = cm.embind_test_optional_small_class_arg(undefined);
             assert.equal(-1, value);
         });
+
+        test("std::optional args can be omitted", function() {
```
sbc100, line 1298: How about arguments that have default values?

src/embind/embind_shared.js
```diff
@@ -220,11 +242,11 @@ var LibraryEmbindShared = {
     argsList = argsList.join(',')
     argsListWired = argsListWired.join(',')
 
-    var invokerFnBody = `
-      return function (${argsList}) {
-      if (arguments.length !== ${argCount}) {
-        throwBindingError('function ' + humanName + ' called with ' + arguments.length + ' arguments, expected ${argCount}');
-      }`;
+    var invokerFnBody = `return function (${argsList}) {\n`;
```
sbc100, line 245: As a followup can this be arrow function in some cases?

src/embind/embind.js
```diff
@@ -778,6 +781,10 @@ var LibraryEmbind = {
 #endif
 #if ASYNCIFY
     '$Asyncify',
+#endif
+#if ASSERTIONS
+  '$getRequiredArgCount',
+  '$checkArgCount',
```
sbc100, line 787: Indentation looks wrong here compared to above.

src/embind/embind.js
```diff
@@ -57,6 +57,9 @@ var LibraryEmbind = {
     // TODO: do we need a deleteObject here?  write a test where
     // emval is passed into JS via an interface
   }`,
+  $EmValOptionalType__deps: ['$EmValType'],
+  $EmValOptionalType: undefined,
+  $EmValOptionalType__postset: 'EmValOptionalType = Object.assign({optional: true}, EmValType);',
```
sbc100, line 62: I think there is a better way to do this.  Does ` $EmValOptionalType: '=Object.assign({optional: true}, EmValType);` work?

sbc100 (approved):

ChangeLog.md
```diff
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 3.1.68 (in development)
 -----------------------
+- The number of arguments passed to Embind function calls is now only verified
+  with ASSERTIONS enabled.
+- Optional arguments can now be omitted from Embind function calls.
```
kripken, line 25: Please add the PR numbers here.

src/embind/embind_shared.js
brendandahl: I'll look into it. This is also used in constructors so having the scope bound, may be an issue.

test/embind/embind.test.js
brendandahl: Embind doesn't really support default C++ arguments. Functions with default arguments just look like regular functions to embind. e.g. `foo(int a = 100)` is just `foo(int a)` in embind's view of things.

src/embind/embind.js
brendandahl: That works! Much nicer.

kripken (approved):

PR #22593, opened 2024-09-19
[test] Cleanup make_reftest code. NFC

sbc100:


test/common.py
```diff
@@ -2191,18 +2191,17 @@ def run_browser(self, html_file, expected=None, message=None, timeout=None, extr
...
     create_file('reftest.js', '''
       function doReftest() {
         if (doReftest.done) return;
         doReftest.done = true;
         var img = new Image();
-        img.onload = function() {
-          assert(img.width == Module.canvas.width, 'Invalid width: ' + Module.canvas.width + ', should be ' + img.width);
-          assert(img.height == Module.canvas.height, 'Invalid height: ' + Module.canvas.height + ', should be ' + img.height);
+        img.src = 'expected.png';
+        img.onload = () => {
```
kripken, line 2202: Is it safe to set `src` before `onload`? I have a vague worry about a race condition here, which is why I've written it in the opposite order.

sbc100: OK I'll revert that.

Issue #12707, opened 2020-11-05
Swap executable bit between .js and .wasm files?


boris-kolpackov:
I've noticed that when linking an executable in a mode that produces both the `.js` and `.wasm` files, the `.wasm` file has the executable bit set:

```
$ em++ -o hello.js hello.cxx
$ ls -l hello.js hello.wasm
-rw-r--r-- 1 boris boris 106336 Nov  5 11:48 hello.js
-rwxr-xr-x 1 boris boris    652 Nov  5 11:48 hello.wasm
```

I wonder what is the reason for this? I suspect it has something to do with [Standalone WebAssembly](https://github.com/emscripten-core/emscripten/wiki/WebAssembly-Standalone) where the `.wasm` file could be executed directly. However, my understanding is that in this mode (with the `.js` file), the only way to execute the `.wasm` file is via the `.js` file (since it depends on the Emscripten runtime).

On the other hand, if we were to make the `.js` file executable, then, at least on Linux, one could register, say, `nodejs` as an interpreter for `.js` files and run them like any other Linux executable. In particular, this would allow running tests without any changes to the build system.

I've just tried this on my machine and it works great:

> The `update-binfmts` helper is only available on Debian-based distributions.

```
$ sudo update-binfmts --install nodejs /usr/bin/nodejs --extension js
$ chmod ugo+x hello.js
$ ./hello.js 
Hello, World!
```

So I am wondering if the executable bit should be swapped between the `.js` and `.wasm` files?

sbc100:
I think the executable bit on the wasm file is set by wasm-ld since that its the default for executable wasm files that it creates.

We have at internal mechanism that will make the output JS file executable (and also add a `#!` line) but its currently only used when running autoconf scripts (to allow the test programs to run during configure).

I was thinking of exposing this as a `--executable` options, but we probably don't want it to be the default as most users are targeting the web and the `#!` line would just add noise.   I guess there is no harm in adding the `x` bit by default though.. we can look into that.  If you feel like taking a look the code is in `emcc.py:make_js_executable`.

PR #22595, opened 2024-09-19
Allow JS files to be executable

vicapow:
Here's a quick comparison of behavior before and after:

Before:

````
# emcc main.c -o main
# ls -alh main
-rw-r--r-- 1 root root  54K Sep 20 06:41 main # NOTE the missing `x` executable flag
````

After:

````
# emcc main.c -o main
# ls -alh main
-rwxr--r-- 1 root root 54K Sep 20 06:37 main
````

For comparison to gcc:

````
gcc main.c -o main
# ls -alh main
-rwxr-xr-x 1 root root 9.1K Sep 20 06:42 main
````

I was noticing some projects I tried to compile to web assembly were failing because they generated intermediary utilities that were then used as part of the build process in later stages. it was non-trivial to modify the original build scripts so this was a way to solve it in my case.

Personally, I think it's better to match the behavior of GCC more closely, in generally, thus the pull request. It looks like a previous workaround for this was added just for `conftest.c` for automake. see: https://github.com/emscripten-core/emscripten/blob/b435a7c46f38750e6c4490d4c0f726f452b67c82/tools/link.py#L652

Related issue: https://github.com/emscripten-core/emscripten/issues/12707

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -774,6 +774,8 @@ def phase_linker_setup(options, state, newargs):
   else:
     # Otherwise the wasm file is produced alongside the final target.
     wasm_target = get_secondary_target(target, '.wasm')
+    if final_suffix in EXECUTABLE_ENDINGS:
+        options.executable = True
 
   if settings.SAFE_HEAP not in [0, 1, 2]:
     exit_with_error('emcc: SAFE_HEAP must be 0, 1 or 2')


sbc100:
I'm generally supportive of this idea.  However, I think there are several issues here:

1. The `#!` line only works on unix, not on windows
2. Not all folks will want this the `#!` line.  Perhaps we can works around this by skipping it if folks don't include `node` in `-sENVIRONMENT`?
3. If the goal is to actually run the build-time binaries, I'm not sure that will work.  For example, by default, unless you also add NODDERAWFS then the program won't have any access the host filesystem, likely making it useless.  It might be better to teach the build system that these intermediate binaries need to be built for the host.
4. I think there may be some situations in which the `#!` line doesn't actually work... for example if the program is not in your current working directory.  I think this effects pthreaded programs, for example.   So we would want to fix that too.

sbc100:
One approach I've thought of is that maybe we can infer that a #! is desirable if the output file has no extension at all.

We might also want to make the default output file `a.out` instead of `a.out.js`?

sbc100:
> 1. Good point. Will the presence of it break windows? I don't have a windows computer to try it on.
> 2. True. I like your idea of potentially inferring this based on the lack of an extension.
> 3. True. For my use case, I also had to add the `-sNODERAWFS=1`
> 4. You mean like this example? It seems to work:
> 
> ```
> # ./main
> Hello, World!
> # cd foo/
> # ../main
> Hello, World!
> ```
> 
> That all said, maybe an easy, first option, is to add it as a flag? something like `--executable`

Yes, I agree, exposing some kind of `--executable` or `-sEXECUTABLE` might make the most sense.

PR #22598, opened 2024-09-20
Stop generating empty `.worker.js` in pthreads builds

sbc100:
We stopped using a separate worker file back in #21701.  That was released in 3.1.58 back in April.

This change ends the transition period by no longer generating a dummy/useless worker.js file alongside the output.


dschuff (approved):

kripken (commented):
Please mention this in the changelog.

PR #22603, opened 2024-09-20
[embind] Use std::disjunction for checking policy arguments.

brendandahl:
This is a little cleaner than having to define a recursive template for each of these `is<SomePolicy>`. std::disjunction isn't available until c++17, so I also added a polyfill.

I also changed conjunction to use std::conjunction when it is available and polyfill otherwise.

### DIFF
--- a/system/include/emscripten/bind.h
+++ b/system/include/emscripten/bind.h
@@ -359,6 +359,27 @@ struct reference : public allow_raw_pointers {};
 
 namespace internal {
 
+#if __cplusplus >= 201703L
+template <typename... Args> using conjunction = std::conjunction<Args...>;
+template <typename... Args> using disjunction = std::disjunction<Args...>;
+#else
+// Helper available in C++14.
+template <bool _Test, class _T1, class _T2>
+using conditional_t = typename std::conditional<_Test, _T1, _T2>::type;
+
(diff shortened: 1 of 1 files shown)


system/include/emscripten/bind.h
```diff
@@ -428,40 +449,10 @@ struct GetReturnValuePolicy<ReturnType, T, Rest...> {
...
-template<typename T, typename... Rest>
-struct isNonnullReturn<T, Rest...> {
-    static constexpr bool value = isNonnullReturn<Rest...>::value;
-};
-
-template<>
-struct isNonnullReturn<> {
-    static constexpr bool value = false;
-};
+using isNonnullReturn = disjunction<std::is_same<nonnull<ret_val>, Policies>...>;
```
sbc100, line 455: Nice!

What are `async` and `non_null` in the above lines by the way?   Are they special values defines somewhere, or just some kind of placeholder?

sbc100 (approved):

system/include/emscripten/bind.h
brendandahl: Policy args are basically options you can pass into function bindings. e.g.

```c++
   function('sleep', &sleep, async());
   function('constructor', &constructor, nonnull<ret_val>());
 ```

PR #22605, opened 2024-09-22
Webpack workaround

seanmorris:
The string of code `new URL('./', import.meta.url)` causes webpack to try and resolve the URL, which causes problems for downstream users who are using that bundler:

https://github.com/webpack/webpack/issues/16878

https://github.com/electric-sql/pglite/issues/328

This PR replaces `new URL('./', import.meta.url)` with `new URL('./', Object(import.meta).url)`, which is a workaround for this behavior.


src/shell.js
```diff
@@ -235,7 +235,9 @@ if (ENVIRONMENT_IS_NODE) {
   // EXPORT_ES6 + ENVIRONMENT_IS_NODE always requires use of import.meta.url,
   // since there's no way getting the current absolute path of the module when
   // support for that is not available.
-  scriptDirectory = require('url').fileURLToPath(new URL('./', import.meta.url)); // includes trailing slash
+  // Using Object(import.meta).url here to prevent Webpack from trying to resolve the URL
+  // https://github.com/webpack/webpack/issues/16878
+  scriptDirectory = require('url').fileURLToPath(new URL('./', Object(import.meta).url)); // includes trailing slash
```
sbc100, line 240: OMG, this is pretty disgusting.    

It makes me wonder if we need some kind of flag such as `WEBPACK_COMPATIBLE` that that we don't need to subject all users to this. 

However, I suppose that closure compiler can/will just remove this so maybe is not a huge deal.

sbc100:
@seanmorris BTW are you targetting node?   If not and you are only targeting that web you might also want to simple remove this code using `-sENVIRONMENT=web`.

sbc100:
How does webpack handle `fileURLToPath(import.meta.url);`?   We could do path.dirname(fileURLToPath(import.meta.url);` here instead maybe?

src/shell.js
seanmorris: I really don't like bundlers, but a lot of people do, so I'd like to do something.

What do you think is best?

sbc100: I think `path.dirname(fileURLToPath(import.meta.url);` would be great it if works.  Is webpack ok with it?

seanmorris: @sbc100 Yea, that notation works.

seanmorris:
> How does webpack handle `fileURLToPath(import.meta.url);`? We could do `path.dirname(fileURLToPath(import.meta.url);` here instead maybe?

That works great. Its just `new URL(___, import.meta.url)` we need to worry about.

seanmorris:
> @seanmorris BTW are you targetting node? If not and you are only targeting that web you might also want to simple remove this code using `-sENVIRONMENT=web`.

My project has separate builds for all targets, but not everyone wants to do that. This is for the countless downstream projects that want to publish universal builds.

There are valid reasons for that, too. I need to be VERY careful about the size of my project, php-wasm. I'm publishing separate builds for all targets in cjs & mjs. If I go over a certain limit, I can't publish my project via CDNs like jsdelivr or unpkg. So far I've solved the problem by splitting the extensions off into their own packages and loading them dynamically as `.so` files, but not everyone has that luxury. I may need to move cjs into its own release channel. You can see how I'm doing the loading here: https://codepen.io/SeanMorris227/pen/Yzbbrre?editors=1000

I'd really like to maximize interoperability. If there's a roadblock we can remove with a **small** notation change, then I say we do it, at least until webpack figures their mess out.

I see a LOT of potential in this project for interoperability. If someone wants to re-use code in a webpack project, but can't because of how its been published, that's going to cut the Emscripten userbase significantly.

src/shell.js
```diff
@@ -235,7 +235,7 @@ if (ENVIRONMENT_IS_NODE) {
   // EXPORT_ES6 + ENVIRONMENT_IS_NODE always requires use of import.meta.url,
   // since there's no way getting the current absolute path of the module when
   // support for that is not available.
-  scriptDirectory = require('url').fileURLToPath(new URL('./', import.meta.url)); // includes trailing slash
+  scriptDirectory = nodePath.dirname(require('url').fileURLToPath(import.meta.url)); // includes trailing slash
```
sbc100, line 238: Can you remove the trailing comment (and perhaps add `+ '/'` like 3 lines below).  It seems that `scriptDirectory` is expected to end in a `/` and `path.dirname` does not include one.

sbc100 (approved):
lgtm with comment addressed

src/shell.js
```diff
@@ -235,11 +235,13 @@ if (ENVIRONMENT_IS_NODE) {
...
   // since there's no way getting the current absolute path of the module when
   // support for that is not available.
-  scriptDirectory = require('url').fileURLToPath(new URL('./', import.meta.url)); // includes trailing slash
+  scriptDirectory = nodePath.dirname(require('url').fileURLToPath(import.meta.url));
 #else
-  scriptDirectory = __dirname + '/';
+  scriptDirectory = __dirname;
 #endif
 
+scriptDirectory += '/' // scriptDirectory is expected to end in a "/"
```
sbc100, line 243: Can you just add `+ '/'` to the end of both lines above rather than adding a second statement here?  I know it looks like code duplication but it helps on code size.

PR #22610, opened 2024-09-24
Fix -sPROXY_TO_WORKER + -sSINGLE_FILE + node

sbc100:
The proxyClient needs to jump through a few more hoops here to load the embedded generated JS file which is encoded as a data URL.

This change also fixes the mime type used when we embed JS as a data URL.

### DIFF
--- a/src/proxyClient.js
+++ b/src/proxyClient.js
@@ -18,7 +18,23 @@
 #if ENVIRONMENT_MAY_BE_NODE
 var ENVIRONMENT_IS_NODE = typeof process == 'object' && typeof process.versions == 'object' && typeof process.versions.node == 'string';
 if (ENVIRONMENT_IS_NODE) {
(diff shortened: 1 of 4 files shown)


src/proxyClient.js
```diff
@@ -18,7 +18,23 @@
...
+    // of the Web API.
+    if (typeof url == 'string' && url.startsWith('data:')) {
+#if EXPORT_ES6
+      // worker_threads always assume data URLs are ES6 modules
+      url = new URL(url);
+#else
+      // For class modules we decode the data URL and use `eval: true`.
+      url = Buffer.from(url.split(",")[1], 'base64').toString();
+      options ||= {}
+      options.eval = true;
```
kripken, line 33: I did not know this was possible...

kripken (approved):

PR #22616, opened 2024-09-24
Fix poppler benchmark.

brendandahl:
std::basic_string<unsigned short> will no longer work with LLVM 19.

### DIFF
--- a/test/third_party/poppler/cpp/poppler-global.h
+++ b/test/third_party/poppler/cpp/poppler-global.h
@@ -79,7 +79,7 @@ typedef unsigned int /* time_t */ time_type;
 #pragma warning(push)
 #pragma warning(disable: 4251) /* class 'A' needs to have dll interface for to be used by clients of class 'B'. */
 #endif
-class POPPLER_CPP_EXPORT ustring : public std::basic_string<unsigned short>
+class POPPLER_CPP_EXPORT ustring : public std::basic_string<char16_t>
 {
 public:
     ustring();


test/third_party/poppler/cpp/poppler-global.h
```diff
@@ -79,7 +79,7 @@ typedef unsigned int /* time_t */ time_type;
 #pragma warning(push)
 #pragma warning(disable: 4251) /* class 'A' needs to have dll interface for to be used by clients of class 'B'. */
 #endif
-class POPPLER_CPP_EXPORT ustring : public std::basic_string<unsigned short>
+class POPPLER_CPP_EXPORT ustring : public std::basic_string<char16_t>
```
sbc100, line 82: This is upstream code so we should avoid modifying it if possible?

Should we update the version of poppler perhaps?

sbc100: It looks like upstream still uses this: https://gitlab.freedesktop.org/poppler/poppler/-/blob/master/cpp/poppler-global.h#L102

Is upstream wrong here?

brendandahl: Upstream has `char16_t` which is what I'm changing it to.

sbc100: Oh I see..  I wonder if we should bump the upstream version then?    I also wonder if we can/should switch to a git submodule?

brendandahl: It looks like there are a number of file changes specifically for emscripten so manually updating this will be a pain. Is there a branch this built off of? I don't really want to turn this in to a bigger project.

sbc100: I only see one change (b956a15fb40aa661c7feb0f8b078f1bb0ea1d622) which looks like it might not be needed anymore. 

I guess its fine to add another change like this, but please also update test/third_party/poppler/readme.txt

sbc100 (approved):

PR #17684, opened 2022-08-19
Fix  bug in Electron environment.

mokeyish:
图片

PR #22619, opened 2024-09-25
Fix Electron renderer process compatability

idranme:
https://www.electronjs.org/docs/latest/api/process#processtype-readonly


src/shell.js
```diff
@@ -96,7 +96,7 @@ var ENVIRONMENT_IS_WEB = typeof window == 'object';
 var ENVIRONMENT_IS_WORKER = typeof importScripts == 'function';
 // N.b. Electron.js environment is simultaneously a NODE-environment, but
 // also a web environment.
-var ENVIRONMENT_IS_NODE = typeof process == 'object' && typeof process.versions == 'object' && typeof process.versions.node == 'string';
+var ENVIRONMENT_IS_NODE = typeof process == 'object' && typeof process.versions == 'object' && typeof process.versions.node == 'string' && process.type != 'renderer';
```
sbc100, line 99: Does electron set `process.versions.node`?  It looks like the check for `process.versions.node` was added specifically to deal with electron.

idranme: > Does electron set `process.versions.node`?

Yes
image

idranme: https://www.electronjs.org/docs/latest/tutorial/electron-timelines
Electron v22.0.0 comes with Node.js as v16.17.
Electron v22.0.0 was released in 2022, while https://github.com/emscripten-core/emscripten/commit/e9971d17e02d3fc3861f35d629a29baa6931842c is for 2019.

idranme: > > Does electron set `process.versions.node`?
> 
> Yes image

`process` from `contextBridge.exposeInMainWorld('process', process)` in the preload script.

sbc100: So its possible to expose the `process` module in some way.. but `process.type` will always be `renderer`?

In that case this LGTM, but can you drop the `process.versions.node` check?

Again I wish we had some kind of testing for electron in our CI :(

sbc100:
Looks like this change replaces https://github.com/emscripten-core/emscripten/pull/17684

src/shell.js
idranme: > So its possible to expose the `process` module in some way.. but `process.type` will always be `renderer`?

Yes
image
This is the result under Electron v32

idranme: > In that case this LGTM, but can you drop the `process.versions.node` check?

Feels like it's possible, but not sure it'll work on a version that's too old.

sbc100 (approved):

src/shell.js
sbc100: Actually I think we do still want the `process.versions.node` check, so that if `process` is defined by some other means its doesn't trigger this to be true.

sbc100 (approved):
lgtm.

You can rebaseline the codesize tests using `test/runner other.*code_size* other.*codesize* --rebaseline`

idranme:
> You can rebaseline the codesize tests using `test/runner other.*code_size* other.*codesize* --rebaseline`

OK

sbc100 (approved):

sbc100:
Oh, I forget to say, its best to do `emsdk install tot` whevever do rebaseline any codesize expectations.

yosoymin:
Hi, I don't really understand the purpose of this PR, but at the moment, this little change `&& process.type != 'renderer'` makes impossible to use NODEFS inside Electron with nodeIntegration. Is there another approach to fix whatever this PR solves?

PR #17752, opened 2022-08-29
Prevent transferredCanvasNames is not iterable (fixes #17751)

boardend:
Partially reverts #17577 and fixes #17751

PR #22620, opened 2024-09-25
Revert part of "Modernize JS usage in library_pthread.js. NFC"

guusw:
Part of this PR breaks the fix previously applied in #17752 the comment specifically mentioned not to use for of.

(NOTE: will edit this PR to only remove part of the change, and add a test possibly)

### DIFF
--- a/src/library_pthread.js
+++ b/src/library_pthread.js
@@ -75,7 +75,7 @@ var LibraryPThread = {
         ) {
           t = _pthread_self();
         }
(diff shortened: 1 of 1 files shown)


sbc100:
Oops, I guess this is the second time I broke this.  We should really have a test for this.

sbc100:
I uploaded a specific for this this and will make sure I add a test this time: #22623

guusw:
Thanks!

PR #22621, opened 2024-09-25
Add support for preMainLoop/postMainLoop functions

sbc100:
Rather than having code from e.g SDL, etc, in event loop code use a callback registration system.

### DIFF
--- a/src/library_eventloop.js
+++ b/src/library_eventloop.js
@@ -149,12 +149,23 @@ LibraryJSEventLoop = {
     clearInterval(id);
   },
 
+  $registerPostMainLoop: (f) => {
+    // Does nothing unless $MainLoop is included/used.
+    typeof MainLoop != 'undefined' && MainLoop.postMainLoop.push(f);
+  },
+
+  $registerPreMainLoop: (f) => {
+    // Does nothing unless $MainLoop is included/used.
+    typeof MainLoop != 'undefined' && MainLoop.preMainLoop.push(f);
+  },
+
   $MainLoop__internal: true,
   $MainLoop__deps: ['$setMainLoop', '$callUserCallback', 'emscripten_set_main_loop_timing'],
   $MainLoop__postset: `
     Module["requestAnimationFrame"] = MainLoop.requestAnimationFrame;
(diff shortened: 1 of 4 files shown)


src/library_eventloop.js
```diff
@@ -149,12 +149,21 @@ LibraryJSEventLoop = {
     clearInterval(id);
   },
 
+  $registerPostMainLoop: (f) => {
```
kripken, line 152: Maybe add a comment that as an optimization, this does not include MainLoop but instead checks for it at runtime and does nothing?

Separately, I wonder if we can do better than this. We could in theory add a "weak linking" mechanism, that would add some code only if MainLoop was actually included. Though if it's complicated it might not be worth it.

kripken (approved):

src/library_eventloop.js
sbc100: Sadly I don't think we have such a weak linking mechanism today.  The way I did it here I think is the closest thing we have to weak linking in JS code.

src/library_eventloop.js
```diff
@@ -149,12 +149,23 @@ LibraryJSEventLoop = {
...
+    // Does nothing unless $MainLoop is included/used.
+    typeof MainLoop != 'undefined' && MainLoop.postMainLoop.push(f);
```
curiousdannii, line 154:
```suggestion
    MainLoop?.postMainLoop.push(f);
```

sbc100: Sadly I don't think the ? operator works like that (i.e. it doesn't work on top level things):

```
$ node
Welcome to Node.js v18.20.3.
Type ".help" for more information.
> a?.foo
Uncaught ReferenceError: a is not defined
> 
```

sbc100: Indeed from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining:

```
Optional chaining cannot be used on a non-declared root object
```

curiousdannii: Oh sorry. That is unfortunate, I wonder why it has that restriction.

PR #17577, opened 2022-08-04
Allow use of modern JS in library_pthread.js. NFC

sbc100:

PR #22623, opened 2024-09-25
library_pthread.js: Ensure transferredCanvasNames is iterable

sbc100:
This was first broken back in #17577 and then fixed in #17752.

I then broke it again in #22545 (yay!) (see #22620).

This time I will include a test to ensure this doesn't happen again.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -12080,6 +12080,14 @@ Aborted(`Module.arguments` has been replaced by `arguments_` (the initial value 
     self.set_setting('PTHREAD_POOL_SIZE', 1)
     self.do_other_test('test_pthread_reuse.c')
 
+  @parameterized({
+    '': ([],),
+    'offscreen_canvas': (['-sOFFSCREENCANVAS_SUPPORT', '-sDEFAULT_LIBRARY_FUNCS_TO_INCLUDE=$GL'],),
+  })
+  @node_pthreads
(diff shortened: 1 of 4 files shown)


test/test_other.py
```diff
@@ -12080,6 +12080,14 @@ def test_pthread_reuse(self):
     self.set_setting('PTHREAD_POOL_SIZE', 1)
     self.do_other_test('test_pthread_reuse.c')
 
+  @parameterized({
+    '': ([],),
+    'offscreen_canvas': (['-sOFFSCREENCANVAS_SUPPORT', '-sDEFAULT_LIBRARY_FUNCS_TO_INCLUDE=$GL'],),
```
kripken, line 12085: Why is GL needed? I don't see it used in the test.

kripken (approved):
lgtm % question

test/test_other.py
sbc100: From library_pthread.js:

```
#if ASSERTIONS                                                                   
        assert(typeof GL == 'object', 'OFFSCREENCANVAS_SUPPORT assumes GL is in use (you can force-include it with \'-sDEFAULT_LIBRARY_FUNCS_TO_INCLUDE=$GL\')');
#endif  
```

kripken: I see, thanks.

PR #22630, opened 2024-09-26
Fix TCP socket send() immediately back to back after connect()

juj:
Previously, if an emulated TCP socket is connect()ed and one attempts to send() back to back right after connecting, the send would fail because the socket is not yet actually connected.

It is a bit hard to imagine where such behavior would be useful.

So this PR changes this behavior to be identical to how connectionless UDP sockets are emulated: instead, connect() calls are always pretended to succeed (since we cannot synchronously establish if the connect would fail, so presume it'll work), and all send() calls that are issued while the socket is connecting will be queued to be sent after the socket actually connects.


test/sockets/send_immediately_after_connect.cpp
```diff
@@ -0,0 +1,65 @@
...
+  if (error)
+  {
+    printf("connect() failed to error %d\n", error);
+    return error;
+  }
+
+  // Immediately send a message back-to-back from connecting to the socket
+  const char *msg = "Hello";
+  ssize_t bytes = send(sock, msg, strlen(msg), 0);
+  if (bytes != strlen(msg))
```
sbc100, line 56: The style guide puts braces on the same line as if/else/etc keyword.

test/sockets/send_immediately_after_connect.cpp
```diff
@@ -0,0 +1,65 @@
...
+  {
+    printf("send() failed to send %d bytes. Return value: %d\n", (int)strlen(msg), (int)bytes);
+    return bytes;
+  }
+
+  // Asynchronously wait until we get the message echoed back
+  emscripten_set_timeout_loop(wait_for_recv, 0, 0);
+  return 0;
+}
```
sbc100, line 65: Can you rename this test from `.cpp` to `.c`

sbc100 (approved):
lgtm % nits

test/test_sockets.py
```diff
@@ -350,6 +350,11 @@ def test_posix_proxy_sockets(self):
         # Build and run the TCP echo client program with Emscripten
         self.btest_exit('websocket/tcp_echo_client.c', args=['-lwebsocket', '-sPROXY_POSIX_SOCKETS', '-pthread', '-sPROXY_TO_PTHREAD'])
 
+  # Test that multiple pthreads calling send() on the same socket produces correct ordering semantics.
+  def test_sockets_send_immediately_after_connect(self):
+    with NodeJsWebSocketEchoServerProcess():
+      self.btest('sockets/send_immediately_after_connect.cpp', args=['-DSOCKET_DEBUG'], expected='0')
```
sbc100, line 356: Can you give the filename the full name of the test?  e.g. `test_sockets_send_immediately_after_connect.c`?   

Even though the file lives in the sockets subdirectory I find it very useful to have filenames and test names align, and we have been moving in that direction as much as possible (e.g. see the other other tests in this directory).

test/test_sockets.py
```diff
@@ -350,6 +350,11 @@ def test_posix_proxy_sockets(self):
         # Build and run the TCP echo client program with Emscripten
         self.btest_exit('websocket/tcp_echo_client.c', args=['-lwebsocket', '-sPROXY_POSIX_SOCKETS', '-pthread', '-sPROXY_TO_PTHREAD'])
 
+  # Test that multiple pthreads calling send() on the same socket produces correct ordering semantics.
+  def test_sockets_send_immediately_after_connect(self):
```
sbc100, line 354: How about calling it `test_sockets_send_while_connecting`?

test/sockets/send_immediately_after_connect.cpp
juj: It was written in C++ to make it simpler to read, but ok.

test/test_sockets.py
juj: sure

sbc100 (approved):

Issue #22607, opened 2024-09-23
Have ___sys_statfs64 return real values when the FS supports it


jeroenpf:
**Version of emscripten/emsdk:**
3.1.61

The [___syscall_statfs64 ](https://github.com/emscripten-core/emscripten/blob/8eb432e21ba19b3f14b2b367891c03f704ad45bf/src/library_syscall.js#L804)function currently returns hardcoded values while some filesystems could potentially return more accurate stats. For example, `NODEFS` could leverage [the Node statfs filesystem function](https://nodejs.org/api/fs.html#fsstatfspath-options-callback) to return actual stats for a given mount point.

Is this something worth considering or are there any problems with such an approach?

sbc100:
This sounds reasonable to me.  Would you have time to send a PR perhaps?

PR #22631, opened 2024-09-26
[nodefs] Return real values for statvfs via __syscall_statfs64

jeroenpf:
See #22607 

statfs syscall functions are returning hardcoded default values which can in some cases lead to unexpected results.

This PR introduces a way for `__syscall_statfs64` to return real values if the underlying filesystem supports this. For now, I've only implemented this for the NODEFS filesystem but we can expand this to other filesystems when it makes sense.

Additionally, in the previous defaults, `ffree` could be larger than `files` which is incorrect, this has also been addressed.

Ive added a test `test/fs/test_nodefs_statfs.c` which can be executed by running:
`test/runner test_fs_nodefs_statfs`

As I am a new contributor I'm happy to get feedback and make improvements where needed.

### DIFF
--- a/src/library_fs.js
+++ b/src/library_fs.js
@@ -690,7 +690,7 @@ FS.staticInit();
       };
 
       var parent = FS.lookupPath(path, {follow: true}).node;
-      if (typeof parent.node_ops?.statfs) {
+      if (parent.node_ops.statfs) {
         Object.assign(rtn, parent.node_ops.statfs(parent.mount.opts.root));
       }
       return rtn;
--- a/src/library_fs.js
+++ b/src/library_fs.js
@@ -673,8 +673,6 @@ FS.staticInit();
       return parent.node_ops.mknod(parent, name, mode, dev);
     },
     statfs(path) {
-      var lookup = FS.lookupPath(path, { follow: true });
-      var parent = lookup.node
 
       // NOTE: None of the defaults here are true. We're just returning safe and
       //       sane values.
@@ -691,11 +689,13 @@ FS.staticInit();
         namelen: 255,
       };
 
-      if (typeof parent.node_ops?.statfs === 'function') {
-        return { ...defaults, ...parent.node_ops.statfs(parent.mount.opts.root) };
-      } else {
+      var lookup = FS.lookupPath(path, {follow: true});
+      var parent = lookup.node;
+      if (typeof parent.node_ops?.statfs !== 'function') {
         return defaults;
       }
+
+      return { ...defaults, ...parent.node_ops.statfs(parent.mount.opts.root) };
     },
     // helpers to create specific types of nodes
     create(path, mode) {
--- a/src/library_fs.js
+++ b/src/library_fs.js
@@ -690,8 +690,7 @@ FS.staticInit();
         namelen: 255,
       };
 
-      var lookup = FS.lookupPath(path, {follow: true});
-      var parent = lookup.node;
+      var parent = FS.lookupPath(path, {follow: true}).node;
       if (typeof parent.node_ops?.statfs !== 'function') {
         return defaults;
       }
--- a/src/library_fs.js
+++ b/src/library_fs.js
@@ -676,26 +676,24 @@ FS.staticInit();
 
       // NOTE: None of the defaults here are true. We're just returning safe and
       //       sane values.
-      var ffree = FS.nextInode - 1; // Free inodes can not be larger than total inodes
-      var defaults = {
+      var rtn = {
         bsize: 4096,
         frsize: 4096,
         blocks: 1e6,
         bfree: 5e5,
         bavail: 5e5,
         files: FS.nextInode,
-        ffree,
+        ffree: FS.nextInode - 1,
         fsid: 42,
         flags: 2,
         namelen: 255,
       };
 
       var parent = FS.lookupPath(path, {follow: true}).node;
-      if (typeof parent.node_ops?.statfs !== 'function') {
-        return defaults;
+      if (typeof parent.node_ops?.statfs) {
+        Object.assign(rtn, parent.node_ops.statfs(parent.mount.opts.root));
       }
-
-      return { ...defaults, ...parent.node_ops.statfs(parent.mount.opts.root) };
+      return rtn;
     },
     // helpers to create specific types of nodes
     create(path, mode) {
--- a/src/library_fs.js
+++ b/src/library_fs.js
@@ -683,7 +683,8 @@ FS.staticInit();
         bfree: 5e5,
         bavail: 5e5,
         files: FS.nextInode,
-        ffree: 1e6,
+        // Free inodes can not be larger than total inodes so calculate a reasonable value here
+        ffree: () => Math.max(0, Math.floor(this.files * 0.9) - this.files),
         fsid: 42,
         flags: 2,
         namelen: 255,
--- a/src/library_fs.js
+++ b/src/library_fs.js
@@ -676,6 +676,7 @@ FS.staticInit();
 
       // NOTE: None of the defaults here are true. We're just returning safe and
       //       sane values.
+      var ffree = FS.nextInode - 1; // Free inodes can not be larger than total inodes
       var defaults = {
         bsize: 4096,
         frsize: 4096,
@@ -683,8 +684,7 @@ FS.staticInit();
         bfree: 5e5,
         bavail: 5e5,
         files: FS.nextInode,
-        // Free inodes can not be larger than total inodes so calculate a reasonable value here
-        ffree: () => Math.max(0, Math.floor(this.files * 0.9) - this.files),
+        ffree,
         fsid: 42,
         flags: 2,
         namelen: 255,
--- a/src/library_fs.js
+++ b/src/library_fs.js
@@ -676,7 +676,8 @@ FS.staticInit();
       var lookup = FS.lookupPath(path, { follow: true });
       var parent = lookup.node
 
-      // Error handling
+      // NOTE: None of the defaults here are true. We're just returning safe and
+      //       sane values.
       var defaults = {
         bsize: 4096,
         frsize: 4096,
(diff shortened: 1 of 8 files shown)


src/library_nodefs.js
```diff
@@ -214,6 +214,13 @@ addToLibrary({
         var path = NODEFS.realPath(node);
         return NODEFS.tryFSOperation(() => fs.readlinkSync(path));
       },
+      statfs(path) {
+        var stats = NODEFS.tryFSOperation(() => fs.statfsSync(path));
```
sbc100, line 218: How about this:

```
// Comment about why node does not provide `frsize` but `bsize is a reasonable substitute
stats.frsize = stats.bsize
return stats;
```

src/library_fs.js
```diff
@@ -672,6 +672,31 @@ FS.staticInit();
...
+        flags: 2,
+        namelen: 255,
+      };
+
+      var parent = FS.lookupPath(path, {follow: true}).node;
+      if (typeof parent.node_ops?.statfs !== 'function') {
+        return defaults;
+      }
+
+      return { ...defaults, ...parent.node_ops.statfs(parent.mount.opts.root) };
```
sbc100, line 698: How about:

```
var rtn = {
   ... // defaults
}
var parent = FS.lookupPath(path, {follow: true}).node;
if (typeof parent.node_ops?.statfs) {
   Object.assign(rtn, parent.node_ops.statfs(parent.mount.opts.root));
}
return rtn;
```

Also, so we really need to merge the defaults with the actual results?  Are we expecting backends to only paritally fill the out the information?

test/fs/test_nodefs_statfs.c
```diff
@@ -0,0 +1,35 @@
...
+
+int main() {
+    // Test the root filesystem (which should be MEMFS by default)
+    test_statfs("/");
+
+    // Mount NODEFS and test it
+    EM_ASM(
+        FS.mkdir('/working');
+        FS.mount(NODEFS, { root: '.' }, '/working');
+    );
```
sbc100, line 30: Can you move this EM_ASM block into separate function called "setup" and call it at the start on `main`?

test/fs/test_nodefs_statfs.c
```diff
@@ -0,0 +1,35 @@
...
+    assert(st.f_blocks > 0 && "Total blocks should be positive");
+    assert(st.f_bfree <= st.f_blocks && "Free blocks should not exceed total blocks");
+    assert(st.f_bavail <= st.f_bfree && "Available blocks should not exceed free blocks");
+    assert(st.f_files > 0 && "Total inodes should be positive");
+    assert(st.f_ffree <= st.f_files && "Free inodes should not exceed total inodes");
+}
+
+int main() {
+    // Test the root filesystem (which should be MEMFS by default)
+    test_statfs("/");
```
sbc100, line 24: Can we call function `test_statvfs` to match the system call name (statvfs)?   Also the filename?

test/test_core.py
```diff
@@ -5779,6 +5779,15 @@ def test_fs_nodefs_readdir(self):
     self.emcc_args += ['-lnodefs.js']
     self.do_runf('fs/test_nodefs_readdir.c', 'success')
 
+  @requires_node
+  def test_fs_nodefs_statfs(self):
+    # externally setup an existing folder structure: existing/a
+    if self.get_setting('WASMFS'):
+      self.set_setting('FORCE_FILESYSTEM')
+    os.makedirs(os.path.join(self.working_dir, 'existing', 'a'))
```
sbc100, line 5787: I think the working_dir part here is implicit.. the current working directory will always be the working dir when the test is running.  So I think you can just do `os.makedirs('existing/a')` here.

test/test_core.py
```diff
@@ -5779,6 +5779,15 @@ def test_fs_nodefs_readdir(self):
     self.emcc_args += ['-lnodefs.js']
     self.do_runf('fs/test_nodefs_readdir.c', 'success')
 
+  @requires_node
```
sbc100, line 5782: Can you mark this tests as `@crossplatform`?

test/test_core.py
sbc100: (Oh i see this is just copied from above.. I guess that test can be updated too).

src/library_fs.js
```diff
@@ -672,6 +672,31 @@ FS.staticInit();
       }
       return parent.node_ops.mknod(parent, name, mode, dev);
     },
+    statfs(path) {
+
+      // NOTE: None of the defaults here are true. We're just returning safe and
+      //       sane values.
+      var ffree = FS.nextInode - 1; // Free inodes can not be larger than total inodes
```
sbc100, line 679: Maybe just inline this down on line 687?

sbc100 (commented):
Thanks for working on this?

I wonder if we can/should make this working under NODERAWFS too?  Or can that be a followup?   I think maybe it would need to happen at the same time?

sbc100:
> Thanks for working on this?
> 
> I wonder if we can/should make this working under NODERAWFS too? Or can that be a followup? I think maybe it would need to happen at the same time?

Actually I think the `NODERAWFS` version can be a followup as it should just keep using the fake/default values up until that point.

src/library_fs.js
jeroenpf: I've applied the suggested changes.

The idea behind merging the defaults it to ensure that all the required fields are always set. In the case of Node, statfs does not return all of the information:

```
StatFs {
  type: 1397114950,
  bsize: 4096,
  blocks: 121938943,
  bfree: 61058895,
  bavail: 61058895,
  files: 999,
  ffree: 1000000
} 
```

We could add the missing fields in the `node_ops.statfs` but it seemed appropriate to have the FS class ensure a valid response. Im happy to change it though, what do you think?

jeroenpf:
> > Thanks for working on this?
> > I wonder if we can/should make this working under NODERAWFS too? Or can that be a followup? I think maybe it would need to happen at the same time?
> 
> Actually I think the `NODERAWFS` version can be a followup as it should just keep using the fake/default values up until that point.

I agree, all filesystems, besides NODEFS, should continue using the hardcoded values so we can implement the real values for the others in follow-ups.

test/fs/test_nodefs_statvfs.c
```diff
@@ -0,0 +1,35 @@
...
+		FS.mount(NODEFS, { root: '.' }, '/working');
+	);
+}
+
+int main() {
+	setup();
+    // Test the root filesystem (which should be MEMFS by default)
+    test_statvfs("/");
+    test_statvfs("/working");
+    puts("success");
```
sbc100, line 34: Can you use 2 space indentation in this file?

sbc100 (approved):

sbc100:
Looks like a test is failing:

```
FAIL [0.000s]: test_codesize_minimal_pthreads (test_other.other)
  File "/usr/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib/python3.8/unittest/case.py", line 676, in run
  File "/usr/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/root/project/test/common.py", line 330, in decorated
    f(self, *args, **kwargs)
  File "/root/project/test/test_other.py", line 8710, in test_codesize_minimal_pthreads
    self.run_codesize_test('minimal_main.c', ['-Oz', '-pthread', '-sPROXY_TO_PTHREAD'])
  File "/root/project/test/test_other.py", line 8686, in run_codesize_test
    self.assertFileContents(filename, data)
  File "/root/project/test/common.py", line 1480, in assertFileContents
    self.assertTextDataIdentical(expected_content, contents, message,
  File "/root/project/test/common.py", line 1438, in assertTextDataIdentical
    return self.assertIdentical(text1, text2, msg, fromfile, tofile)
  File "/root/project/test/common.py", line 1457, in assertIdentical
    self.fail(fail_message)
AssertionError: Unexpected difference:
--- /root/project/test/other/codesize/test_codesize_minimal_pthreads.funcs
+++ /root/project/test/other/codesize/test_codesize_minimal_pthreads.funcs.new
@@ -6,7 +6,6 @@
 $__pthread_mutex_lock
 $__pthread_mutex_trylock
 $__pthread_mutex_unlock
-$__pthread_rwlock_rdlock
 $__pthread_rwlock_timedrdlock
 $__pthread_rwlock_tryrdlock
 $__pthread_rwlock_trywrlock

For full output run with --verbose.
Run with --rebaseline to automatically update expectations

FAIL [0.000s]: test_unistd_fstatfs (test_other.other)
  File "/usr/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib/python3.8/unittest/case.py", line 676, in run
  File "/usr/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/root/project/test/common.py", line 771, in resulting_test
    return func(self, *args)
  File "/root/project/test/common.py", line 384, in metafunc
    f(self, *args, **kwargs)
  File "/root/project/test/test_other.py", line 13502, in test_unistd_fstatfs
    self.do_run_in_out_file_test('unistd/fstatfs.c')
  File "/root/project/test/common.py", line 1777, in do_run_in_out_file_test
  File "/root/project/test/common.py", line 1815, in _build_and_run
  File "/root/project/test/common.py", line 1410, in run_js
AssertionError: JS subprocess failed (/root/emsdk/node/18.20.3_64bit/bin/node --stack-trace-limit=50 --trace-uncaught /tmp/emtest_22wft709/emscripten_test_other_pxlnnu_b/fstatfs.js): 1 (expected=0).  Output:
/tmp/emtest_22wft709/emscripten_test_other_pxlnnu_b/fstatfs.js:2596
        if (parent.node_ops.statfs) {
                   ^

TypeError: Cannot read properties of null (reading 'node_ops')
```

test/test_core.py
Alex089616: test/core/test_statvfs.out

Alex089616 (changes requested):

bgrgicak:
@sbc100 I've taken over this task from @jeroenpf to wrap the work.

I had to [start a new PR to push my fix](https://github.com/emscripten-core/emscripten/pull/22932). When you have time could you please take a look at it?

Issue #22550, opened 2024-09-10
emcc --clear-cache: "Emscripten setup is not complete"


juj:
I wonder if it might make sense to skip the bootstrap check on some commands, e.g.

```
C:\emsdk\emscripten\main>emcc --clear-cache
emcc: error: emscripten setup is not complete ("npm packages" is out-of-date). Run `bootstrap` to update
```
?

sbc100:
Agreed, if its a easy change that would make sense.

On the other hand I expect most of our users are not running bootstrap themselves since they are using a pre-packaged version of emscripten (bootstrap is really only something that emscripten dev / git users should need to run).

juj:
Actually, thinking even further, given that we are able to detect when it is needed to run `bootstrap`, I wonder if we should run `bootstrap` automatically when we are about to touch any `npm` action, if frozen cache is not in effect? Like, why do we ask the user to run bootstrap, and don't just run bootstrap automatically on the fly?

sbc100:
I'm not a fan of the compiler magically running stuff like this on any invocation.   For example, I think its a mistake that we allow the automatic build of library at link time.   I think the compiler commands should be a simple as possible, and running npm and git commands behind the scenes seems like a bad idea.  I can elaborate more on the reasons why I think this.. but I'm actually heading out for the day right now.

brendandahl:
I also would prefer it to autorun. When I jump around branches, I run into bootstrap complaining. Maybe we could make the auto bootstrap a `.emscripten` setting?

sbc100:
That sounds like a great reason to add a git checkout hook, but not something we should build into emcc itself.   

I don't think its reasonable to expect a complex tool like emscripten to just work on checkout without an kind of post-checkout/build step at all. 

If we go down this path why not also have emcc download binaryen and llvm on demand.   There are several reasons I don't like this kind of idea:

1. Too much magic.  A compiler should be a compiler, it should not need to go online and download stuff when you run it.
2. We have users that to run stricly offline.   They have been asking for a way force the compiler never to go out to the network
3. It can be very confusing for tools that like to run, for example, `emcc --version` if the output also contains all kinds of other stuff, but only in some cases.

Anyway, a git checkout hooks it probably something we could have.  I personally don't mind running in manually once in while.  @brendandahl, how do node projects generally deal with this problem?   Are they running `npm install` on every checkout, or automatically in some way, or are developers normally expected to run `npm install` themselves?

juj:
[...]

PR #22632, opened 2024-09-26
Add (optional) post-commit git hook for running bootstrap script

sbc100:
Once installed this script will automatically bootstrap emscripten on each branch checkout.

Use `./bootstrap -i/--install-post-checkout` to install the script.  We could consider doing this automatically too as a followup if folks end up liking this solution.

Fixes: #22550


bootstrap.py
```diff
@@ -67,11 +80,12 @@ def main(args):
     stamp_file = get_stamp_file(name)
     if args.dry_run:
       print(' (skipping: dry run) -> %s' % ' '.join(cmd))
-      return
+      continue
```
kripken, line 83: Is this a bugfix?

kripken: (Sorry, I can't figure out how this relates to the main change.)

kripken (approved):

PR #22637, opened 2024-09-26
Fix combining of test parametization. NFC

sbc100:
As far as I can tell the only test that was relying on this features is `browser.test_sdl_image_prepare` which uses `@also_with_proxied` in addition to `@also_with_wasmfs` and it was getting the names and values backwards.  I confirmed manually that all 4 test combination now do the right thing.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -376,6 +376,8 @@ def also_with_wasmfs(f):
 
   @wraps(f)
   def metafunc(self, wasmfs, *args, **kwargs):
+    if DEBUG:
+      print('parameterize:wasmfs=%d' % wasmfs)
     if wasmfs:
       self.set_setting('WASMFS')
       self.emcc_args.append('-DWASMFS')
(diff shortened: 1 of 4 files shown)


test/common.py
```diff
@@ -719,10 +737,13 @@ def parameterize(func, parameters):
   if prev:
     # If we're parameterizing 2nd time, construct a cartesian product for various combinations.
     func._parameterize = {
-      '_'.join(filter(None, [k1, k2])): v1 + v2 for (k1, v1), (k2, v2) in itertools.product(prev.items(), parameters.items())
+      '_'.join(filter(None, [k1, k2])): v2 + v1 for (k1, v1), (k2, v2) in itertools.product(prev.items(), parameters.items())
```
sbc100, line 740: The rational here is that the outermost wrapper thats the leftmost argument.

So `foo_proxied_wasmfs` takes wasmfs as arg0 and proxied as arg1.

test/common.py
```diff
@@ -719,10 +737,13 @@ def parameterize(func, parameters):
...
     }
   else:
     func._parameterize = parameters
+  if func.__name__ == 'test_sdl_image_prepare':
+    print(func.__name__)
+    print(func._parameterize)
```
kripken, line 746: Temporary debugging?

kripken (approved):

Issue #22588, opened 2024-09-19
missingGlobal() and missingLibrarySymbol() pollute symbols into the global JS scope in MODULARIZE builds and prevent unloading Emscripten content


juj:
At Unity we support a feature to unload the game engine from a web page. This enables sites to hot-swap between multiple Unity builds on a single web page.

To enable seamless integration with site content, this unloading support has been requested to be possible without needing to use iframes, and we build it on top of Emscripten's `-sMODULARIZE` feature.

It looks like since our last Emscripten update, this feature has regressed in debug builds, due to the use of the `globalThis` object in runtime_debug.js.

The functions `missingGlobal()` and `missingLibrarySymbol()` both call to `Object.defineProperty(globalThis, ...);`, which has the effect of leaking symbol definitions into the global scope of the JS context, even when the code is being built with `-sMODULARIZE=1` where symbol definitions would be expected to be local to the MODULARIZEd function.

The defined functions contain closures that hold on to the loaded Wasm page.

Then even if one deletes all the traces of the loaded Module, these polluted global symbols have function scopes that keep the page alive.

Here is an example:

`shell_unload.html`

```html
<html><body>
<textarea rows="8"></textarea>
<input type="button" value="Unload" onclick="unload();">
<script>
  var script = document.createElement('script');
  script.src = "a.js"; // would like to be {{{ TARGET_JS_NAME }}} but that doesn't work in default runtime.
  script.onload = () => {
    var code = SpecialUnityExportedModuleName;
    delete window.SpecialUnityExportedModuleName;
    code({ print: function(...args) { document.querySelector('textarea').value += `${args.join(' ')}\n`; } });
  }
  document.body.appendChild(script);
  function unload() { // Unload button unloads the Module from the page
    script.remove();
    delete window.script;
    document.querySelector('textarea').disabled = document.querySelector('input').disabled = 'true';
    document.querySelector('input').value = 'Unloaded';
  }
  // Must exist for technical purposes even if not used: {{{ SCRIPT }}}
</script></body></html>
```

```
emcc test\hello_world.c -o a.html -sINITIAL_MEMORY=2GB -sMODULARIZE -sEXPORT_NAME=SpecialUnityExportedModuleName --shell-file shell_unload.html
```

Then run the page, click on Unload and take a Chrome heap snapshot:

![image](https://github.com/user-attachments/assets/36ad1c06-346d-47fb-8e8f-ae712c0426c8)

[...]

PR #22638, opened 2024-09-26
Avoid adding hooks for globalThis in MODULARIZE mode

sbc100:
Fixes: #22588

### DIFF
--- a/src/runtime_debug.js
+++ b/src/runtime_debug.js
@@ -56,6 +56,15 @@ function isExportedByForceFilesystem(name) {
  * their build, or no symbols that no longer exist.
  */
 function hookGlobalSymbolAccess(sym, func) {
+#if MODULARIZE && !EXPORT_ES6
+  // In MODULARIZE mode the generated code runs inside a function scope and not
+  // the global scope, and JavaScript does not provide access to function scopes
+  // so we cannot dynamically modify the scrope using `defineProperty` in this
+  // case.
+  //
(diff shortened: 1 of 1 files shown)


src/runtime_debug.js
```diff
@@ -56,6 +56,14 @@ function isExportedByForceFilesystem(name) {
...
  */
 function hookGlobalSymbolAccess(sym, func) {
+#if MODULARIZE && !EXPORT_ES6
+  // In MODULARIZE mode the generated code runs inside a function scope and not
+  // the global scope, and JavaScript does not provide access to function scopes
+  // so we cannot dynamically modify the scrope using `defineProperty` in this
+  // case.
+  //
+  // In this mode we simply ignore requests for `hookGlobalSymbolAccess`.
+#else
```
kripken, line 66: To make it clearer why it is valid to ignore here, how about renaming this function to `addWarningOnGlobalSymbol`? By saying that it is a warning, it is clear that it is optional.

sbc100: OK, will do, however all functions in this file are debug only so I think it should fairly clear.

kripken: Ah, good point. In that case lgtm with just a comment here, like "As everything else in this file, these are just debug checks, so we simply avoid them in this case."

juj:
Yeh, this is probably simplest, rather than trying to codegen the checks. Thanks for thinking about this!

PR #22639, opened 2024-09-26
Improve test_getaddrinfo. NFC

sbc100:
- Add some logging.
- Run the test on desktop to make sure it works there too.

### DIFF
--- a/test/sockets/test_getaddrinfo.c
+++ b/test/sockets/test_getaddrinfo.c
@@ -5,17 +5,52 @@
  * found in the LICENSE file.
  */
 
+#define _GNU_SOURCE
+
+#include <assert.h>
 #include <arpa/inet.h>
 #include <sys/socket.h>
 #include <assert.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


test/sockets/test_getaddrinfo.c
```diff
@@ -187,9 +238,10 @@ int main() {
   memset(&hints, 0, sizeof(hints));
   hints.ai_family = AF_INET6;
   hints.ai_socktype = SOCK_STREAM;
-  err = getaddrinfo("www.mozilla.org", "90", &hints, &servinfo);
+  err = getaddrinfo("www.google.com", "90", &hints, &servinfo);
```
sbc100, line 241: I changed the address here since www.mozilla.org doesn't have an ipv6 address.

kripken: Huh... that is surprising. But I confirmed and it seems to be the case...

kripken (approved):

PR #22640, opened 2024-09-26
Fix getaddrinfo for IPv6 loopback address

sbc100:


src/library.js
```diff
@@ -1073,7 +1073,7 @@ addToLibrary({
         if (family === {{{ cDefs.AF_INET }}}) {
           addr = _htonl({{{ cDefs.INADDR_LOOPBACK }}});
         } else {
-          addr = [0, 0, 0, 1];
+          addr = [0, 0, 0, _htonl(1)];
```
kripken, line 1076: To make sure I follow, this is an ipv6 address, and line 1074 is ipv4? (hence it is valid for one line to return an int, and other a JS array of 4 ints?)

kripken (approved):

src/library.js
sbc100: Yes, ip4 address are presented as a single number, ip6 is an array of 4 numbers.

This changes makes the address report correctly as `::1`.

Issue #22633, opened 2024-09-26
getaddrinfo behaves differently with loopback address


5d-jh:
During building librdkafka, I encountered `Name does not resolve` error. After some research, I found getaddrinfo behaves differencly with native build.

 When loopback address is given, I expect loopback as is instead of local ip address.

**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.67 (2c4461731b8c0bffbdecebc0030e0489712f9886)
clang version 20.0.0git (https:/github.com/llvm/llvm-project b9198a17315757dc0c2e831c9df0498dcab55285)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /Users/jeonghyun.lee/emsdk/upstream/bin
```

**Failing command line in full:**
```
emcc addrinfo.c -o addrinfo.js -sEXIT_RUNTIME=1 -v && node ./addrinfo.js
```
outputs
```
IPv4 address: 172.29.1.0 ((null))
```
---

```
gcc addrinfo.c -o addrinfo && ./addrinfo
```
outputs
```
IPv6 address: ::1 ((null))
IPv4 address: 127.0.0.1 ((null))
```

**Full link command and output with `-v` appended:**
```
 /Users/jeonghyun.lee/emsdk/upstream/bin/clang -target wasm32-unknown-emscripten -fignore-exceptions -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr --sysroot=/Users/jeonghyun.lee/emsdk/upstream/emscripten/cache/sysroot -DEMSCRIPTEN -Werror=implicit-function-declaration -Xclang -iwithsysroot/include/fakesdl -Xclang -iwithsysroot/include/compat -v addrinfo.c -c -o /var/folders/9l/btg_dvc90w76gpcfdrvdyj8w0000gp/T/emscripten_temp_abek1lax/addrinfo_0.o
clang version 20.0.0git (https:/github.com/llvm/llvm-project b9198a17315757dc0c2e831c9df0498dcab55285)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /Users/jeonghyun.lee/emsdk/upstream/bin
 (in-process)
[...]

PR #22641, opened 2024-09-26
Add explicit support for `localhost` DNS lookup

sbc100:
This change also includes a few micro-optimizations for the DNS helpers.
    
Fixes: #22633

### DIFF
--- a/src/library.js
+++ b/src/library.js
@@ -913,27 +913,24 @@ addToLibrary({
   $DNS: {
     address_map: {
       id: 1,
-      addrs: {},
-      names: {}
+      addrs: {'localhost': '127.0.0.1'},
+      names: {'127.0.0.1': 'localhost'}
     },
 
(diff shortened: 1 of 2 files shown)


src/library.js
```diff
@@ -913,27 +913,24 @@ addToLibrary({
...
       // If the name is already a valid ipv4 / ipv6 address, don't generate a fake one.
-      var res = inetPton4(name);
-      if (res !== null) {
+      if (inetPton4(name) != null) {
         return name;
       }
-      res = inetPton6(name);
-      if (res !== null) {
+      // Unlike the inetPton4 above we don't need and explict null comparison
+      // here since there are no valie v6 addresses that are falsey.
```
kripken, line 926:
```suggestion
      // here since there are no valid v6 addresses that are falsey.
```

kripken (approved):

PR #22643, opened 2024-09-27
Remove unnecessary `allowNull` argument from `getSocketAddress`. NFC

sbc100:
The only callsite that was using this third argument was simply doing and early return, so we can just move the check to the caller and simplify the code.

### DIFF
--- a/src/library_syscall.js
+++ b/src/library_syscall.js
@@ -324,11 +324,8 @@ var SyscallsLibrary = {
 #endif
     return socket;
   },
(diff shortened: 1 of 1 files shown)


brendandahl (commented):
r+ but looks like there's a typo or two in the commit message.

brendandahl (approved):

sbc100:
> r+ but looks like there's a typo or two in the commit message.

Thats how you can tell its one of mine :)

Issue #22077, opened 2024-06-10
run emrun on localhost / 127.0.0.1 by default


kolAflash:
Currently `emrun filename.html` listens to `0.0.0.0`. Instead it should listen to `localhost` / `127.0.0.1` by default.
_(checked version `3.1.46`)_

For security reasons debugging and testing tools like emrun should not listen to possibly publicly reachable addresses.  
Yes, there might be a firewall protecting public addresses. But it's better to pick safe defaults.
_(Yes, even the localhost address isn't extremly safe. But that [will](https://wicg.github.io/private-network-access/) [hopefully](https://developer.chrome.com/blog/private-network-access-update?hl=en) [change](https://bugzilla.mozilla.org/show_bug.cgi?id=354493) [soon](https://bugzilla.mozilla.org/show_bug.cgi?id=1481298))_

In case it's needed `emrun` already has a `--hostname` switch where you can manually set `0.0.0.0` to listen on all addresses.

sbc100:
Seems reasonable to me.  @juj WDYT?   @kolAflash would you like to send a PR?  (Please add a ChangeLog entry if you do)

kolAflash:
@sbc100 Can you give me a hand finding out why this failed?  
#22645

&nbsp;

_P.S._  
_I've another pull request which failed too. I'd be thankful if you may also have a look at it._  
_#22644_

juj:
I agree, given that we do have a --hostname=0.0.0.0 option, we would now be able to default to listening to 127.0.0.1.

I would recommend that when emrun listens to localhost/127.0.0.1, that it would print a continuation to the help message in this line https://github.com/emscripten-core/emscripten/blob/f508b432f2c1b8104ce449f4175a5217cb948b74/emrun.py#L508 , something like `, if you want this server to be accessible from another device, pass --hostname=0.0.0.0 option to listen on the network.`

Otherwise people might be surprised "why doesn't this work" kind of a way, and not know what to do to make it work.

PR #22645, opened 2024-09-27
Don't listen to 0.0.0.0 by default for security reasons. (#22077)

kolAflash:
A developer might start emrun without thinking about the --hostname setting. So he might expose sensible data on the webserver to the LAN or to the Internet. It might even happen, that a vulnerable Emsdk version becomes publicly reachable.

Instead use localhost (127.0.0.1) by default, which is usually sufficient for development.


kolAflash:
bug #22077

sbc100:
I normally do all my work over ssh and then point my browser at the IP of the my server.. so I guess I will always be adding `--hostname 0.0.0.0`.    I wonder if there is some way we can make it even more obvious to folks how to opt into this if they (like me) need to.

juj:
> so I guess I will always be adding `--hostname 0.0.0.0`.

Hmm hmm.. yeah, I can see that this can be tedious.

Although emrun already has the default that it attempts to launch a browser, so you've been passing the --no_browser flag as well I presume?

I find it tedious to need to write that `--no_browser` flag all the time, so I ended up creating a `serve.bat` in my own local user bin/ folder, which runs `emrun --no_browser --port %*`, which I then run with `serve 8000` to launch a web page. Maybe something like that might make sense?

If we defaulted to 0.0.0.0 and while doing so, print a warning that this will be accessible to all users, and then ask developers to opt to --hostname=127.0.0.1, then they would find it tedious.

kolAflash:
If you have any ideas for fixing the failed tests or making the discussed suggestions, feel free to amend a patch.

sbc100:
> > so I guess I will always be adding `--hostname 0.0.0.0`.
> 
> Hmm hmm.. yeah, I can see that this can be tedious.
> 
> Although emrun already has the default that it attempts to launch a browser, so you've been passing the --no_browser flag as well I presume?
> 
> I find it tedious to need to write that `--no_browser` flag all the time, so I ended up creating a `serve.bat` in my own local user bin/ folder, which runs `emrun --no_browser --port %*`, which I then run with `serve 8000` to launch a web page. Maybe something like that might make sense?
> 
> If we defaulted to 0.0.0.0 and while doing so, print a warning that this will be accessible to all users, and then ask developers to opt to --hostname=127.0.0.1, then they would find it tedious.

Yes, I run with `--no-browser` already.  Its easy enough for me to add `--hostname=0.0.0.0`.

PR #22648, opened 2024-09-27
Add new proxying modes foo__proxy: 'abort' and foo__proxy: 'abort_debug'

juj:
Add new proxying modes foo__proxy: 'abort' and foo__proxy: 'abort_debug', which will add compile time checks to verify that given JS function is not called in pthreads or Wasm Workers.


juj:
How to fix the lint errors?

```
C:\emsdk\emscripten\main>npm run lint --write

...

  1:4384  error  Strings must use singlequote   quotes
  1:4417  error  Operator '=' must be spaced    space-infix-ops
  1:4418  error  Strings must use singlequote   quotes
  1:4438  error  Operator '=' must be spaced    space-infix-ops
  1:4487  error  Operator '=' must be spaced    space-infix-ops
  1:4501  error  Operator '=' must be spaced    space-infix-ops
  1:4502  error  Strings must use singlequote   quotes
  1:4516  error  Operator '=' must be spaced    space-infix-ops
  1:4518  error  Operator '+' must be spaced    space-infix-ops
  1:4519  error  Strings must use singlequote   quotes
  1:4539  error  Operator '&&' must be spaced   space-infix-ops
  1:4555  error  Operator '=' must be spaced    space-infix-ops
  1:4556  error  Strings must use singlequote   quotes
  1:4579  error  Operator '=' must be spaced    space-infix-ops
  1:4597  error  Operator '=' must be spaced    space-infix-ops
  1:4612  error  Operator '=' must be spaced    space-infix-ops
  1:4628  error  Operator '=' must be spaced    space-infix-ops
  1:4629  error  Strings must use singlequote   quotes
  1:4682  error  Operator '-' must be spaced    space-infix-ops
  1:4731  error  Operator '-' must be spaced    space-infix-ops
  1:4762  error  Operator '=' must be spaced    space-infix-ops
  1:4783  error  Operator '=' must be spaced    space-infix-ops
  1:4789  error  Operator '=' must be spaced    space-infix-ops
  1:4795  error  Operator '=' must be spaced    space-infix-ops
  1:4805  error  Operator '=' must be spaced    space-infix-ops
  1:4816  error  Operator '=' must be spaced    space-infix-ops
  1:4835  error  Operator '=' must be spaced    space-infix-ops
  1:4855  error  Operator '=' must be spaced    space-infix-ops
  1:4875  error  Operator '=' must be spaced    space-infix-ops
  1:4895  error  Operator '=' must be spaced    space-infix-ops
  1:4916  error  Operator '=' must be spaced    space-infix-ops
  1:4937  error  Operator '=' must be spaced    space-infix-ops

✖ 26511 problems (26511 errors, 0 warnings)
  26507 errors and 0 warnings potentially fixable with the `--fix` option.


C:\emsdk\emscripten\main>git diff

C:\emsdk\emscripten\main>npm run check

> check
> prettier --check src/*.mjs tools/*.mjs

Checking formatting...
[warn] src/compiler.mjs
[warn] src/jsifier.mjs
[warn] src/modules.mjs
[warn] src/parseTools_legacy.mjs
[warn] src/parseTools.mjs
[warn] src/utility.mjs
[warn] tools/acorn-optimizer.mjs
[warn] tools/lz4-compress.mjs
[warn] tools/preprocessor.mjs
[warn] tools/unsafe_optimizations.mjs
[warn] Code style issues found in 10 files. Run Prettier with --write to fix.

C:\emsdk\emscripten\main>npm run check --write

> check
> prettier --check src/*.mjs tools/*.mjs

Checking formatting...
[warn] src/compiler.mjs
[warn] src/jsifier.mjs
[warn] src/modules.mjs
[warn] src/parseTools_legacy.mjs
[warn] src/parseTools.mjs
[warn] src/utility.mjs
[warn] tools/acorn-optimizer.mjs
[warn] tools/lz4-compress.mjs
[warn] tools/preprocessor.mjs
[warn] tools/unsafe_optimizations.mjs
[warn] Code style issues found in 10 files. Run Prettier with --write to fix.

C:\emsdk\emscripten\main>npm run check --write

C:\emsdk\emscripten\main>git diff

C:\emsdk\emscripten\main>prettier --check src/*.mjs tools/*.mjs
'prettier' is not recognized as an internal or external command,
operable program or batch file.

C:\emsdk\emscripten\main>npm prettier --check src/*.mjs tools/*.mjs
Unknown command: "prettier"

To see a list of supported npm commands, run:
  npm help

```

test/pthread/proxy_abort.c
```diff
@@ -0,0 +1,53 @@
+#include <emscripten.h>
+#include <pthread.h>
+#include <assert.h>
+
+// Tests behavior of __proxy: 'abort' and __proxy: 'abort_debug' in pthreads.
+
+void proxied_js_function(void);
+
+int might_throw(void(*func)())
+{
```
sbc100, line 10: Opening braces on the them line as the function decl.

src/jsifier.mjs
```diff
@@ -319,39 +319,40 @@ function(${args}) {
...
-if (ENVIRONMENT_IS_PTHREAD)
-  return ${proxyFunc}(${proxiedFunctionTable.length}, 0, ${+sync}${args ? ', ' : ''}${args});
+            if (WASM_WORKERS && ASSERTIONS) {
+              prefix += `\nassert(!ENVIRONMENT_IS_WASM_WORKER, "Attempted to call proxied function '${mangled}' in a Wasm Worker, but in Wasm Worker enabled builds, proxied function architecture is not available!");`;
+            }
+          } else if (proxyingMode == 'abort' || (proxyingMode == 'abort_debug' && ASSERTIONS)) {
+            const insideWorker = (PTHREADS && WASM_WORKERS)
+              ? '(ENVIRONMENT_IS_PTHREAD || ENVIRONMENT_IS_WASM_WORKER)'
+              : (PTHREADS ? 'ENVIRONMENT_IS_PTHREAD' : 'ENVIRONMENT_IS_WASM_WORKER');
+            prefix = `assert(!${insideWorker}, "Attempted to call function '${mangled}' inside a pthread/Wasm Worker, but this function has been declared to only be callable from the main browser thread");`;
```
sbc100, line 348: How about "but this function can only be called from the main thread"?

Using the term `main browser thread` I a little misleading because emscripten application can be run on workers where there is no main browser thread involved.  We sometimes call this "main application thread", but I think just "main thread" here is clear enough.

sbc100 (commented):
Seems reasonable if there is use case for it.   Are there existing system library function that you had in mind when designing this attribute?

src/jsifier.mjs
juj: The proxying modes `__proxy: 'sync'` and `__proxy: 'async'` definitely mean to run the JS code in the main browser thread, and not in the thread that executes 'main'.

I.e. in PROXY_TO_PTHREAD builds, `__proxy: 'sync'` and `__proxy: 'async'` run the proxied JS code in the main browser thread, which is not the same thread as the main application thread (the main runtime thread)

The earlier PROXY_TO_WORKER build mode (which I presume is what you mean by "can be run on workers where there is no main browser thread involved") predates pthreads/SharedArrayBuffer, and does not interact with SAB, or the proxying architecture. So these proxying modes are not relevant there when pthreads are not used. So I think saying "main browser thread" here is the correct thing to do?

juj: > How about "but this function can only be called from the main thread"?

Changed the message into form

```
"Attempted to call function '${mangled}' inside a pthread/Wasm Worker, but by using the proxying directive "'${proxyingMode}'", this function has been declared to only be callable from the main browser thread"
```

I would like to keep the mention about the proxy declaration that prevents the call. Otherwise a developer might be surprised to mistakenly ponder "how does it know this function is not callable from a worker?"

juj:
> Seems reasonable if there is use case for it. Are there existing system library function that you had in mind when designing this attribute?

I scanned library_browser.js and library_wasm_worker.js to see if there might be functions like this, but none stood out.

The motivation for adding this is that at Unity I am seeing more and more feature developers add their own JS functions, and they generally copy-paste the __proxy: 'sync' directives when they are unsure about what that means.

Many of these JS library functions should never use proxying, but we only ever call those JS functions in the main browser thread - if any pthread might call any of those JS functions, there would be trouble. So that is I why I wanted to make the proxying mechanism check this for me: by documenting each function that should be strictly main thread only, I can run the test suites to confirm that it indeeds happens to be the case, and that we are not getting any surprising unwanted proxying taking effect.

src/jsifier.mjs
sbc100: > The proxying modes `__proxy: 'sync'` and `__proxy: 'async'` definitely mean to run the JS code in the main browser thread, and not in the thread that executes 'main'.
> 

The thread on which the emscripten-generated code is first run I call the `main application thread`, and it might be running in worker, in which case there is no main browser thread at all in the picture.  When you write `MAIN_THREAD_EM_ASM` or `__proxy: 'sync'` its is the main application thread that runs you code.   It just happens that this is also normally the main browser thread.

src/library_wasi.js
```diff
@@ -270,6 +270,7 @@ var WasiLibrary = {
 #else
   fd_write__deps: ['$printChar'],
 #endif
+  fd_write__proxy: 'none', // Each Worker can do stdout/stderr printing on its own, without needing to proxy to the main thread.
```
sbc100, line 273: I'm pretty sure `doWritev` won't work if you call it on a thread.   And I think its not true for `printChar` either since you would end up with a different `printCharBuffers` on each thread.. which is not correct.

src/jsifier.mjs
juj: Hmm, has there been some kind of change in configuration that I am not aware of?

My expectation:

Building with `-sPROXY_TO_WORKER`: `main()` is run in a Worker thread. `__proxy` architecture is not available there.
Building with `-pthread`: `main()` is run in main browser thread. main browser thread == main application thread, `__proxy` targets main browser thread.
Building with `-pthread -sPROXY_TO_PTHREAD`: `main()` is run in a separate Worker thread, i.e. main application thread is not main browser thread. `__proxy: 'sync'` and `MAIN_THREAD_EM_ASM()` proxy to the main _browser_ thread, and not to the main _application_ thread.

In each case, proxying does target the main browser thread. Has this changed somehow?

sbc100: No nothing has changed.   I think we are just using different terminology.  

You are correct that that the main browser thread is always that target of proxying whenever we are actually running with on the main browser there.  This is because the main runtime thread and the main browser thread are the same thing in that case.

Its only when you do something like `new Worker("load_emscripten_program")` that you would have a main runtime thread that is *not* running on the main browser thread.  We do have several users that do this.

The docs for `emscripten_is_main_runtime_thread` and `emscripten_is_main_browser_thread` support this I believe: https://github.com/emscripten-core/emscripten/blob/08a147230902cb39d9706a06ca89b08b56d1bdd0/system/include/emscripten/threading.h#L49-L56

Wwith `PROXY_TO_PTHREAD`, `main()` is run on a thread that is **neither** the main runtime thread nor the main browser thread.

src/library_wasi.js
juj: > you would end up with a different `printCharBuffers` on each thread.. which is not correct.

I looked into this some decade ago, and from back then, I could not find defined semantics on how multiple threads writing on stdout/stderr should synchronize, when fflush is not called by the user. In most implementations, like ours, printing `\n` is an implicit fflush.

If multiple threads race to print to stdout/stderr, then it does seem to me it would be better semantics for each thread to construct full \n buffered lines that they would then race to print, rather than each thread mixing their individual character writes that could result in the data from multiple threads inter-mixing with each other?

Each Worker has a copy of MEMFS/library_tty.js in them, so they would each end up calling to default_tty_ops.put_char() and pushing out console.log() calls independently?

Having each thread synchronously block to perform stdout prints would seem like a rough performance overhead to have?

src/jsifier.mjs
sbc100: (I believe it was actually you who added those two functions back in 174a08276d027a8b32ca4da7e130cdf4a6aabbe1)

src/library_wasi.js
sbc100: But the function is not only used for stdout/stderr, its used for all file access. 

If we do make this change it would be very large one (basically making the filesystem multi-threaded) and I would at least expect a separate RP for it.

test/wasm_worker/proxy_abort.c
```diff
@@ -0,0 +1,48 @@
+#include <emscripten.h>
+#include <emscripten/wasm_worker.h>
+#include <assert.h>
+
+// Tests behavior of __proxy: 'abort' and __proxy: 'abort_debug' in Wasm Workers.
+
+void proxied_js_function(void);
+
+int might_throw(void(*func)())
+{
```
sbc100, line 10: Curly braces on the same line as function declarations

src/jsifier.mjs
juj: > We do have several users that do this.

Is this case a scenario where the user manually spawns a new Worker(), and inside that Worker, they load up the  Emscripten-compiled pthread-enabled program?

Such use case was never supported originally, because of the problem that in such a manual "I loaded the app into a Worker" mode, Emscripten would still need to be able to load some stuff (a lot of stuff) into the JS context of the main application thread, in order for any kind of general proxying to work. And Emscripten runtime JS libraries were not developed with the impression that the target thread that proxying occurs to, would not be the main browser thread.

For example, if I start searching for `__proxy: 'sync'` in /src/, there are functions that assume that they are always the main browser thread:

```
emscripten_set_window_title(): document.title only exists in main browser thread
emscripten_get_screen_size(): screen.width / screen.height exist only in main browser thread
emscripten_hide_mouse(): document.styleSheets only in main browser thread
emscripten_set_canvas_size: Browser.setCanvasSize() works only in main browser thread
emscripten_get_canvas_size: Module['canvas'] only works in main browser thread
emscripten_create_worker: requires Subworkers support in order to work (https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#spawning_subworkers)

library_html5.js: everything here that has __proxy: 'sync' expects that proxy target lands to the main browser thread
library_egl.js, library_glut.js: everything with __proxy: 'sync' expects that proxy target is the main browser thread
library_openal.js, library_html5_webgl.js, library_sdl.js: likewise
```
that is why I added the whole PROXY_TO_PTHREAD mode so that users would not need to figure out how to manually launch Emscripten content main() in a Worker - they could reuse the system-provided mechanism for that.

Although now that I think about it, iirc subworkers are supported widely, so that is no longer a problem (in 2015 Chrome did not support subworkers)

So maybe those users that do go and launch their pthread builds in a Worker are careful to not use any of the above APIs, and just instead do generic computation there? I.e. those people are not using Canvases or library_browser.js or library_html5.js or any other DOM accessing things that Emscripten provides?

juj: > (I believe it was actually you who added those two functions back in [174a082](https://github.com/emscripten-core/emscripten/commit/174a08276d027a8b32ca4da7e130cdf4a6aabbe1))

Yeah, I added those, but that is not for supporting the "user manually launches a pthread-enabled build inside their manually created new Worker()".

src/library_wasi.js
juj: Err, you're right of course. I somehow was interpreting this from the perspective that this function was already just for the stdout/stderr writing case. That would not work. I'll think of something here.

src/jsifier.mjs
sbc100: > Is this case a scenario where the user manually spawns a new Worker(), and inside that Worker, they load up the Emscripten-compiled pthread-enabled program?
> 
> Such use case was never supported originally, because of the problem that in such a manual "I loaded the app into a Worker" mode, Emscripten would still need to be able to load some stuff (a lot of stuff) into the JS context of the main application thread, in order for any kind of general proxying to work. And Emscripten runtime JS libraries were not developed with the impression that the target thread that proxying occurs to, would not be the main browser thread.

Yes exactly.   It turns out that this configuration is supported, at least for some programs.  Obviously such programs cannot use DOM or anything like that, but I believe we do at least some testing of this configuration.

sbc100: > Yeah, I added those, but that is not for supporting the "user manually launches a pthread-enabled build inside their manually created new Worker()".

Can you explain what you intended the difference between those two functions to be then?   I seems that unless we support that configuration then then the main runtime thread and the main browser thread would always be the same?   Unless I'm missing something?

juj: Very originally those functions intended to be a generic set to enable common handling of each of single-threaded, -pthread, and PROXY_TO_PTHREAD and PROXY_TO_WORKER build modes, so that users would be able to do multiple variants of builds from the same codebase and we'd be able to write system library code that would be aware of each of those modes.

So, for example, in a PROXY_TO_PTHREAD build, it would be expected to happen that `emscripten_is_main_runtime_thread()` == `emscripten_is_main_browser_thread()`. But then if one would make a `PROXY_TO_WORKER` build of the same, then those functions would change to return the different respective things. So that way one could have one codebase that was ready to mutate to any of those build modes.

I.e. I did not originally intend for pthreads enabled builds to ever be loaded while inside a Worker(), since the __proxy: 'sync' semantics would proxy to the hosting Worker instead of going to the main browser thread, and we have never had any facilities to do proxying to the main browser thread in such am ode.

Although now that I think about it, yeah, it is probably the case that if user avoids all those library_x.js files that I listed above (that suffer from this problem), things will probably be pretty safe for them, even if they manually load pthread-enabled builds. And one can build "reasonable" use cases in the absence of relying on those library_x.js files.

src/jsifier.mjs
```diff
@@ -437,39 +437,40 @@ function(${args}) {
...
-${async_}function(${args}) {
-if (ENVIRONMENT_IS_PTHREAD)
-  return ${proxyFunc}(${proxiedFunctionTable.length}, 0, ${+sync}${args ? ', ' : ''}${args});
+            if (WASM_WORKERS && ASSERTIONS) {
+              prefix += `\nassert(!ENVIRONMENT_IS_WASM_WORKER, "Attempted to call proxied function '${mangled}' in a Wasm Worker, but in Wasm Worker enabled builds, proxied function architecture is not available!");`;
+            }
+          } else if (proxyingMode == 'abort' || (proxyingMode == 'abort_debug' && ASSERTIONS)) {
+            const insideWorker = (PTHREADS && WASM_WORKERS)
+              ? '(ENVIRONMENT_IS_PTHREAD || ENVIRONMENT_IS_WASM_WORKER)'
+              : (PTHREADS ? 'ENVIRONMENT_IS_PTHREAD' : 'ENVIRONMENT_IS_WASM_WORKER');
```
sbc100: I think you can use the `ENVIRONMENT_IS_WORKER_THREAD` helper functions from parseTools here.

src/jsifier.mjs
```diff
@@ -437,39 +437,40 @@ function(${args}) {
...
+            prefix = `assert(!${insideWorker}, "Attempted to call function '${mangled}' inside a pthread/Wasm Worker, but by using the proxying directive '${proxyingMode}', this function has been declared to only be callable from the main browser thread");`;
```
sbc100: IIUC the `assert` function does not exist in release builds (at least not in MINIMAL_RUNTIME or STRICT mode).   So maybe when `ASSERTIONS` is not set this needs to simply be an abort?   

Or just skip the check and call it undefined behaviour in release builds?

src/lib/libwasi.js
```diff
@@ -267,6 +267,7 @@ var WasiLibrary = {
 #else
   fd_write__deps: ['$printChar'],
 #endif
+  fd_write__proxy: 'none', // Each Worker can do stdout/stderr printing on its own, without needing to proxy to the main thread.
```
sbc100: Under what circumstances is this true?    This function is used for all file isn't it, not just stdout/stderr?  Certainly we need  proxying here in a lot of cases don't we?

test/wasm_worker/proxy_abort.c
```diff
@@ -0,0 +1,44 @@
...
+  }, func);
+  return threw;
+}
+
+void test() {
+  proxied_js_function();
+}
+
+void worker_main() {
+  REPORT_RESULT(might_throw(test));
```
sbc100: Can you not just do `might_throw(proxied_js_function)` here?  (i.e. do we need the test function at all?)

src/lib/libwasi.js
sbc100: Regardless, this change to the `fd_write` syscalls seems completely separate to the new `abort` proxy types.  Can we make this change separately?

And are there some existing JS libraries calls that can be marked as `abort`?

ChangeLog.md
```diff
@@ -45,6 +45,13 @@ See docs/process.md for more on how version tagging works.
 - emcc will now error if `MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION` or
   `MINIMAL_RUNTIME_STREAMING_WASM_INSTANTIATION` are used with `SINGLE_FILE`.
   These are fundamentally incompatible but were previously ignored. (#24849)
+- Added two new proxying directives. foo__proxy: 'abort' will abort program
+  execution if JS function foo is called from a pthread or a Wasm Worker.
+  foo__proxy: 'abort_debug' will do the same, but only in ASSERTIONS builds, and
+  in non-ASSERTIONS builds will be no-op.
```
sbc100, line 54: Are there any existing JS library functions that we want to be no-op on worker threads?  I imagine this case is rare and the `abort` case is more widely useful?

If there are just one or two perhaps is better to just add `runIfWorkerThread('return;')` to the top of those rather than adding this new mechanism.

src/jsifier.mjs
```diff
@@ -437,39 +437,40 @@ function(${args}) {
...
-            const rtnType = sig && sig.length ? sig[0] : null;
-            const proxyFunc =
-              MEMORY64 && rtnType == 'p' ? 'proxyToMainThreadPtr' : 'proxyToMainThread';
-            deps.push('$' + proxyFunc);
-            return `
-${async_}function(${args}) {
-if (ENVIRONMENT_IS_PTHREAD)
-  return ${proxyFunc}(${proxiedFunctionTable.length}, 0, ${+sync}${args ? ', ' : ''}${args});
+            if (WASM_WORKERS && ASSERTIONS) {
+              prefix += `\nassert(!ENVIRONMENT_IS_WASM_WORKER, "Attempted to call proxied function '${mangled}' in a Wasm Worker, but in Wasm Worker enabled builds, proxied function architecture is not available!");`;
```
sbc100, line 461: I wonder if we should abort here in release builds too?  Otherwise what is going to happen in release builds?  Just a crash I guess?

ChangeLog.md
juj: The `abort_debug` flag is generally useful for development mode, i.e. for verifying in development builds that workers don't access unwanted JS functions. And then in release mode one can get the small code size.

> If there are just one or two perhaps is better to just add `runIfWorkerThread('return;')` to the top of those rather than adding this new mechanism.

This is also for end users to utilize, not just internally for Emscripten implemented system JS libraries.

src/jsifier.mjs
juj: What happens would depend on the function.

We don't want to unconditionally emit this, since it would increase code size for users, and they wouldn't have a way to get rid of those.

src/jsifier.mjs
juj: How would that work? I can't call

```js
            const insideWorker = ENVIRONMENT_IS_WORKER_THREAD();
```
at least, and 

```js
            const insideWorker = ' {{{ ENVIRONMENT_IS_WORKER_THREAD() }}} ';
```
neither works, I think the {{{ }}}s have maybe been expanded by this point?

sbc100: why can't you do `const insideWorker = ENVIRONMENT_IS_WORKER_THREAD();`?  I think that should work.  

(You would need to do `import { ENVIRONMENT_IS_WORKER_THREAD }  from './parseTools.js'` at the top).

ChangeLog.md
sbc100: I'd still like to see an example of where `abort_debug` is useful (i.e. and example of a function our internal library where its needed/desirable).

src/jsifier.mjs
sbc100: Fair enough

src/jsifier.mjs
juj: oh it was just missing an `export`, thanks.

src/jsifier.mjs
juj: Thanks, good catch. Updated the code to an abort() and added that case as a test.

src/lib/libwasi.js
juj: I don't quite remember why these are marked as proxy: none. Dropped them for now.

test/wasm_worker/proxy_abort.c
juj: No, since that will throw an unhandled exception, and the test run will then hang without calling out to `REPORT_RESULT()` at all.

sbc100: I just mean the `test` wrapper function:

```
void test() {
  proxied_js_function();
}
```

It seems like a pointless wrapper around `proxied_js_function`?   Is it needed for some reason?  Why can't the address of `proxied_js_function` be used in place of the address of `test`?

ChangeLog.md
juj: Well, e.g. any `egl*()` function, or any `gl*()` function: they only work from the main thread, and calling them from a worker would be a programming error (if not targeting OffscreenCanvas/OffscreenFramebuffer for WebGL).

But I wouldn't want to ship my EGL/GL utilizing program with every function having a runtime check, since that would result in a large code size increase and a performance loss.

sbc100: Can you mark one or more of the egl functions as part of this PR then perhaps?   (To show that there is a valid use case internally).

src/lib/libwasi.js
juj: Oh now I remember: you had asked that this PR should use the new annotations somewhere internally, and I think I had thought that logging should be specific to each Worker, so had tried to use them on the write functions when no filesystem is present. But that doesn't quite work out in the current structure.

ChangeLog.md
juj: Or actually, we don't know as Emscripten authors whether users would intend to use `sync` proxying for EGL or GL in such a scenario.

But it is rather the end user who knows "I won't ever do EGL/GL from a Worker." So in that light, we can't quite set that proxying type for all users's behalf in the Emscripten repo.

One use case in Unity repo is a function `js_register_accelerometer_callback(cb)`, which registers accelerometer event handler to the given callback function. We know that we will never want anything else than the main thread to call this function, i.e. a `proxy_mode: 'sync'` would not match the intent that we have for that JS function.

Dozens of other JS functions have semantics like that: they are only intended to be called from the main thread, and any pthreads/wasm workers that might call them, would be an accidental programming error.

sbc100: Hmm.. OK so maybe we don't have any use for these attributes internally at this point?

sbc100: How about `emscripten_create_wasm_worker` ?   Presumably that is only valid to call from the main thread (at least today)?   Can it me marked as `abort`?

juj: For browsers that support nested Workers, one should be able to call `emscripten_create_wasm_worker()` from within a Worker.

test/wasm_worker/proxy_abort.c
juj: Yeah, that can be removed.

ChangeLog.md
sbc100: But as of today the code simply does not handle that, right?

i.e. the worker IDs are based on the `_wasmWorkers` global that lives on the main thread only.  If each thread had its own list of `_wasmWorkers` then the IDs would be get re-used/confused, right?

At least today it would be big error to call `emscripten_create_wasm_worker` from the worker (I think).

juj: The `wasmWorkers` ID map lives always on the parent. So nested Workers will only be understood in the scope of the parent Worker (just like in JavaScript they are)

Same for the IDs, they will be relative to the parent.

Creating nested Wasm Workers was something I explicitly intended for when I first wrote it, to make a departure from pthreads where pthread creation would always be proxied to the main thread. (with code size and perf impact)

The idea was to allow users to get the same "natural" JS side behavior: since JS allows nested Worker creation, the 1:1 mapped API to Wasm would/could as well, without extra scaffolding.

sbc100: So then the call to `emscripten_terminate_wasm_worker(1)` would mean a different thing depend on the thread that ran it?  i.e. worker `1` could exist in more then one place?

That seems odd to me, but regardless, wouldn't you agree that its not ready today and so `abort` makes sense for now?

sbc100: Or perhaps you can think of another API that we have that is only supposed to be called from the main thread?

juj: > `emscripten_terminate_wasm_worker(1)` would mean a different thing depend on the thread that ran it? i.e. worker `1` could exist in more then one place?

It is always the parent that must terminate a child worker. No other worker can terminate a Worker than its parent. That is also how it works in JavaScript.

I did document this when comparing pthreads vs Wasm Workers, but maybe that point wasn't that explicit. In

https://github.com/emscripten-core/emscripten/blob/main/site/source/docs/api_reference/wasm_workers.rst

I mentioned 

```
Worker cannot terminate itself, parent thread terminates by calling

emscripten_terminate_wasm_worker(worker)
```

Given that in JS land, only parent Worker can terminate a child Worker, and if one wants to terminate e.g. a sibling or a nephew, they must postMessage() explicitly to do that - so should one do also with Wasm Workers.

juj: > Or perhaps you can think of another API that we have that is only supposed to be called from the main thread?

The challenge here is that we cannot easily know on behalf of end users whether some APIs should not be called from a pthread/Wasm Worker. Because we can `sync` proxy almost everything transparently (except the ones requiring reverse proxying), then who are we to say if an end user will consider sync proxying to be intended, or a programming error?

juj: I see I wrote this section under a heading `Nested Workers`:

-----------

Pthread topology is flat - Wasm Workers are hierarchical

On the web, if a Worker spawns a child Worker of its own, it will create a nested Worker hierarchy that the main thread cannot directly access. To sidestep portability issues stemming from this kind of topology, pthreads flatten the Worker creation chain under the hood so that only the main browser thread ever spawns threads.

Wasm Workers do not implement this kind of topology flattening, and creating a Wasm Worker in a Wasm Worker will produce a nested Worker hierarchy. If you need to create Wasm Workers from within a Wasm Worker, consider which type of hierarchy you would like, and if necessary, flatten the hierarchy manually by posting the Worker creation over to the main thread yourself.

Note that support for nested Workers varies across browsers. As of 02/2022, nested Workers are [not supported in Safari](https://bugs.webkit.org/show_bug.cgi?id=22723). See [here](https://github.com/johanholmerin/nested-worker) for a polyfill.

----------------

src/lib/libwasi.js
```diff
@@ -262,6 +262,10 @@ var WasiLibrary = {
     if (printCharBuffers[1].length) printChar(1, {{{ charCode("\n") }}});
     if (printCharBuffers[2].length) printChar(2, {{{ charCode("\n") }}});
   },
+  // When we don't have a filesystem enabled, we do not need to proxy fd_write
+  // to the main thread, because all we do is log out() and err() calls.
+  // This enables printf() to work from Wasm Workers.
+  fd_write__proxy: 'none',
```
sbc100, line 268: Can you make this a separate PR?    It seems like a good idea but unrelated to the `abort` stuff.

ChangeLog.md
sbc100: > One use case in Unity repo is a function `js_register_accelerometer_callback(cb)`, which registers accelerometer event handler to the given callback function. We know that we will never want anything else than the main thread to call this function, i.e. a `proxy_mode: 'sync'` would not match the intent that we have for that JS function.
> 
> Dozens of other JS functions have semantics like that: they are only intended to be called from the main thread, and any pthreads/wasm workers that might call them, would be an accidental programming error.

Can we think of an emscripten API that has these properties?    i.e. that we would want users only calling from the main thread?

The callback mechanism in libhtml5.js has support for registering callback on other thread I believe so I guess it wouldn't work for that case.  Perhaps we should not have made it quite so flexible in its design.

src/lib/libwasi.js
juj: I put it back because it is needed by the correctness of this PR.

Adding the new ASSERTIONS time check that functions are not attempted to be proxied in MINIMAL_RUNTIME mode exposed this function being proxied. It is correct here to declare this as proxy none, and otherwise the test suite won't pass in this PR.

ChangeLog.md
juj: > Can we think of an emscripten API that has these properties?

I don't know of one at the top of my head.

> Perhaps we should not have made it quite so flexible in its design.

I don't see how this proxying mechanism would hinder/uncover a problem in libhtml5's event system. The backproxying is very useful, otherwise the PROXY_TO_PTHREAD model wouldn't be nearly as useful to get going with.

sbc100: I was just wondering if the html5 event system we have could have been like your `js_register_accelerometer_callback` API?  But I guess that would break `PROXY_TO_PTHREAD`.   I guess `js_register_accelerometer_callback` doesn't work with `PROXY_TO_PTHREAD` either?

I still thinks is kind of a shame to add this feature, given that non or our internal APIs need/want it, but I guess its because we want to have max flexibility for out APIs and support things like `PROXY_TO_PTHREAD`, but there are external use cases like yours that might want be less flexible.

I'll give on up finding an internal use now I guess :)

src/lib/libwasi.js
sbc100: Ah I see so there is a new assertion at play here, independent of the new `abort` stuff.

I do think it would be nice to split that out so that the adding of the new `abort` feature can be added in isolation, but if you feel strongly that its no worth it then this change LGTM.

From a casual reading of the diff its quite hard to see how/where this new assertion was added (it looks like that assertion is pre-existing, but maybe not in all cases?)

src/jsifier.mjs
```diff
@@ -437,39 +438,38 @@ function(${args}) {
...
 }\n`;
-          });
-        } else if (WASM_WORKERS && ASSERTIONS) {
-          // In ASSERTIONS builds add runtime checks that proxied functions are not attempted to be called in Wasm Workers
-          // (since there is no automatic proxying architecture available)
-          snippet = modifyJSFunction(
-            snippet,
-            (args, body) => `
-function(${args}) {
-  assert(!ENVIRONMENT_IS_WASM_WORKER, "Attempted to call proxied function '${mangled}' in a Wasm Worker, but in Wasm Worker enabled builds, proxied function architecture is not available!");
```
sbc100, line 468: How come the fdwrite syscall wasn't hitting this assertion already?

src/lib/libwasi.js
juj: > From a casual reading of the diff its quite hard to see how/where this new assertion was added (it looks like that assertion is pre-existing, but maybe not in all cases?)

The previous code was

```js
if (PTHREADS) {
   ...
} else if (WASM_WORKERS && ASSERTIONS) {
  // add check: "assert(!ENVIRONMENT_IS_WASM_WORKER, "Attempted to call proxied function '${mangled}' in a Wasm Worker, but in Wasm Worker enabled builds, proxied function architecture is not available!");"
}
```
The refactored code is
```js
if (PTHREADS) {
  ...
}
if (WASM_WORKERS && ASSERTIONS) {
  // add check: "assert(!ENVIRONMENT_IS_WASM_WORKER, "Attempted to call proxied function '${mangled}' in a Wasm Worker, but in Wasm Worker enabled builds, proxied function architecture is not available!");"
}
```

When the original code was written PTHREADS vs WASM_WORKERS were mutually exclusive features - you couldn't build with both of them at the same time, so the two code blocks were effectively equivalent.

But then at some point that restriction was lifted that you could build with both of them, so the new code now accounts for that.

I agree it is a bit subtle, I am not sure if I did that fix knowingly at the time, but I am kind of wrangling a queue of more than 30 PRs here, so I'd prefer not to rewrite Git history PR commits just for the prose. I think this new check is more accurate to what we'll want.

sbc100: Thanks for the explanation.

It still seem odd though since I would have through that most of our tests are for either WASM_WORKERS *or* PTHREADS... so that assertion for fdwrite should be firing in that case?

sbc100: Figuring out exactly what is going on here seems like a good reason to split out the fdwrite change... but i get that you might be tired on splitting up changes at this point :)  

Personally I'd love to see the assertions fix + fdwrite fix land first, then the new modes.   But I really appreciate all the PRs you've been landing recently and I don't want to discourage your too much so lgtm either way.

juj: Ok, I'll defer this feature altogether now for the time being, to avoid any chances.

juj:
I'll close this for later when I have time to split up the PR.

PR #22661, opened 2024-09-30
Use getSocketAddress helper in __syscall_sendmsg. NFC

sbc100:
The code here was replicating what `getSocketAddress` was doing.

### DIFF
--- a/src/library_syscall.js
+++ b/src/library_syscall.js
@@ -441,7 +441,7 @@ var SyscallsLibrary = {
     }
     return -{{{ cDefs.ENOPROTOOPT }}}; // The option is unknown at the level indicated.
   },
-  __syscall_sendmsg__deps: ['$getSocketFromFD', '$readSockaddr', '$DNS'],
+  __syscall_sendmsg__deps: ['$getSocketFromFD', '$getSocketAddress', '$DNS'],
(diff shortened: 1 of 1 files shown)


src/library_syscall.js
```diff
@@ -451,10 +451,9 @@ var SyscallsLibrary = {
     var name = {{{ makeGetValue('message', C_STRUCTS.msghdr.msg_name, '*') }}};
     var namelen = {{{ makeGetValue('message', C_STRUCTS.msghdr.msg_namelen, 'i32') }}};
     if (name) {
-      var info = readSockaddr(name, namelen);
-      if (info.errno) return -info.errno;
```
kripken, line 455: Looks like `getSocketAddress` will throw if there is an errno?

https://github.com/emscripten-core/emscripten/blob/1dd13b720a8b550d65e4fcb68fe41281b0d7ba7a/src/library_syscall.js#L330

Is that difference not noticeable?

kripken: Oh, I guess we have those catches that convert, so it would be.

kripken (approved):

src/library_syscall.js
sbc100: Yes, all these syscalls are wrapped in try/catch to catch ErrnoError.

PR #22662, opened 2024-09-30
Use uniform prefix in websocket debugging. NFC

sbc100:


brendandahl (approved):
Alternatively, there could be a dbg_websocket function, but this seems to match how we do it elsewhere.

sbc100:
> Alternatively, there could be a dbg_websocket function, but this seems to match how we do it elsewhere.

Yes, I did try that approach a while back but I think it got rejected for some reason.

PR #19251, opened 2023-04-25
Add debug range checks for makeSetValue

sbc100:
Trying to store a value outside the range of the type being stored to can result in unexpected behaviour.  For example, storing 256 to a u8 will end up storing 1.

I decided to write this as a followup to #19239 to make it extra clear what the expected behaviour of makeSetValue is.

Adding these checks discovered real bug in out library code in `getaddrinfo`.

I ran the full other and core test suite with these checks enabled at ASSERTIONS=1 before deciding to use ASSERTIONS=2 (at least for now).

PR #22664, opened 2024-10-01
Fix MIN_INT values when checking integer ranges

sbc100:
Fixes: #19251

### DIFF
--- a/test/core/test_getValue_setValue.cpp
+++ b/test/core/test_getValue_setValue.cpp
@@ -4,6 +4,7 @@
 // found in the LICENSE file.
 
 #include <emscripten.h>
+#include <stdint.h>
 
 int main() {
   char buffer_char[8] = { 'x' };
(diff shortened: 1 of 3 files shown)


test/core/test_getValue_setValue.cpp
```diff
@@ -30,6 +31,28 @@ int main() {
     ptr = Module['getValue']($1, '*');
     out('ptr: 0x' + ptr.toString(16) + ' ' + typeof(ptr));
 #endif
-  }, buffer_char, buffer_ptr);
+
+    /* In debug mode we check the limits of input value */
```
kripken, line 35:
```suggestion
    /* In debug mode we check the limits of input values. */
```

test/test_core.py
```diff
@@ -7068,11 +7068,17 @@ def test(args=None, asserts=False):
     # see that direct usage (not on module) works. we don't export, but the use
     # keeps it alive through JSDCE
     test(args=['-DDIRECT'])
+
+    # Test with ASSERTIONS=2 we check the limits of value passed to setValue
```
kripken, line 7072:
```suggestion
    # Test with ASSERTIONS=2 so we check the limits of value passed to setValue.
```

kripken (approved):

PR #22665, opened 2024-10-01
[test] Remove `manually_trigger_reftest` option. NFC

sbc100:
This option was only being used by 5 tests.  These 5 tests are all for SDL_image and they seem to work perfectly fine with automatic reftest which happens when the application is done

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -2204,9 +2204,7 @@ class BrowserCore(RunnerCore):
       time.sleep(5)
       print('(moving on..)')
 
-  # @manually_trigger If set, we do not assume we should run the reftest when main() is done.
-  #                   Instead, call doReftest() in JS yourself at the right time.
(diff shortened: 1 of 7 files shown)


test/common.py
```diff
@@ -2215,6 +2213,7 @@ def make_reftest(self, expected, manually_trigger=False):
       function doReftest() {
         if (doReftest.done) return;
         doReftest.done = true;
+        console.log('doReftest');
```
kripken, line 2216: temporary debugging?

kripken (approved):

test/common.py
sbc100: I'm happy to leave this.. seems useful for debugging.

kripken: Sounds good.

PR #22668, opened 2024-10-02
Cleanup test_memcpy_alignment/test_memset_alignment. NFC

sbc100:
- Remove a bunch of unused C++ headers
- Convert to C
- Tabs to spaces
- Apply coding style.

I'm not sure we really need these tests anymore but maybe its good to have them for test our different optimized versions of memset/memcpy.


sbc100:
Split into two commits, the first is just the rename,  the second you should view with "hide whitespace"

kripken (approved):
Thanks for the separate commits for review!

PR #22671, opened 2024-10-02
Micro-optimize preRun/postRun handling. NFC

sbc100:


src/preamble.js
```diff
@@ -195,11 +195,10 @@ function preRun() {
...
 #if expectToReceiveOnModule('preRun')
-  if (Module['preRun']) {
-    if (typeof Module['preRun'] == 'function') Module['preRun'] = [Module['preRun']];
-    while (Module['preRun'].length) {
-      addOnPreRun(Module['preRun'].shift());
-    }
+  var preRuns = Module['preRun'];
+  if (preRuns) {
+    if (typeof preRuns == 'function') preRuns = [preRuns];
+    preRuns.forEach(addOnPreRun);
```
kripken, line 201: I worried a little here about leaving `Module['preRun']` non-empty, but I guess we never have a reason to ever look at it again, so this is safe?

kripken (approved):

src/preamble.js
sbc100: Yup.  If anything I think its better not to mutate it in place.  I don't think mutation was ever actually intended here.

PR #22672, opened 2024-10-02
[test] Re-instate ability to delay reference image collection. NFC

sbc100:
In #22665 I removed the manually_trigger_reftest. However this resulted in racey behaviour in some tests.  It turns out that running `doReftest` during `postRun` was working, but only because `doReftest` itself was doing async work, thus delaying the actual capture of the canvas bits.

This change re-instates the behaviour but without the need to test setting in python.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -2209,7 +2209,21 @@ class BrowserCore(RunnerCore):
     #   pngcrush -rem gAMA -rem cHRM -rem iCCP -rem sRGB infile outfile
     shutil.copy(expected, 'expected.png')
     create_file('reftest.js', '''
+      // We have some tests that don't perform rendering during `main` so
+      // the normal process of performing `doReftest` in `postRun` doesn't
+      // work.  These tests can delay the call to `doReftest` by calling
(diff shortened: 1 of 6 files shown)


test/common.py
```diff
@@ -2209,7 +2209,21 @@ def make_reftest(self, expected):
     #   pngcrush -rem gAMA -rem cHRM -rem iCCP -rem sRGB infile outfile
     shutil.copy(expected, 'expected.png')
     create_file('reftest.js', '''
+      // We have some tests that don't perform rendering dueing `main` so
```
kripken, line 2212:
```suggestion
      // We have some tests that don't perform rendering during `main` so
```

test/common.py
```diff
@@ -2209,7 +2209,21 @@ def make_reftest(self, expected):
...
+      // the normal process of performing `doReftest` in `postRun` doesn't
+      // work.  These tests can delay the call to `doReftest` but calling
```
kripken, line 2214:
```suggestion
      // work.  These tests can delay the call to `doReftest` by calling
```

kripken (approved):

PR #12502, opened 2020-10-13
Basic support for running Emscripten code as AudioWorklets

tklajnscek:
Basically, this PR lets you run AudioWorkletNodes that can nicely call into Emscripten-generated code. 
It sets up a pthread environment with shared memory in the AudioWorkletGlobalScope that's shared by all the audio worklets.

### Summary of major changes:
- Implemented a futex that works in the worklets so that the mutex etc works nicely. It's the same busy wait as in the main thread, but without all the extra complexity (no proxying etc)
- Added slightly customized pthread bootstrapping for audio worklets. This works via a dummy AudioWorkletNode to get the message port for worker communication.
- All the script loading is done on the main thread in case of AudioWorklets
- Changed 'worker.js' to make it work in an AudioWorkletGlobalScope. This is mainly to stop it trying to import scripts from the worklet context (not allowed) and to replace various `this.XXX` since `this` doesn't exist in AudioWorkletGlobalScope

### Current limitations:
**Doesn't work with MINIMAL_RUNTIME** as I haven't figured out how to export functions so that they can be called from the registered `AudioWorkletProcessor`.

### Examples
There's an example in `tests/audioworklet` that also tests the pthread environment setup and the futex implementation.

We are using (a much less clean) version of this in production and it works great!
Any comments highly appreciated! :)

PR #22681, opened 2024-10-04
[AudioWorklet] replace repeated use of 128 with defined constant

cwoffenden:
The Web Audio API defines the processed sample size as always being `128`, which is hardcoded in both the test code and docs, so this replaces it with a defined constant `WEBAUDIO_QUANTUM_SIZE`.


juj:
I recall that Web Audio has gained a feature to allow a custom render quantum, though I actually never got around to tending to this:
- https://github.com/WebAudio/web-audio-api/issues/2450
- https://github.com/WebAudio/web-audio-api/pull/2469

so I wonder instead of promoting a define (which will be hardcoded in size), we should look into how to expand the api to read the actual quantum in a configurable way? I haven't looked into the details yet, on how to pull this off in a manner that doesn't break existing users (or I wonder if we might even care about that).

cwoffenden:
> I wonder instead of promoting a define (which will be hardcoded in size), we should look into how to expand the api to read the actual quantum in a configurable way?

I did read through most of the proposals but I think I zoned out after a while due to the multi-year nature of it. But you're right, the safest would be to have an API call to get the value. The details are in the draft:

https://webaudio.github.io/web-audio-api/#BaseAudioContext

I can test for [renderQuantumSize](https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-renderquantumsize), and if it exists great, otherwise return the default.

cwoffenden:
@juj The context's `renderQuantumSize` doesn't look to be implemented anywhere yet, so I'll come back to this. I did however add a `quantumSize` field to the `AudioSampleFrame`, as well as a standalone `emscripten_audio_context_quantum_size()`. Both are hardcoded to 128 for now. Both have their uses, the function call for deciding upon buffer/stack sizes, the struct field for the callback.

Whilst making changes to the `audio_worklet.js` I used the `struct_info` to get the sizes, so for example:
```
stackMemoryNeeded = (numInputs + numOutputs) * 8
```
Becomes:
```
stackMemoryNeeded = (numInputs + numOutputs) * {{{ C_STRUCTS.AudioSampleFrame.__size__ }}}
```
Then we never need to keep track of these manually with more magic numbers. And also same with the field offsets (playing nice with wasm64. All of these get resolved to constants.

system/include/emscripten/webaudio.h
```diff
@@ -95,19 +95,27 @@ typedef void (*EmscriptenWorkletProcessorCreatedCallback)(EMSCRIPTEN_WEBAUDIO_T
 // userData3: A custom userdata pointer to pass to the callback function. This value will be passed on to the call to the given EmscriptenWorkletProcessorCreatedCallback callback function.
 void emscripten_create_wasm_audio_worklet_processor_async(EMSCRIPTEN_WEBAUDIO_T audioContext, const WebAudioWorkletProcessorCreateOptions *options, EmscriptenWorkletProcessorCreatedCallback callback, void *userData3);
 
+// Returns the number of samples processed per channel in an AudioSampleFrame, fixed at 128 in the Web Audio API 1.0 specification, and valid for the lifetime of the audio context.
+// For this to change from the default 128, the context would need creating with a yet unexposed WebAudioWorkletProcessorCreateOptions renderSizeHint, part of the 1.1 Web Audio API.
+int emscripten_audio_context_quantum_size(EMSCRIPTEN_WEBAUDIO_T audioContext);
```
sbc100, line 100: I don't see this new API used anywhere in tests.   It looks like the test is instead reading the quantum size from the AudioSampleFrame object.   Do we need both ways to discover this?  If so can we tests this function too, if only to show how/when its useful?

cwoffenden: I could pass the context around in the user data, but that's a very bad usage pattern.

I'll use it to print what the quantum size is, then we at least have coverage.

test/webaudio/audio_worklet_tone_generator.c
```diff
@@ -98,7 +101,12 @@ void AudioWorkletProcessorCreated(EMSCRIPTEN_WEBAUDIO_T audioContext, bool succe
         audioContext.suspend();
       }
     };
-  }, audioContext);
+
+    // Report the buffer size out of interest
+    let quantumText = document.createElement('p');
+    quantumText.innerHTML += `(Buffer quantum size: ${$1})`;
+    document.body.appendChild(quantumText);
```
sbc100, line 108: Can you just use `printf` to report the quantum size rather than JS / DOM?

src/audio_worklet.js
```diff
@@ -38,60 +38,69 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
-      dataPtr = inputsPtr + numInputs * 8;
+      dataPtr = inputsPtr + numInputs * {{{ C_STRUCTS.AudioSampleFrame.__size__ }}};
       for (i of inputList) {
         // Write the AudioSampleFrame struct instance
-        HEAPU32[k++] = i.length;
-        HEAPU32[k++] = dataPtr;
+        HEAPU32[k + {{{ C_STRUCTS.AudioSampleFrame.numberOfChannels / 4 }}}] = i.length;
+        HEAPU32[k + {{{ C_STRUCTS.AudioSampleFrame.quantumSize / 4 }}}] = quantumSize;
+        HEAPU32[k + {{{ C_STRUCTS.AudioSampleFrame.data / 4 }}}] = dataPtr;
+        k += {{{ C_STRUCTS.AudioSampleFrame.__size__ / 4}}};
```
sbc100, line 73: Space after 4 here

sbc100 (approved):

src/audio_worklet.js
```diff
@@ -38,60 +38,69 @@ function createWasmAudioWorkletProcessor(audioParams) {
     }
 
     process(inputList, outputList, parameters) {
+      // hardcoding this until the Web Audio 1.1 API spec makes it into browsers and we can get the AudioWorkletProcessor's context quantum
+      const quantumSize = 128;
```
juj, line 42: Reading the spec at https://webaudio.github.io/web-audio-api/ , I think we can (and should) already implement this, as it will be need to be implemented in a backwards compatible manner anyways.

I.e. something like `const quantumSize = context.renderQuantumSize || 128;`

The reason I say that we should implement this already now, because the data flow needs tending to, in order to get the `renderQuantumSize` posted into the worklet. So it would be good to implement it already pretending that the `renderQuamtumSize` parameter might exist in an implementation. Reading the spec, it seems fairly complete to assume that the quantum size will exist at some point.

Also, I wonder if we should expand the context creation function to allow taking in a quantum size?

juj:
> playing nice with wasm64

I see there are still pointer e.g. assignments to HEAPU32 that won't actually work in Wasm64 mode, which will need HEAPU64 assignments. But that is good work for a separate PR, to keep the wasm64 part orthogonal.

juj:
> Once the 1.1 version of the Web Audio API is supported

I'm getting an impression this might not happen any time soon, though maybe we should still implement the renderQuantumSize request/read flow in a way that is complete/forward-looking for the implementation that should appear.

juj (approved):
Overall looks good. Personally I would squeeze some code size bytes out of the JS code, though I think it's fine like this.

One quirk is that the quantumSize is not part of the AudioSampleFrame, i.e. there cannot be separate quantum sizes for separate sample inputs and outputs, it's always a constant.

But the only way to address that would probably be to change the signature of the audio process callback, which would be a breaking change, so this looks like a nice middle ground.

I'd like to see the setting and reading the renderQuantumSize flow to exist, just to make sure we can actually get the renderQuantumSize information into the worklet cleanly. (we probably need to postMessage it at audio worklet creation time?) But works either way.

src/audio_worklet.js
cwoffenden: > I.e. something like `const quantumSize = context.renderQuantumSize || 128;`

Originally I did, but I chose not to in the end because it's `[SecureContext]` in the spec, and when it finally makes it to a browser I didn't want a exception. I thought of wrapping in a try/catch, but not knowing at all what the behaviour would be, I opted for the safer option.

> Also, I wonder if we should expand the context creation function to allow taking in a quantum size?

I also considered this, in preparation, but since I didn't know how it would be treated in reality I chose the safer option.

The rationale behind both decisions were: if we expose setting it, we need to be able to read it otherwise things will break, and if we read it things might also break. We're itching to see the larger buffer sizes at work, so it's definitely something I or @tklajnscek will revisit as soon as.

cwoffenden:
> Personally I would squeeze some code size bytes out of the JS code, though I think it's fine like this.

I may just shorten some variable names for now. Ultimately, time permitting (which is wishful thinking), I want to look at what it would take to get the worklet code through Closure, the same as sbc100 did recently with general worklets (IIRC).

> One quirk is that the quantumSize is not part of the AudioSampleFrame, i.e. there cannot be separate quantum sizes for separate sample inputs and outputs, it's always a constant.

This also irks me, but as you mention I didn't want to make a breaking change to the callback signature. It might be worth breaking though, since the API didn't work with Closure until #22472 a few months back, I wonder how much production pain this would cause for others?

I could mark `emscripten_create_wasm_audio_worklet_node()` as deprecated, add a new one that takes a different `EmscriptenWorkletNodeProcessCallback` signature, and let the two live side by side for a while?

> I'd like to see the setting and reading the renderQuantumSize flow to exist, just to make sure we can actually get the renderQuantumSize information into the worklet cleanly.

I'd pass it as one of the processor opts, along with the callback, etc. We already know it at that point, so in the creation code I could read it from `emscripten_audio_context_quantum_size()` then we only have one hardcoded `128` and a single place to make the change. I'll try to look at this today.

cwoffenden:
@juj The quantum size has a single getter now, used internally where needed:

https://github.com/emscripten-core/emscripten/blob/8e1115a8095c777bfacb5117257bb00b09788d30/src/library_webaudio.js#L41

The value is also passed in the options to the `WasmAudioWorkletProcessor` at construction, since it's fixed at context creation time:

https://github.com/emscripten-core/emscripten/blob/8e1115a8095c777bfacb5117257bb00b09788d30/src/library_webaudio.js#L299

If this value is expected to change, e.g. the worklet node gets reattached to another context (I don't think this possible) then we could simple read it in `process()` from the same function.

cwoffenden:
(I see the filesize checks are failing in other PRs, it's not just me...)

sbc100:
> (I see the filesize checks are failing in other PRs, it's not just me...)

Filesize tests rebaselined. Please rebase or merge

sbc100:
> I may just shorten some variable names for now. Ultimately, time permitting (which is wishful thinking), I want to look at what it would take to get the worklet code through Closure, the same as sbc100 did recently with general worklets (IIRC).
> 

You should not need to use short variable names in the source code.  We can rely on closure/minifiers to shorten the names.

I think @juj was likely referring to other tricks when he said "I would squeeze some code size bytes out of the JS code"

cwoffenden:
> You should not need to use short variable names in the source code.

The worklet doesn’t get minified currently and I add a few extra lines in there to use the struct sizes/offsets instead of magic numbers.

sbc100:
> > You should not need to use short variable names in the source code.
> 
> The worklet doesn’t get minified currently and I add a few extra lines in there to use the struct sizes/offsets instead of magic numbers.

I'm not sure what you mean.  The code running the worklet is the exact same code that runs on the main thread right?   i.e. we only generate one single program from the JS library files, right?     If you build with `--closure` it will be minified, right?

cwoffenden:
> If you build with `--closure` it will be minified, right?

I’ll need to recheck tomorrow but I’m pretty sure the property and variable names I added are staying the same (e.g. `this.quantumSize` doesn’t get shortened).

cwoffenden:
> We can rely on closure/minifiers to shorten the names.

I just verified to be sure, and the generated `myfile.aw.js` has no difference whether or not it's built with `--closure=1`. It has whitespace, etc., removed when building with `-O2` for example, but Closure doesn't appear to be run.

cwoffenden:
> I'd like to see the setting and reading the renderQuantumSize flow to exist, just to make sure we can actually get the renderQuantumSize information into the worklet cleanly.

@juj I don't think we need to, but it wants more investigation.  The typed arrays passed in the input and output lists here:

https://github.com/emscripten-core/emscripten/blob/429f70c6f3636e26ae5b559bf96d3db89fdf8d31/src/audio_worklet.js#L40

Each is the size of the quantum, so if all calculations were based on the lengths of these, then everything would work now and in the future. It shouldn't need the quantum passing in at all, but I'd like to see it in action.

sbc100:
> > We can rely on closure/minifiers to shorten the names.
> 
> I just verified to be sure, and the generated `myfile.aw.js` has no difference whether or not it's built with `--closure=1`. It has whitespace, etc., removed when building with `-O2` for example, but Closure doesn't appear to be run.

Ah yes, but that is just the bootstrap code the audio worklet, that main program code (all the code that lives in library_xxx.js) goes in `myfile.js` which should be covered by closure, and I think that covers all the code in this PR.

cwoffenden:
> Ah yes, but that is just the bootstrap code the audio worklet

There are changes to both the bootstrap and library, the bootstrap is mostly the likes of `{{{ C_STRUCTS.AudioSampleFrame.__size__ }}}` due to changes to the offsets. I'd like to have used `makeSetValue` but it needs amending to work with how it accesses the worklet heap.

sbc100:
> > Ah yes, but that is just the bootstrap code the audio worklet
> 
> There are changes to both the bootstrap and library, the bootstrap is mostly the likes of `{{{ C_STRUCTS.AudioSampleFrame.__size__ }}}` due to changes to the offsets. I'd like to have used `makeSetValue` but it needs amending to work with how it accesses the worklet heap.

Ah, sorry I didn't see that.  Still I don't think you want to be doing minification of variable names or other things that closure generatelly does in this code.  A better path forward there would be to work on getting closure to run on that code

cwoffenden:
> A better path forward there would be to work on getting closure to run on that code

Unless there's anything hideous I believe this PR is done, then I plan on looking at getting it all through Closure and other tweaks. I need to first move code away from our own Audio Worklet implementation (#12502 which we've been maintaining and shipping since 2020!) beforehand though.

cwoffenden:
Any feedback on this? I've more PRs to follow but I'm building each on the last since they're touching the same files.

I'm happy to mark the current API as deprecated and create a parallel API with the quantum, or go with this current PR and come back to this later. I'm itching to get to my next task of replacing this in the worklet:
```
for (i of outputList) {
  for (j of i) {
    for (k = 0; k < this.quantumSize; ++k) {
      j[k] = HEAPF32[outputDataPtr++];
    }
  }
}
```
Which I think is doable because the address never changes (given the worklet's stack is passed in externally) so we can create a view on the channel data valid for the life of the worklet, and thus can perform a garbage-free copy.

src/library_webaudio.js
```diff
@@ -37,11 +37,21 @@ let LibraryWebAudio = {
...
 
-  // emscripten_create_audio_context() does not itself use
+  // Performs the work of getting the AudioContext's quantum size.
+  $emscriptenGetContextQuantumSize: (contextHandle) => {
+    // TODO: in a future release this will be something like:
+    //   return EmAudio[contextHandle].renderQuantumSize || 128;
+    // It comes two caveats: it needs the hint when generating the context adding to
+    // emscripten_create_audio_context(), and altering the quantum requires a secure
+    // context and fallback implementing. Until then we simply use the 1.0 API value:
+    return 128;
```
juj, line 47: Couldn't this code already today read

```js
return EmAudio[contextHandle].renderQuantumSize || 128;
```

That way the code would also be verifying that `contextHandle` is a valid context.

src/library_webaudio.js
```diff
@@ -293,6 +308,15 @@ let LibraryWebAudio = {
...
 #endif // ~AUDIO_WORKLET
 
+  emscripten_audio_context_quantum_size__deps: ['$emscriptenGetContextQuantumSize'],
+  emscripten_audio_context_quantum_size: (contextHandle) => {
+#if ASSERTIONS
+    assert(EmAudio[contextHandle], `Called emscripten_audio_context_quantum_size() with an invalid Web Audio Context handle ${contextHandle}`);
+    assert(EmAudio[contextHandle] instanceof (window.AudioContext || window.webkitAudioContext), `Called emscripten_audio_context_quantum_size() on handle ${contextHandle} that is not an AudioContext, but of type ${EmAudio[contextHandle]}`);
+#endif
+    return emscriptenGetContextQuantumSize(contextHandle);
+  },
```
juj, line 318: I would delete the second function `$emscriptenGetContextQuantumSize` and just have one function `emscripten_audio_context_quantum_size()` and use that - it would be simpler that way?

src/library_webaudio.js
cwoffenden: It could, but I was worried about its `[SecureContext]` marking in the spec and what that entails, and also since we can't yet set it then it doesn't make a difference.

But I'll happily change it and look at it again if/when it breaks, and when the API allows for the setting.

src/library_webaudio.js
cwoffenden: One was to expose it to JS for use in the worklet or in other code, the other native. Originally I didn't pass the value to the worklet, instead reading it from this JS function, but you'd said to demonstrate the value being passed.

I can re-look at this.

test/webaudio/audio_worklet_tone_generator.c
```diff
@@ -38,12 +40,12 @@ bool ProcessAudio(int numInputs, const AudioSampleFrame *inputs, int numOutputs,
...
   // Produce a sine wave tone of desired frequency to all output channels.
   for(int o = 0; o < numOutputs; ++o)
-    for(int i = 0; i < 128; ++i)
+    for(int i = 0; i < outputs[o].quantumSize; ++i)
     {
       float s = emscripten_math_sin(phase);
       phase += phaseIncrement;
       for(int ch = 0; ch < outputs[o].numberOfChannels; ++ch)
-        outputs[o].data[ch*128 + i] = s * currentVolume;
+        outputs[o].data[ch*outputs[o].quantumSize + i] = s * currentVolume;
```
juj, line 48: Given that the quantum size is a constant, I wonder if it would be better to have all these examples use that global `quantumSize` integer, e.g.

```c
int quantumSize;

bool ProcessAudio(int numInputs, ...) {
  for(int o = 0; o < numOutputs; ++o)
    for(int i = 0; i < quantumSize; ++i)
      ...
}

int main() {
  quantumSize = emscripten_audio_context_quantum_size(context);
}
```
This would communicate the invariant that the quantumSize cannot change between inputs and outputs in a single call to ProcessAudio(). That might help someone learning about AudioWorklets.

Alternatively, I wonder if the field `AudioSampleFrame::quantumSize` might be simpler to rename to `AudioSampleFrame::length` to signal that it just means the length of the data array. Then users could have

```c
for(int i = 0; i < outputs[o].length; ++i) {
  outputs[o].data[i] = ...;
```

with very idiomatic reading C code. The "quantumSize" name in this context reads fancier than it necessarily needs to be (and doesn't match the Web Audio spec which calls it "renderQuantumSize").

Works either way, leaving it up to you to decide.

system/include/emscripten/webaudio.h
juj: >Do we need both ways to discover this?

In more complex audio applications, there will definitely be a need to know what the quantum size is up front, before the audio graph is started running. This is e.g. for the nature of allocating internal data structures for users' own audio processing graphs, which happens before audio playback is started.

So `emscripten_audio_context_quantum_size()` is very useful for allowing that to happen.

> once the quantum is larger we'd need to query the returned hardware size before calling `emscripten_start_wasm_audio_worklet_thread_async()` with a stack size.

It is fine for users to specify stack sizes that can be well larger than any possible quantum size they choose, i.e. there's no need to squeeze oneself so tight with the stack space so that e.g. 128 quantum would fit, but 512 would not. So users might do a 64K stack size for the worklet, which should be plenty to fit all of the data, even in a huge 10.2 channels format.

test/webaudio/audio_worklet_tone_generator.c
cwoffenden: I don't like setting it as a global, and I agree, I also don't like the name `quantumSize` either (or `renderQuantumSize` or the use of quantum in general for this, or the way audio frame is used, etc.).

If I'm going to spend more time with this I'd rather that be spent marking the existing API as deprecated and creating a new callback signature.

src/library_webaudio.js
juj: Ok, works either way.

[SecureContext] just means that the parameter won't be populated if the page is served via an insecure http:// handler. The value won't then be present, and will return `undefined`.

(http://localhost URLs get a special case and is always considered secure).

src/library_webaudio.js
juj: Both versions of the function receive the handle ID? So in that case, I think it would be fine to call `_emscripten_audio_context_quantum_size()` from JS code as well. (need to add the underscore at the call site though, since the function won't start with a `$` prefix)

src/library_webaudio.js
cwoffenden: My concern was it'd throw an exception, not just silently return undefined.

(For localhost it still needs `Cross-Origin-Opener-Policy` and `Cross-Origin-Embedder-Policy` headers for audio demos to run, BTW, since `SharedArrayBuffer` isn't available and throws a reference error)

juj:
> I could mark `emscripten_create_wasm_audio_worklet_node()` as deprecated, add a new one that takes a different `EmscriptenWorkletNodeProcessCallback` signature, and let the two live side by side for a while? But let me do the minor tweaks here and land this first, I don't know when I'll have free time again.

Yeh, maybe breaking the signature of the audio callback won't be worth it, as the information can be acquired already in two different ways.

> I just verified to be sure, and the generated `myfile.aw.js` has no difference whether or not it's built with `--closure=1`. It has whitespace, etc., removed when building with `-O2` for example, but Closure doesn't appear to be run.

Ohh, this looks like an oversight. It definitely will be good to be Closured when --closure is passed. Though that's definitely something for another PR.

> Still I don't think you want to be doing minification of variable names or other things that closure generatelly does in this code.

Yeah, we don't need to manually minify any variable names (when Closure can do the minification for us). In cases where Closure cannot minify variable names (like the key names for postMessage boundaries), then we can do some custom tweaking (like we do with the `_wsc` things etc.)

src/library_webaudio.js
juj: > My concern was it'd throw an exception, not just silently return undefined.

Ah, right. An exception won't happen, that is guaranteed by the IDL spec of [SecureContext]: https://www.w3.org/TR/secure-contexts/#integration-idl which states multiple times

image

But let's leave this later for when an implementation comes along.

test/webaudio/audio_worklet_tone_generator.c
juj: Gotcha, that makes sense. 👍

juj (approved):

juj:
Thanks for working on this!

src/library_webaudio.js
cwoffenden: > Both versions of the function receive the handle ID?

I can tidy this in another PR.

cwoffenden:
> Thanks for working on this!

It's actually fun, so it's not really work!

src/library_webaudio.js
lindell: Work is now ongoing in Chromium to support renderSizeHint (configurable render quantum). It does not yet work for all Node types. But enough is available in Chrome Canary for testing with the `WebAudioConfigurableRenderQuantum` flag.
https://chromium-review.googlesource.com/q/bug:40637820

Should we change this to a non hardcoded version? 😄

cwoffenden: > Should we change this to a non hardcoded version? 😄

I'll look into it now this is testable.

cwoffenden: `{renderSizeHint: 512}` is settable and `AudioContext.renderQuantumSize` takes the value. I'll go ahead and add support.

cwoffenden: Created #25747 to expose this.

PR #22684, opened 2024-10-04
Revert conversion of test_utf32 from cpp to c.

brendandahl:
This test was failing asan after the conversion in #22669.

https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket/8735090587456463425/+/u/Emscripten_testsuite__ASan_/stdout

### DIFF
--- a/test/utf32.cpp
+++ b/test/utf32.cpp
@@ -4,9 +4,9 @@
 // found in the LICENSE file.
 
 #include <stdio.h>
-#include <string.h>
+#include <string>
 #include <emscripten.h>
-#include <assert.h>
+#include <cassert>
 #include <wchar.h>
 
 typedef unsigned int utf32;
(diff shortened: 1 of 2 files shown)


test/utf32.cpp
```diff
@@ -19,24 +19,23 @@ int main() {
...
   // U+20AC is the Euro sign,
   // U+2007C is a Chinese Han character that looks like three raindrops.
-  wchar_t wstr[] = L"abc\u2603\u20AC\U0002007C123 --- abc\u2603\u20AC\U0002007C123";
-  size_t wstr_len = wcslen(wstr);
+  std::wstring wstr = L"abc\u2603\u20AC\U0002007C123 --- abc\u2603\u20AC\U0002007C123";
 
   printf("sizeof(wchar_t): %d.\n", (int)sizeof(wchar_t));
 
   if (sizeof(wchar_t) == 4) {
-    utf32 *memory = malloc(wstr_len*sizeof(utf32));
```
kripken, line 28: Is maybe problem that we are missing the null terminator here, a +4 or such?

But lgtm to land this and figure it out later.

kripken (approved):

test/utf32.cpp
brendandahl: Sam was going to take a look when he got a chance. Though, I'm not sure I see the value in refactoring tests that use c++ features in c.

sbc100: Yeah I guess I was a little over zealous with this one, probably should have left it as C++.

Issue #22682, opened 2024-10-04
WebGPU Regression?


ypujante:
I upgraded to emscripten 3.1.68 (from 3.1.66) and my project [WebGPU Shader Toy](https://github.com/pongasoft/webgpu-shader-toy) which was working fine now fails.

I am seeing error messages like these:

```text
[WebGPU] Validation error | Write range (bufferOffset: 0, size: 1887248) does not fit in [Buffer "Dear ImGui Vertex buffer"] size (119040).
 - While calling [Queue].WriteBuffer([Buffer "Dear ImGui Vertex buffer"], (0 bytes), data, (1887248 bytes))
```

I can see from the list of PRs, that there was this one: #22494 (WebGPU: support Premultiplied canvas output)

I suppose my code can be doing something wrong (but I did not change my code, just the emscripten version). Does anybody know if this PR introduced a regression?

ypujante:
I just wanted to add that I built the demo (example_glfw_wgpu) that comes with ImGui latest version 1.91.3 and I am getting the exact same kind of error (so basically I have eliminated my code/layer)

```text
(index):41 Validation error: Write range (bufferOffset: 0, size: 592656) does not fit in [Buffer "Dear ImGui Vertex buffer"] size (115240).
(index):41  - While calling [Queue].WriteBuffer([Buffer "Dear ImGui Vertex buffer"], (0 bytes), data, (592656 bytes))
(index):41 
28Write range (bufferOffset: 0, size: 592656) does not fit in [Buffer "Dear ImGui Vertex buffer"] size (115240).
 - While calling [Queue].WriteBuffer([Buffer "Dear ImGui Vertex buffer"], (0 bytes), data, (592656 bytes))
Understand this warning
(index):41 Validation error: Write range (bufferOffset: 0, size: 547616) does not fit in [Buffer "Dear ImGui Index buffer"] size (22528).
(index):41  - While calling [Queue].WriteBuffer([Buffer "Dear ImGui Index buffer"], (0 bytes), data, (547616 bytes))
(index):41 
```

```text
> emcc --version
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.68 (ceee49d2ecdab36a3feb85a684f8e5a453dde910)
Copyright (C) 2014 the Emscripten authors (see AUTHORS.txt)
This is free and open source software under the MIT license.
[...]

PR #22689, opened 2024-10-07
Fix regression in i64 argument handling in webgpu+bigint

sbc100:
This was caused by #22557, which was checking the `LibraryWebGPU` object for `__sig` members when the sigs are stored in a separate file and therefore live on the global `LibraryManager.library` object.

Fixes: #22682

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -23,6 +23,8 @@ See docs/process.md for more on how version tagging works.
 - The usage of `EM_BOOL` in the emscripten API has been replaced with C/C++
   bool.  This change should not be observable since `EM_BOOL` has been
   equivalent to `bool` since #22157. (#22155)
+- Fix regression introduced in 3.1.67 (#22557) which broke webgpu / int64
+  integration. (#22689)
 
 3.1.68 - 09/30/24
 -----------------
--- a/src/library_webgpu.js
+++ b/src/library_webgpu.js
@@ -2883,9 +2883,9 @@ for (var value in LibraryWebGPU.$WebGPU.FeatureName) {
 
 for (const key of Object.keys(LibraryWebGPU)) {
   if (typeof LibraryWebGPU[key] === 'function') {
-    const sig = LibraryWebGPU[key + '__sig'];
+    const sig = LibraryManager.library[key + '__sig'];
     if (sig?.includes('j')) {
-      LibraryWebGPU[key + '__i53abi'] = true;
+      LibraryManager.library[key + '__i53abi'] = true;
     }
   }
 }
--- a/test/webgpu_basic_rendering.cpp
+++ b/test/webgpu_basic_rendering.cpp
@@ -225,6 +225,11 @@ void doCopyTestMappedAtCreation(bool useRange) {
         dst = device.CreateBuffer(&descriptor);
     }
 
+    // Write some random data to the buffer, just to verify that
+    // wgpuQueueWriteBuffer works.
+    char data[4];
+    queue.WriteBuffer(dst, 0, data, sizeof(data));
+
     wgpu::CommandBuffer commands;
     {
         wgpu::CommandEncoder encoder = device.CreateCommandEncoder();


sbc100:
Our webgpu testing is somewhat limited so I'm not sure how to go about testing this.  I'll give it a go though.

sbc100:
It looks like adding a test that included a call to `wgpuQueueWriteBuffer` should reproduce this.   Would it be easy to add a call to this to one of these existing webgpu browser tests @kainino0x ?

juj:
Ah right of course.. the issue wasn't about order, but about the pre vs post addition name.

LGTM.

Btw what makes me a bit sad about the autogeneration aspect of library_sigs.js is that it increases to that "rift" of two tiers of JS library files: i.e. Emscripten authors have a different "power" or capabilities to JS library files that Emscripten users don't. This results in Emscripten users sometimes not having as easy time to monkey-see/-do copy stuff into their JS libraries. Ideally the JS libraries we have in Emscripten repository should be written using the same machinery as what end users have at their disposal. (dogfooding, and ensuring that APIs are complete, etc.)

Though I definitely get the usefulness of autogenerating the sigs from the headers.

juj (approved):

sbc100:
> Ah right of course.. the issue wasn't about order, but about the pre vs post addition name.
> 
> LGTM.
> 
> Btw what makes me a bit sad about the autogeneration aspect of library_sigs.js is that it increases to that "rift" of two tiers of JS library files: i.e. Emscripten authors have a different "power" or capabilities to JS library files that Emscripten users don't. This results in Emscripten users sometimes not having as easy time to monkey-see/-do copy stuff into their JS libraries. Ideally the JS libraries we have in Emscripten repository should be written using the same machinery as what end users have at their disposal. (dogfooding, and ensuring that APIs are complete, etc.)
> 
> Though I definitely get the usefulness of autogenerating the sigs from the headers.

Agreed.   I think we should be able to make `tools/maint/gen_sig_info.py` usable for outside libraries perhaps?    Although its a little more tricky than `gen_struct_info.py` since the input C/C++ headers might need more complex setups (e.g. include paths and macros), in order to compile.

juj:
> It looks like adding a test that included a call to `wgpuQueueWriteBuffer` should reproduce this. Would it be easy to add a call to this to one of these existing webgpu browser tests @kainino0x ?

In [wasm_webgpu](https://github.com/juj/wasm_webgpu) I concocted this kind of a test for [wgpu_queue_write_buffer()](https://github.com/juj/wasm_webgpu/blob/main/test/wgpu_queue_write_buffer.cpp). Even though the bindings API is completely different, the same API call flow should work - maybe that can be used as inspiration.

ypujante:
@sbc100 Thanks for fixing the problem 

@juj Thanks for the thorough investigation

kainino0x:
> It looks like adding a test that included a call to `wgpuQueueWriteBuffer` should reproduce this. Would it be easy to add a call to this to one of these existing webgpu browser tests @kainino0x ?

If we only need to call it - and not check the buffer contents afterward - we could easily write to a buffer somewhere, e.g. here:
https://github.com/emscripten-core/emscripten/blob/main/test/webgpu_basic_rendering.cpp#L225-L226
use writeBuffer to put some data in the buffer, before overwriting it afterward.

sbc100:
> > It looks like adding a test that included a call to `wgpuQueueWriteBuffer` should reproduce this. Would it be easy to add a call to this to one of these existing webgpu browser tests @kainino0x ?
> 
> If we only need to call it - and not check the buffer contents afterward - we could easily write to a buffer somewhere, e.g. here: https://github.com/emscripten-core/emscripten/blob/main/test/webgpu_basic_rendering.cpp#L225-L226 use writeBuffer to put some data in the buffer, before overwriting it afterward.

Thanks, that worked.  I verified locally that this test now fails without this change.

BTW, we don't run that test as part of CI because we don't have access to webgpu on our circleci test bots.  Any ideas on how we could get this test running in headless mode on our linux CI bots?

sbc100:
Oops, I should probably have waited for an LGTM on that new test code.  Can you take a look @kainino0x ?

kainino0x:
Test LGTM!

PR #22696, opened 2024-10-08
Make second argument of UTF8ArrayToString optional. NFC

sbc100:
This is mostly a wash from a code size POV but makes the call sites more readable I think.


src/library_strings.js
```diff
@@ -20,14 +20,14 @@ addToLibrary({
...
    * @return {string}
    */`,
 #if TEXTDECODER
   $UTF8ArrayToString__deps: ['$UTF8Decoder'],
 #endif
-  $UTF8ArrayToString: (heapOrArray, idx, maxBytesToRead) => {
+  $UTF8ArrayToString: (heapOrArray, idx = 0, maxBytesToRead = NaN) => {
 #if CAN_ADDRESS_2GB
     idx >>>= 0;
 #endif
```
kripken, line 33: The comment on line 41 should replace "undefined" with "NaN"

src/source_map_support.js
```diff
@@ -103,7 +103,7 @@ if (!isDataURI(wasmSourceMapFile)) {
 
 function getSourceMap() {
   var buf = readBinary(wasmSourceMapFile);
-  return JSON.parse(UTF8ArrayToString(buf, 0, buf.length));
```
kripken, line 106: Is this change NFC? Before we would read `buf.length` bytes at most, but after we read without limit until the end of a UTF8 string. But is wasmSourceMapFile guaranteed to be encoded that way (including a null terminator)?

kripken (commented):
Readability seems worth it, I agree, but see my question below.

src/source_map_support.js
sbc100: The third argument here is `maxBytesToRead` which means will read at most `buf.length`. However `UTF8ArrayToString` will always stop when it reads a null byte I believe

kripken: But is there a null byte in this input? That is what concerns me. It is reading a source map file, which I don't think is encoded that way.

sbc100: Oh I assumed UTF8ArrayToString would not read read past the end of the input array!  I don't see that check though.  That's strange.

sbc100: Oh wait I think its fine because if you read of the end of a array or buffer in JS you get `undefined` back, so every buffer is effectively null terminated.

kripken: Ah, good point. That looks safe then.

kripken (approved):

PR #22698, opened 2024-10-08
Fix test_utf32. NFC

sbc100:
This test was never actually being run with `-fshort-wchar` because it was being passed as `args` rather than `emcc_args`, ever since the test was added back in c33af789.

When actually running with `-fshort-wchar` the test fails under the sanitizers I believe because wcslen (which is part of libc) would also need to be built with this flag.


kripken (approved):
Oh wow, good catch.

brendandahl (approved):

PR #22702, opened 2024-10-09
Allow MapMode.Read in wgpuBufferGetMappedRange

alexandermarinissen:
Emscripten's distribution of the WebGPU API prohibits the usage of MapMode.Read in wgpuBufferGetMappedRange, even though this is allowed in the C++ spec.

This would be useful considering that wgpuBufferGetConstMappedRange is not present in some distributions of WebGPU.
In fact, both wgpuBufferGetMappedRange and wgpuBufferGetConstMappedRange already forward their calls to bufferWrapper.object.getMappedRange.

This pull request makes wgpuBufferGetConstMappedRange behave like wgpuBufferGetMappedRange when it is called with MapMode.Read.

### DIFF
--- a/src/library_webgpu.js
+++ b/src/library_webgpu.js
@@ -2052,13 +2052,14 @@ var LibraryWebGPU = {
     }
 
     var data = _memalign(16, mapped.byteLength);
-    HEAPU8.fill(0, data, mapped.byteLength);
     if (bufferWrapper.mapMode === {{{ gpu.MapMode.Write }}}) {
+      HEAPU8.fill(0, data, mapped.byteLength);
       bufferWrapper.onUnmap.push(() => {
         new Uint8Array(mapped).set(HEAPU8.subarray(data, data + mapped.byteLength));
         _free(data);
       });
     } else {
+      HEAPU8.set(new Uint8Array(mapped), data);
       bufferWrapper.onUnmap.push(() => _free(data));
     }
     return data;
--- a/src/library_webgpu.js
+++ b/src/library_webgpu.js
@@ -2032,9 +2032,9 @@ var LibraryWebGPU = {
 
     {{{ gpu.convertSentinelToUndefined('size') }}}
 
-    if (bufferWrapper.mapMode !== {{{ gpu.MapMode.Write }}}) {
+    if (bufferWrapper.mapMode !== {{{ gpu.MapMode.Read }}} && bufferWrapper.mapMode !== {{{ gpu.MapMode.Write }}}) {
 #if ASSERTIONS
-      abort("GetMappedRange called, but buffer not mapped for writing");
+      abort("GetMappedRange called, but buffer not mapped for reading or writing");
 #endif
       // TODO(kainino0x): Somehow inject a validation error?
       return 0;
@@ -2053,10 +2053,14 @@ var LibraryWebGPU = {
 
     var data = _memalign(16, mapped.byteLength);
     HEAPU8.fill(0, data, mapped.byteLength);
-    bufferWrapper.onUnmap.push(() => {
-      new Uint8Array(mapped).set(HEAPU8.subarray(data, data + mapped.byteLength));
-      _free(data);
-    });
+    if (bufferWrapper.mapMode === {{{ gpu.MapMode.Write }}}) {
+      bufferWrapper.onUnmap.push(() => {
+        new Uint8Array(mapped).set(HEAPU8.subarray(data, data + mapped.byteLength));
+        _free(data);
+      });
+    } else {
+      bufferWrapper.onUnmap.push(() => _free(data));
+    }
     return data;
   },
 


src/library_webgpu.js
```diff
@@ -2052,13 +2052,14 @@ var LibraryWebGPU = {
     }
 
     var data = _memalign(16, mapped.byteLength);
-    HEAPU8.fill(0, data, mapped.byteLength);
     if (bufferWrapper.mapMode === {{{ gpu.MapMode.Write }}}) {
+      HEAPU8.fill(0, data, mapped.byteLength);
```
sbc100, line 2056: Can you replace this with a call to `zeroMemory` to be more explicit?    (you will also need to add `$zeroMemory` to the `__deps` of this function).

kainino0x:
> Emscripten's distribution of the WebGPU API prohibits the usage of MapMode.Read in wgpuBufferGetMappedRange, even though this is allowed in the C++ spec.

It should not be allowed, it's "unsafe" in the sense that it can return a mutably-typed pointer to a constant memory allocation, causing undefined behavior.

What C++ implementation allows this? I'm pretty sure Dawn does not. (There is no C++ spec, only a [C spec](https://github.com/webgpu-native/webgpu-headers), which is not complete yet.)

> This would be useful considering that wgpuBufferGetConstMappedRange is not present in some distributions of WebGPU.

Also which implementation is this? It would be ideal if they implement it. In the meantime, you could work around the differences with `#ifdef __EMSCRIPTEN__`.

alexandermarinissen:
> It should not be allowed, it's "unsafe" in the sense that it can return a mutably-typed pointer to a constant memory allocation, causing undefined behavior.

I'm not too familiar with the JavaScript functions being used in the parts that I edited. Isn't the `data` variable an entirely new byte array that it is freed again when the buffer is unmapped, so don't we fully own it (for the duration that the buffer is mapped)? Maybe I'm just not seeing why this would cause undefined behavior: what makes this pointer a constant allocation? If it's a fresh byte array that's gonna get cleaned up without a second look, what's the harm in mutating it?

> What C++ implementation allows this? I'm pretty sure Dawn does not. (There is no C++ spec, only a [C spec](https://github.com/webgpu-native/webgpu-headers), which is not complete yet.)

wgpu-native has no problems when using wgpuBufferGetMappedRange after mapping the buffer with read mode. I interpreted this as "it's allowed", but I suppose this is just a lack of asserts on their end?

> > This would be useful considering that wgpuBufferGetConstMappedRange is not present in some distributions of WebGPU.
> 
> Also which implementation is this? It would be ideal if they implement it. In the meantime, you could work around the differences with `#ifdef __EMSCRIPTEN__`.

Sorry for the confusion, I meant to say that buffer.getConstMappedRange is not available in (all?) browsers, e.g. it's not present in Firefox's implementation (https://developer.mozilla.org/en-US/docs/Web/API/GPUBuffer)

src/library_webgpu.js
alexandermarinissen: I'm not very familiar with this code, what does HEAPU8.fill do versus zeroMemory?
Note: the HEAPU8.fill was already there in the Write case, this is essentially unchanged.

sbc100: No need to include that change in this PR if you don't want to..  I can make a separate PR to use zeroMemory here.

kainino0x:
> I'm not too familiar with the JavaScript functions being used in the parts that I edited. Isn't the `data` variable an entirely new byte array that it is freed again when the buffer is unmapped, so don't we fully own it (for the duration that the buffer is mapped)? Maybe I'm just not seeing why this would cause undefined behavior: what makes this pointer a constant allocation? If it's a fresh byte array that's gonna get cleaned up without a second look, what's the harm in mutating it?

It would be safe in Emscripten's current implementation, but the API contract says that it wouldn't be safe - for example in native (Dawn/wgpu-native), or if Wasm gets read-only memory pages or something, it would really be invalid. So we shouldn't allow it, to avoid applications relying accidentally on behavior that is likely to change, and doesn't work on native builds.

> wgpu-native has no problems when using wgpuBufferGetMappedRange after mapping the buffer with read mode. I interpreted this as "it's allowed", but I suppose this is just a lack of asserts on their end?

Ah, that makes sense. Yes, that is just a place where the webgpu.h implementations haven't converged yet (we're still working on it). We discussed this in https://github.com/webgpu-native/webgpu-headers/issues/194 but haven't finished documenting it yet.

> Sorry for the confusion, I meant to say that buffer.getConstMappedRange is not available in (all?) browsers, e.g. it's not present in Firefox's implementation (https://developer.mozilla.org/en-US/docs/Web/API/GPUBuffer)

`GetConstMappedRange` is a C/"native language"-only thing, as JavaScript doesn't have read-only ArrayBuffers (yet?). As you pointed out, Emscripten just implements it over a regular allocation right now, but this may change in the future so it's important for the C `GetMappedRange` and `GetConstMappedRange` APIs to behave differently even in Emscripten.

src/library_webgpu.js
alexandermarinissen: A separate PR would be better, thanks :)

I'm not comfortable enough with the codebase to add dependencies myself

alexandermarinissen:
Gotcha, that makes sense. I'll use the const version from now on and just make a copy if I really need to take ownership of it to mutate it within my application.

kainino0x:
Makes sense - sorry the buffer mapping story in Wasm right now is not very efficient. We've considered something like a `readBuffer` to copy data into a user-allocated pointer which might work better. Thank you for the contribution!

alexandermarinissen:
Yeah, trying to have as few allocs as possible. Thanks for the feedback and all that you do for emscripten!

PR #22704, opened 2024-10-09
[webgpu] Remove redundant check in initManagers. NFC

sbc100:
initManagers already has a reentrancey check at the top (which itself I believe in not necessary).


kainino0x (approved):
LGTM (fyi @lokokung - we probably don't have this anymore in the fork but just in case)

PR #22707, opened 2024-10-09
[embind] Replace trivial use of Object.defineProperty. NFC

sbc100:
All this function was doing was exporting a bunch of library methods. If folks want to export these functions they can do it themselves using the normal mechanisms.

### DIFF
--- a/src/embind/embind.js
+++ b/src/embind/embind.js
@@ -180,18 +180,16 @@ var LibraryEmbind = {
     return errorClass;
   },
 
-  $createNamedFunction: (name, body) => Object.defineProperty(body, 'name', {
-    value: name
-  }),
+  $createNamedFunction: (name, func) => Object.defineProperty(func, 'name', { value: name }),
+#if MIN_CHROME_VERSION < 43 || MIN_SAFARI_VERSION < 100101 || MIN_FIREFOX_VERSION < 38
   // All browsers that support WebAssembly also support configurable function name,
   // but we might be building for very old browsers via WASM2JS.
-#if MIN_CHROME_VERSION < 43 || MIN_SAFARI_VERSION < 100101 || MIN_FIREFOX_VERSION < 38
   // In that case, check if configurable function name is supported at init time
   // and, if not, replace with a fallback that returns function as-is as those browsers
   // don't support other methods either.
   $createNamedFunction__postset: `
     if (!Object.getOwnPropertyDescriptor(Function.prototype, 'name').configurable) {
-      createNamedFunction = (name, body) => body;
+      createNamedFunction = (name, func) => func;
     }
   `,
 #endif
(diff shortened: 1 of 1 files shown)


src/embind/embind.js
```diff
@@ -180,18 +180,16 @@ var LibraryEmbind = {
     return errorClass;
   },
 
-  $createNamedFunction: (name, body) => Object.defineProperty(body, 'name', {
-    value: name
-  }),
+  $createNamedFunction: (name, func) => Object.defineProperty(func, 'name', { value: name }),
```
sbc100, line 183: This was just a driveby fix change the name of the second argument to something more accurate.

src/embind/embind.js
```diff
@@ -2226,9 +2224,7 @@ var LibraryEmbind = {
         }
       }.bind(this));
 
-      Object.defineProperty(this, '__parent', {
-        value: wrapperPrototype
-      });
+      this["__parent"] = wrapperPrototype;
```
brendandahl, line 2227: Using `Object.defineProperty` has different default behavior than assigning a property i.e. with defineProperty, configurable and writable default to false, whereas it's the opposite with regular assignment. I'm not sure we should change this since it's been like this for so long.

sbc100: I don't think there is any reason for it to be readonly is here?   Its not performance since this property is not eve used internally anywhere.

There are some downside to using `defineProperty` here though:

1. Its more verbose and makes all programs larger
2. Its confusing to the reader since it makes you think something more is going on.

I don't think "been like this for so long" is great reason not to change something.  At the very least this code warrants an explanatory comment.

I tracked down the original change to 2014: 9759bdbc104d860e8eadbb0f35e0d3dedb48035d. Can you figure out why this change was made?

sbc100: It seems strange to keep this around if we can't find any reason for it...   there should at least be a comment here.

brendandahl: Oh I bet I know what the reason was... the other difference is 'enumerable' is `false` for `defineProperty`. That way if you're doing `for (var key in instance) { if (instance.hasOwnProperty(key)) ...` you won't see the `__parent` property.

sbc100: Ah I see.   Do we think that hiding `$$` and `__parent` like this makes sense?

I still wonder why the PR title was `checkpoint returning JS instances out of embind`?   What does checkpoint mean do you think?

brendandahl: Yeah, it probably makes sense to hide those since you wouldn't want to iterate over them. If we convert these to es6 classes `super` will be hidden like this.

As for checkpoint, I think that was just a commit to save partial work. If you look at the PR #2371 it has tons of commits.

sbc100: OK, I guess we should make a different PR to add a comment here?   And perhaps at a test of `getOwnPropertyNames` to ensure it doesn't include these?

PR #22710, opened 2024-10-10
Allow pthread programs to run without SharedArrayBuffer

sbc100:
As long as they don't actually try to start any threads there is no
reason to prevent them from running/starting.

Such programs will now instead fail when they first try to create
a thread.

This is useful for programs that are built with threading support but
might also want to run in environments without SharedArrayBuffer,
e.g. when deployed without COOP/COEP.


test/test_other.py
```diff
@@ -12065,6 +12065,18 @@ def test_pthread_hello(self, args):
   def test_pthread_relocatable(self):
     self.do_run_in_out_file_test('hello_world.c', emcc_args=['-sRELOCATABLE'])
 
+  @node_pthreads
+  def test_pthread_unavailable(self):
+    # Run a simple hello world program that uses pthreads
+    self.emcc_args += ['-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME']
```
kripken, line 12071: Why EXIT_RUNTIME?

test/test_other.py
```diff
@@ -12065,6 +12065,18 @@ def test_pthread_hello(self, args):
...
+    self.do_run_in_out_file_test('hello_world.c')
+
+    # Now run the same program but with SharedArrayBuffer undefined, it should run
+    # fine and then fail on the first call to pthread_create.
+    create_file('pre.js', 'SharedArrayBuffer = undefined\n')
+    expected = 'pthread_create: environment does not support SharedArrayBuffer, pthreads are not available'
+    self.do_runf('hello_world.c', expected, assert_returncode=NON_ZERO, emcc_args=['--pre-js=pre.js'])
```
kripken, line 12078: Maybe it's worth adding a specific test here that verifies we can do things up until the pthread creation? Like a program with main
```c
int main() {
  printf("This is printed\n");
  pthread_create(..)
  printf("This is not printed\");
}
```
Then we can verify we error at the `pthread_create`, but managed to print the first line.

sbc100: I think that fact that the error message starts with `pthread_create:` is enough to verify this.

test/test_other.py
sbc100: For now this is needed for all node tests that use `PROXY_TO_PTHREAD`.   I've for a fix in flight in #22445 but I'm still working out some issues.

kripken (approved):
Fair enough on both points.

PR #22711, opened 2024-10-10
Disable test_utf32_short_wchar under LTO. NFC

sbc100:


dschuff (approved):
LGTM (I think this is the case, but just to double-check: the failure mode here is a link failure that might be understandable by a user, and not an inscrutable runtime failure, right?)

sbc100:
> LGTM (I think this is the case, but just to double-check: the failure mode here is a link failure that might be understandable by a user, and not an inscrutable runtime failure, right?)

Its link-time error:

```
wasm-ld: error: linking module flags 'wchar_size': IDs have conflicting values in '/usr/local/google/home/sbc/dev/wasm/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libc.a(__stdio_exit.o at 277994)' and 'ld-temp.o'
```

PR #22712, opened 2024-10-10
Use zeroMemory over HEAPU8.fill. NFC

sbc100:


kripken:
> Also, add missing return statement to the legacy path in zeroMemory.

I don't see that?

src/library_webgpu.js
```diff
@@ -2052,7 +2052,7 @@ var LibraryWebGPU = {
     }
 
     var data = _memalign(16, mapped.byteLength);
-    HEAPU8.fill(0, data, mapped.byteLength);
+    zeroMemory(data, mapped.byteLength);
```
kripken, line 2055: This does not look NFC, as before we did
```js
HEAPU8.fill(0, data, mapped.byteLength);
```
while after we are effectively doing
```js
HEAPU8.fill(0, data, data + mapped.byteLength);
```
Perhaps it was wrong before?

kainino0x: Oops, it was definitely wrong before.

kainino0x (approved):
LGTM, but same comment about the return statement.

sbc100:
> > Also, add missing return statement to the legacy path in zeroMemory.
> 
> I don't see that?

Reverted

PR #22715, opened 2024-10-10
[embind] Fix typo in overloadTable error checking. NFC

sbc100:


src/embind/embind.js
```diff
@@ -124,7 +124,7 @@ var LibraryEmbind = {
       // We are exposing a function with the same name as an existing function. Create an overload table and a function selector
       // that routes between the two.
       ensureOverloadTable(Module, name, name);
-      if (Module.hasOwnProperty(numArguments)) {
+      if (Module[name].overloadTable.hasOwnProperty(numArguments)) {
```
kripken, line 127: I'm not familiar with this code, but it seems right. Was no test failing, though? Can we test it, if not?

sbc100: It looks like we don't have test for this since its always been broken AFAICT.  @brendandahl how hard would it be to write a test for this?  Can you point me to the place to do it?

brendandahl: It shouldn't be too hard, but I don't think it can be part of the main embind tests since it's an error during initialization. It'll need to be its own little test. 
```
int main() {}

int foo(int a) {
  return 0;
}

EMSCRIPTEN_BINDINGS(xxx) {
  function("foo", &foo);
  function("foo", &foo);
}
```

sbc100: I added a test thanks!

In the test I actually attempt to register two distinct overloads which happen to both take a single argument.

brendandahl (approved):

sbc100:
This is now a minor codesize win!

kripken (approved):

PR #22721, opened 2024-10-11
Fix wasm workers under node

sbc100:
- Use callUserCallback to invoke callback in _wasmWorkerRunPostMessage.
  Without this calls to exit/emscripten_force_exit within the callback
  don't work as expected (they cause unhandled exception errors).
- Fix `onmessage` handling under node so that the message payload always
  arrives as the `data` member of the message.
- Update a few of the wasm workers tests do they actually exit (required
   for running tests under node).

### DIFF
--- a/src/wasm_worker.js
+++ b/src/wasm_worker.js
@@ -14,7 +14,19 @@ if (ENVIRONMENT_IS_NODE) {
 
   var parentPort = nodeWorkerThreads.parentPort;
 
-  parentPort.on('message', (data) => typeof onmessage === "function" && onmessage({ data: data }));
+  parentPort.on('message', (msg) => global.onmessage?.({ data: msg }));
+
+  // Weak map of handle functions to their wrapper. Used to implement
+  // addEventListener/removeEventListener.
+  var wrappedHandlers = new WeakMap();
+  function wrapMsgHandler(h) {
+    var f = wrappedHandlers.get(h)
+    if (!f) {
+      f = (msg) => h({data: msg});
+      wrappedHandlers.set(h, f);
+    }
+    return f;
+  }
 
   var fs = require('fs');
   var vm = require('vm');
@@ -28,8 +40,8 @@ if (ENVIRONMENT_IS_NODE) {
     importScripts: (f) => vm.runInThisContext(fs.readFileSync(f, 'utf8'), {filename: f}),
     postMessage: (msg) => parentPort.postMessage(msg),
     performance: global.performance || { now: Date.now },
-    addEventListener: (name, handler) => parentPort.on(name, handler),
-    removeEventListener: (name, handler) => parentPort.off(name, handler),
+    addEventListener: (name, handler) => parentPort.on(name, wrapMsgHandler(handler)),
+    removeEventListener: (name, handler) => parentPort.off(name, wrapMsgHandler(handler)),
   });
 }
 #endif // ENVIRONMENT_MAY_BE_NODE
--- a/src/library_wasm_worker.js
+++ b/src/library_wasm_worker.js
@@ -55,18 +55,13 @@ addToLibrary({
   },
 
   // Executes a wasm function call received via a postMessage.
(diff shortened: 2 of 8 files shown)


src/wasm_worker.js
```diff
@@ -14,7 +14,13 @@ if (ENVIRONMENT_IS_NODE) {
 
   var parentPort = nodeWorkerThreads.parentPort;
 
-  parentPort.on('message', (data) => typeof onmessage === "function" && onmessage({ data: data }));
+  parentPort.on('message', (msg) => global.onmessage?.({ data: msg }));
+
+  var wrappedHandlers = {}
```
kripken, line 19:
```suggestion
  var wrappedHandlers = {};
```

src/wasm_worker.js
```diff
@@ -14,7 +14,13 @@ if (ENVIRONMENT_IS_NODE) {
...
+  function wrapMsgHandler(h) {
+    wrappedHandlers[h] ??= (msg) => h({data: msg});
+    return wrappedHandlers[h];
+  }
```
kripken, line 23: What is this function doing? I can't figure it out.

And a specific question, `wrappedHandlers` is an object, and `h` is a function, so does `wrappedHandlers[h]` not stringify the function to generate a key for the object..? Here is what I mean:
```js
> x = {}
{}
> foo = function foo(x) { return x + 1 }
[Function: foo]
> x[foo] = 10
10
> x
{ 'function foo(x) { return x + 1 }': 10 }
```

sbc100: I'm trying to keep a cache of the functions that have been wrapped.

This is because the function I pass to `parentPort.on` has to match the function I pass to `parentPort.off`.   If we create a new wrapper each time then they will never match and `parentPort.off` will not work so `removeEventListener` won't work either.

I guess I can't use an object though and I should instead use a Set?

kripken: A WeakMap might be best here. That could associate a wrapper with a function, without keeping the function alive.

sbc100: Done

kripken (approved):

juj (approved):
LGTM. Though it is not precisely clear to me which part of this PR was the functional fix, and which part of the PR is nonfunctional/cleanup, since there are multiple orthogonal changes layered to this PR. (in particular, was refactoring to use `callUserCallback()` a needed correctness fix for Node, and was removing use of dynCalls in test_wasm_worker_terminate test necessary for Node?)

I feel a bit that if I proposed this PR, it might get reviewed to need to be split to multiple parts with nonfunctional and functional changes separate 😅 - but since I do argue towards having fast development ergonomy and flow, I think it's fine.

src/wasm_worker.js
```diff
@@ -14,7 +14,19 @@ if (ENVIRONMENT_IS_NODE) {
 
   var parentPort = nodeWorkerThreads.parentPort;
 
-  parentPort.on('message', (data) => typeof onmessage === "function" && onmessage({ data: data }));
+  parentPort.on('message', (msg) => global.onmessage?.({ data: msg }));
+
+  // Weak map of handle functions to thier wrapper. Used to implement
```
juj, line 19: their

sbc100:
> LGTM. Though it is not precisely clear to me which part of this PR was the functional fix, and which part of the PR is nonfunctional/cleanup, since there are multiple orthogonal changes layered to this PR. (in particular, was refactoring to use `callUserCallback()` a needed correctness fix for Node, and was removing use of dynCalls in test_wasm_worker_terminate test necessary for Node?)
> 
> I feel a bit that if I proposed this PR, it might get reviewed to need to be split to multiple parts with nonfunctional and functional changes separate 😅 - but since I do argue towards having fast development ergonomy and flow, I think it's fine.

I pretty sure everything here in the PR is essential to get the tests running under node.  Sorry I should have documented in the PR description why each change is needed.  I will do that now.

sbc100:
I updated the PR description to detail why these changes are needed and split out some parts into cleanup PRs.

kripken (approved):

PR #22734, opened 2024-10-14
[embind] Better reflect the actual class hierarchy in TS.

### LINKED ISSUE #22701: Add deleteLater() and setDelayFunction(...) to TypeScript definitions and make them official API
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.68-git
clang version 20.0.0git
Thread model: posix
InstalledDir: /opt/homebrew/Cellar/emscripten/3.1.68/libexec/llvm/bin
```

We are using the `deleteLater()` function a lot in our TypeScript codebase to not forget to delete a short lived reference. This works quite well. The only drawback is that we have to patch the generated TypeScript definitions to include the `deleteLater()` function for each interface that also has the `delete()` function. And we patch the `MainModule` export to include the `setDelayFunction(...)` definition, so that we can set the delay function at MainModule instantiation.

It would be great, if these two could be added to the generated TypeScript definitions and also be made an official API.

brendandahl:
- All classes now extend from ClassHandle, which is what really happens in JS.
 - Add the missing methods that are available on ClassHandles.

Fixes #22701

### DIFF
--- a/src/embind/embind_gen.js
+++ b/src/embind/embind_gen.js
@@ -415,7 +415,7 @@ var LibraryEmbind = {
           'export interface ClassHandle {\n',
           '  isAliasOf(other: ClassHandle): boolean;\n',
           '  delete(): void;\n',
-          '  deleteLater(): void;\n',
+          '  deleteLater(): this;\n',
           '  isDeleted(): boolean;\n',
           '  clone(): this;\n',
           '}\n',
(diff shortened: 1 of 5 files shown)


sbc100 (approved):
Nice!

arnog-ms:
I think deleteLater() returns a pointer to the object itself, doesn't it? Otherwise you couldn't chain calls. So for the TypeScript definition this would be `this`.

PR #22735, opened 2024-10-14
Split the pthread library out of libc. NFC

sbc100:
The idea here is that, as a followup we could then collapse libc-mt and libc-ww into a single libc-mt that didn't have any specific pthread stuff in it.


sbc100:
@eyebrowsoffire

test/test_other.py
```diff
@@ -5995,9 +5995,9 @@ def test_bad_lookup(self):
     # partial list, but ok since we grab them as needed
     'parial': [{'EMCC_FORCE_STDLIBS': 'libc++'}, False],
     # fail! not enough stdlibs
-    'partial_only': [{'EMCC_FORCE_STDLIBS': 'libc++,libc,libc++abi', 'EMCC_ONLY_FORCED_STDLIBS': '1'}, True],
+    'partial_only': [{'EMCC_FORCE_STDLIBS': 'libc++,libc,libpthread,libc++abi', 'EMCC_ONLY_FORCED_STDLIBS': '1'}, True],
```
kripken, line 5998: Worth a changelog mention because this is observable by some users.

kripken (approved):

sbc100:
This change still needs some work.

Issue #22736, opened 2024-10-13
EM_COMPILER_WRAPPER broken from 3.1.51


slowriot:
From 3.1.51 onwards, using `EM_COMPILER_WRAPPER` fails with a slew of python errors when downloading and building ports.

I typically use this `EM_COMPILER_WRAPPER` to wrap the build commands in `ccache`.

In the example below it's zlib, but this seems to happen for all build steps of all projects I tested.  Version 3.1.50 works fine with the same commands.

steps to reproduce:
```sh
emsdk install 3.1.51
emsdk activate 3.1.51

# for any cmake project:
mkdir build
emcmake cmake ..
EM_COMPILER_WRAPPER=ccache emmake make
```

example output:

<details>
  <summary>Click to expand</summary>

make: make
[  3%] Building CXX object CMakeFiles/client.dir/main.cpp.o
shared:INFO: (Emscripten: Running sanity checks)
ports:INFO: retrieving port: zlib from https://github.com/madler/zlib/archive/refs/tags/v1.2.13.tar.gz
ports:INFO: unpacking port: zlib
cache:INFO: generating port: sysroot/lib/wasm32-emscripten/libz.a... (this will be cached in "/usr/src/emscripten/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/libz.a" for subsequent builds)
Traceback (most recent call last):
Traceback (most recent call last):
  File "/usr/src/emscripten/emsdk/upstream/emscripten/emcc.py", line 1605, in <module>
Traceback (most recent call last):
  File "/usr/src/emscripten/emsdk/upstream/emscripten/emcc.py", line 1605, in <module>
  File "/usr/src/emscripten/emsdk/upstream/emscripten/emcc.py", line 1605, in <module>
Traceback (most recent call last):
  File "/usr/src/emscripten/emsdk/upstream/emscripten/emcc.py", line 1605, in <module>
Traceback (most recent call last):
  File "/usr/src/emscripten/emsdk/upstream/emscripten/emcc.py", line 1605, in <module>
    sys.exit(main(sys.argv))
    sys.exit(main(sys.argv))
             ^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/contextlib.py", line 81, in inner
             ^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/contextlib.py", line 81, in inner
    sys.exit(main(sys.argv))
    sys.exit(main(sys.argv))
[...]

PR #13498, opened 2021-02-14
Ccache

juj:
This PR adds support to ccache for Emscripten. The corresponding PR for emsdk is at https://github.com/emscripten-core/emsdk/pull/711 and the corresponding changes to ccache itself are at https://github.com/juj/ccache/tree/emscripten.

The general mechanism is that after activating ccache in Emscripten SDK, emcc and em++ will automatically take usage of the installed ccache tool.

Usage:

```
emsdk install sdk-upstream-master-64bit ccache-git-emscripten-64bit
emsdk activate sdk-upstream-master-64bit  ccache-git-emscripten-64bit
cd %EMSCRIPTEN%
git remote add juj https://github.com/juj/emscripten.git
git fetch emscripten
git checkout ccache

# Example 1: rebuilding libc is fast from cache:
ccache -s # check initial ccache statistics: zero compiled files in cache
emcc --clear-cache

python embuilder.py build libc # cache miss
ccache -s # will show 862 files missed cache

ccache -z # zero ccache statistics

emcc --clear-cache
python embuilder.py build libc # cache hit!
ccache -s # now shows 862 files found from cache

# Example 2: rebuilding custom code is fast from cache:
ccache -z # zero statistics
emcc -c tests\hello_world.c -o a.o # cache miss
ccache -s # shows one file missed cache
emcc -c tests\hello_world.c -o a.o # cache hit!
ccache -s # shows one file hit cache
```

PR #22737, opened 2024-10-14
Use execvp rather then execv in exec_process

sbc100:
When this code was added in #20909 it broke users who were doing `EM_COMPILER_WRAPPER=cache` since it was not using PATH to find the executable (unlike the subprocess module).

Fixes: #22736

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -787,7 +787,7 @@ jobs:
       EMTEST_SKIP_NODE_CANARY: "1"
       EMTEST_SKIP_WASM64: "1"
     steps:
-      - run: apt-get install -q -y ninja-build scons
+      - run: apt-get install -q -y ninja-build scons ccache
       - run-tests-linux:
           # some native-dependent tests fail because of the lack of native
           # headers on emsdk-bundled clang
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -154,6 +154,18 @@ def requires_ninja(func):
   return decorated
 
 
+def requires_ccache(func):
+  assert callable(func)
+
+  @wraps(func)
+  def decorated(self, *args, **kwargs):
+    if not shutil.which('ccache'):
+      self.fail('test requires ccache to be installed (available in PATH)')
+    return func(self, *args, **kwargs)
+
+  return decorated
+
+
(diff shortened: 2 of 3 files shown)


kripken (approved):

kripken:
(do we have ccache on the CI machines?)

kleisauke:
Just wondering, is `EM_COMPILER_WRAPPER=ccache` recommended over `_EMCC_CCACHE=1`? I've been using the latter (internal) setting for a while now:
https://github.com/kleisauke/wasm-vips/blob/v0.0.10/Dockerfile#L45-L47

sbc100:
> Just wondering, is `EM_COMPILER_WRAPPER=ccache` recommended over `_EMCC_CCACHE=1`? I've been using the latter (internal) setting for a while now: https://github.com/kleisauke/wasm-vips/blob/v0.0.10/Dockerfile#L45-L47

Hmm.. I can't remember exactly that `EMCC_CCACHE` was needed, but it works slightly differently.  `EM_COMPILER_WRAPPER` can be used to inject any compiler wrapper, not just ccache.   It gets injected inside the python driver and it adds the wrapper to the internal clang calls.

`_EMCC_CCACHE` inject explicitly `cache` only and it works in the shell/bat files and adds the wrapper before `emcc` rather than `clang`.  So if you you use `_EMCC_CCACHE` then ccache itself will see `emcc` as the compiler but if you use `EM_COMPILER_WRAPPER` then ccache will see `clang` as the compiler.  I think `_EMCC_CCACHE` might require a patched version of ccache?

kleisauke:
Thanks for the information!

> I think `_EMCC_CCACHE` might require a patched version of ccache?

It seems to work without any patches, though I haven't tested it extensively. emsdk also provides a patched version of ccache through the `ccache-git-emscripten-64bit` target.
https://github.com/emscripten-core/emsdk/blob/3.1.69/emsdk_manifest.json#L678-L688

(the same question was raised in https://github.com/emscripten-core/emscripten/pull/13498#issuecomment-1364551716)

sbc100:
Sounds like performance is that reason to prefer _EMCC_CCACHE.  I do wonder if the difference is still as significant these days.

PR #22744, opened 2024-10-15
Fix `__errno_location` under wasm workers. NFC

sbc100:
Previously the wasm workers build of `__errno_location` was using the
single threaded code path which doesn't use a thead local location.

In addition this change avoids the use of `__EMSCRIPTEN_PTHREADS__` in
`__errno_location.c` which is helpful for unifying the shared memory
libc build.  See #22735.

### DIFF
--- a/system/lib/libc/musl/src/errno/__errno_location.c
+++ b/system/lib/libc/musl/src/errno/__errno_location.c
@@ -1,22 +1,19 @@
 #include <errno.h>
 #include "pthread_impl.h"
 
-#if __EMSCRIPTEN_PTHREADS__
-// for pthreads, use the proper location on the thread info, so each
-// thread has its own errno
-int *__errno_location(void)
-{
-	return &__pthread_self()->errno_val;
-}
-#else
-// for single-threaded mode, avoid linking in pthreads support code
-// just for this
-static int __errno_storage = 0;
+#if __EMSCRIPTEN__
+// For emscripten we use TLS here instead of `__pthread_self`, so that in single
+// threaded builds this gets lowered away to normal global variable.
+static _Thread_local int __errno_storage = 0;
+#endif
 
 int *__errno_location(void)
 {
+#if __EMSCRIPTEN__
 	return &__errno_storage;
-}
+#else
+	return &__pthread_self()->errno_val;
 #endif
+}
 
 weak_alias(__errno_location, ___errno_location);
--- a/test/code_size/hello_wasm_worker_wasm.json
+++ b/test/code_size/hello_wasm_worker_wasm.json
@@ -5,8 +5,8 @@
   "a.js.gz": 455,
   "a.ww.js": 115,
   "a.ww.js.gz": 127,
-  "a.wasm": 1850,
-  "a.wasm.gz": 1050,
-  "total": 3248,
-  "total_gz": 2016
+  "a.wasm": 1894,
+  "a.wasm.gz": 1077,
+  "total": 3292,
+  "total_gz": 2043
 }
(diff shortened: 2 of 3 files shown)


sbc100:
@eyebrowsoffire

sbc100:
@eyebrowsoffire I'm running into the fact the `-sSHARED_MEMORY` builds don't think here..  :(   I guess we need your change first?

system/lib/libc/musl/src/errno/__errno_location.c
```diff
@@ -1,22 +1,19 @@
...
-{
-	return &__pthread_self()->errno_val;
-}
-#else
-// for single-threaded mode, avoid linking in pthreads support code
-// just for this
-static int __errno_storage = 0;
+#if __EMSCRIPTEN__
+// For emscripten we use TLS here instead of `__pthread_self`, so that in single
+// threaded builds this get lowered away to normal global variable.
```
kripken, line 6:
```suggestion
// threaded builds this gets lowered away to a normal global variable.
```

test/test_sockets.py
```diff
@@ -328,7 +328,7 @@ def test_nodejs_sockets_echo_subprotocol(self):
   # NOTE: Shared buffer is not allowed for websocket sending.
   @parameterized({
     '': [[]],
-    'shared': [['-sSHARED_MEMORY']],
+    'pthread': [['-pthread']],
```
kripken, line 331: Is this change related to errno?

kripken (approved):

test/code_size/hello_wasm_worker_wasm.json
```diff
@@ -5,8 +5,8 @@
   "a.js.gz": 455,
   "a.ww.js": 115,
   "a.ww.js.gz": 127,
-  "a.wasm": 1850,
-  "a.wasm.gz": 1050,
-  "total": 3248,
-  "total_gz": 2016
+  "a.wasm": 1894,
```
kripken, line 8: Why did this grow by 44 bytes?

sbc100: I think because it fixes the wasm workers implementation of errno_location so that it returns a TLS location, which I think is bug fix.

sbc100: Prior to this change the WW version of errno_location was the same as the non-threaded version (i.e. not thread local or thread safe).

sbc100:
Updated PR description and title to better reflect the purpose of this change.

test/code_size/hello_wasm_worker_wasm.json
kripken: Ah, right... good fix then.

tools/system_libs.py
```diff
@@ -742,7 +742,7 @@ def vary_on(cls):
 
   @classmethod
   def get_default_variation(cls, **kwargs):
-    return super().get_default_variation(is_mt=settings.PTHREADS, is_ww=settings.WASM_WORKERS and not settings.PTHREADS, **kwargs)
+    return super().get_default_variation(is_mt=settings.PTHREADS, is_ww=settings.SHARED_MEMORY and not settings.PTHREADS, **kwargs)
```
kripken, line 745: Is this needed in this PR?

kripken (approved):
lgtm otherwise

tools/system_libs.py
sbc100: @eyebrowsonfire, is this maybe enough to resolve https://github.com/emscripten-core/emscripten/pull/22683

sbc100: @eyebrowsoffire, is this maybe enough to resolve #22683

sbc100: The reason this is needed for this PR is because it makes `__errno_location` into a function that, in non-threads build, has TLS erased, making it incompatible with `-sSHARED_MEMORY`, which triggers the problem found in #22683.   Basically it make the problem found in #22683 happen is way more cases, including simple hello world programs.

PR #22749, opened 2024-10-16
Cleanup wasm worker tests. NFC

sbc100:
- Use `emscripten_out` helper instead of EM_ASM(out(...)).
- Remove tabs
- Curly braces on the same line as keyword
- Parameterize one test

### DIFF
--- a/test/wasm_worker/thread_stack.c
+++ b/test/wasm_worker/thread_stack.c
@@ -1,43 +1,42 @@
-#include <emscripten.h>
+#include <emscripten/console.h>
 #include <emscripten/wasm_worker.h>
 #include <emscripten/stack.h>
 #include <stdio.h>
 #include <assert.h>
 #include <stdlib.h>
 
-#define THREAD_STACK_SIZE 1024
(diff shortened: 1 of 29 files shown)


kripken (approved):

test/wasm_worker/thread_stack.c
```diff
@@ -1,43 +1,42 @@
-#include <emscripten.h>
+#include <emscripten/console.h>
 #include <emscripten/wasm_worker.h>
 #include <emscripten/stack.h>
 #include <stdio.h>
 #include <assert.h>
 #include <stdlib.h>
 
-#define THREAD_STACK_SIZE 1024
+#define THREAD_STACK_SIZE 2048
```
kripken, line 8: Stack size change because of `emscripten_outf`?

sbc100: Yup, essentially printf.  The exact side of the stack in this test is not super important though I think.  I just designed to ensure that stack start/end match our expectations.

PR #22750, opened 2024-10-16
Decouple libwasm_workers from MTLibrary. NFC

sbc100:
The MTLibrary subclass is designed for libraries that build in all different modes.  I think the wasm_workers library and the planned pthreads library should not be inheriting from this class.

This is needed for both #22683 and #22735.


eyebrowsoffire (approved):
LGTM!

eyebrowsoffire (approved):

kripken (approved):

PR #22753, opened 2024-10-16
[AUDIO_WORKLET] Optimise the copy back from wasm's heap to JS

cwoffenden:
Short version: this improves the copy back from the audio worklet's heap to JS by 7-12x depending on the browser. 

Since we pass in the stack for the worklet from the caller's heap, its address doesn't change. And since the _render quantum size_ doesn't change after the audio worklet creation, the stack positions for the audio buffers do not change either. This optimisation adds one-time subarray views and replaces the float-by-float copy with a simple `set()` per channel (per output).

To prove this doesn't break anything, tests of the audio worklet API to compare before and after have already landed in #23394, merged here; they can be run using:

```
test/runner interactive.test_audio_worklet_stereo_io
test/runner interactive.test_audio_worklet_2x_stereo_io
test/runner interactive.test_audio_worklet_mono_io
test/runner interactive.test_audio_worklet_2x_hard_pan_io 
```

A benchmark of the extracted copy before and after is here:

https://wip.numfum.com/cw/2024-10-29/index.html

This PR fulfils the garbage-free requirement as well as having the performance boost. Multiple output buffers are created with addresses at the first entries in the AW's stack, and the number of ins and outs can change dynamically (using or not the predefined buffers). There are tests for buffer overflows as well as sanity.

(Edited to take the content from the conversation below and turn this from a proposal to a summary.)


juj:
Nice idea! It looks like this PR carries the rename from the other PR. Rebase/merge should hide it?

src/audio_worklet.js
```diff
@@ -97,6 +97,20 @@ function createWasmAudioWorkletProcessor(audioParams) {
         // Reserve space for the output data
         dataPtr += bytesPerChannel * i.length;
       }
+      if (!this.outputViews) {
+        this.outputViews = [];
+        k = outputDataPtr;
+        for (/*which output*/ i of outputList) {
+          for (/*which channel*/ j of i) {
+            this.outputViews.push({
+              // dataPtr is the sanity check (to be implemented)
```
sbc100: If its a sanity check then perhaps put it behind `if ASSERTIONS`?

src/audio_worklet.js
```diff
@@ -97,6 +97,20 @@ function createWasmAudioWorkletProcessor(audioParams) {
         // Reserve space for the output data
         dataPtr += bytesPerChannel * i.length;
       }
+      if (!this.outputViews) {
+        this.outputViews = [];
+        k = outputDataPtr;
+        for (/*which output*/ i of outputList) {
```
sbc100, line 110: Can the length of `outputList` be different in each call?

Perhaps `assert(this.outputViews.length == outputList.length)` after this block to be clear?

cwoffenden:
> Nice idea!

A shower thought, as all good ideas are!

> It looks like this PR carries the rename from the other PR. Rebase/merge should hide it?

Yes, this one was rebased off the earlier one so should have the same commit IDs.

src/audio_worklet.js
cwoffenden: It can grow if other nodes are manually added. I’ll need to write contrived examples to see this, since in the real world I don’t think it will.

src/audio_worklet.js
cwoffenden: I’ve some rearranging to do, to make sure the addresses are fixed, and if those change from the base I’ll add assertions.

I’ll make this a draft and work on it next week, adding some robustness.

src/audio_worklet.js
```diff
@@ -93,6 +96,24 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
         // Reserve space for the output data
         dataPtr += bytesPerChannel * i.length;
+        // How many output views are needed in total?
+        requiredViews += i.length;
+      }
+
+      // Verify we have enough views (it doesn't matter if we have too many, any
+      // excess won't be accessed) then also verify the views' start address
+      // hasn't changed.
+      // TODO: allocate space for outputDataPtr before any inputs?
```
juj, line 106: It would be ok to turn this around to allocate outputs before inputs. This way it should be possible to only assert() in debug builds that the invariant `this.outputViews[0].byteOffset == k << 2` holds.

Also, the stack is already preallocated by the time the above `constructor()` is called, so things can be precomputed there in the ctor instead of at process time for any performance wins.

There are two things that I'd recommend:
1) benchmark the effect of this optimization. Since the code does become more sophisticated/more dependent on things having been pre-set up, it would be good to confirm that there is actually a performance benefit to calling TypedArray.set() instead of manually looping the samples.

I think `performance.now()` should work here, so adding calls to performance.now() at very top of process() and at the very end of process() to measure the overall perf differential would probably work.

2) It would be important to test two scenarios: a) there being multiple distinct audio processor classes, and b) there being different node paths in the same graph that have different number of inputs/outputs channels.

With the original code, it was reasonably "trivial" to statically prove that the code will handle any combination of classes or channels, though when state is being cached like this, there will then be the possibility that different schemes could trip up.

I wonder if a good test case might be to have a scenario where there's simultaneously
a) one stereo source clip -> simple passthrough stereo worklet -> stereo audiocontext speaker destination
b) one mono source clip -> simple passthrough mono worklet -> mono output
c) two copies of mono sources from b) above  -> worklet merging (memcpying) these into stereo as Left/Right channels -> stereo audiocontext speaker destination

In such a test, there would be a full combination of "1ch input, 1ch output", "2ch input, 2ch outputs", "two 1ch inputs, 2ch output" cases that I think would cover every possibility (graph would have multiple processor classes and different shaped nodes).

The worklets themselves would be trivial memcpy input -> output implementations so no much complexity to the audio code itself, but they would then test that different shapes of the graph are preserved. Would something like that be a feasible test case to build?

cwoffenden: > It would be ok to turn this around to allocate outputs before inputs. This way it should be possible to only assert() in debug builds that the invariant `this.outputViews[0].byteOffset == k << 2` holds.

I was unexpectedly free this afternoon so already rearranged everything.

 > Also, the stack is already preallocated by the time the above constructor() is called, so things can be precomputed there in the ctor instead of at process time for any performance wins.

I'll look at creating some views in the ctor up front after I've done the timings.

But you're 100% right about timing this, which I'll aim to do once it's in roughly the right shape, and same for the test cases. At work our use is mono mic, stereo out from a wasm mixer, and then combined mic and mixer, but I want to see it tested with more ins and outs like you propose.

(I can dedicate quite a bit of time to this for the next weeks)

src/audio_worklet.js
cwoffenden: > If its a sanity check then perhaps put it behind `if ASSERTIONS`?

Interestingly (unless I'm doing something hideous) but `assert()` isn't available in the `AudioWorkletGlobalScope`, and I can't add it as a dependency the same as `stackAlloc()`, etc. This stuff I'm very rusty with, mind.

src/audio_worklet.js
cwoffenden: > I think `performance.now()` should work here

It's not part of the `AudioWorkletGlobalScope`. `Date.now()` is available but I just get zeros, though I did see a `1` which brought excitement for a brief moment.

cwoffenden: I added the same `Date.now()` to the original code... and it's a lot of zeros, then some occasional ones. I _think_ there are more ones than with the views, but that's not very scientific.

What I'll need to do is write something standalone that runs `process()` not in the `AudioWorkletGlobalScope` and then benchmark it over many runs on different browser and hardware (which is worth doing).

cwoffenden:
@juj or someone who knows more about the internals of this than me: on entering `AudioWorkletProcessor.process()` nothing else will have used the stack beforehand? I've verified stack top is always the same when we enter the `AudioWorkletProcessor`'s contructor and `process()` and it's always the same as the the address (plus length) passed to `emscripten_start_wasm_audio_worklet_thread_async()`.

To rephrase: unless the audio worklet explicitly uses the stack functions from JS, nothing external from Wasm will before `callbackFunction()` is called?

cwoffenden:
Some notes: lots of experiments with the stack allocations, minimum sizes, various flags (`STACK_OVERFLOW_CHECK` and `ASSERTIONS`) and we have some very happy code! View allocations are all in the constructor, with sanity checks performed on the address.

Next is to benchmark it.

cwoffenden:
Benchmarks of the main part of the audio copy done:

https://wip.numfum.com/cw/2024-10-29/index.html

Testing on my M2 Mac Studio in Chrome and Safari this PR is around 15x faster on the float copy, e.g. the original being 0.625µs per `process()` and the new code 0.044µs. On my aging iPhone 11 it's 1.182µs before and 0.199µs after.

Firefox on Mac has the largest difference: 5.090µs before and 0.123µs after, a 41x speed-up. Both fascinating and sad.

@juj if we're in agreement that the simplified standalone JavaScript test code is doing the right thing (it's short and a copy and paste), I'll can gather numbers from regular hardware (we have a wall of Chromebooks at work).

Next I'll need to create tests to show that this still works with various input and output configs.

src/audio_worklet.js
```diff
@@ -52,22 +79,35 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+#endif
+
+      // Allocate the necessary stack space (dataPtr is always in bytes, and
+      // advances as space for structs and data is taken, but note the switching
+      // between bytes and indices into the various heaps, usually in 'k'). This
+      // will be 16-byte aligned (from _emscripten_stack_alloc()), as were the
+      // output views, so we round up and advance the required bytes to ensure
+      // the addresses all work out at the end.
+      i = (stackMemoryNeeded + 15) & ~15;
+      dataPtr = stackAlloc(i) + (i - stackMemoryNeeded);
```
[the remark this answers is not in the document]
cwoffenden: The restore is further down:

https://github.com/emscripten-core/emscripten/blob/f39e26ee210c17cfb8b288fef4e33ab0a6b404ca/src/audio_worklet.js#L188

As for the asserts, all the relevant addresses, offsets and stack calculations are verified here:

https://github.com/emscripten-core/emscripten/blob/f39e26ee210c17cfb8b288fef4e33ab0a6b404ca/src/audio_worklet.js#L154

The stack is passed in externally, either statically allocated or something like `memalign(16, AUDIO_STACK_SIZE)`, e.g. from the test code:

https://github.com/emscripten-core/emscripten/blob/f39e26ee210c17cfb8b288fef4e33ab0a6b404ca/test/webaudio/audio_worklet_tone_generator.c#L127

So we're guaranteed it's at the same address (I'm just following the existing design and making it faster).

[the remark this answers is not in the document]
sbc100: > I'm just following the existing design and making it faster).

But in your new design aren't you assuming the stack pointer value in `constructor` is always identical to the stack pointer value when `process` is called?  I think this assumption is probably fine but I just want to make sure we call it out and verify it (which it sounds like you are doing with asserts).

BTW I wonder if we can just call stackAlloc during `constructor` and then *not* call stack restore?   Then you also wouldn't need to call stackAlloc in the `process` either?  Slightly better code code size and a little less work maybe?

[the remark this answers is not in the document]
cwoffenden: > But in your new design aren't you assuming the stack pointer value in `constructor` is always identical to the stack pointer value when `process` is called?

I'll add an extra assert in the constructor (but yes, it should never change). We could, as you mentioned previously, malloc space for the data views, but this design lets us reuse the same space for input and output buffers. I know it's 5kB currently, but with a render quantum size of 4kB (when the API to change this lands) it becomes 40kB. Tiny in the grand scheme of things

> BTW I wonder if we can just call stackAlloc during `constructor` and then _not_ call stack restore? Then you also wouldn't need to call stackAlloc in the `process` either?

I can take a look later in the week. The bonus to using `stackAlloc` and the stack like this in general is that we benefit from the `STACK_OVERFLOW_CHECK`, plus I'd like to move to using ```{{{ makeValue }}}``` (I think that's what it's called) to fill the structs.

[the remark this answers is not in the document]
sbc100: > I can take a look later in the week. The bonus to using `stackAlloc` and the stack like this in general is that we benefit from the `STACK_OVERFLOW_CHECK`, plus I'd like to move to using `{{{ makeValue }}}` (I think that's what it's called) to fill the structs.

This second suggestion was not to stop using `stackAlloc`, but to just never call `stackRestore`.   i.e. keep the space you need allocated on the stack after `constructor` returns so you don't need to rs-`stackAlloc` in `process`.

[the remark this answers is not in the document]
cwoffenden: Short version: the stack in `process()` changes with the data to process, which can grow and shrink, then the Wasm call uses the stack.

Longer version: the number of inputs or outputs may change, which is probably rare, but certainly anyone using parameters will see that grow and shrink. I reserve space for buffers in `constructor()` based on the size of the stack, but we don't know at construction time how many channels will be used, or even whether output will even be used, and I certainly don't know how much from the constructor alloc how much to leave over for Wasm.

The simple use case of stereo audio playback coming from a mixer running in the worklet will remain constant, and for this I could just run one alloc, agreed.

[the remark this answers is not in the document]
sbc100: Oh, I thought that `stackAlloc` in `process` always matched the `stackAlloc` in `constructor`.  I guess I misunderstood.  Sorry for the noise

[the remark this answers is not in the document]
cwoffenden: > Sorry for the noise

No problem! 😀

cwoffenden:
Note to me: look at #22808

src/audio_worklet.js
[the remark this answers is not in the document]
cwoffenden: > But in your new design aren't you assuming the stack pointer value in `constructor` is always identical to the stack pointer value when `process` is called?

I added a specific assert for this early on:

https://github.com/emscripten-core/emscripten/blob/069f7a4a43f6e5f13546f8af2b857486d7233d18/src/audio_worklet.js#L100

The individual buffer addresses are asserted lower down.

test/webaudio/audio_files/README.md
cwoffenden, line 1: Just for reference, @sbc100, these test files I'll be using really are by Tim Wright, aka CoLD SToRAGE, aka the composer behind Wipeout and Lemmings. He's a work colleague and old friend, so I asked him to create some test loops for the demo mixer code I'm adding.

sbc100: WOW!   I played a lot of both of those in the 90s!

cwoffenden: I couldn't just bring myself to add any old MP3s for testing when I have Tim sitting just round the corner from my office!

cwoffenden:
I'm going to add variants now of the multiple ins and outs. Demo here, 9kB of wasm, not closured to aid stepping through, of two inputs downmixed to one:

https://wip.numfum.com/cw/2024-11-15/index.html

cwoffenden:
@juj  I added various tests: two stereo inputs being mixed in the processor, two stereo inputs being `memcpy`'d, two mono inputs being `memcpy`'d, and two mono inputs being hard-panned L+R. They're these:
```
test/runner interactive.test_audio_worklet_stereo_io
test/runner interactive.test_audio_worklet_2x_stereo_io
test/runner interactive.test_audio_worklet_mono_io
test/runner interactive.test_audio_worklet_2x_hard_pan_io 
```
These test from 1-4 output channels, with stack sizes of 2-6kB, testing everything moving around in the worklet's stack.

The performance improvement is 7-12x the original (details higher up) and no versatility is lost.

(This is ready for review)

test/test_browser.py
```diff
@@ -5382,7 +5382,7 @@ def test_full_js_library_strict(self):
     'minimal_runtime_pthreads_and_closure': (['-sMINIMAL_RUNTIME', '-pthread', '--closure', '1', '-Oz'],),
     'pthreads_es6': (['-pthread', '-sPTHREAD_POOL_SIZE=2', '-sEXPORT_ES6'],),
     'es6': (['-sEXPORT_ES6'],),
-    'strict': (['-sSTRICT'],),
+    'strict': (['-sSTRICT', '-sINCOMING_MODULE_JS_API=canvas,instantiateWasm,monitorRunDependencies,onAbort,onExit,print,setStatus,wasm,wasmMemory'],),
```
sbc100, line 5385: Why is this change needed for this PR?

test/test_browser.py
```diff
@@ -5408,6 +5408,14 @@ def test_audio_worklet_post_function(self, args):
   def test_audio_worklet_modularize(self, args):
     self.btest_exit('webaudio/audioworklet.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-sMODULARIZE=1', '-sEXPORT_NAME=MyModule', '--shell-file', test_file('shell_that_launches_modularize.html')] + args)
 
+  # We can't test playback, so don't need the data files, we're testing it builds and runs
+  @parameterized({
+    '': ([],),
+    'minimal_with_closure': (['-sMINIMAL_RUNTIME', '--closure', '1', '-Oz'],),
```
sbc100, line 5414: We normally write this `--closure=1` as a single arg

test/webaudio/audioworklet_2x_in_hard_pan.c
```diff
@@ -0,0 +1,154 @@
...
+		audio.loop = looping;
+		var track = context.createMediaElementSource(audio);
+		return emscriptenRegisterAudioObject(track);
+	}
+	return 0;
+});
+
+// Toggles the play/pause of a MediaElementAudioSourceNode given its ID
+EM_JS(void, toggleTrack, (EMSCRIPTEN_WEBAUDIO_T srcID), {
+	var source = emscriptenGetAudioObject(srcID);
```
sbc100, line 52: Can you use 2-space indentation for all these new tests?

test/test_interactive.py
```diff
@@ -306,6 +306,34 @@ def test_audio_worklet_tone_generator(self):
...
+    shutil.copy(test_file('webaudio/audio_files/emscripten-beat-mono.mp3'), 'audio_files/')
+    shutil.copy(test_file('webaudio/audio_files/emscripten-bass-mono.mp3'), 'audio_files/')
+    self.btest('webaudio/audioworklet_in_out_mono.c', expected='0', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS'])
+
+  # Tests an AudioWorklet with multiple mono inputs copying in the processor to L+R stereo outputs (3kB stack)
+  def test_audio_worklet_2x_hard_pan_io(self):
+    os.mkdir('audio_files')
+    shutil.copy(test_file('webaudio/audio_files/emscripten-beat-mono.mp3'), 'audio_files/')
+    shutil.copy(test_file('webaudio/audio_files/emscripten-bass-mono.mp3'), 'audio_files/')
+    self.btest('webaudio/audioworklet_2x_in_hard_pan.c', expected='0', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS'])
```
sbc100, line 335: Can you use `btest_exit` and remove the `expected` arg?

src/audio_worklet.js
```diff
@@ -109,17 +141,52 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+        // First that the output view addresses match the stack positions.
+        k = dataPtr - bytesPerChannel;
+        for (i = 0; i < outputViewsNeeded; i++) {
+          console.assert(k == this.outputViews[i].byteOffset, 'AudioWorklet internal error in addresses of the output array views');
+          k -= bytesPerChannel;
+        }
+        // And that the views' size match the passed in output buffers
+        for (i of outputList) {
+          for (j of i) {
+            console.assert(j.byteLength == bytesPerChannel, `AudioWorklet unexpected output buffer size (expected ${bytesPerChannel} got ${j.byteLength})`);
```
sbc100, line 175: Is it possible to use the emscripten `assert` function here?

cwoffenden: `assert()` isn't available to the worklet. I could look into why, I think it might just need exposing on the module then a adding to the `AudioWorkletGlobalScope`. But I think I already looked at this last month.

test/test_browser.py
cwoffenden: I was running the audio tests with `-sSTRICT` and couldn't understand how any of them ran without it. They build, they don't run, e.g., compile any example with `STRICT` and run it:

```
Uncaught RuntimeError: Aborted(`Module.wasmMemory` was supplied but `wasmMemory` not included in INCOMING_MODULE_JS_API)
    at abort (index.js:592:11)
    at ignoredModuleProp (index.js:823:5)
    at checkIncomingModuleAPI (index.js:1548:3)
    at index.js:223:1
```

It's for sure separate, so can remove it and come back to the tests later.

test/test_browser.py
cwoffenden: I just copied an existing test for this, I'll change it tomorrow.

test/test_browser.py
sbc100: If wasmMemory is requires by for AUDIO_WORKLET then its should be added explicitly in link.py, so that its always part of `INCOMING_MODULE_JS_API` (even in strict mode).

test/test_interactive.py
cwoffenden: Question here: I wrote the tests so they do use the exit by verifying playback happened. You'd like me to remove the verifications in the tests too?

test/test_browser.py
cwoffenden: I created #22971 and I'll come back to this, since it's not new.

sbc100: Great.  I'm happy to take a look at that one too if you like.  Or I can help you fix.  Should be fairly simple.

test/test_interactive.py
sbc100: No, I'm just suggesting that if you are using the exit code as the result and the result is zero then you can use `btest_exit` which by default checks for zero exit code.

Having said that, I think calling exit() from a audio worklet currently doesn't' work because proxying doesn't work, so maybe its not easy in this case?

I've been trying to move from `btest` to `btest_exit` whereever we can.

src/audio_worklet.js
sbc100: Fair enough.

sbc100: In the long run I'd like to remove the separate worklet.js file, just like I did already for pthreads.    Having everything in the single file makes things much easier.  For now using `console.assert` here seems fine.

test/test_browser.py
cwoffenden: I took a quick look and I'm a missing some of the concepts. Feel free to take a look or just throw some hints my way. I won't be able to look at it until next week now.

I stalled after a quick look because I wasn't sure things like `canvas` should be automatically added, so wanted to test first. But the locking was higher on the todo list (and I was hoping to get #22949 solved and in, which I think I'll just write because exposing sample rate, etc., in an extensible way is pretty important).

Plus we'd like to ship our code built around this at some point! 😀

test/test_interactive.py
cwoffenden: Ah, makes sense. I'll can take a look next week (I'm done for a few days now).

src/audio_worklet.js
cwoffenden: I'll leave it for now. Having seen what you'd done for pthreads, it's on my list to look at.

test/test_browser.py
cwoffenden: Resolving this and will come back to ##22971.

test/test_interactive.py
cwoffenden: I changed to `btest_exit` and `emscripten_force_exit()`. The test instantly exits.

So I build with `-sNO_EXIT_RUNTIME` and now it doesn't exit at all. I can see the reported result if I switch back to `btest`, so `report_result?exit:0`, but it continues running.

Is this the behaviour you'd expect when you say the proxying doesn't work?

sbc100: You need to add `exit_with_live_runtime()` (or better, `emscripten_runtime_keeplive_push()`) at the end of your `main` function, otherwise returning from `main` signals the end of the program.

sbc100: See https://github.com/emscripten-core/emscripten/pull/22776

cwoffenden: The tests are all migrated to `btest_exit()`

cwoffenden:
@sbc100 any chance of getting this merged in the next week or so? I want to build my API proposal on top of this.

I'll happily produce more numbers showing what a difference it makes on low-spec'd hardware (it's millis saved per second on high-end).

sbc100:
> @sbc100 any chance of getting this merged in the next week or so? I want to build my API proposal on top of this.
> 
> I'll happily produce more numbers showing what a difference it makes on low-spec'd hardware (it's millis saved per second on high-end).

Since @juj wrote this code and knows it best I think we should wait for his lgtm.

test/test_browser.py
```diff
@@ -5417,6 +5417,17 @@ def test_audio_worklet_post_function(self, args):
   def test_audio_worklet_modularize(self, args):
     self.btest_exit('webaudio/audioworklet.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-sMODULARIZE=1', '-sEXPORT_NAME=MyModule', '--shell-file', test_file('shell_that_launches_modularize.html')] + args)
 
+  # We can't test playback, so don't need the data files, we're testing it builds and runs
```
sbc100, line 5420: The comment says we don't need data files, but the code looks like to does copy them?

cwoffenden: Once I changed the test to btest_exit they were needed (because the test actually ran). I’ll make the change.

src/audio_worklet.js
```diff
@@ -81,24 +127,10 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
-        k += {{{ C_STRUCTS.AudioSampleFrame.__size__ / 4 }}};
-        // Reserve space for the output data
-        dataPtr += bytesPerChannel * i.length;
-      }
-
       // Copy parameters descriptor structs and data to Wasm
       paramsPtr = dataPtr;
-      k = paramsPtr >> 2;
       dataPtr += numParams * {{{ C_STRUCTS.AudioParamFrame.__size__ }}};
+      k = paramsPtr >> 2;
```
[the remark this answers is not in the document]
cwoffenden: I recall doing it for a reason, I think it was to match the pattern of the rest of the code. I can move it back.

cwoffenden:
I ran the loop test I wrote on some recent Chromebooks, and on a middle-of-the-road ARM model I'm seeing an 11x speed-up (e.g. 1.818µs before, and 0.160µs after). Same 7-12x range improvement on other devices.

cwoffenden:
> > @sbc100 any chance of getting this merged in the next week or so?
> 
> Since @juj wrote this code and knows it best I think we should wait for his lgtm.

I'd really appreciate some eyes on this and getting it merged. I'm going to start the proposed API changes and further prep for the next version of the AW standard, and need to build it on these changes (and after that I want to do a single file PR like the workers).

tklajnscek:
@juj can we get this merged please? We have other stuff waiting to be built on top of these changes and it's tedious to work with rebases on rebases :)

mttjcksn:
Awesome, the linked test shows a 9x speedup in chrome my M2 Macbook. 
We would welcome this change as soon as it's ready!

cwoffenden:
@sbc100 I'd still like to get this in. I can split out the tests from the code changes, since they're standalone and were the most amount of work.

sbc100:
> @sbc100 I'd still like to get this in. I can split out the tests from the code changes, since they're standalone and were the most amount of work.

That would be awesome, thank you.   I'll try to review the new tests and we can ping @juj again to take a look at the optimizations here.

cwoffenden:
> > I can split out the tests from the code changes
> 
> That would be awesome, thank you.

Tests moved to #23394 (since they don't require these changes).

cwoffenden:
Since this has had a lot of comments and changes since its initial commit in October I've edited the initial comment to contain the findings and results, and merged the tests which landed in main. Besides the comment/error message in `library_webaudio.js` all of the changes are now in a single file.

cwoffenden:
Odd, GH closed this when I renamed the branch to merge into a fork (it's residing in [cw-audio-optimised-copy](https://github.com/cwoffenden/emscripten/tree/cw-audio-optimised-copy)).

If there's any interest in this I can re-open another PR.

PR #22755, opened 2024-10-16
Use emscripten_stack_set_limits in wasm_worker_initialize.S

sbc100:
Followup to #22683.  This change avoids making __stack_base and __stack_end into public/external symbols and avoids reimplementing `emscripten_stack_set_limits`.

This is a minor code size win too.

### DIFF
--- a/system/lib/wasm_worker/wasm_worker_initialize.S
+++ b/system/lib/wasm_worker/wasm_worker_initialize.S
@@ -1,6 +1,4 @@
 .extern __stack_pointer
-.extern __stack_base
-.extern __stack_end
 
 #ifdef __wasm64__
 #define PTR i64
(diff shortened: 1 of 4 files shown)


system/lib/wasm_worker/wasm_worker_initialize.S
```diff
@@ -13,32 +11,37 @@
 #endif
 
 .globaltype __stack_pointer, PTR
-.globaltype __stack_end, PTR
-.globaltype __stack_base, PTR
+.globaltype __tls_size, PTR, immutable
+.functype emscripten_stack_set_limits (PTR /*__stack_base*/, PTR /*__stack_base*/) -> ()
```
eyebrowsoffire, line 15: Is the second comment supposed to be `/*__stack_end*/`?

eyebrowsoffire (approved):
LGTM, with one tiny nitpick. With those globals being refactored out, I'm wondering what is left before this could just be a simple C function instead of being written in assembly. I see that there is still a global reference to a `__stack_pointer` that is needed here. I wonder if it would be possible to expose a function that is just `emscripten_reset_stack_pointer` or something like that, which would just set `__stack_pointer` to `__stack_base`? (Also probably not for this PR, I'm just thinking out loud a bit).

sbc100:
> LGTM, with one tiny nitpick. With those globals being refactored out, I'm wondering what is left before this could just be a simple C function instead of being written in assembly. I see that there is still a global reference to a `__stack_pointer` that is needed here. I wonder if it would be possible to expose a function that is just `emscripten_reset_stack_pointer` or something like that, which would just set `__stack_pointer` to `__stack_base`? (Also probably not for this PR, I'm just thinking out loud a bit).

The problem is that you cannot safely call *any* C function until the stack pointer has been set.   I mean in theory you could try to second guess the C compiler and try to write your C function in way you are think it won't itself use the stack pointer, but its very fragile.   Better to write these very low level startup routines asm I think

test/code_size/hello_wasm_worker_wasm.json
```diff
@@ -5,8 +5,8 @@
...
   "a.ww.js": 115,
   "a.ww.js.gz": 127,
-  "a.wasm": 1894,
-  "a.wasm.gz": 1077,
-  "total": 3292,
-  "total_gz": 2043
+  "a.wasm": 1879,
+  "a.wasm.gz": 1066,
+  "total": 3277,
+  "total_gz": 2032
```
juj, line 11: This is great - I'm puzzled to understand how this saved code size though..?

juj (approved):

test/code_size/hello_wasm_worker_wasm.json
sbc100: I'm also a little surprised.   

My initial guess was that perhaps `emscripten_stack_set_limits` was being included prior to this change and so this change reduces overall duplication... but I don't think thats true.  I could investigate further but but this seems like a win so I think we just take it.

system/lib/wasm_worker/wasm_worker_initialize.S
sbc100: Fixed

test/code_size/hello_wasm_worker_wasm.json
sbc100: I think i double checked that.   I normally rebaseline all tests before landing changes like this, just to be sure I'm seeing the real deltas.

PR #22756, opened 2024-10-16
Fix STACK_OVERFLOW_CHECK + WASM_WORKERS

sbc100:
We we not calling `writeStackCookie()` in the right place which means that `checkStackCookie()` would always fail in wasm workers.

Required for #22721

### DIFF
--- a/src/postamble.js
+++ b/src/postamble.js
@@ -162,13 +162,6 @@ function run() {
     return;
   }
 
-#if STACK_OVERFLOW_CHECK
-#if PTHREADS
-  if (!ENVIRONMENT_IS_PTHREAD)
-#endif
-    stackCheckInit();
-#endif
-
 #if WASM_WORKERS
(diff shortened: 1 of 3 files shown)


sbc100:
Split out from #22721

src/postamble.js
```diff
@@ -192,6 +185,10 @@ function run() {
   }
 #endif
 
+#if STACK_OVERFLOW_CHECK
+  stackCheckInit();
```
kripken, line 189: This code both moved and lost the condition `if (!pthread)` - do we not need that?

sbc100: We already have early return for `if (ENVIRONMENT_IS_PTHREAD)` and `if (ENVIRONMENT_IS_WASM_WORKER)` so by the time we get here know were the main runtime thread.

juj (approved):

Issue #14804, opened 2021-08-05
console.log message are lost when running with node + pthreads


sbc100:
I noticed this when running in PROXY_TO_PTHREAD mode.   If I use printf(), I see all my message... if I use `console.log` (or `out`) I only see the very first of the log messages.

juj:
Hmm, console.log()s should always come through, unless the main thread is blocked/hung(?). Not quite sure about `out()`.

We used to have a manual postMessage-based console.log override to avoid an old SAB-based deadlock "bug" (a non-specced behavior of forward progress interaction between browser APIs and SAB locking that used to be non-observable before SAB) in Firefox and Edge, but I don't think that has existed in a long time.

juj:
Maybe the issue is due to this? https://github.com/emscripten-core/emscripten/blob/d35de0bf3572dfe9759ac3c7e56b9ce5983f68ac/src/worker.js#L77

sbc100:
> Hmm, console.log()s should always come through, unless the main thread is blocked/hung(?). Not quite sure about `out()`.
> 
> We used to have a manual postMessage-based console.log override to avoid an old SAB-based deadlock "bug" (a non-specced behavior of forward progress interaction between browser APIs and SAB locking that used to be non-observable before SAB) in Firefox and Edge, but I don't think that has existed in a long time.

This issue I think might be something specific to the way node does workers.

sbc100:
> Maybe the issue is due to this?
> 
> https://github.com/emscripten-core/emscripten/blob/d35de0bf3572dfe9759ac3c7e56b9ce5983f68ac/src/worker.js#L77

I saw that too, but no, this relates to code in the module itself, not in worker.js.   It also applies equally to `out` and `console.log`.. which makes me think it specific to how node worker use the console.   

I will confirm whether this issue is specific to node.   I think it is.

kripken:
Reopening as this is not actually fixed, just worked around.

Looking into this, if I build this:
```cpp
#include <emscripten.h>
int main() {
  EM_ASM({
    console.log("test");
    console.log("test");
    console.log("test1");
    console.log("test2");
    console.log("test3");
  });
}
```
with `./emcc a.cpp -pthread -s PROXY_TO_PTHREAD  -O3 --profiling -s EXIT_RUNTIME` then I can see the problem: it prints out "test" once and that's it. However:
* If I remove EXIT_RUNTIME then it logs out everything (and then hangs since we don't exit).
[...]

PR #22761, opened 2024-10-17
Use emscripten_out/err in wasm worker tests. NFC

sbc100:
Also replace some EM_ASM that only existed to call `out` or `err`.

For multi-threads tests that run under node its better to use `out` / `err` than `console.log` / `console.error` since the former get redirected to stdout/stderr directly whereas it seem the later requires some sort of messaging with the main thread.

### DIFF
--- a/test/wasm_worker/c11__Thread_local.c
+++ b/test/wasm_worker/c11__Thread_local.c
@@ -1,4 +1,4 @@
-#include <emscripten.h>
+#include <emscripten/console.h>
 #include <emscripten/wasm_worker.h>
 #include <assert.h>
 #include <threads.h>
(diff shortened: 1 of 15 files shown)


kripken:
How does `console.log` require messaging to the main thread? I thought it is independent in each worker.

sbc100:
> How does `console.log` require messaging to the main thread? I thought it is independent in each worker.

There is some kind of synchronization such that if you block the main thread console.log message from workers don't get through. 
  This is why we redirect to stdout/stderr directly.  See https://github.com/emscripten-core/emscripten/blob/ac83519fffe7c628a579a362bc38fd082bd716f3/src/shell.js#L423-L435

sbc100:
See https://github.com/emscripten-core/emscripten/issues/14804

kripken (approved):

Issue #22742, opened 2024-10-15
Undefined symbol `_emscripten_memcpy_js` when linking rust code with -lc -Oz


hoodmane:
You need rust and the emscripten target installed. `rustup target add wasm32-unknown-emscripten` installs the emscripten target if you use rustup.
```sh
echo "fn main(){}" > a.rs
rustc -Copt-level=z a.rs -o a.js --target=wasm32-unknown-emscripten
```
Fails with:
```
          error: undefined symbol: _emscripten_memcpy_js (referenced by root reference (e.g. compiled C/C++ code))
          warning: To disable errors for undefined symbols use `-sERROR_ON_UNDEFINED_SYMBOLS=0`
          warning: __emscripten_memcpy_js may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
          Error: Aborting compilation due to previous errors
          emcc: error: '/home/rchatham/Documents/programming/emsdk/node/18.20.3_64bit/bin/node /home/rchatham/Documents/programming/emsdk/upstream/emscripten/src/compiler.mjs /tmp/tmpogw1glcz.json' failed (returned 1)          
```

If we minimize the failing emcc link command, it looks like:
```sh
emcc -o a.js -Oz -lc \
a.a.<some-hash>-cgu.0.rcgu.o \
a.<some-hash>.rcgu.o \
.../lib/rustlib/wasm32-unknown-emscripten/lib/libstd-1d3ccab4f7990fa9.rlib \
.../lib/rustlib/wasm32-unknown-emscripten/lib/libpanic_abort-bb30bbb88530dc79.rlib \
.../lib/rustlib/wasm32-unknown-emscripten/lib/liballoc-6b694a8bc75de25c.rlib \
.../lib/rustlib/wasm32-unknown-emscripten/lib/libcore-99daaba8bfbbfe7e.rlib \
```

If I remove either `-Oz` or `-lc` then it succeeds. If I pass `EMCC_DEBUG=1` and look at the contents of `tmpogw1glcz.json` it looks like:

<details>
<summary>Details</summary>

{
  "ABORTING_MALLOC": true,
  "ABORT_ON_WASM_EXCEPTIONS": false,
  "AGGRESSIVE_VARIABLE_ELIMINATION": 0,
  "ALIASING_FUNCTION_POINTERS": 0,
  "ALLOW_BLOCKING_ON_MAIN_THREAD": true,
  "ALLOW_MEMORY_GROWTH": false,
[...]

PR #22765, opened 2024-10-18
Drop explicitly passed -lc unless -nodefaultlibs or similar is passed

### LINKED ISSUE #16680: bug: Emscripten doesn't generate PIC libc
Tested 
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.9-git (7e4a0ae4bdd99abbbf299c025aa2ce7b339c8521)
clang version 15.0.0 (https://github.com/llvm/llvm-project b2a7f1c3904ede781565c6ed772f155167aa8325)
Thread model: posix
InstalledDir: /home/hood/Documents/programming/emsdk/upstream/bin
```

Example:
test.c:
```
int f(x){
	return x+4;
}
```

```
emcc -c test.c -o test.o -fPIC
emcc test.o -o test.so -sSIDE_MODULE=1 -lc
```
fails with:
```
wasm-ld: error: unable to find library -lc
emcc: error: '/src/emsdk/emsdk/upstream/bin/wasm-ld -o tst.so --whole-archive tst.o -lc -L/src/emsdk/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/pic --no-whole-archive -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr --import-undefined --import-memory --strip-debug --export-all --no-gc-sections --experimental-pic -shared' failed (returned 1)
```

### LINKED ISSUE #22758: Undefined symbol _emscripten_memcpy_js linking no object files with `-lc`
Edit: To reproduce doesn't require any rust at all:
```console
$ emcc -lc -sDISABLE_EXCEPTION_CATCHING=0 -sWASM_BIGINT
error: undefined symbol: _emscripten_memcpy_js (referenced by root reference (e.g. compiled C/C++ code))
warning: __emscripten_memcpy_js may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
emcc: error: '/home/rchatham/Documents/programming/emsdk/node/18.20.3_64bit/bin/node /home/rchatham/Documents/programming/emsdk/upstream/emscripten/src/compiler.mjs /tmp/tmpc06i0z65.json' failed (returned 1)
```

Similar bug to #22742. If you remove either `-lc` or  `-sWASM_BIGINT` from the link command the error goes away.

```
$ emcc -v
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.69 (c55ff404dbe1eb9f263c48b584d5ba627224763a)
clang version 20.0.0git (https:/github.com/llvm/llvm-project 50866e84d1da8462aeb96607bf6d9e5bbd5869c5)
Thread model: posix
```

```
echo "fn main(){}" > a.rs
rustc -Clink-arg=-sWASM_BIGINT a.rs -o a.js --target=wasm32-unknown-emscripten
```

Link command:
```sh
emcc \
a.a.5e989e15ed0bb312-cgu.0.rcgu.o a.0l3w7uzzjgtykfn4cqm4q6yup.rcgu.o \
/home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libstd-1d3ccab4f7990fa9.rlib \
/home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libpanic_unwind-7c13b9baf1486a91.rlib \
/home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/liballoc-6b694a8bc75de25c.rlib \
/home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libcore-99daaba8bfbbfe7e.rlib \
-o a.js \
-sWASM_BIGINT -lc -v --reproducer=repro.tar
```
Output:
```
 /home/rchatham/Documents/programming/emsdk/upstream/bin/clang --version
 /home/rchatham/Documents/programming/emsdk/upstream/bin/wasm-ld -o a.wasm a.a.5e989e15ed0bb312-cgu.0.rcgu.o a.0l3w7uzzjgtykfn4cqm4q6yup.rcgu.o -L/home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libstd-1d3ccab4f7990fa9.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libpanic_unwind-7c13b9baf1486a91.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/librustc_demangle-fb333a2ac0a82ceb.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libstd_detect-dacc3810688bd02e.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libhashbrown-c5a0e9d03eccf5a2.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/librustc_std_workspace_alloc-81ea19ddb8698fb5.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libminiz_oxide-940a4059390b281a.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libadler-3449d83794adc734.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libunwind-41f0d2fd6457e3c4.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libcfg_if-dbc33fcbc9032cd3.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/liblibc-aaba565dd73ad2b0.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/liballoc-6b694a8bc75de25c.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/librustc_std_workspace_core-499c8f9444f2d48c.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libcore-99daaba8bfbbfe7e.rlib /home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/libcompiler_builtins-592839b635ef17b6.rlib -lc-debug -L/home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib -L/home/rchatham/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-emscripten/lib/self-contained -L/home/rchatham/Documents/programming/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten -lGL-getprocaddr -lal -lhtml5 -lbulkmemory -lstubs-debug -lnoexit -lc-debug -ldlmalloc -lcompiler_rt -lc++ -lc++abi-debug -lsockets -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-cxx-exceptions -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr /tmp/tmp924vfcwvlibemscripten_js_symbols.so --strip-debug --export=emscripten_stack_get_end --export=emscripten_stack_get_free --export=emscripten_stack_get_base --export=emscripten_stack_get_current --export=emscripten_stack_init --export=_emscripten_stack_alloc --export=__cxa_can_catch --export=__cxa_increment_exception_refcount --export=__cxa_decrement_exception_refcount --export=setThrew --export=__cxa_free_exception --export=__wasm_call_ctors --export=_emscripten_stack_restore --export=__get_exception_message --export=free --export-if-defined=__start_em_asm --export-if-defined=__stop_em_asm --export-if-defined=__start_em_lib_deps --export-if-defined=__stop_em_lib_deps --export-if-defined=__start_em_js --export-if-defined=__stop_em_js --export-if-defined=main --export-if-defined=__main_argc_argv --export-if-defined=fflush --export-table -z stack-size=65536 --no-growable-memory --initial-heap=16777216 --no-entry --stack-first --table-base=1
 /home/rchatham/Documents/programming/emsdk/upstream/bin/llvm-objcopy a.wasm a.wasm --remove-section=.debug* --remove-section=producers
 /home/rchatham/Documents/programming/emsdk/node/18.20.3_64bit/bin/node /home/rchatham/Documents/programming/emsdk/upstream/emscripten/src/compiler.mjs /tmp/tmprtajj_6r.json
error: undefined symbol: _emscripten_memcpy_js (referenced by root reference (e.g. compiled C/C++ code))
warning: __emscripten_memcpy_js may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
emcc: error: '/home/rchatham/Documents/programming/emsdk/node/18.20.3_64bit/bin/node /home/rchatham/Documents/programming/emsdk/upstream/emscripten/src/compiler.mjs /tmp/tmprtajj_6r.json' failed (returned 1)
```

hoodmane:
There have been a lot of bugs when the caller passes `-lc` over the years. For example it crashes if we do:
```
emcc -lc -sDISABLE_EXCEPTION_CATCHING=0 -sMIN_SAFARI_VERSION=150000
```

Rust likes to pass `-lc`. Let's drop it and stop causing trouble for Rust.

Resolves #22758, #22742 and would have resolved #16680 if it hadn't disappeared first.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -15169,3 +15169,6 @@ addToLibrary({
 
   def test_embool(self):
     self.do_other_test('test_embool.c')
+
+  def test_user_passed_lc(self):
+    self.run_process([EMCC, '-lc', '-sDISABLE_EXCEPTION_CATCHING=0', '-sMIN_SAFARI_VERSION=150000'])
--- a/tools/link.py
+++ b/tools/link.py
@@ -2749,7 +2749,7 @@ def map_to_js_libs(library_name):
   return None
 
 
-def process_libraries(state, linker_inputs):
+def process_libraries(state, linker_inputs, nodefaultlibc):
   new_flags = []
   libraries = []
   suffixes = STATICLIB_ENDINGS + DYNAMICLIB_ENDINGS
@@ -2761,6 +2761,12 @@ def process_libraries(state, linker_inputs):
       new_flags.append((i, flag))
       continue
     lib = removeprefix(flag, '-l')
+    if lib == 'c' and not nodefaultlibc:
+      # Drop explicitly passed -lc unless they also passed -nodefaultlibs or
+      # similar. It's important to link libc after our other injected system
+      # libraries like libbulkmemory, but user linked libraries go ahead of
+      # system libraries, so if the user passes `-lc` then we can get crashes.
+      continue
 
     logger.debug('looking for library "%s"', lib)
 
@@ -3026,12 +3032,12 @@ def package_files(options, target):
 
 
 @ToolchainProfiler.profile_block('calculate linker inputs')
-def phase_calculate_linker_inputs(options, state, linker_inputs):
+def phase_calculate_linker_inputs(options, state, linker_inputs, nodefaultlibc):
   using_lld = not (options.oformat == OFormat.OBJECT and settings.LTO)
   state.link_flags = filter_link_flags(state.link_flags, using_lld)
 
   # Decide what we will link
-  process_libraries(state, linker_inputs)
+  process_libraries(state, linker_inputs, nodefaultlibc)
 
   # Interleave the linker inputs with the linker flags while maintainging their
   # relative order on the command line (both of these list are pairs, with the
@@ -3072,8 +3078,10 @@ def run(linker_inputs, options, state, newargs):
 
   target, wasm_target = phase_linker_setup(options, state, newargs)
 
+  nodefaultlibc = '-nodefaultlibs' in newargs or '-nolibc' in newargs or '-nostdlib' in newargs
+
   # Link object files using wasm-ld or llvm-link (for bitcode linking)
-  linker_arguments = phase_calculate_linker_inputs(options, state, linker_inputs)
+  linker_arguments = phase_calculate_linker_inputs(options, state, linker_inputs, nodefaultlibc)
 
   # Embed and preload files
   if len(options.preload_files) or len(options.embed_files):


hoodmane:
@sbc100 does this need a rebaseline code size or something like that?

tools/link.py
```diff
@@ -2761,6 +2761,12 @@ def process_libraries(state, linker_inputs):
       new_flags.append((i, flag))
       continue
     lib = removeprefix(flag, '-l')
+    if lib == 'c' and not nodefaultlibc:
+      # Drop explicitly passed -lc unless they also passed -nodefaultlibs or
+      # similar. It's important to link libc after our other injected system
+      # libraries like libbulkmemory, but user linked libraries go ahead of
+      # system libraries, so if the user passes `-lc` then we can get crashes.
+      continue
```
sbc100, line 2769: While I agree this fixed the program at hand, it also seems a little to magical.   What if user explictly wants to do something like this:

`-lmylib_prelibc -lc -lmylib_postlibc`

In this case removing the explicit `-lc` provided by the user seems wrong/non-intuitive.

Perhaps it would be better to fix rust not to pass the explicit `-lc`, do you know why rust insists on doing this?

sbc100: Perhaps we should at least warn the user when we do this?

test/test_other.py
```diff
@@ -15169,3 +15169,6 @@ def test_fp16(self, opts):
 
   def test_embool(self):
     self.do_other_test('test_embool.c')
+
+  def test_user_passed_lc(self):
+    self.run_process([EMCC, '-lc', '-sDISABLE_EXCEPTION_CATCHING=0', '-sMIN_SAFARI_VERSION=150000'])
```
sbc100, line 15174: Aren't you missing a source file here?  What does this command produce.. I would expect to fail with "no input files".

hoodmane: No, it passes:
`test_user_passed_lc (test_other.other) ... ok (1.02s)`
The program it outputs doesn't do anything though.

sbc100: Ah.. thats because we don't require a main function by default. 

We should probably include `test_file('hello_world.c')`.   In fact you might as well check that it runs too by using `do_runf` instead of `run_process`.

tools/link.py
hoodmane: > Perhaps we should at least warn the user when we do this?

I suppose so. But passing `-lc` at all is also very likely to be wrong if it's likely to be depending on other libraries that are automatically injected at the very end to work correctly. I guess the warning could say to pass `-nolibc` if they need to link libc in a particular order though `-nolibc` seems to also disable linking of `libmalloc` and `libnoexit` if they should be added.

What about the option of injecting the system libs ahead of the user object files? Do you think that would have any negative consequences?

hoodmane: > Perhaps it would be better to fix rust not to pass the explicit `-lc`, do you know why rust insists on doing this?

Well it seems to work for them with all of the other linker they use. I'm trying to fix it from rust's side too, but it seems to me like it's the linker's responsibility to correctly handle this and not the consumer's responsibility not to pass it.

test/test_other.py
hoodmane: The empty file does run, it exits with status 0. But sure, makes sense to add an actual program.

tools/link.py
sbc100: Is there any downside to remove it that you know of?   Why is it even there?

hoodmane: No, the workaround on the rust side is pretty short nad has no downsides I know of. It was rejected two years ago when I tried to merge the fix to rust, but I opened a new PR with the same fix. I think their response was roughly "if it breaks to pass `-lc` maybe it breaks when you pass all sorts of things, we need to figure out what the actual cause is and implement a real fix".

sbc100: > I suppose so. But passing `-lc` at all is also very likely to be wrong if it's likely to be depending on other libraries that are automatically injected at the very end to work correctly. I guess the warning could say to pass `-nolibc` if they need to link libc in a particular order though `-nolibc` seems to also disable linking of `libmalloc` and `libnoexit` if they should be added.

Yes, anyone of passes `-lc` in emscripten is likely going to have a hard time because they would need to *also* add `-lmalloc` and `-lnoexit` and maybe `-lbulkmemory`.   i.e. in emscripten libc is split up into multiple libraryies which are order dependent.

So I think that removing `-lc` is likely the right thing to do but I also worry that doing it silently could break some advanced users who are passing all of those libraries.  i.e. it will fix rust but perhaps break some other use case.

Another solution would be to merge all out libc libraries into a single one but then we would end up with even more variants, e.g. `libc-noexit-bulkmemory-debug-mt.a`.   This complexity would be largely hidden, but it would mean would have to build probably over 100 different libc versions :-/    Having separate libraries allows us to avoid that.

> 
> What about the option of injecting the system libs ahead of the user object files? Do you think that would have any negative consequences?

hoodmane: Reading the Rust codebase, I am not at all sure where it gets decided that it should pass `-lc` in the first place, but it's easy to filter it out as it's constructing the linker invocation.

sbc100: > "if it breaks to pass `-lc` maybe it breaks when you pass all sorts of things, we need to figure out what the actual cause is and implement a real fix".

I guess now that we know more about the problem we can make a better argument for removing it.   We can say "supplying -lc on its own just doesn't work with emscripten and will break the build".   Also, since the compiler always includes libc by default this flag is redundant and pointless.  Removing it is simply better for everyone.

hoodmane: Right, I'd just really like either Emscripten or Rust to merge a fix for this, I'll see what comments I get on the Rust PR.

PR #22769, opened 2024-10-19
[Docs] Mark `-sWASM_EXNREF` as a link-only setting

kleisauke:
```
emcc: error: linker setting ignored during compilation: 'WASM_EXNREF' [-Wunused-command-line-argument] [-Werror]
```


sbc100 (approved):

kripken:
This will become compile-time as well once we use LLVM's codegen for it, but we can switch this back when that happens. cc @aheejin

PR #22770, opened 2024-10-19
Allowing all binary C++ operators in WebIDL

### LINKED ISSUE #9415: WebIDL unsupported operators
I notice that it says in the WebIDL Binder docs that support is missing for some operators such as `Operator="*"`, `Operator="+"`, `Operator="-"`, etc.
https://emscripten.org/docs/porting/connecting_cpp_and_javascript/WebIDL-Binder.html

Is there a good reason behind this?
Could support be added?
Where do I start?

jrouwe:
This expands the list of supported operators by the 'Operator' WebIDL attribute to all binary operators that are supported by C++.

Fixes #9415

### DIFF
--- a/tools/webidl_binder.py
+++ b/tools/webidl_binder.py
@@ -614,7 +614,10 @@ def render_function(class_name, func_name, sigs, return_type, non_pointer,
         # this function comes from an ancestor class; for operators, we must cast it
         cast_self = 'dynamic_cast<' + type_to_c(func_scope) + '>(' + cast_self + ')'
       maybe_deref = deref_if_nonpointer(raw[0])
-      if '=' in operator:
+      operator = operator.strip()
+      if operator in ["+", "-", "*", "/", "%", "^", "&", "|", "=",
+                      "<", ">", "+=", "-=", "*=", "/=", "%=", "^=", "&=", "|=", "<<", ">>", ">>=",
+                      "<<=", "==", "!=", "<=", ">=", "<=>", "&&", "||"]:
         call = '(*%s %s %s%s)' % (cast_self, operator, maybe_deref, args[0])
       elif operator == '[]':
         call = '((*%s)[%s%s])' % (cast_self, maybe_deref, args[0])
--- a/site/source/docs/porting/connecting_cpp_and_javascript/WebIDL-Binder.rst
+++ b/site/source/docs/porting/connecting_cpp_and_javascript/WebIDL-Binder.rst
@@ -318,7 +318,7 @@ You can bind to C++ operators using ``[Operator=]``:
 .. note::
 
   - The operator name can be anything (``add`` is just an example).
-  - Support is currently limited to operators that contain ``=``: ``+=``, ``*=``, ``-=`` etc., and to the array indexing operator ``[]``.
+  - Support is currently limited to the following binary operators: ``+``, ``-``, ``*``, ``/``, ``%``, ``^``, ``&``, ``|``, ``=``, ``<``, ``>``, ``+=``, ``-=``, ``*=``, ``/=``, ``%=``, ``^=``, ``&=``, ``|=``, ``<<``, ``>>``, ``>>=``, ``<<=``, ``==``, ``!=``, ``<=``, ``>=``, ``<=>``, ``&&``, ``||``, and to the array indexing operator ``[]``.
 
 
 enums
--- a/test/webidl/post.js
+++ b/test/webidl/post.js
@@ -144,6 +144,8 @@ console.log(new TheModule.Inner().get());
 console.log('getAsArray: ' + new TheModule.Inner().getAsArray(12));
 new TheModule.Inner().mul(2);
 new TheModule.Inner().incInPlace(new TheModule.Inner());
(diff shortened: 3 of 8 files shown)


sbc100 (commented):
Seems reasonable to me.  @kripken WDYT?

kripken:
Hmm, I think the reason we focused on the `+=` form is because the C++ `x += y` turns into JS `x.add(y)`. Both of these operations are asymmetrical, with `x` playing a more central role. OTOH, with this change, we'd have C++ `x * y` turn into JS `x.mul(y)`, so something symmetrical turns into something asymmetrical. So there might be some risk of confusion there.

But maybe that risk is very small? Anyhow, if there is real-world use case I'm not opposed to it. Out of curiosity @jrouwe can you explain a bit about the use case you have in mind for this? I guess the library has only `*` and not `*=`?

jrouwe:
The use case is for a quaternion or matrix class.

These classes look like:

```
class Vector { ... };

class Matrix {
public:
  Vector operator *(const Vector &) { ... };
};
```

In this example the operation `Matrix * Vector => Vector` cannot be expressed as an `operator *=` because the return type is not a `Matrix` but a `Vector`.

Obviously I could define a function like `MultiplyVector` in C++ instead of using `operator *`, but I want my library ([JoltPhysics](https://github.com/jrouwe/JoltPhysics)) to have an intuitive interface in C++:

```C++
Matrix matrix = ...;
Vector vector = ...;
Vector result = matrix * vector;
```

B.t.w. the real class lives here: 
https://github.com/jrouwe/JoltPhysics/blob/9fb538ef9b945fe4d1ccab18bf979715fd90a312/Jolt/Math/Quat.h#L160

The IDL lives here:
https://github.com/jrouwe/JoltPhysics.js/blob/fa0df81fb3cffb4946a9e5740ecd68ced00f1fb7/JoltJS.idl#L382-L384

kripken:
Thanks for the info @jrouwe , yeah, that makes sense. I agree we should support this.

kripken (approved):
lgtm with an update to the docs:

https://emscripten.org/docs/porting/connecting_cpp_and_javascript/WebIDL-Binder.html#operators

File is at `site/source/docs/porting/connecting_cpp_and_javascript/WebIDL-Binder.rst`.

jrouwe:
Documentation updated.

kripken (approved):

tools/webidl_binder.py
```diff
@@ -614,12 +614,10 @@ def make_call_args(i):
...
         cast_self = 'dynamic_cast<' + type_to_c(func_scope) + '>(' + cast_self + ')'
       maybe_deref = deref_if_nonpointer(raw[0])
-      if '=' in operator:
-        call = '(*%s %s %s%s)' % (cast_self, operator, maybe_deref, args[0])
-      elif operator == '[]':
+      if operator == '[]':
         call = '((*%s)[%s%s])' % (cast_self, maybe_deref, args[0])
       else:
-        raise Exception('unfamiliar operator ' + operator)
+        call = '(*%s %s %s%s)' % (cast_self, operator, maybe_deref, args[0])
```
kripken, line 620: Can we error somehow if this is not one of the allowed operators? For example if it is a unary operator. Before this PR we'd error on unsupported things, and it would be good to keep doing so.

jrouwe: I've made the list explicit (based on https://en.cppreference.com/w/cpp/language/operators). Note that previously operators would be allowed as long as they contained a '=', so 'x=' would have been a valid operator (but doesn't compile obviously). Those will be caught now as well.

tools/webidl_binder.py
```diff
@@ -614,7 +614,10 @@ def make_call_args(i):
         # this function comes from an ancestor class; for operators, we must cast it
         cast_self = 'dynamic_cast<' + type_to_c(func_scope) + '>(' + cast_self + ')'
       maybe_deref = deref_if_nonpointer(raw[0])
-      if '=' in operator:
+      operator = operator.strip()
+      if any(x == operator for x in ["+", "-", "*", "/", "%", "^", "&", "|", "=",
+                                     "<", ">", "+=", "-=", "*=", "/=", "%=", "^=", "&=", "|=", "<<", ">>", ">>=",
+                                     "<<=", "==", "!=", "<=", ">=", "<=>", "&&", "||"]):
```
kripken, line 620:
```suggestion
      if operator in ["+", "-", "*", "/", "%", "^", "&", "|", "=",
                      "<", ">", "+=", "-=", "*=", "/=", "%=", "^=", "&=", "|=", "<<", ">>", ">>=",
                      "<<=", "==", "!=", "<=", ">=", "<=>", "&&", "||"]:
```

kripken: Unless I'm missing something?

jrouwe: Sorry, I don't use python very often... I've merged your suggestion.

kripken (approved):

tools/webidl_binder.py
kripken: No worries, looks good!

PR #22771, opened 2024-10-19
[webgpu] Unknown (undefined) features cause undefined to i32 array writes

Tellusim:
subj


kainino0x:
@lokokung I think this one is still valid in `emdawnwebgpu`, could you take a look at it and port forward as needed?
(seems right but haven't looked at it closely yet)

kainino0x (approved):
Finally reviewed. LGTM, thank you!

PR #22772, opened 2024-10-21
Added --thread_safe option to webidl_binder.py

jrouwe:
This emits all temporaries as 'static thread_local <type> temp;' instead of 'static <type> temp;' and prevents race conditions when you use the resulting library in a multithreaded environment.

E.g. if you create an IDL file like:

```
interface Vec3 {
	[Const, Value] Vec3 Cross([Const, Ref] Vec3 inRHS);
};
```

Previously the generated cpp file would look like:

```C++
const Vec3* EMSCRIPTEN_KEEPALIVE emscripten_bind_Vec3_Cross_1(Vec3* self, const Vec3* inRHS) {
  static Vec3 temp;
  return (temp = self->Cross(*inRHS), &temp);
}
```

Which has a race condition when you call into the `Cross` function from multiple JavaScript threads as they all use the same `temp` variable.

With this new parameter the resulting code will look like:

```C++
const Vec3* EMSCRIPTEN_KEEPALIVE emscripten_bind_Vec3_Cross_1(Vec3* self, const Vec3* inRHS) {
  static thread_local Vec3 temp;
  return (temp = self->Cross(*inRHS), &temp);
}
```

Which guarantees that no other threads overwrite `temp`.

### DIFF
--- a/tools/webidl_binder.py
+++ b/tools/webidl_binder.py
@@ -629,7 +629,8 @@ def render_function(class_name, func_name, sigs, return_type, non_pointer,
     if non_pointer:
       return_prefix += '&'
     if copy:
-      pre += '  static %s temp;\n' % type_to_c(return_type, non_pointing=True)
+      # Avoid sharing this static temp var between threads, which could race.
+      pre += '  static thread_local %s temp;\n' % type_to_c(return_type, non_pointing=True)
       return_prefix += '(temp = '
       return_postfix += ', &temp)'
 


tools/webidl_binder.py
```diff
@@ -59,6 +59,8 @@ def getExtendedAttribute(self, _name):
 parser = argparse.ArgumentParser()
 parser.add_argument('--wasm64', action='store_true', default=False,
                     help='Build for wasm64')
+parser.add_argument('--thread_safe', action='store_true', default=False,
```
sbc100, line 62: can you use dashes in the option name (e.g. "--thread-safe" )

tools/webidl_binder.py
```diff
@@ -626,10 +628,13 @@ def make_call_args(i):
     basic_return = 'return ' if constructor or return_type != 'Void' else ''
     return_prefix = basic_return
     return_postfix = ''
+    storage_attribute = ''
     if non_pointer:
       return_prefix += '&'
+    if thread_safe:
+      storage_attribute = 'thread_local '
```
sbc100, line 635: Maybe we should just always emit this, when TLS gets lowered away when not building with thread support?

tools/webidl_binder.py
jrouwe: Fixed.

tools/webidl_binder.py
jrouwe: We could, and you're right that if you build without `-pthread` the resulting code is the same:

https://godbolt.org/z/6cE6T8a47

When adding `-pthread` there is an extra instruction which adds a bit of overhead:

image

I cannot predict if people are ok with this overhead. Their library may use multiple threads internally, but maybe they don't interact with the library from multiple threads. In that case there's no benefit in adding the `thread_local` attribute.

sbc100: I think if you are doing a pthread build we should just default to being thread safe.

jrouwe: Ok, in that case the change becomes really simple. I've pushed the new code.

tools/webidl_binder.py
```diff
@@ -629,7 +629,7 @@ def make_call_args(i):
     if non_pointer:
       return_prefix += '&'
     if copy:
-      pre += '  static %s temp;\n' % type_to_c(return_type, non_pointing=True)
+      pre += '  static thread_local %s temp;\n' % type_to_c(return_type, non_pointing=True)
```
kripken, line 632: Please add a comment, maybe something like "Avoid sharing this static temp var between threads, which could race." Otherwise looks great.

jrouwe: Done.

kripken (approved):

PR #22773, opened 2024-10-22
[wasm64] Remove temporary toIndexType workaround

sbc100:
Now the firefox and chrome both use BigInt for indexing 64-bit memories and tables this workaround is no longer needed.


dschuff (approved):

src/parseTools.mjs
```diff
@@ -964,8 +964,7 @@ function from64Expr(x, assign = true) {
 }
 
 function toIndexType(x) {
-  if (MEMORY64 != 1) return x;
-  return `toIndexType(${x})`;
+  return to64(x);
```
kripken, line 967: Should we replace calls from `toIndexType` to `to64`? Or do we think the distinction is important/might matter later?

kripken (approved):

src/parseTools.mjs
sbc100: Yes, I thought of doing that.  Can always be done as followup I guess?

kripken: Followup sounds fine, but also I am not sure if that is worthwhile. Unless we are certain we won't care about the difference later (which I am not sure of myself).

PR #22774, opened 2024-10-22
Remove use of `Module['websocket']`. NFC

sbc100:
We can still receive the initial value for websocket from the Module object but there is not need to continue to access it via the Module object.

### DIFF
--- a/src/library_sockfs.js
+++ b/src/library_sockfs.js
@@ -10,37 +10,35 @@ addToLibrary({
   },
   $SOCKFS__deps: ['$FS'],
   $SOCKFS: {
+#if expectToReceiveOnModule('websocket')
+    websocketArgs: {},
+#endif
+    callbacks: {},
+    on(event, callback) {
+      SOCKFS.callbacks[event] = callback;
+    },
+    emit(event, param) {
+      SOCKFS.callbacks[event]?.(param);
+    },
     mount(mount) {
-      // If Module['websocket'] has already been defined (e.g. for configuring
-      // the subprotocol/url) use that, if not initialise it to a new object.
-      Module['websocket'] = (Module['websocket'] &&
-                             ('object' === typeof Module['websocket'])) ? Module['websocket'] : {};
-
+#if expectToReceiveOnModule('websocket')
+      // The incomming Module['websocket'] can be used for configuring 
+      // configuring subprotocol/url, etc
+      SOCKFS.websocketArgs = {{{ makeModuleReceiveExpr('websocket', '{}') }}};
       // Add the Event registration mechanism to the exported websocket configuration
       // object so we can register network callbacks from native JavaScript too.
       // For more documentation see system/include/emscripten/emscripten.h
-      Module['websocket']._callbacks = {};
-      Module['websocket']['on'] = /** @this{Object} */ function(event, callback) {
-        if ('function' === typeof callback) {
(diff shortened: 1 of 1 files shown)


src/library_sockfs.js
```diff
@@ -10,37 +10,35 @@ addToLibrary({
   },
   $SOCKFS__deps: ['$FS'],
   $SOCKFS: {
+#if expectToReceiveOnModule('websocket')
+    websocket_args: {},
```
kripken, line 14:
```suggestion
    websocket: {},
```
If we do need "args" in the name, should it be `websocketArgs`?

src/library_sockfs.js
```diff
@@ -207,12 +201,6 @@ addToLibrary({
               opts = subProtocols;
             }
 
-            // some webservers (azure) does not support subprotocol header
-            if (runtimeConfig && null === Module['websocket']['subprotocol']) {
-              subProtocols = 'null';
-              opts = undefined;
-            }
```
kripken, line 214: Separate change? (I don't seem to see this code moved anywhere else.)

sbc100: I specially handle the `=== null` case above.

src/library_sockfs.js
sbc100: There are some other existing props of this object that use snake_case (e.g. websocket_sock_ops) but I made this change anyway.

I think its good to have the name args in there.. since `websocket` on its own is a little misleading.. sounds like a noun.

src/library_sockfs.js
```diff
@@ -10,37 +10,35 @@ addToLibrary({
...
-        }
-        return this;
-      };
-
-      Module['websocket'].emit = /** @this{Object} */ function(event, param) {
-        if ('function' === typeof this._callbacks[event]) {
-          this._callbacks[event].call(this, param);
-        }
-      };
+      (Module['websocket'] ??= {})['on'] = SOCKFS.on;
```
kripken, line 31: Do we not need `'emit'` as well? Looks like before we had both exported.

src/library_sockfs.js
```diff
@@ -169,36 +167,32 @@ addToLibrary({
...
+#if expectToReceiveOnModule('websocket')
+            // Fetch runtime WebSocket URL config.
+            if (SOCKFS.websocketArgs['url']) {
+              url = SOCKFS.websocketArgs['url'];
             }
+            // Fetch runtime WebSocket subprotocol config.
+            if (SOCKFS.websocketArgs['subprotocol']) {
+              subProtocols = SOCKFS.websocketArgs['subprotocol'];
+            } else if (SOCKFS.websocketArgs['subprotocol'] === null) {
+              subProtocols = 'null'
```
kripken, line 187:
```suggestion
              subProtocols = 'null';
```

kripken (approved):

src/library_sockfs.js
sbc100: No, I don't think so.   The `emit` for for internal use to send the message.  The `on` is what users would use to register callback.

sbc100: You can see that the `on` property is quoted so that closure won't minify it but the `emit` helper function is not.

sbc100: See the docs here: https://emscripten.org/docs/api_reference/emscripten.h.html#socket-event-registration

PR #22776, opened 2024-10-22
Convert all remaining wasm worker tests to btest_exit. NFC

sbc100:
Amongst other things this allows all these test to be run under node.

### DIFF
--- a/test/atomic/test_wait64_notify.c
+++ b/test/atomic/test_wait64_notify.c
@@ -47,11 +47,16 @@ void run_test() {
 
 char stack[1024];
 
+void do_exit() {
+  emscripten_out("do_exit");
+  emscripten_terminate_all_wasm_workers();
+  emscripten_force_exit(0);
+}
+
 void worker_main() {
   run_test();
-#ifdef REPORT_RESULT
-  REPORT_RESULT(addr >> 32);
-#endif
+  assert(addr == 0x300000000ull);
+  emscripten_wasm_worker_post_function_v(EMSCRIPTEN_WASM_WORKER_ID_PARENT, do_exit);
 }
 
 #else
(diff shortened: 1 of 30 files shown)


test/atomic/test_wait64_notify.c
```diff
@@ -91,6 +96,7 @@ int main() {
 #ifdef __EMSCRIPTEN_WASM_WORKERS__
   emscripten_wasm_worker_t worker = emscripten_create_wasm_worker(stack, sizeof(stack));
   emscripten_wasm_worker_post_function_v(worker, worker_main);
+  emscripten_runtime_keepalive_push();
```
kripken, line 99: I slightly worry about converting all the tests to this form, as our users may be writing in the simpler way?

sbc100: I guess I can make this part implicit by having `post_function` itself keep the runtime alive.

The other part of replacing `REPORT_RESULT` with normal exit I hope/think is fine.

sbc100: Actually post_function doesn't track when job is done so that don't work.

I think the benefit here of making these test runnable under node, along with using a more normal exit-status-based test setup outweighs the downside here of having code that a user might not write.   Note that `emscripten_runtime_keepalive_push` simply does nothing in default build without `-sEXIT_RUNTIME` since the runtime always stays alive regardless in the default mode.

sbc100: OK, maybe we can land just a small subset for now? #22777

kripken: I'd be ok landing the whole thing, so long as we keep one test in the old form (something like a "wasm workers without exit runtime" test).

Issue #12203, opened 2020-09-13
Replace ENVIRONMENT_* logic with Web default + polyfills?


kripken:
Right now we support various "environments": web, node, other JS shells, etc. We need to use different code paths for loading a file, for example - fetch on the web versus node's `fs` package, and so forth. We have a layer of indirection for this, where in `shell.js` the environment is detected and hooks like `readBinary`, `readAsync` are set up.

This has two current downsides. First, that while you can specify that only a single target environment is supported, the detection + indirection layer may still end up increasing code size. Second, we get requests for more environments to be supported now and then, like electron variants or deno, and they require adding an entire new "environment".

Instead, what if we made the following change:
* The default code path uses Web APIs directly.
* To support other environments, they would polyfill the Web environment. For example, the default code path would use `fetch`, and other environments would need to implement a minimal polyfill for that. So for node, `fs.readFile` would be used in the polyfill, etc.

Benefits:
* In practice, the web target is the one that cares the most about code size. By having the default code path use Web APIs, code size would be minimal there.
* Supporting more environments can be done in a pluggable way: add a file with code to polyfill Web APIs for whatever you want. We can have `env/node.js, env/deno.js` etc. as needed, and a user can also add a custom one of their own just by including the code with `--pre-js`.

curiousdannii:
Great idea. But a few thoughts.

Traditionally polyfills have aimed to completely implement the relevant standard by implementing it as the . So a polyfill for `fetch` would need to completely implement the full spec. We don't want that. We only care about a small subset of the full specification. So rather than talking in terms of polyfills I think it would be better to think of a small Emscripten Environment API. Such an API like that would only have the things Emscripten itself uses. So it wouldn't need to handle all the options of `fetch`, but just the few options we need.

Can I suggest that each env file pushes a function to an array, and then at the appropriate time the shell code will run the functions in series. Each function should check if a component in the API has already been supplied, and if not it checks to see if it can supply it. The function should be an async function - this will allow us to fix #11792. 

My last thought is whether the default should be for modern browsers, or whether it should be for a more minimalistic and agnostic environment that might support the core ES standard but not might else. For example for many people `Math.random` might be adequate randomness, and the extra code for using the Crypto API might not be worth it (probably not a good example as I don't think it actually uses too much code.)

I'm not at my dev system now, but I'll take a look later to try to put together a list of functions this environment API would need.

sbc100:
[...]

PR #22778, opened 2024-10-23
Fix `ENVIRONMENT_IS_WORKER` detection on Deno

kleisauke:
`importScripts()` is always `undefined` on Deno.


kleisauke:
Test case:

Before:
```console
$ cat > test.c << EOF
#include <assert.h>
#include <pthread.h>
#include <stdio.h>

#include <emscripten/em_asm.h>

void* worker_thread(void *arg) {
  int is_worker = EM_ASM_INT(return ENVIRONMENT_IS_WORKER);
  printf("ENVIRONMENT_IS_WORKER: %d\n", is_worker);
  return NULL;
}

int main() {
  int is_worker = EM_ASM_INT(return ENVIRONMENT_IS_WORKER);
  printf("ENVIRONMENT_IS_WORKER: %d\n", is_worker);

  pthread_t thread;
  pthread_create(&thread, NULL, worker_thread, NULL);
  pthread_join(thread, NULL);

  return 0;
}
EOF
$ emcc test.c -o test.js -sASSERTIONS=0 -pthread -sEXIT_RUNTIME
$ node test.js
ENVIRONMENT_IS_WORKER: 0
ENVIRONMENT_IS_WORKER: 1
$ emcc test.c -o test.mjs -sENVIRONMENT=web,worker -sASSERTIONS=0 -pthread
$ echo "import Module from './test.mjs';await Module();Deno.exit()" | deno run --allow-read -
ENVIRONMENT_IS_WORKER: 0 [repeated x times until OOM]
```

After:
```console
$ emcc test.c -o test.mjs -sENVIRONMENT=web,worker -sASSERTIONS=0 -pthread
$ echo "import Module from './test.mjs';await Module();Deno.exit()" | deno run --allow-read -
ENVIRONMENT_IS_WORKER: 0
ENVIRONMENT_IS_WORKER: 1
```

kripken:
Looks reasonable to me. The downside is maybe a few more bytes in size, but I don't see a way around that, and it mostly affects pthreads builds (which are larger anyhow).

`other.test_no_pthread` on CI will need to be updated a little, it looks like (it greps for `Worker`, which this PR adds; can first replace `WorkerGlobalScope` with another string, if there isn't anything nicer).

sbc100 (approved):

src/runtime_pthread.js
```diff
@@ -31,14 +31,9 @@ if (ENVIRONMENT_IS_PTHREAD) {
 
     Object.assign(globalThis, {
       self: global,
-      // Dummy importScripts.  The presence of this global is used
-      // to detect that we are running on a Worker.
-      // TODO(sbc): Find another way?
-      importScripts: () => {
```
sbc100, line 37: I'm a little confused because you say that deno never defines `importScripts`, but from the looks of this code it seems that node doesn't define it either.   Isn't the point of this line to inject it?

kleisauke: Ah, you're right. Node.js doesn't define `importScripts` either.
```js
const { Worker, isMainThread } = require('node:worker_threads');

if (isMainThread) {
  // This re-loads the current file inside a Worker instance.
  new Worker(__filename);
} else {
  console.log(typeof importScripts);  // Prints 'undefined'.
}
```

Commit 4bab494ba88b664241bbff2b4427d5b24ed9e16a removes this altogether, since `ENVIRONMENT_IS_WORKER` no longer depends on this dummy symbol on Node.js:
https://github.com/emscripten-core/emscripten/blob/9e618aaaf4bfec941f04d909549ef76d880e2200/src/shell.js#L144

(I can separate this change into its own PR if needed)

kleisauke: Note that `ENVIRONMENT_IS_NODE` is `false` on Deno. IIUC, Deno tends to favor standard Web Platform APIs over proprietary ones and wasn't originally designed to be a drop-in replacement for Node.js.

kripken: Deno does have a [Node.js compatibility layer](https://docs.deno.com/runtime/fundamentals/node/) these days (they started to compromise on that at some point, though they did start with "pure Web APIs" IIRC). I wonder if maybe we can make `IS_NODE` also accept deno, for simplicity? Separate from this PR of course. (And if it isn't trivial, then https://github.com/emscripten-core/emscripten/issues/12203 is probably the best long-term plan instead.)

This PR itself lgtm as is. Unless @sbc100 you had more thoughts?

sbc100 (approved):

PR #22779, opened 2024-10-23
WebIDL binder: Added 'BindTo' extended attribute

jrouwe:
This attribute can be used to change the C++ function that will be called. It can be used expose functions that have multiple overloads with the same number of parameters.

Example C++ code:

```
class BindToTest {
public:
  int test(const char*) { return 1; }
  int test(int) { return 2; }
};
```
The following IDL will generate an error:

```
interface BindToTest {
  void BindToTest();
  long test([Const] DOMString arg);
  long test(long arg);
};
```

```
AssertionError: overloading must differentiate by # of arguments (cannot have two signatures that differ by types but not by length)
```

New IDL:

```
interface BindToTest {
  void BindToTest();
  [BindTo="test"] long testString([Const] DOMString arg);
  [BindTo="test"] long testInt(long arg);
};
```

Now the two overloaded functions can be called from JavaScript because they have a different name. It can also be used if you just want to rename the function, e.g. in my C++ library the function naming convention is `CamelCase`, in JavaScript it is `camelCase`.

I've run into this issue multiple times and so far I've been forced to choose the overload that I want to expose to JavaScript.

Actual use case: I have an axis aligned box class. This class has multiple `Encapsulate` functions (one to grow the box by including a point, one for including another box and one to include a triangle):

https://github.com/jrouwe/JoltPhysics/blob/9fb538ef9b945fe4d1ccab18bf979715fd90a312/Jolt/Geometry/AABox.h#L53-L75

My new IDL is this:

```
interface AABox {
	void AABox();
	[BindTo="Encapsulate"] void EncapsulateVec3([Const, Ref] Vec3 inV);
	[BindTo="Encapsulate"] void EncapsulateAABox([Const, Ref] AABox inBox);
	[BindTo="Encapsulate"] void EncapsulateTriangle([Const, Ref] Triangle inTriangle);
};
```

Note: I will update the documentation if the idea in this PR is approved.

Note 2: `BindTo` was the best name I could come up with. I'm open to suggestions.


kripken (commented):
Good idea, and the code so far looks right. I don't have a better idea for a name.

lgtm with the docs

jrouwe:
Documentation was updated. 

I moved the unit test code higher up as that seems to fix the strange warning that makes ```./test/runner wasm2js1``` fail:

```
Warning: Enlarging memory arrays, this is not fast! 16908288,25296896
```

(I'm having trouble running unit tests locally, many fail even without my changes)

jrouwe:
I think the failure of `./test/runner other skip:other.test_native_link_error_message` is not related to my change (or at least I don't see how), let me know if it is.

site/source/docs/porting/connecting_cpp_and_javascript/WebIDL-Binder.rst
```diff
@@ -416,6 +416,55 @@ When C++ code has a pointer to a ``Base`` instance and calls ``virtualFunc()``,
   - You *must* implement all the methods you mentioned in the IDL of the ``JSImplementation`` class (``ImplJS``) or compilation will fail with an error.
   - You will also need to provide an interface definition for the ``Base`` class in the IDL file.
 
+Function overloads
+==================
+
+C++ allows function overloads, where multiple member functions have the same name but different arguments. *WebIDL Binder* allows you to bind overloaded functions only if they differ in the number of arguments:
```
kripken, line 422:
```suggestion
C++ allows function overloads, where multiple member functions have the same name but different arguments. By default, the *WebIDL Binder* allows you to bind overloaded functions if they differ only in the number of arguments:
```

site/source/docs/porting/connecting_cpp_and_javascript/WebIDL-Binder.rst
```diff
@@ -416,6 +416,55 @@ When C++ code has a pointer to a ``Base`` instance and calls ``virtualFunc()``,
...
+.. code-block:: idl
+
+  // WebIDL
+  interface OverloadTest {
+    void OverloadTest();
+    void test(long arg1, long arg2);
+    void test(long arg);
+  };
+
+If your functions don't differ in number of arguments, you can use the ``[BindTo]`` attribute:
```
kripken, line 442:
```suggestion
If your overloaded functions differ in some other way (say, in the types) then you can use the ``[BindTo]`` attribute to tell the tool what function name to bind to (that is, to call):
```

PR #22801, opened 2024-10-29
Remove automatic fallback to $HOME/.emscripten_cache. NFC

sbc100:
Putting the cache in the `$HOME` directory is not a great idea since its a location that is shared by any/all versions of emscripten.  If anyone really wants to use that location they can do so via the config file setting `CACHE` or via the environment variable `EMCC_CACHE`.

### DIFF
--- a/tools/cache.py
+++ b/tools/cache.py
@@ -23,12 +23,19 @@ cachelock = None
 cachelock_name = None
 
 
+def is_writable(path):
+  return os.access(path, os.W_OK)
+
+
 def acquire_cache_lock(reason):
   global acquired_count
   if config.FROZEN_CACHE:
     # Raise an exception here rather than exit_with_error since in practice this
     # should never happen
     raise Exception('Attempt to lock the cache but FROZEN_CACHE is set')
+
+  if not is_writable(cachedir):
+    utils.exit_with_error(f'cache directory "{cachedir}" is not writable while accessing cache for: {reason} (see https://emscripten.org/docs/tools_reference/emcc.html for info on setting the cache directory)')
 
   if acquired_count == 0:
     logger.debug(f'PID {os.getpid()} acquiring multiprocess file lock to Emscripten cache at {cachedir}')
@@ -67,7 +74,11 @@ def lock(reason):
 
 def ensure():
   ensure_setup()
-  utils.safe_ensure_dirs(cachedir)
+  if not os.path.isdir(cachedir):
+    parent_dir = os.path.dirname(cachedir)
+    if not is_writable(parent_dir):
+      utils.exit_with_error('unable to create cache directory "{cachdir}": parent directory not writable (see https://emscripten.org/docs/tools_reference/emcc.html for info on setting the cache directory)')
+    utils.safe_ensure_dirs(cachedir)
 
 
 def erase():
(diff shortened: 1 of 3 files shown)


tools/cache.py
```diff
@@ -23,13 +23,20 @@
...
+
 def acquire_cache_lock(reason):
   global acquired_count
   if config.FROZEN_CACHE:
     # Raise an exception here rather than exit_with_error since in practice this
     # should never happen
     raise Exception('Attempt to lock the cache but FROZEN_CACHE is set')
 
+  if not is_writable(cachedir):
+    utils.exit_with_error(f'Cache directory "{cachedir}" is not writable while accesing cache for: {reason}')
```
kripken, line 38:
```suggestion
    utils.exit_with_error(f'Cache directory "{cachedir}" is not writable while accessing cache for: {reason}')
```

kripken: Also, maybe we can link to a doc page on the website (which explains `CACHE` etc.)?

tools/cache.py
```diff
@@ -67,7 +74,11 @@ def lock(reason):
 
 def ensure():
   ensure_setup()
-  utils.safe_ensure_dirs(cachedir)
+  if not os.path.isdir(cachedir):
+    parent_dir = os.path.dirname(cachedir)
+    if not is_writable(parent_dir):
+      utils.exit_with_error('unable to create cache directory "{cachdir}": parent directory not writable')
```
kripken, line 80: (same here)

tools/cache.py
sbc100: I don't know if we have such a page that describes each of the config settings like that.  Perhaps we should.   The best I could find is https://emscripten.org/docs/building_from_source/configuring_emscripten_settings.html. 

Remember that all this stuff only applies to the more advanced users out there, not users of emsdk.

kripken: Agreed it's more advanced users, but I'm just trying to make this breaking change less annoying for them.

I see we have this page

https://emscripten.org/docs/tools_reference/emcc.html

whose entry for `--cache` (I can't seem to link to it directly) mentions `CACHE`. Maybe that's a good place to link to?

sbc100: Ok done.

kripken (approved):
Thanks!

I imagine the link could help someone. Like imagine a dev is maintaining an emcc build that they didn't set up themselves, and it breaks due to this. They aren't a power user and just want to fix it quickly, so the more info we can give them, the better.

neondev9:
is it really expected that emscripten compiler doesn't create --cache/EM_CACHE directories now?
i.e. if you override cache directory, you need to guarantee that directory exists before compilation.

sbc100:
> is it really expected that emscripten compiler doesn't create --cache/EM_CACHE directories now? i.e. if you override cache directory, you need to guarantee that directory exists before compilation.
> 
> Just updated SDK from 3.1.57 and this change affected Github Action pipelines because of `em++: error: unable to create cache directory "/home/runner/work/Project/Project/.tmp/Emscripten/cache": parent directory not writable (see https://emscripten.org/docs/tools_reference/emcc.html for info on setting the cache directory)`

Sorry, no that was not an indented change in behaviour.   Sounds like a case we don't cover in out tests.

Issue #22790, opened 2024-10-25
FS no longer populated after "Micro-optimize preRun/postRun handling. NFC (#22671)"


stbergmann:
As originally described at <https://groups.google.com/g/emscripten-discuss/c/eJyqgMCWnCs/m/SrA2CtuBAQAJ> 'FS no longer populated after 'Micro-optimize preRun/postRun handling. NFC (#22671)'":

When building my app (LibreOffice on top of Qt6) with a build of latest emsdk 3.1.69, I found that the virtual file system would no longer be populated and only contained the bare hardcoded minimum.

Turned out that <https://github.com/emscripten-core/emscripten/commit/483cdcd5306780520a2de8856fcd13910e71497e> "Micro-optimize preRun/postRun handling. NFC (https://github.com/emscripten-core/emscripten/pull/22671)"
```
> diff --git a/src/preamble.js b/src/preamble.js
> index d37cb9fe7..3a6aee9c9 100644
> --- a/src/preamble.js
> +++ b/src/preamble.js
> @@ -195,11 +195,10 @@ function preRun() {
>    assert(!ENVIRONMENT_IS_PTHREAD); // PThreads reuse the runtime from the main thread.
>  #endif
>  #if expectToReceiveOnModule('preRun')
> -  if (Module['preRun']) {
> -    if (typeof Module['preRun'] == 'function') Module['preRun'] = [Module['preRun']];
> -    while (Module['preRun'].length) {
> -      addOnPreRun(Module['preRun'].shift());
> -    }
> +  var preRuns = Module['preRun'];
> +  if (preRuns) {
> +    if (typeof preRuns == 'function') preRuns = [preRuns];
> +    preRuns.forEach(addOnPreRun);
>    }
>  #endif
>    callRuntimeCallbacks(__ATPRERUN__);
```
caused the function `runWithFS` to no longer be called, because it gets added to` Module.preRun` from within the function `runMetaWithFS` that itself is called from within the function `preRun`.

Locally reverting that commit fixed the issue for me for now.

sbc100:
Revert in flight. 

BTW, are you attached to using file preloading?  Does file embedding not for your use case for some reason?

sbc100:
Actually looking at the code I'm struggling to see where `preRun` is added within `runMetaWithFS`: https://github.com/emscripten-core/emscripten/blob/fb2ae6165e786f13d1202403d49767039f7d88ad/tools/file_packager.py#L1105-L1116

[...]

PR #22802, opened 2024-10-29
file_packager: Fix regression in --separate-metadata output

sbc100:
This reverts commit 483cdcd5306780520a2de8856fcd13910e71497e.

This also adds some testing of the `--separate-metadata` file_packager options where we actually run the generated code.  As part of adding this test I also made the code work under node.

Fixes: #22790


kripken (approved):
Can we add a test?

sbc100:
I've added a test as a separate commit.  It also fixes `--separate-metadata` so that it runs under node.  OK to land these together or should I land the test after the revert?

test/test_other.py
```diff
@@ -3620,6 +3620,21 @@ def test_file_packager_separate_metadata(self):
...
+    int main() {
+      struct stat buf;
+      assert(stat("data1.txt", &buf) == 0);
+      assert(stat("subdir/data2.txt", &buf) == 0);
+      printf("done\n");
+      return 0;
+    }
+    ''')
+    self.do_runf('src.c', emcc_args=['--pre-js=immutable.js', '-sFORCE_FILESYSTEM'])
+
```
kripken, line 3637: I'm not opposed to adding the test + node fix in this PR, that sgtm, but can you add an explanation for the test? It just appends to an existing test and I can't figure out what specifically it is testing, and why it failed before this reversion.

kripken: I mean that it looks like many other tests we already have coverage for, so I'm missing what is special here.

sbc100: There were no other node tests that ran with `--separate-metadata`.   It was that options specifically that was broken.

kripken:
Thanks, lgtm

PR #22805, opened 2024-10-30
Use nullish assignment in file_packager.py. NFC

sbc100:


kripken (commented):
Do we run the transpiler on file packager output? When we combine that output with the main JS we do, but users can also run file packager output directly.

sbc100:
> Do we run the transpiler on file packager output? When we combine that output with the main JS we do, but users can also run file packager output directly.

If users run the file packager directly it think its reasonable to generate modern JS and depend on users to transpile it if they want to.   Users of the stand alone file packager I fairly few and likely more advanced so I think its reasonable for them to understand such things.

kripken:
Fair point. Sounds reasonable to me. But perhaps we should add a changelog entry for this, as it is a breaking change potentially for such users? Something like

```
The file packager tool will now emit more modern JS. When emcc transpiles code to support older browsers it will convert that code together with all the rest (using Babel), but if you run the file packager directly yourself then you may need to do so manually (if you want to support legacy browsers).
```

sbc100:
Done

ChangeLog.md
```diff
@@ -21,14 +21,19 @@ See docs/process.md for more on how version tagging works.
...
   `memset` and similar functions for which it is a documented undefined
   behavior (#22798). See https://github.com/llvm/llvm-project/pull/111434
 - The automatic fallback to `$HOME/.emscripten_cache` when the emscripten
   directory is read-only was removed.  This automatic behaviour could cause
   confusion.  Anyone who really wants to use `$HOME/.emscripten_cache` can
   still do so either via an environment variable (`EMCC_CACHE`) or via a config
   file setting `CACHE`.
+- The standalone `file_packager.py` tool now outputs modern JS (specifically it
+  includes nullish assignment).  If you use this output directly and you want
+  to support older browser you may need to transpile it.  If you use
```
kripken, line 34:
```suggestion
  to support older browsers you may need to transpile it.  If you use
```

kripken (approved):

PR #22806, opened 2024-10-30
Fix harfbuzz build after new clang warning was added. NFC

sbc100:
Followup to #22798


sbc100:
Will land TBR to get the roller going again.

PR #22818, opened 2024-10-31
[library_sockfs] Report socket connect failures via select()

sbc100:
When connect fails on a NON_BLOCKING socket the socket is supposed to become writable and the status made available via SO_ERROR.

Without this change `test_sockets_echo_client.c` will loop forever polling for read/write status on the socket fd.

### DIFF
--- a/src/library_sockfs.js
+++ b/src/library_sockfs.js
@@ -281,6 +281,7 @@ addToLibrary({
 #endif
 
           Module['websocket'].emit('open', sock.stream.fd);
+          sock.connecting = false;
 
           try {
             var queued = peer.msg_send_queue.shift();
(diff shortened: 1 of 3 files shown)


src/library_sockfs.js
```diff
@@ -383,6 +383,10 @@ addToLibrary({
           return sock.pending.length ? ({{{ cDefs.POLLRDNORM }}} | {{{ cDefs.POLLIN }}}) : 0;
         }
 
+        if (sock.error) {
+          return {{{ cDefs.POLLOUT }}};
+        }
```
kripken, line 388: Can you explain and maybe add a comment, why all errors are treated as POLLOUT ("writing is now possible")? Are no other errors possible?

sbc100: I added a comment.

While it is possible I'm casting too wide of a net here and other times that `sock.error` are set do not warrant being marked as readable, it would require more state tracking to determine that the source of the error was a call to `connect` since as far as I can tell we don't track the current state of the connection.

kripken (approved):

src/library_sockfs.js
sbc100: Turns out I did need the extra state in order to be more precise here.  Cleaner now.

PR #22819, opened 2024-10-31
implement getentropy via __wasi_random_get.

### LINKED ISSUE #22782: Standalone: implement getentropy() via wasi_snapshot_preview1.random_get()

**Version of emscripten/emsdk:** 3.1.67 (and likely latest)

Compiling absl::random code in standalone mode results in a wasm module that crashes. Sample code:

```cpp
absl::BitGen bitgen;
int digit = absl::Uniform(bitgen, 0, 2);
```

The reason is that Abseil under Emscripten uses the `getentropy()` syscall as a seed: https://github.com/abseil/abseil-cpp/blob/07fff76ed619672bb7bf00939231ba45aef2b043/absl/random/internal/seed_material.cc#L60-L64

And standalone Emscripten defaults to a crashing implementation:
https://github.com/emscripten-core/emscripten/blob/b53978ee3f540dc74761eba127aa7f1b8761a125/system/lib/standalone/standalone.c#L139-L143

Would it be reasonable to use `wasi_snapshot_preview1.random_get()` to implement `getentropy()`?

walkingeyerobot:
fixes #22782 

I'm not sure how to get this tested. Any feedback would be appreciated.


system/lib/standalone/standalone.c
```diff
@@ -139,7 +139,7 @@ weak int __syscall_lstat64(intptr_t path, intptr_t buf) {
 // There is no good source of entropy without an import. Make this weak so that
 // it can be replaced with a pRNG or a proper import.
 weak int getentropy(void* buffer, size_t length) {
-  abort();
```
sbc100, line 142: Can you copy the code from https://github.com/WebAssembly/wasi-libc/blob/main/libc-bottom-half/sources/getentropy.c?   (With a link/attribution I think).

system/lib/standalone/standalone.c
```diff
@@ -139,7 +139,7 @@ weak int __syscall_lstat64(intptr_t path, intptr_t buf) {
 // There is no good source of entropy without an import. Make this weak so that
 // it can be replaced with a pRNG or a proper import.
```
sbc100, line 140: I think this comment can now be removed.

sbc100:
From the CI failures it looks like you need to perhaps update/rebase on top of the main branch?

sbc100 (approved):

system/lib/standalone/standalone.c
walkingeyerobot: done. should I leave it `weak`?

system/lib/standalone/standalone.c
walkingeyerobot: done.

sbc100:
I actually think we should probably just use this implementation everywhere (not just in standalone mode), then we can implement it in JS and get it tested that way.   I'm happy to take over this work if you like?

sbc100:
Closing in favor of #22820

PR #22820, opened 2024-10-31
Implement getentropy via __wasi_random_get

sbc100:
This allows it to work in standalone mode.

This change also includes a fix to actually honor the `__proxy: 'none'` mode.  It seems like we were just falling back to async proxying in this case, but we only had a single function that used this mode which was ` __syscall_fadvise64`.

Fixes: #22782


src/library_wasi.js
```diff
@@ -568,6 +568,65 @@ var WasiLibrary = {
...
+  $randomFill: (view) => {
+    // Lazily init on the first invocation.
+    return (randomFill = initRandomFill())(view);
+  },
+
+  random_get__deps: ['$randomFill'],
+  random_get: (buffer, size) => {
+    randomFill(HEAPU8.subarray(buffer, buffer + size));
+    return 0;
+  },
```
kripken, line 629: There is a previous implementation in this file,

https://github.com/emscripten-core/emscripten/blob/2f79c58ce1390a47673ca8aa07e638b684da396f/src/library_wasi.js#L44-L48

sbc100: Oops, well spotted

kripken (approved):

PR #22827, opened 2024-11-01
[file_packager] Remove support for legacy locateFilePackage

sbc100:
This old name was removed back in 2014 (bdfa7bc963)


kripken (approved):
Removal after a decade of deprecation... good find!

Issue #22394, opened 2024-08-18
Compiled code using pthreads breaks Vite bundler


jamsinclair:
**Description**

When code using threads is compiled with Emscripten and then bundled with Vite it does not bundle and throws the following Vite error:

```
[vite:worker-import-meta-url] Vite is unable to parse the worker options as the value is not static.To ignore this error, please use /* @vite-ignore */ in the worker options.
file
```

Vite has required static worker options for some time, over 2 years. Recent changes to Emscripten have made output code incompatible with the Vite bundler.

**Version of emscripten/emsdk:**

Affects versions >= `v3.1.58`

**Reproduction:**
I have created a repository at https://github.com/jamsinclair/emscripten-worker-options-vite-reproduction which show cases the problem.

**Potential Remedy:**

Refactor [library_pthread.js](https://github.com/emscripten-core/emscripten/blob/main/src/library_pthread.js) to use static values for the worker options. It'll be more verbose, but, it'll make the code compatible with Vite and maintain compatibility with other bundlers.

jamsinclair:
Vite users could potentially write a simple rollup or vite plugin to work around this. Although, if we could fix this in Emscripten itself, that would be amazing 🙏.

I know emscripten is in a tricky spot trying to support bundlers and their different nuances 😅

x0YYYY0:
I have run into this issue as well. Thank you for posting about it.  This is also unfortunate for those of us who are looking to use just shared memory, as the pthreads build option is the most comprehensive build option for that use case.

sbc100:
Yup, we can/should fix this.  I wonder if we can write a test for it so we don't regress again.

Kind of still that vite can't see that the options are actually static.. just declared in a non-escaping var right before I think.

dr-matt:
I've also run into this issue recently. Is there currently a workaround?

zamfofex-dummy:
You can just use version 3.1.57, and it should work.

sbc100:
We should fix this.  I'll try to add some tests to emscripten so that we don't regress again too.

dr-matt:
@zamfofex-dummy thank you, good to know that's the version just before the regression. Unfortunately it would be tricky to move back to that version, as we're relying on other more recent changes.

@sbc100 that would be fantastic. Interestingly, we have one use case that works fine, but adding `-g2` to the build produces this error. Haven't dug into why yet, might be project-specific.

jamsinclair:
[...]

PR #22834, opened 2024-11-02
Fix pthread worker options usage with Vite bundler

jamsinclair:
## Overview

An attempt at solving #22394.

When code using threads is compiled with Emscripten and then bundled with Vite it does not bundle and throws an error.

This is due to a variable reference for worker options being used during worker construction. Vite needs static worker options to bundle the code successfully.

As discussed in the issue, the easiest solution would be to switch to using an object literal for each worker construction.


tools/link.py
```diff
@@ -2098,6 +2098,9 @@ def phase_final_emitting(options, state, target, wasm_target):
 
   target_dir = os.path.dirname(os.path.abspath(target))
 
+  if settings.PTHREADS:
+    replace_worker_options_placeholders(final_js)
```
jamsinclair, line 2102: Based on test failures, is there an early location where we should be replacing these values? 🤔

jamsinclair: After further reflection on how these files fit together, it seems some kind of internal setting could be a more appropriate way to configure this? (i.e. [settings_internal.js](https://github.com/emscripten-core/emscripten/blob/main/src/settings_internal.js))

src/library_pthread.js
```diff
@@ -426,14 +407,14 @@ var LibraryPThread = {
             createScriptURL: (ignored) => new URL("{{{ TARGET_JS_NAME }}}", import.meta.url)
           }
         );
-        worker = new Worker(p.createScriptURL('ignored'), workerOptions);
+        worker = new Worker(p.createScriptURL('ignored'), {{{ WORKER_OPTIONS }}});
```
sbc100, line 410: I simpler way to do this would be to define a new macro here at the top of this file. e.g.

```
{{{
globalThis.WORKER_OPTIONS = {
   ...
};
null;
}}}
```

jamsinclair: Thank you so much! I wasn't too sure how the templating works in these files 😅 . I'll update the PR with this approach when I get a chance.

tools/link.py
jamsinclair: Solved thanks to the maintainers suggested improvement.

src/library_pthread.js
```diff
@@ -31,6 +31,28 @@ globalThis.MAX_PTR = (2 ** 32) - 1
...
+#if ENVIRONMENT_MAY_BE_NODE
+  // This is the way that we signal to the node worker that it is hosting
+  // a pthread.
+  'workerData': 'em-pthread',
+#endif
+#if ENVIRONMENT_MAY_BE_WEB || ENVIRONMENT_MAY_BE_WORKER
+  // This is the way that we signal to the Web Worker that it is hosting
+  // a pthread.
+#if ASSERTIONS
+  'name': 'em-pthread-' + PThread.nextWorkerID,
```
jamsinclair, line 48: Alas, the worker options are not entirely static.

Any elegant ideas of how we could solve this? 🤔

jamsinclair: Vite will let you compile non-static worker options if you include the comment `/* @vite-ignore */` before the worker options definition. It will always treat the worker as a classic worker type if the ignore comment is used.

Perhaps, this would be ok to configure as we only need dynamic name values for debugging?

jamsinclair:
I'm still tinkering around. The learnings so far are
- Emscripten sets a dynamic worker `name` value when `ASSERTIONS` are configured. This complicates how we can define the static worker options
- To keep the code backwards compatible with the above, it looks like we'll need to define worker options per use of `new Worker(...)`
- To also support the dynamic worker `name` values we'll need to output a `/* @vite-ignore */` comment before each worker options definition

Please let me know if anyone has any other suggestions or cleaner ways we could solve this!

src/library_pthread.js
jamsinclair: Changes to Vite are likely to be made that would allow dynamic name values to be used (https://github.com/vitejs/vite/pull/19010)

jamsinclair:
Superseded by #23607

PR #22838, opened 2024-11-03
[docs] Added a tip to explain how to permanently add environment variables

### LINKED ISSUE #20471: Bug in page:Download and install
Suggestion for making install and run easier so newbies can get started more easily. Instructions jump straight to checking the version.

On installation instructions please include setting PATH by adding 

`export PATH="/path/to/emsdk/upstream/emscripten:$PATH"`

to

`/home/username/.bashrc`

URL: ../../docs/getting_started/downloads

lorsanta:
Fixes: #20471

### DIFF
--- a/site/source/docs/getting_started/downloads.rst
+++ b/site/source/docs/getting_started/downloads.rst
@@ -52,7 +52,7 @@ GitHub and set them as :term:`active <Active Tool/SDK>`:
     # Activate PATH and other environment variables in the current terminal
     source ./emsdk_env.sh
 
-  .. tip:: If you want to avoid to execute ``source /path/to/emsdk/emsdk_env.sh`` everytime you open a new terminal, run ``echo source /path/to/emsdk/emsdk_env.sh >> ~/.bashrc``.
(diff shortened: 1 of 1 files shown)


kripken:
Thanks @lorsanta for the PR. Personally, I am not sure this level of detail is worth getting into, however: most users probably know how to use their shell to do this, or know how to search to find out how (I am in the second camp... I always need to look it up, but it just takes a minute). Curious what other people think.

lorsanta:
> Thanks @lorsanta for the PR. Personally, I am not sure this level of detail is worth getting into, however: most users probably know how to use their shell to do this, or know how to search to find out how (I am in the second camp... I always need to look it up, but it just takes a minute). Curious what other people think.

Hi! You're right I agree, I made the pull request just to close the issue that it's a little old. I would be fine with closing it as "won't fix."

> I'd be more comfortable with a general link to a place that explains how to do it on various shells and OSes, for example. Though I'm not sure if there is one?

I'm not aware of one,  but I personally prefer @sbc100 approach, so we can avoid any link rot issues in the future

sbc100 (approved):

Issue #22837, opened 2024-11-03
UTF8ToString does not work with -sWASM_WORKERS if window.crossOriginIsolated is false


ThomasMertes:
I compile with `-sWASM_WORKERS` but the code with `UTF8ToString` should also work if  `crossOriginIsolated` is false.

I open the worker only if `window.crossOriginIsolated` holds. In case of a missing worker I have an alternate solution without worker.
The problem is: I use `UTF8ToString`. E.g.:

```
        windowIdAndFlags = EM_ASM_INT({
          if (typeof window !== "undefined") {
            let windowName = Module.UTF8ToString($4);
            let windowTitle = Module.UTF8ToString($5);
```

and this fails for the combination `-sWASM_WORKERS` and `crossOriginIsolated===false` with:

```
Uncaught (in promise) ReferenceError: SharedArrayBuffer is not defined
    UTF8ArrayToString                    http://localhost:1080/demo/klondike.js:2179
    UTF8ToString                         http://localhost:1080/demo/klondike.js:4882
    6879322                              http://localhost:1080/demo/klondike.js:1103
    runEmAsmFunction                     http://localhost:1080/demo/klondike.js:5183
    _emscripten_asm_const_int            http://localhost:1080/demo/klondike.js:5186
    x                                    http://localhost:1080/demo/klondike.js:5570
    dynCall_ijjjji                       http://localhost:1080/demo/klondike.js:5991
    invoke_ijjjji                        http://localhost:1080/demo/klondike.js:6117
    x                                    http://localhost:1080/demo/klondike.js:5570
    _main                                http://localhost:1080/demo/klondike.js:5917
    callMain                             http://localhost:1080/demo/klondike.js:6156
    doRun                                http://localhost:1080/demo/klondike.js:6190
    run                                  http://localhost:1080/demo/klondike.js:6200
    runCaller                            http://localhost:1080/demo/klondike.js:6139
    removeRunDependency                  http://localhost:1080/demo/klondike.js:532
    receiveInstance                      http://localhost:1080/demo/klondike.js:671
    receiveInstantiationResult           http://localhost:1080/demo/klondike.js:680
    promise callback*instantiateAsync/<  http://localhost:1080/demo/klondike.js:637
    promise callback*instantiateAsync    http://localhost:1080/demo/klondike.js:631
    createWasm                           http://localhost:1080/demo/klondike.js:697
    <anonymous>                          http://localhost:1080/demo/klondike.js:5913
```

[...]

PR #22844, opened 2024-11-04
Avoid reference to SharedArrayBuffer in UTF8ArrayToString

sbc100:
This allows program built with `SHARED_MEMORY` enabled to run in a single threaded context where `SharedArrayBuffer` is not defined (i.e. when crossOriginIsolated is false).

Fixes: #22837

### DIFF
--- a/src/parseTools.mjs
+++ b/src/parseTools.mjs
@@ -1042,9 +1042,11 @@ function getUnsharedTextDecoderView(heap, start, end) {
   // then unconditionally do a .slice() for smallest code size.
   if (SHRINK_LEVEL == 2 || heap == 'HEAPU8') return shared;
 
-  // Otherwise, generate a runtime type check: must do a .slice() if looking at a SAB,
-  // or can use .subarray() otherwise.
-  return `${heap}.buffer instanceof SharedArrayBuffer ? ${shared} : ${unshared}`;
+  // Otherwise, generate a runtime type check: must do a .slice() if looking at
+  // a SAB, or can use .subarray() otherwise.  Note: We compare with
+  // `ArrayBuffer` here to avoid referencing `SharedArrayBuffer` which could be
+  // undefined.
+  return `${heap}.buffer instanceof ArrayBuffer ? ${unshared} : ${shared}`;
 }
 
 function getEntryFunction() {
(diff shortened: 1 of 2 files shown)


src/parseTools.mjs
```diff
@@ -1042,9 +1042,11 @@ function getUnsharedTextDecoderView(heap, start, end) {
...
   if (SHRINK_LEVEL == 2 || heap == 'HEAPU8') return shared;
 
-  // Otherwise, generate a runtime type check: must do a .slice() if looking at a SAB,
-  // or can use .subarray() otherwise.
-  return `${heap}.buffer instanceof SharedArrayBuffer ? ${shared} : ${unshared}`;
+  // Otherwise, generate a runtime type check: must do a .slice() if looking at
+  // a SAB, or can use .subarray() otherwise.  Note: We compare with
+  // `ArrayBuffer` here to avoid referencing `SharedArrayBuffer` which could be
+  // undefined.
+  return `${heap}.buffer instanceof ArrayBuffer ? ${unshared} : ${shared}`;
```
kripken, line 1049: Hmm, I'm surprised this works, since I'd expect `SharedArrayBuffer` to inherit from `ArrayBuffer`, so `x instanceof SharedArrayBuffer` implies `x instanceof ArrayBuffer`. But I verified that it isn't the case, so I guess these are just separate types...

test/test_other.py
```diff
@@ -12193,14 +12193,25 @@ def test_pthread_relocatable(self):
...
+    self.emcc_args += ['-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME', '-sDEFAULT_LIBRARY_FUNCS_TO_INCLUDE=$stringToNewUTF8,$UTF8ToString']
     self.do_run_in_out_file_test('hello_world.c')
 
     # Now run the same program but with SharedArrayBuffer undefined, it should run
     # fine and then fail on the first call to pthread_create.
-    create_file('pre.js', 'SharedArrayBuffer = undefined\n')
-    expected = 'pthread_create: environment does not support SharedArrayBuffer, pthreads are not available'
-    self.do_runf('hello_world.c', expected, assert_returncode=NON_ZERO, emcc_args=['--pre-js=pre.js'])
+    #
+    # We specifically test that we can call UTF8ToString which previously contained
```
kripken, line 12202:
```suggestion
    # We specifically test that we can call UTF8ToString which in earlier Emscripten versions contained
```

kripken (approved):

PR #22846, opened 2024-11-04
Extract JSEvents.memcpy helper. NFC

sbc100:
I notices that were was a bug here in `emscripten_get_devicemotion_status` and `emscripten_get_deviceorientation_status` where `HEAP32` was used instead of `HEAP8`.

### DIFF
--- a/src/library_html5.js
+++ b/src/library_html5.js
@@ -58,6 +58,9 @@ var LibraryHTML5 = {
     currentEventHandler: null,
 #endif
 */
+    memcpy(target, src, size) {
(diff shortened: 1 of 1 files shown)


kripken (approved):

sbc100:
I notices that were was a bug here in emscripten_get_devicemotion_status and emscripten_get_deviceorientation_status where HEAP32 was used instead of HEAP8.   I guess we don't have tests for those!

kripken:
Hmm, yeah... we might not have tests for these, they might need input faking. We have limited tests for those, sadly.

PR #22849, opened 2024-11-05
[library_html5] Simplify JSEvent system. NFC

sbc100:
This change extracts a few helper function to make the JS event system smaller and more internally consistent.

Firstly, we now use a single storage location for the event data pointer used in event callback.  This is allocated on demand in `allocateEventStruct`.  We then use a new `dispatchEvent` all the many loction that previously duplicated the event dispatching logic.

In order to handler the 3 cases where we need to keep a copy of the last dispatched event we also have two new helpers: `getCachedEvent`, `cacheEvent`.  These allocate separate copies of the last generated event, but only for the 3 event types where this is required.

### DIFF
--- a/src/library_html5.js
+++ b/src/library_html5.js
@@ -4,30 +4,64 @@
  * SPDX-License-Identifier: MIT
  */
 
+{{{
+globalThis.maxEventSize = Math.max(
+  C_STRUCTS.EmscriptenKeyboardEvent.__size__,
+  C_STRUCTS.EmscriptenMouseEvent.__size__,
+  C_STRUCTS.EmscriptenWheelEvent.__size__,
+  C_STRUCTS.EmscriptenUiEvent.__size__,
+  C_STRUCTS.EmscriptenFocusEvent.__size__,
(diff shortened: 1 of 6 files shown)


sbc100:
Sadly I don't think we have code size tests that cover library_html5 but I think this would be pretty good code size saving.

sbc100:
For test_html5_mouse_closure we have:

Before:
13897 out/test/test.js
19922 out/test/test.wasm

After:
14189 out/test/test.js
20141 out/test/test.wasm

kripken:
This looks like a rather large refactoring, and I am worried about our lack of testing on this UI code. Are you certain we have coverage of all the things changed here? If not, perhaps before this PR we can look into adding tests?

sbc100:
> This looks like a rather large refactoring, and I am worried about our lack of testing on this UI code. Are you certain we have coverage of all the things changed here? If not, perhaps before this PR we can look into adding tests?

Agreed, I'll put this on ice until I get around to adding more testing.

PR #22850, opened 2024-11-05
Allow avxintrin.h to be used by C compiler

jkl1337:
An internal union defined to implement 256-bit AVX support added in PR emscripten-core/emscripten#22430 is missing `union` keywords at declarations so causes errors when used with the C compiler.

Add the `union` keyword to the declarations of the `m256_data` union in avxintrin.h. Also adds `__` prefix to make type `__m256_data` to avoid further polluting global namespace.

### DIFF
--- a/system/include/compat/avxintrin.h
+++ b/system/include/compat/avxintrin.h
@@ -40,7 +40,7 @@ typedef struct {
   __m128i_u v1;
 } __m256i_u;
 
-union m256_data {
+union __m256_data {
   __m256i int_view;
   __m256d double_view;
   __m256 float_view;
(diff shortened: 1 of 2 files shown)


sbc100 (approved):
lgtm.. but I wonder why this wasn't covered by our `test_standalone_system_headers` test in the `test_other.py`

sbc100:
I just uploaded https://github.com/emscripten-core/emscripten/pull/22854.  Once that land you should be able to remove the exceptions for these headers from `test_standalone_system_headers` as part of this change.

sbc100:
Can you remove these headers from the exception list in `test_standalone_system_headers`?

PR #22855, opened 2024-11-05
Move `preloadedAudios`/`preloadedImages` inside `Browser`. NFC

sbc100:


src/library_browser.js
```diff
@@ -881,7 +882,7 @@ var LibraryBrowser = {
   $getPreloadedImageData: (path, w, h) => {
     path = PATH_FS.resolve(path);
 
-    var canvas = /** @type {HTMLCanvasElement} */(preloadedImages[path]);
+    var canvas = /** @type {HTMLCanvasElement} */(Browser.preloadedImages[path]);
```
brendandahl, line 885: Missing dependency on $Browser?

src/library_sdl.js
```diff
@@ -2209,7 +2209,7 @@ var LibrarySDL = {
 
       if (!raw) {
         filename = PATH_FS.resolve(filename);
-        raw = preloadedImages[filename];
+        raw = Browser.preloadedImages[filename];
```
brendandahl, line 2212: Missing dependency on $Browser? (and the other spots in this file)

src/library_browser.js
sbc100: All symbols in this file implicitly depend on `$Browser` thanks to the `autoAddDeps(LibraryBrowser, '$Browser');` line at the bottom.

src/library_sdl.js
sbc100: This is also handled transitively I think sine we have `autoAddDeps(LibrarySDL, '$SDL');` below and SDL depends on `Browser.`   Maybe I should make this one explicit though.

brendandahl (approved):

Issue #22852, opened 2024-11-05
setjmp/longjmp and sigsetjmp/siglongjmp hang if -sWASM_WORKERS is used


ThomasMertes:
The program `longjmp_error.c` tests the functionality of setjmp/longjmp:
```
#include <stdio.h>
#include <setjmp.h>
int main(int argc, char *argv[]){
  jmp_buf env;
  int ret_code = 4; 
  int count = 2;
  if ((ret_code=setjmp(env)) == 0) {
    count--; 
    longjmp(env, count);
    printf("3\n");
  } else printf("%d\n", ret_code);
  return 0;
}
```
When I compile this program with:
```
emcc longjmp_error.c -o longjmp_error.js
```
I can execute it and it writes `1`:
```
tm@localhost:~/seed7_5/src> node longjmp_error.js 
1
```
If I compile the program with:
```
emcc -sWASM_WORKERS longjmp_error.c -o longjmp_error.js
```
The program hangs when I execute it:
```
node longjmp_error.js
```
It is necessary to terminate the program with ^C.
This is not how setjmp/longjmp should work.

The same holds for setjmp/longjmp which tests sigsetjmp/siglongjmp:
```
#include <stdio.h>
#include <setjmp.h>
int main(int argc, char *argv[]){
  sigjmp_buf env; 
  int ret_code = 4;
  int count = 2;
  if ((ret_code=sigsetjmp(env, 1)) == 0) {
    count--; 
    siglongjmp(env, count); 
    printf("3\n");
  } else printf("%d\n", ret_code);
  return 0;
}
```
**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.69 (c55ff404dbe1eb9f263c48b584d5ba627224763a)
clang version 20.0.0git (https:/github.com/llvm/llvm-project 50866e84d1da8462aeb96607bf6d9e5bbd5869c5)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /home/tm/emsdk/upstream/bin
```
**Full link command and output with `-v` appended:**
```
emcc -sWASM_WORKERS longjmp_error.c -o longjmp_error.js -v
[...]

PR #22856, opened 2024-11-05
[WASM_WORKERS] Ensure TLS is initialized

sbc100:
The main thread TLS region get initialized via a static constructor in `library_wasm_worker.c`.  Without this change programs that don't otherwise reference `library_wasm_worker.c` will not have this static constructor run.

Fixes: #22852

### DIFF
--- a/tools/system_libs.py
+++ b/tools/system_libs.py
@@ -2276,7 +2276,7 @@ def get_libs_to_link(args):
   if force_include:
     logger.debug(f'forcing stdlibs: {force_include}')
 
-  def add_library(libname):
+  def add_library(libname, whole_archive=False):
     lib = system_libs_map[libname]
     if lib.name in already_included:
       return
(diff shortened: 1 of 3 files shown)


tools/system_libs.py
```diff
@@ -2396,7 +2396,7 @@ def add_sanitizer_libs():
   if settings.WASM_WORKERS and (not settings.SINGLE_FILE and
                                 not settings.RELOCATABLE and
                                 not settings.PROXY_TO_WORKER):
-    add_library('libwasm_workers')
+    add_library('libwasm_workers', whole_archive=True)
```
kripken, line 2399: Please add a comment here.

kripken (commented):
I guess there is no size downside here since that library has a single C file, and is almost always referred to anyhow?

sbc100:
> I guess there is no size downside here since that library has a single C file, and is almost always referred to anyhow?

yes, exactly.  Passing `-sWASM_WORKERS` but not references that file should be a outlier case.

kripken (approved):

PR #22864, opened 2024-11-06
[wasm64] Remove the experimental warning from -sMEMORY64

sbc100:
Memory64 is now at phase4 and IIUC all emscripten features now work with it.

### DIFF
--- a/emcc.py
+++ b/emcc.py
@@ -899,10 +899,8 @@ def phase_setup(options, state, newargs):
   if options.target.startswith('wasm64'):
     default_setting('MEMORY64', 1)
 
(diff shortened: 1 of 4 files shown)


kripken (approved):
fyi I am looking into adding fuzzing for memory64 on binaryen right now, but this doesn't need to wait on that.

sbc100:
> fyi I am looking into adding fuzzing for memory64 on binaryen right now, but this doesn't need to wait on that.

Yes, this isn't really about the runtimes being finished so much as the emscripten itself being basically feature complete.

Issue #11792, opened 2020-08-03
EXPORT_ES6: Add testing and support for node.js (replace require(), __dirname)


curiousdannii:
So I'm finally getting around to trying EXPORT_ES6 mode, and it just doesn't work in node (when loaded as an actual ES6 Module), because it still refers to `__dirname`. I assume other people have only been using it Webpack/rollup/whatever?

It doesn't look like there are any tests for EXPORT_ES6 either.

kripken:
I think we don't have tests for it because at the time there wasn't a version of node that supported it. If there is now, we should add some.

I think we have 12.10 on CI, and can maybe add another version. Another option is to use v8, which we already install a new version of, if it has ES6 module support.

It would be good if someone could write up a test for this and then we can check the options.

curiousdannii:
Node 12.10 will definitely be enough.

sbc100:
Just FYI, it looks like out current version of node that we use in emsdk and in all our testing is `12.18.1`.

jvail:
Could this issue be extended to cover also "require" vs "import"? I wonder if replacing all lines in the output such as 

```js
var nodeFS = require('fs');`
```

with 

```js
#if EXPORT_ES6
const { default: nodeFS } = await import('fs');
#else
var nodeFS = require('fs');
```
would be sufficient.

Unfortunately my insight in the build system is very limited. But if someone could start then I could help at least with testing.

curiousdannii:
Realised this is going to be difficult to handle for multi-environment builds. ES import statements can't be put in conditionals, and while the import function can, it needs to be handled async.

RReverser:
> ES import statements can't be put in conditionals, and while the import function can, it needs to be handled async.

They don't need to, Emscripten could emit static import declarations instead.

curiousdannii:
I said it would be difficult for multi-environment builds. A static import declaration for node will fail in the browser.

RReverser:
Ah, right, I see. In that case, the async import sounds fine. The fact that it's async shouldn't be an issue - Emscripten already has a set of async runtime dependencies and a way to wait on them.

Take a look at usages of `addRunDependency` from https://github.com/emscripten-core/emscripten/blob/0cc4a667828be0793db01c8c6b85434b6ccb38dc/src/preamble.js#L488.

matthewp:
@RReverser You mentioned working on this in #13571 , do you have a test case that you were using?

RReverser:
[...]

PR #22867, opened 2024-11-06
Add an experimental way to export a instance ES module.

brendandahl:
Adds a new mode `-sMODULARIZE=instance` which will change the output to be more of a static ES module. As mentioned in the docs, there will be a default `async init` function exported and named exports that correspond to Wasm and runtime exports. See the docs and test for an example.

Internally, the module will now have an `init` function that wraps nearly all of the code except some top level variables that will be exported. When the `init` function is run, the top level variables are then updated which will in turn update the module exports. E.g.

```js
async function init(moduleArgs) {
  function foo() {};
  x_foo = foo;
  x_bar = wasmExports['bar'];
}
var x_foo, x_bar;
export {x_foo as foo, x_bar as bar};
```

Note: I alternatively tried to keep everything at the top level scope and move only the code that reads from moduleArg into an init function. This would make it possible to get rid of the `x_func` vars and directly add `export` to vars/functions we want to export. However, there are lots of things that read from moduleArg in many different spots and ways which makes this challenging.


src/modules.mjs
```diff
@@ -384,6 +384,9 @@ function exportRuntime() {
     // If requested to be exported, export it.  HEAP objects are exported
     // separately in updateMemoryViews
     if (EXPORTED_RUNTIME_METHODS.has(name) && !name.startsWith('HEAP')) {
+      if (MODULARIZE === 'static') {
+        return `x_${name} = ${name};`;
```
sbc100, line 388: Can we use a more meaningful prefix?   How about `__exp_` ?

src/postamble_modularize.js
```diff
@@ -18,7 +18,7 @@ moduleRtn = Module;
 
 #endif // WASM_ASYNC_COMPILATION
 
-#if ASSERTIONS
+#if ASSERTIONS && MODULARIZE != 'static'
```
sbc100, line 21: Why not use `-sMODULARIZE=instance` to match the old `-sMODULARIZE_INSTANCE` setting?  Do you think the word "static" is more descriptive of what is happening here and "instance"?  I think of the two modes as "single instance" vs "factory/multiple instance", but maybe "static" makes sense on some level?

tools/link.py
```diff
@@ -757,7 +757,13 @@ def phase_linker_setup(options, state, newargs):
 
   if options.oformat == OFormat.MJS:
     settings.EXPORT_ES6 = 1
-    settings.MODULARIZE = 1
+    if not settings.MODULARIZE:
```
sbc100, line 760: I think you can use `default_setting` here

tools/settings.py
```diff
@@ -267,7 +267,7 @@ def __setattr__(self, name, value):
     self.attrs[name] = value
 
   def check_type(self, name, value):
-    if name in ('SUPPORT_LONGJMP', 'PTHREAD_POOL_SIZE', 'SEPARATE_DWARF', 'LTO'):
+    if name in ('SUPPORT_LONGJMP', 'PTHREAD_POOL_SIZE', 'SEPARATE_DWARF', 'LTO', 'MODULARIZE'):
```
sbc100, line 270: Can we add a comment above this line?  Something like "# These settings have a variable type so cannot be easily type checked"

tools/link.py
```diff
@@ -2448,9 +2479,18 @@ def modularize():
...
     src += 'export default %s;\n' % settings.EXPORT_NAME
 
+  if settings.MODULARIZE == 'static':
+    exports = settings.EXPORTED_FUNCTIONS + settings.EXPORTED_RUNTIME_METHODS
+    # Declare a top level var for each export so that code in the init function
+    # can assign to it and update the live module bindings.
+    src += "var " + ", ".join(['x_' + export for export in exports]) + ";\n"
+    # Export the functions with their original name.
+    exports = ['x_' + export + ' as ' + export for export in exports]
+    src += "export {" + ", ".join(exports) + "};\n"
```
sbc100, line 2492: Use single quote in this block

src/postamble_modularize.js
brendandahl: I open to changing, but instance to me confusing. When I hear instance, it makes me think I can create multiple instances i.e. a factory. I chose static since the exports to the outside world don't change and there's only one static instance.

Other ideas: static_instance, singleton, esmodule. Or alternatively, we add a whole different flag that will enable all the settings we want in the "new output world" e.g. strict, es modules, no _ exports....

sbc100: I'm kind of leaning towards bring back the old name `-sMODULARIZE_INSTANCE` until we can think of a better one.    It is exactly the same meaning as the old option I believe, right?

Its interesting to me that the word `instance` to you implies multiple instances.  I don't think it has the connotation for me.

kripken:
> Internally, the module will now have an init function that wraps nearly all of the code except some top level variables that will be exported. When the init function is run, the top level variables are then updated which will in turn update the module exports.

I think I follow this, but can you maybe explain a bit about how this improves on the current status? That is, what is changed, and what problems this fixes by changing it?

brendandahl:
> > Internally, the module will now have an init function that wraps nearly all of the code except some top level variables that will be exported. When the init function is run, the top level variables are then updated which will in turn update the module exports.
> 
> I think I follow this, but can you maybe explain a bit about how this improves on the current status? That is, what is changed, and what problems this fixes by changing it?

Do you mean the overall goals of this new ES module output API or why the `init` function is needed?

brendandahl:
To answer both:

The goals of a static ES module:
1) Produce a module that has a more familiar shape to JS developers where the exports are statically declared.
2) Make it easier for bundlers to do tree shaking/dead code elimination on ES modules produced by emscripten. Currently, bundlers like rollup can't handle the modularize output where everything is dynamically assigned to the Module object. The use case for this is that a developer uses a prebuilt emscripten module, but only uses a small part of the library. Ideally, the developer in this case would still be able to do DCE after the module has been produced.

As for the init function, there's no way to pass arguments into an imported ES module, so we have to separate out this step into an init function.

sbc100:
> I think I see, thanks. And, looking at the current output of `-sMODULARIZE -sEXPORT_ES6`, it looks like we have
> 
> ```js
> export default Module;
> ```
> 
> So "more static" here means to replace the big Module object with individual exports per function, as written above,
> 
> ```js
> export {x_foo as foo, x_bar as bar};
> ```
> 
> ?
> 
> What I am still a little confused on is why that is more "static". I guess the big Module object is not introspectable, but `x_foo` is written dynamically in the example above - can bundlers really reason about the runtime assignment to it?

I believe that can/will yes.  You are correct that these still get dynamically assigned internally, but at least static from the POV of the importer/user they are static.  Over time we can work to make these even more static, using things like the esm integration proposal, but even before then IIUC the bundlers can deal with this kind of thing.

kripken:
Interesting. How do they do that, though? It seems that to do any kind of useful optimization here we need to know exactly what can be assigned to these exports. I guess it doesn't matter that they can be assigned more than once, but can they even infer the content written to them? That is, @brendandahl you gave this example now
```js
 _export__foo = function() { console.log('foo'); };
```
but the actual code would be less static, I think? Don't we capture these from wasm exports, like this?
```js
 _export__foo = exports.foo;
```
Do these tools actually manage to figure out what `exports` is, going back through the wasm compile&link stuff, to the wasm binary file?

sbc100:
> Do these tools actually manage to figure out what `exports` is, going back through the wasm compile&link stuff, to the wasm binary file?

I think the idea is that we could extend the bundlers so that they can do that (assuming that can't already do it today).

brendandahl:
> Don't we capture these from wasm exports, like this?

Yes, for the first stage rollup would only be able to get rid of runtime/library JS that is unused. The next stage is to create a rollup plugin that knows about wasm exports and can remove them if they're never used. This is where it gets more tricky and it's unclear exactly how we want to do that. I see two options:
1) Use source phase imports to identify wasm exports in the plugin.
2) Create some special emscripten output that makes it easy for a plugin to identify wasm exports.

Option 1 seems more like more general/standards approach, but I haven't looked into this yet.

Alternatively, we could keep the current module output and try to make bundlers understand that. I don't see this as a great option though, since we'd need to support this several bundlers, whereas using `export` seems like a way tools already support.

tools/link.py
```diff
@@ -2394,7 +2400,20 @@ def modularize():
...
+
+%(src)s
+
+  return await moduleRtn;
+}
+''' % {
+      'src': src,
+    }
+  else:
+    src = '''
```
kripken, line 2416: Do we show an error if MODULARIZE is written anything other than `static` or `1`?

kripken:
I see, thanks! Makes sense. So there are few more steps here, after this, that I didn't fully get before.

curiousdannii:
In the world of ES Modules, do wrapper functions really have any actual advantage? Wouldn't it be most beneficial to have no wrapper? Everything applicable could be exported directly, and DCE-ed by the bundler if not used.

We'd then need a separate setting for factory output.

Old discussion on the same topic: https://github.com/emscripten-core/emscripten/issues/11792#issuecomment-921312864

sbc100:
> In the world of ES Modules, do wrapper functions really have any actual advantage? 

When we can avoid wrappers we obviously can and should.  However, until the Wasm ESM integration proposal (https://github.com/WebAssembly/esm-integration) lands we cannot directly export wasm functions, so we need to export something than is then dynamically updated to point to the wasm function once its available.

The second issue that emscripten startup takes arguments (see `INCOMING_MODULE_JS_API`).   This means that even if we have full ESM integration we would still need to delay module instantiation in most cases until those arguments are provided.

brendandahl:
> Old discussion on the same topic: [#11792 (comment)](https://github.com/emscripten-core/emscripten/issues/11792#issuecomment-921312864)

Thanks for pointing to this issue. I hadn't seen it before, but I think it echos a lot of what I was thinking. I like your idea of two independent options of `FACTORY(aka MODULARIZE)` and then `MODULE_FORMAT` = `none/esm/umd/...`. Ideally, in emscripten we'd just handle creating the factory/instance version, and always produce an ESM. Then we'd let rollup handle outputting the desired module format.

tools/link.py
brendandahl: Added a check in phase_linker_setup.

src/postamble_modularize.js
brendandahl: I've renamed it instance for now. Later I'd like to look into the idea of having two independent options of `FACTORY(aka MODULARIZE)` and then `MODULE_FORMAT` = `none/esm/umd/...` as mentioned below and in the other bug.

brendandahl:
For posterity, it seems webpack is unable to do tree shaking as thoroughly as rollup and closure. Using the example of "exported functions" in [this comment](https://github.com/emscripten-core/emscripten/pull/22867#issuecomment-2462955856), the `bar` function still remains in the output. 

I tried with:
```js
// package.json
{
  "sideEffects": false
}

// webpack.config.json
const path = require('path');

module.exports = {
  entry: "./main.mjs",
  output: {
    path: path.resolve(__dirname, "dist"),
  },
  optimization: {
    usedExports: true
  },
  stats: {
    // Display bailout reasons
    optimizationBailout: true,
  },
};
```

If I instead use an even "more static" module, tree shaking does work.

```js
export let fillSquare = function() {
  console.log('fillSquare');
};
export let fillCircle = function() {
  console.log('fillCircle');
};

export default async function init() {
}
```

src/postamble_modularize.js
```diff
@@ -18,7 +18,7 @@ moduleRtn = Module;
 
 #endif // WASM_ASYNC_COMPILATION
 
-#if ASSERTIONS
+#if ASSERTIONS && MODULARIZE != 'instance'
```
sbc100, line 21: Why this change?

test/modularize_instance_runner.mjs
```diff
@@ -0,0 +1,3 @@
+import init, { _foo as foo } from "./modularize_static.mjs";
+await init();
+foo();
```
sbc100, line 3: Missing newline

test/modularize_instance.cpp
```diff
@@ -0,0 +1,24 @@
...
+extern "C" void bar() {
+  printf("bar\n");
+}
+
+int main() {
+  printf("main1\n");
+#ifdef __EMSCRIPTEN_PTHREADS__
+  std::thread([]{
+    printf("main2\n");
+  }).join();
```
sbc100, line 20: Can you using pthread_create here and keep this test as a single `C` test?   layering on C++ threading makes the test more complex than it needs to be.

Alternatively if you just want to test if threading works at all the easiest thing to do with run with `-sPROXY_TO_PTHREAD` .. then you don't need to write any of your own threading code.

tools/link.py
```diff
@@ -2403,9 +2424,9 @@ def modularize():
   return moduleRtn;
 }
 ''' % {
-    'maybe_async': async_emit,
-    'src': src,
-  }
+      'maybe_async': async_emit,
+      'src': src,
+    }
```
sbc100, line 2429: Is this indentation better than before?  Maybe revert this?

tools/link.py
```diff
@@ -2448,9 +2481,18 @@ def modularize():
     src += f'globalThis.AudioWorkletModule = {settings.EXPORT_NAME};\n'
 
   # Export using a UMD style export, or ES6 exports if selected
-  if settings.EXPORT_ES6:
+  if settings.EXPORT_ES6 and settings.MODULARIZE != 'instance':
     src += 'export default %s;\n' % settings.EXPORT_NAME
 
+  if settings.MODULARIZE == 'instance':
```
sbc100, line 2487: Maybe write these two blocks as:

```
if EXPORT_ES6:
      .. new stuff
    else:
       .. old stuff
```

kripken (approved):
lgtm % @sbc100 's comments

src/postamble_modularize.js
brendandahl: Not needed, this was from before the init function wrapped everything.

tools/link.py
brendandahl: Flake wants this because of the new if/else above.

sbc100 (approved):
I'm still not crazy about reusing the `MODULARIZE` setting but I can't think of a clearly better option, and we can continue to bikeshed this while the feature is experimental.

I also don't love the fact that we cannot put the generated code at the top level of the module, but we can work on fixing that over time.

lgtm with one final comment.

brendandahl:
> I'm still not crazy about reusing the `MODULARIZE` setting but I can't think of a clearly better option, and we can continue to bikeshed this while the feature is experimental.
> 

I'm also not crazy about the current. I was hoping to discuss this during the bundler meeting today, but didn't see you there. My current thinking is trying to progress to what was suggested in another issue. Have two independent options:

1) FACTORY(aka MODULARIZE) This strictly controls if we output a factory or instance style.
2) MODULE_FORMAT = none/esm/umd/etc.

sbc100:
> > I'm still not crazy about reusing the `MODULARIZE` setting but I can't think of a clearly better option, and we can continue to bikeshed this while the feature is experimental.
> 
> I'm also not crazy about the current. I was hoping to discuss this during the bundler meeting today, but didn't see you there. 

Oops, sorry about that.  Forgot to Yes the event.

> My current thinking is trying to progress to what was suggested in another issue. Have two independent options:
> 
> 1. FACTORY(aka MODULARIZE) This strictly controls if we output a factory or instance style.
> 2. MODULE_FORMAT = none/esm/umd/etc.

Yes.  I'm also interested in hiding internals inside and IIFE.  Perhaps that could be `-sMODULE_FORMAT=IIFE` :)    I'd love to also make that the default in strict mode since it should be mostly unobservable.

Issue #15365, opened 2021-10-26
Update website theme for newer Sphinx


kripken:
See https://github.com/emscripten-core/emscripten/pull/15361#issuecomment-952331375, the current theme appears to be tailored for older versions. Before we upgrade Sphinx we'll need to fix the theme.

jonassmedegaard:
Possibly irrelevant if starting over from scratch, but Debian maintainer for Sphinx provided some insight into modernizing at https://bugs.debian.org/1000136

sbc100:
This is awesome progess!  Thanks @lorsanta!

sbc100:
And thanks @jonassmedegaard for making it possible!

lorsanta:
No problem @sbc100 , happy to help!

sbc100:
I noticed that that anchor symbols are missing. e.g. https://emscripten.org/docs/contributing/developers_guide.html#bisecting

There used to be a chainlink symbol after each heading I think.

lorsanta:
The FontAwesome related files are missing from `_static\css`. It's strange, though, because they are present when you build the site locally.

kripken:
Thanks @lorsanta , that was it! I didn't notice I needed to `git add` those new files. It looks fixed to me now.

PR #22872, opened 2024-11-06
Update website theme for newer Sphinx #15365

lorsanta:
Ref #15365

I replaced the custom theme with the last version of sphinx_rtd_theme, and then added the changes found by diffing the emscripten custom theme with sphinx_rtd_theme version 0.1.6


site/source/conf.py
```diff
@@ -392,7 +392,8 @@
 
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'http://docs.python.org/': None}
+intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}
+
```
sbc100, line 396: Maybe just delete these lines?  I don't think we need to link to the python standard library anywhere?

site/source/docs/api_reference/html5.h.rst
```diff
@@ -106,7 +106,7 @@ When the event occurs the callback is invoked with the relevant event "type" (fo
 
 .. _callback-handler-return-em_bool-html5-api:
 
-Callback handlers that return an :c:data:`bool` may specify ``true`` to signal that the handler *consumed* the event (this suppresses the default action for that event by calling its ``.preventDefault();`` member). Returning ``false`` indicates that the event was not consumed — the default browser event action is carried out and the event is allowed to pass on/bubble up as normal.
+Callback handlers that return a ``bool`` may specify ``true`` to signal that the handler *consumed* the event (this suppresses the default action for that event by calling its ``.preventDefault();`` member). Returning ``false`` indicates that the event was not consumed — the default browser event action is carried out and the event is allowed to pass on/bubble up as normal.
```
sbc100, line 109: What happened to `:c:data:`?

site/source/docs/index.rst
```diff
@@ -51,7 +51,6 @@ The full hierarchy of articles, opened to the second level, is shown below.
   compiling/index
   building_from_source/index
   contributing/index
-  api_reference/index
```
sbc100, line 54: What happened to api_reference?

sbc100 (commented):
Wow! Thanks for working on this.  This is awesome.   I wonder if we could publish the output somewhere temporary so that is might be viewed before we land this and so we can compare the old and new output?

sbc100:
You will likely need to update all the sphinx-related stuff in `requirements-dev.txt`.

site/source/docs/index.rst
lorsanta: When building the site I get multiple warnings like ```/emscripten/site/source/docs/api_reference/index.rst: document is referenced in multiple toctrees: ['docs/index', 'index'], selecting: index <- docs/api_reference/index```.

I thought about deal with them by removing the duplicates but then I changed my mind to make the pull request easier to understand, but I forgot to add ```api_reference/index``` back.

site/source/docs/api_reference/html5.h.rst
lorsanta: With ``` :c:data:`bool` ``` I get this error ```emscripten/site/source/docs/api_reference/html5.h.rst:109: WARNING: Unparseable C cross-reference: 'bool'
Invalid C declaration: Expected identifier, got user-defined keyword: bool. Remove it from c_extra_keywords to allow it as identifier.
Currently c_extra_keywords is ['alignas', 'alignof', 'bool', 'complex', 'imaginary', 'noreturn', 'static_assert', 'thread_local']. [error at 4]
  bool
  ----^```

I think it makes more sense to remove the cross reference rather than changing `c_extra_keywords` in `conf.py`

lorsanta:
> Wow! Thanks for working on this. This is awesome. I wonder if we could publish the output somewhere temporary so that is might be viewed before we land this and so we can compare the old and new output?

I guess I could create a github page If necessary

sbc100:
> > Wow! Thanks for working on this. This is awesome. I wonder if we could publish the output somewhere temporary so that is might be viewed before we land this and so we can compare the old and new output?
> 
> I guess I could create a github page If necessary

I wonder if we could setup CI such that the `build-docs` CI step automatically publishes somewhere temporary?

kripken:
@sbc100 that would be ideal, but unless we have an idea for a quick way to do it, I think these updates are rare enough that it isn't worth waiting on such a feature. I'd be fine with building this PR locally to view the changes before merging, in particular.

@lorsanta Looks like there are some errors on `build-docs`?

sbc100:
> @sbc100 that would be ideal, but unless we have an idea for a quick way to do it, I think these updates are rare enough that it isn't worth waiting on such a feature. I'd be fine with building this PR locally to view the changes before merging, in particular.
> 

Yup, that sounds find to me.  As long as the new site can be reviews somewhere before this PR lands.   Perhaps you (@kripken) could publish the results under `https://emscripten.org/docs/testing/`?

kripken:
@sbc100 I can upload it, sure. How about `emscripten.org/testing/`? (`/docs/` is already part of the proper output, so it would mix them) Or maybe even `/_testing/` to make it clear it is internal.

@lorsanta Building locally, I had to do
```diff
-docutils==0.22
+docutils==0.21.2
```
because it couldn't find 0.22.

Pages look valid after that (and also a little nicer/more modern!) but I also ran into a problem: doing a search (typing in the search box on the upper left, where it says "Search docs", and then pressing enter) hangs on "Searching...", and the JS console shows an error
```js
searchtools.js:281 Uncaught ReferenceError: Stemmer is not defined
    at Object._parseQuery (searchtools.js:281:21)
    at Object.query (searchtools.js:416:91)
    at Object.setIndex (searchtools.js:226:14)
    at searchindex.js:1:8
```

kripken:
Uploaded: https://emscripten.org/_testing/

kripken:
Another difference I just noticed is that on the main page, under "About this site" the "Index" has vanished in the new version. Is that intentional?

sbc100:
> Another difference I just noticed is that on the main page, under "About this site" the "Index" has vanished in the new version. Is that intentional?

The anchors seems a little off too.  When I use the index on the left to click on things like https://emscripten.org/_testing/docs/tools_reference/settings_reference.html#maximum-memory it seems to scroll down the that place in the index, but not the `MAXIMUM_MEMORY` anchor on the main page on the right.

lorsanta:
> Pages look valid after that (and also a little nicer/more modern!) but I also ran into a problem: doing a search (typing in the search box on the upper left, where it says "Search docs", and then pressing enter) hangs on "Searching...", and the JS console shows an error

Sorry my bad, I forgot to update `search.html`. I'll commit the change later with everything else.

> @lorsanta Looks like there are some errors on build-docs?

emscripten-ci has `python-3.8.10` so I can't use `sphinx==8.1.3`, I guess I'll have to change `requirements-dev.txt` again

> Another difference I just noticed is that on the main page, under "About this site" the "Index" has vanished in the new version. Is that intentional?

Oh no it isn't, probably I forgot to add something in `layout.html`.

> The anchors seems a little off too. When I use the index on the left to click on things like https://emscripten.org/_testing/docs/tools_reference/settings_reference.html#maximum-memory it seems to scroll down the that place in the index, but not the `MAXIMUM_MEMORY` anchor on the main page on the right.

I'll look into these issues in the next few days.

kripken:
Sounds good, thanks @lorsanta ! This is a really nice PR, will be great to upgrade this.

requirements-dev.txt
```diff
@@ -17,19 +17,19 @@ lxml==4.9.2
...
 sphinxcontrib-jquery==4.0
-alabaster==0.7.14
-pygments==2.17.2
-jinja2==3.1
+alabaster<0.8,>=0.7
+pygments>=2.13
+jinja2>=3.0
 
-docutils==0.22
+docutils<0.21,>=0.18.1
```
sbc100, line 32: Why change this docutils dependency?

requirements-dev.txt
```diff
@@ -17,19 +17,19 @@ lxml==4.9.2
...
+sphinxcontrib-htmlhelp>=2.0.0
+sphinxcontrib-serializinghtml>=1.1.5
 sphinxcontrib-qthelp==1.0.6
 sphinxcontrib-jquery==4.0
-alabaster==0.7.14
-pygments==2.17.2
-jinja2==3.1
+alabaster<0.8,>=0.7
+pygments>=2.13
+jinja2>=3.0
```
sbc100, line 30: Can we require a specific version here to avoid version scew between different people's machines?

requirements-dev.txt
lorsanta: The container used for CircleCI has python 3.8 I had to use `sphinx 7.1.2` that requires `docutils<0.21,>=0.18.1`

lorsanta:
> The anchors seems a little off too. When I use the index on the left to click on things like https://emscripten.org/_testing/docs/tools_reference/settings_reference.html#maximum-memory it seems to scroll down the that place in the index, but not the `MAXIMUM_MEMORY` anchor on the main page on the right.

The issue was caused by `t[0].scrollIntoView()` in `theme.js`, so I commented it out

requirements-dev.txt
```diff
@@ -17,20 +17,19 @@ lxml==4.9.2
...
+sphinxcontrib-qthelp==1.0.3
+sphinxcontrib-jquery==4.0
+alabaster==0.7.13
 pygments==2.17.2
-# See https://github.com/readthedocs/readthedocs.org/issues/9038
-jinja2<3.1
+jinja2==3.0
 
-# Pin docutils because newer versions are not compatible with sphinx 2.4.4
-docutils==0.17.1
```
sbc100, line 33: How about if we simply remove all of these dependencies except for the sphinx one ones?

lorsanta: I saw that `sphinx 7.1.2` has `jinja >= 3.0` as a requirement.

I think it's best to keep `jinja2==3.0` to avoid future releases of `jinja2` to break things.

What do you think?

sbc100: I think its best to only include direct dependencies here, up until the point where we have some reason to do otherwise.

But i'm no expert so if you think this is safer/better then I'm fine with it.

How about the rest of these indirect deps, can  we remove those?

sbc100:
I think you can fix the mypy failure by adding `site/source/_themes/` to `exclude` in `.mypy.ini`

sbc100:
@kripken can you re-publish with the latest changes?

kripken:
@sbc100 uploaded

kripken:
Looks like "Index" is back in the contents, and search works (though the results are different than before - but I don't know if they are better or worse :smile: I tested with "asyncify" in both cases).

Overall things look good to me - I can't see anything wrong.

sbc100 (approved):
LGTM!

kripken (approved):

PR #22875, opened 2024-11-07
Use GLctx in library_egl.js and library_glut.js. NFC

sbc100:
I'm trying to reduce/remove the usage of `Module.ctx` with a view to possibly removing it from the `Module` object.  These usages should be using `GLctx` anyway I believe.


kripken (approved):
Agreed, this is a good direction. `Module.*` for what could be a normal var is best replaced.

PR #22876, opened 2024-11-07
Consistent quoting for Module['canvas']. NFC

sbc100:
127 or the 150 references to Module['canvas'] were already quoted.  In general this is how we handle properties of the Module object that we don't want closure to minify.  Once I do the same for the ctx object we can remove the closure type annotation completely from shell.js.


kripken:
Do we want this to remain on Module? I don't remember a reason for users to access it after construction, but I might be wrong.

If we just need it during construction then we could receive it from Module and then access it like `GLctx`?

kripken (approved):
I see, this sounds good as a step then.

Issue #22866, opened 2024-11-06
library_glfw.js blocks html inputs


uysalibov:
Why is the marked area in the link I gave in the glfw necessary? Although I do not use any key event in my own project, “backspace” and “tab” do not work in the inputs in html because of this marked field.

https://github.com/emscripten-core/emscripten/blob/94f41734f1e9f74a19ac209e35e803ca29fcdd25/src/library_glfw.js#L416-L425

sbc100:
This doesn't answer your question but that code is duplicated from: https://github.com/emscripten-core/emscripten/blob/fa5bc28be30ef78902230fbd4ab185044050de3d/src/library_sdl.js#L687-L694

So I guess SDL programs have the same issue?

uysalibov:
Yes, probably it has same issue too.

Issue #6143, opened 2018-01-25
GLFW key listener blocking backspace and tab


winduptoy:
https://github.com/kripken/emscripten/blob/incoming/src/library_glfw.js#L400

Why is it necessary to call `event.preventDefault()` when the key is backspace or tab here? This blocks backspace and tab on the **entire page, making text inputs on the page completely unusable**.

I'm not sure I understand the purpose of this anyways. Is it to prevent the user from exiting the page when they hit backspace? If so, maybe this should only be checked if the pointer is locked or fullscreen is enabled.

juj:
Commented about this in detail in the PR. Indeed the current behavior blocks on interacting with input on other DOM elements, and generally people have solved this by using the HTML5 Input API, where this is designed to be controllable per event. GLUT/GLFW/SDL each have defaulted to more or less assuming that the only interesting element on the page is the canvas, which has dedicated focus, since that helps most projects get going.

Smarter heuristics are definitely welcome, but we should be careful not to get into a "whose use case is the most important one to be the default?" debate. Perhaps some Emscripten-specific configuration options for GLUT/GLFW/SDL APIs might be in order?

winduptoy:
@juj Do you mind linking to some documentation about the HTML5 Input API? I'm afraid I don't know which API you're talking about, nor can I find it on Google. Is it the same as `addEventListener`/`removeEventListener`?

Are you sure that smarter heuristics are really a worthwhile solution here (read: even more uncontrollable magic code)? What do you think about a configuration option that ties the existing functionality to just the canvas so that everyone can have their own desired level of control?

```
var myModule = Module({
    attachGLFWEventsToCanvas: true
});
```
library_glfw.js, `glfwInit()`:

```
...
var eventTarget = window;
if(Module['attachGLFWEventsToCanvas']) eventTarget = Module['canvas'];
eventTarget.addEventListener("keydown", GLFW.onKeydown, true);
eventTarget.addEventListener("keypress", GLFW.onKeyPress, true);
eventTarget.addEventListener("keyup", GLFW.onKeyup, true);
eventTarget.addEventListener("blur", GLFW.onBlur, true);
...
```

juj:
> @juj Do you mind linking to some documentation about the HTML5 Input API? I'm afraid I don't know which API you're talking about, nor can I find it on Google. Is it the same as addEventListener/removeEventListener?

[...]

PR #22879, opened 2024-11-07
Fixing the bug that prevents any text on the site from being deleted

uysalibov:
If glfw is initizalized, it prevents the deletion of text in the input and textarea elements on the site. If this is not filtered like this, no form on the site works as desired.
Related issue: https://github.com/emscripten-core/emscripten/issues/22866


src/library_glfw.js
```diff
@@ -419,7 +419,10 @@ var LibraryGLFW = {
       // This logic comes directly from the sdl implementation. We cannot
       // call preventDefault on all keydown events otherwise onKeyPress will
       // not get called
-      if (event.keyCode === 8 /* backspace */ || event.keyCode === 9 /* tab */) {
+      // Prevent default behavior for backspace and tab keys when the event target is not an input or textarea element.
+      // If prevented when input and textarea are prevented, no text in the site can be deleted
+      if ( event.target.localName !== "input" && event.target.localName !== "textarea" &&
```
sbc100, line 424: No space after `(`.

src/library_glfw.js
```diff
@@ -419,7 +419,10 @@ var LibraryGLFW = {
       // This logic comes directly from the sdl implementation. We cannot
       // call preventDefault on all keydown events otherwise onKeyPress will
       // not get called
-      if (event.keyCode === 8 /* backspace */ || event.keyCode === 9 /* tab */) {
+      // Prevent default behavior for backspace and tab keys when the event target is not an input or textarea element.
+      // If prevented when input and textarea are prevented, no text in the site can be deleted
```
sbc100, line 423: How about `Otherwise, no text in the site can be deleted`.

Can you wrap this comment at 80 columns?

Should we make the same change to library_sdl.js perhaps?

src/library_glfw.js
sbc100, line 424: Why use `target.localName` rather than, for example, `target.tabName`?  I'm not sure what `localName` is so maybe you can help me out?

sbc100 (commented):
Thanks for working on this BTW!  I think we should probably add an interative test to make sure its working as expected.

sbc100:
> I tried to make the improvements you mentioned. I have not used tabName before, thank you for the advice, it seems to be a better choice. Also I added same things for sdl. @sbc100

Sorry I meant `tagName`: https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName

sbc100:
It looks like you somehow included copies of some other changes that already landed when you rebased/merged here.

uysalibov:
I wanted to remove old one for new 2 commit. But somehow i added unrelated commits.

kripken:
Is this related to https://github.com/emscripten-core/emscripten/issues/6143 ? I don't think we had a clear resolution there.

uysalibov:
> Is this related to #6143 ? I don't think we had a clear resolution there.

Yes, I try to fix that problem.

kripken:
@uysalibov I understand, but see @juj's point in that [older discussion](https://github.com/emscripten-core/emscripten/issues/6143#issuecomment-360939684) - there are tradeoffs here, and it's not obvious that flipping it the other way is better. And it would be disruptive to existing users.

I don't have a strong preference myself, except that if we change a long-standing tradeoff in a disruptive way then we should have a very strong reason.

uysalibov:
@kripken I can't find a way to remove these related preventation in the structure in the old version. Assuming that the whole page is canvas doesn't seem very correct to me today. It's easier to block these keys (backspace and tab) in the project. On the contrary, we cannot find a solution to this situation since 2018.

kripken:
> Assuming that the whole page is canvas doesn't seem very correct to me today.

I agree it might not be the best default. It does help porting fullscreen games and similar applications, which historically has been popular with Emscripten. I think that's why we ended up here.

I agree we should think about how to improve things. Can you explain how the `"INPUT", "TEXTAREA"` changes in this PR work? I'm trying to figure out how disruptive this will be.

uysalibov:
Imagine a page with canvas and input. Normally backspace and tab operations will be prevented because there is a canvas. But if you click on the input element and try to delete something, this if expreson will not be entered and you can delete the text as you wish. The main point is to click on the input and textarea elements. Otherwise it will continue to work as before

kripken:
@uysalibov I understand that, but I am asking specifically about `"INPUT"` and `"TEXTAREA"`. Does the user need to specify those in the HTML as the `id` or something like that? If so we should document it.

Another option might be to check if the event was sent to the canvas, and swallowed if so.

sbc100:
> @uysalibov I understand that, but I am asking specifically about `"INPUT"` and `"TEXTAREA"`. Does the user need to specify those in the HTML as the `id` or something like that? If so we should document it.

No, those are the names of the HTML tags that accept text as input (Here we are looking at the `tagName` of the target).  Seems like a pretty safe change to me.

> 
> Another option might be to check if the event was sent to the canvas, and swallowed if so.

kripken:
Oh, thanks, I was confusing Id with Tag I guess. In that case this does sound pretty safe, sgtm.

I do think we should add testing for it. Relevant tests is `browser.test_glfw_events`, perhaps we can add to that? The source file is

https://github.com/emscripten-core/emscripten/blob/main/test/browser/test_glfw_events.c

and you can run it with `./test/runner browser.test_glfw_events`

uysalibov:
> Another option might be to check if the event was sent to the canvas, and swallowed if so.

I think it's not an option. Because the key event does not always come from the canvas element. Most of the time it comes from the `body`. If we do this it will break porting projects. `input` and `textarea` filtering seems safer.

uysalibov:
I don't know the project well enough to write a test for the emscriptten project. That's why I can't write it. Is there any chance you can write it?
@sbc100 @kripken

sbc100:
I can look into helping to write a test.  Can you rebase and resolve conflicts?

uysalibov:
@sbc100 Is the change I made enough? When I rebased the commit named “Merge branch ‘main’ into main”, I couldn't rebase it because github gave a merge conflict warning. But now I edited it according to the conflict commit.

sbc100 (approved):

uysalibov:
Why test-browser-firefox gives timeout error?.

sbc100:
Thats likely just a flaky test.  However, I don't want to land this PR without a test anyway, so we are still blocked on that. 

I was thinking of looking into the testing side myself, but if you have time to look into that would be great.

kripken:
I wrote tests in #23285 . Things look good for glfw with those tests and this PR, but not the SDL side (see details there). @uysalibov did you test the SDL side of this PR locally and see that it works?

PR #22881, opened 2024-11-07
Reduce usage of legacy `event.keyCode`. NFC

sbc100:
This change replaced the usage of the legacy `keyCode` attribute with the preferred `key` attribute, which also has the advantage of removing some hardcoded constant numbers.

In order to test this change I fixed the `test_glfw_get_key_stuck` test and added `test_sdl_key_test` to the interactive tests.

In doing so I noticed and fixed a crash bug that was introduced in #22874.

### DIFF
--- a/src/library_sdl.js
+++ b/src/library_sdl.js
@@ -696,7 +696,7 @@ var LibrarySDL = {
           // won't fire. However, it's fine (and in some cases necessary) to
           // preventDefault for keys that don't generate a character. Otherwise,
           // preventDefault is the right thing to do in general.
-          if (event.type !== 'keydown' || (!SDL.unicode && !SDL.textInput) || (event.keyCode === 8 /* backspace */ || event.keyCode === 9 /* tab */)) {
+          if (event.type !== 'keydown' || (!SDL.unicode && !SDL.textInput) || (event.key == 'Backspace' || event.key == 'Tab')) {
             event.preventDefault();
           }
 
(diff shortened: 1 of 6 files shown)


sbc100:
I noticed this could be improved while reviewing #22879.  Whichever lands first will have to deal with a conflict.

src/library_sdl.js
```diff
@@ -553,7 +553,7 @@ var LibrarySDL = {
     receiveEvent(event) {
       function unpressAllPressedKeys() {
         // Un-press all pressed keys: TODO
-        for (var keyCode of SDL.keyboardMap) {
+        for (var keyCode of Object.values(SDL.keyboardMap)) {
```
kripken, line 556: Is this the crash fix?

kripken (approved):
lgtm if relevant interactive tests pass. By my reading of the MDN docs this looks valid.

src/library_sdl.js
sbc100: Yup, I see this crash when running the interactive tests.

src/proxyClient.js
```diff
@@ -313,7 +313,7 @@ if (!ENVIRONMENT_IS_NODE) {
 // Only prevent default on backspace/tab because we don't want unexpected navigation.
 // Do not prevent default on the rest as we need the keypress event.
 function shouldPreventDefault(event) {
-  if (event.type === 'keydown' && event.keyCode !== 8 /* backspace */ && event.keyCode !== 9 /* tab */) {
+  if (event.type === 'keydown' && event.key != 'Backspace' && event.key != 'Tab') {
```
sbc100, line 316: BTW the logic here seems to be the inverse of what the comment says.    Shouldn't all the clauses above be `===`?

sbc100:
I split out the crash fix: #22884

src/proxyClient.js
kripken: Looks right to me? Comment above says "prevent default on backspace/tab", the if checks "not backspace, not tab" and returns false if so. So it returns true - prevents default - on backspace/tab as described.

PR #22884, opened 2024-11-07
Fix crash introduced in #22874

sbc100:


kripken (approved):
Can we add a test for this?

sbc100:
> Can we add a test for this?

I think writing an automated test would be non-trivial, and given how many gaps we have in automatic GUI testing I'm not sure its worth it.

sbc100:
I think I would need to simulate a `visibilitychange` or `blur` event in between key presses.

kripken:
I agree it's non-trivial, but I think we can simulate such events like we do keypresses and mouse moves, and interleave it with them. The order of such event handling is deterministic. I'm just worried we have big gaps in testing of those GUI APIs, and it would be good to think about improving it, basically.

sbc100:
> I agree it's non-trivial, but I think we can simulate such events like we do keypresses and mouse moves, and interleave it with them. The order of such event handling is deterministic. I'm just worried we have big gaps in testing of those GUI APIs, and it would be good to think about improving it, basically.

I totally agree.   I'll take a look at this one.   My reaction was basically broken windows syndrome I suppose:  There are way bigger and more important gaps than this one :)

PR #22885, opened 2024-11-08
Use struct info in src/library_sdl.js. NFC

sbc100:


src/library_sdl.js
```diff
@@ -348,13 +348,13 @@ var LibrarySDL = {
     translateRGBAToColor: (r, g, b, a) => r | g << 8 | b << 16 | a << 24,
 
     makeSurface(width, height, flags, usePageCanvas, source, rmask, gmask, bmask, amask) {
-      var is_SDL_HWSURFACE = flags & 0x00000001;
```
sbc100, line 351: It looks like `SDL_HWSURFACE` constant was wrong here.  It should have been `0x08000001`.   Luckily it was still matching because its a subset of the bits.

kripken (approved):
We have some testing of this, but I'm not sure how much. Might be worth checking.

sbc100:
> We have some testing of this, but I'm not sure how much. Might be worth checking.

We have tests at least the keyboard stuff because I made a mistake there and they failed.   I double checked everything again and I think we are good now.

PR #22891, opened 2024-11-08
[JSPI] Automatically add promising JSPI wrappers to thread entry points.

### LINKED ISSUE #22354: JSPI and pthreads shows RuntimeError - invalid suspender object for suspend

Thread model: posix
InstalledDir: /home/jspanchu/dev/emsdk/upstream/bin

```
emcc ./main.c -sJSPI -pthread -sPROXY_TO_PTHREAD -o index.html
```

Note that it works when I use ASYNCIFY.

```cpp
#include <stdio.h>

int main(int argc, char** argv)
{
    int counter = 0;
    while (counter < 4)
    {
        emscripten_sleep(1);
        printf("counter=%d\n", counter);
        ++counter;
    }
    printf("counter=%d\n", counter);
    return 0;
}
```

Error:
```
worker: onmessage() captured an uncaught exception: RuntimeError: invalid suspender object for suspend
index.js:364 
RuntimeError: invalid suspender object for suspend
    at index.wasm.main (index.wasm)
    at index.wasm.__main_void (index.wasm)
    at index.wasm._main_thread (index.wasm)
    at invokeEntryPoint (index.js:1734:42)
    at handleMessage (index.js:459:11)

index.js:1579 Pthread 0x000123b0 sent an error! http://localhost:8000/index.js:489: Uncaught RuntimeError: invalid suspender object for suspend
index.js:1580 Uncaught 
ErrorEvent {isTrusted: true, message: 'Uncaught RuntimeError: invalid suspender object for suspend', filename: 'http://localhost:8000/index.js', lineno: 489, colno: 7, …}
(index):195 [post-exception status] Exception thrown, see JavaScript console
index.js:489 Uncaught 
RuntimeError: invalid suspender object for suspend
    at index.wasm.main (index.wasm)
    at index.wasm.__main_void (index.wasm)
    at index.wasm._main_thread (index.wasm)
    at invokeEntryPoint (index.js:1734:42)
    at handleMessage (index.js:459:11)
```

brendandahl:
Use a WebAssembly.Promising wrapper on the thread entry points to allow
JSPI to be used from worker threads. The code that invokes the wrapper
also adds async/await to handle this async call to entry point.

Fixes https://github.com/emscripten-core/emscripten/issues/22354


src/runtime_pthread.js
```diff
@@ -90,7 +90,7 @@ if (ENVIRONMENT_IS_PTHREAD) {
   // notified about them.
   self.onunhandledrejection = (e) => { throw e.reason || e; };
 
-  function handleMessage(e) {
+  {{{ asyncIf(ASYNCIFY == 2) }}} function handleMessage(e) {
```
sbc100, line 93: It seems like these fixes would apply to JSPI using in the main function of any pthread, not just the PROXY_TO_PTHREAD mode.   Is there something specific to PROXY_TO_PTHREAD here?

brendandahl: Yeah, it's not totally specific to PROXY_TO_PTHREAD, more it showed up as an issue there since the entry point for the main proxy is not the users actual main. 

With a user defined entry point there were ways of handling this (see core.test_pthread_join_and_asyncify). I think I can actually change that test now to not require the user to specify that the entry function is JSPI. Hower, this makes me wonder if we should automatically JSPI wrap thread entry functions or require the user to specify that they need JSPI. FWIW, we currently auto JSPI wrap main.

sbc100: Do an alternative solution to this one would be to auto-wrap ` _main_thread` which is used in `_emscripten_proxy_main` ?

If the clang annotation was working we could use that there?

brendandahl: > So an alternative solution to this one would be to auto-wrap ` _main_thread` which is used in `_emscripten_proxy_main` ?

Yeah, we'd need `EMSCRIPTEN_KEEPALIVE` or add it to `EXPORTED_FUNCTIONS` and also add it to `DEFAULT_ASYNCIFY_EXPORTS`.

> If the clang annotation was working we could use that there?

Yeah, I still need to fix binaryen to handle the annotation.

kripken (approved):

brendandahl:
Per discussion in meeting, we'll auto wrap all thread entry points. I've removed my new test, but updated the original test to cover the case of suspending on them main thread too. 

(sorry for the force push, but I wanted to amend to make sure I removed all new code)

test/core/test_pthread_join_and_asyncify.c
```diff
@@ -13,15 +13,17 @@ EM_ASYNC_JS(int, async_call, (), {
...
-EMSCRIPTEN_KEEPALIVE void *run_thread(void *args) {
+void *run_thread(void *args) {
   int ret = async_call();
   assert(ret == 42);
   return NULL;
 }
 
 int main() {
   pthread_t id;
+  // Test that JSPI works on main thread.
```
kripken, line 24:
```suggestion
  // Test that JSPI works on the main thread.
```

kripken (approved):

sbc100 (approved):

brendandahl:
Wasm64 was broken since it was trying to do `WebAssembly.promising(func)` where func was a wrapper and not the actual wasm export. I added an option to makeDynCall to add the `WebAssembly.promising` there instead.

sbc100 (approved):

kripken (approved):

PR #22894, opened 2024-11-08
Use llrintf over llrint in SSE funcs. NFC

sbc100:
These function operate of floats so this is slighlyt more correct and perhaps faster in some cases I guess.

### DIFF
--- a/system/include/compat/emmintrin.h
+++ b/system/include/compat/emmintrin.h
@@ -1008,7 +1008,7 @@ _mm_cvtsd_si64(__m128d __a)
   // TODO: optimize
   if (isnan(__a[0]) || isinf(__a[0])) return 0x8000000000000000LL;
   long long x = llrint(__a[0]);
-  if (x != 0xFFFFFFFF00000000ULL && (x != 0 || fabsf(__a[0]) < 2.f))
+  if (x != 0xFFFFFFFF00000000ULL && (x != 0 || fabs(__a[0]) < 2.f))
     return x;
   else
     return 0x8000000000000000LL;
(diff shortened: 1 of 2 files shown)


system/include/compat/emmintrin.h
```diff
@@ -1007,7 +1007,7 @@ _mm_cvtsd_si64(__m128d __a)
 {
   // TODO: optimize
   if (isnan(__a[0]) || isinf(__a[0])) return 0x8000000000000000LL;
-  long long x = llrint(__a[0]);
+  long long x = llrintf(__a[0]);
```
tlively, line 1010: But `__m128d` is f64x2, so operating on doubles should be more correct.

sbc100: Do is the `fabsf` on the line below wrong in this case then?

sbc100: Updated the line below instead.

system/include/compat/emmintrin.h
```diff
@@ -1020,7 +1020,7 @@ _mm_cvttsd_si64(__m128d __a)
   // TODO: optimize
   float e = __a[0];
   if (isnan(e) || isinf(e) || e > LLONG_MAX || e < LLONG_MIN) return 0x8000000000000000LL;
-  long long x = llrint(e);
+  long long x = llrintf(e);
```
tlively, line 1023: Same issue here. This should be `llrint` and `fabs`.

tlively (approved):

system/include/compat/emmintrin.h
sbc100: Oops that means `e` has the wrong type too (from #22893).  Fixed

Issue #22847, opened 2024-11-04
Can't scale canvas with CSS without it changing the canvas Width and Height properties


michaelfiber:
**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.70 (b53978ee3f540dc74761eba127aa7f1b8761a125)
clang version 20.0.0git (https:/github.com/llvm/llvm-project f52b89561f2d929c0c6f37fd818229fbcad3b26c)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /home/michael/dev/raylib-test/emsdk/emsdk-main/upstream/bin
```

Starting with version 3.1.51 the width and height properties of a canvas as overridden if you use CSS to change the size of the canvas.  How do we maintain the original way of doing it where you can use CSS to change the canvas.style.width and canvas.style.height without it overriding the canvas.width and canvas.height? 

There is an issue open on the raylib project related to this: https://github.com/raysan5/raylib/issues/4464 - but it doesn't seem like an issue with raylib itself.

The goal is to continue to allow emscripten programs compiled with GLFW=3 to be scaled up to fill a containing element with `object-fit: cover` or something similar so that you could, for instance, use CSS to allow an itch.io game to go fullscreen while the game itself remains a fixed resolution.

sbc100:
Can you confirm this was broken between 3.1.50 and 3.1.51?  Could you perhaps bisect to see exactly which emscripten change broke this: https://emscripten.org/docs/contributing/developers_guide.html#bisecting

sbc100:
I see only 3 changes to the glfw between 3.1.50 and 3.1.51 so I suppose the culprit is likely one these:

```
$ git l 3.1.50..3.1.51  src/library_glfw.js
05821a3ed Leverage coalescing operators. NFC (#20531) 11 months ago [Ingvar Stepanyan]
c26fc388e Add Hi DPI Support for GLFW (#20584) 11 months ago [Yan Pujante]
95591bcfb Use `let` for loop variable instead of IIFEs. NFC (#20823) 11 months ago [Sam Clegg]
7d639f7a6 Fix glfw3 default hints being modified (#20770) 11 months ago [Yan Pujante]
b3b0f2794 Fix touch input handling for glfw. (#20805) 11 months ago [Ruben]
```

My guess is that it would be #20584.  @ypujante WDYT?

curiousdannii:
[...]

PR #22900, opened 2024-11-09
[GLFW] Restore css scaling behavior (removed in 3.1.51)

ypujante:
This is the patch I am proposing for addressing #22847 

- the changes in 3.1.51 to add HiDPI support removed a feature that is not documented nor a feature of GLFW
- raylib wants it back so this patch restores it

Some points:

* I added a function `isCSSScalingEnabled()` with documentation to make it clear what is happening
* Due to the fact that HiDPI / 4k support overrides CSS, CSS scaling is disabled in that mode. There would be a conflict otherwise
* Since this feature is not really standard, I introduced a way to disable it (since it was implemented prior to 3.1.51, I felt it was better to have it on by default with a way to disable it rather than the other way around).

I have tested the changes as best as I could with ImGui and raylib. I tested a few examples with raylib and it seems to be doing what they want (the canvas occupies the fullscreen and the mouse coordinates are correct and resizing the window follows appriopriately)

It would certainly be better if @michaelfiber and/or @raysan5 could test the changes before they go live. It should be a matter of checking out this PR but I don't know if they have any bandwidth to do this.


ypujante:
After the changes are approved I will modify Changelog to add an entry about this.

raysan5:
@ypujante Thanks for the review. I tested this change on latest raylib master branch, with `emsdk 3.1.71`, replacing `library_glfw.js` manually, and it works as expected, the canvas and input are scaled to `width: 100%`.

Please, excuse me if some of my previous messages could be rude. Thanks again for reviewing it.

ypujante:
The only test failure I am seeing is

```text
ERROR [61.246s]: test_offset_converter (test_browser.browser)
  File "/root/project/test/common.py", line 2188, in run_browser
  File "/root/project/test/common.py", line 1518, in assertContained
AssertionError: Expected to find '/report_result?exit:1
' in '[no http server activity]
', diff:
```

which seems to be very unlikely related to these unrelated changes

src/library_glfw.js
```diff
@@ -1308,10 +1321,27 @@ var LibraryGLFW = {
         return false;
     },
 
+    /**
+     * CSS Scaling is a feature that is NOT part of the GLFW API, but for historical reasons, it is available
+     * in Emscripten.
+     * It is enabled by default but can be disabled by setting Module['EMSCRIPTEN_GLFW_DISABLE_CSS_SCALING'] to true.
+     * It is automatically disabled when using Hi DPI (the library overrides CSS sizes). */
+    isCSSScalingEnabled() {
+      return Module['EMSCRIPTEN_GLFW_DISABLE_CSS_SCALING'] !== true && !GLFW.isHiDPIAware();
```
sbc100, line 1330: I'm somewhat loath to add more incoming module API elements like this.   Can we perhaps wait until somebody asks for this setting?

If we do end up adding it it would need to be added to INCOMING_MODULE_JS_API and then this line should be wrapped in `#if expectToReceiveOnModule('INCOMING_MODULE_JS_API')`.

I simpler more pay-as-you-go option could be to add some kind of new API for this either via `glfwSetWindowParam` or some new C API. 

I wonder if we can add a test for this behaviour?   Can we perhaps make a simple reftest that shows that the GLFW windows is scaled to the size of the canvas?  I'm not sure how hard that would be.

ypujante:
FYI, I am aware of the test failure (which does not fail on my local machine). I am trying to figure out why... will report when I do.

src/library_glfw.js
```diff
@@ -1112,7 +1112,8 @@ var LibraryGLFW = {
 
       // Get non alive id
       const canvas = Module['canvas'];
-      var win = new GLFW_Window(id, canvas.clientWidth, canvas.clientHeight, canvas.width, canvas.height, title, monitor, share);
+
```
sbc100, line 1115: Did you mean to add an extra newline here?  Maybe remove it?

test/browser/test_glfw3_css_scaling.c
```diff
@@ -0,0 +1,119 @@
...
+    printf("Use case #1\n");
+    GLFWwindow* window = glfwCreateWindow(640, 480, "test_glfw3_css_scaling.c", NULL, NULL);
+    assert(window);
+    checkWindowSize(window, 640, 480, 1.0);
+    double w, h;
+    emscripten_get_element_css_size("#canvas", &w, &h);
+    printf("CSS Size=%.0fx%.0f\n", w, h);
+    assert(w == 640 && h == 480);
+    glfwDestroyWindow(window);
+  }
```
sbc100, line 78: How about putting each of these test cases in its own C function so the main function looks like:

```
int main() {
  assert(glfwInit() == GLFW_TRUE);
  test_foo();
  test_bar();
  test_baz();
  printf("All tests complete\n");
  glfwTerminate();
  return 0;
}
```

This form has a few advantaged I think.

sbc100 (approved):

raysan5:
Just tested with latest raylib and this PRs `library_glfw.js` copied over `emsdk 3.1.71`. It works with same previous behaviour before `emsdk 3.1.51`. Thank you very much for reviewing this issue so quickly.

sbc100 (approved):

PR #22902, opened 2024-11-10
Fix typo in Closure extern

kleisauke:
Prevents Closure compiler from mangling `require('os').tmpdir()`.


kleisauke:
As a possible follow-up, we might drop all these externs in favor of the ones available in `node_modules/google-closure-compiler/contrib/nodejs/*.js`.

sbc100 (approved):

PR #22912, opened 2024-11-12
Avoid using Module.instantiateWasm for pthreads.

sbc100:
This leave the Module.instantiateWasm API for users to overide as they please.

Fixes: #22905

PR #22913, opened 2024-11-12
Use -sSTRICT in test_codesize_minimal_pthreads. NFC

sbc100:


kripken (commented):
1. Why add this? Perhaps not many users apply it. Or is the idea to add coverage for the users that do, in at least one test?
2. Why does code size decrease with this option?

sbc100:
> 1. Why add this? Perhaps not many users apply it. Or is the idea to add coverage for the users that do, in at least one test?

The idea is that its a good way to minimize code size and in general (I think) we want to measure the minimum code size a user can get if they enable all the right options (this is why we have `-Oz` there).

> 2. Why does code size decrease with this option?

Mostly because it defaults to an empty `INCOMING_MODULE_JS_API`, but there could other things in the list of things that STRICT does that help too:

```
//   - The C define EMSCRIPTEN is not defined (__EMSCRIPTEN__ always is, and        
//     is the correct thing to use).                                             
//   - STRICT_JS is enabled.                                                     
//   - IGNORE_MISSING_MAIN is disabled.                                          
//   - AUTO_JS_LIBRARIES is disabled.                                            
//   - AUTO_NATIVE_LIBRARIES is disabled.                                        
//   - DEFAULT_TO_CXX is disabled.                                               
//   - USE_GLFW is set to 0 rather than 2 by default.                            
//   - ALLOW_UNIMPLEMENTED_SYSCALLS is disabled.                                 
//   - INCOMING_MODULE_JS_API is set to empty by default.   
```

sbc100:
The fact that `INCOMING_MODULE_JS_API` defaults to empty allows the code size saving in #22912 to be demonstrated.

test/test_other.py
```diff
@@ -8758,7 +8758,7 @@ def test_codesize_minimal(self, *args):
 
   @node_pthreads
   def test_codesize_minimal_pthreads(self):
```
kripken, line 8760: Maybe rename this to add `_strict`?

kripken (approved):
I see, thanks. Makes sense.

test/test_other.py
sbc100: The `test_codesize_minimal` test above has `STRICT` too without including it in the name, I think its fair to assume that most minimal codesize tests include these flags.

PR #22920, opened 2024-11-13
Ignore unknown keyCodes in library_sdl.js

sbc100:
Prior to this change any unrecognized DOM keyCode would be directly passed through with the DOM keyCode used as an SDL keycode.

This strange behavior is what allowed `test_sdl_key_proxy` to use SDL key codes when generating new `KeyboardEvent`s.  The number previously hardcoded in the test (1248, 1249, 1250) correspond to (SDLK_LCTRL, SDLK_LSHIFT, SDLK_LALT) but don't make any sense as DOM keyboard codes.

### DIFF
--- a/src/library_sdl.js
+++ b/src/library_sdl.js
@@ -823,13 +823,16 @@ var LibrarySDL = {
 
     lookupKeyCodeForEvent(event) {
       var code = event.keyCode;
-      if (code >= 65 && code <= 90) {
+      if (code >= 65 && code <= 90) { // ASCII A-Z
         code += 32; // make lowercase for SDL
       } else {
+        // Look up DOM code in the keyCodes table with fallback for ASCII codes
+        // which can match between DOM codes and SDL keycodes (allows keyCodes
+        // to be smaller).
+        code = SDL.keyCodes[code] || (code < 128 ? code : 0);
 #if RUNTIME_DEBUG
-        if (!(event.keyCode in SDL.keyCodes)) dbg('unknown keyCode: ', event.keyCode);
+        if (!code) dbg('unmapped keyCode: ', event.keyCode);
 #endif
-        code = SDL.keyCodes[event.keyCode] || event.keyCode;
         // If this is one of the modifier keys (224 | 1<<10 - 227 | 1<<10), and the event specifies that it is
         // a right key, add 4 to get the right key SDL key code.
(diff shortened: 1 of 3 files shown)


src/library_sdl.js
```diff
@@ -823,13 +823,16 @@ var LibrarySDL = {
...
     lookupKeyCodeForEvent(event) {
       var code = event.keyCode;
-      if (code >= 65 && code <= 90) {
+      if (code >= 65 && code <= 90) { // ASCII A-Z
         code += 32; // make lowercase for SDL
       } else {
+        // Look up DOM code in the keyCodes table with fallback for ASCII codes
+        // which can match between DOM codes and SDL keycodes (allows keyCodes
+        // to be smaller).
+        code = SDL.keyCodes[code] || code < 128 ? code : 0;
```
kripken, line 832:
```js
        code = (SDL.keyCodes[code] || code < 128) ? code : 0;
or
        code = SDL.keyCodes[code] || (code < 128 ? code : 0);
```
?

test/test_browser.py
```diff
@@ -990,8 +990,8 @@ def post():
       html = html.replace('</body>', '''
 <script src='fake_events.js'></script>
 <script>
-simulateKeyDown(1250);simulateKeyDown(38);simulateKeyUp(38);simulateKeyUp(1250); // alt, up
-simulateKeyDown(1248);simulateKeyDown(1249);simulateKeyDown(40);simulateKeyUp(40);simulateKeyUp(1249);simulateKeyUp(1248); // ctrl, shift, down
+simulateKeyDown(18);simulateKeyDown(38);simulateKeyUp(38);simulateKeyUp(18); // alt, up
+simulateKeyDown(17);simulateKeyDown(16);simulateKeyDown(40);simulateKeyUp(40);simulateKeyUp(16);simulateKeyUp(17); // ctrl, shift, down
```
kripken, line 994: I'm not sure how to understand this change?

sbc100: I tried to describe this in the PR description.

The old code is passing raw SDL key values here (1248, 1249, 1250 which correspond to SDLK_LCTRL, SDLK_LSHIFT, SDLK_LALT).

This means that resulting DOM keyboard event contains SDL keybaord codes, which doesn't make any sense.  However because the SDL code was just copying any keyboard codes it didn't know about directly from the DOM event to the SDL event this was magically working.. but it should not have been.

The new key codes are the correct key codes from the corresponding DOM keys.

kripken: Now I see, thanks!

kripken (approved):

PR #22921, opened 2024-11-13
Implement SDL_GetKeyName

sbc100:
I found this pretty useful during debugging recently.

### DIFF
--- a/src/library_sdl.js
+++ b/src/library_sdl.js
@@ -1767,9 +1767,16 @@ var LibrarySDL = {
   SDL_GetKeyState: () => _SDL_GetKeyboardState(0),
 
   SDL_GetKeyName__proxy: 'sync',
-  SDL_GetKeyName__deps: ['$stringToNewUTF8'],
+  SDL_GetKeyName__deps: ['$stringToUTF8', 'realloc'],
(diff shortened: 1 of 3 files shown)


src/library_sdl.js
```diff
@@ -1767,9 +1767,16 @@ var LibrarySDL = {
...
   SDL_GetKeyName: (key) => {
-    SDL.keyName ||= stringToNewUTF8('unknown key');
+    var name = '';
+    /* ASCII A-Z or 0-9 */
+    if ((key >= 97 && key <= 122) || (key >= 48 && key <= 57)) {
+      name = String.fromCharCode(key);
+    }
+    var size = lengthBytesUTF8(name) + 1;
+    SDL.keyName = _realloc(SDL.keyName, size);
+    stringToUTF8(name, SDL.keyName, size);
```
kripken, line 1779: Maybe add `keyName: 0` to the SDL class for closure? Though if it doesn't error, maybe it doesn't matter.

kripken (approved):

src/library_sdl.js
sbc100: I tend to do that kind of thing if (and only if) closure complains.

Issue #22889, opened 2024-11-08
Cannot read properties of undefined (reading 'getParameter') since 3.1.70 (WebGL)


FrancoisRecisio:
Hello! Since the version 3.1.70, we have an error at runtime when starting our application using Emscripten and Electron. It worked fine in 3.1.69, I tried to find a commit related to our problem but couldn't find it, so maybe you'll have an idea.
Here is the diff : https://github.com/emscripten-core/emscripten/compare/0486179..12cedf4
If needed, I could test each commit dicotomically, but that would take a lot of time.
The problem is still here in the latest tot version.

**Version of emscripten/emsdk with the problem (first and latest):**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.70 (b53978ee3f540dc74761eba127aa7f1b8761a125)
clang version 20.0.0git (https:/github.com/llvm/llvm-project f52b89561f2d929c0c6f37fd818229fbcad3b26c)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: C:\git\emsdk\upstream\bin

emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.72-git (ffeb76137e765a0023de2635420a4827ca15106d)
clang version 20.0.0git (https:/github.com/llvm/llvm-project 31af00fda73df8e85e30c71f66e096d486ec8c8b)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: C:\git\emsdk\upstream\bin
Build config: +assertions
```

The error is :
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getParameter')
    at emscriptenWebGLGet (MyProject.js:10167:24)
    at _glGetIntegerv (MyProject.js:10295:66)
    at imports.<computed> (MyProject.js:6825:20)
    at MyProject.wasm._do_call (MyProject.wasm:0x2b87f34)
    at MyProject.wasm.em_task_queue_execute (MyProject.wasm:0x2b82a89)
    at MyProject.wasm.receive_notification (MyProject.wasm:0x2b83a15)
    at MyProject.wasm.em_task_queue_execute (MyProject.wasm:0x2b82a89)
    at MyProject.wasm._emscripten_check_mailbox (MyProject.wasm:0x2b9df0c)
    at ret.<computed> (MyProject.js:6851:20)
    at MyProject.js:1078:12
```

The generated js :
```
   case 33307:
[...]

PR #22927, opened 2024-11-14
Fix keepRuntimeAlive for pthreads

sbc100:
Still need to figure out how to test this.

Fixes: #22889


src/library.js
```diff
@@ -2078,16 +2078,11 @@ addToLibrary({
...
-  // When `noExitRuntime` is not include and EXIT_RUNTIME=0 then we know the
+#elif !EXIT_RUNTIME && !PTHREADS
+  // When `noExitRuntime` is not included and EXIT_RUNTIME=0 then we know the
   // runtime can never exit (i.e. should always be kept alive).
-  // However for pthreads we always default to allowing the runtime to exit
-  // otherwise threads never exit and are not joinable.
-#if PTHREADS
-  $keepRuntimeAlive: () => !ENVIRONMENT_IS_PTHREAD,
-#else
+  // However, for pthreads we always always need to allow the runtime to exit
```
kripken, line 2084:
```suggestion
  // However, for pthreads we always need to allow the runtime to exit.
```

kripken, line 2084: Why do we need to always allow the runtime to exit with pthreads? Is that documented somewhere?

And, if so, should we do pthreads => `EXIT_RUNTIME=1`?

sbc100: Maybe I was just wanting to be emphatic!  :)

sbc100: Because pthreads themselves need to be able to exit.    `EXIT_RUNTIME=0` means the main thread never exits but it doesn't mean that pthreads can't exit.

Each pthread has its own "runtime" that can be kept alive and stops the thread from exiting.   `EXIT_RUNTIME` has not effect (or should have no effect) on threads.

sbc100: I updated the comment.

kripken (approved):

src/library.js
kripken: Heh, these double-positives/double-negatives always remind me of the joke,
```
A teacher told the class, "English has double negatives
that cancel out: 'I don't not want to go' means I want
to go, at least somewhat. But a double positive, unlike
in other languages, is basically the same as a positive."

A student at the back of the class scoffed, "Yeah, right."
```

PR #22930, opened 2024-11-15
[embind] Fix signature for 64 bit types.

### LINKED ISSUE #22928: Emscripten bindings do not work with -sWASM_BIGINT and -sASYNCIFY
I am unable to successfully run emscripten bindings with WASM_BIGINT and ASYNCIFY
When asyncify is not used, everything works as expected.
For convenience purposes, here's a repo I created for trying this out: https://github.com/sachit-wp/asyncify-repro but I have also provided details below:

```
Thread model: posix
InstalledDir: /workspaces/asyncify-repro/emsdk/upstream/bin
```

I have a simple program like this:
```cpp
#include <cstdint>
#include <emscripten.h>
#include <emscripten/bind.h>

int64_t getBigInt() {
    return 1000000000000;
}

EMSCRIPTEN_BINDINGS(my_module) {
    emscripten::function("getBigInt", &getBigInt);
}
```

I built it with the following options:
```
-sMODULARIZE -sWASM_BIGINT=1 -lembind -sEXPORT_ES6=1 -sASYNCIFY=1
```

When I attempt to run the following code from nodejs:
```js
import module from "./build/asyncify_bigint.js"
const mymod = await module();
console.log(mymod.getBigInt());
```

I get the error:
```
RuntimeError: null function or function signature mismatch
    at wasm://wasm/0001c73a:wasm-function[98]:0x57be
    at ret.<computed> (file:///workspaces/asyncify-repro/build/asyncify_bigint.js:1679:24)
    at file:///workspaces/asyncify-repro/build/asyncify_bigint.js:649:12
    at dynCallLegacy (file:///workspaces/asyncify-repro/build/asyncify_bigint.js:2028:14)
    at dynCall (file:///workspaces/asyncify-repro/build/asyncify_bigint.js:2045:17)
    at file:///workspaces/asyncify-repro/build/asyncify_bigint.js:2049:27
    at Object.getBigInt (eval at newFunc (file:///workspaces/asyncify-repro/build/asyncify_bigint.js:1447:27), <anonymous>:6:10)
    at file:///workspaces/asyncify-repro/test.js:5:19

Node.js v23.2.0
```

brendandahl:
Long longs were incorrectly being mapped to int for the signature.

Fixes #22928


brendandahl:
I should note when BIGINT is not enabled the `long long` is not a registered type so it will fail as expected.

test/test_other.py
```diff
@@ -3296,6 +3296,14 @@ def test_embind_return_value_policy(self):
 
     self.do_runf('embind/test_return_value_policy.cpp')
 
+  @parameterized({
+    '': [[]],
+    'asyncify': [['-sASYNCIFY=1']]
+  })
+  def test_embind_bigint(self, args):
+    self.do_runf('embind/test_bigint.cpp', '1000000000000n\n-1000000000000n',
+                 emcc_args=['-lembind', '-sWASM_BIGINT'] + args)
```
sbc100, line 3305: Wow I guess we just never had anyone try this before?

Why does this require `-sWASM_BIGINT` to expose the issue?  It looks like that lack of a `long long ` mapping is a problem regardless of this setting?

test/test_other.py
```diff
@@ -3296,6 +3296,14 @@ def test_embind_return_value_policy(self):
 
     self.do_runf('embind/test_return_value_policy.cpp')
 
+  @parameterized({
+    '': [[]],
+    'asyncify': [['-sASYNCIFY=1']]
+  })
+  def test_embind_bigint(self, args):
+    self.do_runf('embind/test_bigint.cpp', '1000000000000n\n-1000000000000n',
```
sbc100, line 3304: Is this new file missing from this PR?

Can you call it `test_embind_bigint` to match the test name?

Should it be called `test_embind_int64` perhaps?  Or `test_embind_long_long`?

test/test_other.py
brendandahl: It seems only ASYNCIFY exposed this because it uses the signature to create a DYNCALL with the legacy mode. Without asyncify bigints were working fine.

test/embind/test_embind_long_long.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+int main() {
+  EM_ASM(
+    console.log(Module.getBigInt());
+    console.log(Module.getBigUInt());
+  );
+}
+
+EMSCRIPTEN_BINDINGS(my_module) {
+  emscripten::function("getBigInt", &getBigInt);
```
sbc100, line 21: How about `getUint64`?  Or `getLongLong`

sbc100 (approved):
lgtm % commentg

Issue #22941, opened 2024-11-15
File "C:\emsdk-main\upstream\emscripten\tools\cache.py", line 68, in lock     acquire_cache_lock(reason)


Benman2785:
latest emscripten installed today
VS2022 installed today

commands for compiling:
> emcmake cmake -S. -Bbuild-em -DCMAKE_BUILD_TYPE=Release
> embuilder build sdl2 sdl2_ttf sdl2_image sdl2_image_jpg sdl2_image_png
> cmake --build build-em -j $(getconf _NPROCESSORS_ONLN)

File "C:\emsdk-main\upstream\emscripten\tools\cache.py", line 68, in lock
    acquire_cache_lock(reason)
File "C:\emsdk-main\upstream\emscripten\tools\cache.py", line 42, in acquire_cache_lock
    assert 'EM_CACHE_IS_LOCKED' not in os.environ, f'attempt to lock the cache while a parent process is holding the lock ({reason})'
AssertionError: attempt to lock the cache while a parent process is holding the lock (sysroot\lib\wasm32-emscripten\thinlto\libSDL2.a)

googled it and found same error in 2020 and 2023 -> all said its fixed on windows -> but it isnt

actually first time i didnt use 
> embuilder build sdl2 sdl2_ttf sdl2_image sdl2_image_jpg sdl2_image_png

but i restarted pc

full log: [emsdk-err.log](https://github.com/user-attachments/files/17780144/emsdk-err.log)

sbc100:
Which version of emsdk are you using?

I can't seem to reproduce this running `./embuilder build sdl2 sdl2_ttf sdl2_image sdl2_image_jpg sdl2_image_png` locally:

```
$ ./embuilder build sdl2 sdl2_ttf sdl2_image sdl2_image_jpg sdl2_image_png --force
embuilder:INFO: building sdl2
cache:INFO: deleting cached file: /usr/local/google/home/sbc/dev/wasm/emscripten/cache/sysroot/lib/wasm32-emscripten/libSDL2.a
cache:INFO: generating port: sysroot/lib/wasm32-emscripten/libSDL2.a... (this will be cached in "/usr/local/google/home/sbc/dev/wasm/emscripten/cache/sysroot/lib/wasm32-emscripten/libSDL2.a" for subsequent builds)
system_libs:INFO: compiled 118 inputs in 2.18s
cache:INFO:  - ok
embuilder:INFO: ...success. Took (2.31s)
[...]

Issue #20204, opened 2023-09-07
Compilation error with -sUSE_SDL_TTF=2 and -sMAIN_MODULE=2


hedwigz:
**Version of emscripten/emsdk:**
3.1.41

**Failing command line in full:**
```
$ emcc -pthread -sUSE_SDL=2 -sUSE_SDL_TTF=2 -sMAIN_MODULE=2 main.c
shared:INFO: (Emscripten: Running sanity checks)
cache:INFO: generating system headers: sysroot_install.stamp... (this will be cached in "/Users/amit/dev/rdp-wasm-core/vendor/emsdk/upstream/emscripten/cache/sysroot_install.stamp" for subsequent builds)
cache:INFO:  - ok
ports:INFO: retrieving port: freetype from https://github.com/emscripten-ports/FreeType/archive/version_1.zip
ports:INFO: unpacking port: freetype
cache:INFO: generating port: sysroot/lib/wasm32-emscripten/pic/libfreetype.a... (this will be cached in "/Users/amit/dev/rdp-wasm-core/vendor/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/pic/libfreetype.a" for subsequent builds)
system_libs:INFO: compiled 52 inputs
cache:INFO:  - ok
ports:INFO: retrieving port: sdl2 from https://github.com/libsdl-org/SDL/archive/release-2.24.2.zip
ports:INFO: unpacking port: sdl2
cache:INFO: generating port: sysroot/lib/wasm32-emscripten/pic/libSDL2-mt.a... (this will be cached in "/Users/amit/dev/rdp-wasm-core/vendor/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/pic/libSDL2-mt.a" for subsequent builds)
system_libs:INFO: compiled 115 inputs
cache:INFO:  - ok
ports:INFO: retrieving port: harfbuzz from https://github.com/harfbuzz/harfbuzz/releases/download/3.2.0/harfbuzz-3.2.0.tar.xz
ports:INFO: unpacking port: harfbuzz
cache:INFO: generating port: sysroot/lib/wasm32-emscripten/pic/libharfbuzz-mt.a... (this will be cached in "/Users/amit/dev/rdp-wasm-core/vendor/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/pic/libharfbuzz-mt.a" for subsequent builds)
[...]

PR #22946, opened 2024-11-16
Add multi-threaded versions of sdl2_image and sdl2_ttf

sbc100:
Because some possible dependencies of these libraries have `-mt` variants (specifically png and harfbuzz) we also needs to declare `-mt` variant of these libraries.

The reason for this is a little complex: When we find the set of transitive dependencies of given library we select the correct variant of each library.  This means that if we are building with `-pthread` then we select and include the `-mt` variants of all dependencies. However if libsdl2_image or libsdl2_ttf themselves then need to be built we end up building them without `-pthread` which means that emcc subprocesses will try to select and build the single threaded variants of the dependencies.  This is mostly a serious problem because we don't allow for nested calls to emcc (we assume all dependencies have been already built before we try to build a given library and the we error out with `EM_CACHE_IS_LOCKED` if that is not the case).

Testing this fix requires the cache to be setup just right so I'm not sure its worth it.

Fixes: #22941, #20204

### DIFF
--- a/test/test_sanity.py
+++ b/test/test_sanity.py
@@ -752,7 +752,7 @@ fi
   def test_embuilder_with_use_port_syntax(self):
     restore_and_set_up()
     self.run_process([EMBUILDER, 'build', 'sdl2_image:formats=png,jpg', '--force'])
-    self.assertExists(os.path.join(config.CACHE, 'sysroot', 'lib', 'wasm32-emscripten', 'libSDL2_image_jpg-png.a'))
+    self.assertExists(os.path.join(config.CACHE, 'sysroot', 'lib', 'wasm32-emscripten', 'libSDL2_image-jpg-png.a'))
     self.assertContained('error building port `sdl2_image:formats=invalid` | invalid is not a supported format', self.do([EMBUILDER, 'build', 'sdl2_image:formats=invalid', '--force']))
 
   def test_embuilder_external_ports(self):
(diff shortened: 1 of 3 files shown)


test/test_sanity.py
```diff
@@ -752,7 +752,7 @@ def test_embuilder_wildcards(self):
   def test_embuilder_with_use_port_syntax(self):
     restore_and_set_up()
     self.run_process([EMBUILDER, 'build', 'sdl2_image:formats=png,jpg', '--force'])
-    self.assertExists(os.path.join(config.CACHE, 'sysroot', 'lib', 'wasm32-emscripten', 'libSDL2_image_jpg-png.a'))
+    self.assertExists(os.path.join(config.CACHE, 'sysroot', 'lib', 'wasm32-emscripten', 'libSDL2_image-jpg-png.a'))
```
kripken, line 755: Is the `_ => -` change needed, or for aesthetics? If it isn't needed then maybe it isn't worth it, as users might have hardcoded CI commands about it.

kripken (approved):
lgtm % question

test/test_sanity.py
sbc100: Tend to use `-` for all our variants (not `_`) so this is just for consistency.

The precise names of these variants should not be hardcoded anywhere.. if it were I think that would be a bug.

kripken: I just worry about a user that has
```
cp libSDL2_image_jpg-png.a $OUT_DIR
```
in their CI workflow. This change would make them need to rename.

But if you are confident that is unlikely, and we haven't had issues with such changes in the past, sgtm as is.

Issue #22943, opened 2024-11-15
Type error in `emscripten_webgl_do_create_context`


JoeOsborn:
If client code asks for explicit swap control, a transferred canvas record is obtained from `GL.offscreenCanvases`, but then that record (and not the underlying OffscreenCanvas) is passed through to `GL.createContext`, which expects an actual offscreen canvas.  The fix is to add `....offscreenCanvas || ....canvas` on the end of line 174, or else to change the code of `emscripten_webgl_do_create_context` to unpack the canvas.

https://github.com/emscripten-core/emscripten/blob/5203d0805881ac4daf841e08cb320b1885b01cfb/src/library_html5_webgl.js#L174

JoeOsborn:
`findCanvasEventTarget` has a similar bug, in that it will sometimes return a key from GL.offscreenCanvases, sometimes return an element from GL.offscreenCanvases, and sometimes return an arbitrary HTML element.  It should always return something of a reasonable type:

```
  $findCanvasEventTarget: (target) => {
    target = maybeCStringToJsString(target);
    return (GL.offscreenCanvases[target.substr(1)] && GL.offscreenCanvases[target.substr(1)].canvas) // Remove '#' prefix
     || specialHTMLTargets[target]
     || (target == 'canvas' && Object.keys(GL.offscreenCanvases) && GL.offscreenCanvases[0].canvas)
#if PTHREADS
     || (typeof document != 'undefined' && document.querySelector(target));
#else
     || document.querySelector(target);
#endif
```

A similar fix is probably due for the deprecated lookup behavior version.

sbc100:
Hmm, so in that case it seems like that `GL.createContext` call will always fail if `explicitSwapControl` is requested and `transferControlToOffscreen` exists (i.e. offscreen canvas is supported). 

I guess that means we don't have any testing of this configuration?

Would you have time to send a PR for that?    It seems like the fix would be to add `.canvas` to the end of line 174 (e.g. `canvas = GL.offscreenCanvases[canvas.id].canvas`?

JoeOsborn:
[...]

PR #22958, opened 2024-11-19
Extract the canvas from the offscreenCanvases entry

JoeOsborn:
Fixes type error in #22943 when creating a context with explicit swap control.


src/library_html5_webgl.js
```diff
@@ -171,7 +171,7 @@ var LibraryHtml5WebGL = {
 #endif
           return 0;
         }
-        canvas = GL.offscreenCanvases[canvas.id];
+        canvas = GL.offscreenCanvases[canvas.id].canvas;
```
sbc100, line 174: Do you think we can add a test for this?

JoeOsborn:
I added a test that fails without the patch and passes with the patch.

sbc100 (approved):
Nice!

I guess this feature simply never worked?  Or maybe it worked at some point in history and bitrotted?  (@juj WDYT?)

Thanks for working on this!

test/browser/webgl_create_context_swapcontrol.c
```diff
@@ -0,0 +1,27 @@
+// Copyright 2016 The Emscripten Authors.  All rights reserved.
```
sbc100, line 1: 2024

sbc100 (approved):

sbc100:
It looks like this code was added back in 1b278917cd325c2aac877e4d09d1a8af07686c9f.

It looks like in that original patch this code would have worked because `offscreenCanvases` was `DOM ID -> OffscreenCanvas mappings of <canvas> elements that have their rendering control transferred to offscreen`.   i.e. the values in the map were canvas elements.

It looks like the breaking change was in 736565ee2827280c3f69126b1151e7b8fb3f947e where the values in `offscreenCanvases` changed to be custom JS object.   So I guess this combination of settings was been broken since 2019?

sbc100 (approved):
lgtm

What do you think @juj?

test/browser/webgl_create_context_swapcontrol.c
```diff
@@ -0,0 +1,27 @@
...
+#include <emscripten/emscripten.h>
+#include <emscripten/html5.h>
+
+int main() {
+  EmscriptenWebGLContextAttributes attrs;
+  emscripten_webgl_init_context_attributes(&attrs);
+  attrs.explicitSwapControl = true;
+  #ifdef USE_OFFSCREEN_FRAMEBUFFER
+  attrs.renderViaOffscreenBackBuffer = true;
+  #endif
```
kripken, line 19:
```suggestion
#ifdef USE_OFFSCREEN_FRAMEBUFFER
  attrs.renderViaOffscreenBackBuffer = true;
#endif
```

kripken: (to match our general coding style)

kripken (approved):
Thanks @JoeOsborn !

test/test_browser.py
```diff
@@ -2612,6 +2612,15 @@ def test_html5_webgl_create_context(self, args):
   def test_html5_webgl_create_context2(self):
     self.btest_exit('webgl_create_context2.cpp')
 
+  @requires_graphics_hardware
+  # Verify bug https://github.com/emscripten-core/emscripten/issues/22943: creating a WebGL context with explicit swap control and offscreenCanvas
+  @parameterized({
+    'offscreencanvas': (['-sOFFSCREENCANVAS_SUPPORT'],),
+    'offscreenframebuffer': (['-sOFFSCREEN_FRAMEBUFFER', '-DUSE_OFFSCREEN_FRAMEBUFFER'],),
+    })
```
kripken, line 2620:
```suggestion
  })
```

sbc100: Flake8 doesn't like this `})`.  It should be indented to match the @parameterized above.  

Sorry for all the back a forth, I think this will be good to go after this fix.

kripken: (I fixed this now)

JoeOsborn:
Thank you!

kripken:
Hmm, the code size failures here seem unrelated. @sbc100 was there a recent update there?

sbc100:
> Hmm, the code size failures here seem unrelated. @sbc100 was there a recent update there?

Yup, I just updated the expectation.  Rebasing should fix those failures.

PR #22959, opened 2024-11-19
Fallback findCanvasEventTarget to findEventTarget

### LINKED ISSUE #22942: findCanvasEventTarget doesn't use specialHTMLTargets

```
Thread model: posix
InstalledDir: /home/jcoa2018/.local/src/emsdk/upstream/bin
```

This line in `library_html5.js`:

https://github.com/emscripten-core/emscripten/blob/f0cc3d004bcbc29a4a5421d54e884d4c42b05218/src/library_html5.js#L366

Seems like it should fall back to findEventTarget, not manual document querying/ID lookup.  At a minimum, it should check `specialHTMLTargets[target]` as well as the other options.

Right now, if you follow the docs and call your canvas e.g. "!canvas" in specialHTMLTargets, createContext seemingly won't be able to find it.

JoeOsborn:
Also fix canvas tag lookup bug, which would return a string rather than an offscreenCanvases entry.

Fixes #22942 , seemingly with no other patches needed; call sites already have to handle a variety of object shapes out of this function.


src/library_html5.js
```diff
@@ -362,13 +362,9 @@ var LibraryHTML5 = {
     return GL.offscreenCanvases[target.substr(1)] // Remove '#' prefix
     // If not found, if one is querying by using DOM tag name selector 'canvas', grab the first
     // OffscreenCanvas that we can find.
-     || (target == 'canvas' && Object.keys(GL.offscreenCanvases)[0])
+     || (target == 'canvas' && Object.keys(GL.offscreenCanvases) && GL.offscreenCanvases[Object.keys(GL.offScreenCanvases)[0]])
```
sbc100, line 365: This line looks like it might be more complex that it needs to be.  

About just `target == 'canvas' && Object.values(GL.offscreenCanvases)[0]`?

Can we make this a separate PR?    Ideally with a test.  I guess we have very little testing for this since the existing code seems simply broken?

JoeOsborn: I agree with that change and will push it here, but I think this whole lookup thing is pretty hacked together/"best effort".

JoeOsborn:
As for a test which shows the bug, I guess there are two bugs here---one is that there's no use of specialEventTargets, and the other is that tag lookups are broken.  I'll prepare one parameterized test for findCanvasEventTarget, varying on whether it's an ID, a tag, or a special target, both before and after an offscreen canvas has been created.

JoeOsborn:
I've added a set of tests that fails when the old function definition is in place, and passes with the new definition.

src/library_html5.js
```diff
@@ -362,13 +362,9 @@ var LibraryHTML5 = {
...
     // If not found, if one is querying by using DOM tag name selector 'canvas', grab the first
     // OffscreenCanvas that we can find.
-     || (target == 'canvas' && Object.keys(GL.offscreenCanvases)[0])
-    // If that is not found either, query via the regular DOM selector.
-#if PTHREADS
-     || (typeof document != 'undefined' && document.querySelector(target));
-#else
-     || document.querySelector(target);
-#endif
+     || (target == 'canvas' && Object.values(GL.offscreenCanvases)[0])
```
sbc100, line 365: Is this also missing the `.canvas` similar to in #22958

test/browser/html5_special_canvas_event_targets.cpp
```diff
@@ -0,0 +1,36 @@
...
+// found in the LICENSE file.
+
+#include <GLES2/gl2.h>
+#include <GLES2/gl2ext.h>
+#include <assert.h>
+
+#include <emscripten/emscripten.h>
+#include <emscripten/html5.h>
+
+const char *TARGETS[]={"!foovas", "#canvas", "canvas"};
```
sbc100, line 13: Spaces around `=`?

test/browser/html5_special_canvas_event_targets.cpp
```diff
@@ -0,0 +1,36 @@
...
+const char *target = TARGETS[WHICH_TARGET];
```
sbc100, line 14: Why not just do `-DTARGET="!foovas"` on the command line rather than use this index/indirection?

test/test_browser.py
```diff
@@ -2618,6 +2618,18 @@ def test_html5_webgl_create_context2(self):
...
     self.btest_exit('html5_special_event_targets.cpp', args=['-lGL'])
 
+  @parameterized({
+    '': ([],),
+    'offscreen_no_pthread': (['-sOFFSCREENCANVAS_SUPPORT'],),
+    'offscreen_pthread': (['-sOFFSCREENCANVAS_SUPPORT', '-sPROXY_TO_PTHREAD', '-pthread'],),
+  })
+  @requires_graphics_hardware
+  # Verify bug https://github.com/emscripten-core/emscripten/issues/22942: findCanvasEventTarget doesn't use specialHTMLTargets
+  def test_html5_special_canvas_event_targets(self, args):
```
sbc100, line 2628: You can do `self.emcc_args += ['-lGL', '-sDISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=1'] + args` here to avoid duplication on the following 3 lines

src/library_html5.js
JoeOsborn: I was worried it would, but in fact we want the whole offscreenCanvases entry because the caller might need the shared canvas pointer.   All of the callers do a check for the presence of an .offscreenCanvas field.

test/browser/html5_special_canvas_event_targets.cpp
JoeOsborn: Good point, I'll delete it.  I was worried about escaping but since there are no spaces it's not so bad.

test/test_browser.py
JoeOsborn: Cool, good to know

test/browser/html5_special_canvas_event_targets.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+int main()
+{
```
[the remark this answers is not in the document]
JoeOsborn: Fixed

sbc100 (approved):
Seems correct to.   We should probably wait on @juj for approval though

JoeOsborn:
I mentioned this on Discord, but we actually have a big problem here: specialHTMLTargets is likely to only have reasonable values on the main thread, since it isn't shared across workers and probably will be configured using MAIN_THREAD_EM_ASM or something (at any rate, it's possible it would be set up on one thread and used on a different thread, which has the same issue).  So a findCanvasEventTarget (or indeed a findEventTarget) won't do what we want for canvases that are referred to by specialHTMLTarget names.  Only ID lookups or tag lookups actually work.

It seemed like it was working because !canvas was the suggested special name and the ID of the objects involved has typically been #canvas, and the offscreenCanvases array is keyed by the ID of the element involved.  So stripping the first character from the ID ends up being like a pun in that case.

What's the right solution?  To key offscreenCanvases based on target, rather than canvas ID?  To ignore specialHTMLTargets completely in findCanvasEventTarget?

JoeOsborn:
Just wanted to ping here—I’m a bit stuck because I’m not sure what the right behavior should be.

JoeOsborn:
Can I get some feedback on this one?  I was stuck because the correct behavior isn't clear and I'd like a maintainer to weigh in.

sbc100:
I'm afraid this is at the limits of my understanding too.   However it does seem expected/understandable that the list of specialHTMLTargets would be thread-specific since canvases themselves are thread specific and any given thread could have a different set of them.

Is the issue that you new test is failing in a certain configuration?   Or is this more of a general concern that this is the wrong fix?

JoeOsborn:
I agree it makes sense that special targets would be thread specific, but it complicates the handoff from main to proxy thread (on screen to offscreen canvas) in a proxy to pthread situation. So the options seem to be:

1. Use the deprecated lookup strategy for findCanvasEventTarget, where #canvas means something special and not the html element with ID `canvas`
2. Proxy accesses to specialHTMLTargets to the main thread, or proxy all calls to find*EventTarget to the main thread; but then we can’t easily get the offscreencanvas after it’s been transferred, right?
3. Make the offscreenCanvases dictionary keyed on target selector strings rather than IDs.  This means selectors besides IDs or !IDs won’t always do the right thing in that two queries that should produce the same canvas/offscreenCanvas might not—probably better to require that canvases have either an ID or are in the specialHTMLTargets, so that an identifier or !identifier will uniquely identify the canvas.
4. Eliminate findCanvasEventTarget and replace its uses with either the regular findEventTarget or a function to get a specific offscreencanvas.
5. Return offscreen canvases as numeric handles or something so that client code is responsible for maintaining references to them, and we don’t need findCanvasEventtTarget anymore?


Other ideas I’m not thinking of?  It’s been a minute since I was deep into this bug.

sbc100:
> I agree it makes sense that special targets would be thread specific, but it complicates the handoff from main to proxy thread (on screen to offscreen canvas) in a proxy to pthread situation.

Just to be clear, is this something that doesn't work today?   And never has worked?

JoeOsborn:
Yes, I believe it only worked if you used specialHTMLTarget[“!canvas”] specifically because canvas is also an element name and likely because #canvas is special cased in the lookup code.  The offscreencanvas would be keyed by “canvas” in the offscreenCanvases map, and by luck it would be retrieved by !canvas (special target) after the ! was trimmed off and produced “canvas”, an element name selector, since the specialHTMLTargets map would be empty in all threads but the main thread.

PR #22972, opened 2024-11-20
Coding style in library_wasm_worker. NFC

sbc100:
I did this by hand rather than using clang-format.  Its just the indentation and the curly braces on the same line.

Best viewed with "ignore whitespace".


brendandahl (approved):
Aside: it'd be nice if we clang-formatted all our C/C++ code and required it going forward.

sbc100:
> Aside: it'd be nice if we clang-formatted all our C/C++ code and required it going forward.

Agreed.  Although its a little tricky because some of our code is either implicitly or explicitly part of musl which has its own style.  Or some other upstream project.

So we need to way to declare specifically which stuff is out own internal code.

PR #22976, opened 2024-11-21
Use simple response file format in test code. NFC

sbc100:
The simple one-line-per-element format is recommended these days.

We have other tests that actually test both formats work (`test_dash_s_list_parsing` and `test_dash_s_response_file_list`)

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -7155,7 +7155,7 @@ void* operator new(size_t size) {
         return 0;
       }
     '''
-    create_file('exps', '["_main","_other_function"]')
+    create_file('exps', '_main\n_other_function\n')
 
     self.set_setting('EXPORTED_FUNCTIONS', '@exps')
     self.do_run(src, '''waka 5!''')
@@ -7170,12 +7170,12 @@ void* operator new(size_t size) {
       extern "C" {
       '''
 
-    js_funcs = []
+    rsp_file_lines = []
     num_exports = 5000
     count = 0
     while count < num_exports:
       src += 'int exported_func_from_response_file_%d () { return %d;}\n' % (count, count)
-      js_funcs.append('_exported_func_from_response_file_%d' % count)
+      rsp_file_lines.append('_exported_func_from_response_file_%d' % count)
       count += 1
 
     src += r'''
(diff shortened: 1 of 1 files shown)


test/test_core.py
```diff
@@ -8193,8 +8193,8 @@ def test_asyncify_unused(self):
     'onlylist_b': (['-sASYNCIFY_ONLY=["main","__original_main","foo(int, double)","baz()","c_baz","Structy::funcy()"]'], True),
     'onlylist_c': (['-sASYNCIFY_ONLY=["main","__original_main","foo(int, double)","baz()","c_baz"]'], False),
     'onlylist_d': (['-sASYNCIFY_ONLY=["foo(int, double)","baz()","c_baz","Structy::funcy()"]'], False),
-    'onlylist_b_response': ([], True,  '["main","__original_main","foo(int, double)","baz()","c_baz","Structy::funcy()"]'),
-    'onlylist_c_response': ([], False, '["main","__original_main","foo(int, double)","baz()","c_baz"]'),
+    'onlylist_b_response': ([], True,  'main\n__original_main\nfoo(int, double)\nbaz()\nc_baz\nStructy::funcy()\n'),
+    'onlylist_c_response': ([], False, 'main\n__original_main\nfoo(int, double)\nbaz()\nc_baz\n'),
```
kripken, line 8197: In this case, isn't the original clearer actually? It avoids `\nc` and such, where the reader needs to think how to separate into tokens.

And isn't it good to keep testing for the older format, or do we have that testing elsewhere?

sbc100: See the PR description (I updated it).  We have plenty of testing for old/new format.

kripken: Sorry I missed that. Then I still think this would be clearer the other way, but I don't feel strongly.

kripken (approved):

test/test_core.py
sbc100: > In this case, isn't the original clearer actually? It avoids `\nc` and such, where the reader needs to think how to separate into tokens.

Its easier to read here in the python, but if you look at the test directory after the test runs and inspect the resp file, the one-element-per-line format is way easier to read.

I'd like to one day deprecate the old rsp json format.. (so not using outside of `test_dash_s_list_parsing` is a good first step)

sbc100: Its a little complex here because this python test file is trying to to encode the contetns of a file into a single-line-string in python, which will always look at little odd.

PR #22977, opened 2024-11-21
Add new simple export mechanism

sbc100:
This change adds a new `-sEXPORT` setting that can by used to export any type of symbols (see #8380)

It also includes a new `-sMANGLED_SYMBOLS` setting which default to true in order to not break backwards compatibility.

Both these new settings are currently experimental and using `-sEXPORT` currently disables `-sMANGLED_SYMBOLS` by default.


sbc100:
@brendandahl FYI still WIP

src/deterministic.js
```diff
@@ -26,7 +26,7 @@ Module['thisProgram'] = 'thisProgram'; // for consistency between different buil
 
 function hashMemory(id) {
   var ret = 0;
-  var len = _sbrk(0);
+  var len = sbrk(0);
```
kripken, line 29: How does this work? I was thinking that `MANGLED_SYMBOLS` controls whether we have `_sbrk` or `sbrk`, so how can this code have only one of the two forms?

sbc100: The idea is that internally the unmangled symbols are always available and ever present.

If somebody wants unmangled symbols they are generated in addition.    I'm hoping DCE should take care the most of the cost of this.

If we didn't do it this way then all of these internal call sites would somehow have to be conditional, which seem like a bad idea.

kripken: I see. Yeah, that makes sense. If we verify that JS optimizations remove the duplicate, then sgtm.

test/code_size/hello_world_wasm2js.js
```diff
@@ -35,33 +35,43 @@ function e(b) {
...
+                f: function(c) {
+                    n = c | 0;
+                },
+                g: function(c) {
+                    n = c = n - (c | 0) & -16;
+                    return c | 0;
+                },
+                h: function() {
+                    return n | 0;
+                }
```
kripken, line 52: These three new functions are maybe worth looking into. The first and last should be getting inlined, surely.

Perhaps we can add a tiny Acorn pass to deduplicate the mangled/unmangled globals, if they are the cause of this, and if Closure is not good enough to do it itself.

test/core/test_em_async_js.c
```diff
@@ -22,17 +22,19 @@ EM_ASYNC_JS(double, foo, (int timeout), {
     () => {
       // Do a simple call back into compiled code synchronously. This is
       // allowed (an async one would not).
-      _inc_result();
+      inc_result();
       // Do the same with ccall.
+      out("XXX");
       ccall("inc_result");
+      out("YYY");
```
kripken, line 29: Temp debugging?

sbc100: Sorry, yes, this is still a draft at this point

kripken: Oh ok, sorry.

Btw, given the size of the PR, if you can avoid force-pushes over time that would be helpful.

tools/webidl_binder.py
```diff
@@ -59,6 +59,8 @@ def getExtendedAttribute(self, _name):
 parser = argparse.ArgumentParser()
 parser.add_argument('--wasm64', action='store_true', default=False,
                     help='Build for wasm64')
+parser.add_argument('--no-mangled-exports', dest='mangle_exports', action='store_false', default=True,
+                    help='Dont use mangled export names')
```
kripken, line 63:
```suggestion
                    help='Avoid use mangled export names')
```
This avoids `Dont` with an apostrophe, which is a typo.

tools/maint/gen_sig_info.py
```diff
@@ -175,7 +175,7 @@ def ignore_symbol(s, cxx):
   if s.startswith('gl') and any(s.endswith(x) for x in ('NV', 'EXT', 'WEBGL', 'ARB', 'ANGLE')):
     return True
   if s in {'__stack_base', '__memory_base', '__table_base', '__global_base', '__heap_base',
-           '__stack_pointer', '__stack_high', '__stack_low', '_load_secondary_module',
+           '__stack_pointer', '__stack_high', '__stack_low', '__load_secondary_module',
```
kripken, line 178: Why does only one of these change? Did we have different prefix rules for functions and globals?

tools/link.py
```diff
@@ -1118,8 +1137,8 @@ def phase_linker_setup(options, state, newargs):
       'emscripten_stack_get_current',
     ]
 
-    # We call one of these two functions during startup which caches the stack limits
-    # in wasm globals allowing get_base/get_free to be super fast.
+    # stackCheckInit calls one of these two functions duing startup which caches
```
kripken, line 1140:
```suggestion
    # stackCheckInit calls one of these two functions during startup which caches
```

kripken (commented):
I skimmed the critical parts. Looks very promising! Great to see this happening.

brendandahl:
Looks good so far! I imagine this will bitrot very quickly. I'm happy to look into failures next week if you want help.

sbc100:
> Looks good so far! I imagine this will bitrot very quickly. I'm happy to look into failures next week if you want help.
Looks good so far! I imagine this will bitrot very quickly. I'm happy to look into failures next week if you want help.
I think I'd like to split this up into two parts:

1. New experimental flag: `-sMANGLE_SYMBOLS=0`
3. New experimental `-sEXPORT` flag that implies `-sMANGLE_SYMBOLS=0`

PR #22982, opened 2024-11-21
[test] Rename EMTEST_SKIP_CARGO -> EMTEST_SKIP_RUST. NFC

sbc100:
See #22964

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -142,7 +142,7 @@ def also_with_wasmfs_all_backends(f):
   return metafunc
 
 
-def requires_tool(tool):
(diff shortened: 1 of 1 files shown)


dschuff (approved):
Looks like we raced with #22891, either one is fine with me.

sbc100:
> Looks like we raced with #22891, either one is fine with me.

I don't think your PR actually works because there is no tool called just "rust",  its either "cargo" or "rustc", but since the tool we use directly is "cargo" I figured its better to use that?

sbc100:
Tested locally, so landing without waiting for CI so we can get the roller running again.

PR #22983, opened 2024-11-21
Use JS string templates in library_sockfs.js. NFC

sbc100:
When trying to log the contents of a typed array the old code was doing `Array.prototype.slice.call(typed_array)` but that doesn't seem necessary.

```
$ node
> console.log(`hello: ${new Uint8Array(new ArrayBuffer(3))}`);
hello: 0,0,0
> console.log(`hello: ${Array.prototype.slice.call(new Uint8Array(new ArrayBuffer(3)))}`);
hello: 0,0,0
```

### DIFF
--- a/src/library_sockfs.js
+++ b/src/library_sockfs.js
@@ -33,12 +33,12 @@ addToLibrary({
 
 #if SOCKET_DEBUG
       // If debug is enabled register simple default logging callbacks for each Event.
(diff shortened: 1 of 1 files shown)


kripken (approved):
Maybe in the far past browsers did not log out elements of typed arrays, but did for JS arrays? Yeah, today it works.

PR #22984, opened 2024-11-21
Avoid try/catch in findEventTarget. NFC

sbc100:
Rather than using try/catch to detect when `document` is missing we explicitly check for it.  This is the same pattern used in the other version of `findEventTarget` on line 336.

This avoids casting a wide net and hiding errors in the rest of the code in this function.  Its also more consistent with the other version of `findEventTarget`.

### DIFF
--- a/src/library_html5.js
+++ b/src/library_html5.js
@@ -332,11 +332,12 @@ var LibraryHTML5 = {
     return cString > 2 ? UTF8ToString(cString) : cString;
   },
 
+  // Find a DOM element with the given ID, or null if none is found.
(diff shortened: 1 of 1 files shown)


sbc100:
I noticed this while reviewing https://github.com/emscripten-core/emscripten/pull/22959

kripken (approved):
lgtm but maybe the last paragraph in the PR description is worth a changelog entry?

sbc100:
> lgtm but maybe the last paragraph in the PR description is worth a changelog entry?

I don't think it could have any material effect.  `findEventTarget` should always be returning a DOM object.

kripken (approved):

PR #22988, opened 2024-11-21
[tsgen] Make commonjs module output compatible with tsc.

brendandahl:
This enables commonjs modules to be imported into TypeScript using the same syntax as ESM modules e.g.

`import moduleFactory from './embind_tsgen.js';`

### DIFF
--- a/test/other/embind_tsgen_main.ts
+++ b/test/other/embind_tsgen_main.ts
@@ -1,6 +1,9 @@
 // Example TS program that consumes the emscripten-generated module to to
 // illustrate how the type definitions are used and test they are workings as
 // expected.
+
+// The imported file will either be an ES module or a CommonJS module depending
+// on the test.
 import moduleFactory from './embind_tsgen.js';
 
 // Async IIFE is required for TSC with commonjs modules. This is not needed for
--- a/test/other/embind_tsgen_main.ts
+++ b/test/other/embind_tsgen_main.ts
@@ -1,30 +1,36 @@
 // Example TS program that consumes the emscripten-generated module to to
 // illustrate how the type definitions are used and test they are workings as
 // expected.
-import moduleFactory from './embind_tsgen.mjs';
+import moduleFactory from './embind_tsgen.js';
 
-const module = await moduleFactory();
+// Async IIFE is required for TSC with commonjs modules. This is not needed for
+// ESM output since top level await can be used.
+(async function() {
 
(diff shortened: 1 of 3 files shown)


tools/link.py
```diff
@@ -2491,9 +2491,11 @@ def modularize():
       src += 'export default %s;\n' % settings.EXPORT_NAME
   elif not settings.MINIMAL_RUNTIME:
     src += '''\
-if (typeof exports === 'object' && typeof module === 'object')
+if (typeof exports === 'object' && typeof module === 'object') {
   module.exports = %(EXPORT_NAME)s;
-else if (typeof define === 'function' && define['amd'])
+  // This default export allows TS to import this commonjs style module.
+  module.exports.default = %(EXPORT_NAME)s;
```
sbc100, line 2497: Do we both of these though?  It seems a little odd..

test/other/embind_tsgen_main.ts
```diff
@@ -1,30 +1,36 @@
 // Example TS program that consumes the emscripten-generated module to to
 // illustrate how the type definitions are used and test they are workings as
 // expected.
-import moduleFactory from './embind_tsgen.mjs';
+import moduleFactory from './embind_tsgen.js';
```
sbc100, line 4: Should we not be testing both of these options?

test/test_other.py
```diff
@@ -3385,21 +3385,24 @@ def test_jspi_add_function(self):
     self.do_runf('other/test_jspi_add_function.c', 'done')
 
   @parameterized({
-    '': [[]],
-    'with_jsgen': [['-sEMBIND_AOT']]
+    'commonjs': [['-sMODULARIZE'], ['--module', 'commonjs', '--moduleResolution', 'node']],
```
brendandahl, line 3388: FYI: this roughly matches how emscripten output is used in g3.

test/other/embind_tsgen_main.ts
sbc100: I see.. so this `.js` file can be either one, depending on the test?

Perhaps add a comment to that effect here?

tools/link.py
sbc100: I wonder if its worth include that link here, or elaborating on the comment.   e.g. "This might seem redundant but without this it is not possible to ...".

sbc100 (approved):

Issue #22962, opened 2024-11-19
Interactive audio worklet futex test fails


cwoffenden:
With the latest Emscripten:

```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.73-git (299be0bbfbbadfa926cb69380f0fffa3703b26c7)
```

This interactive audio worklet futex test fails:

https://github.com/emscripten-core/emscripten/blob/299be0bbfbbadfa926cb69380f0fffa3703b26c7/test/test_interactive.py#L298

Running:

```
test/runner interactive.test_audio_worklet_emscripten_futex_wake
```

It fails to compile due to `_emscripten_thread_supports_atomics_wait()` being internal to the pthread implementation:

```
test_audio_worklet_emscripten_futex_wake (test_interactive.interactive) ... Opening in existing browser session.
/Volumes/Work/Tools/Emscripten/emscripten/main/test/webaudio/audioworklet_emscripten_futex_wake.cpp:17:28: error: use of undeclared identifier '_emscripten_thread_supports_atomics_wait'
   17 |   int supportsAtomicWait = _emscripten_thread_supports_atomics_wait();
      |                            ^
1 error generated.
```

Further, the comment here:

https://github.com/emscripten-core/emscripten/blob/299be0bbfbbadfa926cb69380f0fffa3703b26c7/system/lib/pthread/emscripten_thread_state.S#L56

Is incorrect since `!ENVIRONMENT_IS_WEB` is true for Audio Worklets but `Atomics.wait` is not supported.

cwoffenden:
There's more to this. If I comment out `_emscripten_thread_supports_atomics_wait()` then the `emscripten_futex_wait()` call fails with:

```
test.js:899 Uncaught RuntimeError: Aborted(Assertion failed: emscripten_is_main_browser_thread(), at: ../../../system/lib/pthread/emscripten_futex_wait.c,26,futex_wait_main_browser_thread)
```

Because when the audio worklet code hits here:

https://github.com/emscripten-core/emscripten/blob/299be0bbfbbadfa926cb69380f0fffa3703b26c7/system/lib/pthread/emscripten_futex_wait.c#L132

As the comment says, an audio worklet will take the same path as the main thread:

```
  // For the main browser thread and audio worklets we can't use
  // __builtin_wasm_memory_atomic_wait32 so we have busy wait instead.
```

Which then fails when it arrives here:

[...]

PR #22995, opened 2024-11-23
[AUDIO_WORKLET] Fix spinlocks for audio worklets (and add tests)

cwoffenden:
Fix for #22962 and adds tests:

- ~test for atomic wait exposed as a public function~
- test renamed to match its new use
- test converted to C-only (like the other tests, plus it wasn't using C++ features)

(The `git mv` lost the old to new mapping at some point)


system/lib/wasm_worker/library_wasm_worker.c
```diff
@@ -105,15 +105,19 @@ void emscripten_lock_waitinf_acquire(emscripten_lock_t *lock) {
 }
 
 bool emscripten_lock_busyspin_wait_acquire(emscripten_lock_t *lock, double maxWaitMilliseconds) {
+  // Note: using emscripten_get_now() over emscripten_performance_now() to allow
+  // compatibility with Audio Worklets (but for workers this will defer to the
+  // performance call with its higher granularity). We keep the non-busy locks
+  // using the performance version since they have waits, incompatible with AWs.
```
sbc100, line 111: I'm not really sure we need this comment, and I think we we should probably just convert everything to `emscripten_get_now`.  

If you do go with a comment how about "Use emscripten_get_now over emscripten_performance_now since the former is available in all contexts and uses performance.now under the hood where available.".

system/lib/pthread/emscripten_thread_state.S
```diff
@@ -53,7 +53,8 @@ emscripten_is_main_browser_thread:
   global.get is_main_thread
   end_function
 
-# Semantically the same as testing "!ENVIRONMENT_IS_WEB" in JS
+# Semantically the same as testing "!ENVIRONMENT_IS_WEB" in JS (except for the
```
sbc100, line 56: How about just:

```
# Semantically the same as testing "!ENVIRONMENT_IS_WEB && !ENVIRONMENT_IS_AUDIO_WORKLET" since neither the main thread
on the web, nor audio worklets support atomics waits.
```

test/webaudio/audioworklet_emscripten_locks.cpp
```diff
@@ -1,26 +1,27 @@
...
-int futexLocation = 0;
+// Todo: add/find a way to verify atomic wait fails?
+
+emscripten_lock_t testLock = EMSCRIPTEN_LOCK_T_STATIC_INITIALIZER;
 int testSuccess = 0;
 
 bool ProcessAudio(int numInputs, const AudioSampleFrame *inputs, int numOutputs, AudioSampleFrame *outputs, int numParams, const AudioParamFrame *params, void *userData) {
-  int supportsAtomicWait = _emscripten_thread_supports_atomics_wait();
-  printf("supportsAtomicWait: %d\n", supportsAtomicWait);
-  assert(!supportsAtomicWait);
```
sbc100, line 19: Why not keep these three lines around?

cwoffenden: I don’t see `_emscripten_thread_supports_atomics_wait()` in a public API.

Non-support of wait is a given for AW global contexts, which is why I didn’t think it worth the effort of exposing the function.

system/lib/wasm_worker/library_wasm_worker.c
cwoffenden: All the other lock types used waits, so it was a limited set of functions that could work. There’s no current benefit, unless I’m missing something, to changing all uses over to plain `now`.

cwoffenden: I thought it useful to leave a comment in case someone later thinks to change it.

It does need docs adding in the web audio page to explain what locks can be used. I’ll do this, and finish the tests, in the next days.

sbc100: I guess i just don't see the point of `emscripten_performance_now` at all.  I'm not sure why it was ever added.  Why wouldn't all callsites simply prefer `emscripten_get_now`?   The works in more places (i.e. all places) and lowers the same thing when `performance.now` is available.

test/webaudio/audioworklet_emscripten_locks.cpp
cwoffenden: I thought about this, and I’ll expose the function publicly. I can think of a few use cases, sharing code between threads that can or cannot wait, for example.

cwoffenden: Made `emscripten_thread_supports_atomics_wait()` public (added with the other Emscripten threading functions).

system/lib/wasm_worker/library_wasm_worker.c
cwoffenden: I've had time to think about this: `emscripten_performance_now()` when supported is very lightweight (it maps directly to the browser's call). `emscripten_get_now()` has a bit of indirection, and for high performance code I'd prefer the option to know I'm avoiding indirection.

For this PR I've kept the change to the parts that need changing to prove out a set of lock calls work.

sbc100: FWIW emscripten_get_now() maps directly to `performance.now()` whenever `performance.now() is available.  See https://github.com/emscripten-core/emscripten/blob/7342ee809ac4bcff958d61ac327530ed4b6eaf38/src/library.js#L1393-L1396.  There is no indirection.

cwoffenden: Good to know, thanks.

tklajnscek:
@sbc100 this looks like it's good to go?

cwoffenden:
Same with #22753, I'd still like to get this merged (and unlike #22753 this fixed something broken).

system/include/emscripten/threading.h
```diff
@@ -48,6 +48,11 @@ int emscripten_is_main_runtime_thread(void);
 // this returns 1.
 int emscripten_is_main_browser_thread(void);
 
+// Return non-zero if the calling thread supports Atomic.wait (For example if called from
+// the main browser or audio worklet thread, this function will return zero since blocking
+// is not allowed there).
+int emscripten_thread_supports_atomics_wait(void);
```
sbc100, line 54: Can we avoid adding the public API here?   Its seems like keeping this internal would make this change much smaller.   

If you want to keep using it in the test code specifically you could perhaps forward declare the internal function there rather then making it a part of the public API.

If we do want to make this into a new public API then perhaps that can/should be its own standalone change.

sbc100 (commented):
Can you update the PR title?  Perhaps something like "Fix testing of emscripten locks in audio worklets".

Perhaps you could add to the description why the existing test wasn't actually working?

sbc100:
> Same with #22753, I'd still like to get this merged (and unlike #22753 this fixed something broken).

Sure this change LGTM if we split out or remove the API change (which makes this a test-only change). 

Just to be clear when you say "fixed something broken" you mean the it fixes the testing.. not an actual fix in emscripten itself?  Or am I missunderstanding?

system/include/emscripten/threading.h
cwoffenden: It’s more than the test, it reenables some form of locks for audio worklets. The test is to replace the existing one that was broken (and not being run).

More details in #22962.

sbc100: Ah I see.  Perhaps the PR title and description could be improved then.   The title sounds like this change is about testing, but there is a function change in there.   I guess "testing" is ambiguous in this context.

cwoffenden: Title and description changed (I'll make the API changes tomorrow).

sbc100: What do you think about keeping the public API constant for this change?   Do we need to make this new API public?

cwoffenden: I'm considering moving the call back to being private, as you suggested, and remove it from the test. I had a think about it:

- Audio worklets don't support atomic wait, so the result is a given
- I can't think of a use case where I would test for wait support then run another branch (real use cases, since the AW code is run in a worklet)
- We know we're in an AW from the rest of the API calls

I'll look in the morning.

sbc100: If you want you can still forward declare the private function in the test code... its up to you.

cwoffenden:
Closed to recreate a new PR (since it had too many changes to keep track and undo).

New PR in #23393.

PR #22998, opened 2024-11-25
JS file system: distinguish between atime, mtime, and ctime

hoodmane:
This improves posix compliance of the file system.

Consider the following file:
```C
#include <stdio.h>
#include <unistd.h>
#include <assert.h>
#include <fcntl.h>
#include <string.h>
#include <sys/stat.h>

static void create_file(const char *path, const char *buffer, int mode) {
  printf("creating: %s\n", path);
  int fd = open(path, O_WRONLY | O_CREAT | O_EXCL, mode);
  assert(fd >= 0);

  int err = write(fd, buffer, sizeof(char) * strlen(buffer));
  assert(err ==  (sizeof(char) * strlen(buffer)));

  close(fd);
}

int main() {
    create_file("file", "test", 0555);
    struct stat buf1;
    stat("file", &buf1);
    sleep(1);
    chmod("file", 0777);
    struct stat buf2;
    stat("file", &buf2);
    printf("delta mtime: %ld\n", buf2.st_atim.tv_sec - buf1.st_atim.tv_sec);
    unlink("file");
}
```
Compiling and running with `gcc` vs `emcc` gives different output because the chmod updated the mtime of the file in Emscripten but not in linux: 
```
$ gcc a.c
$ ./a.c
creating: file
delta mtime: 0
$ emcc a.c
$ node a.out.js
creating: file
delta mtime: 1
````

### CODE REVIEW

**Comment by hoodmane** [comment]
@sbc100 the tests assert that `ctime` changed. But `mtime` and `atime` should not change. These are all a single property in the memfs, so without recording distinct times for them we can't satisfy both requirements at once...

**Review by sbc100 — APPROVED** [review]

File: test/stat/test_chmod.c [hunk 1]
```diff
@@ -111,7 +111,7 @@ void test() {
   memset(&s, 0, sizeof s);
   stat("folder", &s);
   assert(s.st_mode == (0300 | S_IFDIR));
-  assert(s.st_ctime != lastctime);
```
sbc100, line 114: I think these assertions are valid, no?  When I run this test on native linux they all seem to pass.

sbc100: Oh I see, so linux does track ctime separately, but emscripten doesn't (at least not today).

I wonder if its worth adding separate ctime tracking to emscripten?  Otherwise I guess we should make these assertions platform dependent?

For example if we want to make this test pass with NODERAWFS then we would want these assertions still I think.

hoodmane: If we tracked atime mtime and ctime separately, that would allow us to be more posix compliant but take up a little bit more memory per file? Seems to me like it would be worth it since I'd expect the contents of files to dominate the metadata.

sbc100: Yes, I imagine its more about the static codesize/complexity overhead.   Perhaps we can make that change before we make this one?  Or as part of it?

sbc100: I created https://github.com/emscripten-core/emscripten/pull/23004 which runs this test also under noderawfs

hoodmane: Okay I added another commit that maintains separate atime, mtime, and ctime. Should we support `O_NOATIME` too?

src/library_lz4.js
```diff
@@ -95,19 +95,18 @@ addToLibrary({
...
       setattr(node, attr) {
-        if (attr.mode !== undefined) {
-          node.mode = attr.mode;
-        }
-        if (attr.timestamp !== undefined) {
-          node.timestamp = attr.timestamp;
+        for (const key of ["mode", "atime", "mtime", "ctime"]) {
+          if (attr[key]) {
+            node[key] = attr[key];
+          }
```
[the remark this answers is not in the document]
hoodmane: We also want to drop `undefined` (this simplifies the logic in `utimensat` a bit). When I thought about using `Object.assign()` I wrote this:
```js
Object.assign(
  node,
  Object.fromEntries(
    Object.entries(attr).filter(([k, v]) => ['mode', 'atime', 'mtime', 'ctime'].contains(k) && v),
  ),
);
```
I think the for loop is simpler... Maybe we could drop the "ignore undefined" condition in trade for a tiny increase in complexity in `utimensat` but I think it's about the same...

src/library_noderawfs.js
```diff
@@ -95,10 +95,10 @@ addToLibrary({
       // -1 here for atime or mtime means UTIME_OMIT was passed.  Since node
```
sbc100, line 95: I guess this comment needs updating too.

Any reason to switch from `-1` to `undefined` ?  (That later take a little more code size).

hoodmane: Well it's useful that undefined is falsey. If we're after minimizing code size, in this block I removed the two conditionals which saved ~22 characters and switched -1 to undefined, which added 14 characters so it was a net improvement. We could save 10 more characters if you want by testing `if ((atime ?? mtime) === undefined)`

hoodmane: What about using `null` that's only 4 characters.

hoodmane: Oh nevermind `if ((atime ?? mtime) === undefined)` is the conditional with `&&`, this is `||`...

sbc100 (commented):
I wonder if this warrants a Changelog entry?

hoodmane:
Probably this merits a changelog entry yes.

src/library_fs.js
```diff
@@ -165,6 +166,12 @@ FS.staticInit();
       get isDevice() {
         return FS.isChrdev(this.mode);
       }
+      get timestamp() {
+        return this.atime;
+      }
+      set timestamp(val) {
+        this.atime = this.mtime = this.ctime = val;
+      }
```
kripken, line 174: What still uses `.timestamp`?

src/tmp.js
```diff
@@ -0,0 +1,6 @@
+Object.assign(
+  node,
+  Object.fromEntries(
+    Object.entries(attr).filter(([k, v]) => ['mode', 'atime', 'mtime', 'ctime'].contains(k) && v),
+  ),
+);
```
kripken, line 6: Was this added intentionally?

hoodmane: Nope!

src/library_fs.js
hoodmane: Well I didn't change `library_idbfs` because I didn't understand what was going on, so I put this in to allow it to keep working as-is. But because we have this setter there are several places in the new code where instead of writing
```js
node.atime = node.mtime = node.ctime = val
```
I wrote
```js
node.timestamp = val
```

kripken: I believe `library_idbfs.js` has its own `timestamp` field on its own objects, but it doesn't use that property name on FS objects. So I think we can remove this getter/setter pair. If I am wrong we should see the IDBFS tests error.

ChangeLog.md
```diff
@@ -24,6 +24,9 @@ See docs/process.md for more on how version tagging works.
 3.1.73 - 11/28/24
 -----------------
 - libunwind was updated to LLVM 19.1.4. (#22394)
+- mimalloc was updated to 2.1.7. (#21548)
```
sbc100, line 27: I wonder how this line ended up here?

sbc100 (approved):
Nice!   Thanks for working on this.

ChangeLog.md
hoodmane: Yeah I managed to remove it in #23017 and add it here...

PR #23000, opened 2024-11-25
Make readlink system call not resolve the link

### LINKED ISSUE #22999: Readlink incorrectly returns fully qualified path
Considering the following:
```C
#include <fcntl.h>
#include <string.h>
#include <stdio.h>
#include <assert.h>
#include <unistd.h>

static void create_file(const char *path, const char *buffer, int mode) {
  int fd = open(path, O_WRONLY | O_CREAT | O_EXCL, mode);
  assert(fd >= 0);

  int err = write(fd, buffer, sizeof(char) * strlen(buffer));
  assert(err ==  (sizeof(char) * strlen(buffer)));

  close(fd);
}

int main() {
    char buf[20];

    create_file("symlink_target", "xyz", 0777);
    symlink("symlink_target", "symlink_src");
    readlink("symlink_src", buf, 20);
    printf("buf: '%s'\n", buf);
    unlink("symlink_src");
    unlink("symlink_target");
}
```
Running with `gcc` we just the contents of the link vs running with `emcc` we get a fully qualified path:
```
$ gcc a.c
$ ./a.out
buf: 'symlink_target'
$ emcc a.c
$ node a.out.js
buf: '/symlink_target'
```
Note the initial `/` in the emcc run.

hoodmane:
Readlink in linux does not resolve the link, it returns the exact link contents. Resolves #22999.

### DIFF
--- a/src/library_fs.js
+++ b/src/library_fs.js
@@ -895,7 +895,7 @@ FS.staticInit();
       }
 #endif
     },
-    readlink(path) {
+    readlink(path, opts = {}) {
       var lookup = FS.lookupPath(path);
       var link = lookup.node;
       if (!link) {
@@ -904,7 +904,11 @@ FS.staticInit();
       if (!link.node_ops.readlink) {
         throw new FS.ErrnoError({{{ cDefs.EINVAL }}});
       }
-      return PATH_FS.resolve(FS.getPath(link.parent), link.node_ops.readlink(link));
+      const target = link.node_ops.readlink(link);
+      if (opts.noResolve) {
+        return target;
+      }
+      return PATH_FS.resolve(FS.getPath(link.parent), target);
     },
     stat(path, dontFollow) {
       var lookup = FS.lookupPath(path, { follow: !dontFollow });
--- a/src/library_fs.js
+++ b/src/library_fs.js
@@ -904,8 +904,7 @@ FS.staticInit();
       if (!link.node_ops.readlink) {
         throw new FS.ErrnoError({{{ cDefs.EINVAL }}});
       }
-      var target = link.node_ops.readlink(link);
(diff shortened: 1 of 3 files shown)


hoodmane:
@sbc100 test failure looks like it may be a flake?

src/library_fs.js
```diff
@@ -904,7 +904,11 @@ FS.staticInit();
       if (!link.node_ops.readlink) {
         throw new FS.ErrnoError({{{ cDefs.EINVAL }}});
       }
-      return PATH_FS.resolve(FS.getPath(link.parent), link.node_ops.readlink(link));
+      const target = link.node_ops.readlink(link);
```
sbc100, line 907: We tend to avoid `let` and `const` in JS library code still.   Sadly using `var` everywhere lends to better minified code size.

src/library_fs.js
```diff
@@ -895,7 +895,7 @@ FS.staticInit();
       }
 #endif
     },
-    readlink(path) {
+    readlink(path, opts = {}) {
```
sbc100, line 898: Do we need the old behaviour?  i.e. Are there other callsites beside `__syscall_readlinkat`?

sbc100 (commented):
lgtm!  great to see the test simplification!

src/library_fs.js
hoodmane: Just that and `test_fs_readlink` which would need to be updated. It's also documented public API but the documentation was wrong before and this updates the behavior to match what the documentation says.

hoodmane: Actually `test_fs_readlink` doesn't cover relative links.

sbc100 (approved):
I love how simple this fix is!   Thanks.

PR #23003, opened 2024-11-25
Remove cleanup code from stdlib tests. NFC

sbc100:
When running tests in emscripten there is no requirement to clean up after yourself.  The test framework automatically gives each test a clean directory to work in.

In fact, cleaning up after yourself like this makes debugging harder because in the failure case you can no longer see the state of the filesystem at the point of failure.

In addition, it meant that all these tests were also relying on and testing `atexit` and `signal` handling, making the tests more complex that less precise.

As part of this change I also made sure all these tests could be compiled outside of emscripten (on my desktop linux machine).

### DIFF
--- a/test/dirent/test_readdir.c
+++ b/test/dirent/test_readdir.c
@@ -40,14 +40,6 @@ void setup() {
   create_file("foobar/file.txt", "ride into the danger zone", 0666);
 }
 
-void cleanup() {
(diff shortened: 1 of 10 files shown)


kripken (approved):
There is one exception to this, the IndexedDB and other persistent storage tests in the browser. But looks like you intentionally left those unchanged, very good.

PR #23004, opened 2024-11-25
Run test_chmod under NODERAWFS. NFC

sbc100:


test/common.py
```diff
@@ -381,15 +381,35 @@ def metafunc(self, wasmfs, *args, **kwargs):
...
+      self.emcc_args += ['-DNODERAWFS']
+      self.set_setting('NODERAWFS')
+    elif wasmfs:
+      self.set_setting('WASMFS')
+      self.emcc_args.append('-DWASMFS')
+    func(self, *args, **kwargs)
+
+  parameterize(metafunc, {'': (False, False),
+                          'rawfs': (True, False),
+                          'wasm2js': (False, True,)})
```
kripken, line 409: Does WasmFS not support NODERAWFS? If not then it could be worth a TODO here to remove this decorator when that is fixed, at which point we could do `@also_with_wasmfs` on top of `@also_with_rawfs`.

kripken (approved):

test/common.py
sbc100: Oh yes it looks like it does.  I'll add another config for that here!

src/library_fs.js
```diff
@@ -936,9 +936,6 @@ FS.staticInit();
         timestamp: Date.now()
       });
     },
-    lchmod(path, mode) {
-      FS.chmod(path, mode, true);
-    },
```
kripken, line 941: Why remove this? It is part of the public API I think.

src/library_noderawfs.js
```diff
@@ -72,7 +72,15 @@ addToLibrary({
     readlink(...args) { return fs.readlinkSync(...args); },
     stat(...args) { return fs.statSync(...args); },
     lstat(...args) { return fs.lstatSync(...args); },
-    chmod(...args) { fs.chmodSync(...args); },
+    chmod(path, mode, dontFollow) {
+      if (dontFollow && fs.lstatSync(path).isSymbolicLink()) {
+        // Node (an indeed linux) does not support chmod on symlinks
```
kripken, line 77:
```suggestion
        // Node (and indeed linux) does not support chmod on symlinks
```

src/library_noderawfs.js
```diff
@@ -72,7 +72,15 @@ addToLibrary({
...
+        // https://nodejs.org/api/fs.html#fslchmodsyncpath-mode
+        throw new FS.ErrnoError({{{ cDefs.EOPNOTSUPP }}});
+      } else {
+        fs.chmodSync(path, mode);
+      }
```
kripken, line 82:
```suggestion
      }
      fs.chmodSync(path, mode);
```

src/library_fs.js
sbc100: Its not needed internally anywhere.  i.e. none of the syscalls depend on this since (they instead depend on the third argument to `FS.chmod`).

You would argue that its part of the public API, but to be fair this syscall is so unusual its very unlikely anyone is using it.  For example, its not supported by linux or by node.

sbc100: I reverted this change for now.

kripken (approved):
I'd be ok with separately considering to remove `FS.lchmod`.

PR #23010, opened 2024-11-25
Parameterize test_unistd_access. NFC

sbc100:


test/test_core.py
```diff
@@ -5916,25 +5916,24 @@ def test_sigaction_default(self, signal, exit_code, assert_identical):
 
   @no_windows('https://github.com/emscripten-core/emscripten/issues/8882')
   @requires_node
-  def test_unistd_access(self):
+  @parameterized({
+    '': (['-DMEMFS'],),
+    'nodefs': (['-DNODEFS', '-lnodefs.js'],),
+    'noderawfs': (['-DNODERAWFS', '-sNODERAWFS'],)
```
kripken, line 5922: Do we need `-l` with nodefs but not noderawfs?

test/test_core.py
```diff
@@ -5916,25 +5916,24 @@ def test_sigaction_default(self, signal, exit_code, assert_identical):
...
-          # TODO: NODEFS in WasmFS
-          continue
-        self.emcc_args += ['-sFORCE_FILESYSTEM']
-      if fs == 'NODEFS':
-        self.emcc_args += ['-lnodefs.js']
-      self.do_run_in_out_file_test('unistd/access.c')
+    self.emcc_args += args
+    if self.get_setting('WASMFS'):
+      if '-DNODEFS' in args:
+        self.skipTest('NODEFS in WasmFS')
```
kripken, line 5929: NodeFS should work in WasmFS, so hopefully this can be removed?

test/test_core.py
```diff
@@ -5916,25 +5916,24 @@ def test_sigaction_default(self, signal, exit_code, assert_identical):
...
+      self.emcc_args += ['-sFORCE_FILESYSTEM']
+
     # Node.js fs.chmod is nearly no-op on Windows
-    # TODO: NODERAWFS in WasmFS
-    if not WINDOWS and not self.get_setting('WASMFS'):
-      self.emcc_args = orig_compiler_opts + ['-DNODERAWFS']
-      self.set_setting('NODERAWFS')
-      self.do_run_in_out_file_test('unistd/access.c')
+    if '-DNODERAWFS' in args and (WINDOWS or self.get_setting('WASMFS')):
+      self.skipTest('NODERAWFS in WasmFS')
```
kripken, line 5934: This part as well can hopefully be removed?

test/test_core.py
sbc100: They both work.  I'll make them consistent.

sbc100: Actually `-sNODERAWFS` on its own doesn't seem to work right now.    So leaving as is for now.

test/test_core.py
sbc100: Removing these causes `wasmfs.test_unistd_access*` to fail, so leaving them in for now.

test/test_core.py
sbc100: Again, seems like it would require some WASMFS fixes.

kripken, line 5934:
```suggestion
      self.skipTest('NODERAWFS in WasmFS/Windows')
```
(or maybe two separate skips?)

kripken (approved):

PR #23013, opened 2024-11-26
Fix stat and chmod on windows under NODERAWFS

sbc100:


test/test_core.py
```diff
@@ -5911,25 +5911,31 @@ def test_sigaction_default(self, signal, exit_code, assert_identical):
...
-
-    # Node.js fs.chmod is nearly no-op on Windows
-    if '-DNODERAWFS' in args and WINDOWS:
-      self.skipTest('NODERAWFS on windows')
-
-    self.do_run_in_out_file_test('unistd/access.c')
+    # On windows we have slighly different output because we the same
+    # level of permissions are not available. For example, on windows
+    # its not possible have a file that is not readable, but writable.
+    # We also report all files as executable since there is no x but
```
kripken, line 5931:
```suggestion
    # We also report all files as executable since there is no x bit
```

kripken (approved):
lgtm, though I wonder if in general it is worth trying to be more portable than Node.js is. That is, we could give up on trying to do this sort of thing, and say "windows support requires code changes, like all Node.js code in general" - ?

sbc100:
> lgtm, though I wonder if in general it is worth trying to be more portable than Node.js is. That is, we could give up on trying to do this sort of thing, and say "windows support requires code changes, like all Node.js code in general" - ?

I think that is already true today.  But I guess we could take it step further?

PR #23014, opened 2024-11-26
Use default argument values in library_fs.js. NFC

sbc100:


src/library_fs.js
```diff
@@ -701,7 +699,7 @@ FS.staticInit();
     mkdev(path, mode, dev) {
       if (typeof dev == 'undefined') {
         dev = mode;
-        mode = 438 /* 0666 */;
+        mode = {{{ 0666 }}};
```
kripken, line 702: Can this not be just `0666` without the `{{{ }}}` ?

sbc100: Done

kripken (approved):

PR #23019, opened 2024-11-26
[tsgen] Ignore checking library types when generating TS definitions.

### LINKED ISSUE #22996: Broken (unrelated) ambient types in the user's node_modules can cause --emit-tsd builds to fail
The `tsc` invocation [here](https://github.com/emscripten-core/emscripten/blob/main/tools/emscripten.py#L652) seems sensitive to unrelated broken types present in the users `node_modules/@types` folder.

I believe TypeScript checks types in `node_modules/@types`, and it's not uncommon for these to be broken in some way (duplicate identifiers when multiple versions of the same package are present in a monorepo, 3rd party type incompatibilities and so on).

The issue is solved by passing the `--skipLibCheck` flag to `tsc`. Or alternatively, perhaps invoking tsc in some temporary `cwd` that is not the user's project, where we're not going to trigger random errors from npm dependencies.

brendandahl:
We don't need to validate other types when generating definitions since they may be in bad state.

Fixes #22996

### DIFF
--- a/tools/emscripten.py
+++ b/tools/emscripten.py
@@ -649,7 +649,11 @@ def create_tsd_exported_runtime_methods(metadata):
     tsc = [tsc]
   else:
     tsc = shared.get_npm_cmd('tsc')
-  cmd = tsc + ['--outFile', tsc_output_file, '--declaration', '--emitDeclarationOnly', '--allowJs', js_doc_file]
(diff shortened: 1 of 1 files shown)


brendandahl:
I attempted to create a test for this, but couldn't get it failing without messing with my own node system packages.

sbc100 (approved):

PR #23021, opened 2024-11-26
WASMFS chunked fetch backend

JoeOsborn:
This extends the existing WASMFS fetch backend ~~in two ways~~:

1. Supporting a configurable chunk size, and only fetching one chunk at a time as needed to satisfy requests
2. ~~Supporting a manifest mapping of file paths to URLs, so that the resources on the server don't need to match the layout of the virtual filesystem.~~

This passes all the existing fetchfs tests (and I added more for new features).  I would especially appreciate review of the C++ code (I'm not much of a C++ person).

On the way here, I also had to make the test http server support range requests.


src/library_fetchfs.js
```diff
@@ -5,10 +5,19 @@
...
+      var manifest = 0;
+      if (opts['manifest']) {
+        manifest = wasmfs_fetch_create_manifest();
+        Object.entries(opts['manifest']).forEach((pair) => {
+          var path = stringToUTF8OnStack(pair[0]);
+          var url = stringToUTF8OnStack(pair[1]);
+          wasmfs_fetch_add_to_manifest(path, url);
+        })
+      }
+      return _wasmfs_create_fetch_backend(stringToUTF8OnStack(opts.base_url), opts.chunkSize | 0, manifest);
```
sbc100, line 20: If you use `stringToUTF8OnStack` in a JS function you need to somehow restore the SP at the end of the function.   One simple way to do that is wrap the whole thing in `withStackSave(() => ...)`

src/library_fetchfs.js
```diff
@@ -5,10 +5,19 @@
...
 addToLibrary({
-  $FETCHFS__deps: ['$stringToUTF8OnStack', 'wasmfs_create_fetch_backend'],
+  $FETCHFS__deps: ['$stringToUTF8OnStack', 'wasmfs_create_fetch_backend', 'wasmfs_fetch_create_manifest', 'wasmfs_fetch_add_to_manifest'],
   $FETCHFS: {
     createBackend(opts) {
-      return _wasmfs_create_fetch_backend(stringToUTF8OnStack(opts.base_url));
+      var manifest = 0;
+      if (opts['manifest']) {
+        manifest = wasmfs_fetch_create_manifest();
+        Object.entries(opts['manifest']).forEach((pair) => {
```
sbc100, line 14: Can you do `([key, value])` or `([path, url])` here?

src/library_wasmfs_fetch.js
```diff
@@ -38,31 +34,71 @@ addToLibrary({
...
+          }
+          var wholeFileData = new Uint8Array(await wholeFileReq.arrayBuffer());
+          var text = new TextDecoder().decode(wholeFileData);
+          wasmFS$JSMemoryRanges[file] = {size:wholeFileData.byteLength, chunks:[wholeFileData], chunkSize:wholeFileData.byteLength};
+          return Promise.resolve();
+        }
+      }
+      var allPresent = true;
+      var i;
+      if(lastChunk * chunkSize < offset+len) {
```
sbc100, line 63: Space after `if` and `for` keywords here an elsewhere.

Spaces around `+` operator.

src/library_wasmfs_fetch.js
```diff
@@ -38,31 +34,71 @@ addToLibrary({
...
-        var buffer = await response['arrayBuffer']();
-        wasmFS$JSMemoryFiles[file] = new Uint8Array(buffer);
+        var bytes = new Uint8Array(await response['arrayBuffer']());
+        for (i = firstChunk; i < lastChunk; i++) {
+          wasmFS$JSMemoryRanges[file].chunks[i] = bytes.slice(i*chunkSize-start,(i+1)*chunkSize-start);
+        }
       } else {
         throw response;
       }
+      return Promise.resolve();
```
sbc100, line 89: Why is this line needed now but wasn't before?

src/library_wasmfs_fetch.js
```diff
@@ -38,31 +34,71 @@ addToLibrary({
...
+      var response = await fetch(url, {headers:{"Range": `bytes=${start}-${end-1}`}});
       if (response.ok) {
-        var buffer = await response['arrayBuffer']();
-        wasmFS$JSMemoryFiles[file] = new Uint8Array(buffer);
+        var bytes = new Uint8Array(await response['arrayBuffer']());
+        for (i = firstChunk; i < lastChunk; i++) {
+          wasmFS$JSMemoryRanges[file].chunks[i] = bytes.slice(i*chunkSize-start,(i+1)*chunkSize-start);
+        }
       } else {
         throw response;
```
sbc100, line 87: How about `if (!response.ok) { throw response; }` and then remove the else case here?  (i.e. early return)

src/library_wasmfs_fetch.js
```diff
@@ -73,17 +109,39 @@ addToLibrary({
...
+        }
+        return readLength;
       },
       getSize: async (file) => {
         try {
-          await getFile(file);
-        } catch (response) {}
-        return jsFileOps.getSize(file);
+          await getFileRange(file, 0, 0);
+        } catch (response) {
```
sbc100, line 141: Should this be called `error` or something other then response?  Do we really want to just ignore this error and return 0?   (I guess that is simply the existing behaviour).

system/include/emscripten/wasmfs.h
```diff
@@ -57,7 +57,10 @@ backend_t wasmfs_create_memory_backend(void);
 //
 // TODO: Add an async version of this function that will work on the main
 // thread.
-backend_t wasmfs_create_fetch_backend(const char* base_url __attribute__((nonnull)));
+backend_t wasmfs_create_fetch_backend(const char* base_url __attribute__((nonnull)), uint32_t, void *);
+
+void *wasmfs_fetch_create_manifest();
+void wasmfs_fetch_add_to_manifest(void *, const char *, const char *);
```
sbc100, line 63: Maybe name the parameters?

system/lib/wasmfs/backends/fetch_backend.cpp
```diff
@@ -60,38 +93,65 @@ class FetchDirectory : public MemoryDirectory {
...
 extern "C" {
-backend_t wasmfs_create_fetch_backend(const char* base_url) {
+  backend_t wasmfs_create_fetch_backend(const char* base_url, uint32_t chunkSize, FetchManifest *manifest) {
   // ProxyWorker cannot safely be synchronously spawned from the main browser
   // thread. See comment in thread_utils.h for more details.
   assert(!emscripten_is_main_browser_thread() &&
          "Cannot safely create fetch backend on main browser thread");
   return wasmFS.addBackend(std::make_unique<FetchBackend>(
     base_url ? base_url : "",
+    chunkSize != 0 ? chunkSize : DEFAULT_CHUNK_SIZE,
```
sbc100, line 127: Just `chunkSize ? chunkSize : DEFAULT_CHUNK_SIZE` maybe?

system/lib/wasmfs/backends/fetch_backend.cpp
```diff
@@ -60,38 +93,65 @@ class FetchDirectory : public MemoryDirectory {
...
+    manifest,
     [](backend_t backend) { _wasmfs_create_fetch_backend_js(backend); }));
-}
+  }
 
 const char* EMSCRIPTEN_KEEPALIVE _wasmfs_fetch_get_file_path(void* ptr) {
   auto* file = reinterpret_cast<wasmfs::FetchFile*>(ptr);
   return file ? file->getPath().data() : nullptr;
 }
+const char* EMSCRIPTEN_KEEPALIVE _wasmfs_fetch_get_file_url(void* ptr) {
```
sbc100, line 136: Add empty lines between top level functions

system/lib/wasmfs/backends/fetch_backend.cpp
```diff
@@ -60,38 +93,65 @@ class FetchDirectory : public MemoryDirectory {
...
+  auto* file = reinterpret_cast<wasmfs::FetchFile*>(ptr);
+  return file ? file->getURL().data() : nullptr;
+}
+uint32_t EMSCRIPTEN_KEEPALIVE _wasmfs_fetch_get_chunk_size(void* ptr) {
+  auto* file = reinterpret_cast<wasmfs::FetchFile*>(ptr);
+  return file ? file->getChunkSize() : DEFAULT_CHUNK_SIZE;
+}
+
+void *EMSCRIPTEN_KEEPALIVE wasmfs_fetch_create_manifest() {
```
sbc100, line 145: Do these need `EMSCRIPTEN_KEEPALIVE` on them?

system/lib/wasmfs/backends/fetch_backend.h
```diff
@@ -6,7 +6,6 @@
 #include "wasmfs.h"
 
 extern "C" {
-
-// See library_wasmfs_fetch.js
-void _wasmfs_create_fetch_backend_js(wasmfs::backend_t);
+  // See library_wasmfs_fetch.js
+  void _wasmfs_create_fetch_backend_js(wasmfs::backend_t);
```
sbc100, line 10: Maybe revert this file?  (We don't normally indent the contents of the extern "C" block).

test/common.py
```diff
@@ -2041,7 +2057,23 @@ def do_GET(self):
...
+          data = read_binary(path)
+          ctype = self.guess_type(path)
+          self.send_header('Content-type', ctype)
+          pieces = self.headers.get("range").split("=")[1].split("-")
+          start = int(pieces[0]) if pieces[0] != '' else 0
+          end = int(pieces[1]) if pieces[1] != '' else len(data) - 1
+          end = min(len(data) - 1, end)
+          length = end - start + 1
+          self.send_header('Content-Length', str(length))
+          self.send_header('Content-Range', "bytes " + str(start) + "-" + str(end) + "/" + str(len(data)))
```
sbc100, line 2072: Use single quotes here and elsewhere in this file.

You can also using pythong's f-strings.  e.g. `f'bytes {start} - {end} / {len(data)}'`

sbc100 (commented):
Nice!

Mostly nits, I'll let @tlively and @kripken chime in too.

src/library_wasmfs_fetch.js
JoeOsborn: I'm actually not sure, it just seemed wrong for it to sometimes return a promise and sometimes return nothing.

JoeOsborn:
Thanks, I'll make these style changes and stack saving.  I think I may have messed up something with respect to closure compiler, but I'm not sure what exactly.

system/lib/wasmfs/backends/fetch_backend.cpp
JoeOsborn: I'm not sure, why would or wouldn't it?  I just tried to mimic other extern C functions I saw nearby.

src/library_fetchfs.js
JoeOsborn: OK--the first argument to create_fetch_backend was like this before, and did not use withStackSave.  Should I wrap that call in the withStackSave function as well?

sbc100: Ah yes, that looks like an existing bug (a leak of stack space).

test/common.py
JoeOsborn: Good idea!

system/lib/wasmfs/backends/fetch_backend.h
JoeOsborn: Got it.

system/lib/wasmfs/backends/fetch_backend.cpp
JoeOsborn: I removed it.

system/lib/wasmfs/backends/fetch_backend.cpp
JoeOsborn: I think I got them all.

system/include/emscripten/wasmfs.h
JoeOsborn: Done

src/library_wasmfs_fetch.js
JoeOsborn: I called it failedResponse, I don't love throwing things that aren't errors but I didn't want to make too many changes.

src/library_wasmfs_fetch.js
JoeOsborn: I think I got them all.

src/library_fetchfs.js
JoeOsborn: Good idea.

src/library_fetchfs.js
JoeOsborn: Done.

src/library_fetchfs.js
```diff
@@ -5,11 +5,20 @@
...
-  $FETCHFS__deps: ['$stringToUTF8OnStack', 'wasmfs_create_fetch_backend'],
+  $FETCHFS__deps: ['$stringToUTF8OnStack', 'wasmfs_create_fetch_backend', 'wasmfs_fetch_create_manifest', 'wasmfs_fetch_add_to_manifest'],
   $FETCHFS: {
     createBackend(opts) {
-      return _wasmfs_create_fetch_backend(stringToUTF8OnStack(opts.base_url));
-    }
+      return withStackSave(() => {
+        var manifest = 0;
+        if (opts['manifest']) {
+          manifest = wasmfs_fetch_create_manifest();
```
kripken, line 14: This should require a `_` prefix, shouldn't it? New line 20 shows such a prefix being used for another function. Does this work without the prefix?

system/include/emscripten/wasmfs.h
```diff
@@ -57,7 +57,18 @@ backend_t wasmfs_create_memory_backend(void);
...
 // TODO: Add an async version of this function that will work on the main
 // thread.
-backend_t wasmfs_create_fetch_backend(const char* base_url __attribute__((nonnull)));
+backend_t wasmfs_create_fetch_backend(const char* base_url __attribute__((nonnull)),
+                                      uint32_t chunkSize,
+                                      void *manifest);
+
+// Create a FetchFS manifest record that can be populated and passed into
+// wasmfs_fetch_create_backend.
+void *wasmfs_fetch_create_manifest();
```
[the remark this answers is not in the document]
JoeOsborn: A manifest is a std::map<string,string> of path to URL. I think there needs to be wasmfs documentation in general, and if there’s an rst file where the usage documentation should go I’d be happy to write up docs for fetchfs once the API sketch here is approved.

src/library_fetchfs.js
JoeOsborn: Good question. Is there a good way to write unit tests for the JS side of interop? Say, a test that used JS to create a fetchfs filesystem and then run a C program that checks for he presence/contents of files in the filesystem?  I’d be happy to add a test case if you can point me to an example of a test that isn’t just C code.

system/include/emscripten/wasmfs.h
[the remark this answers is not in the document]
kripken: Hmm, the best we have for WasmFS docs for now is `site/source/docs/api_reference/Filesystem-API.rst`, but it's very minimal.

I think we can add some explanation in this header file though. It would be nice if it were self-contained so people could understand it without other docs. Though I do agree we should have those other docs too.

src/library_fetchfs.js
kripken: You can embed JS code into the C code using `EM_ASM` etc. See e.g.

https://github.com/emscripten-core/emscripten/blob/5643c2694ce08c4cfcada3529b8271ae5721e56e/test/wasmfs/wasmfs_getdents.c#L100-L108

JoeOsborn: Ok that’s helpful, and would at least tell us if the createBackend code path works.  I was wondering if we could have a test where JS code in a pre run hook can successfully set up a filesystem—is there something like em_asm that runs before main but after global initializers?

vadimkantorov:
An important functionality is to be able to consume via bare HTTP chunked data files - e.g. so that we could still use basic, non-range supporting servers. And to check that GH Pages (and maybe some other bare hostings) works for this scenario...

JoeOsborn:
That’s interesting. Should it be an option in opts, or a fallback in case we get error 416 or no accept-ranges header from the server?

vadimkantorov:
I mean I have nothing against range requests for chunked access, my point is that there should also be support for chunked data files, e.g. so that we can serve off GitHub Pages cacheable <50Mb files `myarchive.part0.data`, `myarchive.part1.data` (or even `myarchive.zip`, `myarchive.part0.zip`, ... and maybe allowing for indexes to be served in a separate files or in the main file to delay downloading the actual unneeded chunks...) -

JoeOsborn:
I think the idea of downloading multiple files into ranges of one file sounds tricky, but I do think it could make sense to map chunks onto remote files.  I’d like to land this first with just range requests and maybe down the line the manifest could be extended to support this use case.

JoeOsborn:
I have fixed some bugs, added tests for the JS API, added docs, and rebased.  I think this is ready to merge now if CI is clean, though of course I'm happy to revisit if there are code quality issues.

I'll be able to look at this again on Monday, so I might not be able to fix it right away if there are CI issues.  But I will get to them!

src/library_fetchfs.js
```diff
@@ -5,11 +5,25 @@
...
 addToLibrary({
-  $FETCHFS__deps: ['$stringToUTF8OnStack', 'wasmfs_create_fetch_backend'],
+  $FETCHFS__deps: ['$stringToUTF8OnStack', 'wasmfs_create_fetch_backend', 'wasmfs_fetch_create_manifest', 'wasmfs_fetch_add_to_manifest'],
   $FETCHFS: {
     createBackend(opts) {
-      return _wasmfs_create_fetch_backend(stringToUTF8OnStack(opts.base_url));
-    }
+      if(opts.base_url === undefined) {
+        opts.base_url = "";
+      }
```
kripken, line 13:
```suggestion
      opt.base_url ??= '';
```

src/library_wasmfs_fetch.js
```diff
@@ -35,34 +31,74 @@ addToLibrary({
         try {
           var u = new URL(fileUrl, self.location.origin);
           url = u.toString();
-        } catch (e) {
+        } catch(_e) {
```
kripken, line 34:
```suggestion
        } catch (_e) {
```

system/lib/wasmfs/backends/fetch_backend.cpp
```diff
@@ -57,40 +135,75 @@ class FetchDirectory : public MemoryDirectory {
...
+    manifest,
     [](backend_t backend) { _wasmfs_create_fetch_backend_js(backend); }));
-}
+  }
 
-const char* EMSCRIPTEN_KEEPALIVE _wasmfs_fetch_get_file_path(void* ptr) {
+const char* _wasmfs_fetch_get_file_path(void* ptr) {
   auto* file = reinterpret_cast<wasmfs::FetchFile*>(ptr);
   return file ? file->getPath().data() : nullptr;
 }
```
kripken, line 184: Is this function still needed?

src/library_wasmfs_fetch.js
```diff
@@ -35,34 +31,74 @@ addToLibrary({
...
+          throw {status: 404};
         }
       }
-      var response = await fetch(url);
-      if (response.ok) {
-        var buffer = await response['arrayBuffer']();
-        wasmFS$JSMemoryFiles[file] = new Uint8Array(buffer);
-      } else {
+      var chunkSize = __wasmfs_fetch_get_chunk_size(file);
+      offset = offset || 0;
```
kripken, line 39:
```suggestion
      offset ??= 0;
```

system/include/emscripten/wasmfs.h
```diff
@@ -48,6 +48,32 @@ typedef backend_t (*backend_constructor_t)(void*);
...
+// files at arbitrary different locations.  The manifest can be built
+// using wasmfs_fetch_create_manifest and
+// wasmfs_fetch_add_to_manifest.  When the fetchfs is mounted,
+// directories and files corresponding to the manifest entries will be
+// automatically created.
+// 
+// For example, a manifest like `{'/test.txt':'resources/file.txt'}`
+// mounted at `/dat` (using either wasmfs_mount or
+// wasmfs_create_directory) will let you write `open("/dat/test.txt",
+// O_RDONLY)`.
```
kripken, line 75: Can the user not implement manifests using symlinks? That is, use redirection locally?

JoeOsborn: It is sometimes possible, but very tedious to arrange the directories and depends on the C code following symlinks properly, doing the right thing with stat calls, etc.  I thought about it, but felt like adding the control over where downloaded files go in the first place seemed better.

kripken: I think the tedious part could be fixed with a helper somehow, but good points about some C code not following symlinks.

Still, I hope we can find a way to separate these two functionalities, of the fetch backend and a "rename/restructure" routing layer. In particular because other backends may want similar things.

One of the core ideas in WasmFS is that it is composable. For example, we have an "ignore-case" backend that ignores lowercase/uppercase (`wasmfs_create_icase_backend`) and which is layered on top of another backend. That approach lets any backend have the property of ignoring the case of filenames. Following that, what do you think about adding a "routing" backend which would receive a manifest like here, and basically map locations to other locations? Then both the fetch backend and other backends could use it.

system/lib/wasmfs/backends/fetch_backend.cpp
JoeOsborn: It is not, good catch!

system/include/emscripten/wasmfs.h
JoeOsborn: Since the manifest maps file paths to relative URLs, I don't immediately see how it's composable (the manifest could in principle map to URLs containing query parameters, for example).  Are you saying the idea is that the composable bit is mapping paths in one filesystem to paths in a different filesystem?  It sounds (to me) like a lot of work to build implementations of getChild, getEntries, etc that respect those remappings because I'm not super familiar with the wasmfs code or its invariants and not much of a C++ programmer, but if the path forward there seems obvious to you I'm happy to collaborate on it.

kripken: Yes, the idea is a mapping layer. For example, a test in this PR contains this:
```JS
FS.mount(FETCHFS, {"manifest":{"/file":"test.txt", "/sub/file":"test.txt"}}, "/dat2");
```
If we had a MAPFS that does such a mapping, we could do
```js
FS.mount(MAPFS, {"/file":"test.txt", "/sub/file":"test.txt"}, "/dat2", FETCHFS);
```
This mapping backend would see `/dat2/file` and map that to the path `test.txt`, and it would route that to the underlying (layered) FETCHFS.

(query parameters can work too, as part of the filename? but that part I'm not sure of - illegal characters might be an issue?)

> It sounds (to me) like a lot of work to build implementations of getChild, getEntries, e

It's not trivial, but it shouldn't be much more than the ignore-case backend:

https://github.com/emscripten-core/emscripten/blob/main/system/lib/wasmfs/backends/ignore_case_backend.cpp

In fact I'd guess 80% of the code could be identical. The only change would be that instead of ignoring the case of a file, we route to a possibly different path.

JoeOsborn: Yes, I can sort of see how that makes sense, but I’m not sure which other backends would benefit since they would typically have more control over the layout of files.  

I struggled in the implementation of the [manifest](https://github.com/emscripten-core/emscripten/blob/87ced784341f0a6581bae8099c87eb75d6d6c2ab/system/lib/wasmfs/backends/fetch_backend.cpp#L74) and ended up creating every file record immediately on mount/root directory creation.  I couldn’t figure out a way for a directory or file object to know its full path and to do this lazily.  This makes sense because of things like symlinks, but I found implementing getChild/getEntries pretty tricky and let the memory directory implementation do the work.

If the manifest were a nested map (string -> filepath | manifest, natural for a json object), it’s clear to see how to implement getChild and friends, but constructing such a nested map in c is just as much work as calling createDirectory or createFile a bunch of times and seems about the same as using the memorydirectory implementation.  In fact I guess a MAPFS would have to call those same functions on the target backend for the target paths, as the source path tree is being constructed, which I guess the case insensitive filesystem does too.

I can try to take a look at this over the next few days.  One last question about your hypothetical example—does WASMFS FS.mount take a fourth parameter?  According to lib/libwasmfs.js it does not.

JoeOsborn:
I wonder if I should separate out the chunking part of this PR from the manifest part, since there’s some more design and implementation to do there and it could be achieved in the short term through e.g. mounting a bunch of single-file fetchfs backends with different base URLs.  That’s less convenient but it’s easier to add more features later than change ones that were added before they were ready.

JoeOsborn:
I've force-pushed a reduced version of this patch here, with the manifest stuff taken out.  I'll introduce a new branch in a little while with the MAPFS idea.

src/lib/libfetchfs.js
```diff
@@ -8,8 +8,11 @@ addToLibrary({
   $FETCHFS__deps: ['$stringToUTF8OnStack', 'wasmfs_create_fetch_backend'],
   $FETCHFS: {
     createBackend(opts) {
-      return _wasmfs_create_fetch_backend(stringToUTF8OnStack(opts.base_url));
-    }
+      opts.base_url ??= "";
+      return withStackSave(() => {
+        return _wasmfs_create_fetch_backend(stringToUTF8OnStack(opts.base_url), opts.chunkSize | 0);
```
kripken, line 13:
```suggestion
      return withStackSave(() => {
        return _wasmfs_create_fetch_backend(stringToUTF8OnStack(opts.base_url ?? ""), opts.chunkSize | 0);
```

src/lib/libwasmfs_fetch.js
```diff
@@ -35,34 +31,74 @@ addToLibrary({
...
         }
       }
-      var response = await fetch(url);
-      if (response.ok) {
-        var buffer = await response['arrayBuffer']();
-        wasmFS$JSMemoryFiles[file] = new Uint8Array(buffer);
-      } else {
+      var chunkSize = __wasmfs_fetch_get_chunk_size(file);
+      offset ??= 0;
+      len = len || chunkSize;
```
kripken, line 40:
```suggestion
      len ??= chunkSize;
```

src/lib/libwasmfs_fetch.js
```diff
@@ -35,34 +31,74 @@ addToLibrary({
...
+      var firstChunk = (offset / chunkSize) | 0;
+      var lastChunk = ((offset+len) / chunkSize) | 0;
+      if (!(file in wasmFS$JSMemoryRanges)) {
+        var fileInfo = await fetch(url,{method:"HEAD", headers:{"Range": "bytes=0-"}});
+        if (fileInfo.ok &&
+           fileInfo.headers.has("Content-Length") &&
+           fileInfo.headers.get("Accept-Ranges") == "bytes" &&
+           (parseInt(fileInfo.headers.get("Content-Length")) > chunkSize*2)) {
+          wasmFS$JSMemoryRanges[file] = {size:parseInt(fileInfo.headers.get("Content-Length")), chunks:[], chunkSize:chunkSize};
```
kripken, line 49:
```suggestion
          wasmFS$JSMemoryRanges[file] = {
            size: parseInt(fileInfo.headers.get("Content-Length")),
            chunks: [],
            chunkSize: chunkSize
          };
```
this matches our general formatting a little better

src/lib/libwasmfs_fetch.js
```diff
@@ -35,34 +31,74 @@ addToLibrary({
...
+        } else {
+          // may as well/forced to download the whole file
+          var wholeFileReq = await fetch(url);
+          if (!wholeFileReq.ok) {
+            throw wholeFileReq;
+          }
+          var wholeFileData = new Uint8Array(await wholeFileReq.arrayBuffer());
+          var text = new TextDecoder().decode(wholeFileData);
+          wasmFS$JSMemoryRanges[file] = {size:wholeFileData.byteLength, chunks:[wholeFileData], chunkSize:wholeFileData.byteLength};
```
kripken, line 58: please format this too

system/include/emscripten/wasmfs.h
```diff
@@ -57,7 +69,9 @@ backend_t wasmfs_create_memory_backend(void);
 //
 // TODO: Add an async version of this function that will work on the main
 // thread.
-backend_t wasmfs_create_fetch_backend(const char* base_url __attribute__((nonnull)));
+//
+backend_t wasmfs_create_fetch_backend(const char* base_url __attribute__((nonnull)),
+                                      uint32_t chunkSize);
```
kripken, line 74: Please document chunkSize.

system/lib/wasmfs/backends/fetch_backend.cpp
```diff
@@ -57,40 +79,53 @@ class FetchDirectory : public MemoryDirectory {
...
-               std::function<void(backend_t)> setupOnThread)
-    : ProxiedAsyncJSBackend(setupOnThread), baseUrl(baseUrl) {}
+std::shared_ptr<Directory> FetchBackend::createDirectory(mode_t mode) {
+  return std::make_shared<FetchDirectory>("", mode, this, proxy);
+}
 
-  std::shared_ptr<DataFile> createFile(mode_t mode) override {
-    return std::make_shared<FetchFile>(baseUrl, mode, this, proxy);
+const std::string FetchBackend::getFileURL(const std::string& filePath) {
+  if(filePath == "") {
```
kripken, line 97:
```suggestion
  if (filePath == "") {
```

test/wasmfs/wasmfs_fetch.c
```diff
@@ -42,7 +43,7 @@ void check_file(int fd, const char* content) {
 void test_url_relative() {
   printf("Running %s...\n", __FUNCTION__);
 
-  backend_t backend2 = wasmfs_create_fetch_backend("test.txt");
+  backend_t backend2 = wasmfs_create_fetch_backend("test.txt",0);
```
kripken, line 46:
```suggestion
  backend_t backend2 = wasmfs_create_fetch_backend("test.txt", 0);
```
below as well

kripken (commented):
Thanks, I think this makes sense split out like this: even if we do decide that a manifest option only fits in the Fetch backend, it is good to land that in a separate PR as a later addition (but I am still optimistic about the layered approach, hopefully that works well).

This mostly looks good to me, just some minor comments.

JoeOsborn:
Thanks, I’ll take a look at this Monday!

src/lib/libwasmfs_fetch.js
```diff
@@ -73,17 +117,36 @@ addToLibrary({
...
         try {
-          await getFile(file);
-        } catch (response) {
-          return response.status === 404 ? -{{{ cDefs.ENOENT }}} : -{{{ cDefs.EBADF }}};
+          await getFileRange(file, offset || 0, length);
+        } catch (failedResponse) {
+          return failedResponse.status === 404 ? -{{{ cDefs.ENOENT }}} : -{{{ cDefs.EBADF }}};
+        }
+        var fileInfo = wasmFS$JSMemoryRanges[file];
+        var fileData = fileInfo.chunks;
```
kripken, line 125:
```suggestion
        var chunks = fileInfo.chunks;
```
It seems clearer not to rename this?

src/lib/libwasmfs_fetch.js
```diff
@@ -73,17 +117,36 @@ addToLibrary({
...
+        var chunkSize = fileInfo.chunkSize;
+        var firstChunk = (offset / chunkSize) | 0;
+        var lastChunk = ((offset+length) / chunkSize) | 0;
+        if (offset + length > lastChunk * chunkSize) {
```
kripken, line 129: Same question as before (perhaps worth a comment?)

system/include/emscripten/wasmfs.h
```diff
@@ -57,7 +74,9 @@ backend_t wasmfs_create_memory_backend(void);
 //
 // TODO: Add an async version of this function that will work on the main
 // thread.
-backend_t wasmfs_create_fetch_backend(const char* base_url __attribute__((nonnull)));
+//
+backend_t wasmfs_create_fetch_backend(const char* base_url __attribute__((nonnull)),
+                                      uint32_t chunk_size);
```
kripken, line 79: Please mention this API change in `ChangeLog.md`

test/wasmfs/wasmfs_fetch.c
```diff
@@ -150,6 +151,108 @@ void test_small_reads() {
...
 }
 
+void test_small_chunks() {
+  // Read the file in small amounts.
+  printf("Running %s...\n", __FUNCTION__);
+
+  char expected[] = "hello";
+  size_t size = 5;
+
+  backend_t backend = wasmfs_create_fetch_backend("small.dat",2);
```
kripken, line 161:
```suggestion
  backend_t backend = wasmfs_create_fetch_backend("small.dat", 2);
```

test/wasmfs/wasmfs_fetch.c
```diff
@@ -150,6 +151,108 @@ void test_small_reads() {
...
+
+  fd = wasmfs_create_file("/testfile6", 0777, backend);
+  for (size_t i = 0; i < size; i+=5) {
+    int read_now = read(fd, buf + i, 5);
+    assert(read_now <= 5);
+    printf("read some bytes much larger than chunk size\n");
+  }
+  buf[size] = 0;
+  printf("buf %s\n",buf);
+  assert(strcmp(buf, "hello") == 0);
```
kripken, line 196: How about making a helper function for these read loops, that receives the chunk size?

test/wasmfs/wasmfs_fetch.c
```diff
@@ -150,6 +151,108 @@ void test_small_reads() {
...
+    printf("read some bytes equal to chunk size\n");
+  }
+  buf[size] = 0;
+  printf("buf %s\n",buf);
+  assert(strcmp(buf, "hello") == 0);
+
+  assert(close(fd) == 0);
+
+  fd = wasmfs_create_file("/testfile6", 0777, backend);
+  for (size_t i = 0; i < size; i+=5) {
```
kripken, line 189:
```suggestion
  for (size_t i = 0; i < size; i += 5) {
```

kripken (commented):
I read this in detail now. Mostly minor comments, and I won't have any more after these.

See also the CI errors, though (it isn't done but looks like `other.test_closure_full_js_library_wasmfs` is failing).

src/lib/libwasmfs_fetch.js
JoeOsborn: Will do.

src/lib/libwasmfs_fetch.js
JoeOsborn: Good idea, I'll simplify the conditions too.

system/include/emscripten/wasmfs.h
JoeOsborn: Will do.

test/wasmfs/wasmfs_fetch.c
JoeOsborn: Good idea

JoeOsborn:
The closure test failure was due to not passing a radix to parseInt, I've resolved that too.

test/common.py
```diff
@@ -2049,23 +2049,39 @@ class TestServerHandler(SimpleHTTPRequestHandler):
...
     # SimpleHTTPRequestHandler.do_GET(self) below.
     def send_head(self):
+      path = self.translate_path(self.path)
+      try:
+        f = open(path, 'rb')
+        fs = os.fstat(f.fileno())
+      except IOError:
+        self.send_error(404, f'File not found {path}')
+        return None
       if self.path.endswith('.js'):
```
sbc100, line 2059: Any idea why `.js` files are singled out here?

test/common.py
```diff
@@ -2160,7 +2176,23 @@ def do_GET(self):
...
+          self.send_header('Content-type', ctype)
+          pieces = self.headers.get('Range').split('=')[1].split('-')
+          start = int(pieces[0]) if pieces[0] != '' else 0
+          end = int(pieces[1]) if pieces[1] != '' else len(data) - 1
+          end = min(len(data) - 1, end)
+          length = end - start + 1
+          self.send_header('Content-Length', str(length))
+          self.send_header('Content-Range', f'bytes {start}-{end}/{len(data)}')
+          self.end_headers()
+          self.wfile.write(data[start:end + 1])
```
sbc100, line 2193: So is there no support for `Range` in `SimpleHTTPRequestHandler` itself?

test/test_browser.py
```diff
@@ -5318,6 +5318,7 @@ def test_wasmfs_fetch_backend(self, args):
     create_file('subdir/backendfile2', 'file 2')
     self.btest_exit('wasmfs/wasmfs_fetch.c',
                     args=['-sWASMFS', '-pthread', '-sPROXY_TO_PTHREAD',
+                          '-sFORCE_FILESYSTEM', '-lfetchfs.js',
```
sbc100, line 5321: How did this test work before with `fetchfs.js` included?

sbc100 (commented):
Sorry, late the party with mostly nits

test/common.py
JoeOsborn: There is not.

test/test_browser.py
JoeOsborn: It didn't use the JS library before, just the WASM backend code which was included either way.

test/common.py
JoeOsborn: Not sure, it was like that before: https://github.com/emscripten-core/emscripten/blob/3084681dd8b8f834a0d9c4cc034f90c0ad4937b1/test/common.py#L2055

But I removed the special case for JS and the tests seem to work fine.

JoeOsborn:
I've addressed all of @sbc100 's comments.

test/common.py
```diff
@@ -2049,23 +2049,32 @@ class TestServerHandler(SimpleHTTPRequestHandler):
...
-        except IOError:
-          self.send_error(404, "File not found: " + path)
-          return None
-        self.send_response(200)
-        self.send_header('Content-type', 'application/javascript')
-        self.send_header('Connection', 'close')
+      path = self.translate_path(self.path)
+      try:
+        fsize = os.path.getsize(path)
+        f = open(path, 'rb')
```
sbc100, line 2055: It looks like `f` is not closed in the non-Range case below.

Is it even needed in that path?   Maybe this whole try/catch can go inside the `Range` check?

sbc100:
I think we can/should land the send_head simplification first/separately: https://github.com/emscripten-core/emscripten/pull/23609

JoeOsborn:
OK, but send_head is needed for the range case so I'll be adding back the simplified version.

test/common.py
JoeOsborn: https://github.com/python/cpython/blob/de84531a3ec5f1f4357983f2e95e0c614b2e744e/Lib/http/server.py#L683C1-L687C22 The HTTP server does close it; send_head should return an open file for do_GET to do the right thing.

JoeOsborn:
Ah dang, I hit the wrong button.  If you don't like merge commits let me know and I can try to undo it/rebase normally

sbc100:
Whatever rebase or merge strategy you like is fine for PRs.  The PR will get squashed and rebased at the end of the review either way.

sbc100:
Is the PR basically ready to land now?  Should we remove the WIP from the title?

src/library_wasmfs_fetch.js
sbc100: OK, I think we can probably replace a bunch of `return Promise.resolve()` with simply `return` (since they both end up resolving to undefined I believe), but we could do that a a followup I think.

JoeOsborn:
Indeed! It's ready to go.

src/lib/libwasmfs_fetch.js
```diff
@@ -35,34 +31,86 @@ addToLibrary({
...
+      // one request for all the chunks we need, rather than one
+      // request per chunk.
+      var start = firstChunk * chunkSize;
+      // We must fetch *up to* the last byte of the last chunk.
+      var end = (lastChunk+1) * chunkSize;
+      var response = await fetch(url, {headers:{"Range": `bytes=${start}-${end-1}`}});
+      if (!response.ok) {
         throw response;
       }
+      var bytes = new Uint8Array(await response['arrayBuffer']());
```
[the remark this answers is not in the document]
sbc100: Looks like this is still not resolved?

ChangeLog.md
```diff
@@ -31,6 +31,12 @@ See docs/process.md for more on how version tagging works.
   source maps, so it has not worked in many years, and there have been no
   requests for it. This has no impact on the source map support in browser
   devtools. (#23553)
+- The WASMFS fetch backend now fetches files in chunks using HTTP range
+  requests (if supported by the server). `wasmfs_create_fetch_backend` now
+  takes a second parameter (`uint32_t chunk_size`) to configure the size of
+  each chunk. If a file is read a few times with random accesses, a small
+  chunk size will minimize bandwidth; if a file is read in larger contiguous
+  ranges, a larger chunk size will reduce the number of requests.
```
kripken, line 39:
```suggestion
  ranges, a larger chunk size will reduce the number of requests. (#23021)
```
mentioning the PR number lets people find it for more info

kripken (approved):
Great! Looks ready to land aside from the changelog comment.

(and unless @sbc100 has other comments)

kripken:
Looks like `other.test_closure_full_js_library_wasmfs` is failing on CI.

src/lib/libwasmfs_fetch.js
```diff
@@ -94,7 +94,7 @@ addToLibrary({
       if (!response.ok) {
         throw response;
       }
-      var bytes = await response.bytes();
+      var bytes = await response['bytes']();
```
sbc100, line 97: I'm OK landing this this way.  I will see if I can find a way to teach closure about this as a followup.

src/lib/libwasmfs_fetch.js
```diff
@@ -35,55 +31,127 @@ addToLibrary({
...
+            chunkSize: chunkSize
+          };
+        } else {
+          // may as well/forced to download the whole file
+          var wholeFileReq = await fetch(url);
+          if (!wholeFileReq.ok) {
+            throw wholeFileReq;
+          }
+          var wholeFileData = new Uint8Array(await wholeFileReq.arrayBuffer());
+          var text = new TextDecoder().decode(wholeFileData);
```
ringoz, line 64: this line should've been deleted

JoeOsborn: Yes that’s true, I left it in by mistake. Good catch!

PR #23022, opened 2024-11-26
Use octal JS literals. NFC

sbc100:
Followup to #23014

### DIFF
--- a/src/library_lz4.js
+++ b/src/library_lz4.js
@@ -8,8 +8,8 @@
 addToLibrary({
   $LZ4__deps: ['$FS', '$preloadPlugins'],
   $LZ4: {
-    DIR_MODE: {{{ cDefs.S_IFDIR }}} | 511 /* 0777 */,
-    FILE_MODE: {{{ cDefs.S_IFREG }}} | 511 /* 0777 */,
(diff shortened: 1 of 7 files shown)


src/library_lz4.js
```diff
@@ -8,8 +8,8 @@
 addToLibrary({
   $LZ4__deps: ['$FS', '$preloadPlugins'],
   $LZ4: {
-    DIR_MODE: {{{ cDefs.S_IFDIR }}} | 511 /* 0777 */,
-    FILE_MODE: {{{ cDefs.S_IFREG }}} | 511 /* 0777 */,
+    DIR_MODE: {{{ cDefs.S_IFDIR }}} | 0o777,
```
kripken, line 11: Does closure optimize this? The octal notation is easier to read but it is larger.

kripken: The acorn optimizer might also optimize this, maybe just by default, if it doesn't track what was octal originally and preserve it?

sbc100: Yes, which ever layer it is that converts this to a normal number, it is happening since this change has zero code size changes.

kripken (approved):

PR #23023, opened 2024-11-27
Fix for core0.test_stat_chmod_rawfs on mac. NFC

sbc100:
I'm not sure how #23018 managed landed with this broken.


sbc100:
Landing NFC to green the CI

PR #23026, opened 2024-11-27
Use preset to set doesNotExistError instead of postset. NFC

hoodmane:
Sometimes we throw `MEMFS.doesNotExistError` before it is initialized. It doesn't really matter because the `undefined` we throw gets caught and thrown away, but it seems better to fix.

### DIFF
--- a/src/library_memfs.js
+++ b/src/library_memfs.js
@@ -7,7 +7,7 @@
 addToLibrary({
   $MEMFS__deps: ['$FS', '$mmapAlloc'],
 #if !ASSERTIONS
-  $MEMFS__postset: `
+  $MEMFS__preset: `
     // This error may happen quite a bit. To avoid overhead we reuse it (and
     // suffer a lack of stack info).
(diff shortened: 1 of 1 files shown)


src/library_memfs.js
```diff
@@ -7,7 +7,7 @@
 addToLibrary({
   $MEMFS__deps: ['$FS', '$mmapAlloc'],
 #if !ASSERTIONS
-  $MEMFS__postset: `
```
sbc100, line 10: Can you explain how `MEMFS.doesNotExistError` could get used before the postset run?  IIUC the `post` set code is emitted at the top level at the end of the library top level code.    Is the problem that other `postset` code wants to run first?

Wouldn't `preset` attempt to set `MEMFS.doesNotExistError` before `MEMFS` itself is created?

sbc100: Actually `__preset` just doesn't exist.  Only `__postset`.  So I'm not sure how this could work.

PR #21548, opened 2024-03-16
Update mimalloc to 2.1.7

kleisauke:

Issue #20645, opened 2023-11-07
emmalloc wastes memory with high alignments


kripken:
```c
#include <emscripten/emmalloc.h>
#include <stdio.h>

int main() {
  // Allocate 32MB with 1MB alignment.
  printf("%p\n", emmalloc_memalign(1024 * 1024, 32 * 1024 * 1024));
}
```

Build with
```
./emcc a.c -sMALLOC=emmalloc -sWASM=0 -sINITIAL_MEMORY=128mb --profiling
```
and then edit `a.out.js` to add some logging at the top of `function sbrk`:
```js
 function sbrk($0) {
  console.log('sbrk', $0);
```
Running the program then shows
```
sbrk 48
sbrk 33554480
sbrk 33554480
0x100000
```
The initial sbrk is for emmalloc's internal data. Then we sbrk around 32MB, but then do so a second time for a total of 64MB.

This seems related to the high alignment, 1MB, as this does not happen without it.

This affects `mimalloc` when layered on `emmalloc`, since `mimalloc` does such large allocations with high alignment.

englercj:
Was this fixed by #20704?

kripken:
Good catch, thanks @englercj ! Yes, this is basically fixed by that PR, aside from that one comment you noticed (unfortunately I don't think that remaining issue is fixable, as I think that high alignment is a design constraint of mimalloc).

In fact it looks like I meant to close this by that PR, by had a typo in the issue number to be closed by it :smile:  (20645/20654...)

englercj:
Awesome, thanks for the info!

kleisauke:
> [This comment](https://github.com/emscripten-core/emscripten/blob/900aee0a2df98b28579d72b17f6fa73e48087e69/system/lib/mimalloc/src/prim/emscripten/prim.c#L82) is still in `main` so I wasn't entirely sure.

It looks like `MI_SEGMENT_ALIGN` will affect this. I was able to halve that alignment by reducing the page size from 32KiB to 16KiB on wasm32, see for example:
https://github.com/kleisauke/emscripten/commit/20c31536251a4f7a8af89a1fa2147a2faa13209e
https://github.com/microsoft/mimalloc/issues/647#issuecomment-1324109021

(that commit also seems to fix a suspicious OOM error on [wasm-vips' benchmarks](https://github.com/kleisauke/wasm-vips/tree/master/test/bench) when switching the memory allocator from `dlmalloc` to `mimalloc`)

[...]

PR #23027, opened 2024-11-27
Update mimalloc to microsoft/mimalloc@9b75377

kleisauke:


kleisauke:
Context: this PR seems to address a suspicious(?) OOM error on wasm-vips' test suite. See for example the CI results for commit https://github.com/kleisauke/wasm-vips/commit/605b53fc7d925507eabb668bbcb9627f325d5a64 and https://github.com/kleisauke/wasm-vips/commit/5cb048be333d5ccf65587e9f0c51c571f8055364.
https://github.com/kleisauke/wasm-vips/actions/runs/12053482829/job/33610497071#step:7:302
https://github.com/kleisauke/wasm-vips/actions/runs/12053529699

It's generally recommended to use tagged upstream versions, but I prefer to avoid those that are known to have issues. Note that the previous version of mimalloc (used prior to PR #21548) was commit https://github.com/microsoft/mimalloc/commit/4e50d6714d471b72b2285e25a3df6c92db944593, which is also an untagged version.

system/lib/mimalloc/include/mimalloc.h
```diff
@@ -8,7 +8,7 @@ terms of the MIT license. A copy of the license can be found in the file
 #ifndef MIMALLOC_H
 #define MIMALLOC_H
 
-#define MI_MALLOC_VERSION 217   // major + 2 digits minor
+#define MI_MALLOC_VERSION 188   // major + 2 digits minor
```
kleisauke, line 11: Ah, I just noticed that I synced with the `dev` branch. Perhaps it should be synced with the `dev-slice` branch instead.
https://github.com/emscripten-core/emscripten/blob/8d59f19ef9aeeabeb23f1977d7ccd67fbd3fe215/system/lib/mimalloc/readme.md?plain=1#L71-L76

kleisauke:
Updating to the latest commit on the `dev-slice` branch did not resolve the issue (see https://github.com/kleisauke/wasm-vips/commit/28b407a1fc10f80f15c93a948cfacd97508c140a and https://github.com/emscripten-core/emscripten/compare/HEAD...kleisauke:mimalloc-update-577246d for details).

Upon further investigation, it appears to be the same issue as outlined in https://github.com/emscripten-core/emscripten/issues/20645#issuecomment-1962964755, i.e. mimalloc requests an alignment of 4 MiB, causing emmalloc to allocate 8 MiB (or perhaps was that fixed via 3cd6861ccdf02a89555b94d6ab5397efb1b0ede8?), which somehow exceeds the `-sINITIAL_MEMORY=1GB` limit in wasm-vips' test suite. Halving the page size seems to fix this.

Closing in favor of #23037.

PR #23029, opened 2024-11-27
Add test for `GLOBAL_BASE` and `SIDE_MODULE` error message. NFC

sbc100:
These two options are not compatible, and `GLOBAL_BASE` is currently just ignored in this mode.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -1273,6 +1273,10 @@ f.close()
     err = self.expect_fail(cmd + ['-sGLOBAL_BASE=1024'])
     self.assertContained('error: --stack-first is not compatible with -sGLOBAL_BASE', err)
 
+  def test_side_module_global_base(self):
+    err = self.expect_fail([EMCC, test_file('hello_world.c'), '-Werror', '-sGLOBAL_BASE=1024', '-sSIDE_MODULE'])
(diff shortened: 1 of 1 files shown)


kripken (approved):

test/test_other.py
```diff
@@ -1273,6 +1273,10 @@ def test_wl_stackfirst(self):
     err = self.expect_fail(cmd + ['-sGLOBAL_BASE=1024'])
     self.assertContained('error: --stack-first is not compatible with -sGLOBAL_BASE', err)
 
+  def test_dylink_global_base(self):
+    err = self.expect_fail([EMCC, test_file('hello_world.c'), '-sGLOBAL_BASE=1024', '-sMAIN_MODULE'])
+    self.assertContained('emcc: error: GLOBAL_BASE is not compatible with relocatable output', err)
+
```
kripken, line 1279: Testing a side module instead would be faster, and also be future compatible if we ever make main modules non-relocatable. But just a suggestion, lgtm either way.

sbc100: The error happens before we do any actual linking work so its just as fast either way.

sbc100:
Oops it turns out the GLOBAL_BASE does work with MAIN_MODULE since it gets embedded in the JS file.  

Updated this RP to just a test for the existing error for SIDE_MODULE.

kripken (approved):
Oh, interesting.

PR #23030, opened 2024-11-28
Cleanup library_sdl.js. NFC

sbc100:


src/library_sdl.js
```diff
@@ -2893,7 +2903,7 @@ var LibrarySDL = {
       if (channelInfo.audio === this || channelInfo.audio.webAudioNode === this) { 
         channelInfo.audio.paused = true; channelInfo.audio = null; 
       }
-      if (SDL.channelFinished)  {{{ makeDynCall('vi', 'SDL.channelFinished') }}}(channel);
+      if (SDL.channelFinished) {{{ makeDynCall('vi', 'SDL.channelFinished') }}}(channel);
```
brendandahl, line 2906: Braces for this too or does the preprocessor not like that?

brendandahl (approved):
r+ w/ one little nit

src/library_sdl.js
sbc100: We do still allow single line if statements and they don't use braces in that case.

PR #23035, opened 2024-11-29
Add AVX2 support

jiepan-intel:
It's successor of PR #22430,  one 256-bit AVX2 intrinsic can be also emulated by two 128-bit intrinsics.


site/source/docs/porting/simd.rst
```diff
@@ -1138,6 +1139,88 @@ The following table highlights the availability and expected performance of diff
...
+   * - _mm_i32gather_epi32
+     - ❌ scalarized
+   * - _mm_i64gather_epi32
+     - ❌ scalarized
+   * - _mm_i32gather_epi64
+     - ❌ scalarized
+   * - _mm_i64gather_epi64
+     - ❌ scalarized
+
+All the 128-bit wide instructions from AVX2 instruction set are listed. Only a small part of the 256-bit AVX2 instruction set are listed, most of the 256-bit wide AVX2 instructions are emulated by two 128-bit wide instructions.
```
sbc100, line 1223: Can you wrap text at 80 here

system/include/compat/avxintrin.h
```diff
@@ -4,20 +4,18 @@
  * University of Illinois/NCSA Open Source License.  Both these licenses can be
  * found in the LICENSE file.
  */
+
+#ifndef __emscripten_immintrin_h__
```
tlively, line 8: Let's use a different macro that gets undefined after all the relevant includes in immintrin.h. Otherwise, we won't emit this error if the user does something like this:

```
#include <immintrin.h>
#include <avxintrin.h>
```

system/include/compat/avx2intrin.h
```diff
@@ -0,0 +1,1800 @@
...
+  return ret;
+}
+
+static __inline__ __m128i __attribute__((__always_inline__, __nodebug__))
+_mm_maskload_epi32(int32_t const* __p, __m128i __m) {
+  // This may cause an out-of-bounds memory load since we first load and
+  // then mask, but since there are no segmentation faults in Wasm memory
+  // accesses, that is ok (as long as we are within the heap bounds -
+  // a negligible limitation in practice)
+  // TODO, loadu or load, 128-bit align?
```
tlively, line 1138: Could this cause ASan failures?

jiepan-intel: The snippet is ported from old code in avxintrin.h, I am not sure, I will check that.

system/include/compat/avxintrin.h
jiepan-intel: Yes, in this case, it won't emit error message, but I think it's suitable. 

Actually, include avxintrin.h directly  will miss prerequisites and cause compile error. Include immintrin.h can avoid this error. 
immintrin.h includes the necessary prerequisites and choose the *intrin.h corresponding to options(-mavx, -mavx2...).
so in the following code, if -mavx option is specified, #include <avxintrin.h> does nothing due to header guards, it's valid code, the behavior is consistent with LLVM.

```
#include <immintrin.h>
#include <avxintrin.h>
```

tlively (approved):
I'm fine with this landing once @sbc100's documentation comments are addressed and the tests are passing, etc.

site/source/docs/porting/simd.rst
jiepan-intel: Done

system/include/compat/avx2intrin.h
jiepan-intel: The implementation is changed, scalarized version is used now, it's more compliant with specifications.

kripken:
Looks like this was lgtm'd pending tests passing, and they just passed, landing.

PR #23037, opened 2024-11-29
[mimalloc] halve the page size

kleisauke:


kripken:
What is the benefit here? Did you measure this @kleisauke ?

kleisauke:
The advantage is that mimalloc uses 2 MiB alignments instead of 4 MiB, which causes emmalloc (and `sbrk()`) to allocate memory in smaller segments (i.e. 4 MiB instead of 8 MiB).
```diff
@@ -1,12 +1,12 @@
 initialize_emmalloc_heap()
 claim_more_memory(numBytes=48)
-claim_more_memory: claimed 0x0050ed80 - 0x0050edb0 (48 bytes) via sbrk()
+claim_more_memory: claimed 0x0050edc0 - 0x0050edf0 (48 bytes) via sbrk()
 claim_more_memory: sbrk() returned a disjoint region to last root region, some external code must have sbrk()ed outside emmalloc(). Creating a new root region
-allocate_memory(align=4194304, size=4194304 bytes)
-claim_more_memory(numBytes=8388656)
-claim_more_memory: claimed 0x0050edb0 - 0x00d0ede0 (8388656 bytes) via sbrk()
+allocate_memory(align=2097152, size=2097152 bytes)
+claim_more_memory(numBytes=4194352)
+claim_more_memory: claimed 0x0050edf0 - 0x0090ee20 (4194352 bytes) via sbrk()
 claim_more_memory: sbrk() returned a region contiguous to last root region, expanding the existing root region
-allocate_memory(align=4194304, size=4194304 bytes)
-attempt_allocate(freeRegion=0x0050ed90,alignment=4194304,size=4194304)
-attempt_allocate - succeeded allocating memory, region ptr=0x007ffffc, align=4194304, payload size=4194304 bytes)
+allocate_memory(align=2097152, size=2097152 bytes)
+attempt_allocate(freeRegion=0x0050edd0,alignment=2097152,size=2097152)
+attempt_allocate - succeeded allocating memory, region ptr=0x005ffffc, align=2097152, payload size=2097152 bytes)
```
(before commit 3cd6861ccdf02a89555b94d6ab5397efb1b0ede8 this issue did not occur because an internal alignment of 1 MiB in mimalloc was always enforced)

This was only verified on wasm-vips' test suite, which somehow consistently exceeds the `-sINITIAL_MEMORY=1GB` limit after the mimalloc v2.1.7 update.

I should also verify this on Emscripten's benchmarks. They need to be redone anyway, the previous benchmarks was done with assertions enabled in emmalloc (i.e. before commit d0a2dce14cc3debad719e111ed7830ee5eb728b0), which may have affect the results.
https://github.com/emscripten-core/emscripten/blob/58889f9f20f74d0dbaed7d49025c49b864359ae1/test/test_benchmark.py#L968-L972

kripken:
Sounds good. Yeah, if this does not regress that benchmark that it seems reasonable to me.

And meanwhile I see that this is recommended for "small systems", which probably describes most wasm usage: https://github.com/microsoft/mimalloc/issues/647#issuecomment-1324109021

kleisauke:
Closing in favor of #26696.

PR #23040, opened 2024-11-29
Ruff rules for comprehensions, performance, and simplicity

cclauss:
Only in the `tools/` directory apply ruff rules for comprehensions, performance, and simplicity.
* https://docs.astral.sh/ruff
* https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4
* https://docs.astral.sh/ruff/rules/#perflint-perf
* https://docs.astral.sh/ruff/rules/#flake8-simplify-sim

This pull request was roughly created via the command:
`ruff --select=C4,PERF,Q,SIM --ignore=SIM108,SIM300 lint.flake8-quotes.inline-quotes=single --fix --unsafe-fixes`


tools/link.py
```diff
@@ -216,8 +216,7 @@ def get_js_sym_info():
   # and can contain full paths to temporary files.
   skip_settings = {'PRE_JS_FILES', 'POST_JS_FILES'}
   input_files = [json.dumps(settings.external_dict(skip_keys=skip_settings), sort_keys=True, indent=2)]
-  for jslib in sorted(glob.glob(utils.path_from_root('src') + '/library*.js')):
-    input_files.append(read_file(jslib))
+  input_files.extend(read_file(jslib) for jslib in sorted(glob.glob(utils.path_from_root('src') + '/library*.js')))
```
sbc100, line 219: Is there a way to configure it to avoid generating overly complex list comprehension like this?    I think there is too much going on here for a single line of code.

tools/maint/gen_sig_info.py
```diff
@@ -181,9 +181,7 @@ def ignore_symbol(s, cxx):
            'stackSave', 'stackRestore', 'stackAlloc', 'getTempRet0', 'setTempRet0',
            }:
     return True
-  if cxx and s in ('__asctime_r') or s.startswith('__cxa_find_matching_catch'):
-    return True
-  return False
+  return bool(cxx and s == '__asctime_r' or s.startswith('__cxa_find_matching_catch'))
```
sbc100, line 184: `bool` seems unnecessary here?

tools/maint/gen_struct_info.py
```diff
@@ -242,10 +242,7 @@ def generate_cmd(js_file_path, src_file_path, cflags):
   # Compile the program.
   show('Compiling generated code...')
 
-  if any('libcxxabi' in f for f in cflags):
-    compiler = shared.EMXX
-  else:
-    compiler = shared.EMCC
+  compiler = shared.EMXX if any('libcxxabi' in f for f in cflags) else shared.EMCC
```
sbc100, line 245: Can we tell it to avoid this kind of thing?   I'm not big a fan of the single line if / else syntax.

tools/ports/__init__.py
```diff
@@ -118,8 +118,8 @@ def dir_is_newer(dir_a, dir_b):
   assert os.path.exists(dir_b)
   files_a = [(x, os.path.getmtime(x)) for x in get_all_files_under(dir_a)]
   files_b = [(x, os.path.getmtime(x)) for x in get_all_files_under(dir_b)]
-  newest_a = max([f for f in files_a], key=lambda f: f[1])
-  newest_b = max([f for f in files_b], key=lambda f: f[1])
+  newest_a = max(list(files_a), key=lambda f: f[1])
```
sbc100, line 121: Can we just drop the `list()` here too?

tools/ports/__init__.py
```diff
@@ -326,10 +326,7 @@ def unpack():
       utils.write_file(marker, url + '\n')
 
     def up_to_date():
-      if os.path.exists(marker):
-        if utils.read_file(marker).strip() == url:
-          return True
-      return False
+      return bool(os.path.exists(marker) and utils.read_file(marker).strip() == url)
```
sbc100, line 329: No need for `bool` here I think?

tools/ports/bullet.py
```diff
@@ -32,10 +32,7 @@ def create(final):
         shared.safe_ensure_dirs(os.path.dirname(target))
         shutil.copyfile(fullpath, target)
 
-    includes = []
-    for base, dirs, _ in os.walk(src_path, topdown=False):
-      for dir in dirs:
-        includes.append(os.path.join(base, dir))
+    includes = [os.path.join(base, dir) for base, dirs, _ in os.walk(src_path, topdown=False) for dir in dirs]
```
sbc100, line 35: Can you revert this one?  I think it makes the line to long.

tools/toolchain_profiler.py
```diff
@@ -152,7 +152,7 @@ def escape_string(arg):
 
     @staticmethod
     def escape_args(args):
-      return map(lambda arg: ToolchainProfiler.escape_string(arg), args)
+      return (ToolchainProfiler.escape_string(arg) for arg in args)
```
sbc100, line 155: Just return a list here maybe?  `[ToolchainProfiler.escape_string(arg) for arg in args]`

sbc100 (commented):
Most of these look great!

Thanks for working on this.

sbc100:
Can you include the command line you used to generate this PR (in the PR description).

tools/link.py
cclauss:
```suggestion
  jslibs = glob.glob(utils.path_from_root('src') + '/library*.js')
  input_files.extend(read_file(jslib) for jslib in sorted(jslibs))
```

tools/maint/gen_sig_info.py
cclauss:
```suggestion
  return cxx and s == '__asctime_r' or s.startswith('__cxa_find_matching_catch')
```

tools/maint/gen_struct_info.py
cclauss: % [`ruff rule SIM108`](https://docs.astral.sh/ruff/rules/if-else-block-instead-of-if-exp/)

tools/toolchain_profiler.py
cclauss: [`map()`](https://docs.python.org/3/library/functions.html#map) returns an iterator in Python 3 so the linter's fix wants to return a generator expression.  Lists take more RAM than maps and generators.

tools/ports/bullet.py
cclauss:
```suggestion
    for base, dirs, _ in os.walk(src_path, topdown=False):
      for dir in dirs:
        includes.append(os.path.join(base, dir))
```

tools/toolchain_profiler.py
sbc100: Sure, this is far from memory critical code though and I think the explicit list is more idiomatic (at least to my old fasioned eyes).

tools/emscripten.py
```diff
@@ -699,15 +699,12 @@ def create_asm_consts(metadata):
...
     max_arity = 16
     arity = 0
     for i in range(max_arity):
-      if ('$' + str(i)) in const:
+      if f'${i}' in const:
         arity = i + 1
-    for i in range(arity):
-      args.append('$' + str(i))
-    args = ', '.join(args)
+    args = ', '.join(f"${i}" for i in range(arity))
```
sbc100, line 707: Can you use single quotes here?

tools/feature_matrix.py
```diff
@@ -106,16 +106,16 @@ def report_missing(setting_name):
...
-  if settings.MIN_FIREFOX_VERSION < min_versions['firefox']:
+  if min_versions['firefox'] > settings.MIN_FIREFOX_VERSION:
     report_missing('MIN_FIREFOX_VERSION')
     return False
-  if settings.MIN_SAFARI_VERSION < min_versions['safari']:
+  if min_versions['safari'] > settings.MIN_SAFARI_VERSION:
     report_missing('MIN_SAFARI_VERSION')
     return False
-  if 'node' in min_versions and settings.MIN_NODE_VERSION < min_versions['node']:
+  if 'node' in min_versions and min_versions['node'] > settings.MIN_NODE_VERSION:
```
sbc100, line 118: Why does the linter want to flip these comparisons?  I think maybe the read more clearly as written.   Maybe revert this file?

sbc100 (approved):
lgtm!   Thanks for working on this.

tools/feature_matrix.py
cclauss: You must be a fan of Star Wars!

% [`ruff rule SIM300`](https://docs.astral.sh/ruff/rules/yoda-conditions)

sbc100: Maybe its misleading since `settings.MIN_NODE_VERSION` is not so much a constant as a user setting.

In this case we are saying "is the user-specified-min-node-version less then the min node version required for this feature".   So in this `min_versions['node']` (the min node version required for the feature) is really the constant.

tools/toolchain_profiler.py
cclauss: I will change this but...

The Py 2 to 3 transition contained a few key transitions.
* `bytes == str` vs. `bytes != str`
* https://docs.python.org/3/whatsnew/3.0.html#print-is-a-function (and related changes)
* https://docs.python.org/3/whatsnew/3.0.html#views-and-iterators-instead-of-lists

Lists were more idiomatic in in legacy Python but the push towards iterators was quite deliberate.

sbc100 (approved):

PR #23043, opened 2024-12-01
Add MAIN_THREAD_EM_ASM_PROMISE_AWAIT

hedwigz:
Introduce a new macro `MAIN_THREAD_EM_ASM_PROMISE_AWAIT` which is used to write JavaScript code that **returns a promise** and block the C code from progressing until the promise is resolved (or errors).  

```C
int main() {
  printf("1st print\n");
  // This call will block until the internal promise resolves (at least 1 second in this case) 
  MAIN_THREAD_EM_ASM_PROMISE_AWAIT({
    out('2nd print');
    // do some async operation that involves the main thread
    return new Promise((resolve,reject) => {
      setTimeout(() => {
        out('3rd print');
        resolve();
      }, 1000);
    });
  });
  printf("4th print\n");
  return 0;
}
```

The motivation for this was [discussed before](https://github.com/emscripten-core/emscripten/discussions/21929)  
We have multiple web workers, each of them calls javascript code which is async and has to be ran on the main thread and need to wait for the promise to be resolved.  
I am open for suggestions on how to add this more elegantly


src/library_pthread.js
```diff
@@ -983,6 +983,27 @@ var LibraryPThread = {
...
+#if MEMORY64
+    // In memory64 mode some proxied functions return bigint/pointer but
+    // our return type is i53/double.
+    if (typeof res == "bigint") {
+      res = bigintToI53Checked(res);
+    }
+#endif
+      __emscripten_proxy_promise_finish(res, promiseCtx);
+      }).catch(err => {
+        __emscripten_proxy_promise_finish(err, promiseCtx);
```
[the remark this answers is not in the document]
hedwigz: It's defined [here](https://github.com/emscripten-core/emscripten/pull/23043/files#diff-ca572ec5c79d581d3657cfcf00705c200889335d5289c917656864187fb6ca62R651-R653)  
I was not sure how to add an internal C function that is available to the runtime so I added it to `REQUIRED_EXPORTS` [here](https://github.com/emscripten-core/emscripten/pull/23043/files#diff-a9cba4e03450a831038c69937bac7910c60b2a341f1eef13b70aaf4e21cb7c5dR505)  
Is there a better way to do this?

test/core/test_main_thread_async_em_asm_promise_await.cpp
```diff
@@ -0,0 +1,24 @@
+// Copyright 2017 The Emscripten Authors.  All rights reserved.
```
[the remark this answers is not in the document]
hedwigz: fixed

test/test_core.py
```diff
@@ -1896,6 +1896,13 @@ def test_main_thread_em_asm(self, args):
   def test_main_thread_async_em_asm(self, args, force_c=False):
     self.do_core_test('test_main_thread_async_em_asm.cpp', emcc_args=args, force_c=force_c)
 
+  @needs_dylink
+  @parameterized({
+    'pthreads': (['-pthread', '-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME'], False),
```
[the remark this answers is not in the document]
hedwigz: I added the empty one, and the test expects for an error. I did not handle the case of calling this macro from the main thread. Do you think we should add that?

hedwigz:
@sbc100 ping 🤞

src/library.js
```diff
@@ -1655,6 +1655,16 @@ addToLibrary({
   emscripten_asm_const_int_sync_on_main_thread__deps: ['$runMainThreadEmAsm'],
   emscripten_asm_const_int_sync_on_main_thread: (emAsmAddr, sigPtr, argbuf) => runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, 1),
 
+  emscripten_asm_const_int_await_promise_on_main_thread__deps: ['$runMainThreadEmAsm'],
+  emscripten_asm_const_int_await_promise_on_main_thread: (emAsmAddr, sigPtr, argbuf) => {
```
sbc100, line 1659: How about we drop the `promise` part so this becomes `emscripten_asm_const_int_await_on_main_thread` and the macro is `MAIN_THREAD_EM_ASM_AWAIT`?

src/library.js
```diff
@@ -1631,7 +1631,7 @@ addToLibrary({
     '$proxyToMainThread'
 #endif
   ],
-  $runMainThreadEmAsm: (emAsmAddr, sigPtr, argbuf, sync) => {
+  $runMainThreadEmAsm: (emAsmAddr, sigPtr, argbuf, sync, promise) => {
```
sbc100, line 1634: Maybe call this argument "await" too.. although that could be confused with the JS keyword.  Maybe "asyncAwait"?

src/library.js
```diff
@@ -1655,6 +1655,16 @@ addToLibrary({
...
   emscripten_asm_const_int_sync_on_main_thread: (emAsmAddr, sigPtr, argbuf) => runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, 1),
 
+  emscripten_asm_const_int_await_promise_on_main_thread__deps: ['$runMainThreadEmAsm'],
+  emscripten_asm_const_int_await_promise_on_main_thread: (emAsmAddr, sigPtr, argbuf) => {
+    #if PTHREADS
+    if (ENVIRONMENT_IS_PTHREAD) {
+      return runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, 1, 1);
+    }
+    #endif
+    throw new Error('call to emscripten_asm_const_int_await_promise_on_main_thread is only supported from pthread (but was called from main thread)');
```
sbc100, line 1665: We probably don't want this long error message in release builds.  How about:

```
#if ASSERTIONS
assert(ENVIRONMENT_IS_PTHREAD. "emscripten_asm_const_int_await_promise_on_main_thread is not available on the main thread");
#endif
return runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, 1, 1);
```

src/library.js
```diff
@@ -1655,6 +1655,16 @@ addToLibrary({
   emscripten_asm_const_int_sync_on_main_thread__deps: ['$runMainThreadEmAsm'],
   emscripten_asm_const_int_sync_on_main_thread: (emAsmAddr, sigPtr, argbuf) => runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, 1),
 
+  emscripten_asm_const_int_await_promise_on_main_thread__deps: ['$runMainThreadEmAsm'],
+  emscripten_asm_const_int_await_promise_on_main_thread: (emAsmAddr, sigPtr, argbuf) => {
+    #if PTHREADS
+    if (ENVIRONMENT_IS_PTHREAD) {
+      return runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, 1, 1);
```
sbc100, line 1662: Can you add comments here for these contants e.g.

```
return runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, /*sync=*/1, /*asyncAwait=*/1);
```

src/library_pthread.js
```diff
@@ -983,6 +983,27 @@ var LibraryPThread = {
     PThread.currentProxiedOperationCallerThread = callingThread;
     var rtn = func(...proxiedJSCallArgs);
     PThread.currentProxiedOperationCallerThread = 0;
+    if (promiseCtx) {
+#if ASSERTIONS
+      assert(!!rtn.then, 'Return value of proxied function expected to be a Promise but got' + rtn);
+#endif
+      if (!rtn.then) {
+        throw new Error('Return value of proxied function expected to be a Promise but got' + rtn);
+      }
```
sbc100, line 992: Need for both that assertion and runtime check here.  I think you can just remove these 3 lines.

system/lib/pthread/proxying.c
```diff
@@ -583,31 +632,35 @@ em_promise_t emscripten_proxy_promise(em_proxying_queue* q,
...
   }
 }
 
+static void run_js_promise_await(void* arg) {
+  proxied_js_func_t* f = (proxied_js_func_t*)arg;
+  f->result = _emscripten_receive_on_main_thread_js(
+    f->funcIndex, f->emAsmAddr, f->callingThread, f->numArgs, f->argBuffer, (void*)f->ctx);
+}
+
+void _emscripten_proxy_promise_finish(void* res, em_proxying_ctx* ctx) {
```
sbc100, line 651: I would expect the context to be the first arugment here.

system/lib/pthread/proxying.c
```diff
@@ -620,12 +673,18 @@ double _emscripten_run_on_main_thread_js(int func_index,
...
       return 0;
     }
     return f.result;
+  } else if (sync && promise) {
+    if (!emscripten_proxy_promise_await(q, target, run_js_promise_await, &f)) {
+      assert(false && "emscripten_proxy_promise failed");
+      return 0;
+    }
+    return f.result;
   }
```
sbc100, line 688: Maybe structure this as:

```
if (sync) {
   if (promise) {
   } else { 
   }
   return f.result;
}
```

Also the fact that we have `int sync` and `int promise` here as a separate arguments, but it doesn't make sense for `promise` to be set without `sync` seem a little odd.

At the very least we should probably assert that promise is never true when sync is not.

sbc100 (commented):
@tlively would you mind taking a look, since this contains changes to proxying.c

system/lib/pthread/proxying.c
tlively, line 1: Since the new code in this file does not implement the API surface in proxying.h, it would be best to move it to a different file and leave proxying.c unchanged. It would also be best to implement the new functionality in terms of the public API in proxying.h rather than using internal implementation details. For example, there should be no need to reach inside the `em_proxying_ctx` object.

tlively: Although I see we already have `_emscripten_run_on_main_thread_js` at the end of proxying.c. Adding more code there would be fine, but it should still use only the public API from proxying.h.

src/library_pthread.js
hedwigz: oops. nice catch

system/lib/pthread/proxying.c
hedwigz: 👍

src/library.js
hedwigz: Yes, I like that better as well

src/library.js
hedwigz: 👍

src/library.js
hedwigz: 👍

src/library.js
hedwigz: 👍

src/library.js
hedwigz: I had to modify the assert condition because it errored: `ReferenceError: ENVIRONMENT_IS_PTHREAD is not defined`  
```javascript
assert((typeof ENVIRONMENT_IS_PTHREAD !== 'undefined' && ENVIRONMENT_IS_PTHREAD), "emscripten_asm_const_int_await_on_main_thread is not available on the main thread");
```
Not sure if there's a prettier way

system/lib/pthread/proxying.c
hedwigz: Do you prefer that I introduce a new function similar to `_emscripten_run_on_main_thread_js` but that uses only public proxying.c API?
The problem with introducing a new function similar to `emscripten_run_on_main_thread_js` is the "path" to that function call.  
It'd a bit tedious. And it'd modify the whole stack leading to this logic.  
This is why I chose to introduce the new function param `asyncAwait` (previously `promise`) just so I could "ride" the whole functions call path.  
But, I don't insist on this at all.  
As for using the public API - I agree that we can do that if I indeed fork from `_emscripten_run_on_main_thread_js`  
Let me know what you think

hedwigz:
@sbc100 @tlively 🙏 I really appreciate the thorough review and comments!  
I addressed all of your comments and I think the code looks way better now.

src/library.js
sbc100: Use `#if ASSERTION && PTHREADS`

system/lib/pthread/proxying.c
tlively: Why do you need to use internal details of the proxying mechanism if you don't create an entirely new version of `_emscripten_run_on_main_thread_js`? Can't you use `emscripten_proxy_sync_with_ctx` to synchronously wait for the promise to be resolved on the main thread instead of introducing a new `emscripten_proxy_async_await` function?

hedwigz: The challenging part was to pass the `em_proxying_ctx` all the way down to `_emscripten_receive_on_main_thread_js`, through the `task` and `proxied_js_func_t` wrappers, because I have to call `_emscripten_proxy_promise_finish` from within `_emscripten_receive_on_main_thread_js`.  
This is also why I added `em_proxying_ctx * ctx` to `proxied_js_func_t` and moved it upwards.  
I made a significant simplification and deleted most of the previous code in `proxying.c`  
WDYT?

hedwigz: Now that I think of it, it is a little weird that `proxied_js_func_t` has a field `ctx` that if set, causes `_emscripten_receive_on_main_thread_js` to expect a promise. Maybe I should name it differently?

system/lib/pthread/proxying.c
```diff
@@ -642,3 +651,34 @@ double _emscripten_run_on_main_thread_js(int func_index,
...
   return 0;
 }
+
+double _emscripten_await_on_main_thread_js(int func_index,
```
tlively, line 655: I do think there is enough code shared with `_emscripten_run_on_main_thread_js` that it may make sense to combine them, but I don't feel strongly about it. If you do keep them separate, you should explicitly initialize the `.ctx` field to `NULL` in `_emscripten_run_on_main_thread_js` to make sure it is initialized .

system/lib/pthread/proxying.c
```diff
@@ -642,3 +651,34 @@ double _emscripten_run_on_main_thread_js(int func_index,
...
+  }
+  return f.result;
+}
+
+void _emscripten_proxy_promise_finish(em_proxying_ctx* ctx, void* res) {
+  task* t = (task*)ctx->arg;
+  proxied_js_func_t* func = (proxied_js_func_t*)t->arg;
+  func->result = (double)(intptr_t)res;
+  emscripten_proxy_finish(ctx);
+}
```
tlively, line 684: Please make sure there is a newline at the end of the file. It would also be good to run `git clang-format` to fix any formatting errors.

system/lib/pthread/proxying.c
tlively: I think the `ctx` and `promiseCtx` names are fine, but it would be good to add comments explaining that they can be null.

tlively (commented):
Great, this looks a lot simpler overall, even though I agree that it's a little awkward to get the `ctx` into the right place. Thanks!

hedwigz:
> Great, this looks a lot simpler overall, even though I agree that it's a little awkward to get the `ctx` into the right place. Thanks!

Awesome :)  
Any tips for how to pass the CI tests?

tlively:
proxying.c looks good to me.

It looks like the flake8 bot is failing because of a formatting issue. The other CI bots might be aborting early because of that issue. Merging in main to pick up any unrelated fixes could also help.

sbc100:
The fact that flake8 is being included at all in CI makes me think that perhaps you are not up-to-date on your branch?  The flake8 step was replaced with a step call `ruff` few days back.

sbc100:
Can you try rebase onto to merging with the latest changes on main?

hybridherbst:
Any updates or blockers here? I'd like to see this land as well, we're also using workers that need to sync await some main thread code.

hedwigz:
> Any updates or blockers here? I'd like to see this land as well, we're also using workers that need to sync await some main thread code.

Hey @hybridherbst , glad to see more interest in this.  The current blocker is that some wasm64 test fail. I was struggling making the test run on my machine due to the need in d8/node24

hedwigz:
@sbc100 Thanks! I'll give it a go

hedwigz:
@tlively @sbc100 All tests are now green :)

test/test_core.py
```diff
@@ -1892,6 +1892,26 @@ def test_main_thread_em_asm(self, args):
...
 
+  @needs_dylink
+  @parameterized({
+    '': (['-sASSERTIONS'], False),
+    'pthreads': (['-pthread', '-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME'], False),
+  })
+  def test_main_thread_async_em_asm_await(self, args, force_c=False):
+    if '-sPROXY_TO_PTHREAD' not in args:
+      # expect runtime to error
+      output = self.do_runf('core/test_main_thread_async_em_asm_await.cpp', expected_output=None, assert_returncode=NON_ZERO, emcc_args=args)
```
sbc100, line 1903: You can remove `expected_output=None` here

test/test_core.py
```diff
@@ -1892,6 +1892,26 @@ def test_main_thread_em_asm(self, args):
   def test_main_thread_async_em_asm(self, args, force_c=False):
     self.do_core_test('test_main_thread_async_em_asm.cpp', emcc_args=args, force_c=force_c)
 
+  @needs_dylink
+  @parameterized({
+    '': (['-sASSERTIONS'], False),
+    'pthreads': (['-pthread', '-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME'], False),
+  })
+  def test_main_thread_async_em_asm_await(self, args, force_c=False):
```
sbc100, line 1900: It looks like force_c is always false.  Maybe just remove it?

test/core/test_main_thread_async_em_asm_await_reject.cpp
```diff
@@ -0,0 +1,34 @@
...
+// Emscripten is available under two separate licenses, the MIT license and the
+// University of Illinois/NCSA Open Source License.  Both these licenses can be
+// found in the LICENSE file.
+
+#include <emscripten.h>
+#include <stdio.h>
+#include <pthread.h>
+
+int main()
+{
```
sbc100, line 11: Curly brace goes on the same line as the function name

test/core/test_main_thread_async_em_asm_await_reject.cpp
```diff
@@ -0,0 +1,34 @@
+// Copyright 2024 The Emscripten Authors.  All rights reserved.
```
sbc100, line 1: 2025

test/core/test_main_thread_async_em_asm_await_reject.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <emscripten.h>
+#include <stdio.h>
+#include <pthread.h>
+
+int main()
+{
+  // start new thread
+  pthread_t thread;
+  pthread_create(&thread, NULL, [](void*) -> void* {
```
sbc100, line 14: Can you split this out into a separate function

tools/link.py
```diff
@@ -514,6 +514,7 @@ def setup_pthreads():
   settings.REQUIRED_EXPORTS += [
     '_emscripten_thread_free_data',
     '_emscripten_thread_crashed',
+    '_emscripten_proxy_promise_finish',
```
sbc100, line 517: I don't think this should be needed here.  At least I don't see runtime_pthread.js depending on this.  Instead I think you can add this to `_emscripten_receive_on_main_thread_js__deps`

src/lib/libpthread.js
```diff
@@ -960,6 +970,24 @@ var LibraryPThread = {
     PThread.currentProxiedOperationCallerThread = callingThread;
     var rtn = func(...proxiedJSCallArgs);
     PThread.currentProxiedOperationCallerThread = 0;
+    if (promiseCtx) {
+#if ASSERTIONS
+      assert(!!rtn.then, 'Return value of proxied function expected to be a Promise but got' + rtn);
```
sbc100, line 975: I don't think you need the `!!` here ?

src/lib/libpthread.js
```diff
@@ -960,6 +970,24 @@ var LibraryPThread = {
...
+#endif
+      rtn.then(res => {
```
sbc100, line 977: We put braces around the `res` here and `err` below.

system/lib/pthread/proxying.c
```diff
@@ -591,12 +591,15 @@ typedef struct proxied_js_func_t {
   double* argBuffer;
   double result;
   bool owned;
+  // Only used when the underlying js func is async.
+  // Can be null when the function is sync.
+  em_proxying_ctx * ctx;
```
sbc100, line 596: No space before `*` here.

test/test_core.py
```diff
@@ -1972,6 +1972,24 @@ def test_main_thread_em_asm(self, args):
...
+    '': (['-sASSERTIONS']),
+    'pthreads': (['-pthread', '-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME']),
+  })
+  def test_main_thread_async_em_asm_await(self, args):
+    if '-sPROXY_TO_PTHREAD' not in args:
+      # expect runtime to error
+      output = self.do_runf('core/test_main_thread_async_em_asm_await.cpp', assert_returncode=NON_ZERO, emcc_args=args)
+      self.assertContained('emscripten_asm_const_int_await_on_main_thread is not available on the main thread', output)
+    else:
+      self.do_core_test('test_main_thread_async_em_asm_await.cpp', emcc_args=args, force_c=False)
```
sbc100, line 1985: You don't need `force_c=False` here

test/test_core.py
```diff
@@ -1972,6 +1972,24 @@ def test_main_thread_em_asm(self, args):
...
+
+  @parameterized({
+    'pthreads': (['-pthread', '-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME', '-sASSERTIONS'], False),
+  })
+  def test_main_thread_async_em_asm_await_reject(self, args, force_c=False):
+    self.do_core_test('test_main_thread_async_em_asm_await_reject.cpp', emcc_args=args, force_c=force_c)
```
sbc100, line 1991: I don't think you need `force_c` here or above.

test/test_core.py
```diff
@@ -1972,6 +1972,24 @@ def test_main_thread_em_asm(self, args):
...
+    if '-sPROXY_TO_PTHREAD' not in args:
+      # expect runtime to error
+      output = self.do_runf('core/test_main_thread_async_em_asm_await.cpp', assert_returncode=NON_ZERO, emcc_args=args)
+      self.assertContained('emscripten_asm_const_int_await_on_main_thread is not available on the main thread', output)
+    else:
+      self.do_core_test('test_main_thread_async_em_asm_await.cpp', emcc_args=args, force_c=False)
+
+  @parameterized({
+    'pthreads': (['-pthread', '-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME', '-sASSERTIONS'], False),
+  })
```
sbc100, line 1989: I think you can just remove the `@parameterized` block since there is only one variant of this test.

test/core/test_main_thread_async_em_asm_await.cpp
```diff
@@ -0,0 +1,25 @@
+// Copyright 2024 The Emscripten Authors.  All rights reserved.
+// Emscripten is available under two separate licenses, the MIT license and the
+// University of Illinois/NCSA Open Source License.  Both these licenses can be
+// found in the LICENSE file.
+
+#include <emscripten.h>
+#include <stdio.h>
+
+int main()
+{
```
sbc100, line 10: Open brace on same line as function decl.

test/core/test_main_thread_async_em_asm_await_reject.cpp
```diff
@@ -0,0 +1,35 @@
...
+    const asyncOp = new Promise((resolve,reject) => {
+      setTimeout(() => {
+        out('Inside asyncOp');
+        reject(new Error('asyncOp rejected'));
+      }, 1000);
+    });
+    return asyncOp;
+  }, 42, 3.5);
+  printf("After MAIN_THREAD_EM_ASM_AWAIT rejected\n");
+  printf("result: %d\n", res);
```
sbc100, line 23: If code that uses threads its better to use `emscripten_outf` rather than `printf` since its writes to the lower level file descriptor under node. (Writing to stdout can have buffering issues).

sbc100:
Thanks for continuing to work on this.

Apologies for the review process taking so long.  I had no imagined this being such an invasive change (i.e.changing the code proxying functions).

hedwigz:
@sbc100 Thanks, I appreciate the continued work on this. I didn't have much time recently to cater to this PR. I'll hopefully get back to it soon.

sbc100:
Sorry for the long delay on this.

I've since landed a JS library feature that is kind of like this one: #26000.  So I think i now have a much better understanding of the use case.

Would you have time to rebase this and we can see about landing it?

test/test_core.py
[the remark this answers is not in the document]
hedwigz: Addressed in the rebased branch.

src/library_pthread.js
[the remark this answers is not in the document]
hedwigz: Addressed or superseded by the rebased implementation.

src/library.js
hedwigz: Addressed or superseded by the rebased implementation.

src/library.js
hedwigz: Addressed or superseded by the rebased implementation.

src/library.js
hedwigz: Addressed or superseded by the rebased implementation.

src/library.js
hedwigz: Addressed or superseded by the rebased implementation.

src/library_pthread.js
hedwigz: Addressed or superseded by the rebased implementation.

system/lib/pthread/proxying.c
hedwigz: Addressed or superseded by the rebased implementation.

system/lib/pthread/proxying.c
hedwigz: Addressed or superseded by the rebased implementation.

system/lib/pthread/proxying.c
hedwigz: Addressed or superseded by the rebased implementation.

system/lib/pthread/proxying.c
```diff
@@ -13,6 +13,7 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdio.h>
```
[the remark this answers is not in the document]
hedwigz: Addressed or superseded by the rebased implementation.

system/lib/pthread/proxying.c
```diff
@@ -35,6 +36,17 @@ static em_proxying_queue system_proxying_queue = {
   .capacity = 0,
 };
 
+typedef struct proxied_js_func_t {
```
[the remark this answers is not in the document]
hedwigz: Addressed or superseded by the rebased implementation.

system/lib/pthread/proxying.c
hedwigz: Addressed or superseded by the rebased implementation.

system/lib/pthread/proxying.c
hedwigz: Addressed or superseded by the rebased implementation.

test/test_core.py
hedwigz: Addressed or superseded by the rebased implementation.

test/test_core.py
hedwigz: Addressed or superseded by the rebased implementation.

test/core/test_main_thread_async_em_asm_await_reject.cpp
hedwigz: Addressed or superseded by the rebased implementation.

test/core/test_main_thread_async_em_asm_await_reject.cpp
hedwigz: Addressed or superseded by the rebased implementation.

test/core/test_main_thread_async_em_asm_await_reject.cpp
hedwigz: Addressed or superseded by the rebased implementation.

tools/link.py
hedwigz: Addressed or superseded by the rebased implementation.

src/lib/libcore.js
```diff
@@ -1572,12 +1572,13 @@ addToLibrary({
...
   },
 
+  $runMainThreadEmAsm__docs: '/** @param {number=} asyncAwait */',
   $runMainThreadEmAsm__deps: ['$readEmAsmArgs',
 #if PTHREADS
     '$proxyToMainThread'
 #endif
   ],
-  $runMainThreadEmAsm: (emAsmAddr, sigPtr, argbuf, sync) => {
+  $runMainThreadEmAsm: (emAsmAddr, sigPtr, argbuf, sync, asyncAwait) => {
```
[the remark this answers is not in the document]
hedwigz: Addressed or superseded by the rebased implementation.

src/lib/libcore.js
```diff
@@ -1572,12 +1572,13 @@ addToLibrary({
     return runEmAsmFunction(code, sigPtr, argbuf);
   },
 
+  $runMainThreadEmAsm__docs: '/** @param {number=} asyncAwait */',
```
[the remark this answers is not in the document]
hedwigz: Addressed or superseded by the rebased implementation.

src/lib/libcore.js
```diff
@@ -1601,6 +1602,19 @@ addToLibrary({
   emscripten_asm_const_int_sync_on_main_thread__deps: ['$runMainThreadEmAsm'],
   emscripten_asm_const_int_sync_on_main_thread: (emAsmAddr, sigPtr, argbuf) => runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, 1),
 
+  emscripten_asm_const_int_await_on_main_thread__deps: ['$runMainThreadEmAsm'],
+  emscripten_asm_const_int_await_on_main_thread: (emAsmAddr, sigPtr, argbuf) => {
+    #if PTHREADS
```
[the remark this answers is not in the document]
hedwigz: Addressed or superseded by the rebased implementation.

src/lib/libpthread.js
hedwigz: Addressed or superseded by the rebased implementation.

src/lib/libpthread.js
hedwigz: Addressed or superseded by the rebased implementation.

system/lib/pthread/proxying.c
hedwigz: Addressed or superseded by the rebased implementation.

test/test_core.py
hedwigz: Addressed or superseded by the rebased implementation.

test/test_core.py
hedwigz: Addressed or superseded by the rebased implementation.

test/test_core.py
hedwigz: Addressed or superseded by the rebased implementation.

test/core/test_main_thread_async_em_asm_await.cpp
hedwigz: Addressed or superseded by the rebased implementation.

test/core/test_main_thread_async_em_asm_await_reject.cpp
hedwigz: Addressed or superseded by the rebased implementation.

src/lib/libcore.js
```diff
@@ -1601,6 +1602,19 @@ addToLibrary({
...
+    if (ENVIRONMENT_IS_PTHREAD) {
+      return runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, /*sync=*/1, /*asyncAwait=*/1);
+    }
+    #endif
+    #if ASSERTIONS
+    assert((typeof ENVIRONMENT_IS_PTHREAD !== 'undefined' && ENVIRONMENT_IS_PTHREAD), "emscripten_asm_const_int_await_on_main_thread is not available on the main thread");
```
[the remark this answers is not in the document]
hedwigz: Addressed: the function is now wrapped in `#if PTHREADS`, with a direct `ENVIRONMENT_IS_PTHREAD` assertion.

hedwigz:
@sbc100 Hey, I rebased it and addressed all of your comments

sbc100:
@guybedford WDYT of this extra macro/tool?

src/lib/libpthread.js
```diff
@@ -1086,7 +1086,10 @@ var LibraryPThread = {
     var rtn = func(...proxiedJSCallArgs);
     PThread.currentProxiedOperationCallerThread = 0;
     if (ctx) {
-      rtn.then((rtn) => __emscripten_run_js_on_main_thread_done(ctx, ctxArgs, rtn));
+      rtn.then(
+        (rtn) => __emscripten_run_js_on_main_thread_done(ctx, ctxArgs, rtn),
+        () => __emscripten_run_js_on_main_thread_done(ctx, ctxArgs, 0),
```
sbc100, line 1091: Does this mean we always return zero on promise rejection?   Is this needed?   What is we have a function that uses a zero return to mean something else?

Do you need this feature?  Maybe we should leave it out for the initial version?

test/core/test_main_thread_async_em_asm_await.out
```diff
@@ -0,0 +1,5 @@
+Before MAIN_THREAD_EM_ASM_AWAIT
+Inside MAIN_THREAD_EM_ASM_AWAIT: 42 3.5
+Inside asyncOp
+After MAIN_THREAD_EM_ASM_AWAIT
+result: 2
```
sbc100, line 5: missing newline here

test/core/test_main_thread_async_em_asm_await.cpp
```diff
@@ -0,0 +1,24 @@
...
+// Emscripten is available under two separate licenses, the MIT license and the
+// University of Illinois/NCSA Open Source License.  Both these licenses can be
+// found in the LICENSE file.
+
+#include <emscripten.h>
+#include <emscripten/console.h>
+
+int main() {
+  emscripten_out("Before MAIN_THREAD_EM_ASM_AWAIT");
+  int res = MAIN_THREAD_EM_ASM_AWAIT({
```
sbc100, line 11: Might it be nice if we could use `await` inside these functions? 

i.e. should we mark these function snippets as `async` on the JS side?

src/lib/libcore.js
```diff
@@ -1675,6 +1675,16 @@ addToLibrary({
   emscripten_asm_const_int_sync_on_main_thread__deps: ['$runMainThreadEmAsm'],
   emscripten_asm_const_int_sync_on_main_thread: (emAsmAddr, sigPtr, argbuf) => runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, 1),
 
+#if PTHREADS
+  emscripten_asm_const_int_await_on_main_thread__deps: ['$runMainThreadEmAsm'],
+  emscripten_asm_const_int_await_on_main_thread: (emAsmAddr, sigPtr, argbuf) => {
+#if ASSERTIONS
+    assert(ENVIRONMENT_IS_PTHREAD, 'emscripten_asm_const_int_await_on_main_thread is not available on the main thread');
+#endif
+    return runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, 2);
```
sbc100, line 1684: I wonder if these is some way we can avoid the magic number here?  

How about `{{{ PROXY_SYNC_ASYNC }}}`?

PR #23045, opened 2024-12-02
Handle trailing % in strftime

hoodmane:
"In glibc a trailing % in strftime() acts like printf, ie it's a literal %". This is breaking some Python tests. I've applied the patch suggested here: https://www.openwall.com/lists/musl/2022/12/19/2

### DIFF
--- a/system/lib/libc/musl/src/time/strftime.c
+++ b/system/lib/libc/musl/src/time/strftime.c
@@ -226,7 +226,7 @@ size_t __strftime_l(char *restrict s, size_t n, const char *restrict f, const st
 			s[l] = 0;
 			return l;
 		}
-		if (*f != '%') {
+		if (*f != '%' || !f[1]) {
 			s[l++] = *f;
 			continue;
 		}
--- a/test/other/test_strftime.c
+++ b/test/other/test_strftime.c
@@ -307,5 +307,8 @@ int main() {
   size = strftime(s, sizeof(s), "%Ec", &tm);
   TEST(!cmp(s, "Mon Dec 17 00:00:00 2018"), "strftime test #36a", s);
 
+  size = strftime(s, sizeof(s), "trailing %", &tm);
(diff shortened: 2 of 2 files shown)


hoodmane:
I guess this would have to be addressed upstream with musl though.

system/lib/libc/musl/src/time/strftime.c
```diff
@@ -226,7 +226,7 @@ size_t __strftime_l(char *restrict s, size_t n, const char *restrict f, const st
 			s[l] = 0;
 			return l;
 		}
-		if (*f != '%') {
+		if (*f != '%' || !f[1]) {
```
sbc100, line 229: Can you mention this in system/lib/libc/README.md and put the change inside `#ifdef __EMSCRIPTEN__`?

sbc100 (commented):
lgtm % comment

sbc100:
Perhaps the python tests themselves should be updated?   Otherwise they cannot run on any musl-based platform (e.g. alpine linux).

hoodmane:
Generally I try to fix both sides =)
https://github.com/python/cpython/pull/127528

hoodmane:
I think it works on other musl systems as long as they support wide characters. The problem affects systems that use musl or bsd libc and where `HAVE_WCSFTIME` is false.

sbc100:
Ah.. In that case we can/should also be including wcsftime.. i don't know why we don't already include that.

hoodmane:
I looked into turning that on too but I couldn't figure out how it was determined. But I think this change is an improvement in either case.

sbc100:
This change lgtm with comments addressed.

sbc100:
wcsftime being added in https://github.com/emscripten-core/emscripten/pull/23061

sbc100 (approved):

hoodmane:
Okay I addressed the comments I think.

PR #23049, opened 2024-12-02
Fix false positive writability check on cache directory

sbc100:
We were incorrectly reporting non-existent parent directories as non-writable.

This was broken in #22801.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -15412,3 +15412,25 @@ addToLibrary({
        return 0;
     }''')
     self.do_runf('main.cpp', 'Hello from rust!', emcc_args=[lib])
+
+  @crossplatform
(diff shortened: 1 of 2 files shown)


test/test_other.py
```diff
@@ -15400,3 +15400,18 @@ def test_rust_integration_basics(self):
        return 0;
     }''')
     self.do_runf('main.cpp', 'Hello from rust!', emcc_args=[lib])
+
+  @crossplatform
+  def test_create_cache_directory(self):
+    # Test that cache directory (including parents directories) is
```
kripken, line 15406:
```suggestion
    # Test that the cache directory (including parent directories) is
```

kripken (approved):

PR #23050, opened 2024-12-02
[test] Avoid seeding random number generators with time. NFC

sbc100:
Using the current time to seed the random number generator in tests will make them non-determinisitic.  Simply using the default seed or seeding with a fixed value seem fine all of these test cases.


test/core/test_rand.c
```diff
@@ -0,0 +1,43 @@
+#include <stdlib.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <assert.h>
+
+int main() {
+  // we need RAND_MAX to be a bitmask (power of 2 minus 1). this assertions guarantees
+  // if RAND_MAX changes the test failure will focus attention on that issue here.
```
kripken, line 8:
```suggestion
  // we need RAND_MAX to be a bitmask (power of 2 minus 1). this assertions guarantees
  // an error if RAND_MAX ever changes.
```

kripken: (pre-existing, but might as well fix it here)

kripken (approved):

PR #23051, opened 2024-12-02
[tsgen] Lower memory64 when generating typescript bindings

sbc100:
Fixes: https://github.com/emscripten-core/emsdk/issues/1494


kripken (approved):
Great!

sbc100:
Actually, this lowering is still needed since we need to run the binary under node.

Issue #17951, opened 2022-09-29
Defining custom Module.instantiateWasm() causes -sWASMFS builds to die horribly


sgbeal:
**Version of emscripten/emsdk:**

```
$ emcc --version
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.23 (5ae63ce7dd955df449cb419bfe5afc51d1bd57f2)
Copyright (C) 2014 the Emscripten authors (see AUTHORS.txt)
This is free and open source software under the MIT license.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ uname -a
Linux nuc 5.15.0-47-generic #51-Ubuntu SMP Thu Aug 11 07:51:15 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

i also tried version 3.1.21, with the same results.

**Failing command line in full:**

It's not quite that simple, unfortunately.

We have a large project (the upcoming official sqlite3 wasm build) for which we are trying to customize the WASM imports. Apparently the only way to do that with Emscripten is to provide a custom Module.instantiateWasm() and modify the imports object which gets passed to it.

Our custom loader works wonderfully in every build combination we've tried but dies horribly in an -sWASMFS build.

The loader:

```js
Module['instantiateWasm'] = function callee(imports,onSuccess){
  imports.foo = function(){}; // the whole goal of these acrobatics is to be able to customize the imports
  console.warn("instantiateWasm() uri =",callee.uri, self.location.href);
  const wfetch = ()=>fetch(callee.uri, {credentials: 'same-origin'});
  const loadWasm = WebAssembly.instantiateStreaming
        ? function loadWasmStreaming(){
          return WebAssembly.instantiateStreaming(wfetch(), imports)
            .then((arg)=>onSuccess(arg.instance, arg.module));
        }
        : function loadWasmOldSchool(){ // Safari < v15
          return wfetch()
            .then(response => response.arrayBuffer())
            .then(bytes => WebAssembly.instantiate(bytes, imports))
            .then((arg)=>onSuccess(arg.instance, arg.module));
        };
  loadWasm();
  return {};
};
/*
  It is literally impossible to get the name of a Worker's own script,
  so impossible to derive X.wasm from script name X.js. Thus we need,
  at build-time, to redefine Module['instantiateWasm'].uri by
  appending it to a build-specific copy of this file with the name of
[...]

Issue #15941, opened 2022-01-10
WasmFS: Node FS backend


tlively:
Add a concrete WasmFS backend for managing files via Node's file system APIs, analogous to [NODEFS](https://emscripten.org/docs/api_reference/Filesystem-API.html#filesystem-api-nodefs).

Depends on #15946.

kripken:
Thinking about this, I wonder if our current backend API is enough for NODEFS etc.

Current state: Backends define `createFile` (and directory and symlink). When the open syscall fails to find a file already present on the parent directory, it calls that method on the backend. Then it creates a generic OpenFileState for it etc.

Perhaps we should also have backends define `openFile`, and allow them to create a subclass of OpenFileState. Doing so would let NODEFS mirror the WasmFS state onto node file operations: when a file is opened in WasmFS, we'd open it in node, and when it is closed in WasmFS we'd close it in node as well. Without adding something like this, all a backend defines is a custom `File` structure, but that stays around forever and there is no signal as to when it is closed.

tlively:
Backends can override the `File` constructor and destructor to implement custom logic for opening and closing files.

kripken:
Sorry if I wasn't clear. There is a single `File` for a particular file. That is create once and never destroyed. But we can create and destroy multiple `OpenFileState`s for that file, say if we open it, write, close it, then much much later open it again. So if we want to track that behavior, we'd need to customize the latter.

In practical terms: In the constructor of `File` we don't know the flags the file is opened in (writing? reading?), and even if we did, they would be constant forever. So we can't open the file using Node.js APIs at that time. We would need to say open it and close it immediately in each write syscall, which doesn't sound good. Does that make more sense now?

tlively:
Oh right, my bad. The `File` destructor is for deleting files, not just closing them. Letting backends provide custom `OpenFileState`s would be one solution, but another option would be for `File` to have open and close hooks that backends could override. I think the hooks would probably be the simpler option since they don't have the overhead of defining a full new class and adding a new top-level function to the backend interface.

kripken:
Interesting, what would those hooks look like? Receiving the OpenFileState perhaps? I think that might work.

> The File destructor is for deleting files, not just closing them

[...]

PR #23058, opened 2024-12-03
Make fstat work on file descriptor with no name

### LINKED ISSUE #16414: fstat doesn't work on pipes
Consider the following simple program:
```C
#include <sys/types.h>
#include <sys/stat.h>
#include <unistd.h>
#include <stdio.h>

int fd[2];

int main(){
    struct stat statbuf;
    pipe(fd);
    int status = fstat(fd[0], &statbuf);
    printf("status: %d\n", status);
	return status;
}
```
If I compile it with `gcc` and run it, it will print `status: 0`. If I compile it with `emcc` and run it with node, it will print `status: -1`.

hoodmane:
This also is a bit more efficient: rather than taking the stream and converting it to a path, then calling `lookupPath` to find the node and calling getattr on it, just lookup `stream.node` and call `getattr` on it directly.

This requires that we add `stream_ops.getattr` in addition to `node_ops.getattr` -- the first to implement lookup from file descriptors to file attributes, the second to implement lookup from file system path to file attributes. Most file systems do not need a `stream_ops.getattr` because `stream.node` contains all the information about the file. However, in the nodefs, `stream` holds the `nfd`. If the file descriptor points to an unlinked file, the `nfd` is needed to access the native file information but the node doesn't have it. Thus, the nodefs needs both `node_ops.getattr` for the path based functions and `stream_ops.getattr` for the file descriptor based functions and similar for `setattr`. This requires a significant refactor. The common logic is moved into helper functions.

Resolves #16414.


hoodmane:
Looks like to fix this in NODEFS, we need to make `stream_ops.getattr` and `stream_ops.setattr` and have `fstat`, `fchmod`, `fchown`, and `ftruncate` use them instead of the corresponding `node_ops` if they exist.

src/library_nodefs.js
```diff
@@ -116,64 +116,68 @@ addToLibrary({
...
       return newFlags;
     },
-
-    node_ops: {
-      getattr(node) {
-        var path = NODEFS.realPath(node);
-        var stat;
-        NODEFS.tryFSOperation(() => stat = fs.lstatSync(path));
+    getattr(func) {
+        var stat = NODEFS.tryFSOperation(func);
```
sbc100, line 120: Indentation looks wrong here.

test/test_core.py
```diff
@@ -5921,6 +5921,21 @@ def test_fs_64bit(self):
...
+    'noderawfs':(['-sNODERAWFS'],),
+  })
+  def test_fs_stat_unnamed_file_descriptor(self, args):
+    self.emcc_args += args
+    nodefs = '-DNODEFS' in args or '-DNODERAWFS' in args
+    if self.get_setting('WASMFS'):
+      if nodefs:
+        self.skipTest('NODEFS in WasmFS')
+      self.set_setting('FORCE_FILESYSTEM')
+    self.do_runf('fs/test_stat_unnamed_file_descriptor.c', 'success')
```
sbc100, line 5937: `emcc_args=args` here instead of `self.emcc_args += args` above.

test/fs/test_stat_unnamed_file_descriptor.c
```diff
@@ -0,0 +1,40 @@
...
+}
+
+void changedir(const char *dir) {
+  int rtn = chdir(dir);
+  assert(rtn == 0);
+}
+
+
+int main() {
+    makedir("working");
```
sbc100, line 23: 2 space indentation

test/fs/test_stat_unnamed_file_descriptor.c
```diff
@@ -0,0 +1,40 @@
...
+#if defined(__EMSCRIPTEN__) && defined(NODEFS)
+    EM_ASM(FS.mount(NODEFS, { root: '.' }, 'working'));
+#endif
+    changedir("working");
```
sbc100, line 27: Can you put the makedir and changedir calls (and the function defs inside the `defined(__EMSCRIPTEN__) && defined(NODEFS)` block?

Also maybe move this code into a separate `setup()` function?

src/library_syscall.js
```diff
@@ -679,8 +679,7 @@ var SyscallsLibrary = {
     return SYSCALLS.doStat(FS.lstat, path, buf);
   },
   __syscall_fstat64: (fd, buf) => {
-    var stream = SYSCALLS.getStreamFromFD(fd);
-    return SYSCALLS.doStat(FS.stat, stream.path, buf);
+    return SYSCALLS.doStat(FS.fstat, fd, buf);
```
sbc100, line 682: Perhaps rename the second argument to `doStat` if we are going to allow it to take things that are not paths?

src/library_fs.js
```diff
@@ -1018,10 +1048,32 @@ FS.staticInit();
...
+        throw new FS.ErrnoError(errCode);
+      }
+      var attrs = {
+        size: len,
+        timestamp: Date.now()
+      };
+      if (stream.stream_ops.setattr) {
+        stream.stream_ops.setattr(stream, attrs);
+      } else {
+        node.node_ops.setattr(node, attrs);
```
sbc100, line 1075: Is this a common pattern where we try to do an operation first on the stream and then on the node if it fails?

Do we need the fallback?

hoodmane: Well we have a map from file descriptors to streams, and a map from paths to nodes. Each stream has a reference to its node. The difficulty is that not every node has a valid path. 

Most file systems store all relevant information about a node on the node itself. However, in the nodefs, each stream has a distinct native file descriptor. In the case where the node has no valid path, the only way to talk to the native object is with `fstat` and friends using the native file descriptor. So in that case we need these `stream_ops`. For every other FS, it's sufficient to have `setattr` on the `node_ops`.

Also, we _do_ need the `node_ops` case because if we call `stat` instead of `fstat` then we have a path but no file descriptor. We could `open` the path to get the file descriptor and then `stat` it and close it. But we're better off just having a separate handler for gettattr/setattr on a path/node than on a file descriptor/stream. But only nodefs seems to need it.

src/library_fs.js
```diff
@@ -1017,11 +1047,36 @@ FS.staticInit();
...
+      }
```
sbc100, line 1079: Seems like a fair amount of duplication here between `truncate` and `ftruncate`.. is there some way to extract the common stuff?

hoodmane: Yeah can do.

src/library_fs.js
```diff
@@ -985,20 +1005,37 @@ FS.staticInit();
...
+          throw new FS.ErrnoError({{{ cDefs.EINVAL }}});
+        }  
+        node = stream.node;
+      } else if (typeof arg == 'string') {
+        var lookup = FS.lookupPath(arg, { follow: true });
         node = lookup.node;
       } else {
-        node = path;
+        node = arg;
       }
```
sbc100, line 1037: Can we leave lines 1025 through 1037 in the caller and also avoid having a polymorphic `arg` here?

How about `truncateCommon(node, stream, len)` here node would be required and stream would be optional.

src/library_nodefs.js
```diff
@@ -232,18 +243,22 @@ addToLibrary({
...
+        NODEFS.setattr(stream.nfd, stream.node, attr, fs.fchmodSync, fs.futimesSync, fs.ftruncateSync);
+      },
       open(stream) {
         var path = NODEFS.realPath(stream.node);
         NODEFS.tryFSOperation(() => {
-          if (FS.isFile(stream.node.mode)) {
-            stream.shared.refcount = 1;
-            stream.nfd = fs.openSync(path, NODEFS.flagsForNode(stream.flags));
-          }
+          stream.shared.refcount = 1;
```
sbc100, line 255: So we are removing all of the `FS.isFile(stream.node.mode)` checks here.. are you sure that is OK / safe?

src/library_syscall.js
```diff
@@ -40,9 +40,10 @@ var SyscallsLibrary = {
       }
       return PATH.join2(dir, path);
     },
-
-    doStat(func, path, buf) {
-      var stat = func(path);
+    // When called by stat, arg is a path. When called by fstat, arg is a file
+    // descriptor.
+    doStat(func, arg, buf) {
+      var stat = func(arg);
```
sbc100, line 46: Can we have the caller call `func(arg)` and just pass the stat result.

Maybe rename this `writeStat(bug, stat)`?

src/library_syscall.js
```diff
@@ -679,8 +680,7 @@ var SyscallsLibrary = {
     return SYSCALLS.doStat(FS.lstat, path, buf);
   },
   __syscall_fstat64: (fd, buf) => {
-    var stream = SYSCALLS.getStreamFromFD(fd);
-    return SYSCALLS.doStat(FS.stat, stream.path, buf);
+    return SYSCALLS.doStat(FS.fstat, fd, buf);
```
sbc100, line 683: So this would be `SYSCALLS.writeStat(buf, FS.fstat(fd))`

src/library_fs.js
hoodmane: I'm not entirely clear what you mean by this but I did something in [989cf22](https://github.com/emscripten-core/emscripten/pull/23058/commits/989cf22b8aa5080ad0ec2f2dfa76b2eeac1cbbb9), let me know if that's not what you were suggesting.

src/library_syscall.js
hoodmane: Sounds good.

src/library_nodefs.js
hoodmane: Yeah, the new code paths require every NODEFS stream to have a `nfd`. Otherwise if we fstat the directory it will fail. I think this also fixes dup for nodefs directories -- I'm pretty sure it has the same problem I've fixed a bunch of times where if you open a directory, dup it, and then close it, attempting to continue using the original file descriptor will fail.

sbc100:
@hoodmane, I assume you are aware of the WasmFS project within emscripten: https://github.com/orgs/emscripten-core/projects/1.   The goal here is to eventually replace the current FS completely, if possible.   All these recent changes you have been making a much appreciate, but I wonder if you have looked at WasmFS yet, and what issues you ran into.   I wonder if it would make sense to focus more effect there, if that is where we will ultimately end up?   As you can see by the list of open issues there are still some limitations, and I wonder how much work it would be to be get to a place where you could use it in your project(s)?

hoodmane:
I haven't tried. The wasmfs does seem like it would solve problems for us. But it seems pretty far from being ready, in particular as far as I can tell it doesn't have a nodefs yet? Also #17951 looks a bit discouraging. So until it's a bit closer to serving our use case, I'd prefer to make improvements to the nearly-working existing file system.

kripken:
@hoodmane 

In general WasmFS is ready for most use cases. It is missing some features (looks like the `Module.instantiateWasm` hook is one of those, in the issue you linked), but if you don't need those features then it is usually a lot better (more correct and more efficient). There are some large real-world applications shipping with WasmFS right now. One reason for that is that only WasmFS supports OPFS (very important for apps with lots of data on the Web), and another is that WasmFS has full multithreading support.

A few other features missing in WasmFS are an IndexedDB backend and a few others. Given WasmFS's good OPFS support, that's been less of a priority. But WasmFS **does** already support NodeFS, include NODERAWFS,

* JS side: https://github.com/emscripten-core/emscripten/blob/main/src/library_wasmfs_node.js
* C++ side: https://github.com/emscripten-core/emscripten/blob/main/system/lib/wasmfs/backends/node_backend.cpp
* NODERAWFS integration: https://github.com/emscripten-core/emscripten/blob/main/system/lib/wasmfs/backends/noderawfs_root.cpp


Overall, I think it might be worth trying WasmFS: if you don't happen to need a missing feature, it might just work. Even if it doesn't, there is a decent chance that getting it to work would be similar or maybe even less effort than the (excellent!) work you've been doing on the old JS FS so far.

hoodmane:
That's good to know @kripken. Should #15941 be closed then? Or replaced with issues that describe more specific problems in wasmfs nodefs? Because the impression I get from that issue being open is that nodefs is not supported.

hoodmane:
Okay @sbc100 I updated the PR description. How does it look now? It also looks like it needs a rebaseline in code size. Maybe I should try again to see if I can get that working...

kripken:
@hoodmane Thanks, yeah, we just forgot and left that issue open! I closed it now.

hoodmane:
Okay, only test failures now are the common flakes. Can this be merged @sbc100?

hoodmane:
@sbc100 ping again on this one. Do you have any remaining concerns?

test/test_core.py
```diff
@@ -5839,6 +5839,20 @@ def test_fs_64bit(self):
       self.set_setting('FORCE_FILESYSTEM')
     self.do_runf('fs/test_64bit.c', 'success')
 
+  @requires_node
+  @parameterized({
+    '': ([],),
+    'nodefs': (['-DNODEFS', '-lnodefs.js'],),
+    'noderawfs': (['-sNODERAWFS'],),
+  })
```
sbc100, line 5847: Use `@also_with_nodefs_both` here and also add `@crossplatform`

test/fs/test_stat_unnamed_file_descriptor.c
```diff
@@ -0,0 +1,44 @@
...
+  assert(rtn == 0);
+}
+
+void setup() {
+#if defined(__EMSCRIPTEN__) && defined(NODEFS)
+  makedir("working");
+  EM_ASM(FS.mount(NODEFS, { root: '.' }, 'working'));
+  changedir("working");
+#endif
+}
```
sbc100, line 27: Everything above here can be removed now that we have `@also_with_nodefs_both` to do it for us

hoodmane: Indeed nice to get rid of all this.

hoodmane:
Oh ci is failing again.

test/test_core.py
```diff
@@ -5839,6 +5839,16 @@ def test_fs_64bit(self):
       self.set_setting('FORCE_FILESYSTEM')
     self.do_runf('fs/test_64bit.c', 'success')
 
+  @crossplatform
+  @also_with_nodefs_both
+  def test_fs_stat_unnamed_file_descriptor(self, args):
```
hoodmane, line 5844:
```suggestion
  def test_fs_stat_unnamed_file_descriptor(self):
```

sbc100 (commented):
Can we leave this larger change for the new year?   Its a little bigger than the others and I think there might be some more refactoring we could do on this one.

hoodmane:
Okay sounds good.

hoodmane:
@sbc100 if you want, I could split this into a first PR with all changes except for nodefs and a second PR with the nodefs changes.

hoodmane:
Closing in favor of #23364, #23470, and #23480.

PR #23056, opened 2024-12-03
Remove OFFSCREEN_FRAMEBUFFER_FORBID_VAO_PATH. NFC

sbc100:
This was the only remaining test-only setting.  The same effect can be achieved instead using some EM_ASM.

PR #8037, opened 2019-02-08
blitOffscreenFramebuffer fixes

kainino0x:
I originally went in to allow the blitFramebuffer path on FF >= 67, but
then I found a bunch of bugs while testing it.

 1. Fix test to test blitFramebuffer path
      - (Previously untested because I forgot to create a webgl2 context.)
 2. Fix blitFramebuffer path on antialias:true canvases
      - (Now tested via 5.)
 3. Fix VAO path on WebGL 1.0
    (Previously wasn't restoring the old VAO; bug was visible as a GL
    error, but I didn't check for errors.)
      - And test
 4. Allow blitFramebuffer path on FF >= 67. Fixed in FF 67 nightly!
      - And ensure both paths are tested on FF65, FF67, Chrome.
 5. Figured out a way to actually test the rendering results in
    test_webgl_offscreen_framebuffer_state_restoration.

Follow-up to #7588.

PR #23064, opened 2024-12-03
Parameterize test_webgl_offscreen_framebuffer_state_restoration. NFC

sbc100:


sbc100:
Split out from #23056

test/test_browser.py
```diff
@@ -4322,22 +4322,22 @@ def test_webgl_vao_without_automatic_extensions(self):
...
-        # blitFramebuffer path on WebGL 2.0 (falls back to VAO on Firefox < 67)
-        ['-sMAX_WEBGL_VERSION=2', '-DTEST_WEBGL2=1', '-DTEST_ANTIALIAS=0'],
-      ]:
-      cmd = args + ['-lGL', '-sOFFSCREEN_FRAMEBUFFER', '-DEXPLICIT_SWAP=1']
-      self.btest_exit('webgl_offscreen_framebuffer_swap_with_bad_state.c', args=cmd)
+  @parameterized({
+    # full state restoration path on WebGL 1.0
+    'gl1_no_vao': (['-sMAX_WEBGL_VERSION=1', '-sOFFSCREEN_FRAMEBUFFER_FORBID_VAO_PATH'],),
+    # VAO path on WebGL 1.0
+    'gl1': (['-sMAX_WEBGL_VERSION=1'],),
```
kainino0x, line 4329:
```suggestion
    'gl1': (['-sMAX_WEBGL_VERSION=1', '-DTEST_VERIFY_WEBGL1_VAO_SUPPORT=1'],),
```

kripken (approved):

test/browser/webgl_offscreen_framebuffer_swap_with_bad_state.c
```diff
@@ -28,7 +28,7 @@ int main() {
   EMSCRIPTEN_WEBGL_CONTEXT_HANDLE ctx = emscripten_webgl_create_context("#canvas", &attr);
   emscripten_webgl_make_context_current(ctx);
 
-#if !TEST_WEBGL2 && TEST_VAO
+#if !TEST_WEBGL2 && TEST_REQUIRE_VAO
   // This test cannot run without browser support for OES_vertex_array_object.
```
kainino0x, line 32: This is not the ideal state but it documents what I believe to be the current reality.

```suggestion
  // This test cannot run without browser support for OES_vertex_array_object.
  // This check is just to verify that the browser has support; otherwise, we
  // will end up testing the non-VAO path. Enabling it here does not actually do
  // anything, because offscreen framebuffer has already been initialized. Note
  // that if GL_SUPPORT_AUTOMATIC_ENABLE_EXTENSIONS=0, then offscreen
  // framebuffer will _never_ use VAOs on WebGL 1 (unless something enables
  // OES_vertex_array_object before createOffscreenFramebuffer runs).
```

test/browser/webgl_offscreen_framebuffer_swap_with_bad_state.c
```diff
@@ -28,7 +28,7 @@ int main() {
   EMSCRIPTEN_WEBGL_CONTEXT_HANDLE ctx = emscripten_webgl_create_context("#canvas", &attr);
   emscripten_webgl_make_context_current(ctx);
 
-#if !TEST_WEBGL2 && TEST_VAO
+#if !TEST_WEBGL2 && TEST_REQUIRE_VAO
```
kainino0x, line 31:
```suggestion
#if !TEST_WEBGL2 && TEST_VERIFY_WEBGL1_VAO_SUPPORT
```

test/test_browser.py
```diff
@@ -4322,22 +4322,22 @@ def test_webgl_vao_without_automatic_extensions(self):
...
+    'gl1_max_gl2': (['-sMAX_WEBGL_VERSION=2'],),
+    # VAO path on WebGL 2.0
+    'gl2': (['-sMAX_WEBGL_VERSION=2', '-DTEST_WEBGL2=1', '-DTEST_ANTIALIAS=1', '-DTEST_REQUIRE_VAO=1'],),
```
kainino0x, line 4332: `TEST_REQUIRE_VAO=1` doesn't actually seem to do anything in this case since `TEST_WEBGL2=1`.
```suggestion
    'gl2': (['-sMAX_WEBGL_VERSION=2', '-DTEST_WEBGL2=1', '-DTEST_ANTIALIAS=1'],),
```

kainino0x (commented):
I didn't realize that this whole mess was fully my doing (#8037). I checked there and the `TEST_VAO` vs `TEST_REQUIRE_VAO` thing was just wrong from the beginning. Probably I tried to refactor something during the PR and did it incompletely.

However from looking at that PR I think I have figured out what the intent was: not to _enable_ the extension, but to verify that it _was supported_ so we could be sure that OffscreenFramebuffer would be _using_ it.

Yesterday I was confused about what would actually be enabling it for OffscreenFramebuffer to use, but that was because I thought `emscripten_webgl_create_context` wouldn't enable it. But it does, because `GL_SUPPORT_AUTOMATIC_ENABLE_EXTENSIONS=1` by default.

Back when I originally wrote this I'm not sure there was such an option, or at least it was phrased differently. But I think what this means is if `GL_SUPPORT_AUTOMATIC_ENABLE_EXTENSIONS=0` then the VAO path will never be used on webgl1, because the extension won't be enabled early enough to hit it.

sbc100:
Thanks @kainino0x !

kainino0x (approved):

PR #23065, opened 2024-12-03
Convert asyncLoad from callbacks to promise. NFC

sbc100:
This is an internal function that should not have any external users.


kripken (approved):
Nice!

PR #23066, opened 2024-12-04
Convert instantiateAsync form callbacks to promises. NFC

sbc100:
I'll hoping to follow up with a conversion to using `await` here.

The code size changes seem to be mostly a wash with some tests improving and others regressing by a few bytes.

PR #23068, opened 2024-12-04
Use async/await instead of promises for wasm loading. NFC

sbc100:
These get lowered away by babel when targetting older engines.

Followup to #23066, with code size savings.


sbc100:
This change builds upon #23066

sbc100:
view with ignore whitespace

src/preamble.js
```diff
@@ -839,56 +832,51 @@ function instantiateAsync(binary, binaryFile, imports) {
...
+      // call receiveInstantiationResult, as that function will use the
+      // offset converter (in the case of pthreads, it will create the
+      // pthreads and send them the offsets along with the wasm instance).
+
+      var arrayBufferResult = await clonedResponsePromise;
+      try {
+        wasmOffsetConverter = new WasmOffsetConverter(new Uint8Array(arrayBufferResult), instantiationResult.module);
+          resolve(instantiationResult);
+      } catch (reason) {
+        (reason) => err(`failed to initialize offset-converter: ${reason}`)
```
brendandahl, line 869: This should just call err.

kripken: Should it also call abort? Before the promise rejected, but now IIUC the try-catch swallows the exception and we'd try to keep going.

sbc100: In the only code we use a promise rejection callback that just logs the error and does nothing else, so I think this is the same thing.

sbc100: When you have a rejection handler that rejection doesn't propagate out (IIUC).

src/library.js
```diff
@@ -128,7 +128,11 @@ addToLibrary({
     if (keepRuntimeAlive() && !implicit) {
       var msg = `program exited (with status: ${status}), but keepRuntimeAlive() is set (counter=${runtimeKeepaliveCounter}) due to an async operation, so halting execution but not exiting the runtime or preventing further async execution (you can use emscripten_force_exit, if you want to force a true shutdown)`;
 #if MODULARIZE
-      readyPromiseReject(msg);
+      if (!runtimeInitialized) {
+        // If the runtime has not yet been initializated then reject the
+        // ready promise instead of logging the error here.
+        readyPromiseReject(msg);
+      } else
```
kripken, line 135: This is during the shutdown code. Why do we need to check if the runtime was initialized right after logging that we exited?

src/postamble.js
```diff
@@ -163,22 +163,21 @@ function run() {
 
 #if WASM_WORKERS
   if (ENVIRONMENT_IS_WASM_WORKER) {
+    initRuntime();
```
kripken, line 166: Why does the order of promise resolution vs `initRuntime` need to be flipped now?

src/preamble.js
```diff
@@ -936,7 +919,13 @@ function getWasmImports() {
 
 // Create the wasm instance.
 // Receives the wasm imports, returns the exports.
+#if WASM_ASYNC_COMPILATION
+// Funnily enough in JS the `async` keyword has to be on the same line as the
+// function keyword.
+async function createWasm() {
```
brendandahl, line 925: Alternatively this could be `  {{{ asyncIf(WASM_ASYNC_COMPILATION) }}} function createWasm() {`

sbc100: Nice!  Thats much better

src/postamble.js
sbc100: This is not longer part of this PR

src/library.js
sbc100: This is no longer part of this PR

sbc100:
A few minor cleanups here.   I finally rememberd to use a followup commit instead of squashing, sorry I didn't do that earlier.

I'm pretty happy with this PR now, I think its fairly straight forward.

src/preamble.js
```diff
@@ -841,56 +833,41 @@ function instantiateAsync(binary, binaryFile, imports) {
...
-#endif
-
-        result.then(
+      // We need the wasm binary for the offset converter. Clone the response
+      // in order to get its arrayBuffer (cloning should be more efficient
+      // than doing another entire request).
+      // (We must clone the response now in order to use it later, as if we
+      // try to clone it asynchronously lower down then we will get a
+      // "response was already consumed" error.)
+      var clonedResponse = (await response).clone();
```
kripken, line 845: Is it perhaps more idiomatic to have the `await` on line 837? (`var x = await fetch(..)`) like line 779 etc.

kripken (approved):

src/preamble.js
sbc100: We could but then we would have an extra await there in the non `USE_OFFSET_CONVERTER` path.  For the normal path the response can be passed as a Promise of a Response (i.e. we don't need to await it it in the normal path).

brendandahl (approved):

PR #23069, opened 2024-12-04
[emmalloc] Fix verbose output

kleisauke:
`toString()` is undeclared.

Fixes a regression introduced in commit 04a0cad.


kleisauke:
Details:
```

building:ERROR: /tmp/emscripten_temp_50snxhja/vips-es6.jso3.js:757:85: ERROR - [JSC_UNDEFINED_VARIABLE] variable toString is undeclared
  757|     out("aligned_realloc(ptr=" + ptrToString($0) + ", alignment=" + $1 + ", size=" + toString(Number($2)));
                                                                                            ^^^^^^^^

1 error(s), 0 warning(s)
```

sbc100 (approved):

sbc100:
Are we missing a test for `-sMALLOC=emmalloc-verbose`?

kripken:
Yes, looks like that isn't covered by any existing test. Might be worth adding, but it is a setting for internal debugging.

kleisauke:
It looks like this is being covered here:
https://github.com/emscripten-core/emscripten/blob/a6ecc86b430dd26a2f8af73c0808d73cb9211d15/test/test_core.py#L797
https://github.com/emscripten-core/emscripten/blob/a6ecc86b430dd26a2f8af73c0808d73cb9211d15/test/test_browser.py#L5229

However, these tests do not verify the debug prints. I'm not sure if that would be useful.

sbc100:
> It looks like this is being covered here:
> 
> https://github.com/emscripten-core/emscripten/blob/a6ecc86b430dd26a2f8af73c0808d73cb9211d15/test/test_core.py#L797
> 
> 
> https://github.com/emscripten-core/emscripten/blob/a6ecc86b430dd26a2f8af73c0808d73cb9211d15/test/test_browser.py#L5229
> 
> However, these tests do not verify the debug prints. I'm not sure if that would be useful.

But you could add `--closure=1` which would cause the current code to fail right?

kleisauke:
> But you could add `--closure=1` which would cause the current code to fail right?

Ah, good point. Addressed with commit 6fef505f12534572e0bca0b270a72e7e7b4731ab, PTAL.

test/test_core.py
```diff
@@ -797,8 +797,8 @@ def test_mainenv(self):
     'memvalidate_verbose': ['-DEMMALLOC_MEMVALIDATE', '-DEMMALLOC_VERBOSE', '-DRANDOM_ITERS=130'],
   })
   def test_emmalloc(self, *args):
-    if '-DEMMALLOC_VERBOSE' in args and self.is_wasm64():
-      self.skipTest('EMMALLOC_VERBOSE is not compatible with wasm64')
```
sbc100, line 801: Oh nice to see this skip being removed!

test/test_core.py
```diff
@@ -797,8 +797,8 @@ def test_mainenv(self):
...
+    self.maybe_closure()
+    self.set_setting('DEFAULT_LIBRARY_FUNCS_TO_INCLUDE', ['$ptrToString'])
```
sbc100, line 801: I don't think this line should be needed.

Can we instead add `EM_JS_DEPS(deps, "$ptrToString");` to `emmalloc.c`?

kleisauke: 81182932afcd553adaded8e4a7d9663cbf4861a6

sbc100 (approved):

kripken (approved):
lgtm and that test failure on looks unrelated.

PR #22994, opened 2024-11-23
Update libcxx and libcxxabi to LLVM 19.1.4

aheejin:
This updates libcxx and libcxxabi to LLVM 19.1.4:
https://github.com/llvm/llvm-project/releases/tag/llvmorg-19.1.4

The initial update was done using `update_libcxx.py` and `update_libcxxabi.py`, and subsequent fixes were made in indidual commits. The commit history here is kind of messy because of CI testing so not all individual commits are noteworthy.

Additional changes:

- Build libcxx and libcxxabi with C++23: https://github.com/emscripten-core/emscripten/pull/22994/commits/8b0bfdf2d3039a3d6799dd7b407595f3ad246623
  https://github.com/llvm/llvm-project/blob/aadaa00de76ed0c4987b97450dd638f63a385bed/libcxx/src/expected.cpp
  was added in https://github.com/llvm/llvm-project/pull/87390 and this file assumes C++23 to be compiled. Apparently libc++ sources are always built with C++23 so they don't guard things against it in `src/`: https://github.com/llvm/llvm-project/pull/87390#discussion_r1575374945
  This commit also builds libc++abi with C++23 because it doesn't seem there's any downside to it.

- Exclude newly added `compiler_rt_shims.cpp`: https://github.com/emscripten-core/emscripten/pull/22994/commits/5bbcbf0f5701583f3ac6a782f532a7ea9842cd21
  We have excluded files in https://github.com/emscripten-core/emscripten/tree/main/system/lib/libcxx/src/support/win32.
  This is a new file added in this directory in https://github.com/llvm/llvm-project/pull/83575.

- Disable time zone support: https://github.com/emscripten-core/emscripten/pull/22994/commits/a5f2cbea31e398e63de004db521cb8e984dc8252
  We disabled C++20 time zone support in LLVM 18 update (#21638): https://github.com/emscripten-core/emscripten/commit/df9af64eb3f36b56133abacb6a0687253bd9563c
  The list of source files related to time zone support has changed in https://github.com/llvm/llvm-project/pull/74928, so this commit reflects it.

- Re-add + update `__assertion_handler` from `default_assertion_handler.in`: https://github.com/emscripten-core/emscripten/pull/22994/commits/41f803725dc9d37063869a6a3653adb83fe0854c
[...]

PR #23070, opened 2024-12-04
Remove basic_string<unsigned char> from embind

aheejin:
Only `char`, `wchar`, `char8`, `char16`, and `char32` are valid specializations for `std::basic_string`:
https://en.cppreference.com/w/cpp/string/basic_string

But libc++ had a base template for `basic_string` that allowed any type to be passed for a long time. It looks there have been several attempts to remove this after which they restored it due to complaints, in chronological order:
https://github.com/llvm/llvm-project/commit/aeecef08c385b1e4955155dd649a2d3724463849 https://github.com/llvm/llvm-project/commit/08a0faf4cd32bce6c51027ea9b5ec351747995b4 https://github.com/llvm/llvm-project/commit/e30a148b098d462d0267c479cd9e4783363c2761 https://github.com/llvm/llvm-project/pull/66153
https://github.com/llvm/llvm-project/pull/72694

The last one, https://github.com/llvm/llvm-project/pull/72694, eventually removed it. So `std::basic_string<unsigned char>` is not allowed anymore. This removes all uses of `std::basic_string<unsigned char>` from embind.

This needs to be done to update libc++ to LLVM 19 (#22994). I'm uploading this as a separate PR because this removes a functionality from embind.


brendandahl:
I suppose this could break some people, though I'm not aware of any active users of it. For c++20, we could change it to register `char8_t`, however I lean towards removing it and bringing it back if needed.

If we do remove it, we should also update the code [here](https://github.com/emscripten-core/emscripten/blob/58889f9f20f74d0dbaed7d49025c49b864359ae1/src/embind/embind.js#L500). The condition `(name === "std::string")` will always be true then, so we can use the preprocessor macro to only emit the code in the corresponding branches of the `if (stdStringIsUTF8)`.

test/embind/embind.test.js
```diff
@@ -449,33 +449,6 @@ module({
             assert.equal('ABCD', e);
         });
 
-        test("can pass Uint8Array to std::basic_string<unsigned char>", function() {
-            var e = cm.emval_test_take_and_return_std_basic_string_unsigned_char(new Uint8Array([65, 66, 67, 68]));
-            assert.equal('ABCD', e);
-        });
```
sbc100, line 455: Should we change the test such that it tests that this does not work?

aheejin:
> If we do remove it, we should also update the code [here](https://github.com/emscripten-core/emscripten/blob/58889f9f20f74d0dbaed7d49025c49b864359ae1/src/embind/embind.js#L500). The condition `(name === "std::string")` will always be true then, so we can use the preprocessor macro to only emit the code in the corresponding branches of the `if (stdStringIsUTF8)`.

I tried to replace `if (stdStringIsUTF8) { then_body } else { else_body }` with
```js
#if EMBIND_STD_STRING_IS_UTF8
then_body
#else
else_body
#endif
```
but `stdStringIsUTF8` variable is also being used elsewhere in that file too:
https://github.com/emscripten-core/emscripten/blob/58889f9f20f74d0dbaed7d49025c49b864359ae1/src/embind/embind.js#L554
https://github.com/emscripten-core/emscripten/blob/58889f9f20f74d0dbaed7d49025c49b864359ae1/src/embind/embind.js#L564
So removing the variable completely and replace its usage with the `EMBIND_STD_STRING_IS_UTF8` predecessor doesn't seem to work. Should I just do
```js
    var stdStringIsUTF8
#if EMBIND_STD_STRING_IS_UTF8
    = true;
#else
    = false;
#endif
```
?

test/embind/embind.test.js
aheejin: I tried that, and to do that I needed to revive `emval_test_take_and_return_std_basic_string_unsigned_char` in `embind_test.cpp` and the error message is like `Cannot call emval_test_take_and_return_std_basic_string_unsigned_char due to unbound types: NSt3__212basic_stringIhNS_11char_traitsIhEENS_9allocatorIhEEEE`, so the first test would be like
```
        test("cannot pass Uint8Array to std::basic_string<unsigned char>", function() {       
            var e = assert.throws(TypeError, function() {                        
              cm.emval_test_take_and_return_std_basic_string_unsigned_char(new Uint8Array([65, 66, 67, 68]));      
            });                                                                  
            assert.equal('Cannot call emval_test_take_and_return_std_basic_string_unsigned_char due to unbound types: NSt3__212basic_stringIhNS_11char_traitsIhEENS_9allocatorIhEEEE', e.message);
        }); 
```

Are you suggesting to change five deleted tests here like this?

brendandahl:
I think we could get rid of `stdStringIsUTF8` since it's now a static value, but that's going to add a lot more preprocessor junk. Leaving `stdStringIsUTF8` should be fine as closure will inline and remove any dead code.

test/embind/embind.test.js
aheejin: Ping 😀 By the way I wasn't able to get that code to work, it says the throwing error kind is different...

aheejin: Ping ping 😀 I think we can just delete this. The libc++ change means they don't allow invalid specializations anymore and `unsigned char` is one of many possible invalid specializations after all (Even `basic_string<int>` was possible before), and I don't think we can cover all invalid specializations by error tests anyway.

sbc100:
I'll defer to brendan on for the lgtm

brendandahl (approved):
r+ but add a note in the changelog that this no longer supported.

aheejin:
> r+ but add a note in the changelog that this no longer supported.

Done

PR #23017, opened 2024-11-26
Refactor lookupPath to avoid recursion. NFC

hoodmane:
This removes the recursion from `lookupPath` and in my opinion makes the control flow much more explicit and comprehensible.

PR #23071, opened 2024-12-04
Restore changelog entry removed in #23017. NFC

kleisauke:


kripken (approved):

kripken:
Good catch, not sure how that happened. We need to be more careful I guess...

sbc100:
I think it might be because we have `ChangeLog.md merge=union` in our `.gitattributes`.  Normally its useful.

PR #23072, opened 2024-12-04
Fix path resolution for symlinks

hoodmane:
If `/a/link` is a symlink to directory `/b/c` then `/a/link/..` should be resolved to `/b/c/..` which is `/b`. Currently, we cancel out `link/..` to get `/a`. The problem is that we apply `PATH_FS.resolve` and `PATH_FS.normalize` to paths. These functions cancel `..` incorrectly.

This at least partially handles the situation. `lookupPath` is modified to avoid calls that call `PATH_FS.normalize()`. We check that `mkdir`, `open`, and `stat` now work correctly.

### DIFF
--- a/src/library_fs.js
+++ b/src/library_fs.js
@@ -171,15 +171,17 @@ FS.staticInit();
     // paths
     //
     lookupPath(path, opts = {}) {
-      path = PATH_FS.resolve(path);
-
       if (!path) return { path: '', node: null };
       opts.follow_mount ??= true
+
+      if (!PATH.isAbs(path)) {
+        path = FS.cwd() + "/" + path;
+      }
 
       // limit max consecutive symlinks to 40 (SYMLOOP_MAX).
       linkloop: for (var nlinks = 0; nlinks < 40; nlinks++) {
         // split the absolute path
-        var parts = path.split('/').filter((p) => !!p);
+        var parts = path.split('/').filter((p) => !!p && (p !== "."));
 
         // start at the root
         var current = FS.root;
@@ -190,6 +192,12 @@ FS.staticInit();
           if (islast && opts.parent) {
             // stop resolving
             break;
+          }
+
+          if (parts[i] === "..") {
+            current_path = PATH.dirname(current_path);
+            current = current.parent;
+            continue;
           }
 
           current = FS.lookupNode(current, parts[i]);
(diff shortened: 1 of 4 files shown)


src/library_syscall.js
```diff
@@ -38,7 +38,10 @@ var SyscallsLibrary = {
         }
         return dir;
       }
-      return PATH.join2(dir, path);
+      if (!dir.endsWith("/")) {
+        dir += "/"
+      }
+      return dir + path;
```
sbc100, line 44: How is this different to the old code?

test/test_core.py
```diff
@@ -5925,6 +5925,20 @@ def test_fs_64bit(self):
...
+    'nodefs': (['-DNODEFS', '-lnodefs.js'],),
+    'noderawfs': (['-sNODERAWFS'],),
+  })
+  def test_fs_symlink_resolution(self, args):
+    nodefs = '-DNODEFS' in args or '-sNODERAWFS' in args
+    if self.get_setting('WASMFS'):
+      if nodefs:
+        self.skipTest('NODEFS in WasmFS')
+      self.set_setting('FORCE_FILESYSTEM')
+    self.do_runf('fs/test_symlink_resolution.c', 'success', emcc_args=args)
```
sbc100, line 5940: Can you name this file that same as the test (i.e. include the `fs_` part)

test/fs/test_symlink_resolution.c
```diff
@@ -0,0 +1,57 @@
...
+    struct stat statBuf;
+    assert(stat("a/link/../x.txt", &statBuf) == 0);
+    assert(stat("b/x.txt", &statBuf) == 0);
+    makedir("a/link/../d");
+    assert(stat("a/link/../d", &statBuf) == 0);
+    assert(stat("b/d", &statBuf) == 0);
+
+    assert(truncate("a/link/../x.txt", 0) == 0);
+    assert(chmod("a/link/../x.txt", 0777) == 0);
+    printf("success\n");
```
sbc100, line 56: 2 space indentation.

src/library_syscall.js
```diff
@@ -826,7 +829,6 @@ var SyscallsLibrary = {
     path = SYSCALLS.calculateAt(dirfd, path);
     // remove a trailing slash, if one - /a/b/ has basename of '', but
     // we want to create b in the context of this function
-    path = PATH.normalize(path);
     if (path[path.length-1] === '/') path = path.substr(0, path.length-1);
```
sbc100, line 832: Should this now be `while` instead of `if` to handle multiple tailing slashes?

src/library_fs.js
```diff
@@ -192,6 +194,12 @@ FS.staticInit();
             break;
           }
 
+          if (parts[i] === "..") {
```
sbc100, line 197: Single quotes (here and elsewhere)

src/library_syscall.js
hoodmane: We can actually just delete this, it works fine to have a trailing slash. `basename()` does:
```js
path = path.replace(/\/$/, "");
```

src/library_syscall.js
hoodmane: That'll probably work fine, I put it in and will let the CI say.

sbc100 (approved):

PR #23074, opened 2024-12-04
Fix renaming a file on top of an existing file in memfs and nodefs

hoodmane:


src/library_memfs.js
```diff
@@ -205,6 +205,10 @@ addToLibrary({
             }
           }
         }
+        try {
+          new_node = FS.lookupNode(new_dir, new_name);
+          FS.hashRemoveNode(new_node);
+        } catch (e) {}
```
sbc100, line 211: Can do just do this the try/catch once above?
e.g.

```
var new_node;
try {
   new_node = FS.lookupNode(new_dir, new_name);
} catch (e) {}

if (new_node) {
   if (FS.isDir(old_node.mode)) {
     ...
   }
   FS.hashRemoveNode(new_node);
}
```

hoodmane: Done.

test/fs/test_rename_on_existing.c
```diff
@@ -0,0 +1,55 @@
...
+}
+
+
+void setup() {
+  makedir("working");
+#if defined(__EMSCRIPTEN__) && defined(NODEFS)
+  EM_ASM(FS.mount(NODEFS, { root: '.' }, 'working'));
+#endif
+  changedir("working");
+}
```
sbc100, line 45: This whole setup function that the makedir/changedir function can be if `__EMSCRIPTEN__` too.

sbc100 (approved):

test/test_core.py
```diff
@@ -5935,6 +5935,16 @@ def test_fs_64bit(self):
...
 
+  @parameterized({
+    '': ([],),
+    'nodefs': (['-DNODEFS', '-lnodefs.js'],),
+    'noderawfs': (['-sNODERAWFS'],)
+  })
+  def test_fs_rename_on_existing(self, args):
+    if self.get_setting('WASMFS'):
+      self.set_setting('FORCE_FILESYSTEM')
+    self.do_runf('fs/test_rename_on_existing.c', 'success', emcc_args=args)
```
sbc100, line 5946: Can you make the filename match the test name (i.e. include the `_fs` part).

src/library_memfs.js
```diff
@@ -192,18 +192,18 @@ addToLibrary({
...
+          new_node = FS.lookupNode(new_dir, new_name);
+        } catch (e) {}
+        if (new_node) {
+          if (FS.isDir(old_node.mode)) {
+            // if we're overwriting a directory at new_name, make sure it's empty.
             for (var i in new_node.contents) {
               throw new FS.ErrnoError({{{ cDefs.ENOTEMPTY }}});
             }
           }
+          FS.hashRemoveNode(new_node);
```
sbc100, line 206: Actually can we move this whole `if` block inside the `try` block and then remove the `if`?

hoodmane: Sure. Though these blanket catches make debugging kind of hard. It would be nice to have a variant of `lookupNode` that returns `undefined` if the node is missing...

hoodmane: Wait, we can't throw inside the catch block so that doesn't work.

sbc100: Why can't we throw inside that catch block?

sbc100: > Sure. Though these blanket catches make debugging kind of hard. It would be nice to have a variant of `lookupNode` that returns `undefined` if the node is missing...

Maybe we should do that first?

hoodmane:
@sbc100 can this one merge now?

sbc100 (approved):

src/library_memfs.js
sbc100: Maybe we can clean this up as a followup?

PR #23083, opened 2024-12-05
Change log level in library_glemu.js

mzukovec:
Change the log level from error to warning during GL emulation.

It's a bit annoying since the console log receives a lot of error level messages, which are actually warnings.

### DIFF
--- a/src/library_glemu.js
+++ b/src/library_glemu.js
@@ -185,9 +185,9 @@ var LibraryGLEmulation = {
 
 
       // Add some emulation workarounds
-      err('WARNING: using emscripten GL emulation. This is a collection of limited workarounds, do not expect it to work.');
+      console.warning('WARNING: using emscripten GL emulation. This is a collection of limited workarounds, do not expect it to work.');
 #if GL_UNSAFE_OPTS == 1
-      err('WARNING: using emscripten GL emulation unsafe opts. If weirdness happens, try -sGL_UNSAFE_OPTS=0');
+      console.warning('WARNING: using emscripten GL emulation unsafe opts. If weirdness happens, try -sGL_UNSAFE_OPTS=0');
 #endif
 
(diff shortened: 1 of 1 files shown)


src/library_glemu.js
```diff
@@ -185,9 +185,9 @@ var LibraryGLEmulation = {
 
 
       // Add some emulation workarounds
-      err('WARNING: using emscripten GL emulation. This is a collection of limited workarounds, do not expect it to work.');
+      console.warning('WARNING: using emscripten GL emulation. This is a collection of limited workarounds, do not expect it to work.');
 #if GL_UNSAFE_OPTS == 1
-      err('WARNING: using emscripten GL emulation unsafe opts. If weirdness happens, try -sGL_UNSAFE_OPTS=0');
+      console.warning('WARNING: using emscripten GL emulation unsafe opts. If weirdness happens, try -sGL_UNSAFE_OPTS=0');
```
sbc100, line 190: Do you see these more than once in you program?  We could use the warnOnce helper here maybe?

mzukovec: There are at least 10 such warnings. I'll use the warnOnce.

mzukovec: Hmm, this message shows up for each JS wroker, so I don't think that warnOnce will help here?

sbc100: Hmm.. why is GL being initialized on each worker?   Is that intended?

mzukovec: We use the predefined thread pool (`-sPTHREAD_POOL_SIZE='navigator.hardwareConcurrency`) in order to synchronously spawn threads within our code. I suspect that worker initialization automatically triggers GL init method?

sbc100: Indeed that seems to be the case.   I wonder if we can/should avoid initializing GL on threads that are not using GL (i.e. can we call init() lazily?)

PR #23085, opened 2024-12-05
Python lint: Use ruff instead of flake8. NFC

cclauss:
https://docs.astral.sh/ruff is an extremely fast Python linter and code formatter, written in Rust.  With over 800 linting rules, ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [Black](https://github.com/psf/black), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [pyupgrade](https://pypi.org/project/pyupgrade/), [autoflake](https://pypi.org/project/autoflake/), and more, all while executing tens or hundreds of times faster than any individual tool.

Set some upper limits on code complexity using ruff rules for McCabe cyclomatic complexity and Pylint.

% `git grep flake8`
% `flake8-to-ruff .flake8 > pyproject.toml`
% `validate-pyproject pyproject.toml`
% `pyproject-fmt pyproject.toml`
% `ruff check`

---

The ___Required___ `ci/circleci: flake8` job was renamed to `ci/circleci: ruff` so the ___Required___ label should be removed from the old job and added to the new one.

---

% [`ruff rule C901`](https://docs.astral.sh/ruff/rules/complex-structure)
# complex-structure (C901)

Derived from the **mccabe** linter.

## What it does
Checks for functions with a high `McCabe` complexity.

## Why is this bad?
The `McCabe` complexity of a function is a measure of the complexity of
the control flow graph of the function. It is calculated by adding
one to the number of decision points in the function. A decision
point is a place in the code where the program has a choice of two
or more paths to follow.

Functions with a high complexity are hard to understand and maintain.

## Example
```python
def foo(a, b, c):
    if a:
        if b:
            if c:
                return 1
            else:
                return 2
        else:
            return 3
    else:
        return 4
```

Use instead:
```python
def foo(a, b, c):
    if not a:
        return 4
    if not b:
        return 3
    if not c:
        return 2
    return 1
```

## Options
- `lint.mccabe.max-complexity`


cclauss:
Please rerun the failing `test-browser-chrome-wasm64-4gb` job on CircleCI.

The ___Required___ `ci/circleci: flake8` job was renamed to `ci/circleci: ruff` so the ___Required___ label should be removed from the old job and added to the new one.

emcc.py
```diff
@@ -1127,7 +1127,13 @@ def version_string():
   return f'emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) {utils.EMSCRIPTEN_VERSION}{revision_suffix}'
 
 
-def parse_args(newargs):
+def parse_args(newargs):  # noqa: C901, PLR0912, PLR0915
+  """
```
sbc100, line 1131: I think docstrings like this are always supposed to start with the first line on the same like as the opening tipple quote.

See https://peps.python.org/pep-0257/

emcc.py
```diff
@@ -1127,7 +1127,13 @@ def version_string():
...
+  Future modifications should consider refactoring to reduce complexity.
+  * The McCabe cyclomatiic complexity is currently 117 vs 10 recommended.
+  * There are currently 115 branches vs 12 recommended.
+  * There are currently 302 statements vs 50 recommended.
```
sbc100, line 1135: In any case, can we not add these comments?  I'm not sure they are useful.

test/jsrun.py
```diff
@@ -106,7 +106,8 @@ def run_js(filename, engine, args=None,
         stderr=stderr,
         cwd=cwd,
         timeout=timeout,
-        universal_newlines=True)
+        universal_newlines=True,
+        check=skip_check)
```
sbc100, line 110: This looks like maybe a semantic change.. did ruff make this?

test/runner.py
```diff
@@ -309,11 +309,7 @@ def load_test_suites(args, modules, start_at, repeat):
 
 
 def flattened_tests(loaded_tests):
-  tests = []
-  for subsuite in loaded_tests:
-    for test in subsuite:
-      tests.append(test)
-  return tests
+  return [test for subsuite in loaded_tests for test in subsuite]
```
sbc100, line 312: I think we should avoid nested list comprehensions

tools/emscripten.py
```diff
@@ -368,13 +368,13 @@ def emscript(in_wasm, out_wasm, outfile_js, js_syms, finalize=True, base_metadat
     # check_call doesn't support the `input` argument.
     if asm_consts:
       validate = '\n'.join([f'var tmp = {f};' for _, f in asm_consts])
-      proc = subprocess.run(config.NODE_JS + ['--check', '-'], input=validate.encode('utf-8'))
+      proc = subprocess.run(config.NODE_JS + ['--check', '-'], input=validate.encode('utf-8'), check=False)
```
sbc100, line 371: Why this?  check is false by default isn't it?

sbc100 (commented):
Sounds like ruff is indeed an improvement over flake8!

Is there any stuff that flake8 does that ruff does now catch?

sbc100:
Can we stick with `.ruff.toml` for now (i.e. a specific config file that is clearly for ruff alone) (at least until we have some other reason to have `project.toml`)

emcc.py
cclauss: I would very much like to leave these comments in place because the complexity numbers on these 5 functions are radically outside the recommended ranges.  The "Why is this bad?" section of these rules states the obvious "Functions with a high complexity are hard to understand and maintain".  The vast majority of developers do not know the `noqa` id by heart.  The purpose of these docstrings is to help both contributors and code reviewers notice that there are extraordinary maintainability issues so that careful review and robust tests need to be added around any change.  It might serve as a learning experience for both mentors and new contributors.

test/runner.py
cclauss: Reverted but it is good for Python developers to learn how to read comprehensions.

tools/emscripten.py
sbc100: I think just leaving it out is fine/clear.

sbc100:
Lets just go with `.ruff.toml` initially maybe?

emcc.py
cclauss: Not auto-generated.  If the numbers are not discussed and reanalyzed in the code review process they will become inaccurate.  This is the advantage of adding the docstrings instead of just `# noqa: C901, PLR0912, PLR0915`.  Experience tells me than any developer that lowers these numbers will proudly update the docstring.  If developers and code reviewers allow the numbers to rise then they will not modify the numbers but they will still be a strong signal to their more diligent successors.

These numbers can be easily recalculate by explicitly overriding the config file. `ruff check --select=C901,PLR091`

sbc100: OK, since there are only a few of these I'm ok adding them.

sbc100: Should we had to this "Run `ruff` with the above `noqa` comment removed to update these numbers".  I fear that without such a hit these comments will never get updated.

test/jsrun.py
sbc100: Can we simply revert this?

tools/emscripten.py
sbc100: Can we revert revert this hunk.

sbc100 (approved):
lgtm % comments

cclauss:
* [x] `ruff check --ignore=PLW1510`
* [x] `To revalidate these numbers, run `ruff check --select=C901,PLR091`.`

cclauss:
This PR will not auto-merge because the Required `ci/circleci: flake8` job no longer exists.

sbc100:
@cclauss it looks like lost the check for the correct number of newlines between methods and functions: https://github.com/emscripten-core/emscripten/pull/23139/files#r1883095264

cclauss:
* https://github.com/emscripten-core/emscripten/pull/23139/files#r1883503464

sbc100:
It looks like we also regressed on the ability to detect trailing whitespace in function calls:

```
$ cat test.py 
print(1, 2 )
$ ruff check test.py
All checks passed!
$ flake8 test.py
test.py:1:11: E202 whitespace before ')'
```

Is this just something we need to enable?

sbc100:
I guess E202 is in preview too: https://docs.astral.sh/ruff/rules/whitespace-before-close-bracket/.   Somewhat worrying that such simple checks are in preview.. and that we have to enable them explicitly on the command line.

cclauss:
This is solved in #23154.

% `echo "print(1, 2 )" | pipx run ruff format -`
```
print(1, 2)
```
Flake8 was created before great code formatters but Ruff was not.
I don't think the Ruff Team puts too many cycles on rules that are better solved by black or ruff format.

sbc100:
It would have been good to know that before we made the transition to ruff.

PR #23096, opened 2024-12-06
Error on `--emrun` with `EXIT_RUNTIME=0`

sbc100:
The emrun post.js file depend on `addOnExit` to report the exit of the program.  We were already forcing `-sEXIT_RUNTIME` but this adds an error if the user tries to explicitly disable it.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -664,7 +664,11 @@ def phase_linker_setup(options, state, newargs):  # noqa: C901, PLR0912, PLR0915
   if options.emrun:
     options.pre_js.append(utils.path_from_root('src/emrun_prejs.js'))
     options.post_js.append(utils.path_from_root('src/emrun_postjs.js'))
+    if settings.MINIMAL_RUNTIME:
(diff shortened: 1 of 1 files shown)


tools/link.py
```diff
@@ -656,7 +656,11 @@ def phase_linker_setup(options, state, newargs):
   if options.emrun:
     options.pre_js.append(utils.path_from_root('src/emrun_prejs.js'))
     options.post_js.append(utils.path_from_root('src/emrun_postjs.js'))
+    if settings.MINIMAL_RUNTIME:
+      exit_with_error('--emrun is not compatible with MINIMAL_RUNTIME')
     # emrun mode waits on program exit
+    if user_settings.get('EXIT_RUNTIME') == '0' or user_settings.get('NO_EXIT_RUNTIME') == '1':
```
kripken, line 662: We normalize this, so `NO_EXIT_RUNTIME` will never be a key.

sbc100: Hmm, you are right, not sure what I was thinking there.

kripken (approved):

PR #23100, opened 2024-12-06
Simplify asyncLoad. NFC

sbc100:
Turns out non of the callers were relying the adding of the run dependencies here.  There are 4 call sites total:

- emscripten_async_wget_data: Explicitly sets noRunDep
- emscripten_wget_data: Explicitly sets noRunDep
- loadDynamicLibrary: The outer loadDylibs already has an explicit calls to addRunDependency/removeRunDependency.
- FS_createPreloadedFile: Already has an explicit calls to addRunDependency/removeRunDependency.


sbc100:
Landing TBR.  Hopefully this is a completely straight forward change and I wan to unblock testing / upload / reviewing followups.

PR #22987, opened 2024-11-21
[library_sockfs.js] Cleanup buffer handling in sendmsg

sbc100:
Extracting the `buffer.slice` expression made the code cleanup and showed the double `new Uint8Array` to be redundant (AFAICT).

PR #23103, opened 2024-12-07
Fix for flakiness in socket tests. NFC

sbc100:
In #22987 I added a new veriant of the test_nodejs_sockets_echo test and chose the next consecutive port number, but it seems that each test ends up using two ports numbers so this could conflicted with the tests that were using both the previous and next number in the sequence (59164 and 59166).


sbc100:
Landing TBR to fix the ongoing take flakiness.

PR #23110, opened 2024-12-09
Refactor startup code to avoid duplication between wasm workers and pthreads. NFC

sbc100:
Split out from #23106


test/other/codesize/test_codesize_minimal_pthreads.jssize
```diff
@@ -1 +1 @@
-8703
+8716
```
kripken, line 1: Probably noise, but might be worth seeing why this regressed.

kripken (approved):

PR #23112, opened 2024-12-09
Remove `runDependencies` handling code from  `emscripten_async_load_script`

sbc100:
This handling of `runDependencies` was included when this code was first added in 1fc6762e3.  While there may have been reason for this back then I can't see how or why this is useful today.  We don't have any internal usage of `emscripten_async_load_script` so there should be no way to call it before the program is running.

### DIFF
--- a/src/library_browser.js
+++ b/src/library_browser.js
@@ -661,28 +661,16 @@ var LibraryBrowser = {
       return;
     }
 #endif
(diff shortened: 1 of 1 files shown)


sbc100:
Actually we do have tests that cover this behaviour.  See test_emscripten_async_load_script

PR #23119, opened 2024-12-10
[test] Remove EMSCRIPTEN_KEEPALIVE from _ReportResult. NFC

sbc100:
On the emscripten side we have `reportResultToServer` already.

### DIFF
--- a/test/webaudio/create_webaudio.c
+++ b/test/webaudio/create_webaudio.c
@@ -23,9 +23,7 @@ int main()
 		startButton.onclick = () => {
 			if (audioContext.state != 'running') {
 				audioContext.resume();
-#ifdef REPORT_RESULT
-				__ReportResult(0);
-#endif
+				reportResultToServer(0);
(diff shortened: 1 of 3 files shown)


test/webaudio/create_webaudio.c
```diff
@@ -23,9 +23,7 @@ int main()
 		startButton.onclick = () => {
 			if (audioContext.state != 'running') {
 				audioContext.resume();
-#ifdef REPORT_RESULT
-				__ReportResult(0);
-#endif
+				reportResultToServer(0);
```
kripken, line 26: Will this not report more than before? Or was `REPORT_RESULT` always defined?

kripken (approved):

test/webaudio/create_webaudio.c
sbc100: I don't think anyone is ever going to want to build this test without browser reporting included

PR #23120, opened 2024-12-10
Use async/await for internal readAsync function. NFC

sbc100:
Followup to #23104.


src/node_shell_read.js
```diff
@@ -9,18 +9,17 @@ readBinary = (filename) => {
...
-    fs.readFile(filename, binary ? undefined : 'utf8', (err, data) => {
-      if (err) reject(err);
-      else resolve(binary ? data.buffer : data);
-    });
-  });
+  var ret = fs.readFileSync(filename, binary ? undefined : 'utf8');
+#if ASSERTIONS
+  assert(!binary || Buffer.isBuffer(ret));
+#endif
+  return ret;
```
sbc100, line 24: The async keyword automatically makes the return value into a promise here.

src/shell.js
```diff
@@ -300,11 +300,7 @@ if (ENVIRONMENT_IS_SHELL) {
     return data;
   };
 
-  readAsync = (f) => {
-    return new Promise((resolve, reject) => {
-      setTimeout(() => resolve(readBinary(f)));
-    });
-  };
+  readAsync = async (f) => readBinary(f);
```
sbc100, line 303: The async keyword automatically makes the return value into a promise here.

src/node_shell_read.js
```diff
@@ -9,18 +9,17 @@ readBinary = (filename) => {
   filename = isFileURI(filename) ? new URL(filename) : filename;
   var ret = fs.readFileSync(filename);
 #if ASSERTIONS
-  assert(ret.buffer);
+  assert(Buffer.isBuffer(ret));
```
sbc100, line 12: I think this is more accurately reflects what this assertion was trying to prove.  i.e. that the returned value is a node `Buffer` object.

kripken: That sounds right, but I recall some differences between node versions... we do have coverage of multiple ones on CI though, so if it passes we should be ok I guess.

kripken (approved):

src/node_shell_read.js
sbc100: I was trying to figure out what this assert could possibly be for... is it to detect older node versions that don't return a buffer?  If so, shouldn't we just do that with the version check?  `Buffer.isBuffer` has been around forever so I certainly think it should be fine to use: https://nodejs.org/api/buffer.html#static-method-bufferisbufferobj

kripken: Oh right, yeah, it was whether a Buffer was returned or not. Yeah, this lgtm.

PR #23126, opened 2024-12-10
Default to -mno-bulk-memory-opt

sbc100:
This should fix the recent CI failures on the test-node-compat bot. Once #22873 lands this can, of course, be removed.


sbc100:
@sunfishcode, its maybe little annoying the users who have `-mno-bulk-memory` can still get bulk memory instruction added.  Should we have `-mno-bulk-memory` imply `-mno-bulk-memory-opt`?

sbc100:
PTAL, to green the CI again

aheejin (approved):

test/other/codesize/test_codesize_files_wasmfs.imports
```diff
@@ -11,5 +11,6 @@ j (_wasmfs_get_preloaded_child_path)
 k (_wasmfs_get_num_preloaded_files)
 l (_wasmfs_get_num_preloaded_dirs)
 m (_wasmfs_copy_preloaded_file_data)
-n (_abort_js)
-o (random_get)
+n (_emscripten_memcpy_js)
```
kripken, line 14: Why did this change?

sbc100: Because without `-mno-bulk-memory-opt` (part of this change) we were defaulting to using the bulk memory instruction here (by mistake).

sbc100: The changes here originally made it in via a rebaseline that I did this morning: ea874233bbef6a24ca1d3caaafb422d8cdfabdea.  I normally don't land re-baselines without review if they include symbol changes like this, but this one slipped through.

kripken: Oh, I see, I didn't review the earlier rebaseline, so this seemed like a regression to me...

Maybe we should add some automation for the automatic rebaselining? Like a script that does it + does a check if it is "trivial". Anything aside from size changes within some range could show an error to the user and suggest investigation?

sbc100: I think that's a great idea.

Then contributors could run the script on their PRs too and we could maybe remove the slack on those tests?

sbc100: See https://github.com/emscripten-core/emscripten/issues/23146

PR #23127, opened 2024-12-10
Add some slack to test_extern_prepost. NFC

sbc100:
This test started failing on the macOS roller.

The js size on this test on my machine is 5006 which is very close the limit of 100 * 50 that this test was checking against.

The main change here is to reduce the limit to from 5k (100 * 50) to 2.5k (50 * 50).


sbc100:
I'm not sure why this would fail on the macOS rollor but not for me locally.

kripken (approved):

Issue #22971, opened 2024-11-20
Audio Worklets fail with STRICT


cwoffenden:
This is in the latest Emscripten:

```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.73-git (299be0bbfbbadfa926cb69380f0fffa3703b26c7)
```

Building any of the Audio Worklet tests with `-sSTRICT` compiles but fails to run:

```
emcc -sAUDIO_WORKLET -sWASM_WORKERS -sSTRICT audio_worklet_tone_generator.c
```

The browser errors being variations of missing API calls, e.g.:

```
Aborted(`Module.wasmMemory` was supplied but `wasmMemory` not included in INCOMING_MODULE_JS_API)
```

The suggestion here being to add it:

https://github.com/emscripten-core/emscripten/pull/22753#discussion_r1849053146

(This can be assigned to me and I'll look at it)

sbc100:
I think the fix here is to explicitly add the needed exports to `INCOMING_MODULE_JS_API`.   Something like this: https://github.com/emscripten-core/emscripten/blob/e1ce10df8cc77f084c3ea24cf674a0f856ab43c4/tools/link.py#L1214-L1216

e.g:

```
if settings.AUDIO_WORKLET:
  # The following incomming API elements are using when starting a audio worklet.
  settings.INCOMING_MODULE_JS_API += [ .. .]
```

cwoffenden:
Small fix in #23130. As per the comment there, this error wasn't getting caught by the parameterised `STRICT` test (because I think, TBC, the test exits before fully starting).

cwoffenden:
I’ll close this now the fix has landed and open another for the test failing.

PR #23130, opened 2024-12-11
[AUDIO_WORKLET] Fix when compiling with STRICT

cwoffenden:
Fix for #22971.

This highlights that the AW tests weren't catching this or other issues in the parameterised tests. Running, for example:
```
test/runner browser.test_audio_worklet_strict
```
Passes, but if compiled with the same settings doesn't run.


cwoffenden:
(The build errors look unrelated)

sbc100:
I guess we can land this now and then ensure that the tests are correctly reporting errors as a followup?

sbc100 (approved):

cwoffenden:
> I guess we can land this now and then ensure that the tests are correctly reporting errors as a followup?

Yes, I plan to look at the tests soon-ish (after the API change draft).

PR #23132, opened 2024-12-11
[tsgen] Fix overriding Embind's ClassHandle clone method.

brendandahl:
The return type of `clone` should actually be `ThisType<this>` to better match what the JS is actually doing. This also makes it possible to override the clone method.

### DIFF
--- a/test/other/embind_tsgen_ignore_1.d.ts
+++ b/test/other/embind_tsgen_ignore_1.d.ts
@@ -31,7 +31,7 @@ export interface ClassHandle {
   delete(): void;
   deleteLater(): this;
   isDeleted(): boolean;
-  clone(): this;
+  clone(): ThisType<this>;
 }
 export interface Test extends ClassHandle {
   x: number;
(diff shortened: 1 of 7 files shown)


test/other/embind_tsgen_ignore_1.d.ts
```diff
@@ -31,7 +31,7 @@ export interface ClassHandle {
   delete(): void;
   deleteLater(): this;
   isDeleted(): boolean;
-  clone(): this;
+  clone(): ThisType<this>;
```
sbc100, line 34: What does this expression do/mean?

brendandahl: IIUC, it's magic marker that changes how the `this` type is bound using the context it was declared instead of where it was called from. This is a workaround for some g3 conflicts, but I'm actually leaning towards abandoning this PR and asking that they don't try to override the clone method.

sbc100: Seems a little magical but LGTM if it fixes the issue!

PR #23137, opened 2024-12-12
Make open O_CREATE mode 0 work

hoodmane:
In nodefs we first create the node with permissions 0 and then try to open it. The open fails because we don't have read permissions.


src/library_fs.js
```diff
@@ -1126,6 +1126,9 @@ FS.staticInit();
       if (stream.stream_ops.open) {
         stream.stream_ops.open(stream);
       }
+      if (created) {
+        FS.chmod(node, mode & 0o777);
+      }
```
sbc100, line 1131: Why not just rely on `mknod` to set the correct mode above?  It seems odd to explicitly call `mknod` with 777 and then call `chmod` afterwards, no?

src/library_fs.js
```diff
@@ -1076,7 +1076,7 @@ FS.staticInit();
           }
         } else {
           // node doesn't exist, try to create it
-          node = FS.mknod(path, mode, 0);
+          node = FS.mknod(path, mode | 0o777, 0);
```
sbc100, line 1079: Is there some reason why mknod with zero permissions doesn't work?

test/fs/test_fs_open_no_permissions.c
```diff
@@ -0,0 +1,44 @@
...
+int main() {
+    setup();
+    
+    int res = open("a", O_CREAT, 0);
+    printf("error: %s\n", strerror(errno));
+    assert(res >= 0);
+    struct stat st;
+    assert(stat("a", &st) == 0);
+    assert((st.st_mode & 0777) == 0);
+    printf("success\n");
```
sbc100, line 43: 2 space indentation.

src/library_fs.js
hoodmane: Because then if we call `open("some_file", O_CREAT, 0x000)` in the nodefs `stream.stream_ops.open(stream)` fails. We call `mknod` to produce a native node with no permissions and then separately try to open it but we didn't give ourselves permission to open it. Instead, for this to work on nodefs we either need to detect this situation and make a single native call to `open("some_file", O_CREAT, 0x000)` and use the file descriptor it returns or to make the change I did here.

src/library_fs.js
hoodmane: The problem is later we try to open it and if we don't have read/write permissions that open will fail.

sbc100: Perhaps add a comment here.

```
// Ignore the permission bits here to ensure we can `open` this new file below.
// We use chmod below the apply the permissions once the file is open.
```

This makes me wonder if we should be able open and create in a single FS API call.. maybe our API has the wrong shape.  No need to fix as part of this PR.

hoodmane: >  maybe our API has the wrong shape

I agree. I looked into this briefly, but it's confusing. Maybe the linux fs apis handle this better.

hoodmane:
Well I think the right test would be `test_fcntl_open.c`. But yet again, that test is currently not running in nodefs or noderawfs and trying to run it on noderawfs reveals a seemingly real failure. It looks fixable, so I'll put a comment saying to merge the tests here and then in a followup I'll try to fix that test and put them together.

sbc100 (approved):

Issue #14449, opened 2021-06-14
Auto-format Python code


rth:
As mentioned by @sbc100 in https://github.com/emscripten-core/emscripten/issues/7181#issuecomment-425252819

> We could also encourage the use of yapf which actually makes all the formatting decisions for you: https://github.com/google/yapf. Similar to clang-format.

To be complete, another popular tool now is https://github.com/psf/black though it won't work with 2 space indentation.

This might make different Python scripts / modules a bit more readable and easier to contribute to, in particular when used in combination with [pre-commit](https://pre-commit.com/). 

Though it might make using git blame [1] and conflict resolution in PRs [2] slightly harder. 

[1] https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame
[2] https://github.com/scikit-learn/scikit-learn/pull/20260#issuecomment-860992307 

(links are for black, but I imagine with yapf it would be a similar situation)

cclauss:
`ruff format` has
* https://docs.astral.sh/ruff/settings/#indent-width = 2
* https://docs.astral.sh/ruff/settings/#format_quote-style = "double" | "single" | "preserve"

PR #23139, opened 2024-12-12
Ensure the ino we give to readdir matches the ino we give to stat

hoodmane:


test/test_core.py
```diff
@@ -5900,6 +5900,17 @@ def test_fs_rename_on_existing(self, args):
...
+    '': ([],),
+    'nodefs': (['-DNODEFS', '-lnodefs.js'],),
+    'noderawfs': (['-sNODERAWFS'],)
+  })
+  def test_fs_readdir_ino_matches_stat_ino(self, args):
+    if self.get_setting('WASMFS'):
+      self.set_setting('FORCE_FILESYSTEM')
+    self.do_runf('fs/test_fs_readdir_ino_matches_stat_ino.c', 'success', emcc_args=args)
+
+
```
sbc100, line 5913: Too many newlines here.. flake8 would normally complain about this but we recently switched to ruff.  I wonder if ruff is missing this check?

cclauss: I believe the number of blank lines here is correct.  `test_fs_readdir_ino_matches_stat_ino()` is a top-level function so PEP8 says it should be surrounded by two blank liines.

https://peps.python.org/pep-0008/#blank-lines
> Surround top-level function and class definitions with two blank lines.

% `ruff rule E303`
## Why is this bad?
PEP 8 recommends using blank lines as follows:
- No more than two blank lines between top-level statements.
- No more than one blank line between non-top-level statements.

Note that ruff rule E303 is currently only available in `--preview` mode because of interactions with `isort` and typing stub files (.pyi).

I sense this returns us to the code formatting questions of #14449.  I see that [`yapf`](https://github.com/search?q=repo%3Aemscripten-core%2Femscripten%20yapf) configuration is present in the codebase but is it run on a consistant basis?  If we ran yapf or psf/black or ruff format on a consistant basis then the code formatter could autofix blank lines instead of it being a linter's responsibility.
* #14449

hoodmane: It's not a top level function though it's a class method.

src/library_syscall.js
```diff
@@ -722,16 +721,17 @@ var SyscallsLibrary = {
...
       }
-#if ASSERTIONS
-      assert(id);
-#endif
-      {{{ makeSetValue('dirp + pos', C_STRUCTS.dirent.d_ino, 'id', 'i64') }}};
+      // We use getattr to decide which ino we return to stat, so make sure to
+      // use it to decide which ino we return to readdir too. noderawfs puts the
+      // native inode into node.id and doesn't define node_ops.getattr so use
+      // that as a fallback.
+      var ino = node.node_ops.getattr?.(node).ino ?? node.id;
```
hoodmane, line 733: This `getattr` will fail if the directory doesn't have execute permissions. So perhaps the better approach is to return the virtual file system inode from stat instead of the native inode.

hoodmane:
Okay I updated this again, but now it conflicts with #23058, like several other changes I want to make. So ideally we should merge that one first and then I can update this PR to reflect those changes.

test/test_core.py
sbc100: > If we ran yapf or psf/black or ruff format on a consistent basis then the code formatter could autofix blank lines instead of it being a linter's responsibility.

I think we would always want the linter running CI just in case developers don't run the formatter locally.

I confirmed locally that ruff is not enforcing black line rules today, so we do have a linting regression here.

cclauss: * #23154

sbc100: Can we focus on making sure that `ruff check` reports when the wrong number of blank lines are using?

I'm not complete averse to a full code reformat, but that is much larger change that I think will require a fair amount of dicussion/time... and it seem orthogonal to the fact that `ruff check` didn't report this issue.

cclauss: * #23155

test/fs/test_fs_readdir_ino_matches_stat_ino.c
```diff
@@ -0,0 +1,76 @@
...
+  DIR *dirp;
+  dirp = fdopendir(dirfd);
+  assert(dirp != NULL);
+  struct dirent *ep;
+  struct stat sta, stb;
+  assert(lstat("a", &sta) == 0);
+  assert(lstat("b", &stb) == 0);
+  // Remove execute permission from directory. This prevents us from stat'ing
+  // files in the directory in the implementation of readdir which we tried to
+  // use to fix this.
```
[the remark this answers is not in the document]
hoodmane: I think it's "needed" in the sense that:
* the first implementation I tried passes the entire test suite without this chmod
* it fails when we do this chmod
so I think it's helpful to show what's wrong with that approach. But if you want I can remove it.

test/fs/test_fs_readdir_ino_matches_stat_ino.c
```diff
@@ -0,0 +1,76 @@
...
+
+void setup() {
+#if defined(__EMSCRIPTEN__) && defined(NODEFS)
+  makedir("working");
+  emscripten_debugger();
+  EM_ASM(FS.mount(NODEFS, { root: '.' }, 'working'));
+  changedir("working");
+#endif
+  mkdir("sub", 0775);
+  changedir("sub");
```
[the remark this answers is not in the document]
hoodmane: If we remove the chmod then no. If we keep it, then the test runner hits a permission error when it tries to remove the directory.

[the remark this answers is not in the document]
sbc100: Hmm.. are you sure?  The test runner should be using a very aggressive method to remove the test directory.  See https://github.com/emscripten-core/emscripten/blob/dfe3356cb1860ebf087de33f85a742e8669cce5b/test/common.py#L705-L707

The idea here is that whatever an individual test does in its own directory should always be removable by the test runner itself.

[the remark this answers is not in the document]
hoodmane: I'll try it again and see what happens.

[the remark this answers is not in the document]
hoodmane: I get this:
```
ERROR: test_fs_readdir_ino_matches_stat_ino_nodefs (test_core.core0)
  File "/home/rchatham/Documents/programming/emscripten/test/test_core.py", line 9694, in setUp
    super(TT, self).setUp()
  File "/home/rchatham/Documents/programming/emscripten/test/common.py", line 1121, in setUp
    force_delete_contents(self.working_dir)
  File "/home/rchatham/Documents/programming/emscripten/test/common.py", line 711, in force_delete_contents
    make_dir_writeable(dirname)
  File "/home/rchatham/Documents/programming/emscripten/test/common.py", line 702, in make_dir_writeable
    make_writable(i)
  File "/home/rchatham/Documents/programming/emscripten/test/common.py", line 689, in make_writable
    new_mode = os.stat(path).st_mode | permission_bits
PermissionError: [Errno 13] Permission denied: '/home/rchatham/Documents/programming/emscripten/out/test/test_fs_readdir_ino_matches_stat_ino.wasm'
```
I think there's a few problems: as I understand it, `is_writable` has an incorrect conditional:
```py
  def is_writable(path):
    return (os.stat(path).st_mode & permission_bits) != permission_bits
```
so `not is_writeable(path)` is equivalent to `(os.stat(path).st_mode & permission_bits) == permission_bits`. The second is that `permission_bits` does not have enough permissions. It needs to include an execute bit for both owner and other (but not group). I'd suggest just setting all permissions so we don't have to think too hard:
```py
def make_dir_writeable(dirname):
  # Some tests make files and subdirectories read-only, so rmtree/unlink will not delete
  # them. Force-make everything writable in the subdirectory to make it
  # removable and re-attempt.
  os.chmod(dirname, 0o777)

  for directory, subdirs, files in os.walk(dirname):
    for item in files + subdirs:
      i = os.path.join(directory, item)
      if not os.path.islink(i):
        os.chmod(i, 0o777)
```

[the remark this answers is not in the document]
hoodmane: I pushed a commit with this change, let me know what you think.

test/fs/test_fs_readdir_ino_matches_stat_ino.c
```diff
@@ -0,0 +1,50 @@
...
+#include <errno.h>
+#include <string.h>
+#include <dirent.h>
+
+
+int main() {
+  int res = open("b", O_CREAT, 0777);
+  assert(res >= 0);
+  assert(close(res) == 0);
+  assert(symlink("b", "a") == 0);
```
sbc100, line 17: Maybe put there 4 lines in a `setup` function since they effectively are preparing the directory.

test/fs/test_fs_readdir_ino_matches_stat_ino.c
```diff
@@ -0,0 +1,50 @@
...
+  int dirfd = open(".", O_RDONLY);
+  assert(dirfd > 0);
+  DIR *dirp;
+  dirp = fdopendir(dirfd);
```
sbc100, line 21: Combine these two lines.

test/fs/test_fs_readdir_ino_matches_stat_ino.c
```diff
@@ -0,0 +1,50 @@
...
+      a_ino = ep->d_ino;
+    }
+    if (strcmp(ep->d_name, "b") == 0) {
+      b_ino = ep->d_ino;
+    }
+  }
+  assert(errno == 0);
+  assert(a_ino >= 0);
+  assert(b_ino >= 0);
+  assert(chmod(".", 0775) == 0);
```
sbc100, line 44: Remove this line.

sbc100 (commented):
I wonder if we can make this test part of an existing `readdir` test that than being its own one?  Maybe not.. but just a thought. 

LGTM % comments

hoodmane:
Maybe `test/dirent/test_readdir.c`.

test/fs/test_fs_readdir_ino_matches_stat_ino.c
```diff
@@ -0,0 +1,52 @@
...
+  assert(lstat("a", &sta) == 0);
+  assert(lstat("b", &stb) == 0);
+  // Remove execute permission from directory. This prevents us from stat'ing
+  // files in the directory in the implementation of readdir which we tried to
+  // use to fix this.
+  assert(chmod(".", 0675) == 0);
+  int a_ino = -1;
+  int b_ino = -1;
+  struct dirent *ep;
+  while ((ep = readdir(dirp))) {
```
sbc100, line 35: I think you can do `while (struct dirent *ep = readdir(dirp)) {` ?

Then you don't need the extra braces since its no longer possible to accidentally write `==`.

hoodmane: The `while ()` condition has to be an expression, but this is a statement. You also have to put an expression in the condition slot of a `for()` but the init and the update are allowed to be statements.

test/fs/test_fs_readdir_ino_matches_stat_ino.c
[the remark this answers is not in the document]
hoodmane: Exactly.

hoodmane:
Okay I think I addressed all the comments here.

sbc100 (approved):

hoodmane:
It seems that if the node working directory does not have execute permissions, this makes `fs.readdir()` fail. So I think I need to put back in the subdirectory.

sbc100:
> It seems that if the node working directory does not have execute permissions, this makes `fs.readdir()` fail. So I think I need to put back in the subdirectory.

IIRC that is exactly what the execute bit on a directory means: "can this directory be listed"

test/fs/test_fs_readdir_ino_matches_stat_ino.c
```diff
@@ -0,0 +1,60 @@
...
+#include <dirent.h>
+
+// TODO: Consider merging this with test/dirent/test_readdir.c?
+// That test doesn't work on nodefs, which has to be fixed before these two
+// tests can be merged .
+
+void setup() {
+  // If the node working directory does not have execute permissions, then
+  // `fs.readdir` will fail. For that reason we have to work in a subdirectory
+  // and remove execute permissions from that.
```
sbc100, line 19: I don't understand.. `fs.readdir` should fail on any directory that doesn't have the execute bit set, regardless of weather that is CWD or a sub directory, right?

hoodmane: Apparently not.

hoodmane: I suppose I can drop the chmod, but cding into a subdirectory does fix the test, at least when I run it locally. I don't really have a clear theory as to why since it seems a bit illogical.

sbc100: Ah, yes, I misremembered.  The `x` bit is about traversing the heirachy, not list-ability.  You can list directory than don't have the `x` bit set.  That is what the `r` bit is for.

PR #23143, opened 2024-12-12
Remove unnecessary functions from jsmath.c. NFC

sbc100:
See the comments at the top of `emscripten/js_math.h` for why JS versions of these functions are not needed.

As a followup I plan to map `jsmath.c` functions to `em_math.h` functions instead of using EM_JS here.

See #19284

### DIFF
--- a/system/lib/jsmath.c
+++ b/system/lib/jsmath.c
@@ -23,9 +23,6 @@ CALL_JS_1_TRIPLE(atan, Math.atan)
 CALL_JS_1_TRIPLE(exp, Math.exp)
 CALL_JS_1_TRIPLE(log, Math.log)
 CALL_JS_1_TRIPLE(sqrt, Math.sqrt)
-CALL_JS_1_TRIPLE(fabs, Math.abs)
-CALL_JS_1_TRIPLE(ceil, Math.ceil)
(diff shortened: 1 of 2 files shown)


sbc100:
See https://github.com/emscripten-core/emscripten/blob/1171adaa21048ba8dc82169b9d79952e8754b201/system/include/emscripten/em_math.h#L25-L36

system/lib/jsmath.c
```diff
@@ -50,11 +47,6 @@ CALL_JS_1_IMPL_TRIPLE(round, {
...
 CALL_JS_1_IMPL_TRIPLE(rint,  {
-  function round(x) {
-    return x >= 0 ? Math.floor(x + 0.5) : Math.ceil(x - 0.5);
-  }
+  var round(x) => x >= 0 ? Math.floor(x + 0.5) : Math.ceil(x - 0.5);
   return (x - Math.floor(x) != .5) ? round(x) : round(x / 2) * 2;
 })
-
-double nearbyint(double x) { return rint(x); }
-float nearbyintf(float x) { return rintf(x); }
```
sbc100, line 60: These two are already implement in musl in terms of rint and rintf

kripken (approved):

PR #23148, opened 2024-12-13
Remove rint from jsmath library. NFC

sbc100:
We  use `__builtin_rint` (which lowers to a single `f64.nearest`
instruction) when compiling the musl version of `rint.c` so there there is
no need to use a JS call for this case.

### DIFF
--- a/system/lib/jsmath.c
+++ b/system/lib/jsmath.c
@@ -45,9 +45,3 @@ CALL_JS_2_TRIPLE(pow)
 CALL_JS_1_IMPL_TRIPLE(round, {
   return x >= 0 ? Math.floor(x + 0.5) : Math.ceil(x - 0.5);
 })
-CALL_JS_1_IMPL_TRIPLE(rint,  {
-  function round(x) {
-    return x >= 0 ? Math.floor(x + 0.5) : Math.ceil(x - 0.5);
-  }
-  return (x - Math.floor(x) != .5) ? round(x) : round(x / 2) * 2;
-})


system/lib/jsmath.c
```diff
@@ -33,21 +33,3 @@ CALL_JS_1_TRIPLE(sqrt)
...
-
-#define CALL_JS_1_IMPL(cname, type, impl) \
-  EM_JS(type, JS_##cname, (type x), impl); \
-  type cname(type x) { return JS_##cname(x); }
-
-#define CALL_JS_1_IMPL_TRIPLE(cname, impl) \
-  CALL_JS_1_IMPL(cname, double, impl) \
-  CALL_JS_1_IMPL(cname##f, float, impl)
-
-CALL_JS_1_IMPL_TRIPLE(round, {
```
kripken, line 45: What round will be used instead? musl's does not look efficient,

https://github.com/emscripten-core/emscripten/blob/main/system/lib/libc/musl/src/math/round.c

sbc100: So maybe we just need to update musl's round to use the advice given in https://github.com/emscripten-core/emscripten/blob/1092ec30a3fb1d46b1782ff1b4db5094d3d06ae5/system/include/emscripten/em_math.h#L31?   According that comment we should be able to do better than calling out to JS, right?

sbc100: Oh wait, `fround` is not `round`.

sbc100:
I updated this change to only effect `rint` which is clearly not necessary.

system/lib/jsmath.c
```diff
@@ -45,9 +45,3 @@ CALL_JS_2_TRIPLE(pow)
 CALL_JS_1_IMPL_TRIPLE(round, {
   return x >= 0 ? Math.floor(x + 0.5) : Math.ceil(x - 0.5);
 })
-CALL_JS_1_IMPL_TRIPLE(rint,  {
-  function round(x) {
-    return x >= 0 ? Math.floor(x + 0.5) : Math.ceil(x - 0.5);
-  }
-  return (x - Math.floor(x) != .5) ? round(x) : round(x / 2) * 2;
-})
```
kripken, line 53: Hmm, I am not certain the new code is smaller, as unless I am doing something wrong, even though `rint.c` just does `__builtin_rint`, it doesn't turn into a single instruction. My testcase:
```cpp
#include <emscripten.h>

int main(int argc, char **argv) {
  double x = double(argc + 1000) / double(argc + 100);
  return __builtin_rint(x);
}
```
`emcc -O3 --profiling` leads to
```wat
 (func $main (param $0 i32) (param $1 i32) (result i32)
  (local $2 f64)
  (if
   (f64.lt
    (f64.abs
     (local.tee $2
      (f64.nearest
       (f64.div
        (f64.convert_i32_s
         (i32.add
          (local.get $0)
          (i32.const 1000)
         )
        )
        (f64.convert_i32_s
         (i32.add
          (local.get $0)
          (i32.const 100)
         )
        )
       )
      )
     )
    )
    (f64.const 2147483648)
   )
   (then
    (return
     (i32.trunc_f64_s
      (local.get $2)
     )
    )
   )
  )
  (i32.const -2147483648)
 )
```
Ignoring the division etc., it looks like `rint` is not a single instruction but `f64.nearest, abs, lt, trunc` operations?

sbc100: `rint` does compile to single instruction:

```
$ ar x ./cache/sysroot/lib/wasm64-emscripten/libc.a rint.o
$ wasm-objdump -d -r -x rint.o
...
000059 func[0] <rint>:
 00005a: 20 00                      | local.get 0
 00005c: 9e                         | f64.nearest
 00005d: 0b                         | end
```

sbc100: The rest of the complixity in your example I think comes from the conversion from double to int because your are returning and double from your main function.

If I do this I see the expected results:

```
$ cat test.c
extern double x;

int main(int argc, char **argv) {
  x = __builtin_rint(x);
}
$ ./emcc -Os test.c -c
$ dump test.o
000043 func[0] <__main_argc_argv>:
 000044: 41 00                      | i32.const 0
 000046: 41 00                      | i32.const 0
 000048: 2b 03 80 80 80 80 00       | f64.load 3 0
           00004a: R_WASM_MEMORY_ADDR_LEB 1 <x>
 00004f: 9e                         | f64.nearest
 000050: 39 03 80 80 80 80 00       | f64.store 3 0
           000052: R_WASM_MEMORY_ADDR_LEB 1 <x>
 000057: 41 00                      | i32.const 0
 000059: 0b                         | end
```

kripken: Oh, you're right, I missed the other conversion, sorry...

kripken (approved):

PR #23150, opened 2024-12-13
More use of async/await keywords in JS library code. NFC

sbc100:
Followup to #23104


src/source_map_support.js
```diff
@@ -106,11 +106,14 @@ function getSourceMap() {
...
 }
 
-function getSourceMapPromise() {
+async function getSourceMapAsync() {
   if (ENVIRONMENT_IS_WEB || ENVIRONMENT_IS_WORKER) {
-    return fetch(wasmSourceMapFile, {{{ makeModuleReceiveExpr('fetchSettings', "{ credentials: 'same-origin' }") }}})
-      .then((response) => response.json())
-      .catch(getSourceMap);
+    try {
+      var response = await fetch(wasmSourceMapFile, {{{ makeModuleReceiveExpr('fetchSettings', "{ credentials: 'same-origin' }") }}})
```
kripken, line 112:
```suggestion
      var response = await fetch(wasmSourceMapFile, {{{ makeModuleReceiveExpr('fetchSettings', "{ credentials: 'same-origin' }") }}});
```

kripken (approved):
Please run the manual audio tests, as we lack automatic coverage IIRC.

sbc100:
Of the two audio functions touched here it looks like we have no testing at all for `emscripten_resume_audio_context_async` and I can't find any tests that actually play audio that use `emscripten_start_wasm_audio_worklet_thread_async`.    Maybe I'll just revert that file for now since audio worklet testing is a little bit in flux/missing

Issue #23092, opened 2024-12-06
keeping node alive while workers are working


bakkot:
**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.73 (ac676d5e437525d15df5fd46bc2c208ec6d376a3)
clang version 20.0.0git (https:/github.com/llvm/llvm-project 1d810ece2b2c8fab77720493864257f0ea3336a9)
```

**Failing command line in full:**
```
emcc  --pre-js pre.js -pthread -s MODULARIZE=1 -s EXPORTED_FUNCTIONS="_spawn" main.cc && node run.mjs
```

----

I have a library which is intended to be driven from JS (i.e., no main loop in C, and using `-s MODULARIZE=1`).

I'm trying to spawn a thread, do some work, and then get the result back in the main thread. I have the following three bits of code, with the first two being the inputs to emcc and the third being the driver:

```js
// pre.js
let spawnTimeout;
let resolve;

Module.async_sum = () => {
  _spawn();
  return new Promise((res) => {
    resolve = res;
  });
};
```

```cpp
// main.cc
#include <thread>
#include <stdio.h>
#include <emscripten.h>

extern "C" void spawn() {
  printf("spawning\n");
  std::thread t([] {
    MAIN_THREAD_ASYNC_EM_ASM({
      clearTimeout(spawnTimeout);
    });

    printf("spawned\n");

    double v = 1.0;
    for (int i = 1; i < 1e8; ++i) {
      v += 1.0 / i;
    }
    printf("sum: %f\n", v);
    MAIN_THREAD_ASYNC_EM_ASM({
      resolve($0);
    }, v);
  });
  t.detach();
  EM_ASM({
    spawnTimeout = setTimeout(() => {}, 5000);
  });
}
```

```js
// run.mjs
import { default as init } from './a.out.js';

let x = await init();

console.log(await x.async_sum());
```

This prints
```
spawning
spawned
Warning: Detected unsettled top-level await at file:///Users/kevin/code/emscripten-thread-test/run.mjs:5
console.log(await x.async_sum());
```
and then exits, without printing the sum or giving me the result.

I was really hoping it would instead do the work and give me the result. That worked prior to #19073 (or to be more precise, it worked in version 3.1.15; I'm assuming that PR is the relevant change). I assume this is because node exits when the main thread has no tasks queued and the workers are all `unref`'d.

[...]

PR #23152, opened 2024-12-13
Add `emscripten_thread_set_strongref()` to the public API 

kleisauke:
Helps: #23092.


sbc100 (commented):
I feel like we might want to consider this be more deeply before exposing / creating a new public API.

A few reason that we might not want to make this public:

- This is very node specific and doesn't make sense on any other platform.
- I think you can get the same effect this using a simple `setInterval(() => !done)` loop.

If we do want expose this, I was wondering if perhaps we should make part of pthread_attr like we did for emscripten_pthread_attr_settransferredcanvases.

kleisauke:
I agree; I'll abandon this PR. 

Here's a possible alternative patch to address that issue:
```diff
--- a/src/library_pthread.js
+++ b/src/library_pthread.js
@@ -188,11 +188,11 @@ var LibraryPThread = {
       // worker pool as an unused worker.
       worker.pthread_ptr = 0;
 
-#if ENVIRONMENT_MAY_BE_NODE && PROXY_TO_PTHREAD
+#if ENVIRONMENT_MAY_BE_NODE
       if (ENVIRONMENT_IS_NODE) {
-        // Once the proxied main thread has finished, mark it as weakly
-        // referenced so that its existence does not prevent Node.js from
-        // exiting.  This has no effect if the worker is already weakly
+        // Once a pthread has finished and the worker becomes idle, mark it
+        // as weakly referenced so that its existence does not prevent Node.js
+        // from exiting.  This has no effect if the worker is already weakly
         // referenced.
         worker.unref();
       }
@@ -633,7 +633,7 @@ var LibraryPThread = {
     msg.moduleCanvasId = threadParams.moduleCanvasId;
     msg.offscreenCanvases = threadParams.offscreenCanvases;
 #endif
-#if ENVIRONMENT_MAY_BE_NODE
+#if ENVIRONMENT_MAY_BE_NODE && HAS_MAIN
     if (ENVIRONMENT_IS_NODE) {
       // Mark worker as weakly referenced once we start executing a pthread,
       // so that its existence does not prevent Node.js from exiting.  This
```

i.e. call `.unref()` only if a `main()` function is present. To retain the current behavior in libraries with this patch, users can include the following code:
```c

int main() {
    emscripten_exit_with_live_runtime();

    return 0;
}
```

However,  I'm uncertain if this is the optimal solution. An alternative could be to guard it behind `EXIT_RUNTIME`, but this might lead to some test failures, e.g. `test_pthread_weak_ref` isn't built with `EXIT_RUNTIME`.

sbc100:
I'm not 100% against this PR.   I actually think something like this might be where we end up.   I just want to investigate more options before adding a new API.

PR #23153, opened 2024-12-13
Webaudio: Consistent use of node handle vs audio handle

slowriot:
There's currently some confusion in webaudio.h about node handle types vs the audio handle type - `EMSCRIPTEN_WEBAUDIO_T` is sometimes used to refer to nodes, where it should be `EMSCRIPTEN_AUDIO_WORKLET_NODE_T`. 

[`emscripten_create_wasm_audio_worklet_node` returns `EMSCRIPTEN_AUDIO_WORKLET_NODE_T`](https://github.com/emscripten-core/emscripten/compare/main...slowriot:emscripten:patch-1#diff-dc30c9339d0b63b1311bb29d4f4a99fee9214c17ff5e8b28b6a9e61a55e79bceR137) but functions `emscripten_destroy_web_audio_node` and `emscripten_audio_node_connect` expect the node handle to be passed as a `EMSCRIPTEN_WEBAUDIO_T`.  This PR straightens out those inconsistencies.

It presently doesn't make any material difference since both handles are `typedef int`, so this PR just makes things clearer without changing any behaviour.  If the typedefs ever change in future, it will become important.

## Changes
- Move the `EMSCRIPTEN_AUDIO_WORKLET_NODE_T` node handle typedef to the top of the file alongside the `EMSCRIPTEN_WEBAUDIO_T` definition.
- `emscripten_destroy_web_audio_node` accepts a node handle.
- `emscripten_audio_node_connect` accepts a node handle as its first argument.


system/include/emscripten/webaudio.h
```diff
@@ -57,7 +58,7 @@ void emscripten_destroy_audio_context(EMSCRIPTEN_WEBAUDIO_T audioContext);
 // Disconnects the given audio node from its audio graph, and then releases
 // the JS object table reference to the given audio node. The specified handle
 // is invalid after calling this function.
-void emscripten_destroy_web_audio_node(EMSCRIPTEN_WEBAUDIO_T objectHandle);
+void emscripten_destroy_web_audio_node(EMSCRIPTEN_AUDIO_WORKLET_NODE_T objectHandle);
```
sbc100, line 61: Can this API not be used without audio worklets?

cwoffenden: These types end up being interchangeable sometimes, see here for example:

https://github.com/emscripten-core/emscripten/blob/d19c16213cdc6f63400bf1a55b3c76704ba663d5/src/library_webaudio.js#L324

cwoffenden: > Can this API not be used without audio worklets?

Parts of it, but the it needs JS to create other node types.

(In general though, some of the functions need a node, others a worklet context, and often they’re interchangeable when creating a graph depending on the use.)

slowriot: If the types should be interchangeable, then perhaps `EMSCRIPTEN_AUDIO_WORKLET_NODE_T` can just be removed entirely, and `EMSCRIPTEN_WEBAUDIO_T` used everywhere?  I can update to do that if preferred.

sbc100:
Let see what @juj thinks

system/include/emscripten/webaudio.h
cwoffenden: Some parts of the API only work with a single type, others have tests like this and work with node or context:

https://github.com/emscripten-core/emscripten/blob/d19c16213cdc6f63400bf1a55b3c76704ba663d5/src/library_webaudio.js#L327

I don't think there's a correct choice.

juj:
Good point here.

It looks like the typedef EMSCRIPTEN_AUDIO_WORKLET_NODE_T really never achieved any real purpose. It is unlikely that we'd want to develop a strong type hierarchy for these low level C handle types (that would be a lot of engineering for these couple of functions gain).

So I posted PR https://github.com/emscripten-core/emscripten/pull/26867 with the idea to just unify to use the single common EMSCRIPTEN_WEBAUDIO_T type uniformly.

PR #23154, opened 2024-12-13
Auto-format Python code with ruff format

cclauss:
Fixes: #14449
* #14449

Use `ruff format` to format Python code.  https://docs.astral.sh/ruff/formatter

Settings in `pyproject.toml` are all subject to discussion:
```toml
format.quote-style = "preserve"
indent-width = 2
line-length = 120
```
* Add a `ruff format` command after the existing `ruff check` test in CircleCI.
* Remove `.style.yapf` because maintainers should agree on a single Python code formatter.


test/common.py
```diff
@@ -407,8 +407,7 @@ def metafunc(self, fs, *args, **kwargs):
       assert fs is None
     func(self, *args, **kwargs)
 
-  parameterize(metafunc, {'': (None,),
-                          'nodefs': ('nodefs',)})
+  parameterize(metafunc, {'': (None,), 'nodefs': ('nodefs',)})
```
sbc100, line 410: All of these uses of `parameterize` are deliberately spread so that there is one test case per line.   Its important for readability.

cclauss: See the `format suppression` comment in the commit message.

cclauss: Adding a trailing comma after the last dict item will cause `ruff format` to make things clear.
```suggestion
  parameterize(
    metafunc,
    {
      '': (None,),
      'nodefs': ('nodefs',),
    },
  )

```

PR #23155, opened 2024-12-13
ruff check --preview --select=E303

cclauss:
As requested at:
* https://github.com/emscripten-core/emscripten/pull/23139#discussion_r1883095264

% [`ruff rule E303`](https://docs.astral.sh/ruff/rules/too-many-blank-lines)
# too-many-blank-lines (E303)

Derived from the **pycodestyle** linter.

Fix is always available.

This rule is in preview and is not stable. The `--preview` flag is required for use.

## What it does
Checks for extraneous blank lines.

## Why is this bad?
PEP 8 recommends using blank lines as follows:
- No more than two blank lines between top-level statements.
- No more than one blank line between non-top-level statements.

## Example
```python
def func1():
    pass



def func2():
    pass
```

Use instead:
```python
def func1():
    pass


def func2():
    pass
```

## Typing stub files (`.pyi`)
The rule allows at most one blank line in typing stub files in accordance to the typing style guide recommendation.

Note: The rule respects the following `isort` settings when determining the maximum number of blank lines allowed between two statements:

* [`lint.isort.lines-after-imports`]: For top-level statements directly following an import statement.
* [`lint.isort.lines-between-types`]: For `import` statements directly following a `from ... import ...` statement or vice versa.

## Options
- `lint.isort.lines-after-imports`
- `lint.isort.lines-between-types`

## References
- [PEP 8: Blank Lines](https://peps.python.org/pep-0008/#blank-lines)
- [Flake 8 rule](https://www.flake8rules.com/rules/E303.html)
- [Typing Style Guide](https://typing.readthedocs.io/en/latest/source/stubs.html#blank-lines)


.circleci/config.yml
```diff
@@ -425,6 +425,7 @@ jobs:
       - checkout
       - pip-install
       - run: ruff check
+      - run: ruff check --preview --select=E303
```
sbc100, line 428: Is there any way to have `ruff check` include E303 whenout adding these extra command line argument?

I'm just thinking for local development its nice to just run `ruff check` from time to time.

cclauss: Nope.  We need to target using both `--preview` and `--select=E303`.  Without `--preview`, `E303` will not be tested.  Without `--select=E303` then `ruff check --preview` with our current ruff config in `pyproject.toml` returns:
```
Found 17355 errors.
[*] 497 fixable with the `--fix` option (221 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

I chose this approach because running both commands one after the other will take a small fraction of one second.

sbc100: Ok that sounds like a reasonable solution for now.  Can you add a comment explaining why this is needed?   Is it simply that the ruff authors are still working on the E303 rule?  Will it become stable at some point and we can remove this?  Perhaps mention that in the comment?

cclauss: Will do… https://docs.astral.sh/ruff/preview

cclauss:
> TODO (cclauss): When ruff supports rule E303 without --preview, remove
> 1. lint.select E303
> 2. lint.explicit-preview-rules = true, and
> 3. lint.preview = true

sbc100 (approved):

cclauss:
I reverted to the original solution.  `tool.ruff.lint.explicit-preview-rules` would have been much easier to use and understand if it took a list of rules but it is just a `bool` so it invoked ruff rule F841 but only when tuple unpacking.  Too confusing.

PR #23157, opened 2024-12-13
Convert `createWasm` async/await where possible

sbc100:
The advantage if using `await` in the cases where we can is that it avoids the generation or wrapper function for each export.

So instead of

```
var wasmExport = createWasm(); // returns empty object
...
var malloc = (..) => (malloc = wasmExports['malloc'])(..);
```

We can just generate:

```
var wasmExport = await createWasm(); // returns actual exports
...
var malloc = wasmExports['malloc'];
```

This only currently works in MODULARIZE mode where the code is running inside a factory function.  Otherwise it would end up using top-level-await.

One wrinkle here is that this is not currently supported when closure is enabled because we run closure only on the contents of the factory function so closure ends up seeing this as a top level await when its not.

There are two minor observable effects of this change:

1. In `MODULARIZE` mode its no longer possible to call `new Foo()` with factory function.  We already added a warning for this debug builds in #23210.
2. Because we now can `await` for createWasm to return, the `run` method can run on first call, which means it runs `main` on first call, which means main will now run before `postjs` code, just like it would with sync instantiation.


test/test_unicode_js_library.out
```diff
@@ -1,2 +1,2 @@
-Unicode postjs: àČšñéáúÍųåêâăščếệüçλληνικάбългарскиPусскийСрпскиУкраїнська한국어中文普通话(中国大陆)普通话(香港)中文(台灣)粵語(香港)日本語हिन्दीภาษาไทย
 Unicode snowman ☃ says hello! àČšñéáúÍųåêâăščếệüçλληνικάбългарскиPусскийСрпскиУкраїнська한국어中文普通话(中国大陆)普通话(香港)中文(台灣)粵語(香港)日本語हिन्दीภาษาไทย
+Unicode postjs: àČšñéáúÍųåêâăščếệüçλληνικάбългарскиPусскийСрпскиУкраїнська한국어中文普通话(中国大陆)普通话(香港)中文(台灣)粵語(香港)日本語हिन्दीภาษาไทย
```
sbc100, line 2: This is the only affected test.   The reason is that when we `await` for the result of `createWasm` it behaves more like sync instantation, where the `run()` call simply runs main rather than waiting for unresolved dependencies.

This means that `postjs` code, which comes after `main` will now run after main like it would do with sync instantiation. 

I think this is a pretty reasonable change.   Basically `main` runs first if the wasm module is ready, and main run later if the wasm module is not ready.

test/test_browser.py
```diff
@@ -4795,28 +4795,19 @@ def test_browser_run_from_different_directory(self):
 
   # Similar to `test_browser_run_from_different_directory`, but asynchronous because of `-sMODULARIZE`
   def test_browser_run_from_different_directory_async(self):
-    for args, creations in [
-      (['-sMODULARIZE'], [
-        'Module();',    # documented way for using modularize
-        'new Module();' # not documented as working, but we support it
```
sbc100, line 4801: This is the other odd behavior which happened with work by accident before but not after.    Basically you can call `new Foo()` with a promise-returning function but not with an explicitly `async` function.

kripken: Makes sense. But please

1. Add this breaking change to the changelog, and
2. Add a clear runtime error in ASSERTIONS mode, with a test.

With those, I think the risk is mitigated.

tools/emscripten.py
```diff
@@ -893,6 +893,14 @@ def create_sending(metadata, library_symbols):
   return '{\n  ' + ',\n  '.join(elems) + '\n}'
 
 
+def can_use_await():
+  # In modularize mode we can use `await` since the factory function itself
+  # is marked as `async` and the generated code all lives inside that factory
+  # function.
+  # FIXME(https://github.com/emscripten-core/emscripten/issues/23158)
```
kripken, line 900: The comment seems positive - "in mode X we can do Y, which is good" - so that there is a FIXME seems surprising. Perhaps
```suggestion
  # FIXME: support this with closure (https://github.com/emscripten-core/emscripten/issues/23158)
```

test/test_browser.py
sbc100: I updated the ChangeLog, but I don't think there is any way to detect or allow `new Module()` if `Module` is `async` so I'm not sure there is anything can do in `ASSERTIONS` builds.

One possible option is to add another wrapper than is not async:

e.g. 

```
function Module() {
   async function InnerModule() {
       ...
   }
   return InnterModule();
}
```

Do you think its worth doing this?

kripken: I think in an ASSERTIONS build that would be useful to do, yeah. A little ugly but if it saves users time, I think it's worth it.

sbc100: This landed in #23210

tools/emscripten.py
```diff
@@ -893,6 +893,16 @@ def create_sending(metadata, library_symbols):
   return '{\n  ' + ',\n  '.join(elems) + '\n}'
 
 
+def can_use_await():
+  # In MODULARIZE mode we can use `await` since the factory function itself
+  # is marked as `async` and the generated code all lives inside that factory
+  # function.
+  # However, because closure does see this (it runs only on the inner code),
```
kripken, line 900:
```suggestion
  # However, because closure does not see this (it runs only on the inner code),
```

tools/emscripten.py
```diff
@@ -998,7 +1008,11 @@ def create_module(receiving, metadata, global_exports, library_symbols):
 
   if not settings.MINIMAL_RUNTIME:
     if settings.WASM_ASYNC_COMPILATION:
-      module.append("var wasmExports;\ncreateWasm();\n")
+      if can_use_await():
+        # In modullize mode the generated code is within a factory function.
```
kripken, line 1012:
```suggestion
        # In modularize mode the generated code is within a factory function.
```

ChangeLog.md
```diff
@@ -33,6 +33,8 @@ See docs/process.md for more on how version tagging works.
   `PATH.basename("a/b/..")` returns `".."` instead of `"a"`. This is in line with
   the behaviour of both node and coreutils, and is already the case when using
   NODERAWFS". (#23180)
+- The factory function exposed in `-sMODULARIZE` mode is now marked as `async`
+  when `WASM_ASYNC_COMPILATION` is enabled (the default). (#23157)
```
kripken, line 37:
```suggestion
  when `WASM_ASYNC_COMPILATION` is enabled (the default). This may also
  allow run() to execute earlier. (#23157)
```

kripken: My suggestion here isn't quite right, but I think it is worth mentioning the other issue (affecting `test_unicode_js..`) because having a mention in the changelog could help users that find weird startup behavior after this.

sbc100: Done

ChangeLog.md
```diff
@@ -33,6 +33,12 @@ See docs/process.md for more on how version tagging works.
   `PATH.basename("a/b/..")` returns `".."` instead of `"a"`. This is in line with
   the behaviour of both node and coreutils, and is already the case when using
   NODERAWFS". (#23180)
+- The factory function exposed in `-sMODULARIZE` mode is now marked as `async`
+  when `WASM_ASYNC_COMPILATION` is enabled (the default). This allows us to use
+  `await` during module creation.  One side effect of this is that code in
+  `--post-js` file will now be delayed until after module creation and after
```
kripken, line 39:
```suggestion
  `--post-js` files will now be delayed until after module creation and after
```

PR #23159, opened 2024-12-13
JS optimizer: Fix scoping issues with non-defined functions

### LINKED ISSUE #22968: __embind_initialize_bindings is stripped from export table in Release + pthreads build
This is a regression in 3.1.70, on previously tested version 3.1.56 it worked correctly. My application gets undefined reference to __embind_initialize_bindings during runtime.

There is a code which adds `__embind_initialize_bindings` to exported symbols:

https://github.com/emscripten-core/emscripten/blob/119a427e18eca02192fea679256c021834835cb4/tools/link.py#L508-L509

but this seems to be stripped during call to binaryen:

https://github.com/emscripten-core/emscripten/blob/119a427e18eca02192fea679256c021834835cb4/tools/building.py#L842-L847

The binaryen takes an DCEGraph as an input, and when I inspect the json of graph it has following entry:
```
{
    "name": "emcc$export$__embind_initialize_bindings",
    "export": "_embind_initialize_bindings",
    "reaches": []
 },
```
 I can fix this issue by manually editing this intermediate file and changing this entry to:
```
   {
    "name": "emcc$export$__embind_initialize_bindings",
    "export": "_embind_initialize_bindings",
    "reaches": [],
    "root": true
  }
```

My guess: `_embind_initialize_bindings` should be marked with `root: true` but for some reason is not.

Another workaround is manually exporting this function via:
`EXPORTED_FUNCTIONS=_main,__embind_initialize_bindings"`
but I believe it should not be needed.

This issue is similiar to: https://github.com/emscripten-core/emscripten/issues/21844

kripken:
Other than simple defined functions `function foo() {}`, JS also has
arrow functions and object methods. We need to be aware of those
in metadce.

This fixes #22968 which was an issue with an object method that
enclosed a function of the same name as another toplevel function.
To fix this, this PR makes us aware of such scopes, and we do not
optimize functions in them. That is, we only ever optimize defined
functions at the toplevel scope, and anything enclosed is considered
fixed.


kripken:
@sbc100 I'm not sure what to do with the prettier error here. On CI the error is without explanation, so I tried to run it locally, `./node_modules/prettier/bin/prettier.cjs tools/acorn-optimizer.mjs --fix`. But that produces 334 lines of diff, on lots of other stuff than I touched?

test/js_optimizer/emitDCEGraph6.js
```diff
@@ -0,0 +1,35 @@
+
```
sbc100, line 1: I wonder if we can give this (and the other tests 1-5) better names?  Can be followup maybe?

sbc100:
> @sbc100 I'm not sure what to do with the prettier error here. On CI the error is without explanation, so I tried to run it locally, `./node_modules/prettier/bin/prettier.cjs tools/acorn-optimizer.mjs --fix`. But that produces 334 lines of diff, on lots of other stuff than I touched?

If you run that on `main` does it do anything?  For me it doesn't.

tools/acorn-optimizer.mjs
```diff
@@ -857,9 +871,19 @@ function emitDCEGraph(ast) {
           }
         }
       }
+    } else if (node.type === 'Property' && node.method) {
```
sbc100, line 874: `assert(specialScopes > 0)` ?

kripken:
Correct, it does nothing on the main branch, but does a 334-like diff on this one, though my changes are an order of magnitude smaller than that. I'm not familiar with prettier - is this a known issue?

tools/acorn-optimizer.mjs
sbc100: (Sorry just saw the one below.. probably don't need both)

test/js_optimizer/emitDCEGraph6.js
kripken: I'll rename the new one in this PR. I can look at the others later.

kripken:
Ok, I pushed the prettier results (and the other feedback). The diff is now larger, but most of that is whitespace. I don't quite understand why it decided to change that whitespace, but it doesn't look bad...

sbc100:
> Ok, I pushed the prettier results (and the other feedback). The diff is now larger, but most of that is whitespace. I don't quite understand why it decided to change that whitespace, but it doesn't look bad...

Its because it changed the indentation of argument to fullWalk on line 711.   Viewing with ignore whitespace works.

tools/acorn-optimizer.mjs
kripken: Why, I think both make equal sense?

sbc100: lgtm either way.

brendandahl:
Do you know offhand if methods in js classes work with metadce?

kripken:
Does that use the same syntax as object methods? (`type: "Property"`, and `method: true`) If so we should be handling them properly after this PR. But we just ignore them, we don't try to optimize away parts of classes or objects.

kripken:
Landing to fix the user issue - if there is more to do here, we can look into that later.

PR #23169, opened 2024-12-15
Remove `createRequire` in favor of ES6 `import`

sbc100:


kripken (approved):

kleisauke:
I think the `require()` compat function is still needed in some cases.
```
file:///home/kleisauke/wasm-vips/lib/vips-node.mjs:186
function fd(a,b,c){if(!(0<=a&&3>=a))return 28;if(0===a)a=Date.now();else if(Hi)a=performance.timeOrigin+performance.now();else return 52;D[c>>3]=BigInt(Math.round(1E6*a));return 0}fd.h="iijp";var gd=()=>{};gd.h="v";var jd=a=>x(P(a));jd.h="vp";var kd=()=>{We+=1;throw"unwind";};kd.h="v";var ld=()=>z.length;ld.h="p";var nd=()=>t?require("os").cpus().length:navigator.hardwareConcurrency;nd.h="i";var od=a=>{pi(a)};od.h="vp";
                                                                                                                                                                                                                                                                                                                                      ^

ReferenceError: require is not defined
```
https://github.com/emscripten-core/emscripten/blob/3d4df5b73be7ca25ef78b45bce98c7c2eed34fcd/src/library_atomic.js#L155-L157

grep-ing through the codebase, I also found:
https://github.com/emscripten-core/emscripten/blob/3d4df5b73be7ca25ef78b45bce98c7c2eed34fcd/src/library_uuid.js#L36
https://github.com/emscripten-core/emscripten/blob/3d4df5b73be7ca25ef78b45bce98c7c2eed34fcd/src/library_wasi.js#L582
https://github.com/emscripten-core/emscripten/blob/3d4df5b73be7ca25ef78b45bce98c7c2eed34fcd/src/runtime_shared.js#L63
https://github.com/emscripten-core/emscripten/blob/3d4df5b73be7ca25ef78b45bce98c7c2eed34fcd/src/wasm_worker.js#L13
https://github.com/emscripten-core/emscripten/blob/3d4df5b73be7ca25ef78b45bce98c7c2eed34fcd/src/wasm_worker.js#L31-L32
etc.

sbc100:
Oh wow!  I guess we don't have tests that cover any of those?  Or at least we don't have tests that cover those in combination with `EXPORT_ES6`.

Lets try to update all those sites rather than being back this pollyfil.   

BTW, do you know which of the over usages is the one you are running into?

kleisauke:
Indeed, I think some test coverage is missing for this. A simple test would be to check whether `require(` appears in the output JS file, if we proceed with this. It might also cause issues for users who depend on `require()` in custom `--pre-js` or `--js-library` files, but those cases are likely negligible.

> BTW, do you know which of the over usages is the one you are running into?

Only the [`require('os').cpus().length`](https://github.com/emscripten-core/emscripten/blob/09a662afee2cd02fc617d95616e0b0900c5bbdfc/src/library_atomic.js#L156C27-L156C54) and [`require('crypto').randomFillSync(view)`](https://github.com/emscripten-core/emscripten/blob/09a662afee2cd02fc617d95616e0b0900c5bbdfc/src/library_wasi.js#L576C24-L576C62) snippets were causing issues for me.

sbc100:
> Indeed, I think some test coverage is missing for this. A simple test would be to check whether `require(` appears in the output JS file, if we proceed with this. It might also cause issues for users who depend on `require()` in custom `--pre-js` or `--js-library` files, but those cases are likely negligible.

Presumably any such users would either:

1. Not expect their code to work in an ES6 module context
2. Use createRequire themselves to add require to the ES6 module context.

PR #9234, opened 2019-08-14
Add a specific flag to enable / disable the use of 'import.meta.url'

adevress:
This fixes the problem reported by #9233 and by #8729 while fixing the regression bug causes by #8940

Thank you for submitting a pull request!

If this is your first PR, make sure to add yourself to AUTHORS.

Issue #8729, opened 2019-06-02
EXPORT_ES6 mode import.meta.url is not supported


k2snowman69:
Issue:
[A recent PR](https://github.com/emscripten-core/emscripten/pull/8306/files#diff-ae40e645f84e893ce5d037b5efc53be6R2850) resulted in the use of import.meta.url for EXPORT_ES6. However this functionality is currently only at stage 3 (proposal) which means support for this functionality is fairly low which includes eslint, babel, Firefox and Edge all not supporting it currently (all implement at stage 4 or later)

Context:
Looks like there was already a discussion about this and how support isn't quite there yet.
https://github.com/emscripten-core/emscripten/issues/7137

VirtualTim:
Whoops, that was my fault. Originally I was using `performance.getEntries().slice(-1)[0].name`, which is probably more supported (if much uglier). https://github.com/emscripten-core/emscripten/pull/8306#discussion_r267321263

An idea might be to use `performance.getEntries().slice(-1)[0].name`, and leave a comment to swap this with `import.meta.url` in the future. But I'm far from a JavaScript expert, so someone else might have a better idea.

k2snowman69:
Another option is to check for the existence of `import` and `import.meta` else back onto the old code base.

davidyxu:
Just wanted to chime in on this conversation, the project I'm working on recently ran into this issue where import.meta.url isn't available (we tried to polyfill it through .

I think it might make sense for this into an opt-in flag where people who are targeting an environment where `import.meta.url` is reliably available can use it.

https://github.com/emscripten-core/emscripten/pull/8306/files#diff-ae40e645f84e893ce5d037b5efc53be6R2849
i.e. replacing this part of the code with:
```
if shared.Settings.EXPORT_ES6 and shared.SETTINGS_EXPORT_USE_IMPORT_META :
        script_url = "import.meta.url"
```

k2snowman69:
Wouldn't checking the existence of it also succeed if it's been polyfilled? Meaning less code and settings for the same goal? Or do you mean that there maybe some people who wouldn't want the setting even if it's available?

davidyxu:
If I'm understanding your suggested approach correctly, I think that would end up being more code being delivered to clients since it would need to be a runtime check.

In general I think putting the flexibility into the hands of the emsdk user is better here because there's a lot of variables in the module's running context that will be unknown to emscripten.

[...]

PR #23171, opened 2024-12-15
Remove USE_ES6_IMPORT_META setting

sbc100:
This setting was added (IIUC) as a temporary workaround for the lack or support for `import.meta` in browser and tooling.

However `import.meta` has been at stage 4 for almost 5 years now: https://github.com/tc39/proposal-import-meta/pull/21

Webback (released in 2020) also has builtin support for `import.meta`, so I would hope the use case for disabling this setting no longer exists.

See #9234 and #8729


brendandahl (approved):

src/parseTools.mjs
```diff
@@ -44,7 +44,7 @@ export function processMacros(text, filename) {
 // Also handles #include x.js (similar to C #include <file>)
 export function preprocess(filename) {
   let text = read(filename);
-  if (EXPORT_ES6 && USE_ES6_IMPORT_META) {
+  if (EXPORT_ES6) {
     // `eval`, Terser and Closure don't support module syntax; to allow it,
```
brendandahl, line 48: Potential follow up: I'm pretty sure closure supports module syntax now. Not sure about terser.

sbc100: I looked into this and will try to come up with followup.   For now is seems like the `eval` phase is the one were we still need this.

VirtualTim (approved):

kleisauke:
FWIW, Cloudflare Workers doesn't support `import.meta` at the moment, see: https://github.com/cloudflare/workerd/issues/2963.

That said, I don't think that should block this PR, since users can simply omit `-sEXPORT_ES6` if they need to deploy there.

sbc100:
> FWIW, Cloudflare Workers doesn't support `import.meta` at the moment, see: [cloudflare/workerd#2963](https://github.com/cloudflare/workerd/issues/2963).
> 
> That said, I don't think that should block this PR, since users can simply omit `-sEXPORT_ES6` if they need to deploy there.

Or presumably one could just use a bundler before deploying to cloudflare?   Which seems like a likely/good choice anyway, right?

JoeOsborn:
This fix breaks cross-origin workers via mainScriptUrlOrBlob.  Since the worker script is now hard-coded as the URL, we don't have the ability to use an inline worker to importScripts or use a downloaded blob.

dlemstra:
I was using this option for a use case that was probably not the intention of this option. I used this option to avoid a bundler from embedding the wasm file. And without this option I now need to find a way in the tooling that uses the output to avoid inlining ` new URL('{{{ WASM_BINARY_FILE }}}', import.meta.url).href` as ` new URL("data:application/wasm;base64,`. I am using `Module['locateFile']` and `Module['wasmBinary']` to give control to the user of my library where to load the `.wasm` from. So for me it would still be nice to have an option to disable this or only allow the usage of `locateFile` and `wasmBinary`. But I can also understand that this would not be something that the team wants to maintain.

For those who are reading this in the future I use the following "hack" to work around this issue. Replace `new URL('myfile.wasm', import.meta.url).href` with `new URL('myfile.wasm.empty', import.meta.url).href` after the output is generated and add an empty file `myfile.wasm.empty` (replace `myfile` with the name of your wasm file). This will of course prevent that part from working so you will need to either set `locateFile` or `wasmBinary` to fix the loading of the wasm data.

sbc100:
Interesting.  It does sounds odd that this would be the option used to prevent the bundler from embedding the wasm.   Can I ask what bundler you are using?   Have you looked for a blunder setting that might avoid this?

dlemstra:
I have not found a bug about that so I will probably need to report one. But that will probably take me some time to create something that can be used to reproduce it. And the work around can be a possible "fix/hack" when someone else also has a problem with their bundler.

Sec-ant:
I have exactly the same use case as what @dlemstra mentioned above for `USE_ES6_IMPORT_META` in one of my project, I also let users to [use `locateFile` to override the wasm serve path](https://github.com/Sec-ant/zxing-wasm?tab=readme-ov-file#configuring-wasm-serving). I ended up [patching the `preamble.js` file](https://github.com/Sec-ant/zxing-wasm/pull/225/commits/d8868d5011f188d26d87701fe89e80769bcd270e) to restore the old behavior, which is far from graceful but yields the expected output. I also use vite 6 to bundle my project.

jinwuwu001:
Removing USE_ES6_IMPORT_META will result in library users being unable to customize the loading location of .js when using multi-threaded .wasm.   
The multi-threaded .js code obtained when USE_ES6_IMPORT_META=0 is as follows:

```js
// emsdk 4.0.1
allocateUnusedWorker() {
  var worker;
  var workerOptions = { type: "module", name: "em-pthread" };
  var pthreadMainJs = _scriptName;
  if (Module["mainScriptUrlOrBlob"]) {
    pthreadMainJs = Module["mainScriptUrlOrBlob"];
    if (typeof pthreadMainJs != "string") {
      pthreadMainJs = URL.createObjectURL(pthreadMainJs);
    }
  }
  worker = new Worker(pthreadMainJs, workerOptions);
  PThread.unusedWorkers.push(worker);
}
```

Users can set the loading location through mainScriptUrlOrBlob

```ts
const moduleConfig = {
  locateFile: (file: string) => {
    const path = `${demoDir}.wasm`;
    return path;
  },
  mainScriptUrlOrBlob: `${demoDir}.js`
};
```

The multi-threaded .js code obtained when USE_ES6_IMPORT_META=1 is as follows:
```js
// emsdk 4.0.4
allocateUnusedWorker() {
  var worker;
  worker = new Worker(new URL("demo.js", import.meta.url), { type: "module", name: "em-pthread" });
  PThread.unusedWorkers.push(worker);
}
```

At this time, users can only load files from the following path
```js
`{import.meta.url}/demo.js`
```


Cannot load files from the specified location unless manually modifying demo.js
for example:

```js
`{import.meta.url}/dir1/demo.js`
`{import.meta.url}/dir1/dir2/demo.js`
`{import.meta.url}/dir1/dir2/dir3/demo.js`
```

How to customize the loading location after removing USE_ES6_IMPORT_META
It is necessary to customize the loading position in the scenario of dynamically loading .wasm

kleisauke:
@jinwuwu001 Could you re-test with Emscripten 4.0.6 or later? I think your use-case was fixed via commit 9e21488c4f500950cdcf37bd7b8405ad64fd1934.

PR #23172, opened 2024-12-16
Avoid some unneccesary path resolution

hoodmane:
There are a bunch of codepaths where we have a node or a stream and we turn it back into a path, then eventually call `lookupPath` on it to turn it back into a node. This removes a few of these instances.

In general, I think it would be good to avoid `stream.path` in favor of `stream.node`, since we eventually will call `lookupPath()` to get back `stream.node`.


src/library_fs.js
```diff
@@ -895,6 +896,9 @@ FS.staticInit();
       FS.readdirNode(FS.lookupPath(path, { follow: true }).node);
     },
     readdirNode(node) {
+      if (!node.parent) {
+        return [];
+      }
```
sbc100, line 901: Any idea what these lines were supposed to be doing?

What does not having a parent mean?  Is that only true for the root directory?  Why would we want that to mean return the empty list?

hoodmane: Well, if we've unlinked a directory we want to return empty list or raise ENOENT. By contrast, if the directory exists but is empty we return `[".", ".."]`. So we need to label the directory node as unlinked somehow when we unlink them. I can add a comment here saying that.

PR #23173, opened 2024-12-16
[ci] Update linux image name from bionic to focal

cclauss:
Reduce confusion by using the correct Ubuntu release name.  Currently, the tag is `bionic` but the image is `focal`.
```diff
-  bionic:
+  focal:
     docker:
       - image: emscripten/emscripten-ci:focal
```
https://releases.ubuntu.com

Should upgrading to Jammy 22.04 or Noble 24.04 be considered?

---

~Also, resolve four problems raised by the [VS Code extension for CircleCI](https://circleci.com/docs/vs-code-extension-overview):~
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
   linux-python:
     docker:
       - image: cimg/python:3.10.7
-  bionic:
+  focal:
     docker:
       - image: emscripten/emscripten-ci:focal
     environment:
@@ -247,7 +247,7 @@ commands:
           # detect flakyness.
           condition: ${CIRCLE_PULL_REQUEST}
           steps:
-            set-retry-flaky-tests
+            - set-retry-flaky-tests
       - run:
           name: run tests (<< parameters.title >>)
           command: |
@@ -388,7 +388,7 @@ commands:
             #  "Firefox is already running, but is not responding."
             # TODO: find out a way to shut down and restart firefox
       - upload-test-results
-  test-sockets-chrome:
+  test-sockets-chrome-command:
     description: "Runs emscripten sockets tests under chrome"
     steps:
       - prepare-for-tests
@@ -411,7 +411,7 @@ commands:
 
 jobs:
   build-docs:
-    executor: bionic
+    executor: focal
     steps:
       - checkout
       - pip-install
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -623,12 +623,12 @@ jobs:
...
     environment:
       EMTEST_SKIP_NODE_CANARY: "1"
-    executor: bionic
+    executor: focal
     steps:
       - run-tests-linux:
           test_targets: "wasm2js1"
   test-wasm64:
-    # We don't use `bionic` here since its tool old to run recent node versions:
+    # We don't use `focal` here since its tool old to run recent node versions:
```
cclauss, line 631: Should we change this comment and also line 665?

sbc100: Hmm.. I wonder if this is still true?   Maybe we can followup be removing this.

.circleci/config.yml
```diff
@@ -388,7 +388,7 @@ commands:
             #  "Firefox is already running, but is not responding."
             # TODO: find out a way to shut down and restart firefox
       - upload-test-results
-  test-sockets-chrome:
+  test-sockets-chrome-command:
```
sbc100, line 391: Perhaps we should just inline this command instead of this change?  Maybe revert this part and we can do the inlinining as a separate chagne.

.circleci/config.yml
```diff
@@ -247,7 +247,7 @@ commands:
           # detect flakyness.
           condition: ${CIRCLE_PULL_REQUEST}
           steps:
-            set-retry-flaky-tests
+            - set-retry-flaky-tests
```
sbc100, line 250: Does the old version not work?

.circleci/config.yml
sbc100: Maybe revert these comment updates and we can look into removing these commends completely.   I wouldn't want to update the comments to something that is not true.

.circleci/config.yml
cclauss: CircleCI flags this as an incorrect data type error which means it could easily stop working in future releases.

sbc100: Interesting, but they don't show it as a config error.  I wonder why not?   LGTM to this part either way.

cclauss: https://circleci.com/docs/vs-code-extension-overview/#config-helper is in-context help..

sbc100 (approved):

Issue #23047, opened 2024-12-02
Support source phase imports JS proposal


sbc100:
The source phase imports proposal should allow us to simplify the emscripten-generated JS code significantly: https://github.com/tc39/proposal-source-phase-imports.

This proposal is at phase 3 so browser should be supporting without a flag soon.   For folks that use bundlers that bundler should also be able to lower this away. 

It seems like we should want a new setting for enable this, which should allow us to elide a lot of the wasm file loading code.

Chromium tracking issue: https://issues.chromium.org/issues/42204365

sbc100:
CC @guybedford

PR #23175, opened 2024-12-16
Add experimental option to use source wasm phase imports

sbc100:
Now that node support has been landed we can test this, at least
against the latest node canary builds.
    
See https://github.com/nodejs/node/pull/56919
    
Fixes: #23047

### DIFF
--- a/src/preamble.js
+++ b/src/preamble.js
@@ -558,11 +558,32 @@ function instrumentWasmTableWithAbort() {
 }
 #endif
 
+#if LOAD_SOURCE_MAP
+function receiveSourceMapJSON(sourceMap) {
+  wasmSourceMap = new WasmSourceMap(sourceMap);
+  {{{ runIfMainThread("removeRunDependency('source-map');") }}}
+}
+#endif
+
+#if (PTHREADS || WASM_WORKERS) && (LOAD_SOURCE_MAP || USE_OFFSET_CONVERTER)
+// When using postMessage to send an object, it is processed by the structured
+// clone algorithm.  The prototype, and hence methods, on that object is then
+// lost. This function adds back the lost prototype.  This does not work with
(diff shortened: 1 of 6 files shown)


sbc100:
@guybedford

src/settings.js
```diff
@@ -2182,6 +2182,13 @@ var LEGACY_RUNTIME = false;
 // [link]
 var SIGNATURE_CONVERSIONS = [];
 
+// Experimental support for wasm source phase imports.
+// This is only currenty implement in the pre-release/nightly version of node,
```
kripken, line 2186:
```suggestion
// This is only currently implemented in the pre-release/nightly version of node,
```

src/preamble.js
```diff
@@ -558,11 +558,40 @@ function instrumentWasmTableWithAbort() {
 }
 #endif
 
+#if LOAD_SOURCE_MAP
+var wasmSourceMap;
+#include "source_map_support.js"
```
kripken, line 563: Looks like this is still included from `runtime_shared.js` - can we not end up including it twice?

src/preamble.js
```diff
@@ -558,11 +558,40 @@ function instrumentWasmTableWithAbort() {
...
+// WasmOffsetConverter.
+function resetPrototype(constructor, attrs) {
+  var object = Object.create(constructor.prototype);
+  return Object.assign(object, attrs);
+}
+#endif
+
+#if USE_OFFSET_CONVERTER
+var wasmOffsetConverter;
+#include "wasm_offset_converter.js"
```
kripken, line 585: Same question here.

sbc100: Sorry, this was left over from a previous version of this PR that got rebased.   Reverted.

kripken (approved):

PR #23177, opened 2024-12-16
[test] Add new decorators for running tests under different FS backends

sbc100:
- @with_all_fs
- @also_with_nodefs
- @also_with_nodefs_both

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -379,12 +379,75 @@ def also_with_wasmfs(f):
     if DEBUG:
       print('parameterize:wasmfs=%d' % wasmfs)
     if wasmfs:
-      self.set_setting('WASMFS')
-      self.emcc_args.append('-DWASMFS')
+      self.setup_wasmfs_test()
+    else:
+      self.emcc_args += ['-DMEMFS']
     f(self, *args, **kwargs)
 
   parameterize(metafunc, {'': (False,),
                           'wasmfs': (True,)})
+  return metafunc
+
+
+def also_with_nodefs(func):
+  @wraps(func)
+  def metafunc(self, fs, *args, **kwargs):
+    if DEBUG:
+      print('parameterize:fs=%s' % (fs))
+    if fs == 'nodefs':
+      self.setup_nodefs_test()
+    else:
+      self.emcc_args += ['-DMEMFS']
+      assert fs is None
+    func(self, *args, **kwargs)
+
+  parameterize(metafunc, {'': (None,),
+                          'nodefs': ('nodefs',)})
+  return metafunc
+
+
+def also_with_nodefs_both(func):
+  @wraps(func)
+  def metafunc(self, fs, *args, **kwargs):
+    if DEBUG:
+      print('parameterize:fs=%s' % (fs))
+    if fs == 'nodefs':
+      self.setup_nodefs_test()
+    elif fs == 'rawfs':
+      self.setup_noderawfs_test()
+    else:
+      self.emcc_args += ['-DMEMFS']
+      assert fs is None
+    func(self, *args, **kwargs)
+
+  parameterize(metafunc, {'': (None,),
+                          'nodefs': ('nodefs',),
+                          'rawfs': ('rawfs',)})
+  return metafunc
+
+
+def with_all_fs(func):
+  @wraps(func)
+  def metafunc(self, fs, *args, **kwargs):
+    if DEBUG:
+      print('parameterize:fs=%s' % (fs))
+    if fs == 'nodefs':
+      self.setup_nodefs_test()
+    elif fs == 'rawfs':
+      self.setup_noderawfs_test()
+    elif fs == 'wasmfs':
+      self.setup_wasmfs_test()
+    else:
+      self.emcc_args += ['-DMEMFS']
+      assert fs is None
+    func(self, *args, **kwargs)
+
+  # TODO(sbc): rather than treat WASMFS as orthogonal we should
+  # probably make it combinatorial with nodefs and noderawfs.
+  parameterize(metafunc, {'': (None,),
+                          'nodefs': ('nodefs',),
+                          'rawfs': ('rawfs',),
+                          'wasmfs': ('wasmfs',)})
   return metafunc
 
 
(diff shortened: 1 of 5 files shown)


test/common.py
```diff
@@ -388,6 +388,72 @@ def metafunc(self, wasmfs, *args, **kwargs):
...
 
+def also_with_nodefs(func):
+  @wraps(func)
+  def metafunc(self, fs, *args, **kwargs):
+    if DEBUG:
+      print('parameterize:fs=%s' % (fs))
+    if fs == 'nodefs':
+      self.setup_nodefs_test()
+    else:
+      self.emcc_args += ['-DMEMFS']
```
hoodmane, line 399: If this one is going to define `-DMEMFS` for consistency `also_with_noderawfs()` should probably do the same?

test/test_core.py
```diff
@@ -5873,34 +5855,19 @@ def test_fs_64bit(self):
...
-      self.skipTest('No symlinks on Windows')
-    if self.get_setting('WASMFS'):
-      if nodefs:
+  @with_all_fs
+  def test_fs_symlink_resolution(self):
+    nodefs = '-DNODEFS' in self.emcc_args or '-DNODERAWFS' in self.emcc_args
+    if nodefs:
+      if WINDOWS:
+        self.skipTest('No symlinks on Windows')
+      if self.get_setting('WASMFS'):
```
hoodmane, line 5864: So this happens if you do `test.runner wasmfs.test_fs_symlink_resolution_nodefs`?

hoodmane: Why does this test need this but the others not?

hoodmane: Maybe this skip could be moved into the decorators?

test/unistd/links.c
```diff
@@ -41,11 +41,11 @@ void changedir(const char *dir) {
 }
 
 void setup() {
-  makedir("working");
+  makedir("/working");
```
hoodmane, line 44: Can't we remove this now?

test/test_core.py
sbc100: > So this happens if you do `test.runner wasmfs.test_fs_symlink_resolution_nodefs`?

Yes, and actually I was looking that current CI setup and it looks like we do run a bunch of wasmfs tests:
https://github.com/emscripten-core/emscripten/blob/a9108f89b5d2bb1b0a219b7a2a167271a287cf32/.circleci/config.yml#L570-L622

test/unistd/links.c
sbc100: Sadly, no, I wasn't able to covert this test to use `test/setup_nodefs.js`.

Right now, unlike other tests, this test relies on absolute paths, and `test/setup_nodefs.js` works by moving the current working directory to `/nodefs/`.

hoodmane (approved):

test/common.py
```diff
@@ -388,6 +388,72 @@ def metafunc(self, wasmfs, *args, **kwargs):
...
+      self.emcc_args += ['-DWASMFS']
+    else:
+      self.emcc_args += ['-DMEMFS']
+      assert fs is None
+    func(self, *args, **kwargs)
+
+  parameterize(metafunc, {'': (None,),
+                          'nodefs': ('nodefs',),
+                          'rawfs': ('rawfs',),
+                          'wasmfs': ('wasmfs',)})
```
kripken, line 453: This won't test wasmfs with nodefs/rawfs, will it? Those should be orthogonal I think. That is, a test could do
```python
@also_with_wasmfs
@also_with_nodefs_both
```

kripken: Makes sense. How about adding a TODO to remove this and use the two separate decorators, to get full coverage?

kripken: (in particular, we should avoid adding more usages of this decorator)

test/common.py
sbc100: Done

test/common.py
sbc100: Added a comment.

test/common.py
```diff
@@ -379,15 +379,78 @@ def metafunc(self, wasmfs, *args, **kwargs):
...
+    elif fs == 'rawfs':
+      self.setup_noderawfs_test()
+    elif fs == 'wasmfs':
+      self.setup_wasmfs_test()
+    else:
+      self.emcc_args += ['-DMEMFS']
+      assert fs is None
+    func(self, *args, **kwargs)
+
+  # TODO(sbc): rather than treat WASMFS and orthogonal we should
```
kripken, line 445:
```suggestion
  # TODO(sbc): rather than treat WASMFS as orthogonal we should
```

test/unistd/links.c
```diff
@@ -41,11 +41,11 @@ void changedir(const char *dir) {
...
 #if defined(__EMSCRIPTEN__) && defined(NODEFS)
-  EM_ASM(FS.mount(NODEFS, { root: '.' }, 'working'));
+  EM_ASM(FS.mount(NODEFS, { root: '.' }, '/working'));
 #endif
-  changedir("working");
+  changedir("/working");
```
kripken, line 48: Why are any test changes needed at all to this file?

kripken (commented):
lgtm % question

Btw, separately, this adds more work in `test_core` that could all possibly be in `other` - not sure we need to run it in all opt modes.

sbc100:
> lgtm % question
> 
> Btw, separately, this adds more work in `test_core` that could all possibly be in `other` - not sure we need to run it in all opt modes.

I think in almost all of these cases I'm not adding more test variants here, just converting existing variants to use the decorators instead.  There may be one or two where I increased it.   I agree we might want to move all of these highly-parameterized FS tests to other.    Perhaps as a followup.

test/unistd/links.c
sbc100: The `@also_with_nodefs` mounts `/nodefs` and sets this as CWD on startup.

Without converting this test to use abs paths here that would break this test.

However I think using the absolute path (`/working`) is actually more correct as the check below assume that this is in the root directory.   See line 69

sbc100: Ideally we would have mount nodefs as `/` but I guess that maybe doesn't work for some reason?

kripken (approved):

test/test_core.py
```diff
@@ -5908,29 +5881,21 @@ def test_sigaction_default(self, signal, exit_code, assert_identical):
...
-  @parameterized({
-    '': (['-DMEMFS'],),
-    'nodefs': (['-DNODEFS', '-lnodefs.js'],),
-    'noderawfs': (['-DNODERAWFS', '-sNODERAWFS'],)
-  })
-  def test_unistd_access(self, args):
-    self.emcc_args += args
-    nodefs = '-DNODEFS' in args or '-DNODERAWFS' in args
+  @with_all_fs
+  def test_unistd_access(self):
```
hoodmane, line 5886: The wasmfs version of this test seems to be failing.

sbc100: Ah, this was one where I snuck in an extra test config.  Revert to just nodefs_both.

PR #23180, opened 2024-12-16
Don't normalize path in basename 

hoodmane:
This brings it in line with the behavior of basename im coreutils and node.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 3.1.75 (in development)
 -----------------------
+- `FS.basename()` no longer calls `PATH.normalize()`, so that
+  `FS.basename("a/.")` returns `"."` instead of `"a"` and `FS.basename("a/b/..")`
+  returns `".."` instead of `"a"`.
 
 3.1.74 - 12/14/24
 -----------------
--- a/src/library_path.js
+++ b/src/library_path.js
@@ -66,7 +66,6 @@ addToLibrary({
     basename: (path) => {
       // EMSCRIPTEN return '/'' for '/', not an empty string
       if (path === '/') return '/';
-      path = PATH.normalize(path);
       path = path.replace(/\/$/, "");
       var lastSlash = path.lastIndexOf('/');
       if (lastSlash === -1) return path;


ChangeLog.md
```diff
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 3.1.75 (in development)
 -----------------------
+- `FS.basename()` no longer calls `PATH.normalize()`, so that
+  `FS.basename("a/.")` returns `"."` instead of `"a"` and `FS.basename("a/b/..")`
+  returns `".."` instead of `"a"`. (#23180)
```
sbc100, line 25: How about adding "This is in line with the behaviour of both node and coreutils, and is already the case when using NODEFS/NODERAWFS".

src/library_path.js
```diff
@@ -66,8 +66,7 @@ addToLibrary({
     basename: (path) => {
       // EMSCRIPTEN return '/'' for '/', not an empty string
```
sbc100, line 67: Is there an extra quote here ?  Maybe remove it as part of this PR?

sbc100 (approved):

ChangeLog.md
sbc100: `s/FS.basename/PATH.basename/`

sbc100:
s/basename/PATH.basename/ in the PR title.

src/library_path.js
hoodmane: Right, I'd removed it in https://github.com/emscripten-core/emscripten/pull/23136.

ChangeLog.md
hoodmane: I think it's only already the case in NODERAWFS and is the cause of bugs in NODEFS.

sbc100: Did you see s/FS.basename/PATH.basename/ ?

hoodmane: Oops, updated that now.

src/library_path.js
```diff
@@ -64,10 +64,9 @@ addToLibrary({
       return root + dir;
     },
     basename: (path) => {
-      // EMSCRIPTEN return '/'' for '/', not an empty string
+      // EMSCRIPTEN return '/' for '/', not an empty string
```
sbc100, line 67: BTW this is another inconsistency with the NODERAWFS version.. makes me wonder if we can remove this too?   Unrelated to this change though.

sbc100 (approved):

src/library_path.js
hoodmane: Well, coreutils basename returns `/` for `/` and `//`  too. Our code returns '/' for input `/` but empty string for input //. If we wanted to be really careful about it we'd need something like:
```js
if (path === '') return '';
path = path.replace(/\/+$/g, "");
if (path === '') return '/';
```

sbc100: Lets leave that as a possible followup.

PR #23185, opened 2024-12-16
[test] Remove also_with_wasmfs_js. NFC

sbc100:
The only different here with `also_with_wasmfs` was that this decorator adds `FORCE_FILESYSTEM`, but the only test to use this decorator was already adding this flag.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -157,23 +157,6 @@ def only_wasm2js(note=''):
   return decorated
 
 
-# Similar to also_with_wasmfs, but also enables the full JS API
-def also_with_wasmfs_js(func):
-  assert callable(func)
-
-  @wraps(func)
-  def decorated(self):
-    func(self)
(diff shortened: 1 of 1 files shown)


test/common.py
```diff
@@ -380,6 +380,7 @@ def metafunc(self, wasmfs, *args, **kwargs):
       print('parameterize:wasmfs=%d' % wasmfs)
     if wasmfs:
       self.set_setting('WASMFS')
+      self.set_setting('FORCE_FILESYSTEM')
```
kripken, line 383: Why is this needed?

This PR only affects the one test `test_fs_writeFile`, which already has the decorator, as you mention in the description. Adding this flag here seems unnecessary IIANM, and has the risk of avoiding us from ever testing the non-forced case with wasmfs.

kripken (approved):

PR #23188, opened 2024-12-17
Automatic rebaseline of codesize expectations. NFC

sbc100:
This is an automatic change generated by tools/maint/rebaseline_tests.py.

The following (5) test expectation files were updated by running the tests with `--rebaseline`:
```
browser/test_small_js_flags.js.size: 4369 => 4345 [-24 bytes / -0.55%]
other/test_INCOMING_MODULE_JS_API.js.size: 3780 => 3758 [-22 bytes / -0.58%]
other/test_unoptimized_code_size.js.size: 53980 => 53979 [-1 bytes / -0.00%]
other/test_unoptimized_code_size_no_asserts.js.size: 29400 => 29399 [-1 bytes / -0.00%]
other/test_unoptimized_code_size_strict.js.size: 52776 => 52775 [-1 bytes / -0.00%]

Average change: -0.23% (-0.58% - -0.00%)
```


sbc100:
This is the first use of this script.  WDYT?   Perhaps we could format the results for easier readability?  Maybe in table?

kripken (approved):

kripken:
Putting the automatic output in triple-quotes, so the font is fixed-size, would be good I think.

aheejin:
Wouldn't it be good to run `./emcc --clear-cache` in the beginning too?

sbc100:
> Wouldn't it be good to run `./emcc --clear-cache` in the beginning too?

Yes we should have an option for that!

PR #23193, opened 2024-12-17
Remove the need for `runCaller`. NFC

sbc100:
This change removes the need for `runCaller` and the `calledRun` global. Instead we simply set `dependenciesFulfilled` to `run` directly each time to return early due to outstanding dependencies.

This also means that we don't need a reentrancy check on `doRun` since `dependenciesFulfilled` is only set in the case where `doRun` didn't actually run.

This both simplifies the code and saves on code size.


kripken (approved):
Nice!

PR #23198, opened 2024-12-17
Micro-optimize PATH.basename. NFC

sbc100:
The following (18) test expectation files were updated by                   
                                                                         
```                                                                      
other/codesize/test_codesize_cxx_ctors1.gzsize: 8507 => 8482 [-25 bytes / -0.29%]
other/codesize/test_codesize_cxx_ctors1.jssize: 20871 => 20818 [-53 bytes / -0.25%]
other/codesize/test_codesize_cxx_ctors2.gzsize: 8491 => 8466 [-25 bytes / -0.29%]
other/codesize/test_codesize_cxx_ctors2.jssize: 20839 => 20786 [-53 bytes / -0.25%]
other/codesize/test_codesize_cxx_except.gzsize: 9537 => 9516 [-21 bytes / -0.22%]
other/codesize/test_codesize_cxx_except.jssize: 24716 => 24663 [-53 bytes / -0.21%]
other/codesize/test_codesize_cxx_except_wasm.gzsize: 8467 => 8443 [-24 bytes / -0.28%]
other/codesize/test_codesize_cxx_except_wasm.jssize: 20765 => 20711 [-54 bytes / -0.26%]
other/codesize/test_codesize_cxx_except_wasm_exnref.gzsize: 8467 => 8443 [-24 bytes / -0.28%]
other/codesize/test_codesize_cxx_except_wasm_exnref.jssize: 20765 => 20711 [-54 bytes / -0.26%]
other/codesize/test_codesize_cxx_lto.gzsize: 8406 => 8389 [-17 bytes / -0.20%]
other/codesize/test_codesize_cxx_lto.jssize: 20455 => 20402 [-53 bytes / -0.26%]
other/codesize/test_codesize_cxx_mangle.gzsize: 9541 => 9523 [-18 bytes / -0.19%]
other/codesize/test_codesize_cxx_mangle.jssize: 24716 => 24663 [-53 bytes / -0.21%]
other/codesize/test_codesize_cxx_noexcept.gzsize: 8507 => 8482 [-25 bytes / -0.29%]
other/codesize/test_codesize_cxx_noexcept.jssize: 20871 => 20818 [-53 bytes / -0.25%]
other/codesize/test_codesize_files_js_fs.gzsize: 7646 => 7621 [-25 bytes / -0.33%]
other/codesize/test_codesize_files_js_fs.jssize: 18784 => 18731 [-53 bytes / -0.28%]
                                                                         
Average change: -0.26% (-0.33% - -0.19%)                                 
```


src/library_path.js
```diff
@@ -67,9 +67,7 @@ addToLibrary({
       // EMSCRIPTEN return '/' for '/', not an empty string
       if (path === '/') return '/';
       path = path.replace(/\/+$/g, "");
-      var lastSlash = path.lastIndexOf('/');
-      if (lastSlash === -1) return path;
-      return path.substr(lastSlash+1);
+      return path.split('/').pop();
```
kripken, line 70: This is smaller, but it will allocate garbage, which the old code avoided in at least one case, and even in the other, it only allocates a substring, not an array of substrings. Maybe avoiding allocation is not as critical here as in say the WebGL code, but I'm not sure.

sbc100: I dont't think filesystem operations that involve pathnames are likely to be on critical path are they?  I can see read/write operations maybe caring about this, but I think for string manipulation is probably just prefer size here.

sbc100: @juj WDYT?

hoodmane: File system operations in Emscripten are noticeably slow. I would say avoiding the allocation is preferable. Minified it's a difference of 53 bytes.

But if you want something even shorter, how about:
```js
(path) => path && path.match(/([^/]+|\/)\/*$/)[1]
```

sbc100: > Minified it's a difference of 53 bytes.

To be clear, are you saying that because you don't think saving 53 is a worth it?    I think we have make changes for 53 bytes in the past.  Those can add to be real benefits, especially for otherwise-small programs.   However, I guess if I you are using the FS library you program is already probably at least 10s of KiB in the FS library so so big.

sbc100: > But if you want something even shorter, how about:
> 
> ```js
> (path) => path && path.match(/([^/]+|\/)\/*$/)[1]
> ```
> 
> This also fixes the output for `'//'` which should be `'/'` not `''`

I love it!  Thanks

sbc100: I guess the `match` method still generates garbage in the form of two strings but I guess its O(1) and not O(elements in path)

sbc100: Done

kripken (approved):

PR #23203, opened 2024-12-17
WasmFS: Enable wasm64.test_unistd_access under wasmfs

sbc100:
Fix the signatures of several native wasmfs exports.  These are only needed/relevant under wasm64.


kripken (approved):

kripken:
Great!

PR #23208, opened 2024-12-18
Avoid excessive variable reuse in `modularize` function. NFC

sbc100:
I think this makes it a little easier to follow.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -2377,7 +2377,7 @@ def node_pthread_detection():
 def modularize():
   global final_js
   logger.debug(f'Modularizing, assigning to var {settings.EXPORT_NAME}')
-  src = read_file(final_js)
+  generated_js = read_file(final_js)
 
   # When targetting node and ES6 we use `await import ..` in the generated code
   # so the outer function needs to be marked as async.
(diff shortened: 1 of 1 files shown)


tools/link.py
```diff
@@ -2390,35 +2390,30 @@ def modularize():
     diagnostics.warning('emcc', 'EXPORT_NAME should not be named "config" when targeting Safari')
 
   if settings.MODULARIZE == 'instance':
-    src = '''
+    wrapper_funcion = '''
```
kripken, line 2393:
```suggestion
    wrapper_function = '''
```
also, this could be a f-string `f''' .. {generated_js}`

tools/link.py
```diff
@@ -2390,35 +2390,30 @@ def modularize():
...
-      'maybe_async': async_emit,
-      'src': src,
-    }
+''' % {'maybe_async': async_emit, 'generated_js': generated_js}
 
   if settings.MINIMAL_RUNTIME and not settings.PTHREADS:
     # Single threaded MINIMAL_RUNTIME programs do not need access to
     # document.currentScript, so a simple export declaration is enough.
-    src = '/** @nocollapse */ var %s = %s' % (settings.EXPORT_NAME, src)
+    src = f'/** @nocollapse */ var {settings.EXPORT_NAME} = {wrapper_funcion};'
```
kripken, line 2416: is the added `;` intentional?

sbc100: Yes, I think it was simply missing before.

tools/link.py
sbc100: I think it was deliberate in this case since it would force all the actual curly braces to be escaped (with `{{` and `}}`)

kripken (approved):

PR #23210, opened 2024-12-18
Assert if `MODULARIZE` factory function is used with `new`.

sbc100:
Once we start using `async` for for this function then `new` stops
working: https://github.com/emscripten-core/emscripten/pull/23157.

Using `new` in this was always an antipattern but there was nothing
stopping users from doing this.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,11 @@ See docs/process.md for more on how version tagging works.
 
 3.1.75 (in development)
 -----------------------
+- When using `-sMODULARIZE` we now assert if the factory function is called with
+  the JS `new` keyword.  e.g. `a = new Module()` rather than `b = Module()`.
+  This paves the way for marking the function as `async` which does not allow
+  `new` to be used.  This usage of `new` here was never documented and is
+  considered and antipattern. (#23210)
 - `PATH.basename()` no longer calls `PATH.normalize()`, so that
   `PATH.basename("a/.")` returns `"."` instead of `"a"` and
   `PATH.basename("a/b/..")` returns `".."` instead of `"a"`. This is in line with
--- a/tools/link.py
+++ b/tools/link.py
@@ -2452,6 +2452,19 @@ var %(EXPORT_NAME)s = (() => {
         'wrapper_function': wrapper_function,
       }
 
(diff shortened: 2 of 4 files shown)


ChangeLog.md
```diff
@@ -20,6 +20,11 @@ See docs/process.md for more on how version tagging works.
 
 3.1.75 (in development)
 -----------------------
+- When using `-sMODULARIZE` we now assert if the factory function is called with
+  the JS `new` keyword.  e.g. `var a = new Module()` rather than `var b =
+  Module()`.  This paves the way for marking the function as `async` which does
```
kripken, line 25: The split of backticks over lines looks like it is breaking formatting here, so maybe best to keep `var b = Module()` on one line

ChangeLog.md
```diff
@@ -20,6 +20,11 @@ See docs/process.md for more on how version tagging works.
...
+  not allow `new` to beused.  This usage of `new` here was never documented and
```
kripken, line 26:
```suggestion
  not allow `new` to be used.  This usage of `new` here was never documented and
```

tools/link.py
```diff
@@ -2377,80 +2377,101 @@ def node_pthread_detection():
...
+        'wrapper_function': wrapper_function,
       }
+  elif add_assertion_wrapper:
+    src = '''\
+var %(EXPORT_NAME)s = (() => {
+  %(wrapper_function)s
+  // Return a small, never-async wrapper around %(wrapper_name)s which
+  // checks for callers using `new`.
+  return function(arg) {
+    if (new.target) throw new Error("%(EXPORT_NAME)s() must not be called with new");
```
kripken, line 2463:
```suggestion
    if (new.target) throw new Error("%(EXPORT_NAME)s() must not be called as new %(EXPORT_NAME)s().");
```
I think otherwise the sentence might not be always understood by users (it could look clipped off, `X must not be called with new` - leading to "new what?" questions).

ChangeLog.md
sbc100: That seems odd..

sbc100:
I re-wrote this change is much more minimal way.  PTAL

PR #20191, opened 2023-09-05
some help for the closure jscompiler

walkingeyerobot:

PR #23216, opened 2024-12-18
Remove unnecessary @nocollape. NFC

sbc100:
This was added back in #20191 but I can't see how it can work.  The thing is being applied to here is a function variable. e.g.:

```
var Modulle = function() { ... return the a new module };
```

From the documentation for @nocollape is not clear to me that it has any effect on normal variables like this.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -2414,7 +2414,7 @@ export default async function init(moduleArg = {}) {
   if settings.MINIMAL_RUNTIME and not settings.PTHREADS:
     # Single threaded MINIMAL_RUNTIME programs do not need access to
     # document.currentScript, so a simple export declaration is enough.
-    src = f'/** @nocollapse */ var {settings.EXPORT_NAME} = {wrapper_function};'
+    src = f'var {settings.EXPORT_NAME} = {wrapper_function};'
   else:
     script_url_node = ''
     # When MODULARIZE this JS may be executed later,


sbc100:
@walkingeyerobot I think perhaps i'm misunderstanding here... do you happen to remember why this was needed?

kripken:
Hmm, good question. The [docs](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#nocollapse) do mention that if applied on an object, it prevents collapse of that object's properties - perhaps this was for properties on top of that function?

sbc100:
> Hmm, good question. The [docs](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#nocollapse) do mention that if applied on an object, it prevents collapse of that object's properties - perhaps this was for properties on top of that function?

But closure runs statically.. there are no static properties of this function.

kripken:
Yeah, I don't recall us setting any properties on this in the past... so not sure what the reason was.

If @walkingeyerobot doesn't recall the reason either, sgtm to remove.

sbc100:
Thanks @walkingeyerobot.   BTW,  I guess we should add some upstream emscripten test for running closure after emscripten runs so that we stay on top of this and don't cause regressions for you?

PR #23217, opened 2024-12-18
Fix dup of pipe

### LINKED ISSUE #22030: “TypeError: Cannot read properties of null (reading 'length')” with pipe + dup&dup2

3.1.54

```shell
# emscripten
emcc pipe.cpp -o main.js
node main.js

# g++
g++ pipe.cpp  -o main
./main

```

Hi, I was using `pipe` with `dup`, `dup2` for read/write redirection. And I found when doing something like the following programs, it will raise an exception `TypeError: Cannot read properties of null (reading 'length')`. 

In brief, here I'm trying to redirect the `stderr` to the write end of the pipe, then things originally written to file descriptor 2(`stderr`) could be read from the pipe. And it uses `original_fd` to restore the status of `stderr`.

(Additionally, here I even tried to replace the `stderr` with some other regular file, like `temp.txt`, but the program still throws.)

It seems to be an issue of pipe working with dup&dup2. It'd be great if you can check this. Thank you!

#### Code
```c++

#include <iostream>
#include <unistd.h>
#include <string.h>

int main() {
    // FILE* file_ = fopen("temp.txt", "w+");
    FILE* file_ = stderr;
    int fd = ::fileno(file_);
    int original_fd;
    int read_end_fd;

    int pipefd[2];


    original_fd = ::dup(fd); // duplicate stderr to original_fd, and original_fd is used to restore stderr later

    if (pipe(pipefd) == -1) {
        perror("pipe");
        return 1;
    }

    read_end_fd = pipefd[0];
    pipefd[0] = -1;
    int write_end_fd = pipefd[1];

    ::dup2(write_end_fd, fd); // now things write to fd(stderr) is redirected to write_end_fd
    ::close(write_end_fd); // close the write end of the pipe after duplicating

    ::write(fd, "x", 1); // write to the stderr, expected to be read from pipe

    ::dup2(original_fd, fd); //  restore fd (stderr) to its original state
    ::close(original_fd);

    char buffer[10];
    printf("flag before read\n");
    int ret = ::read(read_end_fd, buffer, 1); // Read from the pipe
    
    printf("ret is: %d\n", ret);
    printf("read content: %s\n", buffer);
    printf("error: %s\n", strerror(errno));

    ::close(read_end_fd); // Close the read end of the pipe

    return 0;
}
```
#### Results
Here is the execution result. The wasm execution raised an exception of `Cannot read properties of null (reading 'length')`. And for the native execution, it runs successfully.

```shell
# emscripten
flag before read
/data/main.js:128
      throw ex;
      ^

TypeError: Cannot read properties of null (reading 'length')
    at Object.read (/data/main.js:3972:44)
    at Object.read (/data/main.js:3159:43)
    at doReadv (/data/main.js:4171:23)
    at _fd_read (/data/main.js:4186:17)
    at wasm://wasm/0001046a:wasm-function[44]:0xe45
    at wasm://wasm/0001046a:wasm-function[10]:0x4b9
    at wasm://wasm/0001046a:wasm-function[11]:0x574
    at /data/main.js:691:12
    at callMain (/data/main.js:4689:15)
    at doRun (/data/main.js:4739:23)

# g++
flag before read
ret is: 1
read content: x
error: Success

```

#### Another pattern
**I also found another similar pattern which seems to be the same issue, as shown here:**
```c++
#include <iostream>
#include <unistd.h>
#include <string.h>

int main() {
    int pipefds[2];
    if (pipe(pipefds) == -1) {
        perror("pipe");
        return 1;
    }
    int pipefd_write = pipefds[1];
    write(pipefd_write, "x", 1);

    int original_fd = dup(pipefd_write); // to restore the pipefd_write later.

    int tempfds[2]; // only tempfds[1], the write end, is used here.
    if (pipe(tempfds) == -1) {
        perror("pipe");
        return 1;
    }

    dup2(tempfds[1], pipefd_write); // replace pipefd_write with tempfds[1], redirecting pipefd_write to tempfds[1]
    close(tempfds[1]); // close
    

    dup2(original_fd, pipefd_write); // recover pipefd_write

    close(original_fd);
    close(tempfds[0]); // we don't use tempfds[0] in this example
    
    close(pipefd_write);

    char buf[10];
    read(pipefds[0], buf, 1);
    printf("buf: %s\n", buf);

    return 0;
}
```

Compilation command are the same. And the results are also similar to the first code snippet
```shell
# emcc
...
      throw ex;
      ^

TypeError: Cannot read properties of null (reading 'length')
...

# g++
buf: x
```

hoodmane:
Resolves #22030

### DIFF
--- a/src/library_pipefs.js
+++ b/src/library_pipefs.js
@@ -77,6 +77,9 @@ addToLibrary({
 
         return 0;
       },
+      dup(stream) {
+        stream.node.pipe.refcnt++;
+      },
       ioctl(stream, request, varargs) {
         return {{{ cDefs.EINVAL }}};
       },
--- a/test/unistd/pipe.c
+++ b/test/unistd/pipe.c
@@ -58,7 +58,7 @@ void test_poll(int *fd, int data_available) {
   assert(pfds[1].revents == POLLOUT);
 }
 
-int main() {
+int test_most() {
   int fd[2];
   unsigned char wchar = 0;
   unsigned char rchar = 0;
@@ -152,5 +152,38 @@ int main() {
   assert(errno == EBADF);
 
   puts("success");
+  return 0;
+}
+
+int test_redirect_stderr_to_pipe() {
+  int stderrfd = fileno(stderr);
+  int pipefd[2];
+  int original_fd = dup(stderrfd); // duplicate stderr to original_fd, and original_fd is used to restore stderr later
+  assert(original_fd >= 0);
+  assert(pipe(pipefd) == 0);
+
+  int read_end_fd = pipefd[0];
+  int write_end_fd = pipefd[1];
+
+  assert(dup2(write_end_fd, stderrfd) == stderrfd); // now things write to fd(stderr) is redirected to write_end_fd
+  assert(close(write_end_fd) == 0); // close the write end of the pipe after duplicating
+
+  assert(write(stderrfd, "xyz", 3) == 3); // write to the stderr, expected to be read from pipe
+
+  assert(dup2(original_fd, stderrfd) == stderrfd); //  restore fd (stderr) to its original state
+  assert(close(original_fd) == 0);
+
+  char buffer[10];
+  memset(buffer, 0, 10);
+  assert(read(read_end_fd, buffer, 10) == 3);
+  assert(strcmp(buffer, "xyz") == 0);
+  assert(close(read_end_fd) == 0); // Close the read end of the pipe
+  printf("success\n");
+  return 0;
+}
+
+int main() {
+  test_most();
+  test_redirect_stderr_to_pipe();
   return 0;
 }


src/library_pipefs.js
```diff
@@ -77,6 +77,9 @@ addToLibrary({
 
         return 0;
       },
+      dup(stream) {
+        stream.node.pipe.refcnt++;
+      },
```
sbc100, line 82: Without this would the dup syscall simply fail at runtime?

hoodmane: No dup succeeds because `FS.dupStream` assumes that no work needs to be done if `stream_ops.dup` is undefined. But then there are crashes caused by use after free.

sbc100: Should `FS.dupStream` require that `stream_ops.dup` exists?

hoodmane: No, because in many cases it doesn't need to do anything. For instance in memfs, streams can be duplicated with no extra logic. It's pretty much only streams that hold a recounted resource that need a `stream_ops.dup`.

PR #23218, opened 2024-12-18
Opening a directory with O_CREAT should return EISDIR

hoodmane:
So that the behavior matches the behavior on linux. This makes the test pass on nodefs and noderawfs. This allows merging the test added in #23137 into test_fcntl_open.c


sbc100 (approved):

hoodmane:
This seems to require also changing the behavior of `open` on a directory in wasmfs.

hoodmane:
It's a simple enough fix in wasmfs too.

sbc100 (approved):
Great!

PR #23219, opened 2024-12-18
Remove unnecessary `await` in `MODULARIZE=instance` mode. NFC

sbc100:
Since the init function is already marks as async it fine/better to simply return the promise here rather than awaiting on it.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -2392,7 +2392,7 @@ export default async function init(moduleArg = {}) {
 
 %(generated_js)s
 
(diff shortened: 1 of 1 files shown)


kripken (approved):
iiuc this is unneeded as the caller must await anyhow?

sbc100:
> iiuc this is unneeded as the caller must await anyhow?

Yes, either way the caller will receive a promise in return, since that is what `async` on the declaration guarantees.

PR #23222, opened 2024-12-19
Use consistent prefix map when building deterministic system libraries

aheejin:
When `deterministic_paths` is set, we are currently using `-ffile-prefix-map` to produce the same path in data and debug info. In the case of absolute paths, their emscripten path is replaced with a fake path `/emsdk/emscripten`, and in the case of relative paths, all path relative to the emscripten directory is removed, so `../../system/lib/somefile.c` becomes `system/lib/somefiles.c`. https://github.com/emscripten-core/emscripten/blob/f66b5d706e174d9e5cc6122c06ea29dcd2735cd0/tools/system_libs.py#L472-L477 https://github.com/emscripten-core/emscripten/blob/f66b5d706e174d9e5cc6122c06ea29dcd2735cd0/tools/system_libs.py#L495-L501

So this does not make relative paths and absolute paths the same, which can lead to different builds depending on whether the command line uses absolute paths vs. relative ones. Currently we use relative paths when `EMCC_BATCH_BUILD` is set. And Ninja builds cannot use relative paths.

This is also what was suggested in https://github.com/emscripten-core/emscripten/issues/23195#issuecomment-2549784273 while discussins `__FILE__` problem in #23195.

This does not change any code size tests because no code size tests happens to contain `__FILE__` at the moment. (One will be added in #22994)

### DIFF
--- a/tools/system_libs.py
+++ b/tools/system_libs.py
@@ -42,9 +42,9 @@ LIBC_SOCKETS = ['socket.c', 'socketpair.c', 'shutdown.c', 'bind.c', 'connect.c',
 # link time.
 USE_NINJA = int(os.environ.get('EMCC_USE_NINJA', '0'))
 
-# A fake emscripten path to use in __FILE__ macro and debug info to produce
-# reproducible builds across platforms.
-FAKE_EMSCRIPTEN_PATH = '/emsdk/emscripten'
+# A (fake) deterministic emscripten path to use in __FILE__ macro and debug info
+# to produce reproducible builds across platforms.
+DETERMINISITIC_PREFIX = '/emsdk/emscripten'
 
 
 def files_in_path(path, filenames):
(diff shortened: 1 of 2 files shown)


sbc100:
Looks like the test_dwarf_system_lib failure is real.

Can we come up with a better PR title?   Something like "Use consistent prefix map when building deterministic system libraries"?

tools/system_libs.py
```diff
@@ -42,6 +42,10 @@
 # link time.
 USE_NINJA = int(os.environ.get('EMCC_USE_NINJA', '0'))
 
+# A fake emscripten path to use in __FILE__ macro and debug info to produce
+# reproducible builds across platforms.
+FAKE_EMSCRIPTEN_PATH = '/emsdk/emscripten'
```
sbc100, line 47: How about `DETERMINISITIC_PREFIX`?

sbc100 (approved):
lgtm % comments and tests passing

tools/system_libs.py
aheejin: Done

aheejin:
> Looks like the test_dwarf_system_lib failure is real.

Fixed.
 
> Can we come up with a better PR title? Something like "Use consistent prefix map when building deterministic system libraries"?

Done.

sbc100 (approved):

tools/system_libs.py
```diff
@@ -496,9 +500,9 @@ def build_objects(self, build_dir):
       source_dir = utils.path_from_root()
       if batch_inputs:
         relative_source_dir = os.path.relpath(source_dir, build_dir)
-        cflags += [f'-ffile-prefix-map={relative_source_dir}/=']
-      cflags += [f'-ffile-prefix-map={source_dir}=/emsdk/emscripten',
-                 '-fdebug-compilation-dir=/emsdk/emscripten']
+        cflags += [f'-ffile-prefix-map={relative_source_dir}={DETERMINISITIC_PREFIX}']
+      cflags += [f'-ffile-prefix-map={source_dir}={DETERMINISITIC_PREFIX}',
+                 f'-fdebug-compilation-dir={DETERMINISITIC_PREFIX}']
```
sbc100, line 505: I wonder if these two chunks could be combined into `self.get_cflags()` rather than duplicated?   Can be part of the followup of course.

aheejin: Done in #23225.

PR #23227, opened 2024-12-19
Simplify createWasm. NFC

sbc100:
I have no idea who this seems to be a minor code size regression.


src/preamble.js
```diff
@@ -472,9 +472,6 @@ function abort(what) {
   /** @suppress {checkTypes} */
   var e = new WebAssembly.RuntimeError(what);
 
-#if MODULARIZE
-  readyPromiseReject(e);
-#endif
```
kripken, line 477: Why is this no longer needed?

sbc100: I've marked this is a draft for now.  I'll come back to it in the new year .

Issue #19053, opened 2023-03-23
Create versioning policy emscripten


sbc100:
We should decide if we even plan to use the major or minor version numbers again in emscripten or if we want to simply continue to use patch versions.

If we are going to drop major and minor bumps should we just move so a single number the chrome and firefix do it?   

Split out from discussion in https://github.com/emscripten-core/emscripten/pull/18191#issuecomment-1480173774

sbc100:
@tlively @juj @kripken @dschuff 

Thomas has made that argument that there is quite of benefit in being able to compare two version and get a good idea of how far apart they are in time.  For example with Chrome its obvious that v88 is 12 versions older than v100 

I've never been a huge fan of semantic versioning myself, so I'd been keen to drop the pretense that that is what we are doing, and admit that there is no such things as a patch-only, risk-free upgrade.  I think a single monotonic version is easier to reason about and its also what we have been doing anyway for almost all our emscripten releases to years ( we just happen to have been using the patch level as this number).

curiousdannii:
One option is to simply use the date as the version number.

PR #23235, opened 2024-12-19
Make the next version emscripten 4.0.0!

sbc100:
We have a whole bunch of significant change in the next release:

- 3 new wasm features enabled by default
- Use of async/await internally and in MODULARIZE mode
- LLVM libraries updated for v19.1
- Lots of old polyfills finally removed

Also, 74 revisions the 3.1 release is gettting pretty long in the tooth.

Plus it will be the first release of the new year!


sbc100:
The wider discussion of the future of versioning is still up in the air: https://github.com/emscripten-core/emscripten/issues/19053, but that shouldn't causes us version paralysis in the meantime.

kripken (approved):
Sounds good!

Let's give it a longer than usual time to bake before the release, though, given all these changes. Anyhow, we can think about that next year...

msqr1:
Finally...!

cclauss:
Congrats on reaching v4.  Should the testing versions be upgraded, or are they kept back on purpose?

`.circleci/config.yml`
```yaml
executors:
  linux-node:
    docker:
      - image: cimg/node:18.20.3
  linux-python:
    docker:
      - image: cimg/python:3.10.7
  bionic:
    docker:
      - image: emscripten/emscripten-ci:focal
```

sbc100:
> Congrats on reaching v4. Should the testing versions be upgraded, or are they kept back on purpose?
> 
> `.circleci/config.yml`
> 
> ```yaml
> executors:
>   linux-node:
>     docker:
>       - image: cimg/node:18.20.3
>   linux-python:
>     docker:
>       - image: cimg/python:3.10.7
>   bionic:
>     docker:
>       - image: emscripten/emscripten-ci:focal
> ```

I don't think the first two really matter.  But the third one we choose deliberately to make sure we test on older versions of linux.  If we want to drop support for the version of glibc in Ubuntu/Focal we could update this, but it don't any pressing reason for do that (at least not that I know of).

PR #23236, opened 2024-12-19
rebaseline_tests.py: Fix emcc location

sbc100:


sbc100:
Actually I'll just include this in #23226 I think

PR #23237, opened 2024-12-19
Avoid extra whitespace when using `asyncIf` macro. NFC

sbc100:


dschuff (approved):
saving the web, one byte at a time....

sbc100:
> saving the web, one byte at a time....

Its mostly just for readability of the generated code :)   Probably doesn't matter to anyone except me.

Issue #13937, opened 2021-04-20
Inform Acorn that the source type is a Module when EXPORT_ES6


kripken:
That allows syntax that is only valid in modules.

See #13894 for discussion.

Relevant code is in `tools/acorn-optimizer.js` which is where Acorn is invoked. It should get another flag like the existing `closureFriendly` from `tools/building.py` (which calls the JS code, and can check `EXPORT_ES6`).

Issue #13894, opened 2021-04-14
Using JavaScript module export in files supplied via --pre-js/--post-js fails with a SyntaxError


ackh:
I'm attempting to export some symbols in the JavaScript code that I pass to `emcc`'s `--pre-js` and `--post-js` in order to provide an easier public JavaScript interface for my WebAssembly component. However, if I add `export { myFunction1, myFunction2 };` at the end of the `--post-js` file I'm getting the following error message:

````
SyntaxError: 'import' and 'export' may appear only with 'sourceType: module' (3664:0)
export { myFunction1, myFunction2 };
^

    at Parser.pp$4.raise (/mnt/c/Tools/emsdk/upstream/emscripten/node_modules/acorn/dist/acorn.js:2927:15)
    at Parser.pp$1.parseStatement (/mnt/c/Tools/emsdk/upstream/emscripten/node_modules/acorn/dist/acorn.js:870:18)
    at Parser.pp$1.parseTopLevel (/mnt/c/Tools/emsdk/upstream/emscripten/node_modules/acorn/dist/acorn.js:755:23)
    at Parser.parse (/mnt/c/Tools/emsdk/upstream/emscripten/node_modules/acorn/dist/acorn.js:555:17)
    at Function.parse (/mnt/c/Tools/emsdk/upstream/emscripten/node_modules/acorn/dist/acorn.js:578:37)
    at Object.parse (/mnt/c/Tools/emsdk/upstream/emscripten/node_modules/acorn/dist/acorn.js:5095:19)
    at Object.<anonymous> (/mnt/c/Tools/emsdk/upstream/emscripten/tools/acorn-optimizer.js:1819:15)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32) {
  pos: 126980,
  loc: Position { line: 3664, column: 0 },
  raisedAt: 126986
}
emcc: error: '/mnt/c/Tools/emsdk/node/14.15.5_64bit/bin/node /mnt/c/Tools/emsdk/upstream/emscripten/tools/acorn-optimizer.js /tmp/emscripten_temp_a16t4n3q/math.js.pp.js AJSDCE --closureFriendly' failed (1)
````
From what I see, `emcc` invokes [acorn](https://github.com/acornjs/acorn) which eventually causes that error. I'm unsure about how to fix that issue.

[...]

PR #23240, opened 2024-12-19
Remove test_export_es6_allows_export_in_post_js. NFC

sbc100:
This test was added in #13937.  Here we are testing something that we actually decided we did not want to support.  Specifically in #13894 we decided that `--extern-post-js` was the place to allow this.

More importantly the resulting JS module simple doesn't work because in `MODULARIZE` mode the code from `--post-js` does not end up at the top level but instead inside a factory function.  Note that this test does not try to exectute the resulting JS.  If you try to execute it you get `SyntaxError: Unexpected token 'export'`.

Since we don't support it I think its best to remove this test.

I'm also working on a change to move the modularization code before acorn and closure, and with that change this won't even compile (because the modularization process puts all the code inside and inner scope).

### DIFF
--- a/test/export_module.js
+++ b/test/export_module.js
@@ -1,5 +1,0 @@
-function doNothing() {
-  return false;
-}
-
-export {doNothing};
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -425,11 +425,6 @@ class other(RunnerCore):
                             '-sENVIRONMENT=node', '-sEXPORT_ES6', '-sUSE_ES6_IMPORT_META=0'])
     self.assertContained('EXPORT_ES6 and ENVIRONMENT=*node* requires USE_ES6_IMPORT_META to be set', err)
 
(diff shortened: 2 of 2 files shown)


kripken:
Sounds good, but what about leaving the test and changing the post-js to an extern-post-js?

sbc100:
> Sounds good, but what about leaving the test and changing the post-js to an extern-post-js?

That doesn't really test anything.. since we don't run the extern-post-js through any of our tools like acorn or closure.

sbc100:
> > Sounds good, but what about leaving the test and changing the post-js to an extern-post-js?
> 
> That doesn't really test anything.. since we don't run the extern-post-js through any of our tools like acorn or closure.

You can already put whatever you want in extern-post-js.. including completely invalid stuff, and we just copy into the output for you.

kripken (approved):
I see, yeah, I guess we don't need to test that.

PR #23258, opened 2024-12-28
fix memory leak in setOffscreenCanvasSizeOnTargetThread

ypujante:
As I was debugging my code, I noticed that the memory was allocated (`stringToNewUTF8`) but not freed afterwards.

I may be wrong but that looks like a memory leak, hence the fix...

### DIFF
--- a/src/library_html5.js
+++ b/src/library_html5.js
@@ -2307,6 +2307,7 @@ var LibraryHTML5 = {
       targetCanvasPtr = stringToNewUTF8(targetCanvas);
     }
     __emscripten_set_offscreencanvas_size_on_thread(targetThread, targetCanvasPtr, width, height);
+    _free(targetCanvasPtr);
   },
 #endif
 


ypujante:
From the failed test, it seems clear that this change is causing issues:

```text
FAIL [5.183s]: test_webgl_resize_offscreencanvas_from_main_thread_offscreen_proxy (test_browser.browser_2gb)
...
  File "/root/project/test/common.py", line 2265, in run_browser
  File "/root/project/test/common.py", line 1585, in assertContained
AssertionError: Expected to find '/report_result?exit:0
' in '//report_result?exception:Failed to execute 'querySelector' on 'Document': '< ' is not a valid selector. / SyntaxError: Failed to execute 'querySelector' on 'Document': '< ' is not a valid selector.
    at findCanvasEventTarget (http://localhost:8888/test.js:2789:47)
    at setCanvasElementSizeCallingThread (http://localhost:8888/test.js:2793:16)
    at setCanvasElementSizeMainThread (http://localhost:8888/test.js:2836:10)
    at __emscripten_receive_on_main_thread_js (http://localhost:8888/test.js:2143:13)
    at http://localhost:8888/test.wasm:wasm-function[104]:0x28c7
    at http://localhost:8888/test.wasm:wasm-function[102]:0x2741
    at http://localhost:8888/test.wasm:wasm-function[97]:0x25ff
    at http://localhost:8888/test.wasm:wasm-function[73]:0x1b27
    at http://localhost:8888/test.wasm:wasm-function[82]:0x1e94
    at http://localhost:8888/test.wasm:wasm-function[73]:0x1b27
', diff:
+++ actual
```

So I am just going to close it.

sbc100:
I believe this is free'd on the target thread in `do_set_size` in `system/lib/html5/offscreencanvas.c`

ypujante:
@sbc100 Actually I don't think it is:

The pointer `target` in `_emscripten_set_offscreencanvas_size_on_thread` is **copied** into a newly allocated structure `set_cavas_size_t` and then `do_set_size` frees that structure, not the `target` part of the structure.

Unless I am missing something this code does not free the allocated `target` argument.

```c
typedef struct set_cavas_size_t {
  const char* target;
  int width;
  int height;
} set_cavas_size_t;

static void do_set_size(void* arg) {
  set_cavas_size_t* args = (set_cavas_size_t*)arg;
  _emscripten_set_offscreencanvas_size(args->target, args->width, args->height);
  free(arg);
}

void _emscripten_set_offscreencanvas_size_on_thread(pthread_t t,
                                                    const char* target,
                                                    int width,
                                                    int height) {
  set_cavas_size_t* arg = malloc(sizeof(set_cavas_size_t));
  arg->target = target;
  arg->width = width;
  arg->height = height;

  em_proxying_queue* q = emscripten_proxy_get_system_queue();

  // Note: If we are also a pthread, the call below could theoretically be
  // done synchronously. However if the target pthread is waiting for a
  // mutex from us, then these two threads will deadlock. At the moment,
  // we'd like to consider that this kind of deadlock would be an Emscripten
  // runtime bug, although if emscripten_set_canvas_element_size() was
  // documented to require running an event in the queue of thread that owns
  // the OffscreenCanvas, then that might be ok.  (safer this way however)
  if (!emscripten_proxy_async(q, t, do_set_size, arg)) {
    assert(false && "emscripten_proxy_async failed");
  }
}
```

In any case, there is 0 documentation in either the names of the parameters, the name of the functions or comments of what is supposed to happen to the memory. At best this causes confusion, at worst this leads to errors.

sbc100:
Indeed!  It looks like there is a memory leak there.   This is all internal stuff so its no surprise that that there is little or no docs here.  Re-opening so we can get this fixed.

sbc100:
Actually I guess we should open a different bug.

PR #23260, opened 2024-12-30
Cleanup randomFill. NFC

sbc100:
Under node we can rely on `randomFillSync` always being available since
its been around since v9.0.0:

https://nodejs.org/api/crypto.html#cryptorandomfillsyncbuffer-offset-size

Also, remove the return value from this function since it always fills
the entire buffer.

https://caniuse.com/getrandomvalues


src/library_fs.js
```diff
@@ -1411,7 +1411,8 @@ FS.staticInit();
       var randomBuffer = new Uint8Array(1024), randomLeft = 0;
       var randomByte = () => {
         if (randomLeft === 0) {
-          randomLeft = randomFill(randomBuffer).byteLength;
+          randomFill(randomBuffer)
```
kripken, line 1414:
```suggestion
          randomFill(randomBuffer);
```

kripken (approved):

Issue #23158, opened 2024-12-13
Closure only run on the subset of the generated JS code not, for example, modularized code


sbc100:
Since we do modularization after we run closure compiler this can lead to some strange effects.  For example, if we try to use `await` in the main out closure will see it as a top-level-await, even though in modularize mode its inside the factory function.

PR #23261, opened 2024-12-30
Move modularization code into js compiler

sbc100:
Prior to this change the mudularization was done as a post-processing step in python.  This complicated things since acorn and closure passes would only see the inner code and not the whole module.

One concrete benefit is that we can now use `await` in the body of the factory function since closure no longer sees it as top level (which it isn't).

Fixes: #23158


sbc100:
WDYT @brendandahl ?

PR #23262, opened 2024-12-30
Remove `POLYFILL_OLD_MATH_FUNCTIONS` setting

sbc100:
These polyfills are only needed on ancient browser versions that there
should be no need to support.


kripken (commented):
I'm confused, where is this documented?

> Our minimum supported browser versions are currently chrome 45, firefox 40, safari 10.1 and node 10.19.

I searched for some of those version numbers under `src/` and `site/` and can't find them. The docs for MIN_CHROME say something else, 33 instead of 45:

https://github.com/emscripten-core/emscripten/blob/5a8d9e52bd6bd34e0455658af2f4a188f78e8ad3/src/settings.js#L1922

33 would be too old for this change IIANM, I think we need 38 according to MDN. But I think it should be fine to bump to that, separately?

sbc100:
> I'm confused, where is this documented?

Oops, I was looking at feature_matrix.py which was out of sync for some reason: https://github.com/emscripten-core/emscripten/pull/23264.

kripken (approved):

PR #23077, opened 2024-12-04
Bump minimum supported chrome version from 32 to 33

sbc100:
This allows us to remove the promise polyfill: https://caniuse.com/promises

PR #23264, opened 2024-12-30
Update docs on oldest supported browser versions.  NFC

sbc100:
These were changed in #23077 and #23118 but the docs were not updated.

### DIFF
--- a/site/source/docs/tools_reference/settings_reference.rst
+++ b/site/source/docs/tools_reference/settings_reference.rst
@@ -2905,7 +2905,8 @@ are desired to work. Pass -sMIN_FIREFOX_VERSION=majorVersion to drop support
 for Firefox versions older than < majorVersion.
 Firefox 79 was released on 2020-07-28.
 MAX_INT (0x7FFFFFFF, or -1) specifies that target is not supported.
-Minimum supported value is 34 which was released on 2014-12-01.
+Minimum supported value is 40 which was released on 2015-09-11 (see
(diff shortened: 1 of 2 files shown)


tools/feature_matrix.py
```diff
@@ -15,12 +15,12 @@
 
 UNSUPPORTED = 0x7FFFFFFF
 
-# Oldest support browser versions.  These have been set somewhat
-# arbitrarily for now.
+# Oldest support browser versions.  Keep in sync with
+# src/settings.js.
```
kripken, line 19: Please add similar comments in `settings.js` (that would have led me to this file, for example).

Or perhaps we can remove the numbers in this file and just say "see minimum versions in settings.js"?

sbc100: The comments in settings.js are generally for users of the settings, not us developers, so I'm not sure it makes sense to add such comments there.

Sadly we cannot remove these since there are needed at runtime.

kripken: Hmm, good points. But I am quite worried that we will decide to bump the minimum version of some browser, and only remember to do that in `settings.js`. I, for example, completely forgot about version numbers in this file just now, and didn't even think to search in this directory.

Perhaps `settings.js` could just add `see feature_matrix.py for more` comments?

sbc100: OK done.  PTAL

kripken (approved):
Great!

PR #23265, opened 2024-12-30
Revert removal of `createRequire`

sbc100:
This change effectively reverts #23169 because it turns out there are quite a few other places in the codebase that depend on being able to call `require`.


kripken:
What is the advantage? On code size it doesn't seem to help here IIUC

sbc100:
See https://github.com/emscripten-core/emscripten/pull/23169#issuecomment-2556691146.

We have a lot of uses of `require` in the codebase.  So this is basically fixing a regression caused by #23169

kripken:
I see, but if it's fixing a regression, why doesn't code size improve here? Or is the regression not covered by existing tests?

sbc100:
> I see, but if it's fixing a regression, why doesn't code size improve here? Or is the regression not covered by existing tests?

The regression is not covered by the existing tests.   We would need to add tests that target each of our `require` usages, and combine them with `-sMODULARIZE`.

The code size regressions are pretty minor and bascially the inverse of the improvments from #23169

kripken (approved):
Got it, thanks, sounds good.

kleisauke (approved):

PR #23275, opened 2025-01-01
Fix memory leak in `_emscripten_set_offscreencanvas_size_on_thread`

ypujante:
This is the follow up to #23258 to actually fix the memory leak.

### DIFF
--- a/system/lib/html5/offscreencanvas.c
+++ b/system/lib/html5/offscreencanvas.c
@@ -3,7 +3,7 @@
 #include "emscripten_internal.h"
 
 typedef struct set_cavas_size_t {
-  const char* target;
+  char* target;
   int width;
   int height;
 } set_cavas_size_t;
@@ -19,7 +19,7 @@ static void do_set_size(void* arg) {
 // - DO NOT call this function with a string literal
 // - DO NOT free the memory after calling this function
 void _emscripten_set_offscreencanvas_size_on_thread(pthread_t t,
-                                                    const char* target,
+                                                    char* target,
                                                     int width,
                                                     int height) {
   set_cavas_size_t* arg = malloc(sizeof(set_cavas_size_t));
--- a/system/lib/html5/offscreencanvas.c
+++ b/system/lib/html5/offscreencanvas.c
@@ -3,7 +3,7 @@
 #include "emscripten_internal.h"
 
 typedef struct set_cavas_size_t {
-  char* target;
+  const char* target;
   int width;
   int height;
 } set_cavas_size_t;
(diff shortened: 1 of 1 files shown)


system/lib/html5/offscreencanvas.c
```diff
@@ -3,23 +3,27 @@
...
-  const char* target;
+  char* target;
   int width;
   int height;
 } set_cavas_size_t;
 
 static void do_set_size(void* arg) {
   set_cavas_size_t* args = (set_cavas_size_t*)arg;
   _emscripten_set_offscreencanvas_size(args->target, args->width, args->height);
+  free(args->target);
```
sbc100, line 14: Can you just do `free((char*)args->target);` here rather than removing the const elsewhere?

system/lib/html5/offscreencanvas.c
```diff
@@ -3,23 +3,27 @@
...
   free(arg);
 }
 
+// Warning: this function takes ownership of the "target" parameter
+// - DO NOT call this function with a string literal
+// - DO NOT free the memory after calling this function
```
sbc100, line 20: I'm not sure we need to this much emphasis here.   How about just `This function takes ownership of the "target" string`.

sbc100 (approved):

system/lib/html5/offscreencanvas.c
ypujante: I made the changes you requested. I just wanted to point out that using `char *` instead of `char const *` was preventing an issue at compile time:

```text
// void foo(char *s) {}

<source>:8:9: warning: ISO C++11 does not allow conversion from string literal to 'char *' [-Wwritable-strings]
    8 |     foo("abc");
      |         ^
1 warning generated.
```

With `const` this call call will compile but will lead to a UB at runtime...

Again I did what you asked/wanted and changed it back

sbc100: I'm not sure I understand what issue you are referring to exactly.  Clearly its UB if you pass an actual string constant to this function since we cannot call `free` on a string constant. 

Note that this function is purely internal and only ever called from JS so I'm not sure its worth discussing possible compiler errors at call sites since there are not C/C++ call sites for this function,

ypujante: I understand that today it is not an issue. But this internal call may be called in the future and who knows what might happen. Having the safety of the compiler was in my mind a way to future proof the call for what seemed to be a very reasonable change (with 0 impact on performance or code size). It seems weird for me to argue about such a trivial change which made the code more robust, but it is not my project and you make the decisions. I am not trying to impose anything. I implemented this PR as a courtesy because I am the one who discovered the bug and wanted to take it off your plate.

sbc100: I absolutely do appreciate the PR!    I'm happy to land it as it stands now.

Regarding the `const` here I feel like it makes it clear that the string is an input and not an output (i.e. its not going to be written to).

I also like this leaving the `const` keeps this PR very small (just a single line change!).  That helps with reviewing and auditing change I think.

PR #23278, opened 2025-01-02
[ci] Re-enable JSC

sbc100:


sbc100:
Landing TBR since the jsc step passed

PR #23279, opened 2025-01-02
Remove dummy emscripten_set_offscreencanvas_size. NFC

sbc100:
This is an internal-only API that is not directly callable by user code.

The only place that this function is called is from `system/lib/html5/offscreencanvas.c` as a result of `_emscripten_set_offscreencanvas_size_on_thread`.

The only call to `_emscripten_set_offscreencanvas_size_on_thread` is from `setOffscreenCanvasSizeOnTargetThread` which is only ever included when `OFFSCREENCANVAS_SUPPORT` is enabled.


tools/maint/gen_sig_info.py
```diff
@@ -172,6 +172,8 @@ def ignore_symbol(s, cxx):
     return True
   if s.startswith('emscripten_gl') or s.startswith('emscripten_alc'):
     return True
+  if (s.endswith('_calling_thread') or s.endswith('_main_thread')) and not s.startswith('emscripten_asm_const'):
```
kripken, line 175: Maybe add a comment here? I can't tell what this change is doing.

sbc100: Done

tools/maint/gen_sig_info.py
```diff
@@ -166,12 +166,20 @@
 
 
 def ignore_symbol(s, cxx):
+  # We need to ignore certain symbols here. Specifically any symbol that doesn't
+  # appear in C/C++ header file.
```
kripken, line 170:
```suggestion
  # appear in a C/C++ header file (such symbols cannot be called by JS).
```
Or is there another reason for why we need to ignore these?

sbc100: No, its more like "Such symbols cannot be part of the auto-sig-generation process at all since we depend on import these functions with their header-declared signature."

sbc100: Its just the way gen_sig_info works:  It generated a bunch of references to header-declared symbols.  If they symbol are not declared the file cannot be compiled.

kripken: Oh, right, thanks.

kripken (approved):

tools/maint/gen_sig_info.py
sbc100: I've updated this PR such that this change is no longer needed, so I split the doc changes out into a separate PR.

kripken (approved):

PR #23292, opened 2025-01-03
Warn on the incompatibility between source maps and SINGLE_FILE

### LINKED ISSUE #23257: -sSINGLE_FILE and -fsanitize=address,undefined

```
$ emcc -v
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.74-git (1092ec30a3fb1d46b
1782ff1b4db5094d3d06ae5)
clang version 20.0.0git
Thread model: posix
InstalledDir: C:/msys64/clang64/opt/emscripten-llvm/bin
```

```
$ emcc -gsource-map -g3 -Og -sALLOW_MEMORY_GROWTH -sINITIAL_MEMORY=655360000 -sSINGLE_FILE -fsanitize=address,undefined $$(find . -name '*.c' ! -name batch.c ! -name simple.c) --use-port=sdl2 -o index.html --shell-file=shell.html
```

this compiles and runs fine!!

THE PROBLEM: sanitizer works fine with `-sSINGLE_FILE`, but when it catches an error, it displays offsets in hex :(
I "solved" by removing `-sSINGLE_FILE`, then it shows line in source code correctly when it catches an error!!

some solutions:

1. add a warning that `-sSINGLE_FILE` and `-fsanitize=` should not go together
2. make it an error that makes build fail, rather than warning
3. make them work™ together

a warning is enough for me <3

kripken:
Source maps require a source map file, which SINGLE_FILE disallows. We
disable source maps in that case, but did not show a warning before
this PR.

Fixes #23257

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -9541,6 +9541,11 @@ int main() {
     self.assertExists('hello_world.js')
     self.assertFileContents('hello_world.wasm', 'not wasm')
 
+  def test_single_file_disables_source_map(self):
+    cmd = [EMCC, test_file('hello_world.c'), '-sSINGLE_FILE', '-gsource-map']
+    stderr = self.run_process(cmd, stderr=PIPE).stderr
+    self.assertContained('warning: SINGLE_FILE disables source map support', stderr)
+
   def test_wasm2js_no_clobber_wasm(self):
     create_file('hello_world.wasm', 'not wasm')
     self.do_runf('hello_world.c', emcc_args=['-sWASM=0'])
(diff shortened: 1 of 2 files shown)


sbc100 (approved):

sbc100:
I wonder if this should be an error?

kripken:
I think a warning is enough for this corner case. It might be more annoying for users that want to do some quick testing with SINGLE_FILE to have to flip other flags to avoid an error (even if they don't need those flags in their test).

PR #23297, opened 2025-01-04
Remove `noProxy` options from `-sPROXY_TO_WORKER`

sbc100:
The `noProxy` options was added back in 2014 in c2295c7b3, but doesn't have any test coverage.  This means there is is good chance its already broken.  What is more it a fair mount of complexity since it adds an additional way in which certain JS files get pre-processed and included.

I propose that we simply remove it, but and alternative would be to fine a way to test it.  I'm not quite sure who the end user of this feature might be though.


kripken:
> doesn't have any test coverage

Was it not tested here?

https://github.com/emscripten-core/emscripten/commit/c2295c7b3d8c73e12ca3e2c4f0801446af384791#diff-644556461d8f2f2bdd70b1f0d5455342db92e9535b5bca4ca5223224bd1a21f3R751

sbc100:
> > doesn't have any test coverage
> 
> Was it not tested here?
> 
> [c2295c7#diff-644556461d8f2f2bdd70b1f0d5455342db92e9535b5bca4ca5223224bd1a21f3R751](https://github.com/emscripten-core/emscripten/commit/c2295c7b3d8c73e12ca3e2c4f0801446af384791#diff-644556461d8f2f2bdd70b1f0d5455342db92e9535b5bca4ca5223224bd1a21f3R751)

Ah, you are correct.  I was kind of hoping it wasn't tested and that it had been broken for a while :)

Do you want to keep it around?

sbc100:
But how did the browser tests pass then ... ?

kripken:
This option is probably not that important, given it is proxy-to-worker and proxy-to-pthread is much more useful. And IIRC one can just build without the option to run on the main thread, which is all the option did. So I think I'm ok with removing it. Maybe worth asking on the mailing list first?

sbc100:
Oh I see the test was removed in #13244.    I can't really see why .. maybe because it was marked as `@no_wasm_backend` .. but still seems odd that I just removed it like that.

kripken:
Hopefully we can remove it. lgtm if no one complains on the mailing list.

sbc100:
OK, I think we can go ahead and land this since nobody has responded.   Could you LGTM?

kripken (approved):

PR #23303, opened 2025-01-05
Avoid parsing `-o` filename except when linking. NFC

sbc100:
This allows the `-o` flag that to be passed directly through to clang like all the other arguments when just compiling.


sbc100:
Converted to draft for now

PR #23305, opened 2025-01-06
Add a checkOpExists helper for common code pattern. NFC

hoodmane:
We frequently check if a node op is present, if not raise a specific error, if it is present call it. This factors out this common pattern.


src/library_fs.js
```diff
@@ -948,13 +952,7 @@ FS.staticInit();
     stat(path, dontFollow) {
       var lookup = FS.lookupPath(path, { follow: !dontFollow });
       var node = lookup.node;
-      if (!node) {
-        throw new FS.ErrnoError({{{ cDefs.ENOENT }}});
-      }
```
hoodmane, line 953: I don't think `lookupPath` returns an `undefined` node ever, it would throw ENOENT itself so this check is not needed.

src/library_fs.js
```diff
@@ -1016,9 +1008,6 @@ FS.staticInit();
       } else {
         node = path;
       }
-      if (!node.node_ops.setattr) {
-        throw new FS.ErrnoError({{{ cDefs.EPERM }}});
-      }
```
hoodmane, line 1021: This changes which error we throw if there are multiple problems.

src/library_fs.js
```diff
@@ -1044,7 +1033,7 @@ FS.staticInit();
     utime(path, atime, mtime) {
       var lookup = FS.lookupPath(path, { follow: true });
       var node = lookup.node;
-      node.node_ops.setattr(node, {
```
hoodmane, line 1047: Previously there was no error handling for `setattr` being missing, so if we called `utime` on a read only node it would crash.

hoodmane:
I'm surprised that this increased code size by a few bytes...

src/library_fs.js
```diff
@@ -895,10 +902,7 @@ FS.staticInit();
     readdir(path) {
       var lookup = FS.lookupPath(path, { follow: true });
       var node = lookup.node;
-      if (!node.node_ops.readdir) {
-        throw new FS.ErrnoError({{{ cDefs.ENOTDIR }}});
-      }
-      return node.node_ops.readdir(node);
+      return FS.doNodeOp(node, "readdir", {{{ cDefs.ENOTDIR }}});
```
sbc100, line 905: When I first say this change I assumed `Node` here meant  Node JS.. but I just the term just very overloaded.

I worry a little that using strings here will cause closure to break, or at least fail to minify these attribute names.

How about this `return FS.doNodeOp(node.node_ops.readdir, {{{ cDefs.ENOTDIR }}});` .. thus avoid the string lookup and passing instead the result of the lookup?

sbc100: Or maybe we can leave the call at the callsite and just make check function that doesn't need to take the `...` operator (which I believe generates JS garbage (i.e. a new array)).

```
FS.checkExists(node.readdir, {{{ cDefs.ENOTDIR }}});
```

hoodmane: Well then we'd have to pass both `node` and `node.node_ops.readdir`.

sbc100: Still, same number of args that you have today, but avoids the extra strings + lambdas I think?

hoodmane: Okay, I switched to checkExists as you suggested.

sbc100 (commented):
Nice!  That looks much better.  Can you rename to to something less overloaded though?  checkExists sounds like it will check for file existence, no?

hoodmane:
`checkOpExists`?

src/library_fs.js
```diff
@@ -895,10 +901,8 @@ FS.staticInit();
     readdir(path) {
       var lookup = FS.lookupPath(path, { follow: true });
       var node = lookup.node;
-      if (!node.node_ops.readdir) {
-        throw new FS.ErrnoError({{{ cDefs.ENOTDIR }}});
-      }
-      return node.node_ops.readdir(node);
+      var readdir = FS.checkExists(node.node_ops.readdir, {{{ cDefs.ENOTDIR }}});
```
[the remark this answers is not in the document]
hoodmane: I like `checkOpExists` the best.

sbc100 (commented):
lgtm % name bikeshedding.  Its kind of shame that FS is a public API.. there is no easy way for us to stop external folks calling this function.

sbc100 (approved):

hoodmane:
Indeed.

sbc100 (commented):
An alternative to this would be to force all FS impls to implement all operations, and have them report correct errors when they don't, but maybe thats a bigger change..

hoodmane:
I find it quite convenient that they default to ENOSYS personally.

sbc100 (approved):

sbc100 (approved):

sbc100:
Oh, lets try to remember to prefix the PR titles for these changes with `[FS]`?

PR #23307, opened 2025-01-06
Raise ENOSYS if AT_SYMLINK_NOFOLLOW is used with chmod or chown in nodefs

hoodmane:
It cannot support this since node doesn't support it. So we should tell the truth that we can't do it instead of ignoring it.


hoodmane:
I realized this has no interaction with #23058 so I moved it off of that and now the diff here is quite small @sbc100.

sbc100:
Looks like this was missing the code size updates.  I'll land an update now.

PR #23309, opened 2025-01-06
Micro-optimize randomFill. NFC

sbc100:
Only use the node fallback code when its actually needed, otherwise assume that `crypto.getRandomValues` is available out-of-the-box (which it is on node v19 and above.


src/library_wasi.js
```diff
@@ -561,28 +561,24 @@ var WasiLibrary = {
...
-      return (view) => crypto.getRandomValues(view);
-#endif
-    }
-
-#if ENVIRONMENT_MAY_BE_NODE
+#if ENVIRONMENT_MAY_BE_NODE && MIN_NODE_VERSION < 19
     if (ENVIRONMENT_IS_NODE) {
-      return (view) => require('crypto').randomFillSync(view);
+      // Node v19 and above has globalThis.crypto already available
+      crypto = require('crypto');
```
kripken, line 567: Is this modifying the global scope intentionally?

sbc100: Yes, the code below that uses `crypto.getRandomValues` expects it to exist.

This is just the default behavior of modern node versions anyway.  I could add a check that it doesn't already exist perhaps?

kripken: No, looks good, I just wanted to be sure I was reading it right.

kripken (approved):

sbc100:
Interrestingly enough it seems that the builtin `crypto` that comes with node v19 and above is not the same is that one you get when you `require('crypto)'`.  Specifically `crypto.randomFillSync` only exists in the later!


```
var mycrypto = require('node:crypto');

console.log(crypto.randomFillSync);  // undefined
console.log(mycrypto.randomFillSync);  // defined!
```

src/library_wasi.js
```diff
@@ -561,28 +561,20 @@ var WasiLibrary = {
...
-#else
-      return (view) => crypto.getRandomValues(view);
-#endif
-    }
-
-#if ENVIRONMENT_MAY_BE_NODE
+#if ENVIRONMENT_MAY_BE_NODE && MIN_NODE_VERSION < 190000
     if (ENVIRONMENT_IS_NODE) {
-      return (view) => require('crypto').randomFillSync(view);
+      // Node v19 and above has globalThis.getRandomValues builtin
```
kripken, line 566: This comment talks about `getRandomValues` but the code on the line below it is `randomFillSync`?

kripken, line 566: Also, this code block is to support node < 19, and this comment says 19+ - so I am not sure how earlier node are supported?

kripken (changes requested):

src/library_wasi.js
sbc100: The `crypto.getRandomValues` call is below.   It gets used when we don't need to support node < 19

sbc100: I updated the comment.

a.out.js
```diff
@@ -0,0 +1,1649 @@
+// include: shell.js
```
kripken, line 1: Unintended, I assume.

tests.txt
```diff
@@ -0,0 +1,13 @@
+            other
```
kripken, line 1: ditto

kripken (approved):

kripken (changes requested):
Oh, I see there is automerge, so removing lgtm. but lgtm with the unneeded files removed

kripken (approved):

PR #23310, opened 2025-01-06
Correctly handle calls setting timestamp to 0

hoodmane:
If someone calls `utimes("file", 0, 0)` the previous code I wrote would not adjust the time stamp.


sbc100 (approved):

sbc100:
Test?

hoodmane:
I added it to `test_futimens` which of course detected more problems to fix.

src/library_nodefs.js
```diff
@@ -166,9 +166,14 @@ addToLibrary({
...
-            var atime = attr.atime && new Date(attr.atime);
-            var mtime = attr.mtime && new Date(attr.mtime);
+          if (typeof (attr.atime ?? attr.mtime) === "number") {
+            // Unfortunately, we have to stat the current value if we don't want
+            // to change it. On top of that, since the times don't round trip
+            // this will only keep the value nearly unchanged not exactly
+            // unchanged. See:
+            // https://github.com/nodejs/node/issues/56492
+            var atime = new Date(attr.atime ?? fs.lstatSync(NODEFS.realPath(node)).atime);
+            var mtime = new Date(attr.mtime ?? fs.lstatSync(NODEFS.realPath(node)).mtime);
```
sbc100, line 176: Can we avoid the duplicate call to lstatSync here?

src/library_workerfs.js
```diff
@@ -91,7 +91,7 @@ addToLibrary({
       },
       setattr(node, attr) {
         for (const key of ["mode", "atime", "mtime", "ctime"]) {
-          if (attr[key]) {
+          if (typeof attr[key] === 'number') {
```
sbc100, line 94: How about `!= 'undefined'` here and elsewhere ?

hoodmane: typeof null is object though.

src/library_nodefs.js
hoodmane: We could do:
```js
var atime = attr.atime;
var mtime = attr.mtime;
if ((typeof atime !== "number") || (typeof mtime !== "number")) { 
   var st = fs.lstatSync(NODEFS.realPath(node));
   atime ??= st.atime;
   mtime ??= st.mtime;
}
```
It's unfortunate that we don't have an operator whose relationship to `&&` is the same as the relationship of `??` to `||`.

hoodmane: Or more simply, an easy "is nullish" test.

src/library_workerfs.js
hoodmane: Ah but we could do `attr[key] != null` because `0 != null` is `true` and `undefined != null` is `false`.

src/library_nodefs.js
sbc100: It both attr.atime and attr.mtime are non-nullish (the most common case IIUC) then it will call fs.lstatSync twice won't it?

hoodmane: I guess `if (atime == null || mtime == null)` would do. That's only 17 characters longer than what I have now.

hoodmane: If both `attr.atime` and `attr.mtime` are not nullish it won't call lstat at all. It only falls back to calling lstat when the first part is nullish.

sbc100: Ah I see.

Well in that case how about just not repeating it:

```
  // Unfortunately, we have to stat the current value if we don't want
  // to change it. On top of that, since the times don't round trip
  // this will only keep the value nearly unchanged not exactly
  // unchanged. See:
  // https://github.com/nodejs/node/issues/56492
  var currentStat = fs.lstatSync(NODEFS.realPath(node));
  var atime = new Date(attr.atime ?? currentStat.atime);
  var mtime = new Date(attr.mtime ?? currentStat.mtime);
```

sbc100: If `null-ish` is causing problems then perhaps we could instead define some kind of sentinal here such as `UTIME_NONE` which would avoid all the typeof checking?

hoodmane: Well then we would always do at least one stat. We know we won't do two stats but we might do zero or one. Maybe:
```
  var stat = () => fs.lstatSync(NODEFS.realPath(node));
  var atime = new Date(attr.atime ?? stat().atime);
  var mtime = new Date(attr.mtime ?? stat().mtime);
```

hoodmane: That is 4 characters shorter than what I currently have.

sbc100 (approved):

PR #23316, opened 2025-01-06
[nodefs] Remove compatibility code for ancient node versions

hoodmane:


src/library_nodefs.js
```diff
@@ -123,15 +119,6 @@ addToLibrary({
         var stat;
         NODEFS.tryFSOperation(() => stat = fs.lstatSync(path));
         if (NODEFS.isWindows) {
-          // node.js v0.10.20 doesn't report blksize and blocks on Windows. Fake
```
sbc100, line 126: We do still support MIN_NODE_VERSION=101900 (10.19.0) so if we want to drop this I guess we would need to bump that minimum too?

sbc100: Oh wait, thats is 0.10.20!

sbc100:
Can you mention nodefs in the title somehow?  Maybe prefix with `[nodefs]`?

src/library_nodefs.js
sbc100: Can you add some stat tests to `test-node-compat` in `.circleci/config.yml`?

src/library_nodefs.js
```diff
@@ -19,11 +19,7 @@ addToLibrary({
     isWindows: false,
     staticInit() {
       NODEFS.isWindows = !!process.platform.match(/^win/);
-      var flags = process.binding("constants");
-      // Node.js 4 compatibility: it has no namespaces for constants
-      if (flags["fs"]) {
-        flags = flags["fs"];
-      }
+      var flags = process.binding("constants")["fs"];
```
sbc100, line 22: I was going to suggest single quotes here but it looks like this file users double-quote a lot, so maybe we leave as is.

.circleci/config.yml
```diff
@@ -780,6 +780,8 @@ jobs:
             other.test_full_js_library*
             core2.test_hello_world
             core2.test_fs_write_rawfs"
+            core2.*nodefs*
+            core2.*rawfs*
```
sbc100, line 784: Thats quite a few tests but maybe thats fine.  Can you remove  core2.test_fs_write_rawfs above?    

And you need to move the closing quote the last line.

sbc100:
It looks like not all of those tests are actually runnable on the oldest version of node that we support (v10.19.0).  Can you perhaps revert the `.circleci` change and just at one extra test there rather than adding all of them?

We can work on adding all of them in followup PRs perhaps?

sbc100:
For the closure compiler errors I think you can maybe add `blksize` to `src/closure-externs/node-externs.js`?

hoodmane:
I also wonder how long we want to keep supporting EOL node versions. Node v10 has been EOL for about 4 years. But we'd probably need to bump to version 16 to get a big help. Node 14 has only been EOL since April 2023 which might be a bit recent to drop.

sbc100:
> I also wonder how long we want to keep supporting EOL node versions. Node v10 has been EOL for about 4 years. But we'd probably need to bump to version 16 to get a big help. Node 14 has only been EOL since April 2023 which might be a bit recent to drop.

That is certainly valid discussion and we should consider bumping the min version.  Perhaps you could open an issue regarding that?

sbc100 (approved):

PR #23320, opened 2025-01-07
[tsgen] Support JSPI when emitting TypeScript definitions.

### LINKED ISSUE #23272: Build fails when both -sJSPI and --emit-tsd are active
```bash
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.74 (1092ec30a3fb1d46b1782ff1b4db5094d3d06ae5)
clang version 20.0.0git (https:/github.com/llvm/llvm-project 322eb1a92e6d4266184060346616fa0dbe39e731)
Thread model: posix
```

**Issue:**
When enabling the **JSPI** flag and attempting to emit **Typescript definitions** (via --emit-tsd), the build fails. The error indicates that JSPI is not supported in the current environment. It seems that it tries to assert in Node context that `WebAssembly.promising` exists.

I suspect that the issue lies in `tsgen.js`, which appears to validate the generated codes but fails because Node.js does not support JSPI. However, the generated code works correctly in a browser environment when the JSPI flag is enabled.

Demo to reproduce: https://github.com/mroohian/emscripten-jspi-demo

```bash
# This command is executed by CMake
em++ -gsource-map \
     -lembind \
     --emit-tsd test1.d.ts \   # It works without this line
     -s EXPORT_NAME=Test1Module \
     -s JSPI \
     -s STRICT \
     -s USE_ES6_IMPORT_META \
     -s MODULARIZE \
     -s EXPORT_ES6=1 \
     --no-entry \
     test1.cpp.o \
     -o test1.js
```

```shell
RuntimeError: Aborted(Assertion failed: JSPI not supported by current environment. Perhaps it needs to be enabled via flags?)
    at abort (C:\Users\User\AppData\Local\Temp\emscripten_temp_52rs7em6\tsgen.js:498:11)
    at assert (C:\Users\User\AppData\Local\Temp\emscripten_temp_52rs7em6\tsgen.js:231:5)
    at Object.instrumentWasmImports (C:\Users\User\AppData\Local\Temp\emscripten_temp_52rs7em6\tsgen.js:4548:9)
    at getWasmImports (C:\Users\User\AppData\Local\Temp\emscripten_temp_52rs7em6\tsgen.js:639:12)
    at createWasm (C:\Users\User\AppData\Local\Temp\emscripten_temp_52rs7em6\tsgen.js:689:14)
    at Object.<anonymous> (C:\Users\User\AppData\Local\Temp\emscripten_temp_52rs7em6\tsgen.js:4824:1)
    at Module._compile (node:internal/modules/cjs/loader:1469:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1548:10)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module._load (node:internal/modules/cjs/loader:1104:12)

Node.js v20.18.0
em++: error: 'C:/Users/User/Workspace/third-party/emsdk/node/20.18.0_64bit/bin/node.exe C:\Users\User\AppData\Local\Temp\emscripten_temp_52rs7em6\tsgen.js' failed (returned 1)
ninja: build stopped: subcommand failed.
```

brendandahl:
- Adds a Promise<T> wrapper around the return type for async embind functions.
- Disables JSPI while generating the definitions, since it's not needed.

Fixes #23272

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -2001,7 +2001,7 @@ def run_embind_gen(wasm_target, js_syms, extra_settings, linker_inputs):
   setup_environment_settings()
   # Use a separate Wasm file so the JS does not need to be modified after emscripten.emscript.
   settings.SINGLE_FILE = False
-  if settings.JSPI:
+  if settings.ASYNCIFY == 2:
     # JSPI is not needed to generate the definitions.
     # TODO: when the emsdk node version supports JSPI, it probably makes more sense
     # to enable it in node than disabling the setting here.


tools/link.py
```diff
@@ -2001,6 +2001,11 @@ def run_embind_gen(wasm_target, js_syms, extra_settings, linker_inputs):
   setup_environment_settings()
   # Use a separate Wasm file so the JS does not need to be modified after emscripten.emscript.
   settings.SINGLE_FILE = False
+  if settings.JSPI:
+    # JSPI is not needed to generate the definitions.
+    # TODO: when the emsdk node version supports JSPI, it probably makes more sense
+    # to enable it in node than disabling the setting here.
+    settings.ASYNCIFY = 0
```
sbc100, line 2008: Why not `settings.JSPI = 0` here?

brendandahl: JSPI really is just `ASYNCIFY = 2`. I guess it's probably more clear if only use the asyncify setting, e.g. `if settings.ASYNCIFY == 2:`.

sbc100: sgtm

sbc100 (approved):

PR #23330, opened 2025-01-07
Use debug version of malloc when ASSERTIONS=1. NFC

sbc100:
Previously we were only using it with `ASSERTIONS=2`.  

The debug version dlmalloc has many very useful checks, including checking for double-free.  This is something that is done by glibc malloc even in release builds.  It seems reasonable that we should do this basic checking in our debug builds at least.

Required fixing two genuine bugs: One double-free in test code and one invalid free in embind.

Fixes: #23360


sbc100:
Changelog updated.

kripken (approved):
Great!

PR #23332, opened 2025-01-08
Delay processing of --js-library flags. NFC

sbc100:
By delaying the processing of `--js-library` arguments we avoid having to jump through hoops and store them alongside their ordinal number for half the time.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -1240,7 +1240,7 @@ def phase_linker_setup(options, state):  # noqa: C901, PLR0912, PLR0915
       state.append_link_flag('--no-whole-archive')
     settings.FILESYSTEM = 1
     settings.SYSCALLS_REQUIRE_FILESYSTEM = 0
-    settings.JS_LIBRARIES.append((0, 'library_wasmfs.js'))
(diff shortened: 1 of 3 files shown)


tools/link.py
```diff
@@ -1240,7 +1240,7 @@ def phase_linker_setup(options, state):  # noqa: C901, PLR0912, PLR0915
       state.append_link_flag('--no-whole-archive')
     settings.FILESYSTEM = 1
     settings.SYSCALLS_REQUIRE_FILESYSTEM = 0
-    settings.JS_LIBRARIES.append((0, 'library_wasmfs.js'))
```
kripken, line 1243: Did this 0 not ensure that system libs were before user ones?

sbc100: Indeed, and that is still that case.   All these `JS_LIBRARIES.append` happen during `process_libraries` which runs before `process_libraries`.

kripken: First `process libraries` should be `phase_linker_setup` in the comment? If so then yes, that looks right, thanks.

sbc100: Yes sorry, fixed commend

PR #23336, opened 2025-01-08
[link.py] Be explicit about if/when we expand -l flag. NFC

sbc100:
In most cases we prefer to leave it up to the linker to do this.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -36,7 +36,7 @@ from . import extract_metadata
 from .utils import read_file, write_file, delete_file
 from .utils import removeprefix, exit_with_error
 from .shared import in_temp, safe_copy, do_replace, OFormat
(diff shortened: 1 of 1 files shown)


tools/link.py
```diff
@@ -2835,16 +2834,22 @@ def process_libraries(state):
     if js_libs is not None:
       continue
 
-    path = None
-    for suff in suffixes:
-      name = 'lib' + lib + suff
-      path = find_library(name, state.lib_dirs)
-      if path:
-        break
+    if not settings.RELOCATABLE:
```
kripken, line 2837: Why was there no RELOCATABLE check before? Was it not needed?

kripken: (maybe related to the test failures?)

sbc100: Indeed, we were needlessly expanding all `-l` flags even though wasm-ld does this.

kripken (approved):

Issue #23343, opened 2025-01-08
emmalloc_trim broken since #13442?


sbc100:
It looks like `emmalloc_trim` has been broken since #13442.  Prior to that change the result of sbrk would reduce in `test_emmalloc_trim` test but after that change the expecations were updated to reflect an sbrk that never goes down, and indeed if I run that test with `-sASSERTIONS=2` I see that this assertion fires every time:

https://github.com/emscripten-core/emscripten/blob/6b81620f82549be042b0d7d5b65353433deac393/system/lib/emmalloc.c#L1277

Basically it looks like sbrk no longer accepts negative values.. which I think was an indended part of #13442.  So maybe emmalloc_trim cannot be implement using sbrk along?  We would need some kind of new internal function such as sbrk_shrink to make it work?

kripken:
Where do you see that sbrk does not accept negative values? Looking through the code I failed to find anything obvious.

sbc100:
You can see from the expected output of the test.. sbrk could be reduced before but not after.  I think its because the increment that sbrk was interpreted as signed before and unsigned after.    Whatever the reason the side effect can clearly be seen in the test expectations, or by enabling `-sASSERTIONS=2`.

sbc100:
```
-    intptr_t old_brk = *sbrk_ptr;
+    uintptr_t old_brk = *sbrk_ptr;
 #endif
-    intptr_t new_brk = old_brk + increment;
+    uintptr_t new_brk = old_brk + increment;
```

sbc100:
Actually real problem is probably `uintptr_t increment = (uintptr_t)increment_;` which was added

kripken:
(I commented in #13442 before seeing this, but as I said there, does wrapping mean that the cast to unsigned is ok?)

sbc100:
@juj I'd be happy to see `emmalloc_trim` re-enabled if you can figure out a way to make it work again.

PR #13442, opened 2021-02-07
Fix different memory allocation issues

juj:
1. In library.js, my earlier PR gated a >>> 0 instruction to CAN_ADDRESS_2GB builds only. That was troublesome, since if one does malloc(-1), it would no longer be caught (malloc(0xFFFFFFFFull) can never succeed). Restored the >>> 0.
2. Fix STACK_OVERFLOW_CHECK=2 mode to work with MINIMAL_RUNTIME. The call to ___set_stack_limits() was misplaced in MINIMAL_RUNTIME postamble.
3. Fix issue with sbrk() not being able to handle signed 32-bit integer overflow in sbrk limit growth, leading to erroneously reporting succeeding massive 2GB heap grow operations.
   This causes a breaking change to emscripten/heap.h emscripten_get_sbrk_ptr() signature, which will now return a uintptr_t instead of intptr_t. Not worried much since it will lead to clear build error, and there are extremely few users of that function.
4. Fix issues with passing ridiculously large (or small negative) alloc sizes to emmalloc functions (>0xFFFFFFC7u). Add assertions to validate memory overflows in emmalloc.
5. Fix a crash with emmalloc when attempting to malloc() memory when there are absolutely zero bytes available in the heap. (not a single free memory region left)
6. Add a test.
7. Misc comment updates.

PR #23344, opened 2025-01-08
Disable emmalloc_trim

sbc100:
I believe this function has not been working correctly since #13442 when sbrk stopped accepting negative values.

Maybe we can find a way to bring this functionality back but disabling it for now seems safest.

See: #23343

### DIFF
--- a/system/lib/emmalloc.c
+++ b/system/lib/emmalloc.c
@@ -1227,6 +1227,7 @@ struct mallinfo emmalloc_mallinfo() {
 }
 EMMALLOC_ALIAS(mallinfo, emmalloc_mallinfo);
 
+#if 0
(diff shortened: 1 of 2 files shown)


sbc100:
If you look at the changes to test_emmalloc_trim.out which were part of #13442 you can see the sbrk() value stops changing.  i.e. it could be reduces prior to #13442 but can no longer be reduces afterward.

sbc100:
Since its been broken for ~4 years I guess nobody is using it ?

kripken (approved):
Likely unused, yeah.

sbc100:
This is why it would have been good to have https://github.com/emscripten-core/emscripten/pull/23330 all along.    If we had that landed 4 years ago this breakage would have happened since that assert would have fired.

Issue #23338, opened 2025-01-08
Allow user JS libraries via `-Lbar` and `-lfoo.js` 


sbc100:
We already support `-lfoo.js` for system libraries (e.g. `-luuid.js` will include `src/library_uuid.js`).

For parity we should extend this to user libraries.  We could even recommend this and deprecate the `--js-library` flag.

One minor nit here is that our system libraries are called `library_xx.js` rather then `libxx.js`.. should we rename them for consistency?

kripken:
Good point that this could be more consistent. I think the benefit of a change here would be pretty minor though, to be honest. If we are certain we can avoid breaking backwards compatibility that might be fine though, but if not, I'm not sure about it.

sbc100:
Turns out making this work is pretty trivial: #23345

sbc100:
As long as we keep the `--js-library` flag around there is no backwards compat issue here.   I'm not in any hurry to remove it.

PR #23345, opened 2025-01-09
Allow user JS libraries to be specified via `-lfoo.js`

sbc100:
We already allowed this for system libraries but it is now also possible for user libraries.

Fixes: #23338


test/test_other.py
```diff
@@ -4886,6 +4886,25 @@ def test_jslib_exported_functions(self):
...
+    create_file('main.c', r'''
+      #include <stdio.h>
+      int foo();
+      int main() {
+        printf("%d\n", foo());
+        return 0;
+      }''')
+    self.do_runf('main.c', '42\n', emcc_args=['-L.', '-lfoo.js'])
+
+    # If the library path is not include with `-L` we expect the command to fail
```
kripken, line 4904:
```suggestion
    # If the library path is not included with `-L` we expect the command to fail.
```

kripken (approved):

PR #23348, opened 2025-01-09
Rename JS library files from library_foo.js to libfoo.js

sbc100:
At the same time, move all the library file into a separate `src/lib`       
directory.  This helps distinguish library files from other files        
in the `lib` directory, which is currently very overcrowded/overloaded.  
                                                                         
These files were also includ-able via `-lfoo` and we previously had to   
use a special case for system libraries where `library_foo.js` could be  
used in addition to more standed `libfoo.js`.  This change removes to       
special case.                                                            
                                                                         
This only file that was not renamed using the simple pattern was         
`library.py` itself which becomes `libcore.py`.  I also considered          
`libemscripten.js`.                                                      
                                                                         
To some this may seem like needless churn, but for years now the `src`   
directory (IMHO) has been asking for cleanup.  I think having files         
logically separated into directories helps a lot with understanding the  
project structure for both newcomers and old timers alike.               
                                                                         
As a followup, I am also considering moving all the runtime files into   
`src/include` (since these are all files that get `#included` into the   
generated code directly.  That would leave the `src` directory           
containing only the JS compiler code itself which I think is another        
important distinction.


sbc100:
Marking as draft for now.

src/lib/libstrings.js
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  */
 
-#include "arrayUtils.js"
+#include "../arrayUtils.js"
```
kripken, line 7: This might affect user code, I worry? They would need to include one of our libraries from one of theirs, but still. We could avoid that risk by making `#include` add the extra directory layer?

sbc100: OK, I think I can make this work by using and include directory that includes `src` and `.`

sbc100: I update the include search mechanism to handle both cases.. I'm going to split that into its own PR I think.

sbc100:
Much cleaner PR now that #23449 landed

kripken (approved):

sbc100:
Added a ChangeLog entry

kripken:
lgtm

PR #23349, opened 2025-01-09
Simplify file finding/reading files in JS compiler. NFC

sbc100:
- Use import.meta.dirname
- Remove duplicate copy of `find` function.
- Remove `export` of find function

### DIFF
--- a/src/compiler.mjs
+++ b/src/compiler.mjs
@@ -7,28 +7,11 @@
 
 // LLVM => JavaScript compiler, main entry point
 
-import * as fs from 'node:fs';
-import * as path from 'node:path';
-import * as url from 'node:url';
-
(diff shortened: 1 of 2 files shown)


src/compiler.mjs
```diff
@@ -9,26 +9,12 @@
...
-    }
-  }
-  return filename;
-}
-
 // Load default settings
-loadSettingsFile(find('settings.js'));
-loadSettingsFile(find('settings_internal.js'));
+loadSettingsFile('settings.js');
+loadSettingsFile('settings_internal.js');
```
sbc100, line 17: loadSettingsFile already calls `find` internally so there was no need for this call.

sbc100:
Converted to draft

kripken (approved):

PR #23353, opened 2025-01-09
Remove unused isPowerOfTwo function from JS compiler. NFC

sbc100:
Also, add `utf8` to read function to avoid the extra `toString()` call (which itself defaults to `utf8`).

### DIFF
--- a/src/utility.mjs
+++ b/src/utility.mjs
@@ -231,13 +231,9 @@ export function isDecorator(ident) {
   return suffixes.some((suffix) => ident.endsWith(suffix));
 }
 
-export function isPowerOfTwo(x) {
-  return x > 0 && (x & (x - 1)) == 0;
-}
-
 export function read(filename) {
   const absolute = find(filename);
-  return fs.readFileSync(absolute).toString();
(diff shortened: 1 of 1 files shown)


src/utility.mjs
```diff
@@ -231,13 +231,9 @@ export function isDecorator(ident) {
   return suffixes.some((suffix) => ident.endsWith(suffix));
 }
 
-export function isPowerOfTwo(x) {
-  return x > 0 && (x & (x - 1)) == 0;
-}
```
kripken, line 236: This may be used by user JS libraries - is it worth removing? If so, let's at least mention in the changelog?

sbc100: No, the way these "exports" work is only to other ES6 modules that make up the compiler.

The JS libraries don't have visibility of these symbols, by design.  JS libraries can only see symbols that are explictly added using `addToCompileTimeContext`.    This function (isPowerOfTwo) is not one of those.

kripken: Oh, nice, I wasn't aware that ES6 modules limited things what way.

kripken (approved):

PR #23357, opened 2025-01-09
Add the ASYNCIFY_STACK_SIZE to the MEM_SIZE_SETTINGS

oleg-derevenetz:
Currently it is impossible to use kb/mb/gb for `ASYNCIFY_STACK_SIZE`, although, in theory, it should work.


oleg-derevenetz:
Hi @sbc100

> lgtm! Maybe we should add a test though?

Unfortunately, I'm not familiar with your test infrastructure. Is there any specific place where such a test should be added? I have not yet been able to find a place where these size suffixes would be tested at all.

sbc100:
How about adding `self.set_setting('ASYNCIFY_STACK_SIZE', ['64kb'])` to `test_asyncify_advise` in `test/test_other.py`, perhaps with a comment that this just to test that size can be specified in `kb`?

oleg-derevenetz:
> How about adding `self.set_setting('ASYNCIFY_STACK_SIZE', ['64kb'])` to `test_asyncify_advise` in `test/test_other.py`, perhaps with a comment that this just to test that size can be specified in `kb`?

Sure, thank you!

tools/settings.py
```diff
@@ -23,7 +23,8 @@
     'MEMORY_GROWTH_GEOMETRIC_CAP',
     'GL_MAX_TEMP_BUFFER_SIZE',
     'MAXIMUM_MEMORY',
-    'DEFAULT_PTHREAD_STACK_SIZE'
+    'DEFAULT_PTHREAD_STACK_SIZE',
+    'ASYNCIFY_STACK_SIZE'
```
sbc100, line 27: Can you add a trailing comma here so future additions can just be single line change?

oleg-derevenetz: Sure, done.

PR #23366, opened 2025-01-10
[FS] lookupPath should throw an error on an empty string

hoodmane:
It's inconsistent that lookupPath returns `node:null` when the path is an empty string. Normally it either raises `ENOENT` or `node` is truthy. This makes empty string into an edge case that every caller has to handle. `stat` has special handling for the empty string case and correctly returns ENOENT on empty string, but chmod, truncate, and mknod all crash. utime and some other functions mishandle the empty string in an unrelated way.


src/library_fs.js
```diff
@@ -172,7 +172,8 @@ FS.staticInit();
     // paths
     //
     lookupPath(path, opts = {}) {
-      if (!path) return { path: '', node: null };
+      if (!path)
+        throw new FS.ErrnoError({{{ cDefs.ENOENT }}});
```
sbc100, line 176: Use curly braces here.

sbc100 (approved):
lgtm % comment

hoodmane:
Well this makes `other.test_unistd_fstatfs` fail, but it was never working for the right reason...

hoodmane:
Opened #23381 to fix the problems in fstatfs turned up by this PR.

PR #23367, opened 2025-01-10
[contrib.glfw3] reverted changes

ypujante:
@sbc100 Hope this can make it before 4.0.0 is released. The latest change I pushed happens to break ImGui due to how `emscripten_webgl_create_context` handles WebGL version context vs what I was doing before. I will open a ticket for Emscripten because what is happening does not make any sense.

So I am reverting this change entirely until I can figure out what is going. Please push it.

Thanks


ypujante:
I pushed a fix

PR #23368, opened 2025-01-10
Always use `size_t` in dlmalloc's mallinfo

sbc100:
For some reason emscripten was overriding `MALLINFO_FIELD_TYPE` to be `int` even though this defaults `size_t` and `size_t` is what linux uses, and `size_t` allows for correct reporting of numbers larger than 2^32.

What is worse, the `MALLINFO_FIELD_TYPE` override was only applied when `DLMALLOC_DEBUG` was defined which meant that `MALLINFO_FIELD_TYPE` varied between `-sASSERTIONS=1` and `-sASSERTIONS=2` builds of dlmalloc!

This means that `-sMEMORY64 + -sASSERTIONS=2` builds of dlmalloc were simply broken WRT `mallinfo` since the public definition of mallinfo disagreed with the dlmalloc-internal version.

I verified that the follow tests fails without this change:  `EMCC_CFLAGS=-sASSERTIONS=2 ./test/runner wasm64.test_mallinfo`

I'm not updating the actual test code here since test coverage will be coming in #23330.

This change was spit out from #23330.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -64,6 +64,8 @@ See docs/process.md for more on how version tagging works.
 - JavaScript libraries can now be specified via `-lfoo.js`.  This works like the
   existing `--js-library` flag but will search the library path (all paths
   specified with `-L`) for `libfoo.js`. (#23338)
+- The `mallinfo` struct members are now defined as `size_t` which makes them
+  compatible with larger memories, and is also how linux defines them. (#23368)
 
 3.1.74 - 12/14/24
 -----------------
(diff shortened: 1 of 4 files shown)


sbc100:
Added ChangeLog entry.

ChangeLog.md
```diff
@@ -64,6 +64,8 @@ See docs/process.md for more on how version tagging works.
 - JavaScript libraries can now be specified via `-lfoo.js`.  This works like the
   existing `--js-library` flag but will search the library path (all paths
   specified with `-L`) for `libfoo.js`. (#23338)
+- The `mallinfo` struct members are now defined as `size_t` which makes them
+  compatible with larger memories is also how linux defines them. (#23368)
```
kripken, line 68:
```suggestion
  compatible with larger memories, and is also how linux defines them. (#23368)
```

kripken (approved):

PR #23369, opened 2025-01-10
[embind] Simplify string.toWireType. NFC

sbc100:
- Simplify the type check
- Restructure the loop to avoid checking stdStringIsUTF8 twice

### DIFF
--- a/src/lib/libembind.js
+++ b/src/lib/libembind.js
@@ -533,7 +533,8 @@ var LibraryEmbind = {
         var length;
         var valueIsOfTypeString = (typeof value == 'string');
 
-        if (!(valueIsOfTypeString || value instanceof Uint8Array || value instanceof Uint8ClampedArray || value instanceof Int8Array)) {
+        // We accept `string` or array views with single byte elements
+        if (!(valueIsOfTypeString || (ArrayBuffer.isView(value) && value.BYTES_PER_ELEMENT == 1))) {
           throwBindingError('Cannot pass non-string to std::string');
         }
         if (stdStringIsUTF8 && valueIsOfTypeString) {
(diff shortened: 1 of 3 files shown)


src/embind/embind.js
```diff
@@ -560,10 +561,10 @@ var LibraryEmbind = {
         var base = _malloc({{{ POINTER_SIZE }}} + length + 1);
         var ptr = base + {{{ POINTER_SIZE }}};
         {{{ makeSetValue('base', '0', 'length', SIZE_TYPE) }}};
-        if (stdStringIsUTF8 && valueIsOfTypeString) {
-          stringToUTF8(value, ptr, length + 1);
-        } else {
-          if (valueIsOfTypeString) {
+        if (valueIsOfTypeString) {
+          if (stdStringIsUTF8) {
+            stringToUTF8(value, ptr);
```
brendandahl, line 566: Last parameter is required.

sbc100: Indeed!  Thats annoying.   Its a shame because often we know we have enough space

src/embind/embind.js
```diff
@@ -560,10 +561,10 @@ var LibraryEmbind = {
         var base = _malloc({{{ POINTER_SIZE }}} + length + 1);
         var ptr = base + {{{ POINTER_SIZE }}};
         {{{ makeSetValue('base', '0', 'length', SIZE_TYPE) }}};
-        if (stdStringIsUTF8 && valueIsOfTypeString) {
-          stringToUTF8(value, ptr, length + 1);
-        } else {
-          if (valueIsOfTypeString) {
+        if (valueIsOfTypeString) {
+          if (stdStringIsUTF8) {
+            stringToUTF8(value, ptr, length);
```
brendandahl, line 566: Looks like tests are still failing, `+ 1` is needed or something else going on?

sbc100: Yup, you are correct.

Strange.. his code doesn't actually require null termination (it uses lenght-pefix instead)... but stringToUTF8 unconditionally does.

src/lib/libembind.js
```diff
@@ -558,10 +559,10 @@ var LibraryEmbind = {
...
             }
-          } else {
-            for (var i = 0; i < length; ++i) {
-              HEAPU8[ptr + i] = value[i];
-            }
+          }
+        } else {
+          for (var i = 0; i < length; ++i) {
+            HEAPU8[ptr + i] = value[i];
           }
```
kleisauke, line 566: This can (probably) be further simplified to:
```suggestion
          HEAPU8.set(value, ptr);
```

sbc100: Nice!

PR #23381, opened 2025-01-13
[FS] Make fstatfs actually work

hoodmane:
Prior to this change, `fstatfs` calls `statfs64` with the `char*` path pointer argument set to zero. Generally byte zero of the memory seems to be null, so when `statfs64` converts this `char*` to a JS string it gets the empty string. Then it calls `FS.statfs` with the empty string as an argument. `FS.statfs` calls `lookupPath` with an empty string argument which returns a `null` node, so `fstatfs` returns the default values for when there is no statfs handler in the files system.

This fixes the problem and adds testing for statfs/fstatfs on nodefs and rawfs. We have to add separate `FS.statfsNode` and `FS.statfsStream` handlers because in some cases `stream.node` is null and in some cases `stream.path` is useless (like for stdout). Also, in noderawfs we need to give paths for the standard streams because there is unfortunately no `fs.fstatfsSync` so we can only access information from the host via a path. This means the noderawfs  test won't be portable to windows.


src/library_syscall.js
```diff
@@ -801,23 +813,17 @@ var SyscallsLibrary = {
...
     return 0;
   },
   __syscall_fstatfs64__deps: ['__syscall_statfs64'],
   __syscall_fstatfs64: (fd, size, buf) => {
+#if ASSERTIONS
+    assert(size === {{{ C_STRUCTS.statfs.__size__ }}});
+#endif
     var stream = SYSCALLS.getStreamFromFD(fd);
-    return ___syscall_statfs64(0, size, buf);
+    SYSCALLS.writeStatFs(buf, FS.statfsStream(stream));
```
sbc100, line 825: Since this is the only callsite for `statfsStream` can you just do `FS.statfsNode(stream.node)` here and completely remove `statfsStream`.

Also I wonder if `statfsNode` could have a better name since Node is overloaded? But I can't think of a better one.

hoodmane: In the noderawfs example, `stream.node` is undefined. So we can't do that if we want to support rawfs. Node rawfs overrides statfsStream to:
```js
    statfsStream(stream) {
      return fs.statfsSync(stream.path);
    },
```

sbc100: I see. Maybe we can add a comment to that effect to the `statfsStream` override?

sbc100 (approved):

PR #23384, opened 2025-01-13
Enable eslint `no-undef` rule in JS compiler. NFC

sbc100:
This required making eslint aware of the settings globals which are used throughout the compiler.

Running this change revealed several legacy parseTools function that simply do not run due to references to undefined symbols (because of this we know that there are no active/current user of these functions).

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,12 +1,22 @@
 import globals from 'globals';
 import js from '@eslint/js';
 import { FlatCompat } from '@eslint/eslintrc';
+import { loadDefaultSettings } from './src/utility.mjs';
 
 const compat = new FlatCompat({
   baseDirectory: import.meta.dirname,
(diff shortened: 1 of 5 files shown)


eslint.config.mjs
```diff
@@ -1,13 +1,23 @@
...
   baseDirectory: import.meta.dirname,
   recommendedConfig: js.configs.recommended,
   allConfig: js.configs.all
 });
 
+
+// Emscripten settings are made available to the compiler as global
+// variables.  Make sure eslint knows about them.
+const settings = loadDefaultSettings();
+const settings_globals = {};
```
kripken, line 16:
```suggestion
const settingsGlobals = {};
```

kripken (approved):

eslint.config.mjs
kripken: Just for consistency with other names.

PR #23393, opened 2025-01-14
[AUDIO_WORKLET] Fix spinlocks for audio worklets (and add tests)

cwoffenden:
Fix for #22962. In brief:

- test renamed to match its new use
- test converted to C-only (like the other tests, plus it wasn't using C++ features)

This is an update to #22995 which had become hard to track. The spinlocks are moved to using `emscripten_get_now()` so they can be used in Audio Worklets, and a set of interactive tests added for the spinlock use cases.

The test can be run with:
```
test/runner interactive.test_audio_worklet_emscripten_locks
```
Or standalone with:
```
emcc -sAUDIO_WORKLET -sWASM_WORKERS -pthread audioworklet_emscripten_locks.c
```

I can change the other `emscripten_performance_now()` calls to `emscripten_get_now()` (as suggested in #22995) in a separate PR.


cwoffenden:
Odd, GH closed this when I renamed the branch to merge into a fork (it's residing in [cw-audio-spinlock-fix](https://github.com/cwoffenden/emscripten/tree/cw-audio-spinlock-fix)).

If there's any interest in this I can re-open another PR.

PR #23404, opened 2025-01-14
[embind] Export embind exports as ESM exports for MODULARIZE=instance.

brendandahl:
Using the embind AOT mode we can generate a list of all the embind bindings and export them as ES module exports.

### DIFF
--- a/src/lib/libembind.js
+++ b/src/lib/libembind.js
@@ -32,7 +32,11 @@ var LibraryEmbind = {
   $PureVirtualError: class extends Error {},
   $GenericWireTypeSize: {{{ 2 * POINTER_SIZE }}},
 #if EMBIND_AOT
-  $InvokerFunctions: '<<< EMBIND_AOT_OUTPUT >>>',
+  $InvokerFunctions: '<<< EMBIND_AOT_INVOKERS >>>',
+#if MODULARIZE == 'instance'
+  $embindUpdateExports: '<<< EMBIND_AOT_UPDATE_EXPORTS >>>',
+  $embindUpdateExports__postset: 'addOnInit(embindUpdateExports);',
+#endif
 #endif
   // If register_type is used, emval will be registered multiple times for
   // different type id's, but only a single type object is needed on the JS side
@@ -2312,5 +2316,9 @@ var LibraryEmbind = {
     });
   },
 };
+
+#if MODULARIZE == 'instance'
+extraLibraryFuncs.push('$embindUpdateExports');
+#endif
 
 addToLibrary(LibraryEmbind);
--- a/src/lib/libembind.js
+++ b/src/lib/libembind.js
@@ -34,10 +34,10 @@ var LibraryEmbind = {
 #if EMBIND_AOT
   $InvokerFunctions: '<<< EMBIND_AOT_INVOKERS >>>',
 #if MODULARIZE == 'instance'
-  // embindUpdateExports is called in the generated code after embind is
+  // embindAssignExports is called in the generated code after embind is
   // initialized. It will link the bindings to the ES module exports.
-  $embindUpdateExports: '<<< EMBIND_AOT_UPDATE_EXPORTS >>>',
-  $embindUpdateExports__postset: 'addOnInit(embindUpdateExports);',
+  $embindAssignExports: '<<< EMBIND_AOT_ASSIGN_EXPORTS >>>',
+  $embindAssignExports__postset: () => { addAtPostCtor('embindAssignExports();'); },
 #endif
 #endif
   // If register_type is used, emval will be registered multiple times for
@@ -2320,7 +2320,7 @@ var LibraryEmbind = {
 };
 
 #if MODULARIZE == 'instance'
-extraLibraryFuncs.push('$embindUpdateExports');
+extraLibraryFuncs.push('$embindAssignExports');
 #endif
 
 addToLibrary(LibraryEmbind);
--- a/src/lib/libembind.js
+++ b/src/lib/libembind.js
@@ -34,6 +34,8 @@ var LibraryEmbind = {
 #if EMBIND_AOT
   $InvokerFunctions: '<<< EMBIND_AOT_INVOKERS >>>',
 #if MODULARIZE == 'instance'
+  // embindUpdateExports is called in the generated code after embind is
+  // initialized. It will link the bindings to the ES module exports.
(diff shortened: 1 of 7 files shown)


src/embind/embind_gen.js
```diff
@@ -52,6 +52,7 @@ var LibraryEmbind = {
       this.thisType = thisType;
       this.isNonnullReturn = isNonnullReturn;
       this.isAsync = isAsync;
+      this.hasPublicSymbol = true;
```
sbc100, line 55: Maybe just make these class fields?

sbc100 (commented):
To be clear this only works in AOT mode?    What happens in non-AOT mode?

src/embind/embind_gen.js
```diff
@@ -454,14 +458,27 @@ var LibraryEmbind = {
...
           continue;
         }
         def.printJs(out);
       }
-      out.push('}')
-      console.log(out.join(''));
+      out.push('}\n');
+      let updateExports = '() => {\n';
+      for (const publicSymbol of publicSymbols) {
+        updateExports += `__exp_${publicSymbol} = Module['${publicSymbol}'];\n`
```
sbc100, line 474: So these extra assignment happen during `onInit` in the final output module?  They don't happen at AOT time, right?  Or do they happen both at AOT time and at runtime?

brendandahl:
> To be clear this only works in AOT mode? What happens in non-AOT mode?

In non-aot mode, the user would have to get the bindings off the module object from init. e.g.
```js
const module = await init();
module.foo();
```

We could require that modularize instance + embind only works with embind_aot, or enable embind_aot in instance mode by default. I lean towards the latter.

src/embind/embind_gen.js
brendandahl: They happen just in the final output at runtime.

src/embind/embind_gen.js
brendandahl: Nice, didn't realize that worked now.

src/embind/embind.js
```diff
@@ -36,7 +36,11 @@ var LibraryEmbind = {
   $PureVirtualError: undefined,
   $GenericWireTypeSize: {{{ 2 * POINTER_SIZE }}},
 #if EMBIND_AOT
-  $InvokerFunctions: '<<< EMBIND_AOT_OUTPUT >>>',
+  $InvokerFunctions: '<<< EMBIND_AOT_INVOKERS >>>',
+#if MODULARIZE == 'instance'
+  $embindUpdateExports: '<<< EMBIND_AOT_UPDATE_EXPORTS >>>',
```
sbc100, line 41: How about adding some comments here about that `embindUpdateExports` does and when?

sbc100 (approved):

brendandahl:
I've updated this to work with the changes with `exp__` being removed. I was hoping to use the emscripten.py code for the exports, but unfortunately `phase_embind_aot` runs after `emscript`.

src/lib/libembind.js
```diff
@@ -32,7 +32,13 @@ var LibraryEmbind = {
   $PureVirtualError: class extends Error {},
   $GenericWireTypeSize: {{{ 2 * POINTER_SIZE }}},
 #if EMBIND_AOT
-  $InvokerFunctions: '<<< EMBIND_AOT_OUTPUT >>>',
+  $InvokerFunctions: '<<< EMBIND_AOT_INVOKERS >>>',
+#if MODULARIZE == 'instance'
+  // embindAssignExports is called in the generated code after embind is
+  // initialized. It will link the bindings to the ES module exports.
+  $embindAssignExports: '<<< EMBIND_AOT_ASSIGN_EXPORTS >>>',
+  $embindAssignExports__postset: () => { addAtPostCtor('embindAssignExports();'); },
```
sbc100, line 40: I wonder if there is any point in have this be part of the library file at all?  Couldn't the code in link.py just emit the whole `embindAssignExports()` function and call `addOnPostCtor(embindAssignExports)` right after?

tools/link.py
```diff
@@ -2052,10 +2052,23 @@ def phase_emit_tsd(options, wasm_target, js_target, js_syms, metadata):
...
-  src = do_replace(src, '<<< EMBIND_AOT_OUTPUT >>>', out)
+  src = do_replace(src, '<<< EMBIND_AOT_INVOKERS >>>', out['invokers'])
+  if settings.MODULARIZE == 'instance':
+    # Add ES module exports for the embind exports.
+    decls = '\n'.join([f'export var {name};' for name in out['publicSymbols']])
+    # Assign the runtime exports from Module to the ES export.
+    assigns = '\n'.join([f'{name} = Module[\'{name}\'];' for name in out['publicSymbols']])
+    exports = f'''
+// start embind exports
+() => {{ {assigns} }};
```
sbc100, line 2068: This looks like an anonymous function that is never called.. maybe I'm missing something?

test/test_other.py
```diff
@@ -3464,7 +3464,7 @@ def test_embind_tsgen_end_to_end(self, opts, tsc_opts):
     # Check that TypeScript generation works and that the program is runs as
     # expected.
     self.emcc(test_file('other/embind_tsgen.cpp'),
-              ['-o', 'embind_tsgen.js', '-lembind', '--emit-tsd', 'embind_tsgen.d.ts'] + opts)
+              ['-o', 'embind_tsgen.js', '-sEMBIND_AOT', '-lembind', '--emit-tsd', 'embind_tsgen.d.ts'] + opts)
```
sbc100, line 3467: Did you mean to add this here?  I looks like the `esm_with_jsgen` variant already tests this?

src/lib/libembind.js
brendandahl: Yeah, this just provided a place to inject the code. I can just append it to the JS. It will end up after the postamble.js though. Is that fine?

tools/link.py
brendandahl: The function gets assigned to the library symbol`embindAssignExports`. It's kind of a hack to include the declarations after it.

src/lib/libembind.js
sbc100: It it works it seems maybe less hacky?

sbc100: It it works it seems maybe less hacky?

sbc100 (approved):

PR #23405, opened 2025-01-14
[MINIMAL_RUNTIME] Remove WASM_MODULE_EXPORTS replacement. NFC

sbc100:
Followup to #23403

We now generate a new function called `assignWasmExports` which gets called instead of the `WASM_MODULE_EXPORTS` replacement.

As is evident from the lack of code size changes closure is able to inline this function.

The motivation for this change is 3-fold:

1. Remove differences between MINIMAL_RUNTIME and regular runtime.
2. Simplifies and logic by removing magic late stage replacement.
3. Sets the stage for regular runtime to use the same technique.

### DIFF
--- a/test/js_optimizer/minimal-runtime-2-emitDCEGraph.js
+++ b/test/js_optimizer/minimal-runtime-2-emitDCEGraph.js
@@ -65,11 +65,14 @@ var imports = {
  }
 };
 var _main, _unused, wasmExports;
-WebAssembly.instantiate(Module["wasm"], imports).then(((output) => {
- wasmExports = output.instance.exports;
+function assignWasmExports(wasmExports) {
  _main = wasmExports["b"];
  _unused = wasmExports["c"];
- initRuntime();
+}
+WebAssembly.instantiate(Module["wasm"], imports).then(((output) => {
+ wasmExports = output.instance.exports;
+ assignWasmExports(wasmExports);
+ initRuntime(wasmExports);
  ready();
 }));
 
--- a/test/js_optimizer/minimal-runtime-emitDCEGraph.js
+++ b/test/js_optimizer/minimal-runtime-emitDCEGraph.js
@@ -36,7 +36,7 @@ var wasmImports = {
 function run() {
  var ret = _main();
 }
-function initRuntime() {
+function initRuntime(wasmExports) {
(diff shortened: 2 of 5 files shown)


sbc100:
The changes to tools/acorn-optimizer.mjs are mostly just deleting code (view with ignore whitespace to get a better view)

test/js_optimizer/minimal-runtime-emitDCEGraph.js
```diff
@@ -36,7 +36,7 @@ var wasmImports = {
 function run() {
  var ret = _main();
 }
-function initRuntime() {
+function initRuntime(wasmExports) {
```
brendandahl, line 39: Needed?

sbc100: This just updates the test code to look more like the generated code.  Its good if be testing stuff that looks like that actual output.   This test I assume was originally copied from the output of emscripten and as such it can get stale over time.

test/js_optimizer/minimal-runtime-2-emitDCEGraph.js
```diff
@@ -65,11 +65,14 @@ var imports = {
...
  _main = wasmExports["b"];
  _unused = wasmExports["c"];
- initRuntime();
+}
+WebAssembly.instantiate(Module["wasm"], imports).then(((output) => {
+ wasmExports = output.instance.exports;
+ assignWasmExports(wasmExports);
+ initRuntime(wasmExports);
  ready();
 }));
```
kripken, line 77: This looks a little larger, but I guess it gets optimized by closure?

kripken (approved):

test/js_optimizer/minimal-runtime-2-emitDCEGraph.js
sbc100: Yes, closure just inlines this.

sbc100: As you can see its a code size win actually

sbc100: Wait.. sorry, this change is code size neutral.. I was getting confused between changes.

Issue #23396, opened 2025-01-14
TypeError [ERR_INVALID_ARG_TYPE] error when compiling with 4.0.0 (regression)


ypujante:
I guess it is a possibility that I was doing something wrong all along, but until 4.0.0 my project was compiling fine on my local machine.

Steps to reproduce:
* clone [emscripten-glfw](https://github.com/pongasoft/emscripten-glfw)
* follow the [Building](https://github.com/pongasoft/emscripten-glfw/blob/master/docs/Building.md) instructions which are very simple (and you only need to build the example-minimal)
```
emcmake cmake -B build_debug -DCMAKE_BUILD_TYPE=Debug .
cd build_debug
cmake --build . --target example-minimal
```

Running this last command results in the following error:

```text
[100%] Linking CXX executable main.html
cache:INFO: generating system asset: symbol_lists/9fdde3d1fa007822b3315ffcfdd0c9b998a32ecf.json... (this will be cached in "/usr/local/emsdk/upstream/emscripten/cache/symbol_lists/9fdde3d1fa007822b3315ffcfdd0c9b998a32ecf.json" for subsequent builds)
node:internal/errors:478
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:387:5)
    at validateString (node:internal/validators:162:11)
    at Module.join (node:path:1172:7)
    at find (file:///usr/local/emsdk/upstream/emscripten/src/utility.mjs:235:27)
    at read (file:///usr/local/emsdk/upstream/emscripten/src/utility.mjs:229:20)
    at file:///usr/local/emsdk/upstream/emscripten/src/compiler.mjs:25:34
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:530:24)
    at async loadESM (node:internal/process/esm_loader:91:5) {
  code: 'ERR_INVALID_ARG_TYPE'
}
em++: error: '/usr/local/emsdk/node/16.20.0_64bit/bin/node /usr/local/emsdk/upstream/emscripten/src/compiler.mjs /var/folders/wz/zqhny4ks031bqlr6r27bfkzm0000gn/T/tmpeoyuhr3e.json --symbols-only' failed (returned 1)
```

[...]

PR #23407, opened 2025-01-14
Fix import.meta.dirname in node 18

kripken:
Likely I am doing something wrong, but I need this change, or else this happens for me locally with node `v18.20.1`:
```
$ ./emcc test/hello_world.c
node:internal/errors:496
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:405:5)
    at validateString (node:internal/validators:162:11)
    at Module.join (node:path:1171:7)
    at find (file:///emscripten/src/utility.mjs:235:27)
    at read (file:///emscripten/src/utility.mjs:229:20)
    at file:///emscripten/src/compiler.mjs:30:34
    at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:337:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:106:12) {
  code: 'ERR_INVALID_ARG_TYPE'
}

Node.js v18.20.1
emcc: error: 'nodejs emscripten/src/compiler.mjs /tmp/tmp9wv0eyav.json' failed (returned 1)
```
Clearing the cache, running bootstrap, `npm ci` etc., nothing works except for this patch.

### DIFF
--- a/src/utility.mjs
+++ b/src/utility.mjs
@@ -232,6 +232,10 @@ export function read(filename) {
 
 function find(filename) {
   for (const prefix of [process.cwd(), import.meta.dirname]) {
+    // import.meta.dirname does not work in all versions of node.
+    if (!prefix) {
+      continue;
+    }
     const combined = path.join(prefix, filename);
     if (fs.existsSync(combined)) {
       return combined;


sbc100:
No, you are not doing anything wrong.   This was an unfortunate side effect of #23349.   I accidentally caused the compiler to start requiring node v19.

See #23396 

emsdk was updated to v20 a while back now so it went unnoticed.

How did you end up with v18 on your system?  Is it an old emsdk version maybe?  Maybe you also forgot to do `emsdk activate latest`?

kripken:
This is just the system node. I have it set to be used in my `.emscripten` file I guess, from a long time ago.

This does seem to fix things for me. Can we land it, or does it have downsides?

sbc100:
The right way to revert this would be to restore this part of the code to how it was before #23349.

kripken:
Ok, I'll switch to 20 locally then.

sbc100:
No, we should still support v18 if its the one in debian/stable.  Working on anther change now..

PR #23059, opened 2024-12-03
[file_packager] Add missing `support_node` check when using preload cache

dashodanger:
I was receiving an "Uncaught (in promise) ReferenceError: isNode is not defined" error when using the preload cache and building with sENVIRONMENT=web. This appears to fix the issue and seems to be in line with other checks for options.support_node in the script

PR #23408, opened 2025-01-14
Add a way to clear all IndexedDB databases in browser tests

kripken:
Without this, `browser.test_preload_caching` works once, but running it again
finds the cached data and makes it error.

Also clean up the test a little while there.

Unblocks the test for #23059

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -604,13 +604,7 @@ If manually bisecting:
         fclose(f);
         printf("|%%s|\n", buf);
 
-        if (strcmp("load me right before", buf)) {
-          puts("bad data");
-          // Return a code that is easily differentiable from the 0, 1 that
-          // the main test returns, which is the # of cached packages.
-          emscripten_force_exit(99);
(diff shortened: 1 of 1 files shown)


test/test_browser.py
```diff
@@ -584,7 +604,13 @@ def test_preload_caching(self, extra_size):
         fclose(f);
         printf("|%%s|\n", buf);
 
-        assert(strcmp("load me right before", buf) == 0);
+        if (strcmp("load me right before", buf)) {
+          puts("bad data");
+          // Return a code that is easily differentiable from the 0, 1 that
+          // the main test returns, which is the # of cached packages.
+          emscripten_force_exit(99);
+        }
```
sbc100, line 612: What is the actual problem here?    I think `assert` is generally better since it send the line number and filename back.   I think it should trigger onAbort which should never be confused with `exit:N`.

kripken: Ok, reverted.

Issue #23339, opened 2025-01-08
Stop using re-assignment thunks when WASM_ASYNC_COMPILATION is enabled.


sbc100:
In release mode, when WASM_ASYNC_COMPILATION is used (i.e the default) we generate little thunks for all wasm exports. They looks like this:

```
var _foo = Module['_foo'] = () => (_foo = Module['_cfoo'] = wasmExports['foo'])();
```

The idea here is that these would then get re-assigned on first use.   However, there are several problem with this:

1. If somebody takes a reference early they will get a reference to wrapper that will *not* get re-assigned and will cause perpetual re-assignment:

```
// This runs early on
var myfunc = Module['foo'];

// Now my func will forever point to the wrapper:

myfunc();  // always goes via wrapper and always re-assigns.. strange
Module['foo'](); // now goes direct
```

2. The identity of these function changes which can be observable (since we don't this re-assignment thing in debug builds):

```
var myfunc = Module['foo'];
// later after instantiation and usage
if (myfunc == Module['foo']) // would always be true in debug builds, but can be false in release builds
```

See #23337

sbc100:
(1) actually seems fairly bad since it perpetually mutates the `Module` and the `globalThis`.

sbc100:
One possible solution would be to simply delay the assignments and just generate initially-undefined variables:

```
var _foo;  // Will simply be undefined until the instantiation is done.  As will `Module['_foo']`
```

In debug builds we could generate wrappers that simply fail if they are ever called:

```
var _foo = Module['_foo'] = reportModuleNotReady('_foo');
```

The major breakage here would be for folks that are currently taking references to `_foo` or `Module['_foo']` before the module is ready.

PR #23411, opened 2025-01-15
Avoid creating wrapper thunks or wasm exports

sbc100:
In debug builds nothing really changes here since we exported functions still get statically assigned to debug wrappers/thunks.

However, after this change, in release builds, if you attempt to access `_foo` (internally) or `Module['_foo']` externally before the module is loaded you will now get `undefined`.

Since it was always the case that calling these function would crash anyway, the only effected folks would be folks that take the reference early and then call later. e.g.

```
var myfunc = Module['_foo'];  // Run before module is loaded
// wait until loaded
myfunc();  // This will no longer work.
```

Fixes: #23339


sbc100:
Maybe this new behaviour should be behind `-sSTRICT` to avoid breaking folks?

sbc100:
Or perhaps I can introduce a warning when folks do this, and let that cook for a few releases first?

sbc100:
Marking as draft for now

brendandahl:
It seems like if we do this, then debug builds should behave in a similar way. i.e. if you take a reference to a not ready function then that function should never work.

```js
// Before the wasm module is loaded
var myFoo = Module['foo'];
Module['onRuntimeInitialized'] = () => {
  myFoo(); // <<< should not work in debug or opt builds
};
```

sbc100:
> It seems like if we do this, then debug builds should behave in a similar way. i.e. if you take a reference to a not ready function then that function should never work.
> 
> ```js
> // Before the wasm module is loaded
> var myFoo = Module['foo'];
> Module['onRuntimeInitialized'] = () => {
>   myFoo(); // <<< should not work in debug or opt builds
> };
> ```

Yes that is correct.   The only difference I would propose is that maybe we can make the debug builds fail in a nicer way than the release builds:

- release build: attempt to call undefined
- debug build: `abort("call to myFoo reference taken before module is ready")`

That way at least we help the folks who were previously relying on this.

However, I'm not totally sure it worth breaking anybody here, which is why I marked this as a draft.   Do many folks do you think would be broken if that above code (which works today) stops working?

brendandahl:
I feel like most the code I've seen people usually call through the module object, but I've mainly looked at g3 code. We could try testing this in g3 too.

At least if it does break someone, the error is not a subtle one and should be relatively easy to fix.

sbc100:
This change landed as #24384

Issue #23372, opened 2025-01-11
emscripten_webgl_create_context ignores EmscriptenWebGLContextAttributes.majorVersion


ypujante:
I created the following test file:

```cpp
#include <stdio.h>
#include <emscripten.h>
#include <emscripten/html5.h>
#include <GLES3/gl3.h>
int main() {
  constexpr int version = 0;
  EmscriptenWebGLContextAttributes attributes{};
  attributes.majorVersion = version;
  auto handle = emscripten_webgl_create_context("#canvas", &attributes);
  emscripten_webgl_make_context_current(handle);
  printf("%d | WebGL: %s | %s\n", version, glGetString(GL_VERSION), glGetString(GL_RENDERER));
  emscripten_webgl_destroy_context(handle);
  return 0;
}
```

Depending on which compilation flag it is called with, `emscripten_webgl_create_context` simply ignores the provided `EmscriptenWebGLContextAttributes.majorVersion`

### Case 1: NO flag (ignored => 1.0 always)
```text
// emcc issue.cpp -o build/index.html
// 0 | WebGL: OpenGL ES 2.0 (WebGL 1.0 (OpenGL ES 2.0 Chromium)) | WebKit WebGL
// 1 | WebGL: OpenGL ES 2.0 (WebGL 1.0 (OpenGL ES 2.0 Chromium)) | WebKit WebGL
// 2 | WebGL: OpenGL ES 2.0 (WebGL 1.0 (OpenGL ES 2.0 Chromium)) | WebKit WebGL
```

### Case 2: `-sMIN_WEBGL_VERSION=2` (ignored => 2.0 always)
```text
// 0 | WebGL: OpenGL ES 3.0 (WebGL 2.0 (OpenGL ES 3.0 Chromium)) | WebKit WebGL
// 1 | WebGL: OpenGL ES 3.0 (WebGL 2.0 (OpenGL ES 3.0 Chromium)) | WebKit WebGL
// 2 | WebGL: OpenGL ES 3.0 (WebGL 2.0 (OpenGL ES 3.0 Chromium)) | WebKit WebGL
```

### Case 3: `-sMAX_WEBGL_VERSION=2` (uses value)
```text
// 0 | WebGL: OpenGL ES 2.0 (WebGL 1.0 (OpenGL ES 2.0 Chromium)) | WebKit WebGL
[...]

PR #23416, opened 2025-01-15
Add Webgl major version checks

ypujante:
This are the changes we discussed in #23372 

I modified the documentation and changed the code to abort (only when assertions are enabled).

I added a comprehensive test for the changes. I would not be surprised if other tests fail, but I guess I will fix them accordingly. I have manually tested `test_webgl2_runtime_no_context` and it works.

### DIFF
--- a/src/library_html5_webgl.js
+++ b/src/library_html5_webgl.js
@@ -91,6 +91,16 @@ var LibraryHtml5WebGL = {
       renderViaOffscreenBackBuffer: HEAP8[attributes + {{{ C_STRUCTS.EmscriptenWebGLContextAttributes.renderViaOffscreenBackBuffer }}}]
     };
 
+
+#if ASSERTIONS
+    assert(contextAttributes.majorVersion === 1 || contextAttributes.majorVersion === 2, `${contextAttributes.majorVersion} is not a valid value for contextAttributes.majorVersion (should be 1 or 2)`);
+#if MIN_WEBGL_VERSION >= 2
+    assert(contextAttributes.majorVersion === 2, 'Requesting a WebGL context with version 1 is incompatible with linker flag -sMIN_WEBGL_VERSION=2');
+#elif MAX_WEBGL_VERSION == 1
+    assert(contextAttributes.majorVersion === 1, 'Requesting a WebGL context with version 2 requires either -sMIN_WEBGL_VERSION=2 or -sMAX_WEBGL_VERSION=2 linker flag');
+#endif
+#endif
+
     var canvas = findCanvasEventTarget(target);
 
 #if GL_DEBUG
--- a/test/browser/test_webgl_context_major_version.c
+++ b/test/browser/test_webgl_context_major_version.c
@@ -1,0 +1,19 @@
+// Copyright 2025 The Emscripten Authors.  All rights reserved.
+// Emscripten is available under two separate licenses, the MIT license and the
+// University of Illinois/NCSA Open Source License.  Both these licenses can be
+// found in the LICENSE file.
+
+#include <assert.h>
+#include <emscripten/html5.h>
+
+
+int main()
+{
+  EmscriptenWebGLContextAttributes attrs;
+  emscripten_webgl_init_context_attributes(&attrs);
+  attrs.majorVersion = WEBGL_CONTEXT_MAJOR_VERSION; // provided when invoking the test
+  EMSCRIPTEN_WEBGL_CONTEXT_HANDLE context;
+  context = emscripten_webgl_create_context("#canvas", &attrs);
+  assert(context);
+  emscripten_webgl_destroy_context(context);
+}
--- a/site/source/docs/api_reference/html5.h.rst
+++ b/site/source/docs/api_reference/html5.h.rst
@@ -2069,7 +2069,11 @@ Functions
   .. note::
 
     - A successful call to this function will not immediately make that rendering context active. Call :c:func:`emscripten_webgl_make_context_current` after creating a context to activate it.
-    - This function will try to initialize the context version that was *exactly* requested. It will not e.g. initialize a newer backwards-compatible version or similar.
+    - A word of caution about :c:type:`EmscriptenWebGLContextAttributes.majorVersion`:
+
+      - When no (WEBGL) linker flags are set, then this attribute is ignored and the context returned is WebGL 1.0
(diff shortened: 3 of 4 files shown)


src/library_html5_webgl.js
```diff
@@ -91,6 +91,16 @@ var LibraryHtml5WebGL = {
       renderViaOffscreenBackBuffer: HEAP8[attributes + {{{ C_STRUCTS.EmscriptenWebGLContextAttributes.renderViaOffscreenBackBuffer }}}]
     };
 
+
+#if ASSERTIONS
+    assert(contextAttributes.majorVersion === 1 || contextAttributes.majorVersion === 2, `${contextAttributes.majorVersion} is not a valid value for contextAttributes.majorVersion (should be 1 or 2)`);
```
sbc100, line 96: How about just `invalid GL version requested: ${contextAttributes.majorVersion}`.   

I tend to prefer shorted error message, but I get that not everyone agrees with me :)

src/library_html5_webgl.js
```diff
@@ -91,6 +91,16 @@ var LibraryHtml5WebGL = {
...
+#if MIN_WEBGL_VERSION >= 2
+    assert(contextAttributes.majorVersion === 2, 'Requesting a WebGL context with version 1 is incompatible with linker flag -sMIN_WEBGL_VERSION=2');
+#elif MAX_WEBGL_VERSION == 1
+    assert(contextAttributes.majorVersion === 1, 'Requesting a WebGL context with version 2 requires either -sMIN_WEBGL_VERSION=2 or -sMAX_WEBGL_VERSION=2 linker flag');
```
sbc100, line 100: I wonder if we should actually fail in this case?  Even in release mode?

It seems better than just ignoring the invalid values maybe?

Would that break a lot of existing code?   It seems like the asserts are going to break just as many users (just in their debug builds only).

test/browser/test_webgl_context_major_version.c
```diff
@@ -0,0 +1,19 @@
...
+// Emscripten is available under two separate licenses, the MIT license and the
+// University of Illinois/NCSA Open Source License.  Both these licenses can be
+// found in the LICENSE file.
+
+#include <assert.h>
+#include <emscripten/html5.h>
+
+
+int main()
+{
```
sbc100, line 11: Curly braces on the same line as the function decl.

test/browser/test_webgl_context_major_version.c
```diff
@@ -0,0 +1,19 @@
...
+  EmscriptenWebGLContextAttributes attrs;
+  emscripten_webgl_init_context_attributes(&attrs);
+  attrs.majorVersion = WEBGL_CONTEXT_MAJOR_VERSION; // provided when invoking the test
+  EMSCRIPTEN_WEBGL_CONTEXT_HANDLE context;
+  context = emscripten_webgl_create_context("#canvas", &attrs);
+  assert(context);
+  emscripten_webgl_destroy_context(context);
+}
```
sbc100, line 19: Missing newline.

src/library_html5_webgl.js
sbc100: If we do stick with asserts maybe we can shorten these errors a little too.

How about: "WebGL2 requested but only WebGL1 is supported (MAX_WEBGL_VERSION is 1)"

sbc100 (approved):
Thanks!   lgtm % comments

src/library_html5_webgl.js
ypujante: I thought about the "failing in release mode" as well, but I am just worried that it might break a lot of code that doesn't really care.

For example, ImGui requests a WebGL 2 context (via the `GLFW_CONTEXT_VERSION_MAJOR` API), but does not provide any compiler flag and as result really gets a WebGL 1 (+ extensions because that is the default) and clearly that is enough for ImGui to work. Maybe there is a lot of other code that behave in a similar fashion.

Also note that these changes are **only** in place when calling the `emscripten_webgl_create_context` API. I purposely did not put these checks in `GL.createContext` which gets called by other parts of the code like `Browser.createContext`.

I think I would leave it in debug build only.

sbc100: But the number of folks with broken build are the same either way.  Only breaking their debug builds is still breaking them.

How about we use `warn` here in debug builds, which would mean folks would have time to investivate?  We can then turn the `warn` into an error at some point in the future?

ypujante: sounds like a good idea:

1) how do I implement it?
2) how do I test it?

Thanks

sbc100:
```
#if ASSERTIONS
if (...) {
  warm(...);
} 
#endif
```

Testing here is a little harder because getting access to the console messages is tricky... we would have to look at which other tests to this.

ypujante: Ok I will investigate other tests. Thanks for the pointer.

sbc100: I think you might have to do something like in `test/emscripten_console_log_pre.js` in order to detect the warn in code.

Then in the test code would write something like:

```
int main() {
   ...
  EM_ASM(assert(warnOutput == EXPECTED_WARN_OUTPUT));
  ...
}
```

sbc100: I think I'd also be ok with skipping testing in this case, assuming you verify the warnings manually.

ypujante: I am not sure what I am doing wrong, but adding `$warn` to the list of `__deps` does not seem to work:

```
error: undefined symbol: $warn (referenced by emscripten_webgl_do_create_context, referenced by emscripten_webgl_create_context, referenced by root reference (e.g. compiled C/C++ code))
```

and of course not adding it result in:

```
+//report_result?exception:warn is not defined / ReferenceError: warn is not defined
```

any idea?

sbc100: Oh sorry, I thought we had `warn` function but I guess we don't at runtime anyway :)

We have `warnOnce` in the library, but in this case I think you can just do `err(..)` so show the error.  You don't need to depend on `err` since its always available.

ypujante:
I made the changes you requested and I am now calling `err` instead of aborting.

I found a way to still make the test "work" by replacing `err` with `abort` so that the test fail. It is true that it is changing the rest of the code, because what comes after `err` is no longer executed, but I think for our purposes that is ok.

If you do not like this I can remove the tests entirely.

sbc100 (approved):
lgtm % nits

sbc100 (approved):

PR #23419, opened 2025-01-15
[jspi] Fix TSD generation when using asyncify library functions.

### LINKED ISSUE #23418: Build still fails when both -sJSPI and --emit-tsd are active
```bash
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.74 (1092ec30a3fb1d46b1782ff1b4db5094d3d06ae5)
clang version 20.0.0git (https:/github.com/llvm/llvm-project 322eb1a92e6d4266184060346616fa0dbe39e731)
Thread model: posix
```

**Issue:**
*This is a follow-up to https://github.com/emscripten-core/emscripten/issues/23272 because it was closed.*

When enabling the **JSPI** flag and attempting to emit **Typescript definitions** (via --emit-tsd), the build fails. Since we disable Asyncify the `library_async` is not included and the linker complains about missing `emscripten_sleep()` function.

Demo to reproduce: https://github.com/mroohian/emscripten-jspi-demo

```bash
# This command is executed by CMake
em++ -gsource-map \
     -lembind \
     --emit-tsd test1.d.ts \   # It works without this line
     -s EXPORT_NAME=Test1Module \
     -s JSPI \
     -s STRICT \
     -s USE_ES6_IMPORT_META \
     -s MODULARIZE \
     -s EXPORT_ES6=1 \
     --no-entry \
     test1.cpp.o \
     -o test1.js
```

```shell
> emscripten-jspi-demo@1.0.0 cmake:build
> cd build.em && make

-- Configuring done (1.2s)
-- Generating done (0.0s)
-- Build files have been written to: /root/workspace/emscripten-jspi-demo/build.em
[ 33%] Building CXX object CMakeFiles/test1.dir/src/wasm/calc.cpp.o
[ 66%] Building CXX object CMakeFiles/test1.dir/src/wasm/test1.cpp.o
[100%] Linking CXX executable test1.js
em++: warning: -sASYNCIFY=2 (JSPI) is still experimental [-Wexperimental]
error: undefined symbol: emscripten_sleep (referenced by root reference (e.g. compiled C/C++ code))
warning: _emscripten_sleep may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
em++: error: '/root/workspace/emsdk/node/20.18.0_64bit/bin/node /root/workspace/emsdk/upstream/emscripten/src/compiler.mjs /tmp/tmpyi_gjmeb.json' failed (returned 1)
make[2]: *** [CMakeFiles/test1.dir/build.make:114: test1.js] Error 1
make[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/test1.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
```

brendandahl:
Instead of disabling JSPI during TSD generation just skip instrumenting the wasm exports and imports so JSPI isn't needed.

Fixes #23418

### DIFF
--- a/test/other/embind_tsgen_jspi.d.ts
+++ b/test/other/embind_tsgen_jspi.d.ts
@@ -1,17 +1,4 @@
 // TypeScript bindings for emscripten-generated code.  Automatically generated at compile time.
-declare namespace RuntimeExports {
-    let HEAPF32: any;
-    let HEAPF64: any;
-    let HEAP_DATA_VIEW: any;
-    let HEAP8: any;
-    let HEAPU8: any;
-    let HEAP16: any;
-    let HEAPU16: any;
-    let HEAP32: any;
-    let HEAPU32: any;
-    let HEAP64: any;
-    let HEAPU64: any;
-}
 interface WasmModule {
 }
 
@@ -19,4 +6,4 @@ interface EmbindModule {
   sleep(): Promise<void>;
 }
 
-export type MainModule = WasmModule & typeof RuntimeExports & EmbindModule;
+export type MainModule = WasmModule & EmbindModule;
--- a/src/library_async.js
+++ b/src/library_async.js
@@ -38,6 +38,10 @@ addToLibrary({
     rewindArguments: {},
 #endif
     instrumentWasmImports(imports) {
+#if EMBIND_GEN_MODE
+      // Instrumenting is not needed when generating code.
+      return imports;
+#endif
 #if ASYNCIFY_DEBUG
       dbg('asyncify instrumenting imports');
 #endif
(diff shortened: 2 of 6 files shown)


test/other/embind_tsgen_jspi.d.ts
```diff
@@ -1,22 +1,9 @@
...
-    let HEAP_DATA_VIEW: any;
-    let HEAP8: any;
-    let HEAPU8: any;
-    let HEAP16: any;
-    let HEAPU16: any;
-    let HEAP32: any;
-    let HEAPU32: any;
-    let HEAP64: any;
-    let HEAPU64: any;
-}
```
sbc100, line 14: I guess this is a side effect of adding `-sSTRICT`?   Is that related to this change somehow?

sbc100 (approved):

test/other/embind_tsgen_jspi.d.ts
brendandahl: Yeah, strict will remove those definitions. The reporters issue only happened in strict mode, since it turns on erroring on undefined symbols.

PR #23427, opened 2025-01-16
[EH] Rename Wasm EH tests

aheejin:
This renames Wasm EH test modes:
- `wasm` -> `wasm_legacy`
- `wasm_exnref` -> `wasm`

So `_wasm` suffix means the standardized Wasm EH, and `_wasm_legacy` means the legacy EH. This also removes some comments mentioning things like "new Wasm EH with exnref", given that this should be (eventually) the default mode and not the new thing.


test/test_other.py
```diff
@@ -15316,8 +15314,8 @@ def test_SUPPORT_BIG_ENDIAN(self):
     'noexcept': ['-fno-exceptions'],
     'default': [],
     'except': ['-sDISABLE_EXCEPTION_CATCHING=0'],
-    'except_wasm': ['-fwasm-exceptions', '-sWASM_LEGACY_EXCEPTIONS'],
-    'except_wasm_exnref': ['-fwasm-exceptions', '-sWASM_LEGACY_EXCEPTIONS=0']
+    'except_wasm': ['-fwasm-exceptions', '-sWASM_LEGACY_EXCEPTIONS=0'],
```
sbc100, line 15317: Can you just drop the `'-sWASM_LEGACY_EXCEPTIONS=0'` here maybe?

sbc100 (approved):

test/test_other.py
aheejin: This is meant to be Wasm (exnref) so without that it will use the legacy exception instead, no? (`WASM_LEGACY_EXCEPTIONS` is on by default now)

sbc100: Oh I didn't realize that part.  lgtm

PR #23428, opened 2025-01-16
[emcc.py] Handle all pass-through args correctly

sbc100:
We were only handling `-mllvm` and not all the other ones.

### DIFF
--- a/emcc.py
+++ b/emcc.py
@@ -1132,6 +1132,12 @@ def parse_args(newargs):  # noqa: C901, PLR0912, PLR0915
     arg = newargs[i]
     arg_value = None
 
+    if arg in CLANG_FLAGS_WITH_ARGS:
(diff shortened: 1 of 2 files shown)


dschuff (approved):

emcc.py
```diff
@@ -1145,6 +1145,12 @@ def parse_args(newargs):  # noqa: C901, PLR0912, PLR0915
     arg = newargs[i]
     arg_value = None
 
+    if arg in CLANG_FLAGS_WITH_ARGS:
+      # Ignore the next argument rather than trying to parse it.  This is needed
+      # because that next args could, for example, start with `-o` and we don't want
```
kripken, line 1150:
```suggestion
      # because that next arg could, for example, start with `-o` and we don't want
```

kripken (approved):

PR #23430, opened 2025-01-16
[compiler.mjs] Simplify file opening and finding logic. NFC

sbc100:
Removes the magic `find` function and instead require explicit file locations at all the various call sites.  This means that in each circumstance were we want to load a file there is only once place it can/should be.

One side effect of this is that when using the `#include` preprocessor directive the filename must either be absolute or relative to the current file.

### DIFF
--- a/src/utility.mjs
+++ b/src/utility.mjs
@@ -227,21 +227,17 @@ export function isDecorator(ident) {
 }
 
 export function read(filename) {
-  const absolute = find(filename);
-  return fs.readFileSync(absolute, 'utf8');
+  return fs.readFileSync(filename, 'utf8');
 }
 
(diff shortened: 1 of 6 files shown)


src/utility.mjs
```diff
@@ -227,21 +227,15 @@ export function isDecorator(ident) {
...
 
-function find(filename) {
-  for (const prefix of [process.cwd(), __dirname]) {
-    const combined = path.join(prefix, filename);
-    if (fs.existsSync(combined)) {
-      return combined;
-    }
-  }
-  return filename;
+export function localFile(filename) {
```
kripken, line 236:
```suggestion
// Returns an absolute path for a file, under the src/ directory.
export function localFile(filename) {
```

kripken (approved):

PR #23433, opened 2025-01-16
Move addOnInit and __ATINIT__ into a JS library.

brendandahl:
Make it so the `addOnInit` code is not emitted if it is not needed.

Since `preamble.js` is not a library I changed the call to
`__wasm_call_ctors` to be inlined like it is in `MINIMAL_RUNTIME`. I also
had to add various dependencies on `addOnInit` and `__ATINIT__`.


src/library.js
```diff
@@ -2296,6 +2296,16 @@ addToLibrary({
 
   $noExitRuntime: "{{{ makeModuleReceiveExpr('noExitRuntime', !EXIT_RUNTIME) }}}",
 
+  // functions called during startup
```
brendandahl, line 2299: TODO: add changelog.

brendandahl:
I've got an issue with pthreads ordering. `__wasm_call_ctors` needs to happen after some calls, but before other addOnInit code is run. Hmm...

src/library.js
```diff
@@ -2296,6 +2296,16 @@ addToLibrary({
...
+  $__ATINIT__: [],
+  $__ATINIT____postset: () => {
```
sbc100, line 2301: I keep forgetting that `__postset` can be a function...

tools/link.py
```diff
@@ -701,6 +701,7 @@ def phase_linker_setup(options, state):  # noqa: C901, PLR0912, PLR0915
       exit_with_error('PTHREADS_PROFILING only works with ASSERTIONS enabled')
     options.post_js.append(utils.path_from_root('src/threadprofiler.js'))
     settings.REQUIRED_EXPORTS.append('emscripten_main_runtime_thread_id')
+    settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$addOnInit']
```
sbc100, line 704: Why this?  Maybe worth a comment?

src/parseTools.mjs
```diff
@@ -709,7 +709,7 @@ export const ATMAINS = [];
 export const ATINITS = [];
 
 function addAtInit(code) {
-  ATINITS.push(code);
+  ATINITS.unshift(code);
```
sbc100, line 712: Why this change?

src/library.js
sbc100, line 2299: Perhaps this comment could be a little more descriptive?   I'm not sure if this is correct but maybe "Function the are called once the runtime is initialized and static constructors have been called".

sbc100 (commented):
I like the idea in general!

Issue #4214, opened 2016-03-31
Problem with converting the OpenGL glDrawElements function to WebGL in emscripten


hnn0003:
I'm running into an issue with how emscripten converts the OpenGL glDrawElements function to the WebGL equivalent (Glctx.drawElements). I had been getting a number of weird triangles popping around in my graphics, and I believe I’ve tracked it down to this:

In emscripten’s library_gl.js file resides the [glDrawElements](https://github.com/kripken/emscripten/blob/master/src/library_gl.js#L6981) function that replaces the OpenGL call with the same name. Part of what this function does is set up a buffer for the indices (called GLctx.ELEMENT_ARRAY_BUFFER) and (in the [GL.preDrawHandleClientVertexAttribBindings](https://github.com/kripken/emscripten/blob/master/src/library_gl.js#L6996) function call) it sets up a buffer for the vertices (called GLctx.ARRAY_BUFFER).

The problem is that in the [GL.preDrawHandleClientVertexAttribBindings](https://github.com/kripken/emscripten/blob/master/src/library_gl.js#L324) function, it is setting how large the GLctx.ARRAY_BUFFER data array will be based on the variable ‘count’, which is the size of the GLctx.ELEMENT_ARRAY_BUFFER (i.e. the array of indices). This is a problem because what the Glctx.drawElements function will actually do is use the values in the index array to determine where it needs to get its data from in the vertex array (i.e. GLctx.ARRAY_BUFFER). So if the values in the index array point to an index larger than the size of the index array, the Glctx.drawElements function will try to get data from outside the bounds of the GLctx.ARRAY_BUFFER (which just gives junk data).

An example of where the current code would have this happen (i.e. failing to produce the correct graphics) would be if you defined a triangle with an index array (or GLctx.ELEMENT_ARRAY_BUFFER) of [0, 5, 6] and a vertex array (or GLctx.ARRAY_BUFFER) containing only the position in (x,y,z) with value of 
[-1,-1,-5,
  1, 2, 3,
  2, 3, 4,
  3, 4, 5,
  4, 5, 6,
  1, -1, -5,
  1, 1, -5].
After running through emscripten, the program would calculate the size of the index array to be 3 (since only 3 indices were given) and each position has 3 values (x, y, and z), so the GLctx.ARRAY_BUFFER would be made to hold only the first 9 values:
[-1,-1,-5,
  1, 2, 3,
  2, 3, 4]
[...]

PR #23443, opened 2025-01-17
GLES compatibility layer: covers the situation when an element buffer is bound, but an array buffer is not.

caiiiycuk:
I found a case where GLES2/3 emulation does not work correctly. In GLES2/3, there are two ways to bind vertex data:

1. **Using an array buffer:** Bind an array buffer and pass an offset within it to `glVertexAttribPointer`.
2. **Using a direct pointer:** Pass a pointer directly to the vertex data and set the array buffer to 0.

However, there is an issue when using an **element array buffer** (used for indexed drawing) but not using an **array buffer**. In this case, the rendering process fails to set the correct vertices for the `glDrawElements` call. This issue arises due to the following code:

```javascript
glDrawElements: (mode, count, type, indices) => {
#if FULL_ES2
    var buf;
    var vertexes = 0;
    if (!GLctx.currentElementArrayBufferBinding) {
```

As you can see, the `vertexes` variable is initialized to zero. However, when `currentElementArrayBufferBinding` exists, the relevant code to initialize `vertexes` is skipped. This causes `GL.preDrawHandleClientVertexAttribBindings(vertexes)` to be called with `vertexes` set to zero, resulting in no data being bound to the vertex attributes.

To fix this, I initialized `vertexes` with `count` (the number of vertices) and added a test to verify the fix. However, I am not very experienced in 3D programming, so I am unsure if this change might unintentionally break something. Can someone review this and provide feedback?

### DIFF
--- a/src/library_webgl.js
+++ b/src/library_webgl.js
@@ -3851,7 +3851,7 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
   glDrawElements: (mode, count, type, indices) => {
 #if FULL_ES2
     var buf;
-    var vertexes = 0;
+    var vertexes = count;
     if (!GLctx.currentElementArrayBufferBinding) {
       var size = GL.calcBufLength(1, type, 0, count);
       buf = GL.getTempIndexBuffer(size);
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -1839,6 +1839,7 @@ simulateKeyUp(100, undefined, 'Numpad4');
       ('third_party/glbook/Chapter_9/Simple_Texture2D/Simple_Texture2D_orig.c', 'third_party/glbook/CH09_SimpleTexture2D.png'),
       ('third_party/glbook/Chapter_10/MultiTexture/MultiTexture_orig.c', 'third_party/glbook/CH10_MultiTexture.png'),
       ('third_party/glbook/Chapter_13/ParticleSystem/ParticleSystem_orig.c', 'third_party/glbook/CH13_ParticleSystem.png'),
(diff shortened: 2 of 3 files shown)


kripken:
cc @kainino0x @juj - perhaps you can take a look here?

kainino0x:
Sorry for the delay. Thank you very much for the explanation, this would be very difficult to understand otherwise!

Unfortunately, if I understand correctly, this isn't the correct value. The `vertexes` value determines the amount of vertex data to upload from the vertex attrib pointers. The reason that just `vertexes = count` doesn't work is because the index buffer can contain values larger than its size. For example instead of [0, 1, 2] it could have [3, 4, 5]. `count` would still be 3, but vertexes needs to be 6. It should be possible to adjust the test case to catch this issue.

The only way to know the correct `vertexes` value is to know the largest value in the index array. If the index array is on the GPU, then there are a few ways to know that: (1) download the index buffer and find its largest value, (2) keep a shadow copy of the entire index buffer on the CPU and keep that up to date, (3) track the largest value in the index buffer and keep that up to date. IIRC something similar to Option 3 is used internally in browsers to avoid making unsafe draw calls where a value in the index buffer would access out of bounds of any of the attached vertex buffers.

At the very least, I think Emscripten should abort here instead of just silently doing the wrong thing. Option 1 would be very slow - it would work, but if we did that we should `warnOnce` when it happens. Options 3 is better, and I'm pretty sure it's theoretically possible in Emscripten, but it would definitely be more complex - unless Emscripten happens to already be tracking that info (I have no idea; I'm not very familiar with FULL_ES at all).

caiiiycuk:
Thank you for the explanation! I am not sure if we should spend time on this because, in my case, I have already found a workaround. So, maybe it's simpler to just mention this bug in the documentation, as we did for https://github.com/emscripten-core/emscripten/issues/4214. I will close this for now.

PR #14306, opened 2021-05-26
Switch from addAtMain to addAtInit in library_fs.js

sbc100:
Using addAtMain stopped working for programs without main functions
(HAS_MAIN not defined) after #13901.

Code that needs to populate the FS is mostly likely going to be run
during preRun callback which all happen before ATINITs.

PR #23451, opened 2025-01-17
Add `addAtPreRun` function for code injection at preRun.

brendandahl:
This is a first step in trying to unify more of the code used by minimal runtime and the normal runtime and also move all addOn* functions into a library. Similar to the other addAt* functions for injecting code, this adds `addAtPreRun`. `addAtPreRun` is used to have code that runs after the Wasm is loaded, but before ctors.

The new injection point allows us to move `FS.init` into atPreRun and also inline the call to `__wasm_call_ctors` at the same place in minimal runtime and the normal runtime.

### DIFF
--- a/src/parseTools.mjs
+++ b/src/parseTools.mjs
@@ -735,8 +735,8 @@ export const ATMAINS = [];
 
 export const ATPRERUN = [];
 
-// Add code that will be run after the Wasm module is loaded, but before static
-// constructors and main are run.
+// Add code that runs after the Wasm module is loaded and the runtime ATINIT
+// callbacks, but before static constructors and main are run.
 function addAtPreRun(code) {
   ATPRERUN.push(code);
 }
--- a/src/parseTools.mjs
+++ b/src/parseTools.mjs
@@ -733,6 +733,14 @@ function makeEval(code) {
 
 export const ATMAINS = [];
 
+export const ATPRERUN = [];
+
+// Add code that will be run after the Wasm module is loaded, but before static
+// constructors and main are run.
+function addAtPreRun(code) {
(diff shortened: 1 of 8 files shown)


kripken (approved):
Makes sense to me. I assume there is no impact on code size?

sbc100 (approved):

brendandahl:
> Makes sense to me. I assume there is no impact on code size?

There should be a little improvement in some cases with this patch. Closure will be able to get rid of `addOnInit` since it won't always be used by default with normal runtime now.

brendandahl:
One interesting test failure `test_fs_no_main`. It looks like [this pr](https://github.com/emscripten-core/emscripten/pull/14306) moved `FS.init` to `atInit` and added a test to ensure `FS.init` is not run before preRun (which is no longer the case).

kripken:
Hmm, that order may be necessary. It is possible to keep the order unchanged?

brendandahl:
After discussing with Sam, I've swapped it so the add**At**PreRun run after the add**On**PreRuns. i.e. compile time code runs after runtime code. This seems to preserve the order we'll want.

src/parseTools.mjs
```diff
@@ -733,6 +733,14 @@ function makeEval(code) {
 
 export const ATMAINS = [];
 
+export const ATPRERUN = [];
+
+// Add code that will be run after the Wasm module is loaded, but before static
+// constructors and main are run.
```
kripken, line 739: Maybe clarify how it relates to `addOnPreRun`?

brendandahl:
Closing this in favor of #23488.

PR #23452, opened 2025-01-17
[EH] Rename EH library modes and file names

aheejin:
- This renames `NoExceptLibrary` to `ExceptionLibrary`

- This renames the `Exception.WASM` to `Exceptions.WASM_LEGACY`, given that we plan to add a new mode that uses the standard EH proposal soon, which will be named `Exceptions.WASM`. This PR does not add the new mode yet; it just renames what we already have.

- This makes `SjLjLibrary` use `eh_mode` as in `ExceptionLibrary` instead of the current `is_wasm` boolean switch, given that we will need to add another mode that uses the standard EH proposal to it soon. Its `eh_mode` can't be set to `Exceptions.NONE` though.

- Library file suffixes changed
  - `ExceptionLibrary`
    - `NONE`: `-noexcept` -> `-noexcept` (no change)
    - `EMSCRIPTEN`: `` -> `` (no change)
    - `WASM_LEGACY`: `-except` -> `-legacyexcept`
    - `WASM` (will be added later): X -> `-wasmexcept`
  - `SjLjLibrary`
    - `EMSCRIPTEN`: `` -> `` (no change)
    - `WASM_LEGACY`: `-wasm-sjlj` -> `-legacysjlj`
    - `WASM` (will be added later): X -> `-wasmsjlj`

We give the current default mode (`EMSCRIPTEN`) no suffix. We hope we eventually change `WASM` to be the default, in which case we will give it no suffix and give `EMSCRIPTEN` a suffix like `-emexcept`/`-emsjlj`.

Also I removed / didn't add a `-` within a single suffix, so it is `-legacyexcept` rather than `-legacy-except`. The reason is if a library inherits from multiple classes, it will have multiple `-`s in its file name like `libc++abi-debug-legacyexcept.a`, and I think it is clearer if `-` only acts like split points between different suffixes.


sbc100:
I think I would rather keep the default configuration as the unprefixed one.   Do with no options specified you get the default with no prefix.  This means that we would swap the names then we make wasm exceptions that default (which might be a long time still I guess?)

sbc100:
Looks good to me!

embuilder.py
```diff
@@ -30,8 +30,8 @@
 
 # Minimal subset of targets used by CI systems to build enough to be useful
 MINIMAL_TASKS = [
-    'libcompiler_rt',
-    'libcompiler_rt-wasm-sjlj',
+    'libcompiler_rt-emsjlj',
```
sbc100, line 33: This suffix looks like it should no longer exist?

tools/system_libs.py
```diff
@@ -828,15 +826,16 @@ def get_default_variation(cls, **kwargs):
...
     super().__init__(**kwargs)
 
   def get_cflags(self):
     cflags = super().get_cflags()
-    if self.is_wasm:
-      # DISABLE_EXCEPTION_THROWING=0 is the default, which is for Emscripten
-      # EH/SjLj, so we should reverse it.
+    if self.eh_mode == Exceptions.EMSCRIPTEN:
+      cflags += ['-sSUPPORT_LONGJMP=emscripten',
+                 '-sDISABLE_EXCEPTION_THROWING=0']
```
aheejin, line 837: These are the default now so we don't need strictly this, but I added this for clarity

embuilder.py
aheejin: Good catch, thanks. Fixed.

tools/system_libs.py
```diff
@@ -732,7 +732,7 @@ def get_default_variation(cls, **kwargs):
     )
 
   @classmethod
-  def variations(cls):
+  def variations(cls, _):
```
aheejin, line 735: Without this, `mypy` will complain
```console
tools/system_libs.py:911: error: Definition of "variations" in base class "MTLibrary" is incompatible with definition in base class "SjLjLibrary"
```

tools/system_libs.py
```diff
@@ -732,7 +732,7 @@ def get_default_variation(cls, **kwargs):
     )
 
   @classmethod
-  def variations(cls):
+  def variations(cls, **_):
```
aheejin, line 735: Without this, mypy errors out like
```console
tools/system_libs.py:911: error: Definition of "variations" in base class "MTLibrary" is incompatible with definition in base class "SjLjLibrary"
```

sbc100: I can't see why this would be needed now.. maybe something is wrong.. but can't see what on first inspection.

aheejin: `libcompiler_rt` inherits both from `SjLjLibrary` and `MTLibrary`. Before this PR `MTLibrary` had `variations` but `SjLjLibrary` had `vary_on` and not `variations` so it was fine. This PR adds `variations` to `SjLjLibrary` (because soon `SjLjLibrary` will gain a new mode, so for preparation). Then the code is like
```py
class MTLibrary(Library): 
  ...
  def variations(cls):
    combos = super(MTLibrary, cls).variations()                                  
    # To save on # of variations, pthreads and Wasm workers when used together, just use pthreads variation.                                                      
    return [combo for combo in combos if not combo['is_mt'] or not combo['is_ww']]
 
class SjLjLibrary(Library):
  ...
  def variations(cls, **kwargs):
    combos = super().variations()                                                
    return ([dict(eh_mode=Exceptions.EMSCRIPTEN, **combo) for combo in combos] + 
            [dict(eh_mode=Exceptions.WASM_LEGACY, **combo) for combo in combos])
```

`mypy` errors out on this because the signatures for `variations` are different for the two classes but there is a common child class of them (`libcompiler_rt`).

So I tried to change `MTLibrary`'s `variations` to
```py
  def variations(cls, **kwargs):
```
But now `ruff` errors out saying there is an unused argument. I changed it to
```py
  def variations(cls, **_):
```
and it seems to satisfy both mypy and ruff.

sbc100: Hmm, do you know why `variations` ever take anything but just `cls`.. can we just never have `kwargs` there?

aheejin: Huh, right, `kwargs` is not actually used anywhere in `variations()`... Removed all of them.

aheejin:
After pulling this commit, emscripten ToT users need to clear the cache... Is there a way to force it?

sbc100 (approved):

PR #23454, opened 2025-01-18
[emcc.py] Simplify passing arguments to clang in compile-only mode

sbc100:
This change avoids removing the `-o` flag and the input files when only compiling.  We delay the splitting out of the linker flags and the input files until we know actually do to do linking.

This effect of this is that the compile-only phase is now much simply because it doesn't need to re-inject the input files and the `-o` flag.

This also as the effect of keeping the order of those flags preserved with respect to other command line flags when calling clang to do compilation.

### DIFF
--- a/emcc.py
+++ b/emcc.py
@@ -139,6 +139,7 @@ class EmccOptions:
   def __init__(self):
     self.target = ''
     self.output_file = None
+    self.input_files = []
     self.no_minify = False
     self.post_link = False
     self.save_temps = False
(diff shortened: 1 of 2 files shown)


emcc.py
```diff
@@ -1000,34 +1002,26 @@ def get_clang_command_asm():
...
+      cmd = get_clang_command() + newargs
     shared.exec_process(cmd)
     assert False, 'exec_process does not return'
 
   # In COMPILE_AND_LINK we need to compile source files too, but we also need to
   # filter out the link flags
   assert state.mode == Mode.COMPILE_AND_LINK
   assert not options.dash_c
+  compile_args, input_files = split_linker_flags(options, state, newargs)
   compile_args = filter_out_link_flags(compile_args)
```
sbc100, line 1025: As a follow up I hope to combine these two functions, but it makes this PR easier to follow if I leave them separate for now.

test/test_other.py
```diff
@@ -12642,8 +12642,8 @@ def test_default_to_cxx(self):
     self.assertContained("'string' file not found", err)
 
     # But it works if we pass and explicit language mode.
-    self.run_process([EMCC, '-c', 'cxxfoo.h', '-x', 'c++-header'])
-    self.run_process([EMCC, '-c', 'cxxfoo.h', '-x', 'c++'])
+    self.run_process([EMCC, '-x', 'c++-header', '-c', 'cxxfoo.h'])
+    self.run_process([EMCC, '-x', 'c++', '-c', 'cxxfoo.h'])
```
kripken, line 12646: Why change the test?

sbc100: Because emcc now passes these arguments though in the correct order:

Before it was re-ordering to put the source file at the end or `emcc -c cxxfoo.h -x c++-header' was becoming `emcc -x c++-header -c cxxfoo.h`.

Becuase `-x` (filetype) only effects the files that follow it this inadvertent / mistaken behaviour was fixing the command line hiding what should have been an error/warning from clang:

```
$ clang -c cxxfoo.h -x c++-header
clang: warning: '-x c++-header' after last input file has no effect [-Wunused-command-line-argument]
```

sbc100: In other words this was more of a bug in the test that emcc was ignored by emcc because it happened to re-order the command line flags.

kripken: I see, thanks... I didn't realize `-x` affects files after it. I thought it was global...

kripken (approved):

PR #8083, opened 2019-02-14
Static code hooks at compile time

kripken:
This lets us do stuff like this:
```
-  $PIPEFS__postset: '__ATINIT__.push(function() { PIPEFS.root = FS.mount(PIPEFS, {}, null); });',
+  $PIPEFS__postset: function() {
+   addStaticAtInit('PIPEFS.root = FS.mount(PIPEFS, {}, null);');
+  },
```
The code then appears in the right place for at-initialization time, in both minimal and normal runtimes. And then we can remove the `postamble_minimal.js` hack with hardcoded initializers.

There are also hooks for at-main and at-exit. For the minimal runtime, calling those does nothing currently, as the minimal runtime code doesn't have the placeholder to receive those. If we decide we want to support those hooks there, it's easy to add later.

This PR uses those hooks in the relevant library code. Later work can implement LLVM static constructors using this mechanism too, which would allow us to simplify things even more, but I didn't want to do too much in one PR.

To enable this PR, I also made it possible for a library `__postset` to be a function that is called at compile time.

PR #23478, opened 2025-01-22
Remove `ATMAINS` from JS compiler

sbc100:
As far as I can tell the compile-time constant was never documented.

Unlike `ATEXITS` and `ATINITS` there was never any corresponding `addAtMain`, and even the `ATMAIN` symbol itself was already not available in `-sSTRICT` or `-sLEGACY_RUNTIME=0` mode.

It looks like it was specifically added in order to facilitate the sealing of the filesystem in #8083, but its no longer used for that purpose, and we have no current internal usages.

### DIFF
--- a/src/jsifier.mjs
+++ b/src/jsifier.mjs
@@ -10,7 +10,6 @@
 import {
   ATEXITS,
   ATINITS,
-  ATMAINS,
   defineI64Param,
   indentify,
   makeReturn64,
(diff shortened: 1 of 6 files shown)


sbc100:
@brendandahl, interestingly enough ATMAINS was originally added specifically to run the `ignorePermissions = false` code after the rest of the FS.init.

I wonder if we should actually not land this PR and instead split up FS.init?

sbc100:
See https://github.com/emscripten-core/emscripten/pull/8083

sbc100:
Closing favor of #23488

PR #23480, opened 2025-01-23
[FS] Make fstat work on file descriptors with no name in nodefs

hoodmane:
This makes fstat work on anonymous nodefs file descriptors. It is the last part of #23058.

### DIFF
--- a/src/lib/libnodefs.js
+++ b/src/lib/libnodefs.js
@@ -112,65 +112,84 @@ addToLibrary({
       }
       return newFlags;
     },
-
+    getattr(func, node) {
+      var stat = NODEFS.tryFSOperation(func);
+      if (NODEFS.isWindows) {
+        // node.js v0.10.20 doesn't report blksize and blocks on Windows. Fake
+        // them with default blksize of 4096.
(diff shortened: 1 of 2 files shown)


hoodmane:
@sbc100 would appreciate review on this one.

src/lib/libnodefs.js
```diff
@@ -112,65 +112,84 @@ addToLibrary({
...
+      };
+    },
+    // Common code for both node and stream setattr
+    // For node getatrr:
+    //  - arg is a native path
+    //  - chmod, utimes, truncate are fs.chmodSync,  fs.utimesSync,  fs.truncateSync
+    // For stream getatrr:
+    //  - arg is a native file descriptor
+    //  - chmod, utimes, truncate are fs.fchmodSync, fs.futimesSync, fs.ftruncateSync
+    setattr(arg, node, attr, chmod, utimes, truncate, stat) {
```
sbc100, line 154: I can't say I love this method of adding the opaque `arg` along with all the opts for it.. but I can't think of a nicer way, and I think its better than duplicating the logic.

sbc100 (approved):

PR #23483, opened 2025-01-23
Replace eval with indirect eval

### LINKED ISSUE #23464: Consider using an indirect eval for DYNAMIC_EXECUTION
The supporting JS code for `DYNAMIC_EXECUTION` currently uses a direct `eval`, which gives the eval-ed code access to names in the scope it is called from, as bundlers (Rollup, esbuild, and so on) cannot analyze such accesses, they emit a warning and might have to resort to performing less tree shaking.

Using an indirect `eval` (`(0, eval)`) avoids this, and if possible, should be preferred.

Of course, using `eval` at all is a problem with a CSP, though I think it might be possible to workaround that by injecting dynamic script tags with a nonce instead of using eval to allow running without `unsafe-eval`. But that's another issue.

segevfiner:
Indirect eval doesn't expose the local scope to the evaluated code, and won't warn as much from bundlers

Fixes #23464

### DIFF
--- a/src/lib/libcore.js
+++ b/src/lib/libcore.js
@@ -1300,12 +1300,12 @@ addToLibrary({
   // ==========================================================================
 
   emscripten_run_script: (ptr) => {
-    {{{ makeEval('eval(UTF8ToString(ptr));') }}}
+    {{{ makeEval('(0, eval)(UTF8ToString(ptr));') }}}
   },
 
   emscripten_run_script_int__docs: '/** @suppress{checkTypes} */',
   emscripten_run_script_int: (ptr) => {
-    {{{ makeEval('return eval(UTF8ToString(ptr))|0;') }}}
+    {{{ makeEval('return (0, eval)(UTF8ToString(ptr))|0;') }}}
   },
 
   // Mark as `noleakcheck` otherwise lsan will report the last returned string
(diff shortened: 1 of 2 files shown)


hoodmane:
Unfortunately I think indirect eval does not suffice. For instance if you use `EM_JS` you will probably want to use runtime functions like `UTF8toString` in all but the simplest examples. This will fail with indirect eval.

Maybe we could have a mode where we pass `Module` as an extra first argument and then as long as the code uses `Module.UTF8toString` it would with indirect eval. But this requires more work than is present in this patch. I'll approve the ci job so we can see which tests fail.

hoodmane:
Wdyt @sbc100? I see you didn't express a concern about this in the linked issue...

hoodmane:
See in ci there are many errors like: 

`ReferenceError: _malloc is not defined`.

I think the bundler warnings may be an acceptable evil and assuming that people test the output they can find out whether the bundler broke anything or not. For many projects it's necessary to disable minification.

segevfiner:
It might still be possible to pass the needed references in for the code to use. Instead of exposing the entire scope of the module.

But I'm not sure what is expected to be defined by the eval-ed code.

sbc100:
> Unfortunately what's expected is... everything defined on the module. We could add a compiler setting `-sNO_DIRECT_EVAL` which passes `Module` in explicitly and does an indirect eval. This would change the abi for shared libraries. It could also be solved transparently by passing in `Module` and then using a `with` statement as long as we can make sure not to eval the code in strict mode.

I don't think that works either, because the `Module` object only contains the exported/public API.  EM_JS code should be allowed to access all the internals, not just the things that are exported.

Issue #23479, opened 2025-01-23
-O0 + js_library + int to boolean = error


mighty1231:
**Version of emscripten/emsdk:**

docker image with emscripten/emsdk:3.1.74

file `test.js`

```javascript
addToLibrary({
  my_js_1: function() {
    return 1;
  },
  my_js_2: function() {
    return 2;
  },
});
```

file `test.cc`

```c++
#include <emscripten/console.h>

extern "C" {
  extern bool my_js_1(void);
  extern bool my_js_2(void);
}

int main() {
  if (my_js_1()) {
    emscripten_out("my_js_1 is truthy");
  } else {
    emscripten_out("my_js_1 is falsy");
  }

  if (my_js_2()) {
    emscripten_out("my_js_2 is truthy");
  } else {
    emscripten_out("my_js_2 is falsy");
  }

  return 0;
}
```

command

```bash
$ em++ --js-library test.js -o build/O0.js -O0 test.cc
$ em++ --js-library test.js -o build/O1.js -O1 test.cc
```

### What I expected

```bash
$ node build/O0.js
my_js_1 is truthy
my_js_2 is truthy

$ node build/O1.js
my_js_1 is truthy
my_js_2 is truthy
```

### But the result for actual run is

```bash
$ node build/O0.js
my_js_1 is truthy
my_js_2 is falsy // <--- 😲

$ node build/O1.js
my_js_1 is truthy
my_js_2 is truthy
```

`-O2`, `-O3` work same as `-O1`

### Severity: flakiness on wasm worker + `-O0` option

https://github.com/emscripten-core/emscripten/blob/1092ec30a3fb1d46b1782ff1b4db5094d3d06ae5/src/library_wasm_worker.js#L239

```js
  emscripten_current_thread_is_wasm_worker: () => {
#if WASM_WORKERS
    return ENVIRONMENT_IS_WASM_WORKER; // <-- returns number 😭
#else
    // implicit return 0;
#endif
  },
```

sbc100:
I think this is undefined behaviour in C/C++ due the valid values for `bool` type.    I think that the compiler is within its rights to assume that `bool` only has two possible values (0 and 1) when you return 2 from a function that is supposed to return `bool` I believe that is undefined behaviour.  

The C/C++ compiler (if it chose too) could just compare with `1` and then assume it is false if that check failed.   This would make 2 (and all other values) appear as `0`.

You can reproduce in native clang too without emscripten:

```
$ cat main.c 
[...]

PR #23484, opened 2025-01-23
Fix emscripten_current_thread_is_wasm_worker return value.

sbc100:
This function return `ENVIRONMENT_IS_WASM_WORKER` which was being initialized to `Module['$ww']`, i.e. the worker id.

Because `emscripten_current_thread_is_wasm_worker`, is defined to return the `bool` type its only valid values are 0 and 1.  Returning values larger than 1 will result in undefined behaviour.

Fixes: #23479


src/shell.js
```diff
@@ -122,7 +122,7 @@ if (ENVIRONMENT_IS_NODE) {
 #endif // ENVIRONMENT_MAY_BE_NODE
 
 #if WASM_WORKERS
-var ENVIRONMENT_IS_WASM_WORKER = Module['$ww'];
+var ENVIRONMENT_IS_WASM_WORKER = !!Module['$ww'];
```
kripken, line 125: This increases code size - is it necessary? Can we not fix the C method on the C side? (if not, perhaps rephrase the PR title/description)

sbc100: The C function is declared as returning `bool` which allows the compiler assume its going be be either a 1 or 0.  Anything else is undefined behaviour.

We could change the return type to it, but this fix seems better, since it also means that the `ENVIRONMENT_IS_WASM_WORKER` is only ever true or false which is is what its name suggests.  Storing the thread in this variable doesn't make much sense.

kripken: Sorry, I'm not following. Perhaps my confusion is about `Module['$ww']`. I cannot find where in the code that is ever written to. Where is that done?

From the context of what you are saying, I am guessing that the output of `emscripten_current_thread_is_wasm_worker()` is written there. (Otherwise, how does this line connect to the topic of the PR?) Assuming my guess is right, then it should already be either 0 or 1, since it is a bool in C?

kripken: I see now, thanks.

kripken (approved):

PR #23488, opened 2025-01-24
Unify more of the minimal runtime and normal runtime initialization code.

brendandahl:
This is a first step in trying to unify more of the code used by minimal runtime and the normal runtime and also move all addOn* functions into a library.

First, make the call to the static constructors use the same code for both modes. With this change, it becomes necessary to have an injection point that runs before ctors and after ctors. To achieve that, I've replaced the unused ATMAIN with new ATPOSTINIT code injection point. The new ATPOSTINIT is similar to the old ATMAIN in that it runs after the static constructors, but it now runs even if there is no main function.

Second, the FS initialization is now split into two parts (similar to how it used to be before a refactoring).
1) atInit the std<X> stream are setup. This allows ctors to use printf etc.
2) atPostInit the FS is locked down as late as possible but before main is run.

Third, switch embind code generation to use the POSTINIT so it can run after the ctors have setup the bindings.

Fourth, minimal runtime changes slightly in that ATINIT runs before ctors. This makes it consistent with the normal mode, but there is now POSTINIT if anyone needs the old behavior.


brendandahl:
This is a combination of my previous PR and Sam's PR to remove ATMAIN (#23478). Sam and I discussed that it is probably better to re-split the FS initialization that was done in a previous refactor (see comment above on this).

src/lib/libfs.js
```diff
@@ -36,8 +36,8 @@ addToLibrary({
     addAtInit(`
 if (!Module['noFSInit'] && !FS.initialized)
   FS.init();
-FS.ignorePermissions = false;
-`)
+`);
```
sbc100, line 39: Maybe just make this into a single line for consistency?

src/parseTools.mjs
```diff
@@ -731,14 +731,22 @@ function makeEval(code) {
...
+// Add code to run after the Wasm module is loaded, but before static
+// constructors and main (if applicable). The code will be executed before the
+// runtime `__ATINIT__` callbacks.
 function addAtInit(code) {
   ATINITS.push(code);
 }
 
+export const ATPOSTINITS = [];
+
+// Add code to run after static constructors, but before main (if applicable).
```
sbc100, line 745: This change looks great!   The only think i'm not sure about is the naming of this new list of actions.

Reading this comment made me think that since the primary defining factor seems to be related to static ctors, perhaps we should call it something with that in the name?  `ATCTORS` ?  `ATPOSTCTORS`?  I'm not sure I love either of those though.

src/parseTools_legacy.mjs
```diff
@@ -42,7 +42,7 @@ const Runtime = {
 const runOnMainThread = runIfMainThread;
 
 addToCompileTimeContext({
-  ATMAINS,
+  ATPOSTINITS,
```
sbc100, line 45: This can simply be removed since its already exported in parseTools.js

src/parseTools.mjs
```diff
@@ -1086,6 +1094,7 @@ function ENVIRONMENT_IS_WORKER_THREAD() {
 addToCompileTimeContext({
   ATEXITS,
   ATINITS,
+  ATPOSTINITS,
```
sbc100, line 1097: Can we just not export the array itself?

I'm not sure library code needs to access the ATPOSTINITS does it?

src/parseTools.mjs
brendandahl: I'm also not crazy about the name. I think I had `ATPOSTCTORS` in another version. That's probably the most clear.

sbc100 (approved):

sbc100 (approved):

PR #23494, opened 2025-01-24
[eslint] Enable unused argument linting within the JS compiler. NFC

sbc100:
Also, remove the explicit `'no-undef'` specifier since we now get that as part of `js.configs.recommended` above.

Also, remove `vars: 'all'` which is the default.

### DIFF
--- a/tools/acorn-optimizer.mjs
+++ b/tools/acorn-optimizer.mjs
@@ -289,7 +289,7 @@ function JSDCE(ast, aggressive) {
     }
     function cleanUp(ast, names) {
       recursiveWalk(ast, {
-        VariableDeclaration(node, c) {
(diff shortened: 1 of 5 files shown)


tools/acorn-optimizer.mjs
```diff
@@ -289,7 +289,7 @@ function JSDCE(ast, aggressive) {
     }
     function cleanUp(ast, names) {
       recursiveWalk(ast, {
-        VariableDeclaration(node, c) {
```
kripken, line 292: is this always good to do? if the caller provides the extra param, not having it might end up going through a slow path. i mean that matching the caller seems optimal, or at least that was the case years ago.

sbc100: OK, revert this part

PR #23496, opened 2025-01-25
Inline URIUtils.js. NFC

sbc100:
This is only three lines and doesn't warrant a separate file.

Prior to #23297 it did need to be separately usable.


src/URIUtils.js
```diff
@@ -4,17 +4,3 @@
  * SPDX-License-Identifier: MIT
  */
```
brendandahl, line 5: What's the reason for keeping the empty file?

sbc100: Oops, no!

brendandahl (approved):

PR #23502, opened 2025-01-26
Fix: Fix module not working on Deno

sevenc-nanashi:
The written .mjs file does not work on Deno, so I fixed that.
- Use `node:module` instead of `module`, so Deno can import that.
- Added polyfill of `Buffer`.

### DIFF
--- a/src/shell.js
+++ b/src/shell.js
@@ -103,9 +103,13 @@ if (ENVIRONMENT_IS_NODE) {
 #if EXPORT_ES6
   // When building an ES module `require` is not normally available.
   // We need to use `createRequire()` to construct the require()` function.
-  const { createRequire } = await import('module');
+  const { createRequire } = await import('node:module');
   /** @suppress{duplicate} */
   var require = createRequire('/');
+
+  if (typeof Deno !== "undefined") {
+    var Buffer = require('buffer').Buffer;
+  }
 #endif
 
 #if PTHREADS || WASM_WORKERS


sbc100:
We currently don't claim any deno support, but I suppose we could / should.   I guess that means we should add at least some basic testing under deno.

Do you know if this incompatibility is likely to persist in deno or might they be looking to increase node comparability?    Perhaps we could persuade them using this issue?  :)

kripken:
+1 for @sbc100 's hope that this can be fixed on the Deno side. But if that can't be done, or if we want something to work in the meantime, then I would recommend we look into the approach in https://github.com/emscripten-core/emscripten/issues/12203 , which I think will scale better if we want Deno, Bun, etc. etc. support. In that idea, each target, like Deno, would have a JS file to include which then patches things up for it. (If we can't literally patch in a polyfill manner, we could do something else, while keeping the goal of avoiding adding code for Deno in builds that don't need Deno.)

src/shell.js
```diff
@@ -103,9 +103,13 @@ if (ENVIRONMENT_IS_NODE) {
...
   // When building an ES module `require` is not normally available.
   // We need to use `createRequire()` to construct the require()` function.
-  const { createRequire } = await import('module');
+  const { createRequire } = await import('node:module');
   /** @suppress{duplicate} */
   var require = createRequire('/');
+
+  if (typeof Deno !== "undefined") {
+    var Buffer = require('buffer').Buffer;
+  }
```
kleisauke, line 112: This should no longer be necessary after Deno v2.4.0, see commit https://github.com/denoland/deno/commit/ed0c77edb3bce4047d0989cac089d6611bf8f351.

PR #23108, opened 2024-12-09
[AUDIO_WORKLET] Fix emscripten_audio_worklet_post_function_sig

mighty1231:
It fixes emscripten_audio_worklet_post_function_sig, which was previously not worked, and enhances test to cover all
emscripten_audio_worklet_post_function_*.

In order to fix, it changes the old term `readAsmConstArgs` to the new term `readEmAsmArgs`, due to conflict of https://github.com/emscripten-core/emscripten/pull/16449 and https://github.com/emscripten-core/emscripten/pull/18218

Background history:

1. #16449 proposed, introducing AudioWorklet earlier
2. #18218 proposed, renaming readAsmConstArgs to readEmAsmArgs
3. #18218 merged earlier
4. #16449 merged later, using the old term `readAsmConstArgs`

PR #23508, opened 2025-01-27
[AUDIO_WORKLET] Added support for MEMORY64 and 2GB heap (including tests)

cwoffenden:
This is built on #22753 and uses the `_2gb` and `_4gb` tests (plus the parameter test) from #23659.

It adds support for `MEMORY64` to audio worklets and tests that:

- All the pointer offsets and maths now work with >2GB heap
- Everything works building with `MEMORY64` and >4GB heap

The tests already prove out various buffer combinations using the `INITIAL_MEMORY` and `GLOBAL_BASE` flags.

For the code changes, besides parameters expecting `BigInt` it also needed any shifts removing from the offset calculations (signed shifts fail >2GB, unsigned >4GB).

The interactive tests (from #23659) can be run with:

```
test/runner.py interactive_2gb.test_audio_worklet_params_mixing
test/runner.py interactive64_4gb.test_audio_worklet_params_mixing
```


test/webaudio/audioworklet_test_shared.inc
```diff
@@ -83,13 +93,42 @@ void initialised(EMSCRIPTEN_WEBAUDIO_T context, bool success, void* data) {
 }
 
 // Common entry point for the mixer tests
-int main() {
-  static char workletStack[AUDIO_STACK_SIZE];
+// stackSize - audio worklet stack size in bytes
+// return true if the requested memory could be allocated
+bool _main_(int stackSize) {
```
sbc100, line 98: Can we just leave the this `int main() {` and use `AUDIO_STACK_SIZE` directly?

Having both `AUDIO_STACK_SIZE` in addition to `stackSize` seems like needless complexity.

Also having both a bool-returning and int-returning version of main seems needless different to other tests that just use a normal `int main`.

test/webaudio/audioworklet_test_shared.inc
```diff
@@ -83,13 +93,42 @@ void initialised(EMSCRIPTEN_WEBAUDIO_T context, bool success, void* data) {
...
+  if (emptySpace) {
+    printf("Empty space allocated (%dGB)\n", TEST_OFFSET_GB);
+  } else {
+    printf("Failed to allocate the required memory offset\n");
+    return false;
+  }
+#endif
+
+  char* const workletStack = memalign(16, stackSize);
+  if (!workletStack) {
```
sbc100, line 111: Same here

test/webaudio/audioworklet_test_shared.inc
```diff
@@ -83,13 +93,42 @@ void initialised(EMSCRIPTEN_WEBAUDIO_T context, bool success, void* data) {
...
 // Common entry point for the mixer tests
-int main() {
-  static char workletStack[AUDIO_STACK_SIZE];
+// stackSize - audio worklet stack size in bytes
+// return true if the requested memory could be allocated
+bool _main_(int stackSize) {
+  // Optional empty space before the audio worklet's stack
+#if TEST_OFFSET_GB > 0
+  char* const emptySpace = malloc(1073741824L * TEST_OFFSET_GB);
+  if (emptySpace) {
```
sbc100, line 102: Just use `assert(emptySpace)` here? The errors produced by assert are nice because they include the exact line number.

test/test_interactive.py
```diff
@@ -334,6 +334,20 @@ def test_audio_worklet_2x_hard_pan_io(self):
     shutil.copy(test_file('webaudio/audio_files/emscripten-bass-mono.mp3'), 'audio_files/')
     self.btest_exit('webaudio/audioworklet_2x_in_hard_pan.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS'])
 
+  # Same as 'test_audio_worklet_2x_stereo_io' but as wasm64 with the worklet stack starting 2GB in (tests for unsigned shifts in the memory offsets)
+  def test_audio_worklet_stereo_io_2gb(self):
+    os.mkdir('audio_files')
+    shutil.copy(test_file('webaudio/audio_files/emscripten-beat.mp3'), 'audio_files/')
+    shutil.copy(test_file('webaudio/audio_files/emscripten-bass.mp3'), 'audio_files/')
+    self.btest_exit('webaudio/audioworklet_in_out_stereo.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-sMEMORY64', '-sINITIAL_MEMORY=2164260864', '-DTEST_OFFSET_GB=2'])
```
sbc100, line 342: The `_2gb` test are normally testing the high 2gb of wasm32 mode.   In this mode we need to use `>> 0` rather than `>>> 0` for example.

test/test_interactive.py
```diff
@@ -334,6 +334,20 @@ def test_audio_worklet_2x_hard_pan_io(self):
...
+
+  # Same as 'test_audio_worklet_stereo_io_2gb' but with a 4GB offset
+  def test_audio_worklet_stereo_io_4gb(self):
+    os.mkdir('audio_files')
+    shutil.copy(test_file('webaudio/audio_files/emscripten-beat.mp3'), 'audio_files/')
+    shutil.copy(test_file('webaudio/audio_files/emscripten-bass.mp3'), 'audio_files/')
+    self.btest_exit('webaudio/audioworklet_in_out_stereo.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-sMEMORY64', '-sINITIAL_MEMORY=4311744512', '-DTEST_OFFSET_GB=4'])
```
sbc100, line 349: Can you use the same number we use in test_browser.py when defining `browser_2gb` and `browser64_4gb`?

i.e 2200mb or 4200mb for INITIAL_MEMORY and 2gb or 4gb for GLOBAL_BASE.

My setting GLOBAL_BASE that ensure that all stack and heap allocations are above that number, so you don't need to the test to be aware of any offset.   i.e. you can completely remove the TEST_OFFSET_GB thing I think

src/lib/libwebaudio.js
```diff
@@ -294,8 +300,8 @@ let LibraryWebAudio = {
       numberOfOutputs: HEAP32[options+1],
       outputChannelCount: HEAPU32[options+2] ? readChannelCountArray(HEAPU32[options+2]>>2, HEAP32[options+1]) : void 0,
       processorOptions: {
-        'cb': callback,
-        'ud': userData,
+        'cb': {{{ toIndexType('callback') }}},
+        'ud': {{{ toIndexType('userData') }}},
```
sbc100, line 304: And maybe do this at the in audio_worklet when `this.userData` is assigned?

src/lib/libwebaudio.js
```diff
@@ -294,8 +300,8 @@ let LibraryWebAudio = {
       numberOfOutputs: HEAP32[options+1],
       outputChannelCount: HEAPU32[options+2] ? readChannelCountArray(HEAPU32[options+2]>>2, HEAP32[options+1]) : void 0,
       processorOptions: {
-        'cb': callback,
-        'ud': userData,
+        'cb': {{{ toIndexType('callback') }}},
```
sbc100, line 303: Can you do this instead in audio_worklet at the point of ` Module['wasmTable'].get`?

src/lib/libwebaudio.js
cwoffenden: I did originally but it needed doing in multiple places. Let me take another look (since I got the hang of the changes required).

test/test_interactive.py
cwoffenden: Will do (which is better than what I was doing since I'm hoping it will break more things I spotted).

sbc100:
It looks like we already have `interactive64` mode in `test_interactive.py`.  Can you just add `interactive64_4gb` and `interactive_2gb` modes?  (copy them from test_browser.py).

test/webaudio/audioworklet_test_shared.inc
cwoffenden: Makes sense.

test/test_interactive.py
cwoffenden: I'll add a separate wasm32 test (where `GLOBAL_BASE` won't cause `malloc` issues with trying to claim 2GB).

cwoffenden:
> What about the exsiting audio worklet tests in test_browser.py?

I'll take a look tomorrow. The problem I found with these tests, and why I wrote the interactive ones, is that they were silently failing. For example, without this PR's changes it was already passing for wasm64 but should have clearly failed.

sbc100:
> > What about the exsiting audio worklet tests in test_browser.py?
> 
> I'll take a look tomorrow. The problem I found with these tests, and why I wrote the interactive ones, is that they were silently failing. For example, without this PR's changes it was already passing for wasm64 but should have clearly failed.

Fair enough.  I'm ok with adding audio tests to test interactive because we don't/cannot run audio tests in our CI anyway (since they require user interaction).

However, I think fixing the audio worklet tests so they don't fail silently even should really be a priority before we start any new work.

cwoffenden:
> However, I think fixing the audio worklet tests so they don't fail silently even should really be a priority before we start any new work.

I mostly agree, it's just that we're shipping this audio work which is how I can justify the time on it (`MEMORY64` wasn't really practical, but I wanted to understand it before moving onto the in-line implementation and removing the `*.aw.js` file, which is where I want to ultimately end up). My list is: wasm64, tests for audio params (none exist), in-line worklet.

I made notes as to why it's silently failing, but off the top of my head it's something to do with `btest` and `btest_exit` working differently. I will take a look.

src/lib/libwebaudio.js
sbc100: The idea is that we want to pass pointers around as just number and only convert to/from BigInt values when, fir example, directly accessing the table

cwoffenden: Understood, I'll make the change. I've a bunch of changes incoming since adding `GLOBAL_BASE=2gb` (the existing tests weren't catching the failures).

test/test_interactive.py
cwoffenden: I made the 2GB interactive test run with wasm32, so now tests are for `>>`, `>>>` and `/ 4` via `getHeapOffset()`.

cwoffenden:
> What about the exsiting audio worklet tests in test_browser.py?

I'll try to take a look in the next days, they're definitely all failing because even 2GB tests should fail. One of a couple of examples here:

https://github.com/emscripten-core/emscripten/blob/62304e29998a1e51e9d435ed229982ca46f15f0c/src/lib/libwebaudio.js#L244

Then for wasm64 the offset fails, e.g.:

https://github.com/emscripten-core/emscripten/blob/62304e29998a1e51e9d435ed229982ca46f15f0c/src/lib/libwebaudio.js#L246

The offset should be `+2` since the entry before is a pointer.

A simple standalone test with just the noise generator is:
```
emcc -sAUDIO_WORKLET -sWASM_WORKERS -sASSERTIONS=2 -sWEBAUDIO_DEBUG -sINITIAL_MEMORY=2200mb -sGLOBAL_BASE=2gb -o index.html audioworklet.c
```

With this PR's changes so far it now works, but I've more to do so this is still draft.

cwoffenden:
I wrote a test for the web audio parameters, creating a pair of descriptors and using the first to handle fading. It needs some tidying up but the important parts are done and it correctly gets the C structs to JS. Demo here:

https://wip.numfum.com/cw/2025-02-03/index.html

It only properly works in Chrome and Firefox so far (I'll look at it).

This single line does the fade-in/out:

https://github.com/emscripten-core/emscripten/blob/48361fd13accb0ba45943c1a195e2417a1788bf3/test/webaudio/audioworklet_params_mixing.c#L59

And the mix is currently done with this:

https://github.com/emscripten-core/emscripten/blob/48361fd13accb0ba45943c1a195e2417a1788bf3/test/webaudio/audioworklet_params_mixing.c#L42

As before, @sbc100, shall I create a new PR just for these tests? Then another PR that fixes running the tests?

cwoffenden:
> What about the exsiting audio worklet tests in test_browser.py?

All the AW tests are silently failing. I'll look at fixing them next.

cwoffenden:
> It looks like we already have `interactive64` mode in `test_interactive.py`. Can you just add `interactive64_4gb` and `interactive_2gb` modes? (copy them from test_browser.py).

Done.

cwoffenden:
Note for me to look at (`POINTER_TYPE` and `POINTER_SIZE`):

pointer_size

src/audio_worklet.js
```diff
@@ -51,75 +84,122 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+        }
+        // And that the views' size match the passed in output buffers
+        for (i of outputList) {
+          for (j of i) {
+            console.assert(j.byteLength == this.bytesPerChannel, `AudioWorklet unexpected output buffer size (expected ${this.bytesPerChannel} got ${j.byteLength})`);
+          }
+        }
+      }
+#endif
+
```
sbc100, line 193: This file sseems to have changed a lot more than I would expect just for the memory64 change.  Am I missing something?

src/audio_worklet.js
```diff
@@ -193,14 +273,9 @@ class BootstrapMessages extends AudioWorkletProcessor {
         // 'cb' the callback function
         // 'ch' the context handle
         // 'ud' the passed user data
-        p.postMessage({'_wsc': d['cb'], 'x': [d['ch'], 1/*EM_TRUE*/, d['ud']] });
+        p.postMessage({'_wsc': {{{ toIndexType("d['cb']") }}}, 'x': [d['ch'], 1/*EM_TRUE*/, {{{ toIndexType("d['ud']") }}}] });
```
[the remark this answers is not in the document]
cwoffenden: Originally the type conversions were in the calling code, here for example:

https://github.com/emscripten-core/emscripten/blob/0c3df923de46dbb1a98528c4f87e88d56af0d89d/src/lib/libwebaudio.js#L269

But you suggested moving them to where they're needed. Without these it results in type errors, e.g.: `TypeError: Cannot convert 0 to a BigInt`.

src/audio_worklet.js
cwoffenden: It's built on #22753 from last October that was never merged, which contained, besides the performance improvements, the groundwork for the struct offsets which made 2GB and wasm64 support straightforward. The diff between the old PR and this isn't that large, with most of the work going into the tests (which went into their own #23659).

cwoffenden:
Something looks to have broken in the many merges, it was all green ticks a few weeks back. Hmm.

cwoffenden:
Note for me to come back to. These test look to have broken when merged with #23108:
```
browser64 & browser64_4gb
test_audio_worklet_es6
test_audio_worklet_modularize
test_audio_worklet_pthreads_es6
```
They break here with the bigint error:

https://github.com/emscripten-core/emscripten/blob/2e14c6db4e002d0628281479f0e7d107e88fb8e6/src/preamble.js#L477

Called from here:

https://github.com/emscripten-core/emscripten/blob/2e14c6db4e002d0628281479f0e7d107e88fb8e6/src/lib/libwasm_worker.js#L101

Probably the switch to `readEmAsmArgs()`.

The `test_stat_chmod_nodefs` failure looks unrelated.

src/audio_worklet.js
cwoffenden: I'll break this down into a series of smaller PRs on the current main, e.g. move to the `C_STRUCTS` offsets and `makeGetValue`, then introduce wasm64, then the performance changes. I couldn't review in its current state what I wrote myself.

I don't know when this will be though since we're shipping it and it's done what I set out to do (we have millions of users putting in hour long sessions per day, so a few millis per second saved is a big deal on low-end school hardware when juggling 3D content with audio, and if it didn't work we'd really know about it!).

sbc100: We appreciate all the work you have done here. 

I'm excited to see these changes lands.  If you can find time to spit split the PR that would be amazing.  If not, we can circle back in a few weeks.

cwoffenden:
Closing in favour of simplified #24732

PR #23515, opened 2025-01-28
Simplify linker flag handling. NFC

sbc100:
This change simplifies the interface between emcc.py (the compiler driver) and link.py (the linker).

Linker flags are no longer stored in the EmccState or stored alongside their ordinal number. Instead passed directly the link.py as a simple list.

This change brings us closer to de-coupling the linker process with the hope of creating a standalone emld entry point.

### DIFF
--- a/emcc.py
+++ b/emcc.py
@@ -108,31 +108,24 @@ class Mode(Enum):
   COMPILE_AND_LINK = auto()
 
 
+class LinkFlag:
+  """Used to represent a linker flag.
+
+  The flag value is stored along with a bool that distingingishes input
+  files from non-files.
+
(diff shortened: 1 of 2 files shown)


emcc.py
```diff
@@ -747,26 +742,21 @@ def phase_parse_arguments(state):
...
   """
 
   if settings.RUNTIME_LINKED_LIBS:
     newargs += settings.RUNTIME_LINKED_LIBS
 
-  input_files = []
   compiler_args = []
+  linker_args = []
+
+  def add_link_arg(flag, isFile=False):
```
kripken, line 758: `isFile => is_file`?

kripken (approved):
lgtm % style nit

Issue #23503, opened 2025-01-27
Regression in "require" in ES6 modules. Fails to find "ws" module.


RamiHg:
It looks like a [recent change](https://github.com/emscripten-core/emscripten/commit/e352c4e4902cf5faccbc4d4d2b00560f5cc5eafa) to revert a commit that removed uses of `require` didn't quite fully revert to the old behavior.

I was getting a bizarre error where Node would refuse to find the "ws" module even though I had installed it using `npm install ws`. I build my code using `-sMODULARIZE -sEXPORT_ES6`. The error was:

```
Error: Cannot find module 'ws'
Require stack:
- /noop.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1060:15)
    at Module._load (node:internal/modules/cjs/loader:905:27)
    at Module.require (node:internal/modules/cjs/loader:1127:19)
    at require (node:internal/modules/helpers:112:18)
```

"ws" was getting imported from libsockfs.js in the `listen` function:

```js
        var WebSocketServer = require('ws').Server;
        var host = sock.saddr;
```

The [original code](https://github.com/emscripten-core/emscripten/pull/23169/) that was initially reverted called `createRequire` like this:
```js
  const { createRequire } = await import('module');
  let dirname = import.meta.url;
  if (dirname.startsWith("data:")) {
    dirname = '/';
  }
  /** @suppress{duplicate} */
  var require = createRequire(dirname);
```

However, the change that re-introduced the old code instead introduced it as:
```js
  const { createRequire } = await import('module');
  /** @suppress{duplicate} */
  var require = createRequire('/');
```

I'm not very familiar with Javascript and Node, but after much head-scratching, I tracked it down to that piece of code. If I replace the `createRequire` with the original version (the one that first checks for "data:"), my code works!

I can create a simple PR to do that fix if needed.

Thanks!

sbc100:
Interesting.  I assumed that the parameter to `createRequire` was only needed for relative imports like `./foo` not for system imports like `ws`.  I guess I was wrong.   I guess we need some more testing here.

sbc100:
Just to confirm, are you trying to actually build websocket server using emscripten that runs in node?   Or is this code just getting included incidentally somehow?

RamiHg:
Thanks for the quick reply! And yes. I am building a websocket server using emscripten. (Well, I'm porting a POSIX socket server).

sbc100:
[...]

PR #23516, opened 2025-01-28
Fix createRequire so it works for local `node_modules`.

sbc100:
For most uses of `require()` in emscripten this does not matter since we load mostly system libraries.  However for the `ws` module it is needed.  This bug was being masked by the fact that we were setting `NODE_PATH` in our socket test running.  This is no longer needed now that we run (non-parallel) tests in the emscripten tree (in out/test).

This is followup to #23265 which itself was an attempt to revert #23169.

Fixes: #23503


src/shell.js
```diff
@@ -105,7 +105,7 @@ if (ENVIRONMENT_IS_NODE) {
   // We need to use `createRequire()` to construct the require()` function.
   const { createRequire } = await import('module');
   /** @suppress{duplicate} */
-  var require = createRequire('/');
+  var require = createRequire(import.meta.url);
```
kripken, line 108: I don't know enough about support for this - didn't we have an issue with it breaking in some bundler or environment?

sbc100: I believe you are thinking of the use of `import.meta.url` of in `new Worker`.   This usage is perfectly fine AFAIK.  See #23169 which removed it prematurely.

kripken (approved):

PR #23517, opened 2025-01-28
Fix proxy-to-worker + modularize error reporting

sbc100:
This configuration is not supported by was going undetected when `-sEXPORT_ES6` was used due to the order to checking vs setting the `MODULARIZE` setting.

This change doesn't add or remove any checks, it only changes the ordering so we don't check settings before they are set.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -915,6 +915,12 @@ def phase_linker_setup(options, state):  # noqa: C901, PLR0912, PLR0915
     # When using MINIMAL_RUNTIME, symbols should only be exported if requested.
     default_setting('EXPORT_KEEPALIVE', 0)
 
+  if settings.EXPORT_ES6 and not settings.MODULARIZE:
+    # EXPORT_ES6 requires output to be a module
+    if 'MODULARIZE' in user_settings:
+      exit_with_error('EXPORT_ES6 requires MODULARIZE to be set')
+    settings.MODULARIZE = 1
+
   if settings.STRICT_JS and (settings.MODULARIZE or settings.EXPORT_ES6):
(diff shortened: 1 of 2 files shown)


tools/link.py
```diff
@@ -915,8 +915,20 @@ def phase_linker_setup(options, state):  # noqa: C901, PLR0912, PLR0915
...
+        exit_with_error('EXPORT_ES6 requires MODULARIZE to be set')
+      settings.MODULARIZE = 1
+    if settings.ENVIRONMENT_MAY_BE_NODE and not settings.USE_ES6_IMPORT_META:
+      # EXPORT_ES6 + ENVIRONMENT=*node* requires the use of import.meta.url
+      if 'USE_ES6_IMPORT_META' in user_settings:
+        exit_with_error('EXPORT_ES6 and ENVIRONMENT=*node* requires USE_ES6_IMPORT_META to be set')
+      settings.USE_ES6_IMPORT_META = 1
+
+  if settings.STRICT_JS and settings.MODULARIZE:
+    exit_with_error("STRICT_JS doesn't work with MODULARIZE")
```
kripken, line 931: I don't see any `proxy-to-worker` code changed. Is it connected to STRICT_JS somehow?

sbc100: I didn't change this check, only removed the redundant `(settings.MODULARIZE or settings.EXPORT_ES6)`.  The later implies the former so this is the same thing.   I can revert these lines to make this change more clear maybe.

sbc100: Done

sbc100:
Note: This change doesn't add or remove any checks, it only changes the ordering so that we don't check the `MODULARIZE` settings before we set it.=

kripken (approved):

PR #23518, opened 2025-01-28
Fix re-creating webgl context for offscreen canvas

JoeOsborn:
This adds a test and a one-line fix to resolve a bug.  After a canvas has been transferred offscreen, it can't be used again to create a webgl context because the result of findCanvasEventTarget is a {canvas:...} map instead of an actual canvas object.  This patch grabs the underlying offscreencanvas if it exists.

### DIFF
--- a/src/lib/libhtml5_webgl.js
+++ b/src/lib/libhtml5_webgl.js
@@ -109,7 +109,10 @@ var LibraryHtml5WebGL = {
 #endif
 
     var canvas = findCanvasEventTarget(target);
+#if OFFSCREENCANVAS_SUPPORT
+    // If our canvas from findCanvasEventTarget is actually an offscreen canvas record, we should extract the inner canvas.
     if (canvas?.canvas) { canvas = canvas.canvas; }
+#endif
 #if GL_DEBUG
     var targetStr = UTF8ToString(target);
 #endif
--- a/src/lib/libhtml5_webgl.js
+++ b/src/lib/libhtml5_webgl.js
@@ -109,7 +109,7 @@ var LibraryHtml5WebGL = {
 #endif
 
     var canvas = findCanvasEventTarget(target);
-
+    if (canvas.canvas) { canvas = canvas.canvas; }
 #if GL_DEBUG
     var targetStr = UTF8ToString(target);
 #endif
--- a/src/lib/libhtml5_webgl.js
+++ b/src/lib/libhtml5_webgl.js
@@ -109,7 +109,7 @@ var LibraryHtml5WebGL = {
 #endif
 
     var canvas = findCanvasEventTarget(target);
-    if (canvas.canvas) { canvas = canvas.canvas; }
+    if (canvas?.canvas) { canvas = canvas.canvas; }
 #if GL_DEBUG
     var targetStr = UTF8ToString(target);
 #endif
--- a/test/browser/webgl_create_context_swapcontrol.c
+++ b/test/browser/webgl_create_context_swapcontrol.c
@@ -23,5 +23,13 @@ int main() {
   EMSCRIPTEN_RESULT res = emscripten_webgl_make_context_current(context);
   assert(res == EMSCRIPTEN_RESULT_SUCCESS);
   assert(emscripten_webgl_get_current_context() == context);
+  emscripten_webgl_destroy_context(context);
+  // Test that creating a context for a canvas that's already been used once succeeds
+  context = emscripten_webgl_create_context("#canvas", &attrs);
+  assert(context > 0); // Must have received a valid context.
+  res = emscripten_webgl_make_context_current(context);
+  assert(res == EMSCRIPTEN_RESULT_SUCCESS);
+  assert(emscripten_webgl_get_current_context() == context);
+  emscripten_webgl_destroy_context(context);
   return 0;
 }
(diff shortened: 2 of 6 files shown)


test/browser/webgl_create_context_swapcontrol.c
```diff
@@ -23,5 +23,12 @@ int main() {
   EMSCRIPTEN_RESULT res = emscripten_webgl_make_context_current(context);
   assert(res == EMSCRIPTEN_RESULT_SUCCESS);
   assert(emscripten_webgl_get_current_context() == context);
+  emscripten_webgl_destroy_context(context);
+  // Test that creating a context for a canvas that's already been used once succeeds
+  context = emscripten_webgl_create_context("#canvas", &attrs);
+  assert(context > 0); // Must have received a valid context.
+  res = emscripten_webgl_make_context_current(context);
+  assert(res == EMSCRIPTEN_RESULT_SUCCESS);
+  assert(emscripten_webgl_get_current_context() == context);
```
sbc100, line 32: Should we `emscripten_webgl_destroy_context` here at the end?

test/browser/webgl_create_context_swapcontrol.c
```diff
@@ -23,5 +23,12 @@ int main() {
   EMSCRIPTEN_RESULT res = emscripten_webgl_make_context_current(context);
   assert(res == EMSCRIPTEN_RESULT_SUCCESS);
   assert(emscripten_webgl_get_current_context() == context);
+  emscripten_webgl_destroy_context(context);
+  // Test that creating a context for a canvas that's already been used once succeeds
+  context = emscripten_webgl_create_context("#canvas", &attrs);
+  assert(context > 0); // Must have received a valid context.
+  res = emscripten_webgl_make_context_current(context);
```
sbc100, line 30: Which variant of this test fails without the fix?   Is there some kind of thread involved here or are we just rendering to and offscreen buffer on the main thread?

sbc100 (approved):
Are there other places in the codebase that do the same unpacking?  It seems a little odd to have `findCanvasEventTarget` return such disparate types.. but I guess that is the status quo.

test/browser/webgl_create_context_swapcontrol.c
JoeOsborn: This error shows up even without threads, so I thought using the existing webgl explicit swap control test would be simplest.

test/browser/webgl_create_context_swapcontrol.c
JoeOsborn: It wouldn't hurt, I'll add it.

test/browser/webgl_create_context_swapcontrol.c
sbc100: Presumably the issue only shows up when `USE_OFFSCREEN_FRAMEBUFFE` is used above?

JoeOsborn: It's true with and without offscreen framebuffer, it really is triggered by the transfer of the offscreen canvas.

sbc100: I see, but there in this test is offscreencanvas used then?

JoeOsborn: It's used because OFFSCREENCANVAS_SUPPORT is 1 in the build settings and explicitSwapControl is true in the context.

src/lib/libhtml5_webgl.js
```diff
@@ -109,7 +109,7 @@ var LibraryHtml5WebGL = {
 #endif
 
     var canvas = findCanvasEventTarget(target);
-
+    if (canvas.canvas) { canvas = canvas.canvas; }
```
sbc100, line 112: Looks like you canvas can be null here?   How about `canvas?.canvas`?

JoeOsborn: Yes, I think the condition in the if also must be guarded. `if (canvas && canvas.canvas) { canvas = canvas.canvas }`?

src/lib/libhtml5_webgl.js
```diff
@@ -109,7 +109,7 @@ var LibraryHtml5WebGL = {
 #endif
 
     var canvas = findCanvasEventTarget(target);
-    if (canvas.canvas) { canvas = canvas.canvas; }
+    if (canvas && canvas.canvas) { canvas = canvas.canvas; }
```
sbc100, line 112: I think you can write `canvas?.canvas` instead of `canvas && canvas.canvas`?

JoeOsborn: Yes, that's true. I'll fix it.

JoeOsborn:
I think this CI issue is just a code size change?

src/lib/libhtml5_webgl.js
```diff
@@ -109,7 +109,7 @@ var LibraryHtml5WebGL = {
 #endif
 
     var canvas = findCanvasEventTarget(target);
-
+    if (canvas?.canvas) { canvas = canvas.canvas; }
```
kripken, line 112: This is worth a comment I think.

And, if this is for offscreen canvas (as the PR title suggests) then can it be behind an ifdef, so we do not increase code size for most users?

PR #23519, opened 2025-01-28
Remove the need for jscompiler to run with CWD=emscripten/src. NFC

sbc100:
The only place that CWD was relied upon here was the use of the `read()`
function in some JS library code that was expecting to be able to read
files relative to the `src` directory.  To deal with this use case I
created a new `readFile` function for compiler usage and modified the
exported `read` function to explicitly support `src/` relative files.

This change makes the existing `EMCC_BUILD_DIR` environment variable
obsolete since it was only needed because we were modifying CWD when
running the JS compiler.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -4961,14 +4961,15 @@ extraLibraryFuncs.push('jsfunc');
     self.do_runf(test_file('hello_world.c'), emcc_args=['-L', '-lfoo.js'])
 
   def test_EMCC_BUILD_DIR(self):
-    # EMCC_BUILD_DIR env var contains the dir we were building in, when running the js compiler (e.g. when
-    # running a js library). We force the cwd to be src/ for technical reasons, so this lets you find out
-    # where you were.
+    # EMCC_BUILD_DIR was necessary in the past since we used to force the cwd to be src/ for
+    # technical reasons.
     create_file('lib.js', r'''
-printErr('dir was ' + process.env.EMCC_BUILD_DIR);
+printErr('EMCC_BUILD_DIR: ' + process.env.EMCC_BUILD_DIR);
+printErr('CWD: ' + process.cwd());
 ''')
     err = self.run_process([EMXX, test_file('hello_world.c'), '--js-library', 'lib.js'], stderr=PIPE).stderr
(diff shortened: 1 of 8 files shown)


kripken:
What is the motivation here?

sbc100:
I'm been looking to compiler.mjs performance, and as part of profiling it I'd like to be able to easily run it from the command line.  There are few things that make it annoying / difficult to run it isolation today.   One of them is that your CWD has to be `src/`.. which just seem unnecessary.

I want to be able run `time src/compiler.mgs <my_settings_file>` to time / profile it.

With out this change I would need to do `(cd src && ./compiler.mjs <my_settings_file>)`.. not a huge deal, just typing I don't want to do.

Also, it the entire rest of the toolchain avoids `chdir` which means it runs with the invocation directory intact.    I think its a nice invariant in a compile that CWD is unchanged to you get use `os.getcwd()` to see where the compiler was run from.

sbc100:
Updated the PR description with some more rationale.

tools/emscripten.py
```diff
@@ -183,14 +183,15 @@ def compile_javascript(symbols_only=False):
       json.dump(settings.external_dict(), s, sort_keys=True, indent=2)
 
     # Call js compiler
+    # TODO(sbc): Remove EMCC_BUILD_DIR since its no longer needed.
```
kripken, line 186:
```suggestion
    # TODO(sbc): Remove EMCC_BUILD_DIR since it's no longer needed.
```

test/test_other.py
```diff
@@ -4961,14 +4961,15 @@ def test_jslib_has_library(self):
     self.do_runf(test_file('hello_world.c'), emcc_args=['-L', '-lfoo.js'])
 
   def test_EMCC_BUILD_DIR(self):
-    # EMCC_BUILD_DIR env var contains the dir we were building in, when running the js compiler (e.g. when
-    # running a js library). We force the cwd to be src/ for technical reasons, so this lets you find out
-    # where you were.
+    # EMCC_BUILD_DIR was necessary in the part since we used to force the cwd to be src/ for
```
kripken, line 4964:
```suggestion
    # EMCC_BUILD_DIR was necessary in the past since we used to force the cwd to be src/ for
```

kripken (approved):

PR #23521, opened 2025-01-28
Make more use of optional chaining. NFC

sbc100:
I found all these sites using:

```
$ git grep -n "if (\(.*\) && \1\>.*)" src/
```


sbc100:
Reverted the non-NFC part

dschuff (approved):

src/lib/libwebgl.js
```diff
@@ -1184,7 +1184,7 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
       }
       // Make sure the canvas object no longer refers to the context object so
       // there are no GC surprises.
-      if (GL.contexts[contextHandle] && GL.contexts[contextHandle].GLctx.canvas) {
+      if (GL.contexts[contextHandle]?.GLctx?.canvas) {
```
kripken, line 1187: The second `?` was not needed before so I assume it isn't after either?

PR #23522, opened 2025-01-29
Support modularize instance exports in the acorn optimizer.

brendandahl:
Helps identify exports in the style of:

`var foo = __exp__foo = wasmExports['_foo']`

### DIFF
--- a/test/js_optimizer/emitDCEGraph-output.js
+++ b/test/js_optimizer/emitDCEGraph-output.js
@@ -110,6 +110,17 @@
   "reaches": []
  },
  {
+  "name": "emcc$export$expESM1",
+  "export": "expESM1",
+  "reaches": []
+ },
+ {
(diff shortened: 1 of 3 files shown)


sbc100 (approved):

sbc100:
How did this work before this change?

brendandahl:
I don't think it did, since we don't have an end-to-end test of ensuring this works. How do we ensure changes to the JS output won't break metadce, just code size tests?

sbc100:
Oh I see, the side effect is just that fewer things can be DCE'd?  In that case it makes sense that we wouldn't see failures today.

kripken (commented):
Yes, only code size tests can catch a missed optimization like this. Do we have a size test for modularize instance? If not let's add one maybe. lgtm otherwise.

PR #23524, opened 2025-01-29
[embind] Stricter function argument name checking

sbc100:
Distinguish between unknown argument list and the empty argument list when parsing function names.

Also use early return to make the code a little more compact.

### DIFF
--- a/src/lib/libembind_gen.js
+++ b/src/lib/libembind_gen.js
@@ -499,13 +499,13 @@ var LibraryEmbind = {
         }
         argStart = 2;
       }
-      if (argsName.length && argsName.length != (argTypes.length - hasThis - 1)) {
+      if (argsName && argsName.length != (argTypes.length - hasThis - 1)) {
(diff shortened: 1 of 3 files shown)


src/lib/libembind_gen.js
```diff
@@ -499,13 +499,13 @@ var LibraryEmbind = {
...
       }
-      if (argsName.length && argsName.length != (argTypes.length - hasThis - 1)) {
+      if (argsName && argsName.length != (argTypes.length - hasThis - 1)) {
         throw new Error('Argument names should match number of parameters.');
       }
 
       const args = [];
       for (let i = argStart, x = 0; i < argTypes.length; i++) {
-        if (x < argsName.length) {
+        if (argsName) {
```
sbc100, line 508: We know that argsName is the correct length thanks to the Error above, right?

brendandahl (approved):

PR #23527, opened 2025-01-29
Avoid looking up GL methods by string name. NFC

sbc100:


src/lib/libwebgl.js
```diff
@@ -1820,15 +1820,15 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
   },
 
   glGenBuffers: (n, buffers) => {
-    GL.genObject(n, buffers, 'createBuffer', GL.buffers
+    GL.genObjects(n, buffers, GLctx.createBuffer, GL.buffers
```
kainino0x, line 1823: I believe this will pass `createBuffer` as an unbound function, so when it's called `this` will be `undefined` and you'll get a crash, like so:

image

kainino0x: You could do `() => GLctx.createBuffer()`
or `GLctx.createBuffer.bind(GLctx)`

sbc100: Fair enough.  I guess maybe I'll just abandon this then.

PR #23538, opened 2025-01-29
[test] Consistent use of cmake_minimum_required in test code. NFC

sbc100:
Lower values were causing cmake to print verbose warnings on each test run.

I chose 3.16 here since that is the version available in ubuntu/focal which is the oldest version we test against AFAICT.

### DIFF
--- a/test/minimal_webgl/CMakeLists.txt
+++ b/test/minimal_webgl/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.16)
 
 # Default to release build if not specified
 if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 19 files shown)


test/minimal_webgl/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.29)
```
kripken, line 1: why is this one 29?

sbc100: Oops, fixed

kripken (approved):

sbc100:
Downgraded to 3.16 since we still test on ubuntu focal.

kripken (approved):

PR #23539, opened 2025-01-29
[ci] Stop disabling SIMD tests under windows. NFC

sbc100:
These only require node 16, and emsdk ships with node 20 these days.

See https://chromium-review.googlesource.com/c/emscripten-releases/+/6216543


sbc100:
ping..

sbc100:
ping...

dschuff (approved):

PR #23540, opened 2025-01-30
Support closure in modularize instance mode.

### LINKED ISSUE #23283: -sMODULARIZE=instance does not work with closure compiler
This is because the module code that defines the `__exp_` variables is run through closure compiler but the code that uses this is not.

This is fixed by #23158

brendandahl:
ES module exports are added after closure is run, which leads to closure complaining about undefined variables. To get around this, add the exports to an extern file.

I've also changed MODULARIZE=instance to not require a `.mjs` output file. This allows it to be used in test_codesize_minimal without changes and some developers prefer not to use the `.mjs` extension, but use the setting in package.json.

Fixes #23283


tools/building.py
```diff
@@ -566,6 +566,14 @@ def closure_compiler(filename, advanced=True, extra_closure_args=None):
   if settings.MODULARIZE:
     CLOSURE_EXTERNS += [path_from_root('src/closure-externs/modularize-externs.js')]
 
+  # The ES module exports are generated after closure is run, so add the exports
+  # to and extern file to avoid closure complaining about undefined variables.
+  if settings.MODULARIZE == 'instance':
+    temp = shared.get_temp_files().get('.js', prefix='emcc_closure_externs_').name
+    exports = settings.EXPORTED_FUNCTIONS + settings.EXPORTED_RUNTIME_METHODS
+    utils.write_file(temp, ';\n'.join(['var' +  ' __exp_' + export for export in exports]) + ';\n')
```
sbc100, line 574: Can you follow the pattern used below for `emcc_module_exports`?   i.e. `.write(..)` and `.close()`?

tools/link.py
```diff
@@ -769,8 +769,7 @@ def phase_linker_setup(options, state):  # noqa: C901, PLR0912, PLR0915
 
   if settings.MODULARIZE == 'instance':
     diagnostics.warning('experimental', '-sMODULARIZE=instance is still experimental. Many features may not work or will change.')
-    if options.oformat != OFormat.MJS:
-      exit_with_error('emcc: MODULARIZE instance is only compatible with .mjs output files')
+    settings.EXPORT_ES6 = 1
```
sbc100, line 772: Isn't `oformat` set to `MJS` when ever `EXPORT_ES6` is set?

In any case I think this error still needs to be here, but maybe in form of :

```
if not settings.EXPORT_ES6:
  exit_with_error('emcc: MODULARIZE instance is only comatile with ES module output')
```

brendandahl: It's the opposite MJS sets EXPORT_ES6. You can still do EXPORT_ES6 with with a `.js` file.

I don't think we plan to support instance mode in non es6 mode, so maybe it's better to auto set the flag to not force the user to set both?

sbc100: I think for now I would rather force users to specify both flags.    Maybe one day we could change that when we decide to make this stuff non-experimental.

sbc100: Also, even if we did want to auto-set on setting based on another we still need to handle and catch the case where the command line is contradictory.

tools/building.py
```diff
@@ -563,6 +563,16 @@ def closure_compiler(filename, advanced=True, extra_closure_args=None):
   if settings.MODULARIZE:
     CLOSURE_EXTERNS += [path_from_root('src/closure-externs/modularize-externs.js')]
 
+  # The ES module exports are generated after closure is run, so add the exports
+  # to and extern file to avoid closure complaining about undefined variables.
```
kripken, line 567:
```suggestion
  # to an extern file to avoid closure complaining about undefined variables.
```

tools/building.py
```diff
@@ -563,6 +563,16 @@ def closure_compiler(filename, advanced=True, extra_closure_args=None):
...
+  if settings.MODULARIZE == 'instance':
+    exports = settings.EXPORTED_FUNCTIONS + settings.EXPORTED_RUNTIME_METHODS
```
kripken, line 569: Does the current test check a value from `EXPORTED_RUNTIME_METHODS`?

brendandahl:
This has been fixed elsewhere with the removal of the wrapper function.

Issue #15097, opened 2021-09-22
Upstream emscripten compiler-rt changes (In particular the new asan, lsan, platform support)


sbc100:

sbc100:
I think this is largely don't now isn't it @aheejin ?

aheejin:
I don't think so? At least I haven't upstreamed any compiler-rt changes myself. Running `grep` on the upstream LLVM's `compiler-rt` directory doesn't yield much. I'm not sure what those `LIBFUZZER_EMSCRIPTEN` things are but they are not our sanitizer changes.
```console
aheejin@aheejin:~/llvm-project/compiler-rt$ grep EMSCRIPTEN * -R
lib/orc/endianness.h:    defined(__Fuchsia__) || defined(__EMSCRIPTEN__)
lib/fuzzer/FuzzerDriver.cpp:#if !LIBFUZZER_EMSCRIPTEN
lib/fuzzer/FuzzerDriver.cpp:#endif // LIBFUZZER_EMSCRIPTEN
lib/fuzzer/FuzzerExtFunctionsWeak.cpp:    LIBFUZZER_FREEBSD || LIBFUZZER_EMSCRIPTEN
lib/fuzzer/FuzzerExtraCounters.cpp:    LIBFUZZER_FUCHSIA || LIBFUZZER_EMSCRIPTEN
lib/fuzzer/FuzzerUtilLinux.cpp:    LIBFUZZER_EMSCRIPTEN
lib/fuzzer/FuzzerPlatform.h:#define LIBFUZZER_EMSCRIPTEN 0
lib/fuzzer/FuzzerPlatform.h:#define LIBFUZZER_EMSCRIPTEN 0
lib/fuzzer/FuzzerPlatform.h:#define LIBFUZZER_EMSCRIPTEN 0
lib/fuzzer/FuzzerPlatform.h:#define LIBFUZZER_EMSCRIPTEN 0
lib/fuzzer/FuzzerPlatform.h:#define LIBFUZZER_EMSCRIPTEN 0
lib/fuzzer/FuzzerPlatform.h:#define LIBFUZZER_EMSCRIPTEN 0
lib/fuzzer/FuzzerPlatform.h:#elif __EMSCRIPTEN__
lib/fuzzer/FuzzerPlatform.h:#define LIBFUZZER_EMSCRIPTEN 1
[...]

PR #23544, opened 2025-01-30
[libc++abi] Add more checks to ensure our patch is upstreamable

georgthegreat:
We use the same libcxxabi for every platform and emscripten patches break compilation when targeting linux-armv7a

### DIFF
--- a/system/lib/libcxxabi/src/private_typeinfo.cpp
+++ b/system/lib/libcxxabi/src/private_typeinfo.cpp
@@ -1546,7 +1546,7 @@ __base_class_type_info::search_below_dst(__dynamic_cast_info* info,
 
 
 // XXX EMSCRIPTEN
-#ifndef __WASM_EXCEPTIONS__
+#if defined(__wasm__) && !defined(__WASM_EXCEPTIONS__)
 
 #include "cxa_exception.h"
 
@@ -1601,5 +1601,5 @@ void *__cxa_get_exception_ptr(void *thrown_object) throw() {
 
 }  // __cxxabiv1
 
-#endif // !__WASM_EXCEPTIONS__
+#endif // __wasm__ && !__WASM_EXCEPTIONS__
 


sbc100:
To be fair this is downstream patch, so we don't expect this code to build anywhere except emscripten.
 
On the other hand this patch seems perfectly reasonable.   Can you perhaps give a more descriptive title?  i.e. at least mention that its libc++abi releated?

georgthegreat:
I have updated the title.
Btw, do you have any plans to upstream patches to LLVM?
There is certain work regarding upstreaming (some patches have already been merged), but I do not quite understand the status.

sbc100:
Regarding the patches we maintain for the llvm stdlibs, there is some documentation here: https://github.com/emscripten-core/emscripten/blob/main/system/lib/libcxx/readme.txt.

Specifically you can see the changes we maintain here: https://github.com/llvm/llvm-project/compare/llvmorg-19.1.6...emscripten-core:emscripten-libs-19

We have made (specially @aheejin mostly) efforts to upstream as much as possible, but there may be more still.  There will also be some cases (like this one) that probably don't want to be upstream in their current form.

sbc100:
See https://github.com/emscripten-core/emscripten/issues/15097

georgthegreat:
> See #15097

This is about sanitizers.
This is definitely a good work to be done, though we do not support sanitizers for emscripten builds internally.

Any chance you do have such issue regarding libcxxabi?

sbc100:
> > See #15097
> 
> This is about sanitizers. This is definitely a good work to be done, though we do not support sanitizers for emscripten builds internally.
> 
> Any chance you do have such issue regarding libcxxabi?

I'm having trouble finding such a bug, but in general it is our goal to upstream where possible.   This particular file/change might not make sense outside of emscripten though so might not make sense to upstream.

sbc100:
I still think the title is a little misleading.. nobody should be trying to build our version of libc++abi for any target except emscripten, but IIUC we are not actually breaking anybody/anything there.

How about "[libc++abi] Add more checks to ensure our patch is upstreamable".

aheejin (approved):

PR #23548, opened 2025-01-30
[test] Remove require_simd and requires_wasm_legacy_eh helpers. NFC

sbc100:
These helpers were requiring node v16 and v17 which is older than the version of node we ship with emsdk these days (v20).

We do run node compat tests using older version of node but we run an explicit subset of tests in those modes, never the whole suite.


sbc100:
closing in favor of #23547

Issue #23523, opened 2025-01-29
HTML5: Are 'input' and 'change' event callbacks implemented?


alibabashack:
Hi, although the HTML5 API allows to subscribe to a lot of events, I don't see support for the [input](https://developer.mozilla.org/en-US/docs/Web/API/Element/input_event) and [change](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) events required in order to perform form validation tasks from wasm. I think these events are simply not implemented or am I mistaken?

Is there something complex about these event types? Otherwise I would like to try to implement them ...

sbc100:
Sounds like perhaps we just never had demand for those.  Feel free to send a PR.

curiousdannii:
I filed this earlier, including a suggestion of how to implement it: https://github.com/emscripten-core/emscripten/issues/21960

alibabashack:
Now, that @curiousdannii mentions specific Android behavior: Is there a quick way to let the test harness (`test/runner`) serve a test page as HTTP server to an external device (e.g. an Android phone)? Or is there a better approach to verify behavior on other devices? (I am not familiar with web development; embedded C++ guy here :)

Issue #21960, opened 2024-05-17
Improve support in SDL2 for keyboard input on mobile browsers


curiousdannii:
Mobile browsers (with virtual keyboards) don't support `keypress` events, and for `keydown` events [set the code to 229](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event) for most keys. But, there is pretty good support through the `input` event - at least it supports all actual textual input keys plus the backspace key (one of the few keys actually returned in `keydown` events). It would be good if Emscripten (and then SDL) supported the `input` event with little/no manual work by the user.

(I just discovered something new: the `input` event doesn't seem to fire if Chrome dev tools have breakpoints in the event handler. Makes it very hard to test and debug these events! This did not used to be the case.)

What this might involve:

1. Creating an input element, as the `input` event won't fire on document or window. This should be a `<textarea>` not an `<input>` because of [this other unfixed Chrome bug](https://stackoverflow.com/questions/72988591/prevent-android-keyboard-in-chrome-from-showing-password-management-on-html-inpu).
2. The `input` event needs to be handled differently, as it has different properties, or you can access what was typed via the textarea's `.value`. When I've used it (not in an Emscripten project) I also had to empty it each time. And if the character was space, blur then refocus, as otherwise it thinks there are still spaces and backspace doesn't work.
3. Adding something like `emscripten_set_keydown_callback_on_thread` and then using it in SDL. I don't know if it could be directly used or if it would need to be mapped into something more like a `keydown` event.
4. Maybe also ignore `keydown` events which have a code of 229 and don't send them to the wasm? That's what I've previously done, but it may not be necessary.

sbc100:
Do you know if SDL2 has the same issues?  I general I would expect SDL2 support to be more advanced than the SDL1/js code.

curiousdannii:
I was referring to SDL2, I've never tried SDL1.

sbc100:
Adding @Daft-Freak.

curiousdannii:
Well I've got my personal solution working, which can be seen here: https://curiousdannii.github.io/infocom-frotz/zorkzero.html

```html
<textarea id="textinput" autocapitalize="off" rows="1"></textarea>
```

```css
#textinput {
    position: absolute;
    left: -10000px;
}
```

```js
const textinput_elem = document.getElementById('textinput')
Module.canvas = document.getElementById('canvas')

// Mobile input event handlers
Module.canvas.addEventListener('touchstart', ev => {
[...]

PR #23550, opened 2025-01-31
HTML5 API: Add support for 'input' event

alibabashack:


alibabashack:
Implements the HTML DOM 'input' event as suggested in #23523.

In review **please consider** one problem: The EmscriptenInputEvent::data field is potentially a very long UTF8 string on the JS side, because the input event could be a paste of a long text. In my implementation this string is truncated to 256 chars, because I was not sure how to handle the JS to wasm handover otherwise. I think the truncation is ok for most use cases of the input event (e.g. form validation where I want to read the DOM value of the event target anyway). However, I am concerned about truncating a UTF8 string, because this could split a code point midway. I am sure there is some experience with this kind of problem in this code base. Any advice?

You may want to try the new interactive test case:
`test/runner interactive.test_html5_input`

alibabashack:
I think I need some help to interpret the CI result. I don't understand why `test_offset_converter_pthread` and `test_pthread_wait64_notify` tests fail. They seem to be unrelated to my work. Any help is much appreciated, thanks.

sbc100:
> I think I need some help to interpret the CI result. I don't understand why `test_offset_converter_pthread` and `test_pthread_wait64_notify` tests fail. They seem to be unrelated to my work. Any help is much appreciated, thanks.
> 
> **Edit**: Oh, I see, looks `main` is broken...

Those tests are known to be flaky.   We can re-run the CI and they should pass.

system/include/emscripten/html5.h
```diff
@@ -189,6 +190,15 @@ EMSCRIPTEN_RESULT emscripten_set_focus_callback_on_thread(const char *target __a
 EMSCRIPTEN_RESULT emscripten_set_focusin_callback_on_thread(const char *target __attribute__((nonnull)), void *userData, bool useCapture, em_focus_callback_func callback, pthread_t targetThread);
 EMSCRIPTEN_RESULT emscripten_set_focusout_callback_on_thread(const char *target __attribute__((nonnull)), void *userData, bool useCapture, em_focus_callback_func callback, pthread_t targetThread);
 
+typedef struct EmscriptenInputEvent {
+  EM_UTF8 data[EM_HTML5_LONG_STRING_LEN_BYTES];
```
sbc100, line 194: I think we should avoid truncation here and instead consider two possible options:

1. Use stack allocated space.   i.e. `stringToNewUTF8OnStack()` on the JS side.   This would means it would be to up the user to make a copy of the string if they want it to outlive the callback.
2. Use heap allocation.    i.e `stringToNewUTF8` on the JS side.  It would then be incumbent upon the receiver to call `free` on the resulting pointer when they are done with it.

I guess (2) seems safer since it avoids a possible stack overflow.

src/lib/libhtml5.js
```diff
@@ -805,6 +806,48 @@ var LibraryHTML5 = {
...
+        var inputEvent = JSEvents.inputEvent;
+        #endif
+        stringToUTF8(data, inputEvent + {{{ C_STRUCTS.EmscriptenInputEvent.data }}}, {{{ cDefs.EM_HTML5_LONG_STRING_LEN_BYTES }}});
+        stringToUTF8(inputType, inputEvent + {{{ C_STRUCTS.EmscriptenInputEvent.inputType }}}, {{{ cDefs.EM_HTML5_SHORT_STRING_LEN_BYTES }}});
+        HEAP8[inputEvent + {{{ C_STRUCTS.EmscriptenInputEvent.isComposing }}}] = e.isComposing;
+
+        #if PTHREADS
+        if (targetThread) __emscripten_run_callback_on_thread(targetThread, callbackfunc, eventTypeId, inputEvent, userData);
+        else
+            #endif
```
sbc100, line 832: Can you put all preprocessor directives in column 0?  (i.e. no indention for them).

src/lib/libhtml5.js
```diff
@@ -805,6 +806,48 @@ var LibraryHTML5 = {
...
 
+$registerInputEventCallback__deps: ['$JSEvents', '$findEventTarget', 'malloc', '$stringToUTF8'],
+    $registerInputEventCallback: (target, userData, useCapture, callbackfunc, eventTypeId, eventTypeString, targetThread) => {
+    #if PTHREADS
+    targetThread = JSEvents.getTargetThreadForEventCallback(targetThread);
+    #endif
+    JSEvents.inputEvent ||= _malloc({{{ C_STRUCTS.EmscriptenInputEvent.__size__ }}});
+
+    var inputEventHandlerFunc = (e = event) => {
+        var data = e.data ? e.data : '';
```
sbc100, line 817: can you write this as `e.data || ''`?

src/lib/libhtml5.js
```diff
@@ -805,6 +806,48 @@ var LibraryHTML5 = {
...
+        var inputType = e.inputType ? e.inputType : '';
```
sbc100, line 818: Same

src/lib/libhtml5.js
```diff
@@ -805,6 +806,48 @@ var LibraryHTML5 = {
...
+
+        #if PTHREADS
+        var inputEvent = targetThread ? _malloc({{{ C_STRUCTS.EmscriptenInputEvent.__size__ }}}) : JSEvents.inputEvent;
+        #else
+        var inputEvent = JSEvents.inputEvent;
+        #endif
+        stringToUTF8(data, inputEvent + {{{ C_STRUCTS.EmscriptenInputEvent.data }}}, {{{ cDefs.EM_HTML5_LONG_STRING_LEN_BYTES }}});
+        stringToUTF8(inputType, inputEvent + {{{ C_STRUCTS.EmscriptenInputEvent.inputType }}}, {{{ cDefs.EM_HTML5_SHORT_STRING_LEN_BYTES }}});
+        HEAP8[inputEvent + {{{ C_STRUCTS.EmscriptenInputEvent.isComposing }}}] = e.isComposing;
```
sbc100, line 827: Can you use `makeSetaValue` macro here.

src/lib/libhtml5.js
```diff
@@ -805,6 +806,48 @@ var LibraryHTML5 = {
   emscripten_set_focusout_callback_on_thread: (target, userData, useCapture, callbackfunc, targetThread) =>
     registerFocusEventCallback(target, userData, useCapture, callbackfunc, {{{ cDefs.EMSCRIPTEN_EVENT_FOCUSOUT }}}, "focusout", targetThread),
 
+$registerInputEventCallback__deps: ['$JSEvents', '$findEventTarget', 'malloc', '$stringToUTF8'],
+    $registerInputEventCallback: (target, userData, useCapture, callbackfunc, eventTypeId, eventTypeString, targetThread) => {
```
sbc100, line 810: Indentation looks wrong here.

src/lib/libhtml5.js
sbc100: Is it possible for inputType to be null?  What does that mean?

src/lib/libhtml5.js
```diff
@@ -805,6 +806,48 @@ var LibraryHTML5 = {
...
+        handlerFunc: inputEventHandlerFunc,
+        useCapture
+    };
+    return JSEvents.registerOrRemoveHandler(eventHandler);
+},
+
+  emscripten_set_input_callback_on_thread__proxy: 'sync',
+  emscripten_set_input_callback_on_thread__deps: ['$registerInputEventCallback'],
+  emscripten_set_input_callback_on_thread: (target, userData, useCapture, callbackfunc, targetThread) =>
+    registerInputEventCallback(target, userData, useCapture, callbackfunc, {{{ cDefs.EMSCRIPTEN_EVENT_INPUT }}}, "input", targetThread),
```
sbc100, line 849: Since this function has exactly one call site I wonder if we should just inline it here and avoid the extra function?   But maybe this is simply the pattern used in this file?

test/test_html5_input.c
```diff
@@ -0,0 +1,373 @@
...
+static const char* const inputType_insertFromPaste = "insertFromPaste";
+// * drag and drop:
+static const char* const inputType_insertFromDrop = "insertFromDrop";
+
+static void testInstallCallback();
+static void startTestInputSingleLetter();
+static void startTestPasteText();
+static void startTestDropVeryLongText();
+static void startTestChangeCheckbox();
+static void testingDone();
```
sbc100, line 30: Can you implement these function in usage order and avoid these pre-declarations?

test/test_html5_input.c
```diff
@@ -0,0 +1,373 @@
...
+  setInputBValue("");
+  printf("Please check checkbox C.\n");
+}
+
+static void testingDone() {
+  emscripten_html5_remove_all_event_listeners();
+
+#ifdef REPORT_RESULT
+  REPORT_RESULT(testStatus());
+#endif
```
sbc100, line 354: Can you just do `exit(testStatus())` here and use `btest_exit` to run the test?    This might require using `emscriopten_runtime_keeplive_push()` in main and `emscripten_runtime_keeplive_pop()` here. 

In fact maybe just inline `testStatus` here?

src/lib/libhtml5.js
sbc100: Thats great.  Smaller PRs are (almost) always preferred.

system/include/emscripten/html5.h
alibabashack: Thanks for the suggestions. I think (1) is out of question because it is very easy even for an end user to create very large change `data` values by Ctrl+A, Ctrl+V or similar. Crashes _will_ happen.

I don't feel too good about (2) either because there will be different target groups of users . Some can completely ignore the provided `data` field. E.g. for my current use case which is form validation. I just need to know that _something_ changed and will request the value from the DOM element anyhow. It is waste to perform an allocation here and require the user to take ownership. I can already smell all the memory leaks produced in situations where the data field is ignored. Other users typically only care about a single input char because they deal with keystroke-like stuff.

Maybe we could have independent sets of registration and callback functions:

1. Requests input events with full data via `$stringToNewUTF8`: for those who really need all the data and have to pay in terms of performance due to an allocation being made and dealing with ownership semantics
2. Requests input events without data; still fixed fields (`inputType`, etc.) could be provided: for those who don't need any data and are just waiting for an _any-change hook_.
3. Requests input events with truncated data; the data field is partially copied (as currently implemented): for those who deal with keystrokes and other small input data sets (e.g. see use case in #21960) which may occur with high rate and the input delta data is still relevant; where performance is key.

I think case 2 is optional, because the target group could use case 3 as well and ignore the provided truncated data. Any thoughts?

sbc100: How about anther option:  Use a dyanmic / heap allocation, but have it live only for the life of the callback.

You could use `realloc` on a fixed pointer to avoid  allocation except when the region grows.

sbc100: See SDL_GetKeyName in libsdl.js for an example to using a fixed buffer with `realloc`.

src/lib/libhtml5.js
```diff
@@ -805,6 +806,48 @@ var LibraryHTML5 = {
   emscripten_set_focusout_callback_on_thread: (target, userData, useCapture, callbackfunc, targetThread) =>
     registerFocusEventCallback(target, userData, useCapture, callbackfunc, {{{ cDefs.EMSCRIPTEN_EVENT_FOCUSOUT }}}, "focusout", targetThread),
 
+$registerInputEventCallback__deps: ['$JSEvents', '$findEventTarget', 'malloc', '$stringToUTF8'],
+    $registerInputEventCallback: (target, userData, useCapture, callbackfunc, eventTypeId, eventTypeString, targetThread) => {
+    #if PTHREADS
+    targetThread = JSEvents.getTargetThreadForEventCallback(targetThread);
+    #endif
+    JSEvents.inputEvent ||= _malloc({{{ C_STRUCTS.EmscriptenInputEvent.__size__ }}});
```
sbc100, line 814: Actually it looks like you are doing a dynamic allocation anyway here.

You could replace this with a realloc call if (and only if) the buffer needs to grow.

Then you declare the struct with an array of unknown size as the final element:

```
typedef struct EmscriptenInputEvent {
   ...
   EM_UTF8 data[];   // Variable size string as final element in struct.
}

File: src/lib/libhtml5.js (hunk 4 above)
Comment (alibabashack) [reply src/lib/libhtml5.js:818]: > Is it possible for inputType to be null? What does that mean?

I can't find it in the [spec](https://w3c.github.io/input-events/#interface-InputEvent-Attributes), but it is what I see when checking a checkbox input in chrome.

File: system/include/emscripten/html5.h (hunk 1 above)
Comment (alibabashack) [reply system/include/emscripten/html5.h:194]: Great, `realloc` looks perfect, I'll go this route. Thanks!

Comment (alibabashack) [reply system/include/emscripten/html5.h]: Implementation uses `realloc` for events on the default thread, now.

For all other threads I am allocating every single time as other event types seem to do. Is there some kind of `free()` required? I don't understand the semantics involved here.

File: src/lib/libhtml5.js (hunk 2 above)
Comment (alibabashack) [reply src/lib/libhtml5.js]: fixed

File: src/lib/libhtml5.js (hunk 3 above)
Comment (alibabashack) [reply src/lib/libhtml5.js]: fixed

File: src/lib/libhtml5.js (hunk 4 above)
Comment (alibabashack) [reply src/lib/libhtml5.js]: fixed

File: src/lib/libhtml5.js (hunk 5 above)
Comment (alibabashack) [reply src/lib/libhtml5.js]: fixed

File: src/lib/libhtml5.js (hunk 6 above)
Comment (alibabashack) [reply src/lib/libhtml5.js]: fixed

File: test/test_html5_input.c (hunk 8 above)
Comment (alibabashack) [reply test/test_html5_input.c]: Fixed as requested.

File: test/test_html5_input.c (hunk 9 above)
Comment (alibabashack) [reply test/test_html5_input.c]: > Can you just do `exit(testStatus())` here and use ...

I did the best I can. This works when testStatus() returns 0, then the browser will close and the test suite stops. However, on non 0 testStatus(), the browser does not close and the test suite remains stuck. This is a little inconvenient.

> In fact maybe just inline `testStatus` here?

I would like to keep the checks in a function, because the function definition can stay very close to the observed test status variable definitions. In this way it is more obvious the check function needs to be upgraded when tests are amended.

File: src/lib/libhtml5.js (hunk 10 above)
Comment (alibabashack) [reply src/lib/libhtml5.js]: Jep, thanks, I implemented it this way.

**Comment by alibabashack** [comment]
I also added `$registerInputEventCallback__proxy: 'sync',` because it seems to be present on all other callback handlers. To be honest I have no idea what this does. Could someone explain? Thanks.

**Comment by alibabashack** [comment]
rebased to latest master

File: src/lib/libhtml5.js [hunk 11]
```diff
@@ -805,6 +807,59 @@ var LibraryHTML5 = {
...
+    targetThread = JSEvents.getTargetThreadForEventCallback(targetThread);
+#endif
+
+    var inputEventHandlerFunc = (e = event) => {
+        var data = e.data || '';
+        var inputType = e.inputType || '';
+        var isComposing = e.isComposing;
+        const dataLengthWithTermination = lengthBytesUTF8(data) + 1;
+        const eventSize = {{{ C_STRUCTS.EmscriptenInputEvent.__size__ }}} + dataLengthWithTermination;
+        if(JSEvents.inputEvent == null){
```
sbc100, line 823: Space after keywords like `if`.

src/lib/libhtml5.js
```diff
@@ -805,6 +807,59 @@ var LibraryHTML5 = {
...
 
+  $registerInputEventCallback__proxy: 'sync',
+  $registerInputEventCallback__deps: ['$JSEvents', '$findEventTarget', '$stringToUTF8', '$lengthBytesUTF8', 'malloc', 'realloc'],
+  $registerInputEventCallback: (target, userData, useCapture, callbackfunc, eventTypeId, eventTypeString, targetThread) => {
+#if PTHREADS
+    targetThread = JSEvents.getTargetThreadForEventCallback(targetThread);
+#endif
+
+    var inputEventHandlerFunc = (e = event) => {
+        var data = e.data || '';
```
sbc100, line 818: We use 2-space indentation.

src/lib/libhtml5.js
```diff
@@ -805,6 +807,59 @@ var LibraryHTML5 = {
...
+        var inputType = e.inputType || '';
+        var isComposing = e.isComposing;
+        const dataLengthWithTermination = lengthBytesUTF8(data) + 1;
+        const eventSize = {{{ C_STRUCTS.EmscriptenInputEvent.__size__ }}} + dataLengthWithTermination;
+        if(JSEvents.inputEvent == null){
+            JSEvents.inputEvent = _malloc(eventSize);
```
sbc100, line 824: No need to call `_malloc` at all here since `_realloc` works with null pointer as firs argument.

src/lib/libhtml5.js
```diff
@@ -805,6 +807,59 @@ var LibraryHTML5 = {
...
+            JSEvents.inputEventAllocatedSize = eventSize;
+        }
+        else if(JSEvents.inputEventAllocatedSize < eventSize) {
+            JSEvents.inputEvent = _realloc(JSEvents.inputEvent, eventSize);
+            JSEvents.inputEventAllocatedSize = eventSize;
```
sbc100, line 829: Maybe just call `realloc` unconditionally and then you don't need to trace `inputEventAllocatedSize` at all.

system/include/emscripten/html5.h
```diff
@@ -189,6 +190,15 @@ EMSCRIPTEN_RESULT emscripten_set_focus_callback_on_thread(const char *target __a
 EMSCRIPTEN_RESULT emscripten_set_focusin_callback_on_thread(const char *target __attribute__((nonnull)), void *userData, bool useCapture, em_focus_callback_func callback, pthread_t targetThread);
 EMSCRIPTEN_RESULT emscripten_set_focusout_callback_on_thread(const char *target __attribute__((nonnull)), void *userData, bool useCapture, em_focus_callback_func callback, pthread_t targetThread);
 
+typedef struct EmscriptenInputEvent {
+  bool isComposing;
+  EM_UTF8 inputType[EM_HTML5_SHORT_STRING_LEN_BYTES];
+  EM_UTF8 data[];
```
sbc100, line 196: Perhaps worth adding a comment here.   `EM_UTF8 data[];  // variable size string; must be the final element of the struct`

test/test_html5_input.c
```diff
@@ -0,0 +1,348 @@
...
+static void* const expectedUserDataPtr = (void*)0x13370001;
+
+static bool
+inputCallbackWaitingForLetterA(const int eventType,
+                               const EmscriptenInputEvent* const inputEvent
+                               __attribute__((nonnull)),
+                               void* const userData) {
+  printEvent(inputEvent, __func__);
+
+  if (currentGoal != GoalInputSingleLetter) {
```
sbc100, line 274: How about just `assert(currentGoal == GoalInputSingleLetter)` here and elsewhere?

test/test_html5_input.c
```diff
@@ -0,0 +1,348 @@
...
+                EMSCRIPTEN_RESULT_SUCCESS,
+                "install callback on existing input element");
+
+  setInputAValue(expectedPasteText);
+  printf("Please copy&paste all text from A to B (Ctrl+A, Ctrl+C, Ctrl+V).\n");
+}
+
+static void* const expectedUserDataPtr = (void*)0x13370001;
+
+static bool
```
sbc100, line 267: Would it help keep the test simpler (more readable) if you just removed all the `static` keywords?  They don't really add anything to the test I think.

test/test_html5_input.c
```diff
@@ -0,0 +1,348 @@
...
+
+// The following input type strings are provided by the browser depending
+// on the kind of input used to change an element.
+// See https://w3c.github.io/input-events/#interface-InputEvent-Attributes
+// * typing on the keyboard:
+static const char* const inputType_insertText = "insertText";
+// * copy and paste:
+static const char* const inputType_insertFromPaste = "insertFromPaste";
+// * drag and drop:
+static const char* const inputType_insertFromDrop = "insertFromDrop";
```
sbc100, line 23: Maybe drop the second `const` here since it adds visual noise?

Would it be more readable to just write `#define INPUTTYPE_INSERTTEXT "insertText"`?

sbc100 (commented):
lgtm, with a few minor nits, and some comments about the test code.

@kripken and @juj, I don't know much about how the html5 API evolved or why this was not originally part of it.  Does this addition look reasonable to you?  @dschuff too.

test/test_html5_input.c
alibabashack: Well, this is about const correctness. The second `const` safeguards against accidental modification which would always be an error. From my perspective, it were distracting if the pointer would not be const as I would expect it to be changed somewhere.

test/test_html5_input.c
alibabashack: Isn't this just the standard practice of information hiding in C? Make everything static that is not required outside the compilation unit. Same as with constness below, it distracts me if these good practices are skipped. They add intent.

src/lib/libhtml5.js
alibabashack: This does not seem to be enforced by your .clang-format setup. Is there another way to format automatically?

src/lib/libhtml5.js
alibabashack: code was removed

src/lib/libhtml5.js
alibabashack: Always calling `realloc`.

system/include/emscripten/html5.h
alibabashack: Agreed, added suggested note.

test/test_html5_input.c
sbc100: Again, I would agree with you, outside of the context of test code where I tend to prefer brevity / simplicity.

sbc100: Actually for these strings I think it might make more sense just to inline them rather than abstracting them behind a variable.  We can just remove these variables completely and inline the C strings at the point of use below?

This fits the DRY vs DAMP principle that a lot of folks think makes sense for test code: https://benjaminlo-dev.medium.com/dry-vs-damp-16513610cdeb

alibabashack: Ok, the last thing we need is a debate on principles, I guess.  I replaced the usages with string literals as suggested.

test/test_html5_input.c
alibabashack: Removed all `static` keywords for brevity.

alibabashack:
> lgtm, with a few minor nits, and some comments about the test code.
> 
> @kripken and @juj, I don't know much about how the html5 API evolved or why this was not originally part of it. Does this addition look reasonable to you? @dschuff too.

Hi @sbc100 @kripken  @juj @dschuff, I think this is ready, now. Could you take a look, please?

src/lib/libhtml5.js
```diff
@@ -806,6 +807,52 @@ var LibraryHTML5 = {
...
+#if PTHREADS
+    targetThread = JSEvents.getTargetThreadForEventCallback(targetThread);
+#endif
+
+    var inputEventHandlerFunc = (e = event) => {
+      var data = e.data || '';
+      var inputType = e.inputType || '';
+      var isComposing = e.isComposing;
+      const dataLengthWithTermination = lengthBytesUTF8(data) + 1;
+      const eventSize = {{{ C_STRUCTS.EmscriptenInputEvent.__size__ }}} + dataLengthWithTermination;
```
sbc100, line 822: We don't use `const` yet in most of the system JS libraries.  Can you just use `var` here?

system/include/emscripten/html5.h
sbc100: Yes, the `emscripten_run_callback_on_thread` function takes care of calling `free` in this case.

site/source/docs/api_reference/html5.h.rst
```diff
@@ -559,6 +559,85 @@ Functions
...
+
+  The event structure passed in `DOM input events <http://www.w3.org/TR/DOM-Level-3-Events/#idl-inputevent>`_.
+
+  .. c:member:: EM_UTF8 data
+
+    The `data <https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/data>`_ field contains inserted characters
+    or an empty string in case the input event deleted characters or a non-character input was performed
+    (e.g. a checkbox was toggled).
+
+    Maximum size 128 ``char`` (i.e. ``EM_UTF8 nodeName[128]``).
```
sbc100, line 585: This line seems wrong, maybe just remove it?

site/source/docs/api_reference/html5.h.rst
```diff
@@ -559,6 +559,85 @@ Functions
...
+
+  .. c:member:: EM_UTF8 inputType
+
+    The `inputType <https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/inputType>`_ indicating the context
+    of the input event generation (e.g. typing on the keyboard, copy&paste or drag&drop).
+    See the `list of possible input types <https://w3c.github.io/input-events/#interface-InputEvent-Attributes>`_.
+    Null values are represented as zero length string.
+
+    Maximum size 32 ``char`` (i.e. ``EM_UTF8 id[32]``).
```
sbc100, line 594: This line seems wrong since it refers to `id` and not `inputType`

PR #20372, opened 2023-10-02
[EH] Make Wasm EH rethrow print stack traces

aheejin:
In Wasm EH, even if we set `-sASSERTION` or `-sEXCEPTION_STACK_TRACES`, if we rethrow an exception in the code, we lose the effect of that option because previously we called `__throw_exception_with_stack_trace` only in `__cxa_throw`:
https://github.com/emscripten-core/emscripten/blob/9ce7020632aa6f7578c6e40e197b800a4dd7073f/system/lib/libcxxabi/src/cxa_exception.cpp#L294-L296

This adds the same mechanism to `__cxa_rethrow` (which is used for C++ `throw;`) and `__cxa_rethrow_primary_exception` (which is used for `std::rethrow_exception`).

This does not solve the problem of losing stack traces _before_ the rethrowing. libc++abi's `__cxa_rethrow` and
`__cxa_rethrow_primary_exception` are implemented as throwing a pointer in the same way we first throw it and they are not aware of any metadata. This happens even in the native platform with GDB; GDB's backtrace only shows stack traces after rethrowing. We may try to fix this later by passing `exnref`
(https://github.com/WebAssembly/exception-handling/issues/281) to the library, but this is likely to take some time.

Partially fixes #20301.

Issue #14035, opened 2021-04-28
Some wasm EH tests fail with ASan


aheejin:
Some wasm EH tests fail with asan, including `test_exceptions_uncaught_count_wasm_eh`:

```
$ ./tests/runner.py asan.test_exceptions_uncaught_count_wasm_eh 
runner:WARNING: use EMTEST_ALL_ENGINES=1 in the env to run against all JS engines, which is slower but provides more coverage
Test suites:
['test_core']
Running test_core: (1 tests)
(checking sanity from test runner)
shared:INFO: (Emscripten: Running sanity checks)
test_exceptions_uncaught_count_wasm_eh (test_core.asan) ... -- begin program output --
in inner catch: 0
exception thrown: RuntimeError: unreachable,RuntimeError: unreachable
    at main (<anonymous>:wasm-function[33]:0xdc7)
    at Module._main (test_exceptions_uncaught_count.js:1:83212)
    at callMain (test_exceptions_uncaught_count.js:1:89234)
    at doRun (test_exceptions_uncaught_count.js:1:89794)
    at run (test_exceptions_uncaught_count.js:1:89966)
    at runCaller (test_exceptions_uncaught_count.js:1:89089)
    at removeRunDependency (test_exceptions_uncaught_count.js:1:13048)
    at receiveInstance (test_exceptions_uncaught_count.js:1:16918)
    at receiveInstantiationResult (test_exceptions_uncaught_count.js:1:17076)
-- end program output --
FAIL

======================================================================
FAIL: test_exceptions_uncaught_count_wasm_eh (test_core.asan)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/google/home/aheejin/emscripten/tests/runner.py", line 345, in resulting_test
    return func(self, *args)
  File "/usr/local/google/home/aheejin/emscripten/tests/test_core.py", line 105, in metafunc
    f(self)
  File "/usr/local/google/home/aheejin/emscripten/tests/test_core.py", line 1432, in test_exceptions_uncaught_count
    self.do_core_test('test_exceptions_uncaught_count.cpp')
  File "/usr/local/google/home/aheejin/emscripten/tests/test_core.py", line 288, in do_core_test
    self.do_run_in_out_file_test('core', testname, **kwargs)
  File "/usr/local/google/home/aheejin/emscripten/tests/runner.py", line 1036, in do_run_in_out_file_test
    self._build_and_run(srcfile, expected, **kwargs)
[...]

PR #23552, opened 2025-01-31
[EH] Make lsan/asan work with Wasm EH

### LINKED ISSUE #17004: memory leak in `test_exceptions_longjmp3_wasm`
```
$ ./tests/runner lsan.test_exceptions_longjmp3_wasm --save-dir
..
Direct leak of 40 byte(s) in 1 object(s) allocated from:
    #0 0x9490 in malloc+0x9490 (test_exceptions_longjmp3.wasm+0x9490)
    #1 0x914 in test()+0x914 (test_exceptions_longjmp3.wasm+0x914)
    #2 0xa97 in __original_main+0xa97 (test_exceptions_longjmp3.wasm+0xa97)
    #3 0xb6f in main+0xb6f (test_exceptions_longjmp3.wasm+0xb6f)
    #4 0x8000054d  (JavaScript+0x54d)
    #5 0x800013ee in callMain /tmp/emscripten_test/test_exceptions_longjmp3.js:5102:15
    #6 0x80001427 in doRun /tmp/emscripten_test/test_exceptions_longjmp3.js:5159:23
```

Seems like there must be an internal allocation at setjmp that is not fee'd

Sadly we don't run this particular test on the releases waterfal because we don't have d8/v8 available there (which this test requires).

### LINKED ISSUE #21124: std::rethrow_exception leaks memory with wasm EH (or at least lsan thinks it does)
`lsan.test_exceptions_primary_wasm` and `lsan.test_exceptions_multiple_inherit_rethrow_wasm` currently fail with a memory leak.

Sadly we don't run these tests on the emscripten-releases waterfall since we only have node 16 here and therefore skip all wasm EH tests by setting `EMTEST_SKIP_EH`.

aheejin:
`std::rethrow_exception` leaked memory because dependent exceptions were not recognized as such and thus not freed properly here: https://github.com/emscripten-core/emscripten/blob/44fd294e8244afa710cecb2c10c9a644a6549b8e/system/lib/libcxxabi/src/cxa_exception.cpp#L575-L583 The reason for that was, in `__cxa_rethrow_primary_exception`, we were incorrectly throwing the primary exception, not the dependent one. I think the error was introduced during copy-pasting.

`isDependentException` method was copied from `cxa_exception.cpp`, because it is a static function, as in the case for other utility functions in that file.

This also adds handling for dependent exceptions in `__get_exception_message`.

Fixes #21124.


test/test_core.py
```diff
@@ -1499,8 +1495,6 @@ def test_exceptions_longjmp2(self):
 
   @with_all_eh_sjlj
   def test_exceptions_longjmp3(self):
-    if '-fwasm-exceptions' in self.emcc_args:
-      self.skipTest('https://github.com/emscripten-core/emscripten/issues/17004')
```
aheejin, line 1503: To be precise this was not fixed by this PR. It was already working under lsan and asan. Not sure what the cause for the failure was. Also many tests reported in #14035 were already working too.

And we were disabling this test as a whole and not only under lsan! 😨

system/lib/libcxxabi/src/cxa_exception.cpp
```diff
@@ -789,9 +789,9 @@ __cxa_rethrow_primary_exception(void* thrown_object)
         // In debug mode, call a JS library function to use
         // WebAssembly.Exception JS API, which enables us to include stack
         // traces
-        __throw_exception_with_stack_trace(&exception_header->unwindHeader);
+        __throw_exception_with_stack_trace(&dep_exception_header->unwindHeader);
 #else
-        _Unwind_RaiseException(&exception_header->unwindHeader);
+        _Unwind_RaiseException(&dep_exception_header->unwindHeader);
```
sbc100, line 794: It looks like this reverts this line to match upstream again?   If so, LGTM.  I wonder how this change changed in the first place

system/lib/libcxxabi/src/cxa_exception_js_utils.cpp
```diff
@@ -78,6 +86,13 @@ void __get_exception_message(void* thrown_object, char** type, char** message) {
     static_cast<const __shim_type_info*>(&typeid(std::exception));
   int can_catch = catch_type->can_catch(thrown_type, thrown_object);
   if (can_catch) {
+#if __WASM_EXCEPTIONS__
+    if (isDependentException(&exception_header->unwindHeader))
+      thrown_object =
+        reinterpret_cast<__cxa_dependent_exception*>(exception_header)
+          ->primaryException;
```
sbc100, line 93: I wish I could say I understood the whole exception / dependent exception / unwind header stuff... but if this fixes the issue then lgtm!

sbc100 (approved):

system/lib/libcxxabi/src/cxa_exception.cpp
aheejin: > It looks like this reverts this line to match upstream again?

Yes

> I wonder how this change changed in the first place

In #20372, I copied the code from `__cxa_throw` and forgot to modify it 🥲

system/lib/libcxxabi/src/cxa_exception_js_utils.cpp
aheejin: It looks when you do `std::rethrow_exception`, internally a dependent exception of the original (primary) exception is created and thrown. But that dependent exception apparently cannot be cast to `__cxa_exception` and can't be called `what()` to get the message. So this is just getting back the original primary exception to get the message.

Issue #13089, opened 2020-12-22
`emscripten_get_callstack` + source maps is currently broken


sbc100:
I believe the only test we have for this is `browser.test_emscripten_log`.

Running this test manually and inspecting the return value of `emscripten_get_callstack` gives the following output:

```
at bar(int, char*, double) (test.wasm:wasm-function[8]:0x37d)
    at void Foo<int>() (test.wasm:wasm-function[14]:0x1184)
    at __original_main (test.wasm:wasm-function[13]:0x1102)
    at main (test.wasm:wasm-function[15]:0x118c)
    at http://0.0.0.0:2345/test.js:1544:22
    at callMain (emscripten_log.cpp:135:0)
     = callMain (test.js:2748:15)
    at doRun (emscripten_log.cpp:135:0)
     = doRun (test.js:2812:23)
```

This is clearly wrong.   The innermost stack frames (doRun, callMain) are clearly not native functions where as the outer ones are.   `main`, `__original_main`, and `Foo` should be showing as coming from `emscripten_log.cpp` and `callMain` should not.

I'm not sure if its worth fixing this but I guess there is some kind of bug in either `$EMSCRIPTEN/src/emscripten-source-map.min.js` or in our usage of this.

sbc100:
I'm also not sure if this ever worked...

PR #23553, opened 2025-01-31
Remove EM_LOG_C_STACK feature of emscripten_log

sbc100:
This was supported by `emscripten-source-map.min.js` which seems like minified source map parser, but I don't think this feature has worked for at least 4 years since we switched to using wasm and wasm source maps.

This feature was part of `emscripten_log` since it was first added in 26d6ad370 but I know of anyone that ever used it.

Fixes: #13089


ChangeLog.md
```diff
@@ -25,6 +25,10 @@ See docs/process.md for more on how version tagging works.
   emcc itself.  This enables support for C++20 in cmake projects. (#21987)
 - The version of python required to run emscripten was bumped from 3.6 to 3.8.
   (#23417)
+- The `EM_LOG_C_STACK` flag to `emscripten_log` was deprecated and the helper
+  file on which it was based (`emscripten-source-map.min.js`) was deteted.  It
+  seems this feature has been broken for at least 4 years and making it work
+  with WebAssembly was no feasible. (#23553)
```
```suggestion
  with WebAssembly is not feasible. (#23553)
```

ChangeLog.md
```diff
@@ -25,6 +25,10 @@ See docs/process.md for more on how version tagging works.
...
```
kripken, line 32: Why is this not feasible? Wasm stores the source map on the side, like JS, so can't we read the source map there?

sbc100: Perhaps infeasible is the wrong word.  I think it would take significant effort.  I will re-phrase.

sbc100:
@kripken do you think its worth trying to resurect this feature?  It seems like it was some kind of user-space source-map parser.. I'm not sure it worth the effort given that nobody has asked for it so many years.

kripken:
I agree it's not worth keeping given the lack of users. Have people been using it with wasm2js perhaps though? If not sgtm to remove.

sbc100:
> I agree it's not worth keeping given the lack of users. Have people been using it with wasm2js perhaps though? If not sgtm to remove.

Source maps do not work with wasm2js (` error: wasm2js does not support source maps yet (debug in wasm for now)`)

ChangeLog.md
```diff
@@ -25,6 +25,9 @@ See docs/process.md for more on how version tagging works.
   emcc itself.  This enables support for C++20 in cmake projects. (#21987)
 - The version of python required to run emscripten was bumped from 3.6 to 3.8.
   (#23417)
+- The `EM_LOG_C_STACK` flag to `emscripten_log` was deprecated and the helper
+  file on which it was based (`emscripten-source-map.min.js`) deteted.  It
+  seems this feature has been broken for at least 4 years now. (#23553)
```
kripken, line 30:
```suggestion
- The `EM_LOG_C_STACK` flag to `emscripten_log` was deprecated and the helper
  file on which it was based (`emscripten-source-map.min.js`) deleted.  This
  feature (userspace source map parsing in logs) was not ported to wasm
  source maps, so it has not worked, and we have had no requests for it. This
  has no impact on browser devtools support of source maps (which work
  fully). (#23553)
```

kripken (commented):
lgtm, but how about this rephrasing of the changlog? To clarify this has no impact on source maps in general.

ChangeLog.md
sbc100: Done.  With a little re-phrasing of my own.

kripken (approved):

PR #23554, opened 2025-01-31
Add wasm exception handling support to sdl2_image port

hoodmane:
`sdl2_image` uses setjmp/longjmp so it needs to be aware of wasm exception handling.


tools/ports/sdl2_image.py
```diff
@@ -46,6 +46,8 @@ def get_lib_name(settings):
     libname += '-' + formats
   if settings.PTHREADS:
     libname += '-mt'
+  if settings.SUPPORT_LONGJMP == 'wasm':
+    libname += '-wasm-sjlj'
```
sbc100, line 50: Should we add these to the `variants` list above too?

sbc100 (approved):

sbc100:
Should we add `with_all_sjlj` to on the the browser tests that uses `-sUSE_SDL_IMAGE=2`?

Can you confirm that it fails without this change too?

tools/ports/sdl2_image.py
hoodmane: There's a lot of missing variants. The one I'm using is called `libSDL2_image_bmp-gif-jpg-png-wasm-sjlj.a`.

hoodmane:
The following tests fail:
* test_sdl2_image_formats
* test_sdl2_gl_frames_swap
* test_sdl2_canvas_proxy

There's also the problem that `require_wasm_eh` currently doesn't believe that the browser supports wasm-eh tests.

hoodmane:
After this PR, `test_sdl2_image_formats_wasm` passes but `test_sdl2_gl_frames_swap_wasm` and `test_sdl2_canvas_proxy_wasm` still fail. All other tests pass with wasm sjlj.

sbc100:
> After this PR, `test_sdl2_image_formats_wasm` passes but `test_sdl2_gl_frames_swap_wasm` and `test_sdl2_canvas_proxy_wasm` still fail. All other tests pass with wasm sjlj on the main branch.

If `test_sdl2_image_formats_wasm` fails without this change and passes with it, then I would just add the decorator to that one test are part of this PR.

hoodmane:
You don't think it's worth documenting that the other two fail like this?

hoodmane:
This is blocked though on #23547.

sbc100 (approved):

hoodmane:
@sbc100 any idea why it says:
```
Error: not compiled for this environment (did you build to HTML and try to run it not on the web, or set ENVIRONMENT to something - like node - and run it someplace else - like on the web?)
```
I don't reproduce locally. Maybe it would be a good time to rerun with ssh and see if I can fix it that way.

sbc100:
> @sbc100 any idea why it says:
> 
> ```
> Error: not compiled for this environment (did you build to HTML and try to run it not on the web, or set ENVIRONMENT to something - like node - and run it someplace else - like on the web?)
> ```
> 
> I don't reproduce locally. Maybe it would be a good time to rerun with ssh and see if I can fix it that way.

Looks like the problem is with `require_wasm_eh` ... it adds `-sENVIRONMENT=shell` in some cases.  We should probably fix that.

sbc100:
> > @sbc100 any idea why it says:
> > ```
> > Error: not compiled for this environment (did you build to HTML and try to run it not on the web, or set ENVIRONMENT to something - like node - and run it someplace else - like on the web?)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't reproduce locally. Maybe it would be a good time to rerun with ssh and see if I can fix it that way.
> 
> Looks like the problem is with `require_wasm_eh` ... it adds `-sENVIRONMENT=shell` in some cases. We should probably fix that.

I think you can put an early return in `require_wasm_eh` if `self.is_browser_test()` is set.

PR #23559, opened 2025-01-31
Inject `minimal_runtime_exit_handling.js` as part of `also_with_minimal_runtime`. NFC

sbc100:
This allows `also_with_minimal_runtime` to be used in more places removing a bunch of duplication.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -5596,21 +5596,12 @@ got: 10
     self.do_runf('benchmark/benchmark_utf8.c', 'OK.')
 
   # Test that invalid character in UTF8 does not cause decoding to crash.
+  @also_with_minimal_runtime
   @parameterized({
-    '': [[]],
-    'textdecoder': [['-sTEXTDECODER']],
+    '': ([],),
+    'textdecoder': (['-sTEXTDECODER'],),
   })
   def test_utf8_invalid(self, args):
-    self.do_runf('test_utf8_invalid.c', 'OK.', emcc_args=args)
-
-  # Test that invalid character in UTF8 does not cause decoding to crash.
-  @parameterized({
-    '': [[]],
-    'textdecoder': [['-sTEXTDECODER']],
-  })
-  def test_minimal_runtime_utf8_invalid(self, args):
-    self.set_setting('MINIMAL_RUNTIME')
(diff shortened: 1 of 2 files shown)


test/test_core.py
```diff
@@ -5597,24 +5597,14 @@ def test_utf8_textdecoder(self):
...
   @parameterized({
-    '': [[]],
-    'textdecoder': [['-sTEXTDECODER']],
+    '': ([],),
+    'textdecoder': (['-sTEXTDECODER'],),
   })
   def test_utf8_invalid(self, args):
     self.do_runf('test_utf8_invalid.c', 'OK.', emcc_args=args)
 
-  # Test that invalid character in UTF8 does not cause decoding to crash.
```
brendandahl, line 5607: It looks like a few tests are removed. Is that suppose to be in this PR?

sbc100: The tests that were removed were just dupliates of existing tests that ran with `MINIMAL_RUNTIME`.   Instead if duplicating those tests I added `@also_with_minimal_runtime` which basically does the same thing (i.e. creates a duplicate of the test).

test/test_core.py
```diff
@@ -9359,20 +9334,12 @@ def test_Module_dynamicLibraries(self, args):
...
 
   # Tests the emscripten_get_exported_function() API.
+  @also_with_minimal_runtime
   def test_get_exported_function(self):
     self.set_setting('ALLOW_TABLE_GROWTH')
     self.emcc_args += ['-lexports.js']
     self.do_core_test('test_get_exported_function.cpp')
 
-  # Tests the emscripten_get_exported_function() API.
-  @no_asan('TODO: ASan support in minimal runtime')
```
brendandahl, line 9368: Does `@also_with_minimal_runtime` somehow skip asan automatically?

sbc100: No, but it maybe it should?   Let me check

sbc100: The test now passes under Asan and minimal runtime!

sbc100: I split that out into its own change: https://github.com/emscripten-core/emscripten/pull/23562

sbc100: OK, #23562 has now landed so there should be no more change here WRT asan.

brendandahl (approved):

PR #23564, opened 2025-01-31
Enable compiler.mjs to take output filename. NFC

sbc100:
This is an internal-only change mostly useful for debugging.  When the compiler is run internally we always use stdout.

### DIFF
--- a/src/compiler.mjs
+++ b/src/compiler.mjs
@@ -16,12 +16,15 @@ loadDefaultSettings();
 const options = {
   help: {type: 'boolean', short: 'h'},
   'symbols-only': {type: 'boolean'},
+  output: {type: 'string', short: 'o'},
 };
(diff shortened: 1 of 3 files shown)


sbc100:
I use hyperfine to confirm that this change doesn't regress performance:

Before:

```
$ hyperfine './emcc hello.o'
Benchmark 1: ./emcc hello.o
  Time (mean ± σ):     456.8 ms ±   4.4 ms    [User: 382.4 ms, System: 169.2 ms]
  Range (min … max):   450.2 ms … 463.9 ms    10 runs
```

After:

```
$ hyperfine './emcc hello.o'
Benchmark 1: ./emcc hello.o
  Time (mean ± σ):     458.9 ms ±   3.5 ms    [User: 381.2 ms, System: 169.4 ms]
  Range (min … max):   453.7 ms … 464.0 ms    10 runs
```

PR #23565, opened 2025-02-01
Move all addOn<X> functions into a JS library.

brendandahl:
- Make all `addOn<X>` functions have a corresponding compile time version `addAt<X>`.
- Make all the `addOn<X>` functions use the compile time `addAt<X>` helper so the `addOn<X>` code is not emitted if it is not needed.
- Add the missing injection points to the minimal runtime.
- Add tests to ensure the order in regular runtime and mininimal runtime.


src/lib/libcore.js
```diff
@@ -2296,6 +2296,74 @@ addToLibrary({
...
+  // compiled time version named addAt<X> that will instead inline during
+  // compilation (see parseTools.mjs).
+  // Note: if there are both runtime and compile time code, the runtime
+  // callbacks will be invoked before the compile time code.
+
+  // See ATPRERUNS in parseTools.mjs for more information.
+  $__ATPRERUN__: [],
+  $__ATPRERUN____deps: ['$callRuntimeCallbacks'],
+  $__ATPRERUN____postset: () => {
+    ATPRERUNS.unshift('callRuntimeCallbacks(__ATPRERUN__);');
```
sbc100, line 2310: I find this pretty hard to wrap my head around... but I think I get it.  The naming is very confusing.

src/lib/libdylink.js
```diff
@@ -612,6 +612,7 @@ var LibraryDylink = {
     '$currentModuleWeakSymbols',
     '$updateTableMap',
     '$wasmTable',
+    '$__ATPOSTCTOR__',
```
sbc100, line 615: Where is this needed?

src/parseTools.mjs
```diff
@@ -1093,7 +1117,11 @@ function ENVIRONMENT_IS_WORKER_THREAD() {
 
 addToCompileTimeContext({
   ATEXITS,
+  ATPRERUNS,
   ATINITS,
+  ATPOSTCTORS,
+  ATMAINS,
+  ATPOSTRUNS,
```
sbc100, line 1124: Do do need to make these array available to the macro context as well as the function for adding to them?

src/lib/libcore.js
```diff
@@ -2296,6 +2296,74 @@ addToLibrary({
...
+  // See ATPOSTCTORS in parseTools.mjs for more information.
+  $__ATPOSTCTOR__: [],
+  $__ATPOSTCTOR____deps: ['$callRuntimeCallbacks'],
+  $__ATPOSTCTOR____postset: () => {
+    ATPOSTCTORS.unshift('callRuntimeCallbacks(__ATPOSTCTOR__);');
+  },
+  $addOnPostCtor__deps: ['$__ATPOSTCTOR__'],
+  $addOnPostCtor: (cb) => {
+    __ATPOSTCTOR__.unshift(cb);
+  },
```
sbc100, line 2335: These can just be on-lineser maybe: `$addOnPostCtor: (cb) => __ATPOSTCTOR__.unshift(cb),`?

test/test_other.py
```diff
@@ -8463,9 +8463,21 @@ def test_binaryen_stack_ir(self, opts, disable_and_enable):
...
+    addAtPreMain("console.log(`addAtPreMain`);");
+    addAtExit("console.log(`addAtExit`);");
+    addAtPostRun("console.log(`addAtPostRun`);");
+    // Add all the runtime equivalents of the above events.
+    return 'addOnPreRun(() => console.log("addOnPreRun"));\n' +
+           'addOnInit(() => console.log("addOnInit"));\n' +
+           'addOnPostCtor(() => console.log("addOnPostCtor"));\n' +
+           'addOnPreMain(() => console.log("addOnPreMain"));\n' +
+           'addOnExit(() => console.log("addOnExit"));\n' +
+           'addOnPostRun(() => console.log("addOnPostRun"));\n';
```
sbc100, line 8480: I wonder if we should put these lines in a pre.js file?    I think that would make match more closely the use case.

tools/link.py
```diff
@@ -1415,6 +1417,11 @@ def phase_linker_setup(options, state):  # noqa: C901, PLR0912, PLR0915
       # The following symbols need exposing to load and bootstrap the audio worklet:
       settings.INCOMING_MODULE_JS_API += ['instantiateWasm', 'wasm', 'wasmMemory']
 
+  if 'preRun' in settings.INCOMING_MODULE_JS_API and not settings.MINIMAL_RUNTIME:
+    settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE.append('$addOnPreRun')
+  if 'postRun' in settings.INCOMING_MODULE_JS_API and not settings.MINIMAL_RUNTIME:
+    settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE.append('$addOnPostRun')
```
sbc100, line 1423: How about putting these 4 lines in a `if not MINIUMAL_RUNTIME` block?

sbc100: Are there not other INCOMING_MODULE_JS_API elements that need to be included here?

sbc100 (commented):
Nice!

A little confusing but great to see.   Do we have any code size changes?

src/parseTools.mjs
brendandahl: Yes, to make the compile time/runtime callback order consistent. I need the raw array so I can push the runtime callbacks to be the very first thing called. Alternatively, the `addAt<x>` function could take an optional argument that adds to the front of the queue instead of the back.

src/lib/libdylink.js
brendandahl: Down below in some of the initialization code.

tools/link.py
brendandahl: PreRun and PostRun are the only ones that can come from the Module object in this style:
```js
#if expectToReceiveOnModule('preRun')
  if (Module['preRun']) {
    if (typeof Module['preRun'] == 'function') Module['preRun'] = [Module['preRun']];
    while (Module['preRun'].length) {
      addOnPreRun(Module['preRun'].shift());
    }
  }
#endif
```

src/lib/libcore.js
brendandahl: Yeah, the difference of `at` vs `on` is not easy to reason at from a glance. We could probably rename the compile time versions without much breakage. Maybe a follow up?

tools/link.py
sbc100: It looks like at least postRun, preRun and preInit are in the INCOMING_MODULE_JS_API list: https://github.com/emscripten-core/emscripten/blob/1e2d8343f4dc3273aa3788414336d0a26f66a09b/src/settings.js#L997-L1011

Maybe `preInit` should not be in that list becasuse it doesn't work?

sbc100: Actually I do see `preInit` is tests on the Module API.

sbc100: See test_preinit in test_other.py

brendandahl: Oh, but it doesn't use callruntimecallbacks

sbc100: Ah yes.. I wonder why?  I think I tried to make it use `callRuntimeCallbacks` once and failed.. I can't remember why that didn't work.

Presumably it should?

sbc100: I see.  Fair enough maybe lets just leave that as then.

brendandahl: On another note, this does make we wonder if we really need preInit and preRun. They look like they will run basically one after the other.

test/test_other.py
sbc100: Should we include preInit in this test, so that we can be sure of the relative order?

Maybe we don't need `test_preinit` in that case since it seems to do that same thing

sbc100: Can you move the into a pre.js file so we don't have a `__postset` that has side effects *and* returns a string at the same time? 

I think that would be more obvious what is going on and I think it would how most ppl use these functions.

test/test_other.py
```diff
@@ -8481,7 +8493,28 @@ def test_run_order(self):
...
+      'addAtPreMain',
+      'main',
+      'addOnExit',
+      'addAtExit',
+      'addOnPostRun',
+      'addAtPostRun',
+    ]
+    if self.get_setting('MINIMAL_RUNTIME'):
+      # Remove postRun events since they will not happen with MINIMAL_RUNTIME +
+      # EXIT_RUNTIME since it throws an exception during _proc_exit.
```
sbc100, line 8514: I think that is only true when `exit()` is called.  When main returns normally (as in this test) I don't think anything get thrown.

sbc100: Indeed when I build the minimal runtime version of this test I don't see `proc_exit` in the generated JS file.

brendandahl: I'm seeing something different. When I use `EMCC_DEBUG=1 test/runner other.test_run_order_minimal_runtime` `src.js` contains _proc_exit that throws.

brendandahl: Are you building with '-sEXIT_RUNTIME'?

test/test_other.py
```diff
@@ -8493,28 +8489,42 @@ def test_run_order(self):
...
+      # exception during _proc_exit.
+      ['addOnPostRun', False],
+      ['addAtPostRun', False]
     ]
     if self.get_setting('MINIMAL_RUNTIME'):
-      # Remove postRun events since they will not happen with MINIMAL_RUNTIME +
-      # EXIT_RUNTIME since it throws an exception during _proc_exit.
-      del expected_order[-2:]
+      expected_order = [item[0] for item in expected_order if item[1]]
       self.emcc_args += ['--pre-js', test_file('minimal_runtime_exit_handling.js')]
```
sbc100, line 8517: You can remove this line since its part of `also_with_minimal_runtime` now

test/test_other.py
```diff
@@ -8480,7 +8488,41 @@ def test_run_order(self):
...
+      ['addOnPostCtor', True],
+      ['addAtPostCtor', True],
+      ['addOnPreMain', True],
+      ['addAtPreMain', True],
+      ['main', True],
+      ['addOnExit', True],
+      ['addAtExit', True],
+      ['modulePostRun', False],  # Not supported in minimal runtime.
+      # The following two events are supported in miminal runtime, but they will
+      # not happen with MINIMAL_RUNTIME + EXIT_RUNTIME since it throws an
```
sbc100, line 8509: Technically you don't need to use `EXIT_RUNTIME` here.. but then I guess you might not see other events like `addOnExit`?

src/lib/libcore.js
```diff
@@ -2296,6 +2296,74 @@ addToLibrary({
 
   $noExitRuntime: "{{{ makeModuleReceiveExpr('noExitRuntime', !EXIT_RUNTIME) }}}",
 
+  // The following addOn<X> functions are for adding runtime callbacks at
+  // various executions points. Each addOn<X> function has a corresponding
+  // compiled time version named addAt<X> that will instead inline during
```
sbc100, line 2301: s/compiled/compile/

src/lib/libcore.js
sbc100: How about this?

src/lib/libdylink.js
sbc100: Can we call `addOnPostCtor` instead?    I think it would be good if we use those helpers everywhere (or nowhere).

src/lib/libcore.js
```diff
@@ -2296,6 +2296,74 @@ addToLibrary({
...
+  $__ATPRERUN____deps: ['$callRuntimeCallbacks'],
+  $__ATPRERUN____postset: () => {
+    ATPRERUNS.unshift('callRuntimeCallbacks(__ATPRERUN__);');
+  },
+  $addOnPreRun__deps: ['$__ATPRERUN__'],
+  $addOnPreRun: (cb) => {
+    __ATPRERUN__.unshift(cb);
+  },
+  // See ATINITS in parseTools.mjs for more information.
+  $__ATINIT__: [],
```
sbc100, line 2317: Do these arrays need to be visible to user code?   If not, maybe we can mark all these arrays as `__internal: true`?

Also, perhaps we could give them better names without the double underscores?   Now sure what the best thing might be?   `runtimeCallbacksOnInit` ?   Or something shorter:  `onInits`?

test/test_other.py
brendandahl: Yeah, I'm trying to test at much of the regular runtime ordering as I can.

src/lib/libcore.js
brendandahl: They don't need to be visible. I'll add and rename.

src/lib/libcore.js
```diff
@@ -2296,6 +2296,62 @@ addToLibrary({
...
+    ATPOSTCTORS.unshift('callRuntimeCallbacks(__ATPOSTCTOR__);');
+  },
+  $addOnPostCtor__deps: ['$__ATPOSTCTOR__'],
+  $addOnPostCtor: (cb) => __ATPOSTCTOR__.unshift(cb),
+  // See ATMAINS in parseTools.mjs for more information.
+  $__ATMAIN__: [],
+  $__ATMAIN____deps: ['$callRuntimeCallbacks'],
+  $__ATMAIN____postset: () => {
+    ATMAINS.unshift('callRuntimeCallbacks(__ATMAIN__);');
+  },
```
sbc100, line 2335: These can be one-liners too I think?

sbc100 (approved):
LGTM! 

I do wonder if we really do need all these different hooks,  but as long as they are pay-as-you-go I guess there isn't much harm in having a lot of them.

src/lib/libcore.js
brendandahl: I tried, but it doesn't work because the return value of `unshift` is the length of the array, so that gets put into the js file as a postset. At least the one liner here doesn't matter for output code size.

sbc100: Ah yes, that old double meaning of `__postset`..   grrr.

sbc100 (approved):

PR #23569, opened 2025-02-02
Update GoogleTest tested against to version 1.15.2

mcbarton:
The GoogleTest which is currently being tested works with emscripten as part of the automated tests is a commit sometime after version 1.13.0 going by the readme when you cd into the submodule. This PR updates the version being tested to the latest available release 1.15.2 . I updated it to the tag v1.15.2 in GoogleTests tags, so it pinned to a specific release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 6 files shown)


mcbarton:
@sbc100 Not entirely sure how to fix the check-expectations ci job. I ran the command the ci told me to fix the issue, but the job still failed.

PR #23579, opened 2025-02-04
Fix regression with instantiateWasm + MODULARIZE

sbc100:
We didn't have test coverage of this case and it was broken in #23157.

Fixes: #23577


test/test_other.py
```diff
@@ -124,6 +124,20 @@ def metafunc(self, backend, *args, **kwargs):
   return metafunc
 
 
+def also_with_modularize(f):
+  assert callable(f)
+
+  @wraps(f)
+  def metafunc(self, molularize, *args, **kwargs):
```
kripken, line 131:
```suggestion
  def metafunc(self, modularize, *args, **kwargs):
```

test/test_other.py
```diff
@@ -124,6 +124,20 @@ def metafunc(self, backend, *args, **kwargs):
...
+    if molularize:
```
kripken, line 132:
```suggestion
    if modularize:
```

kripken (commented):
lgtm otherwise

PR #23581, opened 2025-02-04
Fix JSPI suspend through call from side module to main module

hoodmane:
This fixes a basic use case where a side module calls a suspending main module symbol. Without this change, we call through a promising JS wrapper produced by `instrumentWasmExports()` and fail to suspend.

We have to dlsym() the side module symbol since if we call it directly it will fail because of another JS trampoline on the main -> side module edge.


src/lib/libdylink.js
```diff
@@ -715,7 +715,13 @@ var LibraryDylink = {
           }
           if (prop in wasmImports && !wasmImports[prop].stub) {
             // No stub needed, symbol already exists in symbol table
-            return wasmImports[prop];
+            var res = wasmImports[prop];
+#if JSPI
+            if ('orig' in res) {
```
sbc100, line 720: I think maybe just `if (res.orig)` would would here?   (Oh I sees this is copied from line 1250.. maybe update both  of them?)

sbc100: Is this not needed for `ASYNCIFY=1` too?  The version on line 1250 just uses `if ASYNCIFY`.   Can you copy the comment from there too?

test/other/test_jspi_dylink_main.c
```diff
@@ -0,0 +1,29 @@
...
+
+EM_ASYNC_JS(int, test, (), {
+    console.log("sleeping");
+    await new Promise(res => setTimeout(res, 500));
+    console.log("slept");
+    return 77;
+});
+
+int test_wrapper() {
+    return test();
```
sbc100, line 18: Wrong indentation here and above.

test/other/test_jspi_dylink_side.c
```diff
@@ -0,0 +1,7 @@
+#include <stdio.h>
+int test_wrapper(void);
+
+int f() {
+    printf("f\n");
+    return test_wrapper();
```
sbc100, line 6: Wrong indentation

test/test_other.py
```diff
@@ -3411,6 +3411,15 @@ def test_jspi_add_function(self):
       '-Wno-experimental']
     self.do_runf('other/test_jspi_add_function.c', 'done')
 
+  @requires_jspi
```
sbc100, line 3414: Should we test this with both ASYNCIFY modes?

hoodmane: Is there a decorator to parameterize over the asyncify modes? Or do I just do it manually>

src/lib/libdylink.js
hoodmane: The test fails with `ASYNCIFY=1` in the same way regardless of this change.

hoodmane: Though it looks like the problem is probably fixable by adding another `blah = blah.orig` somewhere else.

hoodmane: Okay, I have no idea how to fix this test with `ASYNCIFY`, so I'd prefer to keep it focused on JSPI.

test/other/test_jspi_dylink_main.c
```diff
@@ -0,0 +1,29 @@
+// Copyright 2023 The Emscripten Authors.  All rights reserved.
```
sbc100, line 1: 2025

src/lib/libdylink.js
```diff
@@ -715,7 +715,14 @@ var LibraryDylink = {
           }
           if (prop in wasmImports && !wasmImports[prop].stub) {
             // No stub needed, symbol already exists in symbol table
-            return wasmImports[prop];
+            var res = wasmImports[prop];
+#if JSPI
```
sbc100, line 719: Do any tests fail if you make this `if ASYNCIFY` (if only to symmetry with below)

sbc100 (commented):
lgtm % comment

I do think it would be nice keep both ASYNCIFY mode in parity so maybe @kripken or myself can take a look after this lands.

test/test_other.py
```diff
@@ -3411,6 +3411,15 @@ def test_jspi_add_function(self):
...
+  def test_jspi_dylink(self):
+      self.run_process([
+        EMCC,
+        '-o', 'side.so',
+        test_file('other/test_jspi_dylink_side.c'),
+        '-sSIDE_MODULE'])
+      self.do_run_in_out_file_test('other/test_jspi_dylink_main.c', emcc_args=['side.so', '-sMAIN_MODULE=2', '-sJSPI'])
```
sbc100, line 3421: We do already have `test_dlfcn_asyncify` in `test_core.py`  perhaps this new test would make sense alongside that one?

sbc100: For reference we do also have `with_asyncify_and_jspi` in test_core.py

hoodmane: I can move it next to `test_dlfcn_asyncify`.

src/lib/libdylink.js
hoodmane: I doubt it, let's see.

src/lib/libdylink.js
```diff
@@ -715,7 +715,14 @@ var LibraryDylink = {
           }
           if (prop in wasmImports && !wasmImports[prop].stub) {
             // No stub needed, symbol already exists in symbol table
-            return wasmImports[prop];
+            var res = wasmImports[prop];
+#if ASYNCIFY || JSPI
```
sbc100, line 719: Just `if ASYNCIFY` is enough.  JSPI is the same things as ASYNCIFY == 2

test/test_core.py
```diff
@@ -3747,6 +3747,17 @@ def test_dlfcn_asyncify(self):
       '''
     self.do_run(src, 'before sleep\nafter sleep\n42\n')
 
+  @requires_jspi
+  @needs_dylink
+  def test_jspi_dylink(self):
```
sbc100, line 3752: Can you call this `test_dlfcn_jspi` maybe?

hoodmane: Ideally it would be nice to have a better description but it's hard to come up with anything concise enough to be a function name.

test/test_core.py
```diff
@@ -3743,6 +3743,8 @@ def test_dlfcn_asyncify(self):
   @requires_jspi
   @needs_dylink
   def test_dlfcn_jspi(self):
+    if self.get_setting('MEMORY64'):
+      self.skipTest('fails with MEMORY64')
```
sbc100, line 3747: Could you open a bug for that and use the bug URL as the argument to self.skipTest?

test/core/test_dlfcn_jspi_main.c
```diff
@@ -0,0 +1,29 @@
+// Copyright 2025 The Emscripten Authors.  All rights reserved.
+// Emscripten is available under two separate licenses, the MIT license and the
```
sbc100, line 2: Could you call this file just `test_dlfcn_jspi.c` , I think its useful to have the filename match the test name where possible.

test/core/test_dlfcn_jspi_side.c
```diff
@@ -0,0 +1,7 @@
+#include <stdio.h>
+int test_wrapper(void);
+
+int f() {
+  printf("f\n");
```
sbc100, line 5: Could you call this function something with more meaning.  How about `side_module_function`.. or also use `sidey` is some other tests.

Then the printf could print that name too.

test/test_core.py
hoodmane: Sounds good. #23588 seems related so it would make sense to test again after merging that.

test/core/test_dlfcn_jspi_side.c
hoodmane: Maybe `side_module_trampoline()` since it just calls back into the main module?

test/test_core.py
hoodmane: The error is `LinkError: WebAssembly.instantiate(): cannot import i64 memory as i32` so I wonder if the problem might just be that I don't pass `-sMEMORY64` when linking the side module.

hoodmane: Okay I fixed that and got a more real error which does sound like what @brendandahl is already working on.

hoodmane:
Since #23588 merged just now I guess I'll merge that and try again with memory64.

hoodmane:
That was the problem.

sbc100 (approved):
Awesome!

test/test_core.py
```diff
@@ -3758,6 +3758,20 @@ def test_dlfcn_asyncify(self):
...
+  def test_dlfcn_jspi(self):
+    side_flags = ['-sSIDE_MODULE']
+    if self.get_setting('MEMORY64'):
+      side_flags.append('-sMEMORY64')
+    self.set_setting('JSPI')
+    self.run_process([
+      EMCC,
+      '-o', 'side.so',
+      test_file('core/test_dlfcn_jspi_side.c'),
+      *side_flags])
```
sbc100, line 3772: Maybe better do put `-sSIDE_MODULE` here and then add `+ self.get_emcc_args()` to the end of the command.  That way to the global flags (such as `-sMEMORY64) will be added there.

test/test_core.py
```diff
@@ -3758,6 +3758,20 @@ def test_dlfcn_asyncify(self):
       '''
     self.do_run(src, 'before sleep\nafter sleep\n42\n')
 
+  @requires_jspi
+  @needs_dylink
+  def test_dlfcn_jspi(self):
+    side_flags = ['-sSIDE_MODULE']
+    if self.get_setting('MEMORY64'):
+      side_flags.append('-sMEMORY64')
+    self.set_setting('JSPI')
```
sbc100, line 3767: `requires_jspi` already takes care of setting this so I think this line can be removed.

test/test_core.py
```diff
@@ -3761,16 +3761,22 @@ def test_dlfcn_asyncify(self):
...
+              "-o",
+              "side.so",
+              test_file("core/test_dlfcn_jspi_side.c"),
+              "-sSIDE_MODULE",
+          ]
+          + self.get_emcc_args()
+      )
+      self.do_run_in_out_file_test(
+          "core/test_dlfcn_jspi.c", emcc_args=["side.so", "-sMAIN_MODULE=2"]
+      )
```
sbc100, line 3783: We use 2 space indention.   We also allow tend to allow long line in test code so I think these can bot be just single lines.

test/test_core.py
```diff
@@ -3761,16 +3761,22 @@ def test_dlfcn_asyncify(self):
...
-    self.set_setting('JSPI')
-    self.run_process([
-      EMCC,
-      '-o', 'side.so',
-      test_file('core/test_dlfcn_jspi_side.c'),
-      *side_flags])
-    self.do_run_in_out_file_test('core/test_dlfcn_jspi.c', emcc_args=['side.so', '-sMAIN_MODULE=2'])
+      side_flags = []
+      if self.get_setting("MEMORY64"):
+          side_flags.append("-sMEMORY64")
```
sbc100, line 3770: These 3 lines can be removed I think?

sbc100 (commented):
lgtm % final nits

sbc100:
Perhaps the PR title could be a little more specific than "a bit"?   What is being fixed here?

hoodmane:
How's this sound?

sbc100 (approved):

sbc100:
@brendandahl does this looks to you?

test/core/test_dlfcn_jspi.c
```diff
@@ -0,0 +1,29 @@
...
+// University of Illinois/NCSA Open Source License.  Both these licenses can be
+// found in the LICENSE file.
+
+#include <emscripten.h>
+#include <stdio.h>
+#include <dlfcn.h>
+
+EM_ASYNC_JS(int, test, (), {
+  console.log("sleeping");
+  await new Promise(res => setTimeout(res, 500));
```
brendandahl, line 12: I'd say either make the timeout 0 or 1. Either will give us a turn of the event loop and should test what we want. No reason the artificially slow down the test.

brendandahl (approved):
Looks good, one minor comment.

test/core/test_dlfcn_jspi.c
hoodmane: I originally set it this high during manual testing since I wanted a perceptible delay. And then I forgot about it.

hoodmane:
Thanks @sbc100 and @brendandahl for the reviews!

Issue #20289, opened 2023-09-19
Async JS library funcs should not be marked for Asyncify transform


RReverser:
After #19093 any `async function () { }` in JS libraries is marked for Asyncify transform. It's a bit hard to describe why, but I don't think this is correct:

1) For Asyncify (`-s ASYNCIFY`), functions must be wrapped into `Asyncify.handleAsync` to work correctly, which would make them not recognised by this check. And, vice versa, if an unwrapped `async function () {}` is recognised by this check, as it is now, it will be passed to Binaryen and transformed - which expands the file size - for no reason, since it's going to be called only directly and not inside `Asyncify.handleAsync`/`Asyncify.handleSleep`.
2) For JSPI, we can sort of get away with it since it's much less reliant on wrapping, so it at least seems to work, but since such functions are still not wrapped into `handleAsync`, they don't get the runtime keepalive counter increased/decreased upon entering/exit correspondingly.
3) For `PROXY_TO_PTHREAD` mode, such functions need to be treated as `void` as async wakeup is happening manually via `proxying.h` functions. Which means that again, they shouldn't be marked as Asyncify functions.

For example, looking at current usages in https://github.com/emscripten-core/emscripten/blob/992d1e91d5c0af2f21dd1ee3d572a03c6727122c/src/library_wasmfs_jsimpl.js#L92-L99, if someone compiles this with `-s ASYNCIFY -s PROXY_TO_PTHREAD` (because their project depends on both of those features), right now this function will be passed to and transformed by Binaryen as Asyncify function, increasing the Wasm size, but then never used as Asyncify function because it's not wrapped into `Asyncify.handleAsync`.

This seems like a footgun.

I'd suggest that we should either 1) require users to explicitly mark functions as `__async: true`, which is easy to put into `#ifdef` so that it only activates when the function is really meant to be used with Asyncify or 2) provide some more magic & automatic wrapping for all async functions that would perform `proxying.h`-based wakeup in `PROXY_TO_PTHREAD` mode and Asyncify-based wakeup otherwise.

cc @brendandahl

brendandahl:
[...]

PR #23588, opened 2025-02-05
[jspi] Fix dlopen with MEMORY64 and JSPI.

brendandahl:
Adds the various `async` and `awaits` before the conversion to BigInt.

### DIFF
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -1159,9 +1159,9 @@ var LibraryDylink = {
 #if ASYNCIFY
   _dlopen_js__async: true,
 #endif
-  _dlopen_js: (handle) => {
+  _dlopen_js: {{{ asyncIf(ASYNCIFY == 2) }}} (handle) => {
 #if ASYNCIFY
-    return Asyncify.handleSleep((wakeUp) => {
+    return {{{ awaitIf(ASYNCIFY == 2) }}} Asyncify.handleSleep((wakeUp) => {
       dlopenInternal(handle, { loadAsync: true })
         .then(wakeUp)
         .catch(() => wakeUp(0));
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -1161,7 +1161,7 @@ var LibraryDylink = {
 #endif
   _dlopen_js: {{{ asyncIf(ASYNCIFY == 2) }}} (handle) => {
 #if ASYNCIFY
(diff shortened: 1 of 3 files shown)


src/lib/libdylink.js
```diff
@@ -1159,9 +1159,9 @@ var LibraryDylink = {
 #if ASYNCIFY
   _dlopen_js__async: true,
 #endif
-  _dlopen_js: (handle) => {
+  _dlopen_js: {{{ asyncIf(ASYNCIFY == 2) }}} (handle) => {
 #if ASYNCIFY
-    return Asyncify.handleSleep((wakeUp) => {
+    return {{{ awaitIf(ASYNCIFY == 2) }}} Asyncify.handleSleep((wakeUp) => {
```
sbc100, line 1164: if the `await` is being done here why does it also need to be done in in the wrapper generated by jsifier?

Should we not do just one of them?

src/lib/libdylink.js
```diff
@@ -1159,9 +1159,9 @@ var LibraryDylink = {
 #if ASYNCIFY
   _dlopen_js__async: true,
 #endif
-  _dlopen_js: (handle) => {
+  _dlopen_js: {{{ asyncIf(ASYNCIFY == 2) }}} (handle) => {
```
sbc100, line 1162: Does `__async` not imply this?  If not, what *does* `__async` do?

brendandahl: `__async` just tells emscripten to wrap the import with JSPI. Not requiring `async function` was an optimization that allowed for synchronous returns. However, JSPI is now going to suspend no matter what the return value is, so we could now probably change it to imply `async` or require the function to be marked `async`.

We talked about getting rid of `__async`, and just using `async function`, but there are cases where we don't actually want to wrap with JSPI if the function is `async`. More discussion in https://github.com/emscripten-core/emscripten/issues/20289

src/lib/libdylink.js
brendandahl: Ah yeah, started with a different approach and this wasn't needed anymore.

src/lib/libdylink.js
brendandahl: I'd say land this first. I probably won't have time to look into the bigger overhaul for a bit.

sbc100: Fair enough.  Resolve this then

sbc100 (approved):

PR #23589, opened 2025-02-05
[test] Use assert and zero exit code in more browser tests. NFC

sbc100:


test/browser/test_force_exit.c
```diff
@@ -10,31 +10,19 @@
...
-
 void EMSCRIPTEN_KEEPALIVE later() {
   printf("later, now force an exit\n");
-  result += 10;
-  emscripten_force_exit(result);
+  emscripten_force_exit(0);
 }
 
 int main() {
-  atexit(success);
```
kripken, line 27: This seems to no longer test that the atexit is executed?

sbc100: I'm not sure the test was checking that before since the result would be reported to the server by the onExit handler which was triggered by `emscripten_force_exit`. 

I guess that atexit handler would be called before the onExit.. but I"m not sure this test verifies that.   I think maybe I can update it so that we check for the output messages.

sbc100: I split this change out into #23597

sbc100:
Rebased on top of #23597. PTAL

kripken (approved):

PR #23592, opened 2025-02-05
[test] Remove test/emscripten_log/emscripten_log_with_closure.cpp. NFC

sbc100:
This test was added in #5808, but we can do better by just running the entire emscripten_log test in closure.


test/test_core.py
```diff
@@ -7867,11 +7867,9 @@ def test_modularize_closure_pre(self):
   @also_with_wasm_bigint
   def test_emscripten_log(self):
     self.emcc_args += ['-g', '-DRUN_FROM_JS_SHELL', '-Wno-deprecated-pragma']
-    self.do_run_in_out_file_test('emscripten_log/emscripten_log.cpp', interleaved_output=False)
-    # test closure compiler as well
     if self.maybe_closure():
-      self.emcc_args += ['-g1'] # extra testing
```
kripken, line 7873: the comment here was useful, in that it clarifies the purpose of `-g1` is for test coverage, and not necessary to build or something like that?

sbc100: Hmm.. I actually thought it was there to make it compile.. but I must have been mistaken.  Restored comment.

kripken (approved):

Issue #23591, opened 2025-02-05
Prefer `libx.a` over `libx.so` when dynamic linking is off?


hoodmane:
When using Emscripten 4.0.0 and 4.0.1, I'm getting errors like `attempted static link of dynamic object ./libx.so`. I haven't diagnosed what changed in Emscripten that causes this, this is a feature request not a report of a regression. The following reproduces the problem:
```console
$ ar -cq libx.a
$ touch a.c
$ emcc a.c -L. -lx
$ echo $?
0
$ touch x.c
$ emcc -sSIDE_MODULE x.c -o libx.so
$ emcc a.c -L. -lx
wasm-ld: error: attempted static link of dynamic object ./libx.so
```
I would like it if instead of complaining that it can't use `libx.so` it would find `libx.a` and link with that.

sbc100:
I think this was an change in wasm-ld: https://github.com/llvm/llvm-project/pull/108263

I'll see if its reasonable to avoid the error in that case.

sbc100:
Actually tracked this down a bug in emscripten.   Fix on the way.

hoodmane:
Thanks!

PR #23593, opened 2025-02-05
Skip the search for dynamic libraries when static library exists

sbc100:
This is a fix for a regression caused by #23336.  

Prior to #23336 all `-l` flags were expanded by emcc to full filenames, with `.a` files being searched for first.   After #23336 we only expand the the paths of fake shared libraries which wasm-ld itself don't look for.   

Fixes: #23591

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -6736,7 +6736,7 @@ This locale is not the C locale.
       if be_clean:
         assert len(clutter) == 0, 'should not leave clutter ' + str(clutter)
       else:
-         assert len(clutter) == 2, 'should leave .o files'
+        assert len(clutter) == 2, 'should leave .o files'
     test(['-o', 'c.so', '-r'], True)
     test(['-o', 'c.js'], True)
     test(['-o', 'c.html'], True)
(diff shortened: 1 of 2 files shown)


test/test_other.py
```diff
@@ -9322,6 +9322,19 @@ def test_side_module_folder_deps(self):
     self.run_process([EMCC, test_file('hello_world.c'), '-sSIDE_MODULE', '-o', 'subdir/libside2.so', '-L', 'subdir', '-lside1'])
     self.run_process([EMCC, test_file('hello_world.c'), '-sMAIN_MODULE', '-o', 'main.js', '-L', 'subdir', '-lside2'])
 
+  def test_side_module_ignore(self):
+    self.run_process([EMCC, test_file('hello_world.c'), '-sSIDE_MODULE', '-o', 'libside.so'])
+
+    # Attempting link statically against a side module (libside.so) should fail.
```
kripken, line 9328:
```suggestion
    # Attempting to link statically against a side module (libside.so) should fail.
```

kripken (approved):
Is this a regression? If so from what?

sbc100:
Yes, its a regression.  Do you mean what change caused it?   I would have to bisect to find out.

kripken:
Yes, I think it would be useful to know what change caused the regression. Perhaps there is more context there (like a reason this was changed).

sbc100:
The regression was caused by  #23336 which changed the way emcc itself expands `-l` flags.

PR #23596, opened 2025-02-05
[Wasm64] Automatically add node memory64 flag for mem64 tests.

brendandahl:


test/common.py
```diff
@@ -1003,6 +1003,7 @@ def require_wasm64(self):
       version = shared.get_node_version(nodejs)
       if version >= (24, 0, 0):
         self.js_engines = [nodejs]
+        self.node_args += ['--experimental-wasm-memory64']
```
sbc100, line 1006: Is this needed?  It doesn't seem to be needed in CI.   I guess I had though that this flag was not needed with node-canary?

sbc100: Maybe you need to update your local version of node canary to match the CI version?

brendandahl: I needed it locally to run with `node-v24.0.0-v8-canary202411045365c9d9be-linux-x64`. Maybe newer versions it's not needed?

sbc100: Yup, I would update to canary202501036428a2737e which is used by the CI.  I think older version don't have the full memory64 spec implemented.

PR #23601, opened 2025-02-05
[test] Fix test_side_module_ignore on windows. NFC

sbc100:
This slipped past CI because our windows builder we out of commission at the time when #23593 landed.


sbc100:
Landing TBR since the windows bot has now passed.

dschuff (commented):
LGTM

PR #23603, opened 2025-02-05
Make WASMFS FS.createPath match the FS version behavior on EEXIST

### LINKED ISSUE #23602: WASMFS FS.createPath doesn't ignore EEXIST
https://github.com/emscripten-core/emscripten/blob/db1b8c77f3ae8e486d6d4d2fbb5ffbc8b743c12e/src/lib/libwasmfs.js#L110-L116

VS

https://github.com/emscripten-core/emscripten/blob/db1b8c77f3ae8e486d6d4d2fbb5ffbc8b743c12e/src/lib/libfs.js#L1623-L1630

This is probably a bug for two reasons.  First, the libfs.js version *does* check for and ignore EEXIST (and any other errors besides).  Second, the file packager generates JS that calls FS_createPath repeatedly in a way that will trigger an exception if any of the directories exist already.

```
Module['FS_createPath']("/", "home", true, true);
Module['FS_createPath']("/home", "web_user", true, true);
Module['FS_createPath']("/home/web_user", "assets", true, true);
```

JoeOsborn:
Fixes #23602.

### DIFF
--- a/src/lib/libwasmfs.js
+++ b/src/lib/libwasmfs.js
@@ -111,7 +111,11 @@ FS.init();
         if (!wasmFSPreloadingFlushed) {
           wasmFSPreloadedDirs.push({parentPath: parent, childName: part});
         } else {
-          FS.mkdir(current);
+          try {
+            FS.mkdir(current);
+          } catch (e) {
+            if (e.errno != {{{ cDefs.EEXIST }}}) throw e;
+          }
         }
         parent = current;
       }
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -1881,6 +1881,22 @@ simulateKeyUp(100, undefined, 'Numpad4');
     shutil.copy(Path('sub/test.data'), '.')
     self.btest_exit('test_emscripten_async_load_script.c', args=['-sFORCE_FILESYSTEM'])
 
+  @also_with_wasmfs
+  def test_emscripten_overlapped_package(self):
+    def setup():
+      create_file('script1.js', '''
+        Module._set(456);
+      ''')
+      ensure_dir('sub')
+      create_file('sub/file1.txt', 'first')
+      create_file('sub/file2.txt', 'second')
+
+    setup()
+    self.run_process([FILE_PACKAGER, 'test.data', '--preload', 'sub/file1.txt@/target/file1.txt'], stdout=open('script2.js', 'w'))
+    self.run_process([FILE_PACKAGER, 'test2.data', '--preload', 'sub/file2.txt@/target/file2.txt'], stdout=open('script3.js', 'w'))
+    self.btest_exit('test_emscripten_overlapped_package.c', args=['-sFORCE_FILESYSTEM'])
+    self.clear()
+
   def test_emscripten_api_infloop(self):
     self.btest_exit('emscripten_api_browser_infloop.cpp', assert_returncode=7)
 
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -1883,17 +1883,17 @@ simulateKeyUp(100, undefined, 'Numpad4');
 
   @also_with_wasmfs
   def test_emscripten_overlapped_package(self):
+    # test that a program that loads multiple file_packager.py packages has a correctly initialized filesystem.
+    # this exercises https://github.com/emscripten-core/emscripten/issues/23602 whose root cause was a difference
+    # between JS FS and WASMFS behavior.
     def setup():
-      create_file('script1.js', '''
-        Module._set(456);
-      ''')
       ensure_dir('sub')
       create_file('sub/file1.txt', 'first')
       create_file('sub/file2.txt', 'second')
 
     setup()
-    self.run_process([FILE_PACKAGER, 'test.data', '--preload', 'sub/file1.txt@/target/file1.txt'], stdout=open('script2.js', 'w'))
-    self.run_process([FILE_PACKAGER, 'test2.data', '--preload', 'sub/file2.txt@/target/file2.txt'], stdout=open('script3.js', 'w'))
+    self.run_process([FILE_PACKAGER, 'test.data', '--preload', 'sub/file1.txt@/target/file1.txt'], stdout=open('script1.js', 'w'))
+    self.run_process([FILE_PACKAGER, 'test2.data', '--preload', 'sub/file2.txt@/target/file2.txt'], stdout=open('script2.js', 'w'))
     self.btest_exit('test_emscripten_overlapped_package.c', args=['-sFORCE_FILESYSTEM'])
     self.clear()
 
--- a/test/test_emscripten_overlapped_package.c
+++ b/test/test_emscripten_overlapped_package.c
@@ -1,0 +1,66 @@
+// Copyright 2013 The Emscripten Authors.  All rights reserved.
+// Emscripten is available under two separate licenses, the MIT license and the
+// University of Illinois/NCSA Open Source License.  Both these licenses can be
+// found in the LICENSE file.
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <assert.h>
+
+#include <emscripten.h>
+
+int value = 0;
+
+EMSCRIPTEN_KEEPALIVE void set(int x) {
+  printf("set! %d\n", x);
+  value = x;
+}
+
+void load3() {
+  printf("load3\n");
+  char buffer[10];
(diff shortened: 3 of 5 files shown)


src/lib/libwasmfs.js
```diff
@@ -111,7 +111,11 @@ FS.init();
         if (!wasmFSPreloadingFlushed) {
           wasmFSPreloadedDirs.push({parentPath: parent, childName: part});
         } else {
-          FS.mkdir(current);
+          try {
+            FS.mkdir(current);
+          } catch (e) {
+            if (e.errno != {{{ cDefs.EEXIST }}}) throw e;
+          }
```
sbc100, line 118: Is there some test that can be updated now?   Maybe some test we couldn't run before but now we can?

JoeOsborn:
I've added a test.  The trigger I noticed is when you load multiple packages which overlap at all in their target paths, or when you load a package into a filesystem that already has content (even if it doesn't overlap with the actual package contents).

createPath was also just less pleasant to use in WASMFS compared to JS FS due to this difference.

test/test_emscripten_overlapped_package.c
```diff
@@ -0,0 +1,66 @@
+// Copyright 2013 The Emscripten Authors.  All rights reserved.
```
kripken, line 1:
```suggestion
// Copyright 2025 The Emscripten Authors.  All rights reserved.
```

test/test_browser.py
```diff
@@ -1881,6 +1881,22 @@ def setup():
     shutil.copy(Path('sub/test.data'), '.')
     self.btest_exit('test_emscripten_async_load_script.c', args=['-sFORCE_FILESYSTEM'])
 
+  @also_with_wasmfs
+  def test_emscripten_overlapped_package(self):
+    def setup():
```
kripken, line 1886: Please add comments explaining what the test is testing.

test/test_browser.py
```diff
@@ -1881,6 +1881,22 @@ def setup():
...
+      create_file('script1.js', '''
+        Module._set(456);
```
kripken, line 1888: IIUC the test is testing for overlapping filenames. How is this `_set()` part helping?

test/test_browser.py
```diff
@@ -1881,6 +1881,22 @@ def setup():
     shutil.copy(Path('sub/test.data'), '.')
     self.btest_exit('test_emscripten_async_load_script.c', args=['-sFORCE_FILESYSTEM'])
 
+  @also_with_wasmfs
+  def test_emscripten_overlapped_package(self):
```
sbc100, line 1885: Perhaps we can instead test the JS API directly?   i.e. write a test that just called `createPath` twice on the same path?   I think there are already some JS API tests for the FS API.

src/lib/libwasmfs.js
```diff
@@ -111,7 +111,11 @@ FS.init();
         if (!wasmFSPreloadingFlushed) {
           wasmFSPreloadedDirs.push({parentPath: parent, childName: part});
         } else {
-          FS.mkdir(current);
+          try {
+            FS.mkdir(current);
+          } catch (e) {
+            if (e.errno != {{{ cDefs.EEXIST }}}) throw e;
```
sbc100, line 117: I looks like the code in `libjs.js` is even more relaxed than this: 

```
        try {                                                                         
          FS.mkdir(current);                                                          
        } catch (e) {                                                                 
          // ignore EEXIST                                                            
        }   
```

I guess they should at least be consistent?

JoeOsborn: I feel like it intended to specifically ignore EEXIST, but was written in a buggy way; it’s plausible existing code relies on that permissiveness so I hesitated to change it.

test/test_browser.py
JoeOsborn: Yeah, that makes sense.  I thought instead of a test that just does createPath twice, I would make a test that showed why the difference in behavior was a bug.

test/test_browser.py
JoeOsborn: My bad, it’s a leftover from the async loading test I copied to get this test started.  I’ll reduce it.

src/lib/libwasmfs.js
sbc100: I think we we should change libfs.js too.  If the tests pass that should be good signal that its good.   I don't think there are very many users of this API outside of the file packager.

JoeOsborn:
So I didn't add a test specifically for createPath, because I do like that the current test exercises the separate file packager and createPath isn't documented to succeed or fail with respect to whether the paths exist already.  I can add a separate JS test if you think it's important.

src/lib/libwasmfs.js
JoeOsborn: Done!

test/test_browser.py
sbc100: I think this test is a little too high level for this fix.  I don't think we need/want to involve the file packager to get this testing.   Would it be OK to just update `test_fs_js_api.c` instead of adding this new test?

JoeOsborn:
The fs_js_api tests seem a bit broken.  They don't work in strict mode (and changing the octal literals to 0o777 causes breakage in other conditions) or WASM64, and there's a memory leak in wasmfs found during read.  Do you still want me to add a WASMFS condition to test_fs_js_api?

sbc100:
> The fs_js_api tests seem a bit broken. They don't work in strict mode (and changing the octal literals to 0o777 causes breakage in other conditions) or WASM64, and there's a memory leak in wasmfs found during read. Do you still want me to add a WASMFS condition to test_fs_js_api?

My understanding is the test_fs_js_api was added precisely so that we could test the wasmfs versions of the APIs, so I think we should fix them if they are broken, yes.   I'm a little confused as to why its not marked as `@also_with_wasmfs` but I guess its run explicitly as `wasmfs.test_fs_js_api` in circleci.

test/test_browser.py
JoeOsborn: Even if this test is too high level for this fix, I do think it's an important regression test to have since having multiple packages is valuable and there's no obvious reason they couldn't e.g. mount to /assets/pack-a and /assets/pack-b.  Can we keep it there as well as the js_fs_api test?

test/third_party/googletest
JoeOsborn, line 1: Sorry, I didn't mean to include this.

test/test_browser.py
sbc100: Fair enough.   Lets group it with the other file packager tests then.   How about calling it `test_file_packager_overlapping`?

And can you move it to test_other with all the other file_packager tests (below `test_file_packager_modularize`)

JoeOsborn: Done!

kripken (approved):

JoeOsborn:
So I tried to move my packager test into test_other, but it actually only breaks in browser (node has its own filesystem stuff used by the file_packager scripts).  Is there a good way to write a test_other test that still runs in a browser environment?  Currently my packager test is still in test_core

sbc100:
> So I tried to move my packager test into test_other, but it actually only breaks in browser (node has its own filesystem stuff used by the file_packager scripts). Is there a good way to write a test_other test that still runs in a browser environment? Currently my packager test is still in test_core

No, the point of test_other is that doesn't run in the browser.

src/lib/libwasmfs.js
```diff
@@ -111,7 +111,11 @@ FS.init();
         if (!wasmFSPreloadingFlushed) {
           wasmFSPreloadedDirs.push({parentPath: parent, childName: part});
         } else {
-          FS.mkdir(current);
+          try {
+            FS.mkdir(current);
```
sbc100, line 115: I'm actually little confused by this fix.   I tried running `wasmfs.test_fs_js_api` and it passed both with or without this change.

Looking into it it looks the repeated calls to  `FS.mkdir` here always seem to succeed.  I modified the code here to add  some extra debugging and I see:

```

mkdir /home/nested1
0
mkdir /home/nested2
0
```

I'm a little confused how you were hitting an error here?   I must be missing something.

sbc100: Hmm even stranger, browser.test_emscripten_overlapped_package seem to pass both before and after this test.

sbc100: Ah I wasn't running `browser.test_emscripten_overlapped_package_wasmfs`.. that one does fail.

PR #23607, opened 2025-02-06
Fix pthread worker options usage with Vite bundler

jamsinclair:
### Overview

An attempt at solving #22394.

I had previously opened an attempt at #22834, however, it got a bit messy with the commits. I have opened a new PR with a cleaner context which should be easier to focus on the reasoning and necessary changes.

### The Problem

When code using threads is compiled with Emscripten and then bundled with Vite it does not bundle and throws an error.

This is due to a variable reference for worker options being used during worker construction. Vite needs inline static worker options to bundle the code successfully.

e.g.
```js
const workerOptions = { type: 'module',  name: 'foo'  } ;

// ❌ this will break vite build as the worker options are dynamic
new Worker(new URL('./my-worker.js', import.meta.url), workerOptions); 

// ✅ The literal worker option values can be statically analysed by Vite 
// and bundled correctly
new Worker(new URL('./my-worker.js', import.meta.url), {
  type: 'module',
  name: 'foo'
}); 
```

### The Solution

Emscripten needs to output literal static values instead of using the `workerOptions` variable.


src/lib/libpthread.js
```diff
@@ -427,14 +408,50 @@ var LibraryPThread = {
...
+#if ENVIRONMENT_MAY_BE_NODE
+            // This is the way that we signal to the node worker that it is hosting
+            // a pthread.
+            'workerData': 'em-pthread',
+#endif
+#if ENVIRONMENT_MAY_BE_WEB || ENVIRONMENT_MAY_BE_WORKER
+            // This is the way that we signal to the Web Worker that it is hosting
+            // a pthread.
+#if ASSERTIONS
+            'name': 'em-pthread-' + PThread.nextWorkerID,
```
jamsinclair, line 424: Previously @sbc100 you had kindly suggested making use of a macro.
```
{{{
globalThis.WORKER_OPTIONS = {
   ...
};
null;
}}}
```

This would have been a great solution, however, because we use a dynamic `PThread.nextWorkerID` reference I couldn't think of a way to make it work. Let me know if you have any other ideas.

test/test_other.py
```diff
@@ -380,7 +380,15 @@ def test_emcc_output_worker_mjs(self, args):
...
-    self.assertContained("new Worker(new URL('hello_world.mjs', import.meta.url), workerOptions)", src)
+    self.assertContained("""new Worker(new URL('hello_world.mjs', import.meta.url), {
+        'type': 'module',
+        // This is the way that we signal to the node worker that it is hosting
+        // a pthread.
+        'workerData': 'em-pthread',
+        // This is the way that we signal to the Web Worker that it is hosting
+        // a pthread.
+        'name': 'em-pthread',
+    })""", src)
```
sbc100, line 391: Rather than checking for this specific output I wonder if we can instead test directly that it works with vite.

I recently added some vite tests to test_browser.py so I think we could add them there.   Then we could test that it actually runs though vite and executed.

I also wonder if its enough to test this in just rollup, since IIUC vite uses rollup?

sbc100 (commented):
Thanks for working on this!  I'm excited to get more test coverage here.

test/test_other.py
jamsinclair: Awesome! I'll look into adding some tests there 👀 

For this particular Vite issue, rollup would not be enough to test. The special logic around these workers is implemented on the Vite side.

test/test_browser.py
```diff
@@ -5507,8 +5507,13 @@ def test_webpack(self, es6):
     shutil.copy('webpack/src/hello.wasm', 'webpack/dist/')
     self.run_browser('webpack/dist/index.html', '/report_result?exit:0')
 
-  def test_vite(self):
+  @parameterized({
+    '': ([],),
+    'pthreads': (['-pthread', '-sPTHREAD_POOL_SIZE=1', '-sENVIRONMENT=web,worker'],),
```
jamsinclair, line 5512: We need to compile for only browser based environments, otherwise Vite runs into issues with bundling.

Perhaps it could be worth moving `-sEnvironment` to the default args array below.

package.json
```diff
@@ -11,7 +11,7 @@
     "rollup": "^4.34.1",
     "source-map": "0.7.4",
     "typescript": "^5.7.3",
-    "vite": "^6.0.11",
+    "vite": "^6.1.0",
```
jamsinclair, line 14: It's important we bump Vite to at least `6.1.0` for the test build.

We introduced a fix to Vite in https://github.com/vitejs/vite/pull/19010 to allow dynamic fields in the literal worker options. This allows Emscripten to keep on using `name: 'em-pthread-' + PThread.nextWorkerID,` for debugging.

sbc100: Nice!  I just updated our deps 3 days ago in #23574, but I'll do it again now as a separate PR I think,

sbc100: https://github.com/emscripten-core/emscripten/pull/23610

test/test_other.py
sbc100: Good to know.  Thanks for working on this!

As a followup question.  Is there some way we can test this stuff using node, or do we really need to test the output in a browser?  I test to prefer non-browser-tests since they are much easy to work with, but maybe this is a case where we really need the browser?

src/lib/libpthread.js
sbc100: I'm pretty sure we can still find a way to avoid these 4-fold duplication.    I'll see if I can some up with something

jamsinclair: Is it possible to capture a raw string and then output that in these templates? Is there a particular templating framework being used? 🤔

test/test_other.py
jamsinclair: For the particular issue concerning the PR, I think the browser is the only way.

Vite is primarily a bundler for the browser, although, as it is built on top of rollup it can be customised and you could bundle a JS file only and test with node (providing the bundled code is Node compatible).

I think the [documentation for bundling a library](https://vite.dev/guide/build#library-mode) could possibly be used by some tests and then tested on the command line with node. I could try create PoC, if you like?

package.json
jamsinclair: Awesome, thanks! I've synced those changes in.

test/test_other.py
sbc100: That would be most welcome.

BTW, are you a vite contributor/maintainer?

jamsinclair: > are you a vite contributor/maintainer?

No, I'm not. I have contributed a couple of small updates to Vite to fix issues with Emscripten. The Vite team has been very responsive to helping solve them (https://github.com/vitejs/vite/pull/14113, https://github.com/vitejs/vite/pull/19010).

sbc100:
I took a stab at doing this with a macro: #23707

jamsinclair:
Ah that's a lot cleaner! Thanks @sbc100 🙌 

I'll close this PR then.

PR #23613, opened 2025-02-06
Move test_emscripten_log. NFC

sbc100:
This test is the only thing left in the subdirectory.  This change          
also fixes search that was being done on the generated stackstrace.         
The old code was simply searching for `_emscripten_log`, which happens   
to be substring of `test_emscripten_log`.  Since we are already using       
more advanced regex below to extract the symbol name we can just use        
that instead.


src/lib/libstack_trace.js
```diff
@@ -15,8 +15,8 @@ var LibraryStackTrace = {
     // Find the symbols in the callstack that corresponds to the functions that
     // report callstack information, and remove everything up to these from the
     // output.
-    var iThisFunc = callstack.lastIndexOf('_emscripten_log');
-    var iThisFunc2 = callstack.lastIndexOf('_emscripten_get_callstack');
+    var iThisFunc = callstack.lastIndexOf(' _emscripten_log ');
+    var iThisFunc2 = callstack.lastIndexOf(' _emscripten_get_callstack ');
```
kripken, line 19: Is this always correct? Do some engines not have something like `func@src.js:10:20` (with no space after `func`)?

Also, is this testable?

sbc100: You are correct, it was broken in FF.

Updated the test and the PR description accordingly.

kripken: > Updated the test

Where is that part? I can't see it.

sbc100: Basically I move the check/test for `_emscripten_log` / `_emscripten_get_callstack` down to where we use proper regexes to find the symbols names.

kripken (approved):

PR #23614, opened 2025-02-06
Simplify EGL/GLFW/SDL/GLUT/Browser canvas access. NFC

sbc100:
All of these frameworks still use `Module['canvas']`.  This change simplifies and unifies the access to the canvas to single helper function per library.

As a followup we may want to extend this to all use of `Module['canvas']`.

### DIFF
--- a/src/lib/libbrowser.js
+++ b/src/lib/libbrowser.js
@@ -37,6 +37,8 @@ var LibraryBrowser = {
     preloadedImages: {},
     preloadedAudios: {},
 
+    getCanvas: () => Module['canvas'],
+
     init() {
       if (Browser.initted) return;
       Browser.initted = true;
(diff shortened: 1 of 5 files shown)


kripken (commented):
What is the impact on code size?

sbc100:
> What is the impact on code size?

We don't have codesize tests for any of these frameworks, but I imagine it can only be positive since it avoid duplicating the string `canvas` a bunch of times.   I'll do a comparison for a simple SDL example.

sbc100:
For browser.test_sdl_alloctext I see a tiny regression from 157890 to 157905 bytes of JS (0.009%)

sbc100:
For `test_sdl_image_prepare_data_closure_webonly` that uses closure I see a size reduction from 98389 to 98326 (0.06%)

kripken (approved):
Sounds good, thanks for measuring.

PR #23617, opened 2025-02-07
[WASMFS] Fix remaining TODO items with wasmfs + noderawfs

sbc100:
In order make sure this gets tested I added the combination of wasmfs
and rawfs to `@with_all_fs`.

Fixes: #23618


kripken (approved):

sbc100:
I ended up just fixing the resulting failures, so this change now includes a bunch of wasmfs fixes.

system/lib/wasmfs/backends/node_backend.cpp
```diff
@@ -45,7 +45,7 @@ class NodeState {
           result = _wasmfs_node_open(path.c_str(), "r");
           break;
         case O_WRONLY:
-          result = _wasmfs_node_open(path.c_str(), "w");
+          result = _wasmfs_node_open(path.c_str(), "r+");
```
kripken, line 48:
```suggestion
          result = _wasmfs_node_open(path.c_str(), "w");
```
or am I missing something?

sbc100: Ah yes, this part at least warrents a comment.

The short answer is that I don't think `O_WRONLY` on its own can be supported (at least not easily).  From the `fopen` man page:

```

       The file descriptor associated with the stream is opened as if by a call to open(2) with the following flags:
              ┌──────────────┬───────────────────────────────┐
              │ fopen() mode │ open() flags                  │
              ├──────────────┼───────────────────────────────┤
              │      r       │ O_RDONLY                      │
              ├──────────────┼───────────────────────────────┤
              │      w       │ O_WRONLY | O_CREAT | O_TRUNC  │
              ├──────────────┼───────────────────────────────┤
              │      a       │ O_WRONLY | O_CREAT | O_APPEND │
              ├──────────────┼───────────────────────────────┤
              │      r+      │ O_RDWR                        │
              ├──────────────┼───────────────────────────────┤
              │      w+      │ O_RDWR | O_CREAT | O_TRUNC    │
              ├──────────────┼───────────────────────────────┤
              │      a+      │ O_RDWR | O_CREAT | O_APPEND   │
              └──────────────┴───────────────────────────────┘
```

Note: there is not way to open write only without either truncation, or appending.    So I chose to just open in read/write mode..

sbc100: Done

system/lib/wasmfs/backends/node_backend.cpp
```diff
@@ -45,8 +45,13 @@ class NodeState {
           result = _wasmfs_node_open(path.c_str(), "r");
           break;
         case O_WRONLY:
-          result = _wasmfs_node_open(path.c_str(), "w");
-          break;
+          // TODO(sbc): Specific handling of O_WRONLY.
+          // There is no simple way to map O_WRONLY to an fopen-style
+          // mode string since the only two modes that are write only
+          // are `w` and `a`.  The problem with the former is that it
+          // truncates to file.  The problem with the later is that it
```
kripken, line 52:
```suggestion
          // truncates to file.  The problem with the latter is that it
```

kripken (approved):

hoodmane:
Okay. I will try building python against wasmfs and see what happens in the test suite.

Issue #23600, opened 2025-02-05
Should JSPI be tested without `--experimental-wasm-type-reflection`?


hoodmane:
Currently `require_jspi()` adds `--experimental-wasm-stack-switching --experimental-wasm-type-reflection`. But (I hope that) JSPI is going to stage 4 pretty soon and type reflection seems like it'll stay in stage 3 a while longer. So if JSPI support relies on type reflection, it may delay people being able to use it.

sbc100:
Good point, I think its was previously needed but no longer.  I think it can be removed.   @brendandahl can confirm.  Would you like to send a PR?

hoodmane:
Yeah I can send a pr. Though I mention this because wasm type reflection makes fixing #23395 more complicated. So I was thinking about first doing a fix that requires type reflection and then removing the dependency on type reflection.

brendandahl:
JSPI should no longer require type reflection.

PR #23619, opened 2025-02-07
Fix late-binding symbols with JSPI

### LINKED ISSUE #23395: JSPI and dynamic linking messed up by stub trampoline
The dynamic linking code will generate a stub if it's not resolved when first imported:
```js
          // Return a stub function that will resolve the symbol
          // when first called.
          if (!(prop in stubs)) {
            var resolved;
            stubs[prop] = (...args) => {
              resolved ||= resolveSymbol(prop);
              return resolved(...args);
            };
          }
```
https://github.com/emscripten-core/emscripten/blob/main/src/library_dylink.js?plain=1#L724-L727
This causes trouble when mixed with stack switching. Working around it will be a bit convoluted, but it should be possible with a little helper wasm module.

hoodmane:
Late-binding symbols get a JS stub import that resolves the symbol and then makes an onward call. This breaks JSPI. (It also canonicalizes NaNs by round tripping them through JS).

To fix this, we have to walk the import table and make wasm module stubs for each of the missing function-type symbols. These stubs will call a JS function to resolve the symbol but then insert it into a table and return control back to the wasm stub which has to make the onward call.

This currently relies on --experimental-wasm-type-reflection. To avoid that, we will need to parse the import table of the wasm binary manually. Also, without wasm-type-reflection, we have no way to handle the case where someone calls `loadWebAssemblyModule()` on a WebAssembly module instead of a Uint8Array.

This half way fixes #23395. I say half way because to be truly fixed the solution shouldn't rely on wasm type reflection. See also #23600.

### DIFF
--- a/test/core/test_dlfcn_jspi.c
+++ b/test/core/test_dlfcn_jspi.c
@@ -51,7 +51,9 @@ EM_JS(void, js_trampoline, (), {
 })
 
 int main() {
+  printf("Suspending test\n");
   helper(test_suspending_wrapper);
+  printf("Non suspending test\n");
   js_trampoline();
   printf("done\n");
   return 0;
--- a/test/core/test_dlfcn_jspi.out
+++ b/test/core/test_dlfcn_jspi.out
@@ -1,8 +1,10 @@
+Suspending test
 side_module_trampoline_a
 side_module_trampoline_b
 sleeping
 slept
 okay 77
+Non suspending test
 side_module_trampoline_a
 side_module_trampoline_b
 sync


sbc100:
> I say half way because to be truly fixed the solution shouldn't rely on wasm type reflection.

Why do you say this?  Is that simply because wasm type reflection is not standardized yet?   Presumably if it was standardized this would be good tool to solve this problem?

test/test_core.py
```diff
@@ -3769,13 +3769,23 @@ def test_dlfcn_jspi(self):
...
       + self.get_emcc_args()
     )
-    self.do_run_in_out_file_test("core/test_dlfcn_jspi.c", emcc_args=["side.so", "-sMAIN_MODULE=2"])
+    self.run_process(
+      [
+        EMCC,
+        "-o",
+        "side_b.so",
+        test_file("core/test_dlfcn_jspi_side_b.c"),
+        "-sSIDE_MODULE",
```
sbc100, line 3784: We use single quotes

src/lib/libdylink.js
```diff
@@ -877,16 +1057,28 @@ var LibraryDylink = {
...
 
       if (flags.loadAsync) {
-        if (binary instanceof WebAssembly.Module) {
-          var instance = new WebAssembly.Instance(binary, info);
-          return Promise.resolve(postInstantiation(binary, instance));
-        }
-        return WebAssembly.instantiate(binary, info).then(
-          (result) => postInstantiation(result.module, result.instance)
-        );
+        return (async function() {
```
sbc100, line 1060: Hmm.. nice trick to use `await` inside a function that is not `async`..

src/lib/libdylink.js
```diff
@@ -596,7 +596,179 @@ var LibraryDylink = {
...
+      }
+#endif
+      var sig = wasmSigToEmscripten(type);
+      var mod = getStubImportModule(sig);
+      const t = new WebAssembly.Table({element: 'funcref', initial: 1});
+      const r = getStubImportResolver(name, sig, t, resolveSymbol);
+      const inst = new WebAssembly.Instance(mod, {e: {t, r}});
+      stubs[name] = inst.exports.o;
+    }
+  },
```
sbc100, line 771: Can you put all of this new code inside `#if JSPI`?

src/lib/libdylink.js
```diff
@@ -697,6 +872,7 @@ var LibraryDylink = {
       // to add the indirection for, without inspecting what A's imports
       // are. To do that here, we use a JS proxy (another option would
       // be to inspect the binary directly).
+      const stubs = {};
```
sbc100, line 875: This can go down to line 920 I think.

src/lib/libdylink.js
```diff
@@ -596,7 +596,179 @@ var LibraryDylink = {
...
+    var lookup = {i32: 'i', i64: 'j', f32: 'f', f64: 'd', externref: 'e'};
+    var sig = 'v';
+    if (type.results.length) {
+      sig = lookup[type.results[0]];
+    }
+    for (var v of type.parameters) {
+      sig += lookup[v];
+    }
+    return sig;
+  },
```
sbc100, line 719: Can any of this new code be shared with the existing code for binary module creation in in libraryadd_function.js?

sbc100 (commented):
Oh wow, we really have to jump through a bunch of hoops to avoid JS thunks huh?

hoodmane:
Yes in particular my concern is that I expect jspi to become available without a feature flag before wasm-type-reflection so if we depend on it there will be engines that otherwise would be able to use our jspi but can't because they haven't yet shipped type reflection.

src/lib/libdylink.js
hoodmane: Perhaps it would make sense to factor out the wasm prelude (magic + version) as an export from `libaddfunction`. Other than that, the only code that's really shared is `generateFuncType` which we already extracted as a separate function.

sbc100:
> @sbc100 do you have any specific feedback on this PR?

It feel like a lot of complexity to be adding.   Perhaps we can find way to reduce it somehow?   I need to take a deeper look at some point.

hoodmane:
Unfortunately, I don't think we can do anything much simpler than this unless we change the shared object files.

src/lib/libdylink.js
```diff
@@ -722,16 +896,21 @@ var LibraryDylink = {
           // Return a stub function that will resolve the symbol
           // when first called.
           if (!(prop in stubs)) {
+#if JSPI
+            throw new Error("Missing stub for " + prop);
```
sbc100, line 900: Should this instead be an assertion?

```
#if JSPI
#if ASSERTIONS
  assert(prop in stubs, 'missing JSPI stub');
#endif
#else
  if (!(prop in stubs)) {
     ...
  }
#endif
```

src/lib/libdylink.js
```diff
@@ -871,16 +1050,28 @@ var LibraryDylink = {
...
+          }
+#if JSPI
+          var module = await WebAssembly.compile(binary);
+          addStubImports(module, stubs, resolveSymbol);
+          var instance = await WebAssembly.instantiate(module, info);
+#else
+          var {module, instance} = await WebAssembly.instantiate(binary, info);
+#endif
+          return postInstantiation(module, instance);
+        })();
```
sbc100, line 1069: Can we land this conversion to using a async function here as a separate PR?

src/lib/libdylink.js
```diff
@@ -582,7 +582,178 @@ var LibraryDylink = {
...
+    "$getStubImportResolver"
+  ],
+  $addStubImports: (mod, stubs, resolveSymbol) => {
+    // Assumes --experimental-wasm-type-reflection to get type field of WebAssembly.Module.imports().
+    // TODO: Make this work without it.
+    for (const {module, name, kind, type} of WebAssembly.Module.imports(mod)) {
+      if (module !== 'env' || kind !== 'function') {
+        continue;
+      }
+      if (name in wasmImports && !wasmImports[name].stub) {
```
sbc100, line 737: Can this be `if (resolveGlobalSymbol(name)) continue` instead?   Perhaps with a comment about how a stub won't be needed in that case?

src/lib/libdylink.js
```diff
@@ -582,7 +582,178 @@ var LibraryDylink = {
...
+        }
+        var wrapped = convertJsFunctionToWasm(res, sig);
+        table.set(0, wrapped);
+      }
+    };
+  },
+  $addStubImports__deps: [
+    "$getStubImportModule",
+    "$wasmSigToEmscripten",
+    "$getStubImportResolver"
```
sbc100, line 728: Using single quotes for strings.

src/lib/libdylink.js
```diff
@@ -582,7 +582,178 @@ var LibraryDylink = {
...
+  ],
+  $addStubImports: (mod, stubs, resolveSymbol) => {
+    // Assumes --experimental-wasm-type-reflection to get type field of WebAssembly.Module.imports().
+    // TODO: Make this work without it.
+    for (const {module, name, kind, type} of WebAssembly.Module.imports(mod)) {
+      if (module !== 'env' || kind !== 'function') {
```
sbc100, line 734: We also import stuff from the wasi namespace, so perhaps just drop the direct clause?

src/lib/libdylink.js
```diff
@@ -582,7 +582,178 @@ var LibraryDylink = {
...
+      if (name.startsWith("invoke_")) {
+        // JSPI + JS exceptions probably doesn't work but maybe nobody will
+        // attempt stack switch inside a try block.
+        stubs[name] = createInvokeFunction(name.split('_')[1]);
+        continue;
+      }
+#endif
+      var sig = wasmSigToEmscripten(type);
+      var mod = getStubImportModule(sig);
+      const t = new WebAssembly.Table({element: 'funcref', initial: 1});
```
sbc100, line 750: Since this is a JSPI-only feature can we skip the table have have `getStubImportResolver` return a funcref that we can use `call.ref` with instead of call_indirect?

src/lib/libdylink.js
hoodmane: Probably even better would be `if (resolveSymbol(name)) continue`. Then it's sort of self explanatory why we don't need a stub, but I will add a comment.

src/lib/libdylink.js
hoodmane: Well it makes the code slightly harder to maintain because wabt doesn't know how to handle it. But I can do that.

hoodmane:
@sbc100 I think this is ready for another round of review when you have a chance.

hoodmane:
@sbc100 would appreciate review on this one as well.

src/lib/libdylink.js
```diff
@@ -582,7 +582,169 @@ var LibraryDylink = {
...
+    };
+  },
+  $addStubImports__deps: [
+    '$getStubImportModule',
+    '$wasmSigToEmscripten',
+    '$getStubImportResolver',
+  ],
+  $addStubImports: (mod, stubs, resolveSymbol) => {
+    // Assumes --experimental-wasm-type-reflection to get type field of WebAssembly.Module.imports().
+    // TODO: Make this work without it.
```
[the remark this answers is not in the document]
hoodmane: If wasm-type-reflection is not available and no late bound symbols exist then we'll never attempt to use the `type` field and there won't be a problem. If a late-bound symbol exists then we'll crash inside of `wasmSigToEmscripten`.

src/lib/libdylink.js
```diff
@@ -582,7 +582,169 @@ var LibraryDylink = {
...
+      sig += lookup[v];
+    }
+    return sig;
+  },
+  $getStubImportResolver: (name, resolveSymbol) => {
+    return function r() {
+      var res = resolveSymbol(name);
+      if (res.orig) {
+        res = res.orig;
+      }
```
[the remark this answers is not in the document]
hoodmane: It looks sensible to me and makes the code a bit shorter.

src/lib/libdylink.js
```diff
@@ -582,7 +582,169 @@ var LibraryDylink = {
...
+  // is ii, the wat for the generated module looks like this:
+  //
+  // (module
+  //   (type $r (func ))
+  //   (type $ii (func (param i32) (result i32)))
+  //   (import "e" "t" (table 0 funcref))
+  //   (import "e" "r" (func $resolve))
+  //   (global $isResolved (mut i32) i32.const 0)
+  //
+  //   (func (export "o") (param $p1 i32) (result i32)
```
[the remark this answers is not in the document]
hoodmane: Well this is the argument to the function that we are going to resolve. We're going to pass it onward. I'm not sure what we call wrapper arguments like this -- in the `invoke_` functions, the arguments go unnamed.

src/lib/libdylink.js
```diff
@@ -582,7 +582,169 @@ var LibraryDylink = {
...
+  //     global.get $isResolved
+  //     i32.eqz
+  //     if
+  //       call $resolve
+  //       i32.const 1
+  //       global.set $isResolved
+  //     end
+  //     local.get $p1
+  //     i32.const 0
+  //     call_indirect (type $ii)
```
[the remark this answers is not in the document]
hoodmane: Yup, updated it.

src/lib/libdylink.js
```diff
@@ -874,18 +1048,29 @@ var LibraryDylink = {
...
           var instance;
           if (binary instanceof WebAssembly.Module) {
+#if JSPI
+            addStubImports(binary, stubs, resolveSymbol);
+#endif
             instance = new WebAssembly.Instance(binary, info);
           } else {
+#if JSPI
+            binary = await WebAssembly.compile(binary);
+            addStubImports(binary, stubs, resolveSymbol);
```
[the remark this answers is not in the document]
hoodmane: No, I think that would work.

hoodmane:
@sbc100 this is ready for another round of review

sbc100:
> @sbc100 this is ready for another round of review

When discussing this change with @kripken he suggested that we could use an alternative approach which used a little extra machinary in the JS wrapper to avoid the need the wasm code generation.   He can probably explain better than me.

I'm not sure if this alternative approach would avoid the wasm type reflection dependency?

kripken:
> Late-binding symbols get a JS stub import that resolves the symbol and then makes an onward call. This breaks JSPI.

I wonder if another approach here could be to "forward" the Promise through JS? That is, if we have a call stack with JS in the middle,
```
wasm (top)
JS (middle)
wasm (bottom)
```
The VM can't suspend the JS like it can the wasm. All it can do is suspend the wasm on top, which then returns a Promise to the JS in the middle. But that JS can await that Promise, leading to the wasm on the bottom being suspended. So we don't have a single wasm suspension, but two of them, with JS "glueing" them together.

For this to work, the middle JS would need to add the JSPI wrapper stuff for suspending, on both sides. This might not be that simple, I'm really not sure, but maybe it's another option to consider here?

hoodmane:
When I tried that approach previously I found the performance was unsatisfactory. But there is a decent chance that v8 has improved its stack switching machinery since then. I'll try implementing it that way and we can see if it sucks less than this.

sbc100:
> When I tried that approach previously I found the performance was unsatisfactory. But there is a decent chance that v8 has improved its stack switching machinery since then. I'll try implementing it that way and we can see if it sucks less than this.

I think even if the perf isn't great it might be better to have something that works without depending on type reflection (and therefore works without flags).

hoodmane:
So the alternative approach could look something like:
```
            if (!(prop in stubs)) {
              var resolved;
              var func = (...args) => {
                resolved ||= resolveSymbol(prop);
                try {
                  resolved = WebAssembly.promising(resolved);
                } catch(e) {}
                return resolved(...args);
              }
              stubs[prop] = new WebAssembly.Suspending(func);
            }
```

brendandahl:
The other idea we talked about was doing something similar to wasm-split in binaryen. It would work roughly like this with two modules `A` and `B`:

1) All the calls from `A` into `B` would be converted into call_indirect
2) In `A` insert a `Proxy` or stubs in the wasm table
3) On first call from `A`->`B`
  a) Load `B`
  b) Wrap all `B` exports with promising (and any imports as suspending if needed)
  c) `B` replaces the table in `A` with it's functions
  d) Call the function in `B` and resolve with the result

After the load of `B`, all the call_indirects would no longer go through the JS stubs and be calls from Wasm->Wasm.

hoodmane:
After thinking about this more, I still think this is the best approach:
1. Making an approach like #24161 work both when we can stack switch and we can't will make it as least as bad as this.
2. Fixing the problem with binaryen as Brendan Dahl suggests seems to require a few but not all dynamic libraries to be compiled separately for JSPI in order to work correctly even if they are not trying to stack switch. But also I think the proposal there won't work when we can't stack switch either.

The fix on this branch isn't _that_ much code and it is all distributed in one place in the main executable. If the main executable requires both dynamic linking and JSPI, the relevant code is injected. If it doesn't require both, the code isn't injected. The same shared libraries can be used whether building with stack switching off, or building with it on but not actually inside a promising export, or inside a promising export.

Issue #21791, opened 2024-04-20
When I use fflush(stdin), the output is wrong. Is it a bug?


laudedd:
Source code:
```cpp
#include <stdio.h>
#include <stdlib.h>

int main(int argc, char** argv) {
    int a,b;
    while( scanf("%d %d",&a,&b)!=EOF){
        fflush(stdin);
        printf("%d %d\n", a, b);
    }
 return 0;   
}
```
In input.txt:
```cpp
1 2
3 4
5 6
```
Run with "wasmtime a.wasm < input.txt" or "wasmer a.wasm < input.txt", the output is:
```cpp
1 2
```
Is it a bug?

sbc100:
Looks like probably an emscripten bug yes.  Something to do with the stdin handling most likely.

You can repro this without wasmtime or wasmer:

```
$ emcc test.c
node ./a.out.js < input.txt 
1 2
```

Surprisingly it also occurs with `-sNODERAWFS` which normally bypasses any bugs in the emscripten FS layers.

sbc100:
Actually I turns out that calling `fflush` on `stdin` is undefined behaviour.  Thanks @sunfishcode for pointing this out: https://en.cppreference.com/w/c/io/fflush:  "For input streams (and for update streams on which the last operation was input), the behavior is undefined.".

It turns out that above program happens to work on linux+musl because linux happen to allow lseek on stdin when its a file.  The above program happens to also work on emscripten now after  #23624, but perhaps we should revert that?

PR #23624, opened 2025-02-08
Handle failure of lseek in fflush

sbc100:
Unlike on linux (it seems) the emscripten lseek can fail for various reasons.  For example, we do not support lseek on TTY or pipe file descriptors.

Musl, it seems, just assumes that the seek succeeds and adjusts its internal buffers accordingly.   This PR update fflush such that it can handle the failure of lseek.

The alternative to this approach would be to add `seek` support to file our stdin implementation but that looks like a bigger change.   Since it seems reasonable for stdout and stdin to not be seekable updating musl to handle this case seems like the right thing to do for now.   If we find a other code in the wild that depends on stdin supporting lseek we can reconsider.

The following (14) test expectation files were updated by                
                                                                         
```                                                                      
other/codesize/test_codesize_cxx_ctors1.size: 129141 => 129147 [+6 bytes / +0.00%]
other/codesize/test_codesize_cxx_ctors2.size: 128553 => 128559 [+6 bytes / +0.00%]
other/codesize/test_codesize_cxx_except.size: 170809 => 170815 [+6 bytes / +0.00%]
other/codesize/test_codesize_cxx_except_wasm.size: 144517 => 144523 [+6 bytes / +0.00%]
other/codesize/test_codesize_cxx_except_wasm_legacy.size: 142092 => 142098 [+6 bytes / +0.00%]
other/codesize/test_codesize_cxx_lto.size: 122093 => 122099 [+6 bytes / +0.00%]
other/codesize/test_codesize_cxx_mangle.size: 232619 => 232625 [+6 bytes / +0.00%]
other/codesize/test_codesize_cxx_noexcept.size: 131704 => 131710 [+6 bytes / +0.00%]
other/codesize/test_codesize_cxx_wasmfs.size: 169111 => 169117 [+6 bytes / +0.00%]
other/codesize/test_codesize_files_wasmfs.size: 50049 => 50055 [+6 bytes / +0.01%]
other/codesize/test_codesize_hello_O0.size: 15101 => 15105 [+4 bytes / +0.03%]
other/codesize/test_codesize_minimal_O0.size: 1156 => 1160 [+4 bytes / +0.35%]
other/test_unoptimized_code_size.wasm.size: 15101 => 15105 [+4 bytes / +0.03%]
other/test_unoptimized_code_size_strict.wasm.size: 15101 => 15105 [+4 bytes / +0.03%]
                                                                         
Average change: +0.03% (+0.00% - +0.35%)                                 
```      

Fixes: #21791


system/lib/libc/musl/src/stdio/fflush.c
```diff
@@ -34,11 +34,22 @@ int fflush(FILE *f)
 	}
 
 	/* If reading, sync position, per POSIX */
+#if __EMSCRIPTEN__
+	/* Handle failues of lseek, which can happen in emscripten, e.g. for stdin etc */
+	if (f->rpos != f->rend) {
+		if (f->seek(f, f->rpos-f->rend, SEEK_CUR) == 0) {
```
hoodmane, line 40: If fseek fails shouldn't flush return -1 to indicate failure? Or is this seek optional for some reason?

hoodmane:
What happens in musl when seeking stdout? I would think the right thing would be returning ENOSYS or ENOTSUP or something. Does it return success and do nothing?

system/lib/libc/musl/src/stdio/fflush.c
hoodmane: I guess if we're flushing stdout we don't want it to fail because it's not seekable

system/lib/libc/musl/src/stdio/fflush.c
```diff
@@ -34,11 +34,22 @@ int fflush(FILE *f)
 	}
 
 	/* If reading, sync position, per POSIX */
+#if __EMSCRIPTEN__
+	/* Handle failues of lseek, which can happen in emscripten, e.g. for stdin etc */
+	if (f->rpos != f->rend) {
```
hoodmane, line 39: In the reproduction, what are rpos and rend? Are we sure the problem isn't that they should be equal always on non seekable streams? and the bug is that we are setting them to different things?

sbc100:
> What happens in musl when seeking stdout? I would think the right thing would be returning ENOSYS or ENOTSUP or something. Does it return success and do nothing?

The code here is only effect files that are being read-from not written-to.   In linux it turns out that you can see `stdin` just fine. I was actually surprised to discover this.     I ran this test of alpine linux (which uses musl) and the stdin file handle is successfully "seeked" backwards by fflush.

system/lib/libc/musl/src/stdio/fflush.c
sbc100: This seeks I believe is kind of optional.  What is trying to do is that read-ahead buffer should be discarded and the FD should be seek'd back to the current position.   On linux you can seek backward in stdin, but not in emscripten.

We should try to fix emscripten's stdin to make it seekable, but that seem harder than making this small change to fflush.

system/lib/libc/musl/src/stdio/fflush.c
sbc100: The stdio code in musl will read `BUFSIZE` bytes from stdin at once.  See: https://github.com/emscripten-core/emscripten/blob/242af002f8dee0366e86002fec187d447dcd26e5/system/lib/libc/musl/src/stdio/stdin.c#L8 

This means that for the test code it only make a single read syscall and then future call to fread just read from the `rpos` in the buffer.

Then fflush is called it ties to discard the buffer completely and reset the stdin fd back to the current read position.

I supposed you could observe difference if yous ran `lseek(fileno(stdout), 0, SEEK_CUR)` after the fflush.   I suppose the idea is that after `fflush` the `lseek(CUR)` position should match the `ftell` position, but normally you would expect `lseek(CUR)` to be ahead due the read-read that the `FILE*` operations are doing.

sbc100:
See #23627 for more the `universal_newlines` vs `text` thing.  I'll probably let that change land first.

hoodmane:
> the stdin file handle is successfully "seeked" backwards by fflush.

Should we consider copying that behavior for the stdin device instead of making this change? Since otherwise there is an observable difference in the behavior when seeking `stdin`.

sbc100:
> > the stdin file handle is successfully "seeked" backwards by fflush.
> 
> Should we consider copying that behavior for the stdin device instead of making this change? Since otherwise there is an observable difference in the behavior when seeking `stdin`.

Can you explain what you mean exactly?

Are you suggesting that we update emscripten to allow stdin to support seeking?   I think that would be much bigger change and without much benefit (as I mentioned in https://github.com/emscripten-core/emscripten/pull/23624#discussion_r1947833097).   I also think its fairly reasonable that stdin and stdout don't support seeking, no?

Here is the code for stdin handling BTW: https://github.com/emscripten-core/emscripten/blob/242af002f8dee0366e86002fec187d447dcd26e5/src/lib/libfs_shared.js#L107-L177.  I did take a look to see if we could enable seeking without too much trouble, but it looked like it would be relatively complex change.. without much benefit that I can see.

hoodmane:
> I also think its fairly reasonable that stdin and stdout don't support seeking, no?

I agree that they shouldn't. But I think if linux does something unreasonable we should try to do the same unreasonable thing whenever possible.

hoodmane:
But yeah if it's more effort than it's worth that is fair.

hoodmane:
I think this would be enough to make stdin seekable in the JS fs:
```patch
--- a/src/lib/libfs.js
+++ b/src/lib/libfs.js
@@ -1493,7 +1493,7 @@ FS.staticInit();
       // have been overwritten we create a unique device for
       // them instead.
       if (input) {
-        FS.createDevice('/dev', 'stdin', input);
+        FS.createDevice('/dev', 'stdin', input, null, {seekable: true});
       } else {
         FS.symlink('/dev/tty', '/dev/stdin');
       }
@@ -1659,7 +1659,7 @@ FS.staticInit();
         FS.chmod(node, mode);
       }
     },
-    createDevice(parent, name, input, output) {
+    createDevice(parent, name, input, output, opts) {
       var path = PATH.join2(typeof parent == 'string' ? parent : FS.getPath(parent), name);
       var mode = FS_getMode(!!input, !!output);
       FS.createDevice.major ??= 64;
@@ -1668,7 +1668,7 @@ FS.staticInit();
       // the old behavior.
       FS.registerDevice(dev, {
         open(stream) {
-          stream.seekable = false;
+          stream.seekable = opts.seekable ?? false;
         },
         close(stream) {
           // flush any pending line data
```

hoodmane:
But maybe that will cause a hard crash when you try to seek it, I didn't try it =)

sbc100:
> > I also think its fairly reasonable that stdin and stdout don't support seeking, no?
> 
> I agree that they shouldn't. But I think if linux does something unreasonable we should try to do the same unreasonable thing whenever possible.

I don't think emscripten should go out of its way to emulate ever linux corner case like this.  At least not until we have evidence that real world programs depend on them.   Our libc for example, even though its based on musl, lacks most linux-specific stuff.  We can make exceptions if we have a lot of real work code the needs it but in general we don't support linux-specific stuff.

sbc100:
> I think this would be enough to make stdin seekable in the JS fs:

Its more complex than that I fear.  See https://github.com/emscripten-core/emscripten/blob/242af002f8dee0366e86002fec187d447dcd26e5/src/lib/libtty.js#L62-L85.   We would need to implement some kind of extra buffering layer, or change the way `get_char` works here.

hoodmane:
Okay I think I finally understand. You're saying that `stdin` is saving recent input so that we can seek back a little bit and replay the recent bytes? That indeed sounds too complicated to bother with.

sbc100:
> Okay I think I finally understand. You're saying that `stdin` is saving recent input so that we can seek back a little bit and replay the recent bytes? That indeed sounds too complicated to bother with.

The linux stdin fd certainly seems to be doing that yes.  At least in my tests on alpine linux.  I'm not sure why.

hoodmane (approved):
Thanks for the explanation @sbc100. I think this makes sense to me now. It might be helpful if you could add a summary of your answers to my questions to the comment.

sbc100:
I updated the comment.  Thanks for the review @hoodmane, and thanks for all your recent contributions.

sbc100:
Actually I turns out that calling fflush on stdin is undefined behaviour: https://en.cppreference.com/w/c/io/fflush: "For input streams (and for update streams on which the last operation was input), the behavior is undefined.".

Should we revert this?

PR #23627, opened 2025-02-08
Replace python subprocess's `universal_newlines` with `text`. NFC

sbc100:
In python 3.7 the `universal_newlines` parameter was deprecated in favor of the more readable `text` parameter.

We updated our minimum supported python version to 3.8 in #23417.

### DIFF
--- a/tools/maint/check_for_closed_issues.py
+++ b/tools/maint/check_for_closed_issues.py
@@ -16,7 +16,7 @@ root_dir = os.path.dirname(os.path.dirname(script_dir))
 
 
 def run(*args, **kwargs):
-  kwargs['universal_newlines'] = True
+  kwargs['text'] = True
   return subprocess.check_output(*args, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


emrun.py
```diff
@@ -738,13 +738,13 @@ def do_POST(self):
     self.wfile.write(b'OK')
 
 
-# Returns stdout by running command with universal_newlines=True
-def check_output(cmd, universal_newlines=True, *args, **kwargs):
+# Returns stdout by running command with text=True
+def check_output(cmd, text=True, *args, **kwargs):
   if hasattr(subprocess, "run"):
```
hoodmane, line 743: You can delete this check, it's only needed to support Python < 3.5.

sbc100: Thanks for spotting that!

sbc100: Oh wait, sorry I forgot that emrun.py is the exception and it need to run on older python versions :(

sbc100: Reverted emrun.py part of the change.

hoodmane (approved):

PR #23629, opened 2025-02-08
Late module assignment

sbc100:


sbc100:
CC @sgbeal

sgbeal:
> CC @sgbeal

Thank you for the heads up! i'll keep an eye out for any fallout on our end.

> When assertions are enabled, we now detect this too-late-assignment and abort with a useful error.

What happens when asserts are _not_ enabled? Does that make such usage "legal but possibly ill-advised" or "undefined behavior"?

sbc100:
> > CC @sgbeal
> 
> Thank you for the heads up! i'll keep an eye out for any fallout on our end.
> 
> > When assertions are enabled, we now detect this too-late-assignment and abort with a useful error.
> 
> What happens when asserts are _not_ enabled? Does that make such usage "legal but possibly ill-advised" or "undefined behavior"?

The same thing that happens today: nothing at all.   The assignment will go undetected and un-used.

sbc100:
> > > CC @sgbeal
> > 
> > 
> > Thank you for the heads up! i'll keep an eye out for any fallout on our end.
> > > When assertions are enabled, we now detect this too-late-assignment and abort with a useful error.
> > 
> > 
> > What happens when asserts are _not_ enabled? Does that make such usage "legal but possibly ill-advised" or "undefined behavior"?
> 
> The same thing that happens today: nothing at all. The assignment will go undetected and un-used.

When we try to detect such API misused, or change the behaviour of emscripten like this, we normally focus on reported these errors only when assertions are enabled, to avoid increasing code size in release builds.

kripken (approved):
Good idea!

Issue #23608, opened 2025-02-06
Built-in SDL3 port


viglietta:
It would be nice to have support for SDL3, the same way as it was included for SDL2.

sbc100:
Sounds like a good idea.  Would you have time to work on PR?

kripken:
Note that SDL3 should already work with Emscripten, see details here:

https://bsky.app/profile/emscripten.org/post/3lchbwh4e5k2j

viglietta:
Sorry, I am unable to work on PR.

Also, I am aware that SDL3 works with Emscripten. However, I don't think it's as well integrated as SDL1 or SDL2.
For instance, is preloading files possible with SDL3?
https://emscripten.org/docs/porting/files/packaging_files.html#preloading-files

sbc100:
WIP is here: https://github.com/emscripten-core/emscripten/pull/23630

PR #23630, opened 2025-02-08
Add initial port of SDL3

sbc100:
I just got a couple of test working, but everything seems good so far.

Fixes: #23608


sbc100:
CC @slouken @icculus

tools/ports/sdl3.py
```diff
@@ -0,0 +1,130 @@
...
+
+  return [shared.cache.get_lib(get_lib_name(settings), create, what='port')]
+
+
+def clear(ports, settings, shared):
+  shared.cache.erase_lib(get_lib_name(settings))
+
+
+def process_args(ports):
+  return ['-isystem', ports.get_include_dir('SDL3')]
```
sbc100, line 126: @icculus should we inject this include path like this?  Or are all uses expected to include via "SDL3/xxxx.h"?

slouken: For SDL3, all users are supposed to include the SDL3 prefix:
```c
#include <SDL3/SDL.h>
```

sbc100: When you say "not supposed to" does that mean that `include/SDL3` is not on the default include path and `#include <SDL.h>` will not work?  e.g. in the cmake config for SDL3?   Or is it just a recommendation not to rely on that include path?   I feel like emscripten should do that same thing that the cmake or pkg-config does.. whatever that is.

slouken: That's correct, `#include <SDL.h>` will not work.
https://github.com/libsdl-org/SDL/blob/9bd6d36471784e31109eab554d39e2aaf93cd56f/docs/README-migration.md?plain=1#L46

sbc100: Nice!  Thats good to know.  I will remove this.

tools/ports/sdl3.py
```diff
@@ -0,0 +1,126 @@
...
+      'joystick/emscripten/*.c',
+      'joystick/virtual/*.c',
+      'power/emscripten/*.c',
+      'misc/emscripten/*.c',
+      'audio/emscripten/*.c',
+      'video/emscripten/*.c',
+      'video/offscreen/*.c',
+      'audio/disk/*.c',
+      'loadso/dlopen/*.c',
+      # Dummy backends, do we need to support these?
```
sbc100, line 93: @slouken, sorry for another dumb question, but do you think its worth including the dummy backends?    I guess it will increase code size, but maybe allow SDL3 to be used in more cases/places?

slouken: I'm not sure how useful they are for emscripten, maybe if someone needs to disable audio or something like that? @icculus, thoughts?

test/browser/test_sdl2_canvas_write.c
```diff
@@ -5,73 +5,75 @@
 
 #include <assert.h>
 #include <SDL.h>
```
kripken, line 7: is this test change unrelated?

sbc100: Yeah, these are the sdl2 tests that I cloned in order to create the new sdl3 tests.  I cleaned up the formatting in these two tests before cloning them

sbc100: View with ignore-whitespace to see that they are just cleanups.

test/browser/test_sdl3_canvas_write.c
```diff
@@ -0,0 +1,85 @@
+// Copyright 2015 The Emscripten Authors.  All rights reserved.
```
kripken, line 1:
```suggestion
// Copyright 2025 The Emscripten Authors.  All rights reserved.
```

test/test_other.py
```diff
@@ -2476,6 +2476,13 @@ def test_sdl2_linkable(self):
     self.emcc(test_file('browser/test_sdl2_misc.c'), ['-sLINKABLE', '-sUSE_SDL=2'], output_filename='a.out.js')
     self.emcc(test_file('browser/test_sdl2_misc.c'), ['-sLINKABLE', '--use-port=sdl2'], output_filename='a.out.js')
 
+  def test_sdl3_linkable(self):
+    # Ensure that SDL2 can be built with LINKABLE.  This implies there are no undefined
+    # symbols in the library (because LINKABLE includes the entire library).
+    self.emcc_args.append('-Wno-experimental')
+    self.emcc(test_file('browser/test_sdl3_misc.c'), ['-sLINKABLE', '-sUSE_SDL=3'], output_filename='a.out.js')
+    self.emcc(test_file('browser/test_sdl3_misc.c'), ['-sLINKABLE', '--use-port=sdl3'], output_filename='a.out.js')
```
kripken, line 2484: Why test both ways of including SDL3 here? isn't that just a linker UI feature (not related to linkable/undefined symbols)?

tools/ports/sdl3.py
```diff
@@ -0,0 +1,126 @@
+# Copyright 2014 The Emscripten Authors.  All rights reserved.
```
kripken, line 1:
```suggestion
# Copyright 2025 The Emscripten Authors.  All rights reserved.
```

tools/ports/sdl3.py
```diff
@@ -0,0 +1,126 @@
...
+
+  def create(final):
+    diagnostics.warning('experimental', 'sdl3 port is still experimental')
+    root_dir = ports.get_dir('sdl3', SUBDIR)
+
+    # copy includes to a location so they can be used as 'SDL3/'
+    source_include_path = os.path.join(root_dir, 'include', 'SDL3')
+    ports.install_headers(source_include_path, target='SDL3')
+
+    # In addtion copy our pregenerated SDL_build_config.h file.
```
kripken, line 38:
```suggestion
    # In addition, copy our pregenerated SDL_build_config.h file.
```

kripken:
lgtm otherwise.

test/test_other.py
sbc100: Yes, its just two ways to link the same library.  Testing that both work.  This a direct copy of the SDL2 test right above.

tools/ports/sdl3.py
sbc100: Wow, this makes me feel old... 11 years since that sdl2 port was done

test/test_other.py
```diff
@@ -2476,6 +2476,13 @@ def test_sdl2_linkable(self):
     self.emcc(test_file('browser/test_sdl2_misc.c'), ['-sLINKABLE', '-sUSE_SDL=2'], output_filename='a.out.js')
     self.emcc(test_file('browser/test_sdl2_misc.c'), ['-sLINKABLE', '--use-port=sdl2'], output_filename='a.out.js')
 
+  def test_sdl3_linkable(self):
+    # Ensure that SDL2 can be built with LINKABLE.  This implies there are no undefined
```
slouken, line 2480:
```suggestion
    # Ensure that SDL3 can be built with LINKABLE.  This implies there are no undefined
```

ypujante:
@sbc100  I guess I am late, but I wonder why did you choose to do `-sUSE_SDL=3` (which I thought was a syntax we were trying to deprecate), instead of the new `--use-port=sdl3`?

ypujante:
@sbc100 I was also in the process of doing the sdl3 port... so here is my version

[sdl3.py.zip](https://github.com/user-attachments/files/18770110/sdl3.py.zip)

The way I did is "scientific":

I ran the emcmake / cmake build.

I took the output of the run which generated the exact list of  files that are compiled (`file_list`). And the exact config file generated as well `build_config_h`

Hope this helps

ypujante:
@sbc100 I guess you can do whatever you want but I personally don't like the addition of the `sdl3/SDL_build_config.h` in the ports folder. I feel like if I had proposed such a change it would have been rejected. 

So far all ports have been self-contained and I feel it is better that way.

I also demonstrated in the sdl3.py file I just posted in my previous message that you can simply include the .h inside the port file (and BTW other ports do it as well)

sbc100:
> @sbc100 I guess you can do whatever you want but I personally don't like the addition of the `sdl3/SDL_build_config.h` in the ports folder. I feel like if I had proposed such a change it would have been rejected.
> 
> So far all ports have been self-contained and I feel it is better that way.
> 
> I also demonstrated in the sdl3.py file I just posted in my previous message that you can simply include the .h inside the port file (and BTW other ports do it as well)

Indeed, it could be argued that self-contained headers are better.  However I've always found a it rather odd to have large amounts of C/C++ header code inside of python strings.   I've always been hoping to find a way to avoid that, and this is my first attempt.   I was thinking of following up with the extraction of such headers into their own files for other ports.    I get that for external ports its kind of nice to represent them as a single file, so there are arguments both ways.

sbc100:
I created a PR to inline the header for now: https://github.com/emscripten-core/emscripten/pull/23660

I may try to un-inline all of these headers in the future, but I agree that being consistent is good for now.

sbc100:
> I created a PR to inline the header for now: #23660
> 
> I may try to un-inline all of these headers in the future, but I agree that being consistent is good for now.

I now have a PR out to remove all the embedded files from the ports #24308

PR #23631, opened 2025-02-09
Update webgpu-externs.js

fs-eire:
Add type declaration for `GPUAdapter.info` and `GPUDevice.adapterInfo`

Dawn is maintaining an forked up-to-date version of library_webgpu.js, but the webgpu-extern.js is not forked.


kainino0x (approved):
LGTM, thank you!

Issue #17891, opened 2022-09-20
Should addFunction grow the table less frequently?


kg:
Right now addFunction grows the wasm table by 1 every time it adds a new function. This seems like it could be a performance issue, since growing tables could invalidate jitcode by changing the range of valid indices, etc. Would it make sense to grow it in chunks, say 256 items at a time? It doesn't seem like the cost of using an extra say (let's be conservative here) `(32 * 255)` bytes of memory would be a problem. Is it a known thing that this isn't an issue?

I've been jitting wasm functions and importing them via addFunction and seeing mysterious perf issues, so I've been trying to track down the causes and this seemed like a possible factor.

sbc100:
Sounds reasonable.  We could even use some kind of quadratic increment.   Seems like relatively small change to `getEmptyTableSlot`.

Thats very interesting that you are jitting wasm functions..   out of interest are you creating an entirely new module each time or using the new WebAssembly.Function from https://github.com/WebAssembly/js-types/blob/main/proposals/js-types/Overview.md?

kg:
I'm jitting modules right now similar to what addFunction does, and may try to cluster them together into larger modules since it seems like having too many small modules kicking around probably would create problems in various parts of browser runtimes. I still have to investigate more where all the performance pits are. Is WebAssembly.Function actually available in a near timeframe in browsers? I hadn't realized it was an option, should definitely investigate it.

FWIW I started growing the table manually by 256 and filling it myself, and didn't notice a big performance improvement so I don't think the problem was grow. (It looks like the set operation may be what actually has the invalidation overhead, from looking at the v8 source code - but it has a few different paths it can take depending on the value you're inserting.)

Growing in larger steps may still be worth it for workloads where adding happens more frequently, though.

martenrichter:
I looked in the chromium profiler for the [jupyterlite xeus kernel](https://github.com/jupyterlite/xeus). 
At longer periods, it is most often inside the emscripten grow function (between extracts). 
According to the data, it is most often inside `getEmptyTableSlot` and `addFunction`, growing the table by one.
Is there a way one may override this function? (For testing?)

hoodmane:
But another thing that works is if you `./emsdk install 4.0.3` then you can go into `upstream/emscripten/src/lib/libaddfunction.js` and change it there. There is no compilation step for the js runtime libraries, if you change it in place the next time you link something your change will get picked up.

martenrichter:
[...]

PR #23633, opened 2025-02-09
Reduce number of calls to wasmTable.grow

martenrichter:
Before, getEmptyTableSlot increased the wasmTable by 1, increasing the overhead if many calls of addFunction exceeded the table length. Now, an exponential scaling strategy is implemented.

https://github.com/emscripten-core/emscripten/issues/17891


martenrichter:
Note: I have tested on a compiled Emscripten module, so I am not sure if the formatting is correct. I hope the test will show it.

src/lib/libaddfunction.js
```diff
@@ -151,15 +151,17 @@ addToLibrary({
     }
     // Grow the table
     try {
+      var growBy = {{{ from64Expr('wasmTable.length') }}};
+      freeTableIndexes.push(...Array.from({ length: growBy }, (_, i) => {{{ from64Expr('wasmTable.length') }}} + i));
```
hoodmane, line 155: It would use less memory to record the last used entry separately. Also I'm not sure we want to grow the table exponentially, @sbc100 was suggesting linear or quadratic.

martenrichter: Okay, I have changed the behavior. Unfortunately, the last commit tests failed, and I did not get a clue what was wrong. (It must be something with the templating.) Maybe the new one is better.
Regarding linear or quadratic scaling, is the initial size of the table somewhere available? We need it to achieve linear or quadratic scaling. I also think it is important that the growth is in a similar order of magnitude as the current size.

hoodmane:
I was imagining something roughly like:
```
  $freeTableIndexes: [],
  $usedTableLength: 0, // TODO: initialize this correctly

  // Weak map of functions in the table to their indexes, created on first use.
  $functionsInTableMap: undefined,

  $getEmptyTableSlot__deps: ['$freeTableIndexes', '$wasmTable', '$usedTableLength'],
  $getEmptyTableSlot: () => {
    // Reuse a free index if there is one, otherwise grow.
    if (freeTableIndexes.length) {
      return freeTableIndexes.pop();
    }
    if (usedTableLength < wasmTable.length) {
      return usedTableLength++;
    } 
    // Grow the table
    try {
      /** @suppress {checkTypes} */
      wasmTable.grow({{{ toIndexType('wasmTable.length') }}});
    } catch (err) {
      if (!(err instanceof RangeError)) {
        throw err;
      }
      throw 'Unable to grow wasm table. Set ALLOW_TABLE_GROWTH.';
    }
    return usedTableLength++;
  },
```

martenrichter:
Ok, I try if this works. But I am not completely sure, what needs to be changed in libdylink.js .

hoodmane:
I think it should be something like:
```
        var tableBase = metadata.tableSize ? usedTableLength : 0;
        if (handle) {
          {{{ makeSetValue('handle', C_STRUCTS.dso.mem_allocated, '1', 'i8') }}};
          {{{ makeSetValue('handle', C_STRUCTS.dso.mem_addr, 'memoryBase', '*') }}};
          {{{ makeSetValue('handle', C_STRUCTS.dso.mem_size, 'metadata.memorySize', 'i32') }}};
          {{{ makeSetValue('handle', C_STRUCTS.dso.table_addr, 'tableBase', '*') }}};
          {{{ makeSetValue('handle', C_STRUCTS.dso.table_size, 'metadata.tableSize', 'i32') }}};
        }
      } else {
        memoryBase = {{{ makeGetValue('handle', C_STRUCTS.dso.mem_addr, '*') }}};
        tableBase = {{{ makeGetValue('handle', C_STRUCTS.dso.table_addr, '*') }}};
      }

      var tableGrowthNeeded = tableBase + metadata.tableSize - {{{ from64Expr('wasmTable.length') }}};
      if (tableGrowthNeeded > 0) {
          wasmTable.grow({{{ toIndexType('tableGrowthNeeded') }}});
      }
      usedTableLength = tableBase + metadata.tableSize
```

hoodmane:
And probably `$dumpTable` in `libdylink.js` should use `usedTableLength` instead of `wasmTable.length`.

martenrichter:
I agree, I came up with similar changes. But I am not familiar with the code....

martenrichter:
But if you use dylink, would it not generally interfere, if you have added functions in the meantime...? (Beside this change)

hoodmane:
No, dynamic linking currently works fine with `addFunction`. The library exports get placed in a block at the end of the wasmTable, ignoring any freelist entries.

hoodmane:
I think `usedTableLength` should also be initialized in two places: line 911 of `preamble.js` if `!RELOCATABLE`:
```js
#if '$wasmTable' in addedLibraryItems && !RELOCATABLE
    wasmTable = wasmExports['__indirect_function_table'];
#if ALLOW_TABLE_GROWTH
    usedTableLength = wasmTable.length;
#endif
    {{{ receivedSymbol('wasmTable') }}}
```
and line 2276 of libcore.js:
```js
#if RELOCATABLE
  // In RELOCATABLE mode we create the table in JS.
  $wasmTable: `=new WebAssembly.Table({
  'initial': {{{ toIndexType(INITIAL_TABLE) }}},
#if !ALLOW_TABLE_GROWTH
  'maximum': {{{ toIndexType(INITIAL_TABLE) }}},
#endif
#if MEMORY64 == 1
  'address': 'i64',
   // TODO(sbc): remove this alias for 'address' once both firefox and
   // chrome roll out the spec change.
   // See https://github.com/WebAssembly/memory64/pull/92
  'index': 'i64',
#endif
  'element': 'anyfunc'
});
`,
#if ALLOW_TABLE_GROWTH
   $wasmTable_postset: `usedTableLength = wasmTable.length`,
#endif
#else
```

But I'm not really sure where to stick the `$usedTableLength` definition to make that work...

martenrichter:
Ok, I was just wondering, as it contains the if clause with one branch being first load. I was wondering, if a wasm was loaded a second time with an increased number of functions. As it recycles the position in the table, it will overwrite the function pointers. If it would be running on the os and not in the sandbox, we would have a vulnerability.

hoodmane:
Are you saying that you're concerned that `addFunction` isn't thread safe? It's probably not.

martenrichter:
No, I am concerned with the following scenario
1.) You load a lib first with say x functions.
2.) Then you use addfunction with some  functions
3.) You load the lib a second time but now it has x +1 functions, and the x+1 will overwrite the function added in 2.
But I am not sure if this would work, I was just wondering because of  `firstLoad`  .

hoodmane:
dynamic linking + pthreads is experimental:
https://emscripten.org/docs/compiling/Dynamic-Linking.html#pthreads-support

I think the safe way to use it is to make sure all library loading and addFunction calls happen before you start any threads.

hoodmane:
`firstLoad` is about threads. It's `true` when one thread has already loaded the library and it is being loaded into a second thread. If you load two different versions of the same library that will probably cause crashes.

martenrichter:
Ok, I understand now. But should it check, that it is not having a different number of table entries and throw?

kg:
With fast table growing it's quite possible to hit the limit for how big a function table can be if your function table already started big. I've heard of applications with tables containing 100k entries or more.

Given that, you probably want to handle the failure case and attempt one retry where you grow by a smaller amount. Otherwise if you have an application with a table that has 400k entries and the size limit is 500k, it'd never be able to grow, even though in practice there's plenty of room to grow.

martenrichter:
Ok, good point, a factor of 10 down? Or better 100 ?

kg:
Your measurements showed that 'grow by 100' and 'grow by 1000' were both big improvements over 'grow by 1'. I'd suggest just falling back to a flat 100 or 1000. 100 is more likely to work, but if an application is so close to the limit that 1000 won't work, it's probably close to broken anyway, so I could understand going with 1000 for better performance. I'd personally pick 100.

src/lib/libaddfunction.js
```diff
@@ -149,17 +149,21 @@ addToLibrary({
     if (freeTableIndexes.length) {
       return freeTableIndexes.pop();
     }
+    // record last index
+    var lastUnused =  {{{ from64Expr('wasmTable.length') }}};
     // Grow the table
     try {
+      var growBy = lastUnused;
+      freeTableIndexes.push(...Array.from({ length: growBy - 1 }, (_, i) => lastUnused + i + 1));
```
kg, line 157: This is probably going to be really expensive for large tables. You might want to track free ranges instead of free indices.

martenrichter: This already gone, I have changed it to the suggestions by hoodmane. I will push soon, but I have to fix a failing test.

sbc100:
> No, I am concerned with the following scenario 1.) You load a lib first with say x functions. 2.) Then you use addfunction with some functions 3.) You load the lib a second time but now it has x +1 functions, and the x+1 will overwrite the function added in 2. But I am not sure if this would work, I was just wondering because of `firstLoad` .

You can't load the same dynamic library twice.   If you were to give them different names and load two version of the same library each version will get its own separate table and memory allocation.

martenrichter:
Ok, I have added the changes and the test failing before now works. But I did not run the full suite, I am waiting for CI.

martenrichter:
> Your measurements showed that 'grow by 100' and 'grow by 1000' were both big improvements over 'grow by 1'. I'd suggest just falling back to a flat 100 or 1000. 100 is more likely to work, but if an application is so close to the limit that 1000 won't work, it's probably close to broken anyway, so I could understand going with 1000 for better performance. I'd personally pick 100.

I have now implemented growing by 1000 and then falling back to 10. But other variants are also possible.

sbc100:
> > Your measurements showed that 'grow by 100' and 'grow by 1000' were both big improvements over 'grow by 1'. I'd suggest just falling back to a flat 100 or 1000. 100 is more likely to work, but if an application is so close to the limit that 1000 won't work, it's probably close to broken anyway, so I could understand going with 1000 for better performance. I'd personally pick 100.
> 
> I have now implemented growing by 1000 and then falling back to 10. But other variants are also possible.

It seems to me that unless you fall all the way back to "grow by 1" you might not ever be able to use the final available slots.

I was going to suggest that instead of implementing falloff we could just query the max length .. but I don't think WebAssembly.Memory exposes that.

martenrichter:
> > > Your measurements showed that 'grow by 100' and 'grow by 1000' were both big improvements over 'grow by 1'. I'd suggest just falling back to a flat 100 or 1000. 100 is more likely to work, but if an application is so close to the limit that 1000 won't work, it's probably close to broken anyway, so I could understand going with 1000 for better performance. I'd personally pick 100.
> > 
> > 
> > I have now implemented growing by 1000 and then falling back to 10. But other variants are also possible.
> 
> It seems to me that unless you fall all the way back to "grow by 1" you might not ever be able to use the final available slots.
> 
> I was going to suggest that instead of implementing falloff we could just query the max length .. but I don't think WebAssembly.Memory exposes that.

Ok, then we go with 1 . And yes the spec does not have this function. I could look in the source code of Chromium if this helps, but if it is not in the specs it can change anytime.

src/preamble.js
```diff
@@ -910,6 +910,7 @@ function getWasmImports() {
 
 #if '$wasmTable' in addedLibraryItems && !RELOCATABLE
     wasmTable = wasmExports['__indirect_function_table'];
+    usedTableLength = wasmTable.length;
```
sbc100, line 913: Ideally we could avoid adding any code here, since it will effect all programs.

You can use the library system to inject some startup code if (and only if) usedTableLength is used.

```
$usedTableLength__postset: () => addAtInit(`usedTableLength = wasmTable.length`);     
```

This way only programs that reference `usedTableLength` will pay the code.

src/lib/libdylink.js
```diff
@@ -660,6 +661,7 @@ var LibraryDylink = {
 #endif
         wasmTable.grow({{{ toIndexType('tableGrowthNeeded') }}});
```
sbc100, line 662: It might be worth converting this code to call some kind of `getEmptyTableSlots(count)` function that could return an empty contiguous range and possibly avoid growing the table on every DLL load.

src/lib/libdylink.js
```diff
@@ -668,6 +669,7 @@ var LibraryDylink = {
 #endif
         wasmTable.grow({{{ toIndexType('metadata.tableSize') }}});
       }
+      usedTableLength = tableBase + metadata.tableSize;
```
sbc100, line 672: I think this line should be moved up inside the above `if (metadata.tableSize)` check, no?

martenrichter: I think this makes sense.

test/interop/test_add_function.cpp
```diff
@@ -27,7 +27,7 @@ int main(int argc, char **argv) {
   EM_ASM({
     // Get an export that isn't in the table (we never took its address in C).
     var baz = wasmExports["baz"];
-    var tableSizeBefore = wasmTable.length;
+    var tableSizeBefore = usedTableLength;
```
sbc100, line 30: Should use update the `We can reuse indexes` part of the test below too?

src/lib/libaddfunction.js
```diff
@@ -139,27 +139,38 @@ addToLibrary({
...
       return freeTableIndexes.pop();
     }
+    if (usedTableLength < {{{ from64Expr('wasmTable.length') }}}) {
+      return usedTableLength++;
+    }
     // Grow the table
     try {
       /** @suppress {checkTypes} */
-      wasmTable.grow({{{ toIndexType('1') }}});
+      wasmTable.grow({{{ toIndexType('1000') }}});
```
sbc100, line 158: I don't love having this hardcoded constant here here, but I also don't want to introduce something like `MEMORY_GROWTH_GEOMETRIC_STEP` that we have for memory. 

Should we perhaps make it a power of 2?  How about 256?

test/interop/test_add_function.cpp
martenrichter: Sure..

src/preamble.js
martenrichter: I tried this, but I have reverted, as I suspect failing test to be connect with it, but I am not sure.

src/lib/libaddfunction.js
martenrichter: I am not sure, if this will be beneficial, I would assume the struct in the UA being already aligned. And I would more lean towards 1024

src/lib/libdylink.js
sbc100: Yes, this code needs to run on every thread, not just the first one.  This is because each thread has its own copy of the wasm table (there no shared wasm tables yet).

src/lib/libaddfunction.js
sbc100: If we do go with a fixed constant like this I think it at least warrants a comment explaining how we came to it.

martenrichter: Ok, I have changed to 1024 and added a comment.

test/interop/test_add_function.cpp
martenrichter: Done.

src/lib/libdylink.js
martenrichter: Changed

src/preamble.js
martenrichter: It failed again, I have to setup a more complex dev environment to be able to run browser tests.

src/lib/libaddfunction.js
```diff
@@ -139,27 +139,39 @@ addToLibrary({
...
     if (freeTableIndexes.length) {
       return freeTableIndexes.pop();
     }
+    if (usedTableLength < {{{ from64Expr('wasmTable.length') }}}) {
+      return usedTableLength++;
+    }
     // Grow the table
     try {
+      // we select a number of base 2 for alignment and 1024 showed in benchmarks
+      // a resonable performace increase
```
kripken, line 158:
```suggestion
      // We select a power of 2 for alignment, and 1024 showed good performance
      // in benchmarks.
```

src/preamble.js
sbc100: What test failed?

src/lib/libdylink.js
martenrichter: Multiple dylink related.
For example `test_dylink_many_postsets_reversed`, with 
```
Aborted(Assertion failed: unexpected table size while loading liblib.so: 1033)
/tmp/emtest_zctth__t/emscripten_test_core0_r5bfmre_/liblib.js:680
          ^

RuntimeError: Aborted(Assertion failed: unexpected table size while loading liblib.so: 1033)
    at abort (/tmp/emtest_zctth__t/emscripten_test_core0_r5bfmre_/liblib.js:680:11)
    at assert (/tmp/emtest_zctth__t/emscripten_test_core0_r5bfmre_/liblib.js:278:5)
    at loadModule (/tmp/emtest_zctth__t/emscripten_test_core0_r5bfmre_/liblib.js:1568:11)
Thrown at:
    at abort (/tmp/emtest_zctth__t/emscripten_test_core0_r5bfmre_/liblib.js:680:11)
    at assert (/tmp/emtest_zctth__t/emscripten_test_core0_r5bfmre_/liblib.js:278:5)
    at loadModule (/tmp/emtest_zctth__t/emscripten_test_core0_r5bfmre_/liblib.js:1568:11)
```
So it looks like the dylink related changes have a problem, or that there is an assertion with needs to be adapted to the changes.

martenrichter: And actually this one is simple...., I will try to fix it. I have not seen the meaningful debug output before.

martenrichter: Ok posted the text related to the failing test at the wrong source code location...

martenrichter: Anyway next failure in test-nodecompat:
```
-- begin program output --
/tmp/emtest_8uhwu3mk/emscripten_test_core2_lr19s7wp/create.js:1
"use strict";var Module=typeof Module!="undefined"?Module:{};var ENVIRONMENT_IS_WEB=false;var ENVIRONMENT_IS_WORKER=typeof WorkerGlobalScope!="undefined";var ENVIRONMENT_IS_NODE=true;var ENVIRONMENT_IS_PTHREAD=ENVIRONMENT_IS_WORKER&&self.name?.startsWith("em-pthread");if(ENVIRONMENT_IS_NODE){var worker_threads=require("worker_threads");global.Worker=worker_threads.Worker;ENVIRONMENT_IS_WORKER=!worker_threads.isMainThread;ENVIRONMENT_IS_PTHREAD=ENVIRONMENT_IS_WORKER&&worker_threads["workerData"]=[..]
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    [..]

ReferenceError: usedTableLength is not defined
    at receiveInstance (/tmp/emtest_8uhwu3mk/emscripten_test_core2_lr19s7wp/create.js:1:7511)
    at receiveInstantiationResult (/tmp/emtest_8uhwu3mk/emscripten_test_core2_lr19s7wp/create.js:1:7694)
    at createWasm (/tmp/emtest_8uhwu3mk/emscripten_test_core2_lr19s7wp/create.js:1:8077)
Thrown at:
    at receiveInstance (/tmp/emtest_8uhwu3mk/emscripten_test_core2_lr19s7wp/create.js:1:7511)
    at receiveInstantiationResult (/tmp/emtest_8uhwu3mk/emscripten_test_core2_lr19s7wp/create.js:1:7694)
    at createWasm (/tmp/emtest_8uhwu3mk/emscripten_test_core2_lr19s7wp/create.js:1:8077)
```
This one I do not understand. Probably, somewhere usedTableSize is not properly referenced.

src/lib/libcore.js
```diff
@@ -2290,6 +2290,8 @@ addToLibrary({
   'element': 'anyfunc'
 });
 `,
+  $wasmTable__deps: ['$usedTableLength'],
+  $wasmTable__postset: `usedTableLength = wasmTable.length`,
```
sbc100, line 2294: I don't think the `wasmTable` itself should depend on `usedTableLength`... it should only be folks that want to use `getEmptyTableSlot` that have this included.

Also, I think the `__postset` need to look more like `__postset: () => addAtInit(`usedTableLength = wasmTable.length`),`

I.e. that assignment of `usedTableLength` can only happen after the wasm module has been loaded.

martenrichter: Ok, I introduced the dependencies, as the tests complained about not finding  `usedTableLength`, but it did not fix it.

src/lib/libcore.js
```diff
@@ -2290,6 +2290,7 @@ addToLibrary({
   'element': 'anyfunc'
 });
 `,
+  $wasmTable__postset: () => addAtInit('usedTableLength = wasmTable.length'),
```
hoodmane, line 2293: I think you need to add `$usedTableLength` to `wasmTable__deps`.

hoodmane: Oh I see the other discussion.

martenrichter: Sure, I just removed it. Let's see if the test in ci/cd again complains about a missing usedTableLength.

martenrichter: Ok I have a test environment. But seems impossible (after tracing the calls with logging for wasm allocation) to do the initial setting correctly. I will abandon the idea of usedTableLength and do something else, which is closer to my initial idea, but uses less memory and has some desirable benefits. It will come during the next hours.

martenrichter:
The new attempt uses a different structure for `freeTableIndexes`, which uses less memory and may also be used to easily reuse freed indices.

PR #23634, opened 2025-02-09
Simplify dynamic linking table allocation. NFC

sbc100:
Also add an extra assert that should catch table layout inconsistencies between threads.

I noticed this could be improved while reviewing #23633.


hoodmane (approved):
Looks good to me.

src/lib/libdylink.js
```diff
@@ -648,21 +650,27 @@ var LibraryDylink = {
...
+      if (metadata.tableSize) {
+#if ASSERTIONS
+        assert({{{ from64Expr('wasmTable.length') }}} == tableBase, `unexpected table size while loading ${libName}: ${wasmTable.length}`);
+#endif
 #if DYLINK_DEBUG
-        dbg("loadModule: growing table: " + tableGrowthNeeded);
+        dbg("loadModule: growing table by: " + metadata.tableSize);
 #endif
-        wasmTable.grow({{{ toIndexType('tableGrowthNeeded') }}});
+        wasmTable.grow({{{ toIndexType('metadata.tableSize') }}});
```
kripken, line 669: Before we grew  by `tableBase + metadata.tableSize - wasmTable.length`, that is, we grew to size `tablebase + metadata.tableSize`. Now we grow by `metadata.tableSize`, which grows to `wasmTable.length + metadata.tableSize`. Why is that right?

hoodmane: There's a new assertion that `tableBase` is equal to `wasmTable.length`.

kripken: oh, now I see, thanks. makes sense.

sbc100: > So before it would be possible to load `liba.so` and then load `libb.so` on the main thread and load them in the opposite order on the side thread. But this rules that out?

Yes, this was should already have been impossible, and we have various check and assertions that try to verify that this can never happen.

This just adds one more place we that we can assert this invariant.

PR #23639, opened 2025-02-10
[Draft] FETCHFS: allow dirent operations with http fileserver

Labnann:
Currently FetchFS does not provide any directory information. Due to this, getdents or similar syscalls fail to present proper directory structure to the program.

This patch solves this issue and allows working with directories seamlessly.


Labnann:
Considerations:
TODO:
- Add tests

Later we could use a manifest object to provide:
- Pattern to match to fetch file
- Fetch strategy: whether to start fetching file when getting a directory
  or when requesting for a file.
- Supporting other fileserver structures.

PR #14875, opened 2021-08-16
Remove unused `objectTransfer` command. NFC

sbc100:
This command appears to be unused.  `receiveObjectTranster` is
done when the `run` command is received in the worker and nowhere
else.

PR #23641, opened 2025-02-10
[ptheads] Rename receiveObjectTransfer. NFC

sbc100:
Followup to #14875.

The name receiveObjectTransfer implies that all kinds of things can be received but its really only for the offscreen canvas.

### DIFF
--- a/src/lib/libpthread.js
+++ b/src/lib/libpthread.js
@@ -203,8 +203,8 @@ var LibraryPThread = {
       // linear memory.
       __emscripten_thread_free_data(pthread_ptr);
     },
-    receiveObjectTransfer(data) {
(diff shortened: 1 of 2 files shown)


src/lib/libpthread.js
```diff
@@ -203,17 +203,17 @@ var LibraryPThread = {
       // linear memory.
       __emscripten_thread_free_data(pthread_ptr);
     },
-    receiveObjectTransfer(data) {
 #if OFFSCREENCANVAS_SUPPORT
+    receiveOffscreenCanvas(data) {
```
kripken, line 207: Should this be Canvases, plural, as line 209 receives Canvases?

kripken (approved):

PR #23643, opened 2025-02-10
Move compiler.mjs to tools/ directory. NFC

sbc100:
For consistency with the other JS entry points we have (preprocessor.mjs and acorn-optimizer.mjs) that already live in the tools directory.

### DIFF
--- a/site/source/docs/contributing/developers_guide.rst
+++ b/site/source/docs/contributing/developers_guide.rst
@@ -82,7 +82,7 @@ The :ref:`Emscripten Compiler Frontend (emcc) <emccdoc>` is a python script that
 - **emcc** then calls `emscripten.py <https://github.com/emscripten-core/emscripten/blob/main/tools/emscripten.py>`_
   which performs the final transformation to Wasm (including invoking
   **wasm-emscripten-finalize** from Binaryen) and calls the JS compiler
-  (see ``src/compiler.mjs`` and related files) which emits the JS.
+  (see ``tools/compiler.mjs`` and related files) which emits the JS.
 - If optimizing Wasm, **emcc** will then call **wasm-opt**, run meta-dce, and
   other useful things. It will also run JS optimizations on the JS that is
   emitted alongside the Wasm.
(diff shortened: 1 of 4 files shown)


kripken:
> For consistency with the other JS entry points we have (preprocessor.mjs and acorn-optimizer.mjs) that already live in the tools directory.

Hmm, I would say that `compiler.mjs` is not a standalone tool like those others. It is very tightly bound to the other contents of `src/` like parseTools and the libraries and so forth?

sbc100:
> > For consistency with the other JS entry points we have (preprocessor.mjs and acorn-optimizer.mjs) that already live in the tools directory.
> 
> Hmm, I would say that `compiler.mjs` is not a standalone tool like those others. It is very tightly bound to the other contents of `src/` like parseTools and the libraries and so forth?

In that case we should move `proprocessor.mjs` instead into `src`?  `proprocessor.mjs` and `compiler.mjs` are just two entry points into the same code, and equally not supported as standalone tools, right?

sbc100:
acorn-optimizer.mjs BTW is also not a standalone tool, it is designed to only be run from the emcc.

kripken:
`preprocessor.js` is fairly generic, isn't it? It could be used to preprocess lots of other things. It could also work with very different JS underneath (so the `#if` conditions could be very different). I see it more as a generic tool like say `lit`.

But I guess we may have some assumptions in there that are specific to emscripten atm. And some definitely exist in the acorn optimizer.

But moving the compiler to a directory for helper tools also feels a little odd as it isn't a tool, it is the core thing here.

sbc100:
> `preprocessor.js` is fairly generic, isn't it? It could be used to preprocess lots of other things. It could also work with very different JS underneath (so the `#if` conditions could be very different). I see it more as a generic tool like say `lit`.

It is, but its also just a very smaller wrapper around the code in `src/` (just like `compiler.mjs`).  Its only certainly not something we support being used outside of emcc.

kripken (approved):
Fair enough, I don't feel strongly.

PR #23644, opened 2025-02-10
Revert "Handle failure of lseek in fflush"

sbc100:
Reverts emscripten-core/emscripten#23624

This turned out to be a fix for something that is undefined behaviour.


hoodmane:
What specifically is undefined behavior?

kripken:
@hoodmane `fflush` on an input stream like `stdin` is UB, it turns out.

kripken (approved):

PR #23645, opened 2025-02-10
[test] Improve createPath testing in test_fs_js_api. NFC

sbc100:
This added test coverage for #23603.

### DIFF
--- a/test/fs/test_fs_js_api.c
+++ b/test/fs/test_fs_js_api.c
@@ -41,6 +41,7 @@ EM_JS(void, test_fs_open, (), {
 // createPath should succeed when called on existing paths ( https://github.com/emscripten-core/emscripten/issues/23602 )
 EM_JS(void, test_fs_createPath, (), {
     FS.createPath('/', 'home', true, true);
+    FS.createPath('/', 'home', true, true); // Creating an already existing path
     FS.createPath('/home', 'nested1', true, true);
     FS.createPath('/home', 'nested2', true, true);
     FS.writeFile('/home/nested1/test.txt', 'a=1\nb=2\n');


test/fs/test_fs_js_api.c
```diff
@@ -41,6 +41,7 @@ EM_JS(void, test_fs_open, (), {
 // createPath should succeed when called on existing paths ( https://github.com/emscripten-core/emscripten/issues/23602 )
 EM_JS(void, test_fs_createPath, (), {
     FS.createPath('/', 'home', true, true);
+    FS.createPath('/', 'home', true, true); // Creating an already existing path
     FS.createPath('/home', 'nested1', true, true);
     FS.createPath('/home', 'nested2', true, true);
```
kripken, line 46: Do these two lines not add `/home/` twice, as the prefix of `nested1/2`? This API creates all parent directories as it goes. And that should have tested the other PR AIUI

JoeOsborn: This API worked like mkdirTree in the JS FS, ignoring failures while creating the intermediate directories.  In the WASMFS JS API, it did /not/ ignore failures and errored out, which caused a problem if you loaded two file_packager packages that overlapped at any prefix of their target paths.

sbc100: The problem is the test as its written is that these two lines did not trigger the exception:

```
    FS.createPath('/home', 'nested1', true, true);
    FS.createPath('/home', 'nested2', true, true);
```

The is that the first argument to `FS.createPath` is assumed to already exist so these two lines are the equivalent of running `mkdir('/home/nested1')` then running `mkdir('/home/nested1')`.

In order to recreate the issue the second argument needs to match so that the same mkdir command is run twice.

kripken: I see, the parent is treated differently... makes sense, good catch.

kripken (approved):

test/fs/test_fs_js_api.c
sbc100: > Do these two lines not add `/home/` twice, as the prefix of `nested1/2`? This API creates all parent directories as it goes. And that should have tested the other PR AIUI

The API create all directories as it goes from arg1 (path) but simply assumes that arg0 (parent) already exist.   In order to recreate the bug from #23603 we need repetition in arg1 (path)

PR #23647, opened 2025-02-11
Use the same json interchange format preprocessor.mjs as compiler.mjs. NFC

sbc100:
I'm not sure why used `.js` in once case and `.json` in another.

The other nice thing about this change is that it has the preprocessor read the default settings (just like compiler.mjs) so the json file can contain just the changes from the default.

### DIFF
--- a/tools/preprocessor.mjs
+++ b/tools/preprocessor.mjs
@@ -15,7 +15,7 @@
 import assert from 'node:assert';
 import {parseArgs} from 'node:util';
 
-import {loadSettingsFile} from '../src/utility.mjs';
+import {readFile, loadDefaultSettings, applySettings} from '../src/utility.mjs';
 
 const options = {
   'expand-macros': {type: 'boolean'},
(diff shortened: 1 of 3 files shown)


tools/preprocessor.mjs
```diff
@@ -31,11 +31,17 @@ Usage: preprocessor.mjs <settings.json> <input-file> [--expand-macros]`);
...
 
+loadDefaultSettings();
+
 assert(positionals.length == 2, 'Script requires 2 arguments');
+
+// Load settings from JSON passed on the command line
 const settingsFile = positionals[0];
-const inputFile = positionals[1];
+assert(settingsFile, 'settings file not specified');
+const user_settings = JSON.parse(readFile(settingsFile));
```
kripken, line 41:
```suggestion
const userSettings = JSON.parse(readFile(settingsFile));
```

kripken (commented):
lgtm otherwise

tools/preprocessor.mjs
sbc100: Done.  I also updated compiler.mjs where this code was copied from.

PR #23648, opened 2025-02-11
Only serialize modified settings. NFC

sbc100:
This means that when we run the compiler or the preprocessor we don't serialize all setting to JSON, only setting that were are set to something other than the default.

### DIFF
--- a/tools/settings.py
+++ b/tools/settings.py
@@ -144,6 +144,7 @@ def default_setting(name, new_default):
 
 class SettingsManager:
   attrs: Dict[str, Any] = {}
+  defaults: Dict[str, tuple] = {}
   types: Dict[str, Any] = {}
   allowed_settings: Set[str] = set()
   legacy_settings: Dict[str, tuple] = {}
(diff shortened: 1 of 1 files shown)


sbc100:
Depends on #23647

tools/settings.py
```diff
@@ -207,11 +216,10 @@ def dict(self):
...
 
   def external_dict(self, skip_keys={}): # noqa
-    external_settings = {k: v for k, v in self.dict().items() if k not in INTERNAL_SETTINGS and k not in skip_keys}
-    # Only the names of the legacy settings are used by the JS compiler
-    # so we can reduce the size of serialized json by simplifying this
-    # otherwise complex value.
-    external_settings['LEGACY_SETTINGS'] = [l[0] for l in external_settings['LEGACY_SETTINGS']]
+    external_settings = {}
+    for key, value in self.dict().items():
+      if ((key in self.legacy_settings and value) or value != self.defaults.get(key)) and key not in INTERNAL_SETTINGS and key not in skip_keys:
```
kripken, line 221: perhaps add a comment about legacy settings, or keep the old one? otherwise it seems odd why they are mentioned here.

also, why `and value` for legacy settings - do we not need to send falsey values too?

sbc100: Good point, I expanded this add added a comment.

PR #23649, opened 2025-02-11
Avoid stdout file in read_and_preprocess. NFC

sbc100:
The comment here suggests there is bug with stdout streaming in python but we use `stdout=PIPE` when we run `compiler.mjs`, which tends to generate a lot more output that `preprocessor.mjs`

See `compile_javascript` in `emscripten.py`.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -15409,6 +15409,18 @@ addToLibrary({
     self.do_runf('hello_world.c', 'assertions disabled\n4', emcc_args=['-sASSERTIONS=0'])
     self.assertNotContained('#preprocess', read_file('hello_world.js'))
 
+  @crossplatform
+  def test_js_preprocess_huge_file(self):
+    # Check that huge files can be preprocessed.  We once had issues with files
+    # larger then 64k being sent over stdout using python's subprocess.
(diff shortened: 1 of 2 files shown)


test/test_other.py
```diff
@@ -15409,6 +15409,18 @@ def test_js_preprocess_pre_post(self):
     self.do_runf('hello_world.c', 'assertions disabled\n4', emcc_args=['-sASSERTIONS=0'])
     self.assertNotContained('#preprocess', read_file('hello_world.js'))
 
+  @crossplatform
+  def test_js_preprocess_huge_file(self):
+    # Check that huge files can be preprocessed.  We one had issues with files
```
kripken, line 15414:
```suggestion
    # Check that huge files can be preprocessed.  We once had issues with files
```

test/test_other.py
```diff
@@ -15409,6 +15409,18 @@ def test_js_preprocess_pre_post(self):
...
+    # larger then 64k being send over stded using python's subprocess.
```
kripken, line 15415:
```suggestion
    # larger then 64k being sent over stdin using python's subprocess.
```

kripken (commented):
lgtm

PR #23650, opened 2025-02-11
Use stdin to pass settings to compiler.mjs / preprocessor.mjs     

sbc100:
This avoids writing more files to disc than we need to.  One of my
hopes here is that this will be a little more performant, especially
on windows, where the filesystem isn't the fastest.

In order to aid debugging we stash copied of the these generated
settings using the save_intermediate mechanism.

### DIFF
--- a/tools/building.py
+++ b/tools/building.py
@@ -1224,16 +1224,45 @@ def run_wasm_opt(infile, outfile=None, args=[], **kwargs):  # noqa
   return run_binaryen_command('wasm-opt', infile, outfile, args=args, **kwargs)
 
 
-def save_intermediate(src, dst):
+intermediate_counter = 0
+
+
+def new_intermediate_filename(name):
+  assert DEBUG
+  global intermediate_counter
+  basename = 'emcc-%02d-%s' % (intermediate_counter, name)
+  intermediate_counter += 1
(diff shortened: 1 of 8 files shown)


tools/building.py
```diff
@@ -1224,18 +1224,43 @@ def run_wasm_opt(infile, outfile=None, args=[], **kwargs):  # noqa
...
     shutil.copyfile(src, dst)
 
 
+def write_intermediate(content, name):
+  if DEBUG:
+    dst = 'emcc-%02d-%s' % (save_intermediate.counter, name)
+    save_intermediate.counter += 1
+    dst = os.path.join(shared.CANONICAL_TEMP_DIR, dst)
+    logger.debug('saving intermediate file %s' % dst)
+    utils.write_file(dst, content)
```
kripken, line 1242:
```suggestion
  if DEBUG:
    utils.write_file(dst, content)
    save_intermediate(content, name)
```
Can this not work, to reuse the code?

kripken: Oh, I guess it generates the name there... but perhaps another way to share the code is to pass `content` as an optional param to `save_intermediate`?

sbc100: Refactored.

kripken (approved):

Issue #11496, opened 2020-06-25
Allow catching JavaScript exceptions from C++


RReverser:
Somewhat wider version of #11434, but applicable to synchronous functions as well: right now, even with exception handling enabled, there is no way to catch an exception from JavaScript call in C++ and extract its properties.

This means that things like DOMException can't be gracefully handled on C++ side or e.g. when integrating with C libraries where you want to convert different kinds of DOMException into different error codes.

I propose to make it possible to catch any JS exceptions and rethrow them on C++ side with Embind, so that things like

```cpp
try {
  someValue.call("method");
} catch (val e) {
  // ...handle val
}
```

would work and return a JS handle to the caught exception.

sbc100:
To actually catch real JS exceptions in C++ i think that would require change to the exception handling proposal,  (Is that correct @aheejin).  

However, it seems like it should be possible for embind to catch JS exceptions and wrap them into C++ val objects at the boundary then throw those as C++ exceptions.

sbc100:
Sorry I didn't fully read your plan.. I see that you are proposing the same thing.. deleting my first comment.

aheejin:
@RReverser Are you talking about [new wasm EH proposal](https://github.com/WebAssembly/exception-handling) or the existing Emscripten EH?

RReverser:
@aheejin Those are implementation details that are kind of orthogonal to my proposal; the proposal is mainly about catching exceptions on JS side and rethrowing them as JS value handles with Embind on C++ side. How C++ exceptions are implemented shouldn't matter in this context.

gerald-dotcom:
I badly need this to catch errors related to Fetch API since emscripten's own Fetch API is callback based, and I prefer my own version with val's await()

I'm currently using ugly hacky workaround to catch exceptions.

datdinhquoc:
How's the progress of this feature? I want to catch exception by JSON.parse

RReverser:
I don't think there was any but I'm really hoping it will get picked someday. Offtop: didn't realise it's been 1.5 years since I've submitted it 🤯

aheejin:
I'm not sure what this is proposing.

In the original post,
> there is no way to catch an exception from JavaScript call in C++ and extract its properties.

In https://github.com/emscripten-core/emscripten/issues/11496#issuecomment-649854573,
> the proposal is mainly about catching exceptions on JS side and rethrowing them as JS value handles with Embind on C++ side.

[...]

PR #23653, opened 2025-02-11
Propagate errors through val coroutine hierarchy.

davits:
Fix bug when failure or exception happening in a >1 level of the val coroutine call hierarchy is not being propagated. As a result only deep most level coroutine promise is rejected, other val promises including top most one remain in a perpetual pending state.

The bug can be reproduced by the following example:
``` cpp
emscripten::val fetch(const std::string& url) {
  co_return co_await emscripten::val::global("fetch")(url);
}
emscripten::val json(const std::string& url) {
  auto response = co_await fetch(url);
  co_return co_await response.call<emscripten::val>("json");
}
EMSCRIPTEN_BINDINGS(module) {
  function("json", &json);
}
```
```js
let p = Module.json('invalid.url'); // JS error is printed, but p remains pending
try {
  let p = await Module.json('invalid.url'); // JS error is printed
} catch(e) {
  console.log('Caught coro error!'); // this line is never printed
}
```


davits:
I will add tests with separate commit...
If there is a maintainer for the coroutine part, I would like to discuss couple of small improvements I have in mind.

brendandahl:
> I will add tests with separate commit... If there is a maintainer for the coroutine part, I would like to discuss couple of small improvements I have in mind.

@RReverser added the coroutine part. I think he's been busy with other projects, but may have time.

RReverser:
This PR looks reasonable, and I wanted this functionality myself in the past, but I wasn't (stil am not) sure if we should have coroutines / promises in Embind behave differently from synchronous function calls.

Basically, I believe we should have a central flag that tells Embind "hey, store JS exceptions and make them catch-able from C++" regardless of whether function you're calling is synchronous, Asyncify one or a coroutine one - see https://github.com/emscripten-core/emscripten/issues/11496 for original issue.

Having partial implementation where only one of those types of functions throws in C++ but not others might be surprising to users.

davits:
I understand. I think I can propogate rejections through the promise hierarchy as well...
Side note: I just noticed that coroutine frames are leaking, couldn't find `handle.destroy()` call anywhere, will add to this PR.

The #11496 had been dragged for quite some time, without conclusion...
I see value in the ability to write C++ coroutines which can catch errors from the "fetch" API for the current project my company is working on. If I pursue addition of a new experimental flag, which switches val coroutine rejection propogation to exceptions, will it be merged ? (Note for the future: initial implementation works via exceptions.)

About the improvement:
I want to enhance val to awaiter transformation with `promise_type::await_transform` with an ability to define custom awaitable types via template trait types.
I use that approach in my small coroutine engine which we use internally, and with its help I was able to get seamless integration with emscripten::val coroutines, like so:
```
struct JSAwaitable {...};
template <>
struct await_ready_trait<emscripten::val> {
    static JSAwaitable await_transform(Executor*, emscripten::val&& awaitable) {
        return JSAwaitable(std::move(awaitable));
    }
};
...
emscripten::val fetch(const std::string& url) {
  auto response = co_await emscripten::val::global("fetch")(url);
  co_return co_await response.call<emscripten::val>("text");
}
Task<std::string> do_work() {
  auto text = co_await fetch("data.url");
  ...
}
``` 
After adding similar await_transform in the val::promise it will be possible to write/integrate custom awaitables for custom user coroutines. For my case I will be able to seamlessly integrate in opposite direction and do `co_await do_work();` in the val coroutine.
Basically add these function to the val::promise_type:
```
  val&& await_transform(val&& v) {
    return std::move(v);
  }
  template <typename T>
  decltype(auto) await_transform(T&& obj) {
    using RawT = std::remove_cvref_t<T>;
    return await_ready_trait<RawT>::await_transform(std::forward<T>(obj));
  }
```
What do you think ?

RReverser:
> Side note: I just noticed that coroutine frames are leaking, couldn't find `handle.destroy()` call anywhere, will add to this PR.

Do you actually see the leaks or do you only suspect it because there is no `destroy()`?

Explicit destroy is necessary only in some situations. I believe the way I implemented it back then was to ensure that coroutine destroys itself automatically, but if that doesn't actually work, please raise a separate issue demonstrating the problem.

davits:
Sorry, my mistake: `auto final_suspend() noexcept { return std::suspend_never{}; }` suspend_never at the finish ensures automatic coroutine frame destruction.

davits:
Lets move the discussion about that into a separate PR (or another platform), to not clutter this one.

Getting back to the current PR. It is ready for review.
And can someone please help me identify the problem with the tests ? I didn't change any compile settings and am getting compile settings incompatibility failure for bind.o file. The thing is that tests passed with previous commit and I didn't even change anything in bind.cpp with my last commit.

brendandahl:
As for the compiler errors, it seem something broke from an LLVM update and it appears to be fixed now. I've restarted the testers.

system/include/emscripten/val.h
```diff
@@ -779,7 +788,19 @@ class val::promise_type {
...
+      reject(error);
+    } catch (...) {
+      val error = val(internal::_emval_coro_exception_to_error());
+      reject(error);
+    }
+  }
+
+  // Reject the stored promise due to rejection deeper in the call chain
+  void reject_with(val&& error) {
+    reject(std::move(error));
```
davits, line 803: The tests pass when I comment out this line.
Still no clue why this line is causing link error...

davits: After digging further:
There is a `_tls_base` undefined entry in the symbol table of the bind.o.
When I comment out this line the `_tls_base` entry disappears.
Given it is related to "Thread local storage" I assume that it may be the cause of the link error:
```
wasm-ld: error: --shared-memory is disallowed by bind.o because it was not compiled with 'atomics' or 'bulk-memory' features.
```
If this new information has shed some light on the issue, or if I am digging in the wrong direction, please let me know.

sbc100: I think the issue is that libembind doesn't have a multi-threaded variant.  If you want to add multi-threaded features (such as TLS) then I think you would want to have it inherit from `MTLibrary` instead of `Library` in `tools/system_libs.py`

davits: Found it:
reject() call expands to Signature<...>::get_method_caller() and there is a thread_local variable in that function.
https://github.com/emscripten-core/emscripten/blob/b9dbeac2106b5640ff43da8268806da89df91d08/system/include/emscripten/val.h#L144
When I remove thread_local specifier the test passes. 
Shouldn't the tread_local be ignored for the single threaded build ?

davits:
OK, so it seems like that embind-rtti library is compiled without -pthread flag while the test itself (I am experimenting with the core0.test_pthread_create_embind_stack_check) is compiled with the -pthread flag and they are linked together...
I added the `_emval_coro_reject()` function to the bind.cpp, which calls the above mentioned `reject()` function at the end of the call hierarchy. And since reject is a JS function it goes through the `Signature::get_method_caller()` adding a thread_local storage thus requirement to be compiled with the -pthread flag.

Is it OK to require libembind and libembind-rtti libraries to be compiled with the same threadness as the application and how do I do so for the tests ?

sbc100:
> Turned the libembind into a MTLibrary. The question whether it is acceptable remains.

Seems like a reasonable solution yes.   Its a pretty small library anyway.

davits:
This is my first time contributing to the project, so I am not familiar with the process.
If I need to change something (title for example) to signal that the PR is ready for the review, please let me know.
Currently I am assuming that someone has to find some time to review it, but wanted to make sure :)

src/lib/libemval.js
```diff
@@ -467,33 +467,33 @@ var LibraryEmVal = {
...
-      };
-
       {{{ makeSetValue('resolveHandlePtr', '0', 'Emval.toHandle(resolve)', '*') }}};
-      {{{ makeSetValue('rejectHandlePtr', '0', 'Emval.toHandle(rejectWithCurrentException)', '*') }}};
+      {{{ makeSetValue('rejectHandlePtr', '0', 'Emval.toHandle(reject)', '*') }}};
     }));
   },
+
+  _emval_coro_exception_to_error__deps: ['$Emval', '__cxa_rethrow'],
+  _emval_coro_exception_to_error: () => {
```
sbc100, line 486: Could this have a better name maybe?    It doesn't seem specific to co-routines really.  How about something more like `emval_from_current_cxa_exception`.. or something like that?

davits: Done. Combined existing naming convention with the proposed one.

src/lib/libemval.js
```diff
@@ -467,33 +467,33 @@ var LibraryEmVal = {
...
+  _emval_coro_val_from_current_cxa_exception: () => {
+    try {
+      // Use __cxa_rethrow which already has mechanism for generating
+      // user-friendly error message and stacktrace from C++ exception
+      // if EXCEPTION_STACK_TRACES is enabled and numeric exception
+      // with metadata optimised out otherwise.
+      ___cxa_rethrow();
+    } catch (e) {
+      return Emval.toHandle(e);
+    }
```
sbc100, line 495: @aheejin does this look reasonable to you?  Or is the maybe a better way to do this?

src/lib/libemval.js
RReverser: @davits It really doesn't need to be prefixed with `emval_coro` since it's not specific to coroutines at all as @sbc100 said, and can be reused for any emval invocations in the future.

I think just `emval_from_` should be fine.

davits: Done.

src/lib/libemval.js
```diff
@@ -467,33 +467,33 @@ var LibraryEmVal = {
...
   },
 
-  _emval_coro_suspend__deps: ['$Emval', '_emval_coro_resume'],
-  _emval_coro_suspend: async (promiseHandle, awaiterPtr) => {
-    var result = await Emval.toValue(promiseHandle);
-    __emval_coro_resume(awaiterPtr, Emval.toHandle(result));
+  _emval_coro_suspend__deps: ['$Emval', '_emval_coro_resume',  '_emval_coro_reject'],
+  _emval_coro_suspend: (promiseHandle, awaiterPtr) => {
+    Emval.toValue(promiseHandle)
+      .then(result => __emval_coro_resume(awaiterPtr, Emval.toHandle(result)))
```
RReverser, line 473: You can merge these two into a single `.then(result => ..., error => ...)`.

davits: I find the `.then().catch()` variant more appealing. If you prefer the version you mentioned, or there is another reason to switch, let me know and I'll update it.

RReverser: They are not strictly equivalent / not just API sugar - in general, you don't want the errors from emval_coro_resume handler to reenter C++ again for emval_coro_reject handler in the same coroutine step, they are meant to be caught by the next handler instead.

`then(A).catch(B)` says "resume with function A, and if either the promise or the resumed function A threw, then handle either error with function B", whereas `then(A, B)` says "if succeeded, resume with A, otherwise reject with B" which is closer to our desired semantics.

(another reason is that it saves us a few bytes, but that's a bit less important)

davits: Done.

src/lib/libemval.js
RReverser: Do we wait for @aheejin's review or merge as-is? This piece of code was before this PR (I added it a while back because it was the only way to make it work that I found at the time), so I'm leaning towards let's merge it anyway.

sbc100: ping @aheejin

src/lib/libemval.js
```diff
@@ -467,33 +467,33 @@ var LibraryEmVal = {
...
 
-  _emval_coro_suspend__deps: ['$Emval', '_emval_coro_resume'],
-  _emval_coro_suspend: async (promiseHandle, awaiterPtr) => {
-    var result = await Emval.toValue(promiseHandle);
-    __emval_coro_resume(awaiterPtr, Emval.toHandle(result));
+  _emval_coro_suspend__deps: ['$Emval', '_emval_coro_resume',  '_emval_coro_reject'],
+  _emval_coro_suspend: (promiseHandle, awaiterPtr) => {
+    Emval.toValue(promiseHandle)
+      .then(result => __emval_coro_resume(awaiterPtr, Emval.toHandle(result)),
+            error => __emval_coro_reject(awaiterPtr, Emval.toHandle(error)));
```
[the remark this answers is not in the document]
davits: Done.

sbc100 (approved):

davits:
Just to be clear. Are there any actions required from my side?

Issue #23654, opened 2025-02-11
WebGL: INVALID_OPERATION: useProgram: attempt to use a deleted object


rickg-hcl:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
Please include the output `emcc -v` here
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.73 (cbdb74973483dba58d9a78bf7ef6c0543d1a055f)
clang version 20.0.0git (https:/github.com/llvm/llvm-project 1d810ece2b2c8fab77720493864257f0ea3336a9)
Target: wasm32-unknown-emscripten
Thread model: posix

**Failing command line in full:**
If this is compile or link-time failure please include the full failing command
along with its entire output.

**Full link command and output with `-v` appended:**
Even for runtime issues it helps a lot if you can include the full link command.
Adding `-v` to the link command will show all of the sub-commands run which
can help us diagnose your issue.

at runtime, in a specific timing scenario, the console shows this error:
```
  | blitOffscreenFramebuffer | @ | 206cdb5……e19ccd8ff25.js:9661
-- | -- | -- | --
  | _emscripten_webgl_do_commit_frame | @ | 206cdb5……19ccd8ff25.js:12731
  | $_do_call | @ | 9cb0897……5399.wasm:0x10853ba
  | $em_task_queue_execute | @ | 9cb0897……05399.wasm:0xbbc83e
  | $receive_notification | @ | 9cb0897……5399.wasm:0x18ff27d
  | $em_task_queue_execute | @ | 9cb0897……05399.wasm:0xbbc83e
  | $_emscripten_check_mailbox | @ | 9cb0897……5399.wasm:0x1a4aaf1
  | (anonymous) | @ | 206cdb5……0e19ccd8ff25.js:684
  | callUserCallback | @ | 206cdb5……e19ccd8ff25.js:9203
  | checkMailbox | @ | 206cdb5……e19ccd8ff25.js:9235
```

I have created a PR that avoids showing this harmless error

https://github.com/rickg-hcl/emscripten/compare/master...rickg-hcl-patch-1

sbc100:
Thanks @rickg-hcl.  Would you mind opening a pull request against this repo?

@kainino0x, WDYT?

rickg-hcl:
sorry, I screwed up the base.
https://github.com/emscripten-core/emscripten/pull/23658

PR #23658, opened 2025-02-12
check for valid buffer and program

rickg-hcl:
https://github.com/emscripten-core/emscripten/issues/23654

### DIFF
--- a/src/lib/libwebgl.js
+++ b/src/lib/libwebgl.js
@@ -1083,8 +1083,12 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
 
         gl.bindTexture(0xDE1 /*GL_TEXTURE_2D*/, prevTextureBinding);
         gl.activeTexture(prevActiveTexture);
-        gl.bindBuffer(0x8892 /*GL_ARRAY_BUFFER*/, prevVB);
-        gl.useProgram(prevProgram);
+        if (gl.isBuffer(prevVB)) {
+          gl.bindBuffer(0x8892 /*GL_ARRAY_BUFFER*/, prevVB);
+        }
+        if (gl.isProgram(prevProgram)){
+          gl.useProgram(prevProgram);
+        }
       }
       gl.bindFramebuffer(0x8D40 /*GL_FRAMEBUFFER*/, prevFbo);
       if (prevScissorTest) gl.enable(0xC11 /*GL_SCISSOR_TEST*/);


kainino0x (changes requested):
Thanks for the contribution, but I'm not sure this change works correctly.

In particular I don't really understand how things could get into a state where this is a problem. For example for `prevVB`: if the buffer was already deleted before the state was saved, then nothing will be bound, so `getParameter(gl.ARRAY_BUFFER)`  (line 1004) returns `null`, so `prevVB == null`. When we restore the state we'll call `gl.bindBuffer(gl.ARRAY_BUFFER, null)`, which is correct - it unbinds the temporary array buffer. We don't want to leave a buffer bound if there wasn't one previously, which I think this PR would do.

Obviously there is some sequence of events that triggers the error you provided in the bug report, I'm just not sure what it could be. If you could provide a repro case or at least a step-by-step of how it happens, that would be really useful to understanding what the correct fix is here. Thanks!

rickg-hcl:
I attempted to make a sample, but could not make one that duplicated the situation.
The issue is that somewhere in between gathering the previous program and restoring the previous program, that program was deleted by another thread.
The interaction between the threads is hard to put into a sample, but happens regularly in my application.

@kainino0x ,I can give you credentials (in a private channel) on our sandbox server so you can see the log yourself
https://nomad.myhclsandbox.com:9443/nomad/#/

kainino0x:
A threading issue shouldn't be possible. WebGL itself is single-threaded, so if you're using multi-threaded GL in your program that's achieved by Emscripten's GL proxying feature. When that's used, all WebGL calls still happen on the same JS thread. All of the code in this function happens at once in a single JS task, which can't be preempted by any other JS code.

It's possible there's a browser bug that causes `getParameter` to not update immediately when the buffer is deleted? Does this issue happen in more than one browser engine? (chromium / firefox / safari)

It's also possible there's some way GL can get into this state that I'm not aware of. I'm rusty on knowledge of the GL state machine.

You could try capturing the actual sequence of WebGL calls using [SpectorJS](https://spector.babylonjs.com/) and that might give some insight into how this happens.

You can feel free to email me at kainino@google.com but I won't be able to take much time to look at it. Hopefully these pointers help you find out what's going on.

rickg-hcl:
according to copilot:

The `blitOffscreenFramebuffer` function in the provided code can cause the thread to pause due to the following reasons:

1. **WebGL Context Loss**: If the WebGL context is lost, operations like `bindFramebuffer`, `blitFramebuffer`, and other WebGL calls may block or fail, causing the thread to pause.

2. **GPU Synchronization**: Operations like `blitFramebuffer` involve GPU commands that may require synchronization between the CPU and GPU. This synchronization can cause the thread to pause while waiting for the GPU to complete its tasks.

3. **State Changes**: The function makes multiple state changes (e.g., enabling/disabling scissor test, binding framebuffers, switching programs, etc.). Each state change can introduce overhead and potential pauses, especially if the GPU needs to flush or synchronize its state.

4. **Resource Binding**: Binding resources like textures and buffers can cause the thread to pause if the resources are not readily available or if there is contention for these resources.

I don't think it's the context loss, but it could any of the others.
I will email you credentials and instructions that display this situation readily.

kainino0x:
Sorry I didn't get a chance to look at your link. (I think the link isn't working anymore.)

The screenshot you sent me does confirm that you're using Emscripten's proxying (though if I'd looked at the stack in the bug you'd filed, I would have seen this). But that shouldn't really matter; it's still the case that this function can't be interrupted even with proxying.

> For example for `prevVB`: if the buffer was already deleted before the state was saved, then nothing will be bound, so `getParameter(gl.ARRAY_BUFFER)` (line 1004) returns `null`, so `prevVB == null`. When we restore the state we'll call `gl.bindBuffer(gl.ARRAY_BUFFER, null)`, which is correct - it unbinds the temporary array buffer.

I just decided to try a [standalone test](https://jsgist.org/?src=cb64%2CeJzVVN9P2zAQ_lcs76GBtc0Dbx1UK9B2SMBQAE0TQchNrqlR4nS2w4qq_O87x2lHU4vuhzaJl8S5--67u-_OWVLBMqA9-gUm43PSIQen5JRLiDTPBUvJOU9mmouEtqkCbU6K9pZlm055Cni8WxNwEcOiO9NZitgoFxqERvNhxMQTU4THRyG155D2D3177IciFIcqknyuLeQJpIZFR81YDLJzEIeU6Oc5oGdl9BcdC0KeUDzlPCYZ48LbI8tQlEjnW75t7qlkSYZl7WBfwX6HX8kISWZaz1XP97_DJEmnhYhRGqFZqrq5TKzVl6DyQkag7Hen0Bz9j1aVmpmW7YaukVIbsm4jHjcBIS0UEKUlj7CRD6ZW3ydj0GRA7Kwr7EKjfX-ffDQykOWnm4vzk2oywxRM7SXZ91EEJkk9xyMS51FhXN1vBcjna0hxWXLphfTdarx7mM6EJCnCrbGbgD6xCRFZdW5xFjmXeSJZhvDKdVtpEklgGq6sayTz7LpSR3lJ2iZ3rlVpE_eQ71cVTYrpFCSmSdKa_riyeLYUCdiT0gPBM2YuwAgTg4eTP-qb4ROjmcpT6KZ54rWOb0ejYdBq16yGooEgZILDqVMMgmDw9cEGrWNaeyYIq3kBxC8n1sXPVR1UY0inT1rkvWFsupzxOJYrZrrUjQofjs8uT88uxy8ZN8CNKtd4Z5oYt2St9br1Cok8Lu9bbdaR-Cr4PA4GF7go9Zo7i8PrWq-6t4L9VMjhdAvURG1I1HTuFunkNgiGlzcPdQuv6dOEvrIH231uLMLb7xT7Maadv5NGSYPRzTBwR_XqXXAt29_fk_92U36x_D-Y7T-ebmle-KDlffkDzxgPTw%3D%3D) of this. I realized that I tested the buffer case, but I did not test the program case. It turns out that `gl.getParameter(gl.CURRENT_PROGRAM)` _will_ return deleted objects, unlike `gl.getParameter(gl.ARRAY_BUFFER_BINDING)`:

```
BUFFER [object WebGLBuffer]
  bindBuffer(ARRAY_BUFFER, buffer)
  isBuffer(buffer) -> true
  getParameter(ARRAY_BUFFER_BINDING) -> [object WebGLBuffer]
  deleteBuffer(buffer)
  isBuffer(buffer) -> false
  getParameter(ARRAY_BUFFER_BINDING) -> null
PROGRAM [object WebGLProgram]
  useProgram(program)
  isProgram(program) -> true
  getParameter(CURRENT_PROGRAM) -> [object WebGLProgram]
  deleteProgram(program)
  isProgram(program) -> true
  getParameter(CURRENT_PROGRAM) -> [object WebGLProgram]
```

I did not know about this, but it is the [specified behavior](https://registry.khronos.org/OpenGL-Refpages/es2.0/xhtml/glDeleteProgram.xml):

> If a program object is in use as part of current rendering state, it will be flagged for deletion, but it will not be deleted until it is no longer part of current state for any rendering context.

But of course we _do_ take it out of the current state, so then it gets deleted. This seems to make it impossible to fully restore the previous GL state. However, we can at least avoid this error. _Hopefully_ the application is not still relying on the deleted program being usable.

For buffers, I'm not aware of any such logic, however. Can you confirm whether you've actually seen the crash on `prevVB`, or if you've only seen it on `prevProgram`?

kainino0x:
I have a proposed fix for `prevProgram`: https://github.com/emscripten-core/emscripten/pull/23980
It's very similar to yours but it also will unbind the program if there was none bound in the first place.

If you're still able to test it, please try this.

Also please let me know about `prevBuffer`. If there's still an issue there I will have to keep digging....

kainino0x:
Thank you for confirming! I will land that soon, then.
(Apologies for taking over the change, it was just a bit easier!)

PR #23659, opened 2025-02-12
[AUDIO_WORKLET] add parameter tests

cwoffenden:
This is splits out the test changes from #23508 adding an interactive parameter test:

```
```

It also tidies a little the shared code (and related tests) in the process. Interactive tests had `_2gb` and `_4gb` options added (which for the audio fail until #23508 lands).


test/webaudio/audioworklet_test_shared.inc
```diff
@@ -5,18 +5,32 @@
 // Helper for MEMORY64 to cast a void* to an audio context or type
 #define VOIDP_2_WA(ptr) ((EMSCRIPTEN_WEBAUDIO_T) (intptr_t) ptr)
 
-// Count the audio callbacks and return after 375 frames (1 second with the default 128 size)
+// Attribute that marks a function, parameter, variable, etc., as unused
+// (silencing any -Wunused warnings).
+#ifndef __unused
+#define __unused __attribute__((unused))
+#endif
```
sbc100, line 12: I don't remember having to do this for other tests.   Are these tests build with different warnings enabled?

Maybe just build with `-Wno-unused` to keep the code simple?

test/webaudio/audioworklet_test_shared.inc
```diff
@@ -28,13 +42,18 @@ EM_JS(EMSCRIPTEN_WEBAUDIO_T, createTrack, (EMSCRIPTEN_WEBAUDIO_T ctxID, const ch
   var context = emscriptenGetAudioObject(ctxID);
   if (context) {
     var audio = document.createElement('audio');
+#if __wasm64__
+    // Number() wrapper is a workaround for UTF8ToString() needing a JS number
+    // and from64() not being available in EM_JS macros. Fix in UTF8ToString?
+    url = Number(url);
+#endif
     audio.src = UTF8ToString(url);
```
sbc100, line 50: Just do `UTF8ToString(Number(url));` unconditionally to avoid the `if/endif` to keep the code simple?

sbc100 (commented):
lgtm in general, thanks for continuing to work on this.

test/webaudio/audioworklet_test_shared.inc
cwoffenden: I wrote these tests standalone and I'm used to working as `-Wpedantic`.  I can take these out, it'll build without.

test/webaudio/audioworklet_test_shared.inc
cwoffenden: I did, then I didn't, and flip-flopped over it. I'll keep it simple.

test/test_interactive.py
```diff
@@ -311,33 +311,55 @@ def test_audio_worklet_stereo_io(self):
     os.mkdir('audio_files')
     shutil.copy(test_file('webaudio/audio_files/emscripten-beat.mp3'), 'audio_files/')
     shutil.copy(test_file('webaudio/audio_files/emscripten-bass.mp3'), 'audio_files/')
-    self.btest_exit('webaudio/audioworklet_in_out_stereo.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS'])
+    self.btest_exit('webaudio/audioworklet_in_out_stereo.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-DTEST_AND_EXIT=1'])
```
sbc100, line 314: With interactive tests so you want `TEST_AND_EXIT` defined?  I thought maybe the point of that macro was for use in the browser tests, not the interactive tests.

If its defiend in both then maybe we don't need it at all?

test/test_browser.py
```diff
@@ -5495,17 +5495,21 @@ def test_audio_worklet_post_function(self, args):
...
-  def test_audio_worklet_stereo_io(self, args):
+  @no_wasm64('https://github.com/emscripten-core/emscripten/pull/23508')
+  @no_2gb('https://github.com/emscripten-core/emscripten/pull/23508')
+  @requires_sound_hardware
+  def test_audio_worklet_params_mixing(self, args):
     os.mkdir('audio_files')
     shutil.copy(test_file('webaudio/audio_files/emscripten-beat.mp3'), 'audio_files/')
     shutil.copy(test_file('webaudio/audio_files/emscripten-bass.mp3'), 'audio_files/')
-    self.btest_exit('webaudio/audioworklet_in_out_stereo.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-DBROWSER_TEST'] + args)
+    self.btest_exit('webaudio/audioworklet_params_mixing.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-DTEST_AND_EXIT'] + args)
```
sbc100, line 5512: You use `-DTEST_AND_EXIT=1` in some places just `-DTEST_AND_EXIT` here.   If the later is enough, perhaps just use that everywhere?

cwoffenden: Without the parameter one second, with it's 10 otherwise the interactive test is too short to interact.

Maybe I change to `1` for a second, `10` for ten seconds?

test/test_interactive.py
cwoffenden: If interactive is meant to run continuously then it can be removed. I wasn't sure of the use case.

test/test_browser.py
sbc100: Maybe just make the interactive tests run forever?  Or until some button is clicked.   I'm not really sure that the overriding pattern is for them.   But whatever it is perhaps you don't need to defined `TEST_AND_EXIT` at all?

test/test_interactive.py
cwoffenden: I guess interactive needs the user to shut it down, in which case I'll remove it here and delete the one or 10 second selection.

test/test_browser.py
sbc100: BTW, do you ever run the interactive tests more than one at a time (i.e. a sequence of them?).  I never got into the habit of doing so and I imagine many of them don't exit correctly.. but maybe thats fine.

cwoffenden: I've never run them unattended, since some will just time out and fail waiting, and I only run the ones related to what I'm looking at.

I'll remove `TEST_AND_EXIT` for interactive, that simplifies things more. It'll be tomorrow now.

cwoffenden:
All review suggestions done:

- `__unused` removed
- `TEST_AND_EXIT` only for browser tests
- Interactive runs until the window is closed

test/webaudio/audioworklet_2x_in_hard_pan.c
```diff
@@ -65,6 +75,13 @@ void processorCreated(EMSCRIPTEN_WEBAUDIO_T context, bool success, void* data) {
...
 
-  // Register the counter that exits the test after one second of mixing
+#ifdef TEST_AND_EXIT
+  // Register the counter that exits the test after one second of playback
   emscripten_set_timeout_loop(&playedAndMixed, 16, NULL);
+#endif
+}
+
+// This implementation has no custom start-up requirements
+EmscriptenStartWebAudioWorkletCallback getStartCallback(void) {
```
sbc100, line 85: Why does this need to be a function that returns a function?  Why not just declare a function called `startCallback`?

sbc100 (commented):
lgtm % comment

test/webaudio/audioworklet_2x_in_hard_pan.c
cwoffenden: It was because most tests return and use this default function in the shared code:

https://github.com/emscripten-core/emscripten/blob/a55981e08e59d14774f81f550a83904fa197f099/test/webaudio/audioworklet_test_shared.inc#L78

Whereas one needed a custom implementation, so this was the cleanest way to keep most of the code shared.

cwoffenden: (Cleanest way and still keeping it C)

sbc100 (approved):

PR #23660, opened 2025-02-12
[SDL3] Inline build config header. NFC

sbc100:
This matches the behavior of existing ports.  I don't love these huge inline headers, but we should probably be consistent for now.


sbc100:
See https://github.com/emscripten-core/emscripten/pull/23630#issuecomment-2654246391

kripken (approved):

ypujante:
@sbc100 

1) I am definitely in favor of consistency. If you decide to go down the route of extracting the headers from the port file, then at the minimum I would suggest have a subfolder for them (something like `tools/ports/includes/sdl3/....h`) instead of "polluting" `tools/ports`

2) When I was working on the sdl3 port as well, I found it cumbersome and error prone to have to essentially reverse engineer what cmake is already doing. It would be amazing if the port file could instead do:

```python
# fetch the zip file
# run emcmake cmake ...
# cmake --build . --target libSDL3
# export libSDL3.a
```

Since cmake is the supported way of building, then it is going to do the right thing. Also, it would be more trivial to add port option that would simply be translated to cmake options...

I know that cmake is not part of emcc so I am sure that is not going to happen, but that would be the right way to do it though...

sbc100:
> @sbc100
> 
> 1. I am definitely in favor of consistency. If you decide to go down the route of extracting the headers from the port file, then at the minimum I would suggest have a subfolder for them (something like `tools/ports/includes/sdl3/....h`) instead of "polluting" `tools/ports`
> 2. When I was working on the sdl3 port as well, I found it cumbersome and error prone to have to essentially reverse engineer what cmake is already doing. It would be amazing if the port file could instead do:
> 
> ```python
> # fetch the zip file
> # run emcmake cmake ...
> # cmake --build . --target libSDL3
> # export libSDL3.a
> ```
> 
> Since cmake is the supported way of building, then it is going to do the right thing. Also, it would be more trivial to add port option that would simply be translated to cmake options...
> 
> I know that cmake is not part of emcc so I am sure that is not going to happen, but that would be the right way to do it though...

Yes, that probably with that approach is that we don't have cmake or ninja/make as emcc dependencies, so right now no core ports can call out to those executables.   External ports, can run cmake if they want to take on the extra dependencies.     The other potential problem with executing thinks like make and cmake from inside the compiler relates to locking.. we hold a global lock on the cache while building a port.

Note that if we were to extend ports to include basically all cmake projects I think that we should probably instead look into just using an external ports system such as vcpkg (https://vcpkg.io/en/) rather then having ports magically built into the compiler like we have today.   I believe vcpkg already has some support for emscripten.

PR #23662, opened 2025-02-12
Update copyright year reported by --version. NFC

sbc100:
It seems the reporting the most recent year is what tools like gcc do:

```
$ gcc --version
gcc (Debian 14.2.0-3+build3) 14.2.0
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -7,6 +7,7 @@
 # noqa: E241
 
 from functools import wraps
+from datetime import datetime
 import glob
 import gzip
 import importlib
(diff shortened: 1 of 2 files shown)


kripken (commented):
Perhaps add a test that the date is the current date? That way the test will error once a year, reminding us to update emcc.py with the new date.

sbc100:
> Perhaps add a test that the date is the current date? That way the test will error once a year, reminding us to update emcc.py with the new date.

Hmm... do we really want tests that that will fail based on current time though?  I'm not sure its worth it.

sbc100:
> I agree that's not ideal. But how else will we remember to update this?

I think the risk of not remembering is very low..

sbc100:
Happy to add the check if you think its worth it.

kripken:
I just think that if we are going to change policy from "initial year" to "latest year" then it makes sense to verify that.

Otherwise when it is 2030 we'll get questions "why is it saying 2025, that's an odd year to pick specifically, since the project began much earlier... is my install out of date?" etc.

sbc100:
Done

test/test_other.py
```diff
@@ -484,8 +485,10 @@ def test_emcc_basics(self, compiler, suffix):
     # --version
     output = self.run_process([compiler, '--version'], stdout=PIPE, stderr=PIPE)
     output = output.stdout.replace('\r', '')
+    # This test will break once per year. Sorry.
```
kripken, line 488:
```suggestion
    # This test will break once per year if we forget to update (sorry).
```

kripken (approved):
lgtm with or without the suggested change.

Sorry for insisting here, but I do feel that moving from "initial year" to "arbitrary year" is a regression. And that it would end up arbitrary without something to make sure we remember to update. I'd be open to other ways to remind us, but a test seems the most robust...

PR #23664, opened 2025-02-13
Avoid converting --pre-js/--post-js files to absolute paths. NFC

sbc100:
Now that we change CWD when running `compiler.mjs` (See #23519) we no longer need to convert these to absolute paths in `link.py`.

This also makes explicit the difference between system and user includes in `jsifier.mjs`.


kripken (approved):
Nice!

Issue #23131, opened 2024-12-11
[AUDIO_WORKLET] tests in the CI don't appear to be run (in test_browser)


cwoffenden:
EDIT: short version, since I think the error is misleading: they're compiling but they're instantly exiting without running.

Testing with the latest Emscripten locally (3.1.74) and seeing the same in the result from CI. Open any of the CI runs and look at the browser tests, e.g.:

https://app.circleci.com/pipelines/github/emscripten-core/emscripten/38495/workflows/353c8dd2-db7a-4691-a83b-c1b8088f8959/jobs/859340

Search for `test_audio_worklet` and the parameterised tests all have the same 404 error with missing files:

```
  test_audio_worklet (test_browser.browser) ... cache:INFO: generating system asset: symbol_lists/c9305f6ae33005a7b005afb88a5689b698a1f55c.json... (this will be cached in "/root/cache/symbol_lists/c9305f6ae33005a7b005afb88a5689b698a1f55c.json" for subsequent builds)
cache:INFO:  - ok
127.0.0.1 - - [11/Dec/2024 12:08:43] code 404, message File not found: /root/project/test.aw.js
```

I noticed this locally, and thought it was just the way my system was set up, then given that I'd seen there was no way the CI was passing tests that should be failing, I dug deeper. The test runner doesn't find `test.aw.js` and if that is copied (or a link created) then it next fails on `test.js`.

The tests are all passing though, I'm guessing because the main page opens and exits (interestingly it's only the Audio Worklets missing files, so 14 404s in `test_browser`'s log).

cwoffenden:
@sbc100 As per the discussion here:

https://github.com/emscripten-core/emscripten/pull/23508#issuecomment-2616858893

I'd like to get these tests running (at least not silently failing). I've put some fixes in already (e.g. `STRICT`), with the remaining fixes in #23508.

My question is, if I fix these they'll break the CI until other fixes are merged. What's the best way to do this? Would you like a separate PR that breaks things, or do they want rolling into #23508?

The more I delve into it the more issues I'm finding, the latest being audio params (`WebAudioParamDescriptor`) don't appear to be wired up at all (and no tests exist). I can fix the code to correctly read structs based on the correct offsets, but without any tests and being actually wired up, it's a shot in the dark.

cwoffenden:
[...]

PR #23665, opened 2025-02-13
[CI] Enable audio tests in Chrome

cwoffenden:
Step one of a fix for #23131 (the second part in #23695).

Enabled Chrome's `FakeAudioOutputStream` for the CI machines and bypassed the need for user interaction. The Chrome tests are now enabled with `@requires_sound_hardware`.

Many attempts at replicating the same functionally were tried with Firefox (details below) but audio would never run so its `EMTEST_LACKS_SOUND_HARDWARE` was left unchanged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -362,8 +362,6 @@ commands:
             user_pref("javascript.options.shared_memory", true);
             user_pref("javascript.options.wasm_memory64", true);
             user_pref("dom.postMessage.sharedArrayBuffer.bypassCOOP_COEP.insecure.enabled", true);
-            user_pref("media.navigator.streams.fake", true);
-            user_pref("media.autoplay.default", 0);
             EOF
       - run:
           name: run tests (<< parameters.title >>)
@@ -373,12 +371,12 @@ commands:
             #       support in headless mode) resolves
             EMTEST_LACKS_GRAPHICS_HARDWARE: "1"
             EMTEST_LACKS_WEBGPU: "1"
+            # TODO: replicate with FF the same as CHROME_FLAGS_AUDIO
+            EMTEST_LACKS_SOUND_HARDWARE: "1"
             # OffscreenCanvas support is not yet done in Firefox.
             EMTEST_LACKS_OFFSCREEN_CANVAS: "1"
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
             DISPLAY: ":0"
-            # Note: as per Chrome, the FF profile enables a dummy audio device
-            # (so doesn't need the EMTEST_LACKS_SOUND_HARDWARE var)
           command: |
             export EMTEST_BROWSER="$HOME/firefox/firefox -headless -profile $HOME/tmp-firefox-profile/"
             # There are tests in the browser test suite that using libraries
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,8 @@ commands:
       - run:
           name: download chrome
           command: |
+            # Try adding a dummy sound device for Chrome...
+            modprobe snd-dummy
             # TODO: Make these part of the base image
             apt-get install -q -y libu2f-udev libvulkan1 xdg-utils
             # wget -O ~/chrome.deb https://dl.google.com/linux/direct/google-chrome-beta_current_amd64.deb
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -306,7 +306,6 @@ commands:
       - run:
           name: run tests (<< parameters.title >>)
           environment:
-            EMTEST_LACKS_SOUND_HARDWARE: "1"
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
             # --no-sandbox because we are running as root and chrome requires
             # this flag for now: https://crbug.com/638180
@@ -314,7 +313,9 @@ commands:
             CHROME_FLAGS_HEADLESS: "--headless=new --remote-debugging-port=1234"
             CHROME_FLAGS_WASM: "--enable-experimental-webassembly-features --js-flags=\"--experimental-wasm-stack-switching --experimental-wasm-type-reflection\""
             CHROME_FLAGS_NOCACHE: "--disk-cache-dir=/dev/null --disk-cache-size=1 --media-cache-size=1 --disable-application-cache --incognito"
-            CHROME_FLAGS_AUDIO: "--autoplay-policy=no-user-gesture-required --use-fake-device-for-media-stream"
+            # The runners lack sound hardware so fallback to a dummy device (and
+            # bypass the user gesture so audio tests work without interaction)
+            CHROME_FLAGS_AUDIO: " --use-fake-device-for-media-stream --autoplay-policy=no-user-gesture-required"
           command: |
             export EMTEST_BROWSER="/usr/bin/google-chrome $CHROME_FLAGS_BASE $CHROME_FLAGS_HEADLESS $CHROME_FLAGS_WASM $CHROME_FLAGS_NOCACHE $CHROME_FLAGS_AUDIO"
             # There are tests in the browser test suite that using libraries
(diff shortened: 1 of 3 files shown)


sbc100 (commented):
nice!

Does the test pass locally for you now?

cwoffenden:
> Does the test pass locally for you now?

Now checked, and yes, all the parameterised `test_audio_worklet` tests apart from the `_memory64` variant work (and now with added asserts that one fails early instead of timing out).

I'll work my way through the other browser tests next week now.  It would be useful to have #23659 merged (which doesn't affect anything other than my code) so I can move all the browser audio tests to use the same pattern.

cwoffenden:
The latest flags for Chrome I tried appear to work, with the AW tests for regular Chrome passing:

https://app.circleci.com/pipelines/github/emscripten-core/emscripten/40523/workflows/bca5a785-e04e-4bf1-8187-01a87982d059/jobs/899865?invite=true#step-108-24450_168

For Firefox it's failing as expected (since I've not added its fake audio flags (`media.navigator.streams.fake` from what I can make out), and the wasm64/2GB tests are failing (needing my `MEMORY64` PR).

.circleci/config.yml
```diff
@@ -314,8 +314,9 @@ commands:
             CHROME_FLAGS_HEADLESS: "--headless=new --remote-debugging-port=1234"
             CHROME_FLAGS_WASM: "--enable-experimental-webassembly-features --js-flags=\"--experimental-wasm-stack-switching --experimental-wasm-type-reflection\""
             CHROME_FLAGS_NOCACHE: "--disk-cache-dir=/dev/null --disk-cache-size=1 --media-cache-size=1 --disable-application-cache --incognito"
+            CHROME_FLAGS_AUDIO: "--autoplay-policy=no-user-gesture-required --use-fake-device-for-media-stream"
           command: |
-            export EMTEST_BROWSER="/usr/bin/google-chrome $CHROME_FLAGS_BASE $CHROME_FLAGS_HEADLESS $CHROME_FLAGS_WASM $CHROME_FLAGS_NOCACHE"
+            export EMTEST_BROWSER="/usr/bin/google-chrome $CHROME_FLAGS_BASE $CHROME_FLAGS_HEADLESS $CHROME_FLAGS_WASM $CHROME_FLAGS_NOCACHE $CHROME_FLAGS_AUDIO"
```
sbc100, line 319: If we make this change should we also remove the `EMTEST_LACKS_SOUND_HARDWARE` setting above?

I'm kind of inclined to make this change as its one separate PR, e.g. `[CI] Enable audio testing on chrome`?

cwoffenden: I didn't have much time today so it was mostly experimenting and coming back to it, but:

- I'll remove the`EMTEST_LACKS_SOUND_HARDWARE`
- I'll rename to match its new purpose (I wasn't sure what I'd achieve when starting)

Is it fine to always enable the dummy sound hardware and bypass the media interaction? I can't think what this would break (and where real sound hardware is present its used as a preference, at least from my manual testing).

sbc100: I imagine enabling dummy sounds hardware and bypass the interaction check should be fine for all out tests..

cwoffenden:
Hmm, FF still broken. To do:

- look at `media.audio_loopback_dev`
- `EMTEST_LACKS_SOUND_HARDWARE` for FF

test/webaudio/audioworklet.c
```diff
@@ -132,4 +143,9 @@ int main() {
   // and kick off Audio Worklet scope initialization, which shares the Wasm
   // Module and Memory to the AudioWorklet scope and initializes its stack.
   emscripten_start_wasm_audio_worklet_thread_async(context, wasmAudioWorkletStack, sizeof(wasmAudioWorkletStack), WebAudioWorkletThreadInitialized, 0);
+
+#ifdef TEST_AND_EXIT
+  // We're in the test harness and exiting is via main_thread_tls_access()
+  emscripten_exit_with_live_runtime();
+#endif
```
sbc100, line 150: I was thinking or a separate PR to enable audio testing CI.   

PR1: enable audio testing in CI
PR2: fix audio worklet test so they fail correctly

cwoffenden: Sure, NP. I’ll see how I get on with FF on Monday but I might split it for Chrome first too (I set up a VM with the same Debian and FF and so far, even with no audio device, it doesn’t fail; left to try is headless).

cwoffenden: After getting an audio-less Debian VM set-up without all the fallback sound devices removed, I finally managed to recreate what Firefox does. It fails here:

https://github.com/emscripten-core/emscripten/blob/f9a781b5b69161b31e3bdcd785fd7edadb24ddb8/test/webaudio/audioworklet.c#L60

The promise calls neither the success nor failure route, it just silently fails and the audio remains permanently suspended. I'll call this a Firefox bug and keep the `EMTEST_LACKS_SOUND_HARDWARE` (I'm satisfied that I've spent enough time going down all avenues and what remains is creating a repro in minimal JS for Firefox).

.circleci/config.yml
cwoffenden: I limited removing `EMTEST_LACKS_SOUND_HARDWARE` to Chrome (FF I can't get to work, documented in the comments).

test/test_browser.py
```diff
@@ -5490,8 +5489,9 @@ def test_audio_worklet_post_function(self, args):
     '': ([],),
     'closure': (['--closure', '1', '-Oz'],),
   })
+  @requires_sound_hardware
   def test_audio_worklet_modularize(self, args):
-    self.btest_exit('webaudio/audioworklet.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-sMODULARIZE=1', '-sEXPORT_NAME=MyModule', '--shell-file', test_file('shell_that_launches_modularize.html')] + args)
+    self.btest_exit('webaudio/audioworklet.c', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-sMODULARIZE=1', '-sEXPORT_NAME=MyModule', '--shell-file', test_file('shell_that_launches_modularize.html'), '-DTEST_AND_EXIT'] + args)
```
sbc100, line 5494: Revert the `-DTEST_AND_EXIT` part of this change

sbc100 (approved):
Great!

Please update the PR description too

test/test_browser.py
cwoffenden: I'm just doing so (and moving it to the other PR). 😀

cwoffenden:
Tested locally, this doesn't break anything, whereas #23695 breaks the CI as expected.

sbc100 (approved):
LGTM with update PR description.   Do you want to also remove the "Draft" status of this PR?

cwoffenden:
Funny, wasm64 tests like `test_sdl2_mixer_wav_dash_l` are failing now audio is enabled (since they were flagged as needing audio). I'll disable those tomorrow for now.

sbc100:
Ah.. good to know.  I can take a look at `browser64.test_sdl2_mixer_wav_dash_l`.   If you find any more of those failing tests can you open bugs for them?

cwoffenden:
There seems to be quite a few. I'll make a list tomorrow and mark them as `@no_wasm64`.

cwoffenden:
They're parameterised so it's only `test_sdl2_mixer_wav` and `test_sdl2_mixer_music`. I marked them not to run for now just to verify. I'll see how the CI finished and create bugs tomorrow (I really am done for the day...!).

test/test_browser.py
```diff
@@ -3244,6 +3245,7 @@ def test_sdl2_mixer_wav(self, flags):
     # TODO: need to source freepats.cfg and a midi file
     # 'mod': (['mid'],    'MIX_INIT_MID', 'midi.mid'),
   })
+  @no_wasm64('TODO: SDL2 audio memory64')
```
sbc100, line 3248: I just submitted a PR on the SDL side: https://github.com/libsdl-org/SDL/pull/12332

Can you use that URL as the string argument to `no_wasm64` here.

cwoffenden: Done.

sbc100 (approved):

Issue #23667, opened 2025-02-13
Expose wasmfs_create_file to JS


JoeOsborn:
Right now there's no way for JS code to call `wasmfs_create_file` directly with a specific backend, which is a key use case for WASMFS.  I suggest adding the following function to `libwasmfs.js`:

```
createFile(path, mode, backend) {
  return FS.handleError(withStackSave(() => __wasmfs_create_file(module.stringToUTF8OnStack(path), mode, backend)));
}
```

We already have ways to create backends in JS code (e.g. OPFS and FETCHFS both define `createBackend`), but no way to use them apart from `FS.mount` which is unsuitable for single files.

PR #19825, opened 2023-07-10
WasmFS JS API: Implement mount and unmount

jameshu15869:
This PR implements `FS.mount` and `FS.unmount`. There were existing tests, so I just added some additional test cases to those files. Tests that now pass in WasmFS mode were also added to CI.

There are some differences in the behavior of the WasmFS `mount` and the legacy API. For example, WasmFS `mount` will create a new directory to mount to if one does not already exist. Existing directories must be nonempty, since WasmFS will unlink the directory before creating a new directory with the specified backend. 

Currently, WasmFS does not support mounting `IDBFS`, `WORKERFS`, and `PROXYFS`, although this may be added in the future using a bridge to legacy backends (https://github.com/emscripten-core/emscripten/issues/18935)

PR #23668, opened 2025-02-13
Add createFile to JS API for wasmfs (#23667)

JoeOsborn:
Per #23667, add a createFile function to the WASMFS JS API that calls out to `wasmfs_create_file`.

### DIFF
--- a/src/lib/libwasmfs.js
+++ b/src/lib/libwasmfs.js
@@ -98,6 +98,7 @@ FS.init();
       FS.ErrnoError.prototype = new Error();
       FS.ErrnoError.prototype.constructor = FS.ErrnoError;
     },
+    createFile: (path, mode, backend) => FS.handleError(withStackSave(() => _wasmfs_create_file(stringToUTF8OnStack(path), mode, backend))),
     createDataFile(parent, name, fileData, canRead, canWrite, canOwn) {
       FS_createDataFile(parent, name, fileData, canRead, canWrite, canOwn);
     },


sbc100 (commented):
lgtm, but can you date the test_fs_js_api test?

JoeOsborn:
I'm read "date" as "update", hope that's what you meant!

test/fs/test_fs_js_api.c
```diff
@@ -469,6 +469,17 @@ void cleanup() {
     remove("closetestfile");
 }
 
+#if WASMFS
```
sbc100, line 472: This API is not specific to WASMFS is it?   I see it exists in `libfs.js` too.

JoeOsborn: Oh indeed it does, I had no idea.  But this version of the test is wasm-specific since it uses a backend.  It also looks like the libfs version ignores the "properties" field...

sbc100: I think the goal with the JS API is mostly feature parity with the existing/old FS.   If possible I thin we should avoid extending the JS API.

But yes, if we decide we really do need a new API we should give it a new name.

Would you make making a version of createFile that *is* compatible with the old FS and testing that as part of this PR?  Or as a separate PR?

JoeOsborn: I've made WASMFS createFile compatible with the JS version.  The JS version never used the third parameter, so I'm using it for the backend here.

test/fs/test_fs_js_api.c
```diff
@@ -469,6 +469,23 @@ void cleanup() {
...
+  EM_ASM(
+         var backend = MEMFS.createBackend({});
+         var file = FS.createFile("/", "test.txt", backend, true, true);
+         if (file <= 0) throw "No file created";
+         FS.close(file);
+         );
+#else
+  EM_ASM(
+         var file = FS.createFile("/", "test.txt", {}, true, true);
```
sbc100, line 482: Can we make sure that this simple usage (without a backend) is works with both the old FS and with wasmfs.

In fact, it might be best not to expose the backend directly here, since then the two APIs would not be compatible.   

In fact I think the for in `libfs.js` which has createFile call through to `FS.create` should probably work fine under wasmfs too, so maybe just copy it verbatim and at a test for it.

Do we really need a new API for creating a backend-specific file?

sbc100: The fact that no other `FS.xxx` APIs for wasmfs take the backend as an argument suggest that perhaps this is not needed in this case either.

JoeOsborn: It can’t be copied verbatim mainly because the JS version handles either a path or a node for the parent, but the wasm version doesn’t have the concept of a node. AFAIK there’s no way to get the path to a file descriptor/wasmfs file.

The reason the FS.apis in wasmfs don’t take a backend is presumably because they’re meant to mimic JS FS APIs which only support a 1:1 relationship between backends and mount points.  But wasmfs backends aren’t necessarily mounted somewhere. Fetchfs, for example, has supported this on the C side (a fetchfs whose url is a specific file, meant to be used to create a file). The tests for the JS wasmfs backend show a similar usage.  The wasmfs C API seems to use the metaphor of creating files and directories within specific backends, but we don’t have a way to express that in JS yet.

As for compatibility, in this case the third argument seems to have no meaning in the JS version (it’s unused), so it shouldn’t be a compatibility issue.  But a different name can be picked if this is a problem.

sbc100: I think we have two different issue here perhaps.

The first issue is that WASMFS lacks the FS.createFile API that the old FS has.   Step 1 would be to add this API and add tests for it.  (it should work in both WASM FS and the old FS and it should operate on the root mountpoint just like the all the other APIs).

If you want to then extend the JS API to include backend-specific APIs that is anther question, perhaps one that @kripken and @tlively could weigh in on.  Perhaps you could explain why there is a need to be able to manipulate backends that are not mounted anywhere in the filesytem?

JoeOsborn: Maybe the tension here is that the only way I can see for JS code to use WASMFS is through the WASMFS port of the JS FS API, and I'm trying to extend that.  If there were a "native" wasmfs API in JS, which exposed the same API as `emscripten/wasmfs.h` (with conveniences around strings for example) then JS code could use the WASMFS API in exactly the way C code can (wasmfs backends aren't necessarily mounted anywhere, they just get to create directories or create files on the backend at different paths on the filesystem).

I know there's a TODO on wasmfs_create_file saying that in the future only directories should be mounted, but that's inconsistent with the existing tests (test/wasmfs/wasmfs_jsfile.c, test/wasmfs/wasmfs_fetch.c) and it's convenient to be able to work with a backend that semantically is a single file.  For now, that means I could achieve what I had talked about earlier with that "mapfs" or manifest idea purely on the JS side, by creating and mounting single-file backends where I need them.

sbc100: Regarding mounting a single file:  The ability to "mount" a file is not something that maps to the existing concept of `mount` from linux/posix.  Perhaps the use case is very strong here and we can/should deviate to allow the idea of mounting just a file?  I would suggest that we avoid doing that if possible, but again I'll probably defer to @kripken / @tlively who were more involved in the original wasmfs design.

kripken: In general the WasmFS design does allow for a file to appear under another file with a different backend. For example the root can be the default memory backend, and a JSFile can be mounted underneath it (using `mountChild()`). Yes, that is more "general" than normal POSIX mounting.

Whether that ability is also allowed on the JS side could be a separate question - we've worried about code size issues, for example - but this PR looks quite reasonable to me. The backend param in createFile is the relevant information that is needed there. So this lgtm, but curious to hear from @tlively too.

tlively: > I know there's a TODO on wasmfs_create_file saying that in the future only directories should be mounted, but that's inconsistent with the existing tests (test/wasmfs/wasmfs_jsfile.c, test/wasmfs/wasmfs_fetch.c) and it's convenient to be able to work with a backend that semantically is a single file.

This TODO [was my idea](https://github.com/emscripten-core/emscripten/pull/19825#discussion_r1275472048), but I have to admit I don't remember why I thought being able to mount individual files was a problem.

I am surprised that we allow doing anything with backends that haven't been mounted anywhere, although I can't say whether this would have been unsurprising to me at some point in the past.

JoeOsborn: Yeah, compared to the JS FS API it is surprising that WASMFS backends aren’t technically mounted, they’re just used to create files and directories at various places (and to support those directories’ ability to create more files and directories).

tlively: Oh hmm, that triggers a memory of it being bad if a backend is used to create directories or files at several disjoint places in the file system. The OPFS backend definitely can’t handle that correctly. Maybe that’s why using a backend to mount individual files seems like a bad idea.

sbc100: Hm.. now that I've looked at this a little longer I think it might be a good idea to remove `wasmfs_create_file` completely and rename `wasmfs_create_directory` to `wasmfs_mount`.   

Then wasmfs would look much more like a normal unix FS.   We could even add `mount()` support with `MS_REMOUNT` to replace the root directory/mount point.

JoeOsborn: I don't know if there are any users of wasmfs_create_file besides me and the unit tests, but I would find it annoying to replace. I'd have to implement the MAPFS backend idea from https://github.com/emscripten-core/emscripten/pull/23021 and get it landed in emscripten.

sbc100: Can you say a little about how `wasmfs_create_file` is necessary in your case?  Is there some reason why mounting a directory won't work in your case?

JoeOsborn: Yeah. I want to use the fetch backend to fetch files from many distinct paths on a web server into one folder (with specific file names) in the filesystem.  I had proposed a manifest parameter for fetchfs to support this kind of thing:
```
https://server/path1/file1.txt => dir/fileA
https://server/path2/file2.txt => dir/fileB
```
If it’s hard to implement or slow to land a mapping backend, I guess I could change my server to use a common root url (these are DB records grouped as a result of a query so this is nontrivial), but currently I solve it by making a (small) number of single file fetchfs backends and use create_file for each.

sbc100: In the above example could we not have have `fetchfs` mounted at `dir` with two files withing it?

Is that idea that each file gets its own backend/mount point? 

If so it doesn't seems like maybe better design would be to allow a single fetch backend to have any number of files under a common root (`dir` in the example above).   Maybe that right-hand-side entries in the manifest could be relative to where the backend is mounted rather than relative to where the actual root FS.

JoeOsborn: fetchfs doesn’t include manifest support right now (I pulled it out in favor of a hypothetical future mapfs that maps one path to another path); it’s true that the manifest would actually be /fileA, /fileB.  

As it is, because fetchfs uses the path relative to the mount point as the URL relative to the baseURL, if I didn’t have create_file I would either need to be ok with a filesystem like dir/path1/file1.txt, dir/path2/file2.txt (using a baseURL of /), change my server (nontrivial due to the rationale above), or implement mapfs to map a path like fileA->/fetch/path1/file1 and fileB->/fetch/path2/file2 and mount my fetchfs backend at /fetch.

I do actually like the mapfs idea, but it’s a bigger change and I’m not sure when I’ll have bandwidth to do it.  Even if you don’t want to commit to including createFile in the JS api, leaving it in the C api for a few versions would give me time to try and build and land a PR for mapfs.

As a separate question, is it possible to write out-of-tree wasmfs backend implementations?  What would compiling them and linking them into a build look like since they’d potentially provide both C and JS symbols?

JoeOsborn: I got a little obsessed with this problem so I experimented and ended up with #23808, though I haven't fully explored it for my use case yet I think it should probably work.

JoeOsborn: It also occurs to me that wasmfs fetch backend as far back as I’m aware requires users to create individual files before reading them.  I don’t know how to (as in, I haven’t looked into) make the backend support creating files on read/stat (which feels like a better user experience anyway).

PR #23669, opened 2025-02-13
Fix an edge case in wasmfs_fetch for reads far past end of file

JoeOsborn:
Use the minimum of the requested last chunk and the actual last chunk when accessing chunked data.

### DIFF
--- a/src/lib/libwasmfs_fetch.js
+++ b/src/lib/libwasmfs_fetch.js
@@ -55,7 +55,7 @@ addToLibrary({
             chunks: [],
             chunkSize: chunkSize
           };
-          lastChunk = Math.min(lastChunk, ((size-1) / chunkSize));
+          lastChunk = Math.min(lastChunk, ((size-1) / chunkSize)) | 0;
         } else {
           // may as well/forced to download the whole file
           var wholeFileReq = await fetch(url);
--- a/src/lib/libwasmfs_fetch.js
+++ b/src/lib/libwasmfs_fetch.js
@@ -122,7 +122,10 @@ addToLibrary({
 
       // read/getSize fetch the data, then forward to the parent class.
       read: async (file, buffer, length, offset) => {
-        if (offset < 0 || length == 0) {
+        // This function assumes that offset is non-negative and length is positive.
+        // C read() doesn't take an offset and so doesn't have to deal with the former situation,
+        // and if the length is 0 or the offset is negative there's no reasonable read we can make.
+        if (offset < 0 || length <= 0) {
           return 0;
         }
         try {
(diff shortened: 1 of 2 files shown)


src/lib/libwasmfs_fetch.js
```diff
@@ -49,11 +49,14 @@ addToLibrary({
...
             fileInfo.headers.get('Accept-Ranges') == 'bytes' &&
             (parseInt(fileInfo.headers.get('Content-Length'), 10) > chunkSize*2)) {
+          var size = parseInt(fileInfo.headers.get('Content-Length'), 10);
           wasmFS$JSMemoryRanges[file] = {
-            size: parseInt(fileInfo.headers.get('Content-Length'), 10),
+            size,
             chunks: [],
             chunkSize: chunkSize
           };
+          firstChunk = 0;
```
kripken, line 58: Please add a comment here, as why firstChunk should be 0 is not obvious to me.

src/lib/libwasmfs_fetch.js
```diff
@@ -49,11 +49,14 @@ addToLibrary({
...
+          lastChunk = Math.min(lastChunk, ((size-1) / chunkSize));
```
kripken, line 59: Do we need ` | 0` as on line 45?

src/lib/libwasmfs_fetch.js
```diff
@@ -133,7 +136,9 @@ addToLibrary({
         var chunkSize = fileInfo.chunkSize;
         var firstChunk = (offset / chunkSize) | 0;
         // See comments in getFileRange.
-        var lastChunk = ((offset+length-1) / chunkSize) | 0;
+        var lastChunk = Math.min(
+          ((offset+length-1) / chunkSize),
+          ((fileInfo.size-1) / chunkSize)) | 0;
```
kripken, line 141: Can we truncate offset and/or length to handle this? That is, when the size is small, the effective length is lower.

src/lib/libwasmfs_fetch.js
JoeOsborn: You're right, that's an error.

src/lib/libwasmfs_fetch.js
JoeOsborn: That would also work.  I'll revise my PR, I like that better.

src/lib/libwasmfs_fetch.js
```diff
@@ -129,6 +131,10 @@ addToLibrary({
           return failedResponse.status === 404 ? -{{{ cDefs.ENOENT }}} : -{{{ cDefs.EBADF }}};
         }
         var fileInfo = wasmFS$JSMemoryRanges[file];
+        length = Math.min(length, fileInfo.size-offset+1) | 0;
+        if (offset < 0 || length <= 0) {
```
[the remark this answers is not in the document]
JoeOsborn: I put it in there because it's an assumption of the remainder of the code, and we can't return a reasonable read in such a case.  The C read() call uses a pointer that's already been offset so it isn't possible for the offset to be negative, so I wasn't sure what the right move was.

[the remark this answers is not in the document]
kripken: ok, sounds reasonable, please just add that in a comment then, so it is documented and the code is clearer.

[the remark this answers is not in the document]
JoeOsborn: Done.

kripken (approved):

PR #23678, opened 2025-02-14
Support ES6 Set/Map in stringifyWithFunctions

### LINKED ISSUE #23672: Using `Map` in library code seems not allowed
I try to use a ES6 `Map` inside library code like this:

```diff
 // library_webgpu.js
 
 var LibraryWebGPU = {
   $WebGPU__deps: [],
   $WebGPU: {
     Internals: {
+      buffers: new Map(),
       ...
     },
     ...
   },
 };
 
 autoAddDeps(LibraryWebGPU, '$WebGPU');
 mergeInto(LibraryManager.library, LibraryWebGPU);
 

```

And the final generated JS files is like this:

```
var WebGPU = {
  Internals: {
    buffers: {},

```

This will cause the following error later:
```
Uncaught TypeError: buffers.set is not a function
```

I tried to figure out whether this is an expected behavior, but didn't see document about this. And I don't understand why this behavior happens. If this is expected behavior (re-writing a Map to a plain object), is there a way to use ES6 `Map` in library code?

fs-eire:
Fixes #23672

### DIFF
--- a/src/jsifier.mjs
+++ b/src/jsifier.mjs
@@ -82,10 +82,7 @@ function stringifyWithFunctions(obj) {
   }
 
   // preserve the type of the object if it is one of [Map, Set, WeakMap, WeakSet].
-  const builtinContainers = vm.runInContext(
-    '[Map, Set, WeakMap, WeakSet]',
-    compileTimeContext,
-  );
+  const builtinContainers = vm.runInContext('[Map, Set, WeakMap, WeakSet]', compileTimeContext);
   for (const container of builtinContainers) {
     if (obj instanceof container) {
       const className = container.name;
--- a/src/jsifier.mjs
+++ b/src/jsifier.mjs
@@ -9,7 +9,6 @@
 
 import assert from 'node:assert';
 import * as fs from 'node:fs/promises';
-import * as vm from 'node:vm';
 import {
   ATEXITS,
   ATINITS,
@@ -34,6 +33,7 @@ import {
   compileTimeContext,
   printErr,
   readFile,
+  runInMacroContext,
   warn,
   warnOnce,
   warningOccured,
@@ -82,7 +82,9 @@ function stringifyWithFunctions(obj) {
   }
 
   // preserve the type of the object if it is one of [Map, Set, WeakMap, WeakSet].
-  const builtinContainers = vm.runInContext('[Map, Set, WeakMap, WeakSet]', compileTimeContext);
+  const builtinContainers = runInMacroContext('[Map, Set, WeakMap, WeakSet]', {
+    filename: '<internal>',
+  });
   for (const container of builtinContainers) {
     if (obj instanceof container) {
       const className = container.name;
--- a/src/jsifier.mjs
+++ b/src/jsifier.mjs
@@ -9,6 +9,7 @@
 
 import assert from 'node:assert';
 import * as fs from 'node:fs/promises';
+import * as vm from 'node:vm';
 import {
   ATEXITS,
   ATINITS,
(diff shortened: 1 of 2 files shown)


fs-eire:
@sbc100

sbc100:
I suggest adding a new test to `test_other.py` see all the existing tests that start with `def test_jslib_*` in that file.

For example:

```
$ ./test/runner other.test_jslib_native_deps
Test suites:
['test_other']
Running test_other: (1 tests)
test_jslib_native_deps (test_other.other.test_jslib_native_deps) ... ok

Ran 1 test in 0.714s

OK
```

fs-eire:
could you suggest a name for the test case so that it matches the style and naming convention best?

sbc100:
How about `test_jslib_new_objects`?

fs-eire:
I am working on adding a test and I found this strange behavior:

```py
  def test_jslib_new_objects_basic(self):
    create_file('lib.js', '''
      addToLibrary({
        $obj: {
          a: new Map(),
          b: new Set(),
          c: new WeakMap(),
          d: new WeakSet(),
        }
      });
      ''')
    self.run_process([EMCC, test_file('hello_world.c'), '--js-library=lib.js', '-sEXPORTED_FUNCTIONS=obj,_main'])
```

I have this line in `function stringifyWithFunctions(obj)`:
```js
  console.log(`##Debug ${inspect(obj)}, ${obj instanceof Set}, ${obj.size}`);
```

And I get the following outputs:
```
##Debug {
  a: Map(0) {},
  b: Set(0) {},
  c: WeakMap { <items unknown> },
  d: WeakSet { <items unknown> }
}, false, undefined
##Debug Map(0) {}, false, 0
##Debug Set(0) {}, false, 0
##Debug WeakMap { <items unknown> }, false, undefined
##Debug WeakSet { <items unknown> }, false, undefined
```

@sbc100 Did you have any tricks in jsifier.mjs? I cannot explain `b instanceof Set === false`.

sbc100:
That could be related to the fact that we evaluate the library code in a separate JS context?  See `compileTimeContext = vm.createContext` in `utility.mjs`

sbc100:
From reading https://nodejs.org/api/vm.html#what-does-it-mean-to-contextify-an-object it sounds like you may be able to add `vm.constants.DONT_CONTEXTIFY` to the `vm.createContext` call and then use `instanceof compileTimeContext.Set`?

src/jsifier.mjs
```diff
@@ -79,6 +80,17 @@ function stringifyWithFunctions(obj) {
   if (Array.isArray(obj)) {
     return '[' + obj.map(stringifyWithFunctions).join(',') + ']';
   }
+
+  // preserve the type of the object if it is one of [Map, Set, WeakMap, WeakSet].
+  const builtinContainers = vm.runInContext('[Map, Set, WeakMap, WeakSet]', compileTimeContext);
```
sbc100, line 85: I think we have a helper called runinmacrocontext or something like that, which would avoid the direct use of VM here

fs-eire: There is a function `runInMacroContext`, which requires a `filename` in options. I don't think that function works here.

do you want me to add the `runInContext` into utility.mjs like this:
```js
export function runInContext(code) {
  return vm.runInContext(code, compileTimeContext);
}
```

sbc100: I think you can just modify `runInMacroContext` such that the filename is optional, or (perhaps better) pass `filename: '<internal>'`.

sbc100: Also, can you look into doing this instead:

```
export const compileTimeContext = vm.createContext(vm.constants.DONT_CONTEXTIFY);
addToCompileTimeContext({process, console})
```

Then I think you can simply access `compileTimeContext.Set` .. without running any code.

sbc100 (approved):

sbc100:
Thanks for working on this!

fs-eire:
The following test failure:
```
file:///home/circleci/project/src/utility.mjs:295
export const compileTimeContext = vm.createContext(vm.constants.DONT_CONTEXTIFY);
                                                                ^

TypeError: Cannot read properties of undefined (reading 'DONT_CONTEXTIFY')
```

From API history of [`vm.runInNewContext`](https://nodejs.org/api/vm.html#vmruninnewcontextcode-contextobject-options)

```
v22.8.0, v20.18.0 | The contextObject argument now accepts vm.constants.DONT_CONTEXTIFY.
```

Does emscripten uses lower version of this?

fs-eire:
@sbc100 do you want to update minimum node version requirement or revert this change to use `vm.runInContext('[Map, Set, WeakMap, WeakSet]'`?

sbc100:
> @sbc100 do you want to update minimum node version requirement or revert this change to use `vm.runInContext('[Map, Set, WeakMap, WeakSet]'`?

The later please.  Sorry for the back a forth on this :(

src/jsifier.mjs
sbc100, line 85: Sorry to go back on this again, but can we use `runInMacroContext('...', {filename: '<internal>'})` and avoid importing `vm`?

sbc100 (approved):

Issue #23677, opened 2025-02-14
emcc doesn't pass -u arguments to linker


JoeOsborn:
**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.4-git (3f29dc6b86ca14624be09ca534d2bbc3dccaa57a)
clang version 21.0.0git (https:/github.com/llvm/llvm-project ff8f6abe20a9dfbb1de8d978f865092413d3416c)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /home/jcoa2018/.local/src/emsdk/upstream/bin
Build config: +assertions
```
**Failing command line in full:**
```
$ emcc -u main hello_world.c
emcc: error: main: No such file or directory ("main" was expected to be an input file, based on the commandline arguments provided)
```

Compared to `clang -u main hello.c`, which succeeds without error.

What should happen is that the "-u main" is passed along to the linker, where it hides/undefines a symbol.

sbc100:
Thanks for the bug report.  Should be a simple fix.

BTW, `-u` doesn't "hide" anything (at least not that I know of) it just forces the symbol to be undefined in the linker so that (for example) that symbol will be extracted from the a library file.

```
       -u symbol
           Pretend  the  symbol  symbol  is undefined, to force linking of library modules to
           define it.  You can use -u multiple times with different symbols to force  loading
           of additional library modules.
```

PR #23679, opened 2025-02-14
Allow `-u` flag to pass through to the linker

sbc100:
`-u` is one of the few linker flags that is also a compiler flag, and the compiler is supposed to pass it through.  This is similar to `-z`, for example.

Fixes: #23677

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -12375,6 +12375,29 @@ Aborted(`Module.arguments` has been replaced by `arguments_` (the initial value 
     err = self.run_process([EMCC, test_file('hello_world.c'), '-zfoo'], stderr=PIPE).stderr
     self.assertContained('wasm-ld: warning: unknown -z value: foo', err)
 
+  def test_linker_flags_pass_through_u(self):
+    # Create a static library that contains a reference to an undefined symbol: bar
+    create_file('test.c', '''
+    extern int bar();
+    int foo() {
+      return bar();
+    }
+    ''')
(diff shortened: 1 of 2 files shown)


test/test_other.py
```diff
@@ -12375,6 +12375,29 @@ def test_linker_flags_pass_through(self):
     err = self.run_process([EMCC, test_file('hello_world.c'), '-zfoo'], stderr=PIPE).stderr
     self.assertContained('wasm-ld: warning: unknown -z value: foo', err)
 
+  def test_linker_flags_pass_through_u(self):
+    # Create a static library that contains a reference undefined symbol: bar
```
kripken, line 12379:
```suggestion
    # Create a static library that contains a referenced but undefined symbol: bar
```

test/test_other.py
```diff
@@ -12375,6 +12375,29 @@ def test_linker_flags_pass_through(self):
...
+      return bar();
+    }
+    ''')
+    self.run_process([EMCC, '-c', 'test.c'])
+    self.run_process([EMAR, 'crs', 'libtest.a', 'test.o'])
+    # This test uses --no-gc-sections otherwise the foo symbol never actually gets included
+    # in the link.
+    cmd = [EMCC, test_file('hello_world.c'), '-Wl,--no-gc-sections']
+
+    # Verify that the library is normally ignored that the link succeeds
```
kripken, line 12392:
```suggestion
    # Verify that the library is normally ignored, and the link succeeds
```

kripken (approved):

PR #23681, opened 2025-02-14
[test] Ensure setUpClass is called in the parallel test runner. NFC

sbc100:


kripken (approved):
To make sure I follow, looks like this was not noticed before as we only use setUpClass in non-parallel test suites so far?

sbc100:
> To make sure I follow, looks like this was not noticed before as we only use setUpClass in non-parallel test suites so far?

Yes, thats about right.  I remember being confused in the past about why setUpClass wasn't called in those cases.   I runs out the `run` method is supposed to do that, but we were not doing it.

PR #23682, opened 2025-02-15
New contrib port: lua

ypujante:
I don't know if there is any interest, but I wrote a port of lua, which allows to embed lua directly in C/C++ in Emscripten without having to deal with how to fetch/compile (which is the point of Emscripten ports...).

It's ok if you do not want to add it to Emscripten, since with external port, I can push it instead to my other [project](https://github.com/pongasoft/emscripten-ports).

A "hello world" program that would use this is

```c
#include <lua.h>
#include <lualib.h>
#include <lauxlib.h>
#include <stdio.h>

int main()
{
  // Create a new Lua state
  lua_State *L = luaL_newstate();

  // Load standard Lua libraries (for print)
  luaL_openlibs(L);

  // Execute Lua code
  if(luaL_dostring(L, "print('hello world')") != LUA_OK)
  {
    printf("Error running Lua code %s\n", lua_tostring(L, -1));
    lua_pop(L, 1); // Remove the error message from the stack
  }

  // Close the Lua state
  lua_close(L);

  return 0;
}
```

and is compiled like this:

```sh
emcc --use-port=contrib.lua main.c -o /tmp/build/index.html
```

I did not change any documentation or Changelog, since I did not want to do extra work if this PR does not go any further. But I will of course update this PR with these changes, if it there is interest.

Let me know. I am fine either way.

### DIFF
--- a/test/other/test_port_contrib_lua.c
+++ b/test/other/test_port_contrib_lua.c
@@ -1,17 +1,16 @@
 /*
-* Copyright 2025 The Emscripten Authors.  All rights reserved.
-* Emscripten is available under two separate licenses, the MIT license and the
-* University of Illinois/NCSA Open Source License.  Both these licenses can be
-* found in the LICENSE file.
-*/
+ * Copyright 2025 The Emscripten Authors.  All rights reserved.
+ * Emscripten is available under two separate licenses, the MIT license and the
+ * University of Illinois/NCSA Open Source License.  Both these licenses can be
+ * found in the LICENSE file.
+ */
 
 #include <lua.h>
 #include <lualib.h>
 #include <lauxlib.h>
 #include <stdio.h>
 
-int main()
-{
+int main() {
   // Create a new Lua state
   lua_State *L = luaL_newstate();
 
--- a/test/other/test_port_contrib_lua.c
+++ b/test/other/test_port_contrib_lua.c
@@ -1,0 +1,41 @@
+/*
+* Copyright 2025 The Emscripten Authors.  All rights reserved.
+* Emscripten is available under two separate licenses, the MIT license and the
+* University of Illinois/NCSA Open Source License.  Both these licenses can be
+* found in the LICENSE file.
+*/
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+#include <stdio.h>
+
+int main()
+{
+  // Create a new Lua state
+  lua_State *L = luaL_newstate();
+
+  // Load standard Lua libraries
+  luaL_openlibs(L);
+  
+  char const *chunks[3] = {
+    "print('hello world')",                                // test Basic Library
+    "print(string.upper('hello world'))",                  // test string library
+    "print('sqrt(16)=' .. math.tointeger(math.sqrt(16)))", // test math library
+  };
+
+  // Iterate over each chunk of Lua code and execute it
+  for (size_t i = 0; i <= 3; i++) {
+    if (luaL_dostring(L, chunks[i]) != LUA_OK) {
+      printf("Error running Lua code: %s\n", lua_tostring(L, -1));
+      lua_pop(L, 1); // Remove the error message from the stack
+      lua_close(L);
+      return 1; // returning error so the test can fail
+    }
+  }
+
+  // Close the Lua state
+  lua_close(L);
+
+  return 0;
+}
--- a/tools/ports/contrib/lua.py
+++ b/tools/ports/contrib/lua.py
@@ -15,10 +15,7 @@ DESCRIPTION = 'Lua is a powerful, efficient, lightweight, embeddable scripting l
 LICENSE = 'MIT License'
 
 port_name = 'contrib.lua'
-
-
-def get_lib_name(settings):
-  return f'lib_{port_name}.a'
+lib_name = 'liblua.a'
 
 
 def get(ports, settings, shared):
@@ -44,8 +41,8 @@ def get(ports, settings, shared):
 
     ports.build_port(source_path, final, port_name, srcs=srcs, flags=flags)
 
-  return [shared.cache.get_lib(get_lib_name(settings), create, what='port')]
+  return [shared.cache.get_lib(lib_name, create, what='port')]
 
 
 def clear(ports, settings, shared):
-  shared.cache.erase_lib(get_lib_name(settings))
+  shared.cache.erase_lib(lib_name)
--- a/tools/ports/contrib/lua.py
+++ b/tools/ports/contrib/lua.py
@@ -7,7 +7,6 @@ import os
 
 TAG = '5.4.7'
 HASH = '98c5c8978dfdf867e37e9eb3b3ec83dee92d199243b5119505da83895e33f10d43c841be6a7d3b106daba8a0b2bd25fe099ebff8f87831dcc55c79c78b97d8b8'
-ZIP_URL = f'https://www.lua.org/ftp/lua-{TAG}.tar.gz'
 
 # contrib port information (required)
 URL = 'https://www.lua.org/'
@@ -20,7 +19,7 @@ lib_name = 'liblua.a'
 
 def get(ports, settings, shared):
   # get the port
-  ports.fetch_project(port_name, ZIP_URL, sha512hash=HASH)
+  ports.fetch_project(port_name, f'https://www.lua.org/ftp/lua-{TAG}.tar.gz', sha512hash=HASH)
 
   def create(final):
     root_path = os.path.join(ports.get_dir(), port_name, f'lua-{TAG}')
--- a/tools/ports/contrib/lua.py
+++ b/tools/ports/contrib/lua.py
@@ -1,0 +1,55 @@
+# Copyright 2025 The Emscripten Authors.  All rights reserved.
+# Emscripten is available under two separate licenses, the MIT license and the
+# University of Illinois/NCSA Open Source License.  Both these licenses can be
+# found in the LICENSE file.
+
+import os
+
+TAG = '5.4.7'
+HASH = '98c5c8978dfdf867e37e9eb3b3ec83dee92d199243b5119505da83895e33f10d43c841be6a7d3b106daba8a0b2bd25fe099ebff8f87831dcc55c79c78b97d8b8'
+ZIP_URL = f'https://www.lua.org/ftp/lua-{TAG}.tar.gz'
+
+# contrib port information (required)
+URL = 'https://www.lua.org/'
+DESCRIPTION = f'Lua is a powerful, efficient, lightweight, embeddable scripting language'
+LICENSE = 'MIT License'
+
+port_name = 'contrib.lua'
+
+
+def get_lib_name(settings):
+  return f'lib_{port_name}.a'
+
+
+def get(ports, settings, shared):
+  # get the port
+  ports.fetch_project(port_name, ZIP_URL, sha512hash=HASH)
+
+  def create(final):
+    root_path = os.path.join(ports.get_dir(), port_name, f'lua-{TAG}')
+    source_path = os.path.join(root_path, 'src')
+
+    # install headers
+    includes = ['lua.h', 'lua.hpp', 'luaconf.h', 'lauxlib.h', 'lualib.h']
+    for f in includes:
+      ports.install_headers(source_path, pattern=f, target='lua')
+
+    srcs = '''
+       lapi.c lcode.c lctype.c ldebug.c ldo.c ldump.c lfunc.c lgc.c llex.c lmem.c lobject.c lopcodes.c 
+       lparser.c lstate.c lstring.c ltable.c ltm.c lundump.c lvm.c lzio.c lauxlib.c lbaselib.c lcorolib.c
+       ldblib.c liolib.c lmathlib.c loadlib.c loslib.c lstrlib.c ltablib.c lutf8lib.c linit.c
+       '''.split()
+
+    flags = ['-DLUA_COMPAT_5_3']
+
+    ports.build_port(source_path, final, port_name, srcs=srcs, flags=flags)
+
+  return [shared.cache.get_lib(get_lib_name(settings), create, what='port')]
+
+
+def clear(ports, settings, shared):
+  shared.cache.erase_lib(get_lib_name(settings))
+
+
+def process_args(ports):
+  return ['-isystem', ports.get_include_dir('lua')]
--- a/tools/ports/contrib/lua.py
+++ b/tools/ports/contrib/lua.py
@@ -11,7 +11,7 @@ ZIP_URL = f'https://www.lua.org/ftp/lua-{TAG}.tar.gz'
 
 # contrib port information (required)
 URL = 'https://www.lua.org/'
-DESCRIPTION = f'Lua is a powerful, efficient, lightweight, embeddable scripting language'
+DESCRIPTION = 'Lua is a powerful, efficient, lightweight, embeddable scripting language'
 LICENSE = 'MIT License'
 
 port_name = 'contrib.lua'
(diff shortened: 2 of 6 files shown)


sbc100:
Adding it as a contrib port seems reasonable.  

Can you add a simple test to test_other?   I think we have some existing tests of lua there maybe.. perhaps they can be converted to use this port instead of whatever they use today?

ypujante:
Sounds great. I will take a look at the tests, fix the ruff failing test and add documentation.

ypujante:
I added:

* a test for the contrib port itself
* documentation (in Contrib Ports)
* changelog entry

I did take a look at the Lua usage in tests that you are referring to. I am not too sure how to proceed. I am not too familiar with this code and what it is testing. It seems that it compiles Lua natively and through Emscripten then runs the same lua code to measure performance differences? Maybe?

If that is the case, I am not too sure how this test can be ported. The (contrib) port downloads the Lua code and generates the Lua library. It does not generate an executable. That being said, the source code of the executable (`lua.c`) is part of what is downloaded. So maybe these tests could be rewritten to somehow use the port so that it gets downloaded, then somehow accesses the cache of where the port was downloaded since it contains the Lua source code?

For your reference this is the whole section that uses Lua (and of course the Lua code is checked in under `test/third_party/lua` and could be removed if we could use the port...)

```python
  def lua(self, benchmark, expected, output_parser=None, args_processor=None):
    self.emcc_args.remove('-Werror')
    shutil.copyfile(test_file(f'third_party/lua/{benchmark}.lua'), benchmark + '.lua')

    def lib_builder(name, native, env_init):
      # We force recomputation for the native benchmarker because this benchmark
      # uses native_exec=True, so we need to copy the native executable
      return self.get_library(os.path.join('third_party', 'lua_native' if native else 'lua'), [os.path.join('src', 'lua.o'), os.path.join('src', 'liblua.a')], make=['make', 'generic'], configure=None, native=native, cache_name_extra=name, env_init=env_init, force_rebuild=native)

    self.do_benchmark('lua_' + benchmark, '', expected,
                      force_c=True, args=[benchmark + '.lua', DEFAULT_ARG],
                      emcc_args=['--embed-file', benchmark + '.lua', '-sFORCE_FILESYSTEM', '-sMINIMAL_RUNTIME=0'], # not minimal because of files
                      lib_builder=lib_builder, native_exec=os.path.join('building', 'third_party', 'lua_native', 'src', 'lua'),
                      output_parser=output_parser, args_processor=args_processor)

  def test_zzz_lua_scimark(self):
    def output_parser(output):
      return 100.0 / float(re.search(r'\nSciMark +([\d\.]+) ', output).group(1))

    self.lua('scimark', '[small problem sizes]', output_parser=output_parser)

  def test_zzz_lua_binarytrees(self):
    # js version: ['binarytrees.lua', {0: 0, 1: 9.5, 2: 11.99, 3: 12.85, 4: 14.72, 5: 15.82}[arguments[0]]]
    self.lua('binarytrees', 'long lived tree of depth')
```

In any case I feel like this is a bigger effort and should probably require its own PR?

ypujante:
I checked that the new test works fine (`test_port_contrib_lua`):

```text
test_port_contrib_lua (test_other.other) ... ok (1.05s)
```

The sole failing test is:

```text
ERROR [0.003s]: test_emmalloc_memvalidate_verbose (test_core.wasm64)
```

tools/ports/contrib/lua.py
```diff
@@ -0,0 +1,55 @@
...
+TAG = '5.4.7'
+HASH = '98c5c8978dfdf867e37e9eb3b3ec83dee92d199243b5119505da83895e33f10d43c841be6a7d3b106daba8a0b2bd25fe099ebff8f87831dcc55c79c78b97d8b8'
+ZIP_URL = f'https://www.lua.org/ftp/lua-{TAG}.tar.gz'
+
+# contrib port information (required)
+URL = 'https://www.lua.org/'
+DESCRIPTION = 'Lua is a powerful, efficient, lightweight, embeddable scripting language'
+LICENSE = 'MIT License'
+
+port_name = 'contrib.lua'
```
sbc100, line 17: I wonder if we need the `contrib.` prefix here, since unlike glfw we don't need to distinguish it from an internal port.

Did we decide we wanted all contib ports to have this prefix?   Either way it seems like this name can/should be based on soley on the filename.   It shouldn't need to explicit like this.

test/other/test_port_contrib_lua.c
```diff
@@ -0,0 +1,41 @@
...
+* found in the LICENSE file.
+*/
+
+#include <lua.h>
+#include <lualib.h>
+#include <lauxlib.h>
+#include <stdio.h>
+
+int main()
+{
```
sbc100, line 14: Can you put the opening brace on the same line as the function name?

test/other/test_port_contrib_lua.c
```diff
@@ -0,0 +1,41 @@
+/*
+* Copyright 2025 The Emscripten Authors.  All rights reserved.
+* Emscripten is available under two separate licenses, the MIT license and the
+* University of Illinois/NCSA Open Source License.  Both these licenses can be
+* found in the LICENSE file.
+*/
```
sbc100, line 6: Lines 2-6 should be indented one space.

sbc100 (commented):
Regarding the existing lua benchmark, I agree we should probably leave that in place.

However, I do wonder if we should delete `test_lua` from `test_core.py`.   It seems odd to have two different lua tests, one in `other` and one in `core` that test two different versions of lua.  However the existing test in test_core.py does test the `lua` main interpreter, whereas this new test tests lua-as-a-library so maybe its fine?

tools/ports/contrib/lua.py
ypujante: You cannot change the name otherwise it fails because the port is registered as `contrib.lua`:

```text
  File "/usr/local/emsdk/emscripten/main/emcc.py", line 1593, in <module>
    sys.exit(main(sys.argv))
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/usr/local/emsdk/emscripten/main/emcc.py", line 1586, in main
    ret = run(args)
  File "/usr/local/emsdk/emscripten/main/emcc.py", line 668, in run
    linker_args = phase_compile_inputs(options, state, newargs)
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/usr/local/emsdk/emscripten/main/emcc.py", line 987, in phase_compile_inputs
    arg.value = compile_source_file(input_file)
  File "/usr/local/emsdk/emscripten/main/emcc.py", line 961, in compile_source_file
    cmd = get_clang_command()
  File "/usr/local/emsdk/emscripten/main/emcc.py", line 920, in get_clang_command
    return compiler + get_cflags(state.orig_args)
  File "/usr/local/emsdk/emscripten/main/tools/utils.py", line 123, in helper
    results[key] = func(*args)
  File "/usr/local/emsdk/emscripten/main/emcc.py", line 447, in get_cflags
    ports.add_cflags(cflags, settings)
  File "/usr/local/emsdk/emscripten/main/tools/ports/__init__.py", line 577, in add_cflags
    port.get(Ports, settings, shared)
  File "/usr/local/emsdk/emscripten/main/tools/ports/contrib/lua.py", line 26, in get
    ports.fetch_project(port_name, ZIP_URL, sha512hash=HASH)
  File "/usr/local/emsdk/emscripten/main/tools/ports/__init__.py", line 352, in fetch_project
    Ports.clear_project_build(name)
  File "/usr/local/emsdk/emscripten/main/tools/ports/__init__.py", line 356, in clear_project_build
    port = ports_by_name[name]
KeyError: 'lua'
```

ypujante:
#### 1.`test_lua` removed

I did remove `test_lua` from `test_core.py` as I agree with you there is kind of no point with this test anymore. Testing the `lua` executable adds no value to Emscripten since the `lua` executable is not part of the final distribution. The contrib port test is enough and tests what actually ships (now the lua (contrib) port generating the lua lib).

#### 2. name of port unchanged

As I commented inline,  I cannot change the name of the port because it is used as the key in the map of ports and because it is in the contrib folder, the name of the port is `contrib.lua`. I included the stack trace that shows the error

#### 3. lua includes

I ran the following command in the lua project

```
> make -n generic install
...
cd src && install -p -m 0644 lua.h luaconf.h lualib.h lauxlib.h lua.hpp /usr/local/include
...
``` 

So it is true that when installed locally the Lua headers are available with no subfolder (`#include <lua.h>`).

In the port, I am installing the headers like this:

```python
      ports.install_headers(source_path, pattern=f, target='lua')
```

and then making them available like that:

```python
def process_args(ports):
  return ['-isystem', ports.get_include_dir('lua')]
```

which has the exact same effect: you still write `#include <lua.h>`. Although it does have the advantage (or disadvantage?) to let you write as well `#include <lua/lua.h>` since the root is also available...

We could install the headers directly at the root if you prefer. I wasn't sure if this is something that was okay. Just let me know what you prefer.

ypujante:
Failing tests (unrelated I presume):

```text
# test-sockets-chrome
ERROR [12.187s]: test_sockets_echo_bigdata_websockify (test_sockets.sockets)

# test-browser-chrome-2gb
ERROR [61.512s]: test_offset_converter_pthread (test_browser.browser_2gb)
```

Lua contrib port test passing fine:

```text
# test-other
test_port_contrib_lua (test_other.other) ... ok (0.99s)
```

tools/ports/contrib/lua.py
sbc100: > You cannot change the name otherwise it fails because the port is registered as `contrib.lua`:

Does this mean `port_name` cannot be set to anything except one fixes value?   Why do we even have `port_name` in that case?

ypujante: Because I don't like to repeat myself multiple time in the rest of the code. it is error prone

sbc100: Oh I see, the port_name global is only used locally within this file.   That seems reasonable.  I thought maybe this was configuring the port name.

ypujante: The port name is "injected" by the [port system](https://github.com/emscripten-core/emscripten/blob/a3b9e324224e19c243d89633fc420dc6ca9fcc92/tools/ports/__init__.py#L43) in the `name` variable. So technically we could remove `port_name` entirely and use `name` instead which would serve the same purpose (of not having to repeat it multiple times).

The only catch is that when looking at the code it is not obvious what `name` is since it is "injected" by the outside code. And also editors flag it as an error (since they don't know that it is injected).

tools/ports/contrib/lua.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def get_lib_name(settings):
+  return f'lib_{port_name}.a'
```
sbc100, line 21: Can we call the library just `liblua.a`.   You don't need `get_lib_name` then I think.

ypujante: done

ypujante:
@sbc100 I did change the lib name to be static as you requested. That being said, in other ports, the lib name is dynamic usually because of pthreads...

Ex: (sdl2)

```python
def get_lib_name(settings):
  return 'libSDL2' + ('-mt' if settings.PTHREADS else '') + '.a'
// ...
    if settings.PTHREADS:
      flags += ['-pthread']
    ports.build_port(src_dir, final, 'sdl2', srcs=srcs, includes=includes, flags=flags)
```

I am not sure what the implications are to add the `-pthread` flag when the port itself is being build. Is this something that all ports should be doing? Do we need to do this here as well for Lua? How would this flag affect (or not) the Lua library?

sbc100:
I would not worry about adding a `-mt` variant at this point.  It can always be added later if needed.

tools/ports/contrib/lua.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2025 The Emscripten Authors.  All rights reserved.
+# Emscripten is available under two separate licenses, the MIT license and the
+# University of Illinois/NCSA Open Source License.  Both these licenses can be
+# found in the LICENSE file.
+
+import os
+
+TAG = '5.4.7'
+HASH = '98c5c8978dfdf867e37e9eb3b3ec83dee92d199243b5119505da83895e33f10d43c841be6a7d3b106daba8a0b2bd25fe099ebff8f87831dcc55c79c78b97d8b8'
+ZIP_URL = f'https://www.lua.org/ftp/lua-{TAG}.tar.gz'
```
sbc100, line 10: I guess this should be renamed.   Actually it looks like most ports jut inline this into the `fetch_project` call.

sbc100:
```
tools/ports/bullet.py:  ports.fetch_project('bullet', f'https://github.com/emscripten-ports/bullet/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/bzip2.py:  ports.fetch_project('bzip2', f'https://github.com/emscripten-ports/bzip2/archive/{VERSION}.zip', sha512hash=HASH)
tools/ports/cocos2d.py:  ports.fetch_project('cocos2d', f'https://github.com/emscripten-ports/Cocos2d/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/contrib/glfw3.py:  ports.fetch_project(port_name,
tools/ports/freetype.py:  ports.fetch_project('freetype', f'https://github.com/freetype/freetype/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/giflib.py:  ports.fetch_project('giflib', f'https://storage.googleapis.com/webassembly/emscripten-ports/giflib-{VERSION}.tar.gz', sha512hash=HASH)
tools/ports/harfbuzz.py:  ports.fetch_project('harfbuzz', f'https://github.com/harfbuzz/harfbuzz/releases/download/{VERSION}/harfbuzz-{VERSION}.tar.xz', sha512hash=HASH)
tools/ports/icu.py:  ports.fetch_project('icu', f'https://github.com/unicode-org/icu/releases/download/{TAG}/icu4c-{VERSION}-src.zip', sha512hash=HASH)
tools/ports/libjpeg.py:  ports.fetch_project('libjpeg', f'https://storage.googleapis.com/webassembly/emscripten-ports/jpegsrc.v{VERSION}.tar.gz', sha512hash=HASH)
tools/ports/libmodplug.py:  ports.fetch_project('libmodplug', f'https://github.com/jancc/libmodplug/archive/v{TAG}.zip', sha512hash=HASH)
tools/ports/libpng.py:  ports.fetch_project('libpng', f'https://storage.googleapis.com/webassembly/emscripten-ports/libpng-{TAG}.tar.gz', sha512hash=HASH)
tools/ports/mpg123.py:  ports.fetch_project('mpg123', f'https://www.mpg123.de/download/mpg123-{TAG}.tar.bz2', sha512hash=HASH)
tools/ports/ogg.py:  ports.fetch_project('ogg', f'https://github.com/emscripten-ports/ogg/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/regal.py:  ports.fetch_project('regal', f'https://github.com/emscripten-ports/regal/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/sdl2.py:  ports.fetch_project('sdl2', f'https://github.com/libsdl-org/SDL/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/sdl2_gfx.py:  ports.fetch_project('sdl2_gfx', f'https://github.com/svn2github/sdl2_gfx/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/sdl2_image.py:  ports.fetch_project('sdl2_image', f'https://github.com/libsdl-org/SDL_image/archive/refs/tags/{TAG}.zip', sha512hash=HASH)
tools/ports/sdl2_mixer.py:  ports.fetch_project('sdl2_mixer', f'https://github.com/libsdl-org/SDL_mixer/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/sdl2_net.py:  ports.fetch_project('sdl2_net', f'https://github.com/emscripten-ports/SDL2_net/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/sdl2_ttf.py:  ports.fetch_project('sdl2_ttf', f'https://github.com/libsdl-org/SDL_ttf/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/sdl3.py:  ports.fetch_project('sdl3', f'https://github.com/libsdl-org/SDL/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/sqlite3.py:  ports.fetch_project('sqlite3', f'https://www.sqlite.org/{VERSION_YEAR}/{release}.zip', sha512hash=HASH)
tools/ports/vorbis.py:  ports.fetch_project('vorbis', f'https://github.com/emscripten-ports/vorbis/archive/{TAG}.zip', sha512hash=HASH)
tools/ports/zlib.py:  ports.fetch_project('zlib', f'https://github.com/madler/zlib/archive/refs/tags/v{VERSION}.tar.gz', sha512hash=HASH)
```

ypujante: ok

sbc100 (approved):

Issue #23687, opened 2025-02-16
Error: 'await' is only allowed within async functions and at the top levels of modules.


jrouwe:
Hello,

Between version 3.1.74 and 4.0.0 the debug build of [JoltPhysics.js](https://github.com/jrouwe/JoltPhysics.js/tree/0.32.0) broke with error:

```
SyntaxError: /tmp/emscripten_temp_562xmtnb/jolt-physics.js: 'await' is only allowed within async functions and at the top levels of modules. (1002:18)

  1000 |
  1001 | };
> 1002 | var wasmExports = await createWasm();
       |                   ^
```

I think it's because of this change: https://github.com/emscripten-core/emscripten/pull/23157

I've created a minimal repro here (just run `./build.sh` to trigger the error):

[Repro.zip](https://github.com/user-attachments/files/18816296/Repro.zip)

In JoltPhysics.js you would reproduce it by running `./build.sh Debug`.

I can reproduce almost the same error by downloading and building [Ammo.js](https://github.com/kripken/ammo.js/).

It seems to be the combination of `-s MODULARIZE=1` and `-s LEGACY_VM_SUPPORT=1` that causes the issue.

**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.3 (a9651ff57165f5710bb09a5fe52590fd6ddb72df)
```

**Failing command line in full:**
```
emcc glue.o -s MODULARIZE=1 -s LEGACY_VM_SUPPORT=1 -o /mnt/c/Users/jrouw/Documents/Code/JoltPhysics.js/jolt-physics.js
SyntaxError: /tmp/emscripten_temp_562xmtnb/jolt-physics.js: 'await' is only allowed within async functions and at the top levels of modules. (1002:18)

  1000 |
  1001 | };
> 1002 | var wasmExports = await createWasm();
       |                   ^
  1003 | var ___wasm_call_ctors = createExportWrapper('__wasm_call_ctors', 0);
  1004 | var _fflush = createExportWrapper('fflush', 1);
  1005 | var _emscripten_stack_init = wasmExports['emscripten_stack_init']
    at constructor (/home/jorrit/emsdk/upstream/emscripten/node_modules/@babel/parser/lib/index.js:360:19)
    at Parser.raise (/home/jorrit/emsdk/upstream/emscripten/node_modules/@babel/parser/lib/index.js:3327:19)
[...]

PR #23693, opened 2025-02-18
Fix -sMODULARIZE + -sLEGACY_VM_SUPPORT

sbc100:
When we run babel in `-sMODULARIZE` mode the input to babel contains `await` calls.  This code appears to be top-level code at the point at which we run babel since we have yet to perform the modularization process which wraps the output in a factory function.

Fixes: #23687


kripken (approved):

sbc100:
Closing in favor of #23704

Issue #23692, opened 2025-02-18
Sqlite3 support with WASMFS OPFS


irfan2502:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.61-git (67fa4c16496b157a7fc3377afd69ee0445e8a6e3)
clang version 19.0.0git (https:/github.com/llvm/llvm-project 7cfffe74eeb68fbb3fb9706ac7071f8caeeb6520)

**Failing command line in full:**
em++ ex1.cpp -o ex1.html -sUSE_SQLITE3 -sWASMFS -g -pthread -sPROXY_TO_PTHREAD

```
#include <iostream>
#include <sqlite3.h>
#include <emscripten/wasmfs.h>

void checkSQLiteError(int rc, sqlite3* db, const char* msg) {
    if (rc != SQLITE_OK) {
        std::cerr << "Error: " << msg << " - " << sqlite3_errmsg(db) << std::endl;
        sqlite3_close(db);
        exit(1);
    }
}

int main() {
    backend_t opfs = wasmfs_create_opfs_backend();
    auto err = wasmfs_create_directory("/opfs", 0777, opfs);
    sqlite3* db;
    char* errMsg = nullptr;

    // Open (or create) the database
    int rc = sqlite3_open("/opfs/example.db", &db);
    checkSQLiteError(rc, db, "Cannot open database");

    // Set PRAGMA settings for truncate journal mode and full synchronous mode
    rc = sqlite3_exec(db, "PRAGMA journal_mode = TRUNCATE;", nullptr, nullptr, &errMsg);
    checkSQLiteError(rc, db, "Failed to set journal mode");

    return 0;
}
```

When I try to compile the above code with the options mentioned above and try opening in browser, it failed saying "Failed to set journal mode". On checking further, it was failing in fcntl check [here](https://github.com/emscripten-core/emscripten/blame/main/system/lib/wasmfs/syscalls.cpp#L1525) Does this mean sqlite is not supported with WASMFS? Or Am I missing something? Is there any plan to support F_SETLK and F_SETLKW in future? Thanks in advance.

sbc100:
For the old FS we currently ignore F_SETLK/F_SETLKW:

https://github.com/emscripten-core/emscripten/blob/a3b9e324224e19c243d89633fc420dc6ca9fcc92/src/lib/libsyscall.js#L795-L797

[...]

PR #23694, opened 2025-02-18
[WasmFS] Ignore fcntl F_SETLK (inter-process file locking)

sbc100:
This is the current behaviour of the legacy FS and the flock function so it seems reasonable to do the same thing, at least for now.  See #23697

This fixes sqlite which makes use of this fcntl to lock the database when accessing it from multiple different processes.

Fixes: #23692

### DIFF
--- a/system/lib/wasmfs/syscalls.cpp
+++ b/system/lib/wasmfs/syscalls.cpp
@@ -1521,8 +1521,11 @@ int __syscall_fcntl64(int fd, int cmd, ...) {
     case F_SETLKW: {
       static_assert(F_SETLK == F_SETLK64);
       static_assert(F_SETLKW == F_SETLKW64);
-      // Always error for now, until we implement byte-range locks.
-      return -EACCES;
+      // Pretend that the locking is successful. These are process-level locks,
+      // and Emscripten programs are a single process. If we supported linking a
+      // filesystem between programs, we'd need to do more here.
+      // See https://github.com/emscripten-core/emscripten/issues/23697
+      return 0;
     }
     default: {
       // TODO: support any remaining cmds
--- a/test/sqlite/test.out
+++ b/test/sqlite/test.out
@@ -1,7 +1,5 @@
 count(*) = 2
 
-datetime('2012-04-16 12:35:57', '+1 days') = 2012-04-17 12:35:57
-
 a = 1
 b = 13153
 c = thirteen thousand one hundred fifty three
--- a/src/lib/libsyscall.js
+++ b/src/lib/libsyscall.js
@@ -794,7 +794,11 @@ var SyscallsLibrary = {
       }
       case {{{ cDefs.F_SETLK }}}:
       case {{{ cDefs.F_SETLKW }}}:
-        return 0; // Pretend that the locking is successful.
+        // Pretend that the locking is successful. These are process-level locks,
+        // and Emscripten programs are a single process. If we supported linking a
+        // filesystem between programs, we'd need to do more here.
+        // See https://github.com/emscripten-core/emscripten/issues/23697
+        return 0;
 #if SYSCALL_DEBUG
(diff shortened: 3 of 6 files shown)


system/lib/wasmfs/syscalls.cpp
```diff
@@ -1521,8 +1521,7 @@ int __syscall_fcntl64(int fd, int cmd, ...) {
     case F_SETLKW: {
       static_assert(F_SETLK == F_SETLK64);
       static_assert(F_SETLKW == F_SETLKW64);
-      // Always error for now, until we implement byte-range locks.
-      return -EACCES;
+      return 0; // Pretend that the locking is successful.
```
kripken, line 1524: This is safe in the old JS FS because no reads can happen in parallel, but for WasmFS there is a real danger of data races.

If we want something in the meanwhile before we implement full byte-range locks, perhaps we could add a whole-file lock (the entire range)?

sbc100: > This is safe in the old JS FS because no reads can happen in parallel, but for WasmFS there is a real danger of data races.

Can you elaborate?  What kind of races are you talking about?   Is this something that is specific to the backend?   At least for the memoryfs (i.e. the default) this should be fine.

sbc100: The locking we are talking about here is not inter-thread locking, but inter-process locking.

IIUC, these locks do not effect threads at all.  And thread with access the the file descriptor effectively holds any locks since the locks at the FD level.   They are also only "advisory" locks. 

See the man page for flock and fcntl.  e.g.:

```
   Advisory record locking
       Linux  implements  traditional  ("process-associated")  UNIX record locks, as standardized by POSIX.  For a Linux-specific alternative
       with better semantics, see the discussion of open file description locks below.

       F_SETLK, F_SETLKW, and F_GETLK are used to acquire, release, and test for the existence of record locks  (also  known  as  byte-range,
       file-segment, or file-region locks).  The third argument, lock, is a pointer to a structure that has at least the following fields (in
       unspecified order).
```

sbc100: The idea is the if you have mulitple sqlite processed (not threads) running on the same file they can be mutually exclusive with each other.   

For in-memory filesystems this makes no sense as there is only one process.  For OPFS I guess we would need support for OPFS itself for this feature?

kripken: Oh, thanks, I missed that this was at the process level.

Yes, for OPFS maybe there are questions here at the spec level.

kripken, line 1524:
```suggestion
      // Pretend that the locking is successful. These are process-level locks,
      // and Emscripten programs are a single process. If we supported linking a
      // filesystem between programs, we'd need to do more here.
      return 0;
```

test/sqlite/test.out
```diff
@@ -1,7 +1,5 @@
 count(*) = 2
 
-datetime('2012-04-16 12:35:57', '+1 days') = 2012-04-17 12:35:57
```
kripken, line 3: I can't figure out what change led to this no longer being needed?

kripken (commented):
lgtm % comments/questions

test/sqlite/test.out
sbc100: I'm running a completely different test file, so this is a brand new test output.   The fact that github thinks I renamed and modified and existing file is misleading.

system/lib/wasmfs/syscalls.cpp
sbc100: Done, plus I opened a bug to track this.

test/sqlite/test.out
kripken: Ah, makes sense.

kripken (approved):

PR #23695, opened 2025-02-18
[test] AudioWorklet tests now correctly run and exit

cwoffenden:
All the audio tests in `test_browser.py` (except the mixer) were updated now the CI has audio, specifically:

- Tests that require audio were flagged as so (tests that don't are documented)
- Tests that currently fail for 2GB and wasm64 were flagged (only tests that play audio fail)
- All tests were migrated to `btest_exit()` as previously discussed
- (Tests moved `printf` calls to Emscripten's API and not stdio)
- All tests were run manually and the audio output verified
- All tests were run with `emscripten_force_exit` returning non-zero to verify they failed
- All of this was also verified that they're running on a Debian VM without audio to manually check the console for `"Test success"` and other outputs for tests without `requires_sound_hardware`

The mixer/AW struct test is a standalone PR in #23659.

Fixes: #23131

### DIFF
--- a/test/webaudio/audioworklet.c
+++ b/test/webaudio/audioworklet.c
@@ -79,7 +79,7 @@ bool main_thread_tls_access(double time, void *userData) {
 // This callback will fire after the Audio Worklet Processor has finished being
 // added to the Worklet global scope.
 void AudioWorkletProcessorCreated(EMSCRIPTEN_WEBAUDIO_T audioContext, bool success, void *userData) {
-  assert(success && "Stopped in AudioWorkletProcessorCreated" && success);
+  assert(success && "Stopped in AudioWorkletProcessorCreated");
 
   // Specify the input and output node configurations for the Wasm Audio
   // Worklet. A simple setup with single mono output channel here, and no
--- a/test/webaudio/audioworklet.c
+++ b/test/webaudio/audioworklet.c
@@ -80,8 +80,7 @@ bool main_thread_tls_access(double time, void *userData) {
 // added to the Worklet global scope.
 void AudioWorkletProcessorCreated(EMSCRIPTEN_WEBAUDIO_T audioContext, bool success, void *userData) {
   if (!success) {
-    emscripten_out("Stopped in AudioWorkletProcessorCreated");
-    assert(0);
+    assert("Stopped in AudioWorkletProcessorCreated" && success);
     return;
   }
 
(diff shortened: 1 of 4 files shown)


cwoffenden:
@sbc100 Would you prefer all of the AW tests working/enabled in batches like this, or all as one?

sbc100:
If you are asking about landing the fixes at the same time as the test updates, I think those are good to keep separate.    I

I like this this change simply updates test code, for example.

test/webaudio/audioworklet.c
```diff
@@ -80,8 +80,7 @@ bool main_thread_tls_access(double time, void *userData) {
 // added to the Worklet global scope.
 void AudioWorkletProcessorCreated(EMSCRIPTEN_WEBAUDIO_T audioContext, bool success, void *userData) {
   if (!success) {
-    emscripten_out("Stopped in AudioWorkletProcessorCreated");
-    assert(0);
+    assert("Stopped in AudioWorkletProcessorCreated" && success);
     return;
   }
```
sbc100, line 85: I would hop you can just do `assert(success && "Stopped in AudioWorkletProcessorCreated");` without the extra if / return.

test/webaudio/audioworklet_worker.c
```diff
@@ -13,17 +11,19 @@
 int workletToWorkerFutexLocation = 0;
 int workletToWorkerFlag = 0;
 
+void do_exit() {
+  emscripten_terminate_all_wasm_workers();
+  emscripten_force_exit(EXIT_SUCCESS);
```
sbc100, line 16: We normally just use `0` in test code rather than `EXIT_SUCCESS`

sbc100 (approved):

test/webaudio/audioworklet.c
cwoffenden: In this case, yes. I wasn't sure about code which also serves as examples that would be built without assertions (and would simply fail). But I'll change this.

test/webaudio/audioworklet_worker.c
cwoffenden: I'm ancient and proper C is `EXIT_SUCCESS`. But I'm not fussy.

sbc100 (approved):

cwoffenden:
> Is this change ready to land now?

Yes, this change is all done (as is the mixer PR).

PR #23700, opened 2025-02-19
[bigendian] Fix crash during PThread initialization

### LINKED ISSUE #23689: SUPPORT_BIG_ENDIAN+PThread does not work on big-endian machines
Trivial program (`int main(){}`) fails during startup with the following error messages
```
TypeError: wait.value.then is not a function
    at __emscripten_thread_mailbox_await (***/main.js:1:19568)
    at wasm://wasm/00015506:wasm-function[14]:0x553
    at callRuntimeCallbacks (***/main.js:1:17193)
    at initRuntime (***/main.js:1:7154)
    at doRun (***/main.js:1:27532)
    at run (***/main.js:1:27802)
    at removeRunDependency (***/main.js:1:8194)
    at ***/main.js:1:14279
```

while running on a big-endian machine with PThreads and SUPPORT_BIG_ENDIAN.

Using emscriptern 4.0.2., I built the program with the following command:
```
em++ -o main.js -s USE_PTHREADS=1 -s PROXY_TO_PTHREAD=1 -s PTHREAD_POOL_SIZE=8 -s SUPPORT_BIG_ENDIAN=1 -pthread -s EXIT_RUNTIME=1 -s ALLOW_MEMORY_GROWTH=1 -O2 main.cpp
```

If I manually adjust (byteswap) values in the following code (both `pthread_ptr` in `Atomics.waitAsync` and `1` in `Atomics.store`), everything starts working again

https://github.com/emscripten-core/emscripten/blob/2e2ec08220aa6e1ac664dee7d29aa6a5448ab312/src/lib/libpthread.js#L1231-L1237

slavek-kucera:
Notes:
- There are other places using Atomic.something that likely have the same issue.
- I tried keeping the `nativeByteOrder32` function in the `libpthread.js` and use the dependency mechanism, but it was always optimized into `x => x`.
- I tried to consider the MEMORY64 option, but it seems to not work at all in node (23 + appropriate experimental flag).
- Simple hello world output was built using the instructions from the issue and tested on s390x z/os machine.

Fixes: #23689


src/runtime_shared.js
```diff
@@ -83,6 +83,19 @@ function updateMemoryViews() {
 #endif
 }
 
+#if SUPPORT_BIG_ENDIAN
+var nativeByteOrder32 = (() => {
+    var h16 = new Int16Array(1);
+    var h8 = new Int8Array(h16.buffer);
+    h16[0] = 42;
+    return h8[0] === 42 && h8[1] === 0; // little endian ordering
```
sbc100, line 91: Surely we must already have helpers that perform this kind of thing in emscripten?  Since all the non-atomic read/writes already work?

slavek-kucera: Other accesses use the DataView methods.

slavek-kucera: Similar test is performed in the runtime_debug.js, but it is only included with ASSERTIONS option.

kripken:
The general mechanism we use for this is

https://github.com/emscripten-core/emscripten/blob/bb7fd251a48b4c4ef02111b9e4fa1fa4e0fb53f2/tools/acorn-optimizer.mjs#L1103-L1105

Perhaps the best thing is to update that pass so it handles Atomic operations?

slavek-kucera:
@kripken 
- The runtime test will need to be performed somewhere.
- The generic transformation might get quite complicated for e.g. waitAsync - wouldn't it be better at that point to introduce wrappers around these functions?

kripken:
>  The runtime test will need to be performed somewhere.

I agree, yes, that is necessary. I am just saying that doing it in the processing pass has advantages. Specifically it will find all atomic operations, guaranteeing that we don't forget any (as you wrote above, "There are other places using Atomic.something that likely have the same issue").

> The generic transformation might get quite complicated for e.g. waitAsync - wouldn't it be better at that point to introduce wrappers around these functions?

Oh, definitely, yes, we want to use wrapper functions. The processing pass should just add calls to those wrappers. That is what it does today: it replaces e.g. `HEAP16[x] = a` with `LE_HEAP_STORE_I16(x * 2, a)`. All the real work happens in the called function. For more examples, see

* test/js_optimizer/LittleEndianHeap.js
* test/js_optimizer/LittleEndianHeap-output.js

And `LE_HEAP_STORE` etc. is not defined in the pass, but in the JS, see

* src/lib/liblittle_endian_heap.js

I am suggesting that Atomics be implemented in a similar way.

slavek-kucera:
@kripken @sbc100 Is this more in line with the direction you had in mind?

src/runtime_shared.js
```diff
@@ -81,8 +81,64 @@ function updateMemoryViews() {
...
+   HEAPU16.unsigned = (x => x >>> 0);
+   HEAPU32.unsigned = (x => x >>> 0);
+#if WASM_BIGINT
+   HEAPU64.unsigned = (x => x >= 0 ? x : BigInt(2**64) + x);
+#endif
+#endif
 }
 
+#if SUPPORT_BIG_ENDIAN
+var nativeByteOrder = (() => {
```
sbc100, line 95: I think this can live in `liblittle_endian_heap.js` too as a library function.

src/runtime_shared.js
```diff
@@ -81,8 +81,64 @@ function updateMemoryViews() {
   {{{ maybeExportHeap('HEAP64')  }}}HEAP64 = new BigInt64Array(b);
   {{{ maybeExportHeap('HEAPU64') }}}HEAPU64 = new BigUint64Array(b);
 #endif
+
+#if SUPPORT_BIG_ENDIAN
+   HEAPU16.unsigned = (x => x >>> 0);
+   HEAPU32.unsigned = (x => x >>> 0);
+#if WASM_BIGINT
+   HEAPU64.unsigned = (x => x >= 0 ? x : BigInt(2**64) + x);
+#endif
```
sbc100, line 90: Ideally we could avoid adding this here and keep all the code chagnes to liblittle_endian_heap.js.

Ideally we could also avoid adding extra methods to these heaps, but maybe thats not easy.

tools/link.py
```diff
@@ -1122,7 +1122,20 @@ def phase_linker_setup(options, linker_args):  # noqa: C901, PLR0912, PLR0915
...
+      '$LE_ATOMICS_EXCHANGE',
+      '$LE_ATOMICS_ISLOCKFREE',
+      '$LE_ATOMICS_LOAD',
+      '$LE_ATOMICS_NOTIFY',
+      '$LE_ATOMICS_OR',
+      '$LE_ATOMICS_STORE',
+      '$LE_ATOMICS_SUB',
+      '$LE_ATOMICS_WAIT',
+      '$LE_ATOMICS_WAITASYNC',
+      '$LE_ATOMICS_XOR',
```
sbc100, line 1138: While I agree it would be good to add all of these, I also thing we might want to only add the ones that we actually use (and also test).

src/runtime_shared.js
slavek-kucera: I didn't really find a better way to keep the extra methods out while not introducing per-type functions (and associated transformations).

src/runtime_shared.js
```diff
@@ -81,8 +81,64 @@ function updateMemoryViews() {
...
+              | ((x & BigInt(0xff000000)) << BigInt(8) )
+              | ((x & BigInt(0x00ff0000)) << BigInt(24))
+              | ((x & BigInt(0x0000ff00)) << BigInt(40))
+              | ((x & BigInt(0x000000ff)) << BigInt(56));
+      if ((x & BigInt(0x80)) == 0)
+        return res;
+      else
+        return res - BigInt(2**64);
+    }
+    ),
```
sbc100, line 137: Do we need to support more than 4 byte access?  Maybe lets just skip then completely?    If anyone tries to do an atomic 64-bit access they would see a crash accessing nativeByteOrder OOB.

tools/link.py
sbc100: Yes, I suppose you right, seems reasonable to leave them all in then.

But lets skip 64-bit atomics and just fail in that case, to keep things simple.  I don't think anyone should be doing that.

slavek-kucera: I've tested it on the following program:
```cpp
#include <atomic>
#include <print>
#include <emscripten.h>

int main(){
        std::atomic<unsigned long long> x = 0x0123456789ABCDEF;

        std::println("x: {:X}", x.load());

        EM_ASM({ Atomics.compareExchange(HEAP64, ($0)/8, BigInt("0x0123456789ABCDEF"), BigInt("0xFEDCBA9876543210")); }, &x);

        std::println("x: {:X}", x.load());

}
```
built with
```
em++ -o main.js -s USE_PTHREADS=1 -s PROXY_TO_PTHREAD=1 -s PTHREAD_POOL_SIZE=8 -s SUPPORT_BIG_ENDIAN=1 -pthread -s EXIT_RUNTIME=1 -s ALLOW_MEMORY_GROWTH=1 -std=c++23 main.cpp
```
And I do get the expected result on both LE and BE:
```
x: 123456789ABCDEF
x: FEDCBA9876543210
```

src/runtime_shared.js
slavek-kucera: See the post above... I think that the 8B atomics will be needed eventually for the MEMORY64.

sbc100: I dont think we want to be adding complexity to support APIs that we don't use in emscripten.

src/runtime_shared.js
sbc100: Can you just use the uppercase version here?    i.e. why do we need the new lower case versions?

sbc100: I wonder if wen also move these lines to liblittle_endian_heap.js?

Perhaps they could be part of the `__postset` too?

slavek-kucera: re the new variables - if the uppercased variables are used they get replaced by e.g. the growable heap wrappers.

slavek-kucera: One option would be to rename `updateMemoryViews` to `_updateMemoryViews`. Assign `_updateMemoryViews` to `updateMemoryViews` and use to `__postset` to override the new `updateMemoryViews` variable.

If you prefer this option, I can look into that on Monday.

src/runtime_shared.js
slavek-kucera: That was just an example, perhaps more realistic would be just accessing 8B atomic or perhaps 2x4B pointer+counter pair value.

sbc100: Sure, but until we get a request that somebody wants to to actually do that from JS then I don't think we should support it.  Better to simply error out for now maybe?

slavek-kucera: I don't really understand your motivation here.
The current (production) implementation is just silently producing wrong results on BE. If WASM_BIGINT is off, this code is turned off, if it is on I don't really see the reason for not including it, given it is the identical problem as in the case of 4B accesses i.e. byte ordering.
And 8B atomic operations from JS on LE machines currently do work.

src/runtime_shared.js
sbc100: Regarding the `HEAPXX` variable substitution, we do have a list of function for which this is not performed.  I don't know if that will be useful in this case but the list is here: https://github.com/emscripten-core/emscripten/blob/cb8d502aa7213167bc773aab9c10d9666dd52c92/tools/acorn-optimizer.mjs#L1484-L1491

src/runtime_shared.js
sbc100: Perhaps I can explain my motivation a little better:   Emscripten has a lot of different settings and configurations in which it can be run.  One of our biggest issues is controlling the combinatorial complexity of all these settings.   We are always looks for way to avoid new options, or limit their scope.   I am also always looking for settings can can be removed completely or downscoped.

This often means I will, by default, argue against new settings of the expansion of existing ones.    Of course I often agree that certain settings are useful and will agree, especially if we can understand or limit their scope.

The BIG_ENDIAN support we have in emscripten has, IIUC, very few users, and almost zero testing.  If posssible, I think it would be great to delete it one day, but I'm also fine with fixing bugs in it, like this one, as along is it continues to exist.  

Ideally we could add some real testing on a big endian machine in order to prevent/detect regression, but that doesn't exist today.

Going back to this PR in particular, I think its great to get this bug fixed.   However, I would like to limit to scope of the fix to features that emscripten actually uses.  Because we don't have any usage of 64-bit atomics in JS I think we it would be reasonable to simple crash/assert if 64-bit JS atomics are used.   Otherwise we are shipping code that we neither use or test.

slavek-kucera: I removed the 8B support, even though I think it is a mistake.

slavek-kucera: As for testing in CI, there is an option on e.g. Ubuntu:
```
apt-get install -y qemu-user libc6-s390x-cross libstdc++6-s390x-cross
wget https://nodejs.org/dist/v22.14.0/node-v22.14.0-linux-s390x.tar.xz
tar xJf node-v22.14.0-linux-s390x.tar.xz
qemu-s390x -L /usr/s390x-linux-gnu/ node-v22.14.0-linux-s390x/bin/node program.js
```

sbc100 (commented):
Nice, this looks much cleaner now.  

Do you think we can find some way to test this in CI?  Or do we just have to rely on folks like you to manually test it once in while?   Can we perhaps run a few tests in quemu or some such?

src/runtime_shared.js
sbc100: Sorry I just saw this comment.  This looks like a good idea.   Should we do this as part of this PR, or open a bug to do this as a followup?

slavek-kucera:
I could be tested in CI via QEMU
```
apt-get install -y qemu-user libc6-s390x-cross libstdc++6-s390x-cross
wget https://nodejs.org/dist/v22.14.0/node-v22.14.0-linux-s390x.tar.xz
tar xJf node-v22.14.0-linux-s390x.tar.xz
qemu-s390x -L /usr/s390x-linux-gnu/ node-v22.14.0-linux-s390x/bin/node program.js
```

sbc100 (approved):
lgtm, either with some testing, or a bug to add some testing as a followup.

Can you update the PR description before we land?

@kripken can you take a final look at this?

src/runtime_shared.js
slavek-kucera: I suggest a follow up. It is not strictly connected to this PR.

kripken:
About testing, it does seem like QEMU or such is the only option. I would definitely not block on that, but it would be nice to have eventually, though only if it isn't very slow on CI.

sbc100:
> About testing, it does seem like QEMU or such is the only option. I would definitely not block on that, but it would be nice to have eventually, though only if it isn't very slow on CI.

True but we should just run a handful of tests in this most (maybe just one even).

PR #23701, opened 2025-02-19
[CI] Try enabling audio tests in Firefox

cwoffenden:
Work in progress. Following on from #23665 having succeeded in a Debian VM.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -356,12 +356,17 @@ commands:
       - run:
           name: configure firefox
           command: |
+            # Some before and after tests to try enabling a sound device:
+            lsmod | grep snd
+            modprobe snd-aloop
+            lsmod | grep snd
             mkdir ~/tmp-firefox-profile/
             cat > ~/tmp-firefox-profile/user.js \<<EOF
             user_pref("gfx.offscreencanvas.enabled", true);
             user_pref("javascript.options.shared_memory", true);
             user_pref("javascript.options.wasm_memory64", true);
             user_pref("dom.postMessage.sharedArrayBuffer.bypassCOOP_COEP.insecure.enabled", true);
+            user_pref("media.autoplay.default", 0);
             EOF
       - run:
           name: run tests (<< parameters.title >>)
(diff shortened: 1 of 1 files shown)


sbc100:
The circleCI runners support "retry with SSH" and you can log into them directly (in case that helps)

cwoffenden:
> The circleCI runners support "retry with SSH" and you can log into them directly

I'll need to create an account and look at it, but it's taking me away from what I'm actually trying to do.

I may still quickly try adding the PulseAudio package (which can run in the user account, AFAIK), but I'm going to need to come back to it.

sbc100:
Feel free to de-prioritize this if you other things you are working on.   We've be operating without any audio testing in CI for over 10 years now so a little more delay shouldn't be a problem.

cwoffenden:
Tomorrow I'll make #23659 work with the now working CI, specifically making `test_audio_worklet_params_mixing` run as part of the browser tests, since it covers most of the AW API (the other tests cover features and are really interactive, this single tests proves out the audio API pretty much as a whole).

Then I'm sitting on the code that fixes wasm64 and 2GB for audio.

cwoffenden:
Testing again locally, adding PulseAudio with its dummy mixer works, then starting it for the user:
```
apt-get install -q -y pulseaudio
pulseaudio --start
```
The CI machine seems to start installing this but fails with gstreamer and its dependencies:

https://app.circleci.com/pipelines/github/emscripten-core/emscripten/40699/workflows/ce65623e-e7d3-469f-a7e9-7c766d3a554b/jobs/903792?invite=true#step-108-1722_106

cwoffenden:
Woo-hoo! They're running:

https://app.circleci.com/pipelines/github/emscripten-core/emscripten/40706/workflows/d392b6a1-ea7b-41df-b410-ed763a9909d9/jobs/903939?invite=true#step-109-12288_96

I'd like to move the install to either its own section or with the Firefox installer, and possibly look at the warnings given (PulseAudio's not meant to be run as root).

.circleci/config.yml
```diff
@@ -353,15 +353,26 @@ commands:
           command: |
             wget -O ~/ff.tar.bz2 "https://download.mozilla.org/?product=firefox-nightly-latest-ssl&os=linux64&lang=en-US"
             tar -C ~ -xf ~/ff.tar.bz2
+      - run:
+          name: Add audio dependencies
+          command: |
+            # This should add and start PulseAudio's dummy mixer. It will warn
+            # that "This program is not intended to be run as root" but it can
+            # be ignored.
+            apt-get update -y; true
```
sbc100, line 362: Why `; true` there?  Does `apt-get update` fail for some reason?

sbc100 (approved):
lgtm % comment

.circleci/config.yml
cwoffenden: Yes, it's just in case it fails. The install is the important one, which if that fails then the test will definitely fail, so I was trying to reduce any flakiness.

cwoffenden: Removed.

PR #17915, opened 2022-09-22
Add support for `-sEXPORT_ES6`/`*.mjs` on Node.js

kleisauke:
As described in #11792, `require()` and `__dirname` doesn't exist in
an ES6 module. Emscripten uses this to import built-in core Node.js
modules. For example, the `node:fs` module is used for synchronously
importing the `*.wasm` binary, when not linking with `-sSINGLE_FILE`.

To work around this, ES6 modules on Node.js may import
`createRequire()` from `node:module` to construct the `require()`
function, allowing modules to be imported in a CommonJS manner.

Emscripten targets a variety of environments, which can be
categorized as:
1. Multi-environment builds, which is the default when
   `-sENVIRONMENT=*` is not specified at link time.
2. Single-environment, e.g. only web or Node.js as target.

For use case (1), this commit ensures that an `async` function is
emitted, allowing Node.js modules to be dynamically imported. This is
necessary given that static import declarations cannot be put in
conditionals. Inside the module, for Node.js only, it's using the
above-mentioned `createRequire()`-construction.

For use case (2), when only Node.js is targeted, a static import
declaration utilize the same `createRequire()`-construction.

For both use cases, `-sUSE_ES6_IMPORT_META=0` is not allowed, when
Node.js is among the targets, since it is not possible to mimic
`__dirname` when `import.meta` support is not available.

This commit does not change anything for use case (2), when only the
web is targeted (`-sENVIRONMENT=web`).

Resolves: #11792.
Resolves: #17797.

PR #23704, opened 2025-02-19
Enable createWasm to use async/await with closure

sbc100:
I really which we had a better solution here, but this makes the use of async/await independent of closure usage.

Fixes: #23158


kleisauke:
FWIW, I once successfully implemented a better solution for this, but it turned out to be a large change, see:
https://github.com/emscripten-core/emscripten/pull/17915#issuecomment-1311871913

tools/link.py
```diff
@@ -2071,7 +2071,8 @@ def fix_es6_import_statements(js_file):
   src = read_file(js_file)
   write_file(js_file, src
              .replace('EMSCRIPTEN$IMPORT$META', 'import.meta')
-             .replace('EMSCRIPTEN$AWAIT$IMPORT', 'await import'))
+             .replace('EMSCRIPTEN$AWAIT$IMPORT', 'await import')
+             .replace('EMSCRIPTEN$AWAIT(createWasm())', 'await creasteWasm()'))
```
kleisauke, line 2075:
```suggestion
             .replace('EMSCRIPTEN$AWAIT(createWasm())', 'await createWasm()'))
```

sbc100: Ooops.. how did that even work.

sbc100: I see.. the only effected test that I ran doesn't actually try to execute the resulting code: other.test_codesize_minimal_esm

brendandahl (approved):

PR #23707, opened 2025-02-19
Fix for vite bundling with pthreads enabled

sbc100:
Replaces #23607
Fixes: #22394

### DIFF
--- a/src/lib/libpthread.js
+++ b/src/lib/libpthread.js
@@ -29,6 +29,30 @@ globalThis.MAX_PTR = Number((2n ** 64n) - 1n);
 #else
 globalThis.MAX_PTR = (2 ** 32) - 1
 #endif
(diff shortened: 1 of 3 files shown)


src/lib/libpthread.js
```diff
@@ -29,6 +29,30 @@ globalThis.MAX_PTR = Number((2n ** 64n) - 1n);
 #else
 globalThis.MAX_PTR = (2 ** 32) - 1
 #endif
+// Use a macro to avoid duplicating pthread worker options.
+// We cannot use normal JS variable since the vite bundler requires that worker
```
kripken, line 33:
```suggestion
// We cannot use a normal JS variable since the vite bundler requires that worker
```

kripken (approved):
``

PR #23708, opened 2025-02-19
[test] Add test of rollup using node in test_other. NFC

sbc100:
This complements the existing browser test in `test_browser.py`.

### DIFF
--- a/test/rollup_node/index.mjs
+++ b/test/rollup_node/index.mjs
@@ -1,0 +1,5 @@
+import Module from './hello.mjs';
+
+await Module();
+
+console.log('done');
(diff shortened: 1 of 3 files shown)


test/rollup_node/index.mjs
```diff
@@ -0,0 +1,5 @@
+import Module from './hello.mjs';
+
+Module().then((instance) => {
```
brendandahl, line 3: could be `const instance = await Module();` ...

sbc100: Sure!

brendandahl (approved):

PR #23712, opened 2025-02-20
Update npm dependencies. NFC

sbc100:
This change is simply the result of running `npm update`.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -20,7 +20,10 @@ STAMP_DIR = os.path.join(__rootdir__, 'out')
 from tools import shared, utils
 
 actions = [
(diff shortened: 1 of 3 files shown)


kripken (approved):

sbc100:
I can't for the life of me figure out why this change causes a slight codesize improvement for just one test.

kripken:
Closure compiler change perhaps?

sbc100:
> Closure compiler change perhaps?

But closure compiler itself doesn't change version here.  I think maybe because the lockfile version changed the way dependencies work between node modules looks different.. but its strange that its just that one test that changed.  I wish I understood what is doing on here.

PR #23721, opened 2025-02-21
Add preprocessor builtin __EMSCRIPTEN_HAS_ASYNCIFY__

kainino0x:
Proposal.
This can be useful in libraries which want to provide additional features when Asyncify is available. It is similar to emscripten_has_asyncify(), but since it's a macro it can reduce code size. (We want to use it in `webgpu.cpp`, but without requiring some build file setup that will set a custom preprocessor macro correctly.)

### DIFF
--- a/emcc.py
+++ b/emcc.py
@@ -439,6 +439,10 @@ def get_cflags(user_args):
   if settings.WASM_WORKERS:
     cflags.append('-D__EMSCRIPTEN_WASM_WORKERS__=1')
 
+  if settings.ASYNCIFY:
+    # Note this macro doesn't care whether we're using JSPI or not.
+    cflags.append('-D__EMSCRIPTEN_HAS_ASYNCIFY__=1')
+
   if not settings.STRICT:
     # The preprocessor define EMSCRIPTEN is deprecated. Don't pass it to code
     # in strict mode. Code should use the define __EMSCRIPTEN__ instead.
--- a/site/source/docs/api_reference/emscripten.h.rst
+++ b/site/source/docs/api_reference/emscripten.h.rst
@@ -1090,6 +1090,7 @@ Functions
 .. c:function:: int emscripten_has_asyncify()
 
   Returns whether pseudo-synchronous functions can be used.
+  (Use the preprocessor macro ``__EMSCRIPTEN_HAS_ASYNCIFY__`` if compile-time switching is needed.)
 
   :rtype: int
   :returns: 1 if program was compiled with -sASYNCIFY, 0 otherwise.


kainino0x:
I added partial documentation by mentioning it in `emscripten_has_asyncify()`, but is there a good place for primary documentation? (if otherwise this looks good to you)

sbc100 (commented):
I would really love to avoid adding this if we can find a way to.

I've been on a mission for a few years to make emscripten as close to bare clang as possible and these kind of macros make it hard to get there.

I fought to avoid adding `__EMSCRIPTEN_WASM_WORKERS__` above but was convinced in the end... so maybe I could be convinced again, but maybe we can find an alternative?

sbc100:
> (We want to use it in `webgpu.cpp`, but without requiring some build file setup that will set a custom preprocessor macro correctly.

But wouldn't you want to just build libwebgpu.a once?   It would seem odd to have two different library flavors, once for asyncify and one for not-asyncify?

kainino0x:
> But wouldn't you want to just build libwebgpu.a once? It would seem odd to have two different library flavors, once for asyncify and one for not-asyncify?

I don't think this particularly matters to us, we're going to distribute webgpu.cpp, not prebuilt binaries (regardless of whether it ends up inside or outside Emscripten), so it'll get built along with people's projects.

The way I wanted to use it would remove some symbols that are unused (a JS function called from Wasm that we only need with Asyncify). However we don't _really_ need to do that, we could rely on the compiler to do it if `emscripten_has_asyncify()` were `constexpr`:
`if constexpr (emscripten_has_asyncify()) { function_that_may_get_optimized_out(); }`

Should I pursue that as an alternative?

kainino0x:
Wait, that isn't possible because it would have to be defined inside `emscripten.h` which would need to know whether asyncify is enabled or not. Which requires something like this (or some kind of intrinsic).

kainino0x:
Is clang responsible for implementing JSPI? If so maybe it has intrinsics to check whether it's enabled? (it would be fine if we can only optimize this code out when using JSPI, since I'm assuming JSPI is the future)

sbc100:
> Is LLVM responsible for implementing JSPI? If so maybe it has intrinsics to check whether it's enabled? (it would be fine if we can only optimize this code out when using JSPI, since I'm assuming JSPI is the future)

Both ASYNCIFY and JSPI are implemented at post-link time.  So normally the compiler doesn't know if these things are available.

The upside of that of course is that you can share object files and libraries between users of JSPI and ASYNCIFY and neither.

I think the alternative would be simply to ask your users to build with `-DASYNCIFY` when building your source files.

emcc.py
```diff
@@ -439,6 +439,10 @@ def get_cflags(user_args):
   if settings.WASM_WORKERS:
     cflags.append('-D__EMSCRIPTEN_WASM_WORKERS__=1')
 
+  if settings.ASYNCIFY:
+    # Note this macro doesn't care whether we're using JSPI or not.
+    cflags.append('-D__EMSCRIPTEN_HAS_ASYNCIFY__=1')
```
sbc100, line 444: Ah yes, the other reason this doesn't work today is that folks don't use `-sASYNCIFY` is the link-only setting.   So you will actually see a warning if you use it at compile time:

```
$ ./emcc -c ~/test/hello.c -sASYNCIFY
emcc: warning: linker setting ignored during compilation: 'ASYNCIFY' [-Wunused-command-line-argument]
```

kainino0x: Ah, fair, I totally forgot about that. OK, `-DASYNCIFY` it is for now then. Thanks!

PR #23726, opened 2025-02-22
[jscompiler] Simplify pushing and popping the current file being processed. NFC

sbc100:


src/modules.mjs
```diff
@@ -257,6 +258,8 @@ export const LibraryManager = {
       },
     });
 
+
+
```
kripken, line 262: excessive spacing?

src/utility.mjs
```diff
@@ -46,17 +46,19 @@ export function warningOccured() {
...
+
+export function popCurrentFile(f) {
+  currentFile.pop(f)
 }
 
 function errorPrefix() {
-  if (currentFile) {
-    return currentFile + ': ';
+  if (currentFile.length > 0) {
+    return currentFile[currentFile.length -1 ] + ': ';
```
kripken, line 61:
```suggestion
    return currentFile[currentFile.length - 1] + ': ';
```

kripken (approved):

PR #23727, opened 2025-02-22
[jscompiler] Cleanup warn/warnOnce utility functions. NFC

sbc100:
I'm not sure why we have the two argument versions of these functions.

The two-argument version of warnOnce was introduced in ca58e84149 but doesn't seem to have ever had any callers.

The two-argument version of `warn` was introduced in 46e3b2a787, but I can't see any users of it from back then.  It was modified to take a single argument in 06b58cbbe.

### DIFF
--- a/src/utility.mjs
+++ b/src/utility.mjs
@@ -62,26 +62,16 @@ function errorPrefix() {
   }
 }
 
-export function warn(a, msg) {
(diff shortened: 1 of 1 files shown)


sbc100:
ping

kripken (approved):

PR #23728, opened 2025-02-22
[docs] Document limitation on std::set_terminate and __cxa_begin_catch

### LINKED ISSUE #23720: `std::current_exception` is NULL in termination handler
`std::current_exception` is nullptr when called from a termination handler.

# Steps to reproduce
- Create `main.cpp` with the following contents:
```cpp
#include <exception>
#include <iostream>
#include <stdexcept>
#include <utility>

static void printCurrentException() {
	auto exPtr = std::current_exception();
	std::cerr << "exception_ptr = ";
	if (exPtr)
		try {
			std::rethrow_exception(std::move(exPtr));
		} catch (const std::runtime_error &ex) {
			std::cerr << "runtime_error: " << ex.what() << '\n';
		} catch (...) {
			std::cerr << "non-runtime_error\n";
		}
	else std::cerr << "NULL\n";
}

void noThrow() noexcept {
	try {
		throw std::runtime_error("error1");
	} catch (...) {
		printCurrentException();
	}
	std::set_terminate([] {
		printCurrentException();
		std::abort();
	});
	throw std::runtime_error("error2");
}

int main() {
	noThrow();
}
```
- Now compile with exceptions (optionally with debug symbols):
```shell
em++ -fexceptions main.cpp -o hello.js
```
- Now run the file (e.g. with Node.js, or use `-o hello.html` and open with a browser), and observe output:
```
$ node hello.js

exception_ptr = runtime_error: error1
exception_ptr = NULL
Aborted(native code called abort())
```
- Note that `std::current_exception()` was properly set in the `catch` block, but the termination handler got a `nullptr`, while it should've been a pointer wrapping `std::runtime_error("error2")`.

The same happens with `-fwasm-exceptions` and `-fwasm-exceptions -sNO_WASM_LEGACY_EXCEPTIONS` (run with `--experimental-wasm-exnref` as first parameter to `node`).

# Details

The standard says [the following](https://eel.is/c++draft/propagation#lib:current_exception)
> ```cpp
> constexpr exception_ptr current_exception() noexcept;
> ```
> 9. Returns: An exception_ptr object that refers to the [currently handled exception](https://eel.is/c++draft/except.handle#def:exception_handling,currently_handled_exception) or a copy of the currently handled exception, or a null exception_ptr object if no exception is being handled[.](https://eel.is/c++draft/propagation#9.sentence-1)
[...]

[and](https://eel.is/c++draft/except.handle#def:exception_handling,currently_handled_exception):
> 8. If no matching handler is found, the function std​::​terminate is invoked; whether or not the stack is unwound before this invocation of std​::​terminate is implementation-defined ([[except.terminate]](https://eel.is/c++draft/except.terminate))[.](https://eel.is/c++draft/except.handle#8.sentence-1)
> 9. A handler is considered [active](https://eel.is/c++draft/except.handle#def:exception_handling,handler,active) when initialization is complete for the parameter (if any) of the catch clause[.](https://eel.is/c++draft/except.handle#9.sentence-1)
[Note [4](https://eel.is/c++draft/except.handle#note-4): The stack will have been unwound at that point[.](https://eel.is/c++draft/except.handle#9.sentence-2)
— end note]
Also, an implicit handler is considered active when the function std​::​terminate is entered due to a throw[.](https://eel.is/c++draft/except.handle#9.sentence-3)
A handler is no longer considered active when the catch clause exits[.](https://eel.is/c++draft/except.handle#9.sentence-4)
> 10. The exception with the most recently activated handler that is still active is called the [currently handled exception](https://eel.is/c++draft/except.handle#def:exception_handling,currently_handled_exception)[.](https://eel.is/c++draft/except.handle#10.sentence-1)

So, as I understand it, the "implicit handler" should be active at this point. Besides, it works fine when compiled for x86 using Clang or GCC.

# Version of emscripten/emsdk
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.3 (a9651ff57165f5710bb09a5fe52590fd6ddb72df)
clang version 21.0.0git (https:/github.com/llvm/llvm-project 6dc41a639334b913e762f65410fcd14a722b137f)
Thread model: posix
InstalledDir: /home/swdv/emsdk/upstream/bin
```

aheejin:
This documents the problems reported in #23720.

1. Due to the lack of two-phase EH, we, both in Emscripten EH and Wasm EH, do not support `std::set_terminate`. This documents the limitation.
2. Due to another LLVM implementation reason, `__cxa_begin_catch` is not called before when EH handling calls `std::terminate`. This is documented more in detail in #23779.


site/source/docs/porting/exceptions.rst
```diff
@@ -171,3 +171,14 @@ Using Exceptions and setjmp-longjmp Together
...
 
 See :ref:`using-exceptions-and-setjmp-longjmp-together`.
+
+
+Limitations
+===========
+
+Currently `std::set_terminate
+<https://en.cppreference.com/w/cpp/error/set_terminate>`_ is NOT supported, both
+in Emscripten-style and WebAssembly exceptions. The functionality requires the
```
kripken, line 181:
```suggestion
in Emscripten-style and WebAssembly exceptions. The functionality requires
```

aheejin:
Added documentation on another problem described in #23779.

aheejin:
@sbc100 
> Can/should we have calls to `set_terminate` trap (or warn in debug builds)?

Turns out `set_terminate` is not supported only when an exception is not caught, and we don't know whether an exception will be caught beforehand, so I don't think we can.

kripken (approved):

sbc100:
> Turns out `set_terminate` is not supported only when an exception is not caught, and we don't know whether an exception will be caught beforehand, so I don't think we can.

Do you mean that it is support when the program exits normally?  i.e. it acts like `atexit()`?

aheejin:
> > Turns out `set_terminate` is not supported only when an exception is not caught, and we don't know whether an exception will be caught beforehand, so I don't think we can.
> 
> Do you mean that it is support when the program exits normally? i.e. it acts like `atexit()`?

No it runs only when `std::terminate` runs due to an exception. I think my previous comment was either incorrect or unclear. So in this program (which I used in #23779):
```
#include <iostream>
#include <exception>

int main() noexcept {
  std::set_terminate([] {
    auto ptr = std::current_exception();
    if (ptr)
      std::cerr << "exception_ptr is NOT null" << std::endl;
    else
      std::cerr << "exception_ptr is null" << std::endl;
    std::abort();
  });
  throw 3;
}
```

Because of `noexcept`, the stack is not unwound, so this is not the case where we can't run the terminate handler when an uncaught exception unwinds all the stack. Rather, the function calls `std::terminate` within this function when an exception is thrown (because `noexcept`). In this case, that custom terminate handler, which prints either `exception is null / NOT null` runs. It only cannot access the current exception, because of the limitation I listed in the second bullet point and wrote in detail in #23779.

aheejin:
@sbc100 Do you have any other comments?

site/source/docs/porting/exceptions.rst
```diff
@@ -171,3 +171,40 @@ Using Exceptions and setjmp-longjmp Together
...
+Limitations
+===========
+
+  * Currently `std::set_terminate
+    <https://en.cppreference.com/w/cpp/error/set_terminate>`_ is NOT supported
+    when a thrown exception does not have a matching handler and unwinds all the
+    stack and crashes the program. This applies to both Emscripten-style and
+    WebAssembly exceptions. That functionality requires `two-phase exception
+    handling <https://itanium-cxx-abi.github.io/cxx-abi/abi-eh.html>`_, which
+    neither supports.
```
sbc100, line 185: This seems to say that `set_terminate` will not work.. but in the example below you show it working (although without a valid `current_exception`).   These two things seems to be in contradiction with each other.

I guess maybe the `noexcept` thing on `main` is important here and that meas that there is maybe some kind of handler implicitly added to `main`?   but that isn't obvious to me at least ..

aheejin: > This seems to say that `set_terminate` will not work.. but in the example below you show it working (although without a valid `current_exception`). These two things seems to be in contradiction with each other.

So there's a condition written in the sentence:
> ... is NOT supported when a thrown exception does not have a matching handler and unwinds all the stack and crashes the program.

When we call `std::terminate` is not when all stack is unwound due to an exception.
 
> I guess maybe the `noexcept` thing on `main` is important here and that means that there is maybe some kind of handler implicitly added to `main`? but that isn't obvious to me at least ..

If a function is marked `noexcept`, the code catches an exception within the function and calls `std::terminate` (or `__clang_call_terminate`, which calls `std::terminate`), and it does not throw the exception to the caller. So the stack unwinding does not happen.

sbc100: I still think its a little confusing though since the main function you wrote looks like it doesn't catch anything (since it contains no catches).   

Understanding the `noexcept` injects a invisible catch-all handler is required in order to understand what is going on there.   That seems pretty non-obvious to me.

Maybe add comment right above main saying somthing like "The use of `noexcept` here means that the `throw 3` will turn into a termination condition".

aheejin: Added "The use of `noexcept` here means that the `throw 3` will turn into a termination condition" as you suggested in the second bullet point, and also added a little more clarifying word in the first bullet point.

aheejin: Also added an example program to the first bullet point to make the situation clearer.

site/source/docs/porting/exceptions.rst
```diff
@@ -171,3 +171,57 @@ Using Exceptions and setjmp-longjmp Together
 ============================================
 
 See :ref:`using-exceptions-and-setjmp-longjmp-together`.
+
+
+Limitations
```
sbc100, line 176: Since both of these limitation are regarding `std::set_terminate` should we call this section `Limitations regarding std::terminate`?

site/source/docs/porting/exceptions.rst
```diff
@@ -171,3 +171,57 @@ Using Exceptions and setjmp-longjmp Together
...
+        int main() noexcept {
+          std::set_terminate([] {
+            auto ptr = std::current_exception();
+            if (ptr)
+              std::cerr << "exception_ptr is NOT null" << std::endl;
+            else
+              std::cerr << "exception_ptr is null" << std::endl;
+            std::abort();
+          });
+          throw 3;
```
sbc100, line 224: Would writing `std::terminate()` here have the same effect?

site/source/docs/porting/exceptions.rst
aheejin: Done

sbc100 (approved):

PR #23729, opened 2025-02-22
[AUDIO_WORKLET] Enable AW spinlocks, verify them in the browser tests

### LINKED ISSUE #22962: Interactive audio worklet futex test fails

cwoffenden:
- test converted to C-only (still in a C++ file for now, see below)

This was unmerged last year (#22995), visited again now the browser tests are running with audio. It's been moved over to `btest_exit()` and verified that it runs (and will only exit after all tests have run or asserted).

The only change I'd make is to rename the original test file, which has grown beyond its original use. A `git mv` command won't survive multiple commits, so I want to either leave it until last, live with the resulting deletion, or rename as a separate PR.

Fixes: #22962

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -5522,7 +5522,7 @@ Module["preRun"] = () => {
   @requires_sound_hardware
   @also_with_minimal_runtime
   def test_audio_worklet_emscripten_locks(self):
-    self.btest_exit('webaudio/audioworklet_emscripten_futex_wake.cpp', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-pthread', '-sPTHREAD_POOL_SIZE=2'])
+    self.btest_exit('webaudio/audioworklet_emscripten_futex_wake.cpp', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-pthread'])
 
   def test_error_reporting(self):
     # Test catching/reporting Error objects
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -5519,7 +5519,7 @@ Module["preRun"] = () => {
   # Tests AudioWorklet with emscripten_lock_busyspin_wait_acquire() and friends.
   @requires_sound_hardware
   @also_with_minimal_runtime
-  def test_audio_worklet_emscripten_futex_wake(self):
+  def test_audio_worklet_emscripten_locks(self):
     self.btest_exit('webaudio/audioworklet_emscripten_futex_wake.cpp', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-pthread', '-sPTHREAD_POOL_SIZE=2'])
 
   def test_error_reporting(self):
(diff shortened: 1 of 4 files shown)


sbc100 (approved):

test/test_browser.py
```diff
@@ -5516,12 +5516,13 @@ def test_audio_worklet_params_mixing(self, args):
...
+  # Tests AudioWorklet with emscripten_lock_busyspin_wait_acquire() and friends
+  @no_wasm64('https://github.com/emscripten-core/emscripten/pull/23508')
+  @no_2gb('https://github.com/emscripten-core/emscripten/pull/23508')
   @requires_sound_hardware
   @also_with_minimal_runtime
-  @disabled('https://github.com/emscripten-core/emscripten/issues/22962')
-  def test_audio_worklet_emscripten_futex_wake(self):
-    self.btest('webaudio/audioworklet_emscripten_futex_wake.cpp', expected='0', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-pthread', '-sPTHREAD_POOL_SIZE=2'])
+  def test_audio_worklet_emscripten_locks(self):
+    self.btest_exit('webaudio/audioworklet_emscripten_futex_wake.cpp', args=['-sAUDIO_WORKLET', '-sWASM_WORKERS', '-pthread', '-sPTHREAD_POOL_SIZE=2'])
```
sbc100, line 5525: Does this test actually start any pthreads?  I.e. can we remove `-sPTHREAD_POOL_SIZE`?

cwoffenden: I’ll take a look tomorrow, I kept this from the original interactive test.

sbc100 (commented):
lgtm. 

CC'ing @juj in case he has some input.

PR #23730, opened 2025-02-23
Support ES6 module syntax in Acorn and Closure

kleisauke:
Acorn and Closure are now made aware of ES6 module syntax when
linking with `-sEXPORT_ES6`. The previous mangling of top-level
await, `import.meta` and dynamic `import()` is tightened to only
mangle top-level await when Closure is enabled, since this is the
only construct not supported there.


test/code_size/hello_webgl_wasm.json
```diff
@@ -1,10 +1,10 @@
 {
   "a.html": 454,
   "a.html.gz": 328,
-  "a.js": 3920,
-  "a.js.gz": 2089,
+  "a.js": 5629,
```
sbc100, line 4: This is pretty big regression sadly.

kleisauke: Here's the consolidated and prettified diff between main and this PR for `other.test_minimal_runtime_code_size_hello_webgl_wasm`:

<details>
  <summary>Details</summary>

```diff
--- a/a.js
+++ b/a.js
@@ -1,4 +1,4 @@
-var Module = async function (B = {}) {
+async function Module(B = {}) {
   var l,
     p,
     r,
@@ -7,6 +7,88 @@ var Module = async function (B = {}) {
     y,
     D,
     E,
+    h = {
+      xa: 0,
+      Ha: 0,
+      Ka: 0,
+      Qa: 0,
+      hb: 0,
+      eb: 0,
+      Ea: 0,
+      Ca: 0,
+      Sa: 0,
+      Ba: 0,
+      Ga: 0,
+      Ta: 0,
+      gb: 0,
+      cb: 0,
+      Oa(a, b, c) {
+        r.set(r.subarray(b, b + c), a);
+      },
+      Ya() {
+        for (; h.L.length; ) h.V(h.L.length - 1);
+        h.N = [];
+      },
+      X: 0,
+      N: [],
+      za(a, b, c) {
+        function d(k, p) {
+          if (k.length != p.length) return !1;
+          for (var q in k) if (k[q] != p[q]) return !1;
+          return !0;
+        }
+
+        for (var f of h.N) if (f.Y == a && d(f.$, c)) return;
+        h.N.push({Y: a, da: b, $: c});
+        h.N.sort((k, p) => k.da < p.da);
+      },
+      Za(a) {
+        h.N = h.N.filter((b) => b.Y != a);
+      },
+      ja() {
+        return navigator.userActivation ? navigator.userActivation.isActive : h.X && h.la.wa;
+      },
+      ea() {
+        if (h.ja()) {
+          var a = h.N;
+          h.N = [];
+          for (var b of a) b.Y(...b.$);
+        }
+      },
+      L: [],
+      sa: (a, b) => {
+        for (var c = 0; c < h.L.length; ++c) h.L[c].target != a || (b && b != h.L[c].T) || h.V(c--);
+      },
+      V(a) {
+        var b = h.L[a];
+        b.target.removeEventListener(b.T, b.ba, b.va);
+        h.L.splice(a, 1);
+      },
+      Xa(a) {
+        if (!a.target) return -4;
+        if (a.ya)
+          (a.ba = function (c) {
+            ++h.X;
+            h.la = a;
+            h.ea();
+            a.Ja(c);
+            h.ea();
+            --h.X;
+          }),
+            a.target.addEventListener(a.T, a.ba, a.va),
+            h.L.push(a);
+        else
+          for (var b = 0; b < h.L.length; ++b)
+            h.L[b].target == a.target && h.L[b].T == a.T && h.V(b--);
+        return 0;
+      },
+      Ia(a) {
+        return a ? (a == window ? '#window' : a == screen ? '#screen' : a?.nodeName || '') : '';
+      },
+      fullscreenEnabled() {
+        return document.fullscreenEnabled || document.webkitFullscreenEnabled;
+      },
+    },
     M = new TextDecoder(),
     w = (a, b) => {
       if (!a) return '';
@@ -283,4 +365,8 @@ var Module = async function (B = {}) {
     L();
   });
   return {};
-};
+}
+
+'object' === typeof exports && 'object' === typeof module
+  ? ((module.exports = Module), (module.exports.default = Module))
+  : 'function' === typeof define && define.amd && define([], () => Module);
```
</details>


One factor is that the CJS/AMD module exports are now present under `MINIMAL_RUNTIME` (see PR #23282). The other factor is likely(?) that JSDCE is unable to remove certain parts of `libhtml5.js`.

sbc100: I think we should try to land f53a92d as its own patch maybe?

kleisauke: PR #23850 resolves this size regression instead. 

It seems that `other.test_minimal_runtime_code_size_hello_webgl_wasm` depends on:
- `emscripten_set_element_css_size()`
- `emscripten_set_canvas_element_size()`
- `emscripten_webgl_create_context()`

All of these had an unused dependency on `$JSEvents`, which, for some reason, JSDCE fails to remove in this changeset.

sbc100: Are you sure its JSDCE and not closure compile that is failing to remove these things?

That was my experience too in #23261.. I figured the optimizers were just less good at removing non-top-level code?

kleisauke: Closure might also be failing to remove this unused dependency - I haven't checked.

We could try replacing Closure with https://github.com/swc-project/swc, as it shows promising results in https://github.com/privatenumber/minification-benchmarks. However, I'm unsure how much effort that would require.

sbc100: I would *love* to replace closure with pretty much anything else :)

However, so far nothing as come along that can't do as good of a job at minifying.   Maybe its worth looking into swc!

kleisauke: > I figured the optimizers were just less good at removing non-top-level code?

I think you're right. After rebasing and revising this PR, it looks like the code size increases in `test_minimal_runtime_code_size_hello_webgl*` are due to the code no longer being defined at the top level, which prevents further optimization.

kleisauke: ... I revised this PR again to remove the dependence on PR #23261, which avoids any code size regressions. PTAL.

kleisauke:
PTAL, this would help PR #26384. Though, perhaps PR #26604 should land first, as it would make the changes to `strip-node-prefix.mjs‎` and `test_other.py` unnecessary.

src/shell.js
```diff
@@ -112,8 +112,7 @@ if (ENVIRONMENT_IS_NODE) {
   // When building an ES module `require` is not normally available.
   // We need to use `createRequire()` to construct the require()` function.
   const { createRequire } = await import('node:module');
-  /** @suppress{duplicate} */
-  var require = createRequire(import.meta.url);
+  global.require = createRequire(import.meta.url);
```
sbc100: Should we prefer `globalThis` here?

third_party/terser/README.md
```diff
@@ -7,7 +7,7 @@ version published in npm.
 This `terser.js` bundle in this directory was built from our fork of terser
 which lives at: https://github.com/emscripten-core/terser/
 
-The current patches are stored in the `emscripten_patches_v4.8.0` branch.
+The current patches are stored in the `emscripten_patches_v5.18.2` branch.
```
sbc100, line 10: Is this a driveby fix?  Was this just no updated at some point?

tools/emscripten.py
```diff
@@ -427,6 +427,11 @@ def emscript(in_wasm, out_wasm, outfile_js, js_syms, finalize=True, base_metadat
   assert '//FORWARDED_DATA:' in out, 'Did not receive forwarded data in pre output - process failed?'
   glue, forwarded_data = out.split('//FORWARDED_DATA:', 1)
 
+  # Unmangle previously mangled `import.meta` references in lib*.js.
+  # See also: `LibraryManager.load` in modules.mjs.
+  if settings.EXPORT_ES6:
+    glue = glue.replace('EMSCRIPTEN$IMPORT$META', 'import.meta')
```
sbc100: Is there a reason we cannot do the demangling in the JS compiler (i.e. the same place where we do the mangling?)

src/shell.js
kleisauke: We use `global.Worker` below, so aligning with that seemed reasonable (and it saves 4 bytes). Though, it appears to be legacy:
https://nodejs.org/api/globals.html#global

(there was a reason I originally did this, likely something related to Closure, but I can't recall the exact reason anymore)

src/modules.mjs
```diff
@@ -329,6 +329,15 @@ export const LibraryManager = {
...
     }
 
+    if (EXPORT_ES6) {
+      // `vm.runInContext` doesn't support module syntax; to allow it,
+      // we need to temporarily replace `import.meta` usages with
+      // placeholders during the JS compile phase, then in Python
+      // we reverse this replacement.
+      // See also: `emscript` in emscripten.py.
+      contents = contents.replace(/\bimport\.meta\b/g, 'EMSCRIPTEN$IMPORT$META');
+    }
```
sbc100: Why does this need to move from `preprocess` to here?

Its a little confusing to have this mangling done in two different places.

third_party/terser/README.md
kleisauke: Yep, a drive-by fix, since we're now using Terser 5.18.2.

tools/emscripten.py
kleisauke: Ah, good idea. Let me try to move this to the JS compiler, I think that would work.

src/shell.js
kleisauke: ... looking at commit ef36474b9347f074f01dbdbe6be96827413be2ef, I think that I know the reason behind this. This was likely necessary for the now-removed `strip-node-prefix.mjs` Babel plugin, which required `require()` to remain unminified. Let me revert this.

kleisauke: Reverted this with commit af43dd135c8e0f53797f9c39f054935664a235c8, which further reduces the codesize expectations of `codesize.test_codesize_minimal_esm`.

tools/emscripten.py
kleisauke: ... that didn't work because the library contents (e.g. `libpthread.js` where `import.meta` is used) are streamed to `stdout`.
https://github.com/emscripten-core/emscripten/blob/426fb2df42aadcebf49a47420956da24227609e4/src/jsifier.mjs#L394-L401

So, this is probably the only place where the full file contents is available and where you can a single replace in one go. Added a comment with cf6417c3c351c531eac0b6315367110bd1f16172 instead.

src/modules.mjs
kleisauke: Moved back to `preprocess` with commit 5c2f12b00778bc827aaa2eca29dab9c882ef9d84 (and also updated to use `replaceAll` in commit 0b218f275571c43bb57d480f843670d81aa8f49f to avoid regex).

tools/emscripten.py
sbc100: Maybe we could run this transform over all of the `libraryItems` before we call writeOutput?

tools/emscripten.py
```diff
@@ -427,6 +427,13 @@ def emscript(in_wasm, out_wasm, outfile_js, js_syms, finalize=True, base_metadat
   assert '//FORWARDED_DATA:' in out, 'Did not receive forwarded data in pre output - process failed?'
   glue, forwarded_data = out.split('//FORWARDED_DATA:', 1)
 
+  # Unmangle previously mangled `import.meta` references in lib*.js.
+  # See also: `preprocess` in parseTools.mjs.
+  # Note: this needs to be done here because this is the only stage
+  # where the complete file contents are available.
```
sbc100: But why can't this be done in `fix_js_mangling`?

tools/emscripten.py
kleisauke: Commit 11a3324fc6738fe5cf5d09d30ee6ef103ed4f986 moves this to `writeOutput`. I also tried to do it here:
https://github.com/emscripten-core/emscripten/blob/d984d060ea4d89ea119a3070f09b1b29d9e3b9a4/src/jsifier.mjs#L888-L892

However, `shell.js` also uses `import.meta`, so that won't work.
https://github.com/emscripten-core/emscripten/blob/d984d060ea4d89ea119a3070f09b1b29d9e3b9a4/src/jsifier.mjs#L882-L883

tools/emscripten.py
kleisauke: `import.meta` is supported by Acorn, Babel and Terser when `allowImportExportEverywhere` is enabled and by Closure when using `--chunk_output_type ES_MODULES`. So, this (un-)mangling is only applied for `vm.runInContext`, where `import.meta` is not supported.

`fix_js_mangling` is now only used under Closure in `MODULARIZE` mode, which is a nice win. Once https://github.com/google/closure-compiler/issues/3835 is fixed, we should be able to remove this Closure-specific (un-)mangling for top-level await entirely.

kleisauke:
Note to self: PR #26677 would make the Babel changes unnecessary.

src/jsifier.mjs
```diff
@@ -397,6 +397,12 @@ export async function runJSify(outputFile, symbolsOnly) {
   }
 
   async function writeOutput(str) {
+    // Unmangle previously mangled `import.meta` references.
+    // See also: `preprocess` in parseTools.mjs.
```
sbc100: The mangling was moved to `processMacros` and its now done by a function called `mangleUnsupportedSyntax` as of #27059.  Perhaps reference that new function name?

src/jsifier.mjs
```diff
@@ -397,6 +397,12 @@ export async function runJSify(outputFile, symbolsOnly) {
...
+    if (EXPORT_ES6) {
+      str = str.replaceAll('EMSCRIPTEN$IMPORT$META', 'import.meta');
+    }
```
sbc100, line 404: Perhaps mention why we are doing this?  

Something like "Unmangle import.meta which was only mangled to allow the code to flow through vm.runInContext".

tools/link.py
```diff
@@ -2073,19 +2073,28 @@ def phase_source_transforms(options):
   save_intermediate('transformed')
 
 
-# Unmangle previously mangled `import.meta` and `await import` references in
+# Unmangle previously mangled `await import` and `await` references in
 # both main code and libraries.
-# See also: `preprocess` in parseTools.js.
+# See also: `preprocess` in parseTools.mjs.
```
sbc100: `preprocess` -> `mangleUnsupportedSyntax`

tools/link.py
```diff
@@ -2073,19 +2073,28 @@ def phase_source_transforms(options):
...
-             .replace('EMSCRIPTEN$AWAIT(', 'await ('))
+
+  if settings.EXPORT_ES6:
+    # Also remove the line containing `export{};`, which is inserted by
+    # Closure to mark the file as an ES6 module.
+    # https://github.com/google/closure-compiler/issues/4084#issuecomment-1505056519
+    # https://github.com/google/closure-compiler/blob/v20260401/src/com/google/javascript/jscomp/ConvertChunksToESModules.java#L111-L113
+    src = src \
+      .replace('EMSCRIPTEN$AWAIT$IMPORT', 'await import') \
+      .replace('export{};\n', '')
```
sbc100, line 2093: Should we at `^` to the start of the `export{}` line to be more precise?

sbc100 (approved):

sbc100:
I'm excited to see this land!

src/jsifier.mjs
kleisauke: Goot catch! I forgot to do this within the merge commit, fixed with 10e638306bfb5b19ad0050f0328de1ac854e2af8.

tools/link.py
kleisauke: 10e638306bfb5b19ad0050f0328de1ac854e2af8

src/jsifier.mjs
kleisauke: This is already mentioned in the comment that is referenced.
https://github.com/emscripten-core/emscripten/blob/10e638306bfb5b19ad0050f0328de1ac854e2af8/src/parseTools.mjs#L38-L43

tools/link.py
kleisauke: That will require use of `re.sub`, I thought this was simpler.

sbc100: Fair enough

sbc100 (approved):

Issue #23732, opened 2025-02-24
Clearer linker error due to sjlj ABI mismatch?


hoodmane:
Source file:
```a.c
#include <setjmp.h>

jmp_buf buf;

int main() {
    longjmp(buf, 8);
}
```
Compile/link:
```sh
$ emcc -c a.c
$ emcc a.o -fwasm-exceptions
error: undefined symbol: emscripten_longjmp (referenced by root reference (e.g. compiled C/C++ code
```
Would be nice if the error could say something roughly like:
```
a.o uses setjmp/longjmp but was compiled with incompatible ABI. Pass -fwasm-exceptions as a compile flag.
```

anutosh491:
Thanks for the report. Can try looking into this soon !

sbc100:
I created an llvm-side fix that (llvm/llvm-project#128564) as well as an emscripten-side fix that should help make this more clear #23738.

hoodmane:
Thanks for the quick fix @sbc100!

PR #23738, opened 2025-02-24
Add better error message when mixing exceptions modes. NFC

sbc100:
Depends on https://github.com/llvm/llvm-project/pull/128564 to allow wasm-ld to report the object file in question.

Fixes: #23732

### DIFF
--- a/src/lib/libexceptions.js
+++ b/src/lib/libexceptions.js
@@ -257,6 +257,7 @@ var LibraryExceptions = {
   },
 
 #endif
+
 #if WASM_EXCEPTIONS || !DISABLE_EXCEPTION_CATCHING
   $getExceptionMessageCommon__deps: ['__get_exception_message', 'free', '$stackSave', '$stackRestore', '$stackAlloc'],
   $getExceptionMessageCommon: (ptr) => {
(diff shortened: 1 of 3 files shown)


aheejin (approved):

src/lib/libexceptions.js
```diff
@@ -277,17 +278,23 @@ var LibraryExceptions = {
     return [type, message];
   },
 #endif
+
 #if WASM_EXCEPTIONS
+#if ASSERTIONS
+  // Try to give a more useful error message than simply `Undefined reference to `emscripten_longjmp`
+  emscripten_longjmp__deps: [() => error('undefined reference to `emscripten_longjmp`. One or more object files was not compiled with `-fwasm-exceptions`.  Build with `-sASSERTIONS=0` to have wasm-ld report which one.')],
```
hoodmane, line 285: Just to confirm, `-sASSERTIONS=0` is the default right?

test/test_other.py
```diff
@@ -9438,6 +9438,22 @@ def test_exceptions_c_linker(self):
...
+      jmp_buf buf;
+      longjmp(buf, 8);
+    }
+    ''')
+    self.emcc('main.c', ['-c'])
+    err = self.expect_fail([EMCC, 'main.o', '-fwasm-exceptions'])
+    self.assertContained('error: undefined reference to `emscripten_longjmp`. One or more object files was not compiled with `-fwasm-exceptions`.  Build with `-sASSERTIONS=0` to have wasm-ld report which one.', err)
+
+    err = self.expect_fail([EMCC, 'main.o', '-fwasm-exceptions', '-sASSERTIONS=0'])
+    self.assertContained('wasm-ld: error: main.o: undefined symbol: emscripten_longjmp', err)
```
hoodmane, line 9454: Looks like a big improvement to me.

When Emscripten invokes wasm-ld it would be nice if it could detect this error and add the extra context about this since one doesn't see the instruction to pass `-fwasm-exceptions` as a cflag here. I assume we are already redirecting stderr when we invoke `wasm-ld`?

If it's too inconvenient this by itself is quite helpful.

src/lib/libexceptions.js
sbc100: `-sASSERTIONS=1` is the default for `-O0` but `-sASSERTIONS=0` is the default of `-O1` and above.

test/test_other.py
sbc100: Emscripten doesn't currently capture and parse the output of wasm-ld.  Its just fails if/when wasm-ld fails.

I'm not sure I want to go down the route of trying to get fancy with parsing the error output of wasm-ld.    I decided that showing the more specific error when `-O0` / `-sASSERTIONS=1` are enabled was a good compromise.

sbc100:
> What's the procedure for an emscripten change that requires an llvm change? You test emscripten locally against the llvm change and then wait for the llvm commit to land and then wait for the roller to ingest it and then rebase the emscripten pr?

Yes, exactly.

Issue #18112, opened 2022-10-28
WasmFS: Current directory ignores custom backend


mere-human:
getcwd() returns empty string for directories that use custom backends.

**Version of emscripten/emsdk:**
```
emcc -v
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.25-git (8647224628dfc10dec6e5c6f3621d625a5f16eda)
clang version 16.0.0 (https://github.com/llvm/llvm-project 2234f582c91e8511443f3e71b6e06abe3e47db4b)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /Users/user/src/emsdk/upstream/bin
```

**Failing code:**
```cpp
#include <cassert>
#include <dirent.h>
#include <emscripten/wasmfs.h>
#include <stdio.h>
#include <sys/stat.h>
#include <unistd.h>
#include <cerrno>

int main() {
  const char *dir_path = "/subdir";

#ifdef WASMFS
  const char *url = "http://localhost:8080";
  printf("url: %s\n", url);
  backend_t backend = wasmfs_create_fetch_backend(url);
  assert(backend);
  int res = wasmfs_create_directory(dir_path, 0777, backend);
  if (errno) perror("wasmfs_create_directory");
  assert(errno == 0);
  assert(res == 0);
#else
  int res = mkdir(dir_path, 0777);
  if (errno) perror("mkdir");
  assert(errno == 0);
  assert(res == 0);
#endif

  char buf[100];
  assert(getcwd(buf, sizeof(buf)));
  printf("cwd: %s\n", buf);

  printf("chdir: %s\n", dir_path);
  chdir(dir_path);

  assert(getcwd(buf, sizeof(buf)));
  printf("cwd: %s\n", buf);

  return 0;
}
```

**Full link command and output with `-v` appended:**
```sh
EXE=common
rm -rf build
mkdir -p build
FLAG_WASMFS="-DWASMFS -sWASMFS=1"
em++ -v -g $(echo $FLAG_WASMFS) -pthread -sPTHREAD_POOL_SIZE=2 -std=c++11 main.cpp -o build/$EXE.html
ret=$?
if [ $ret -eq 0 ]; then
    echo "build ok"
    emrun build/$EXE.html --browser chrome --port 8080
fi;
```

**Build output:**
<details>
  <summary>Click to expand</summary>

  ```
 "/Users/user/src/emsdk/upstream/bin/clang" --version
[...]

PR #23756, opened 2025-02-25
Fix retrieval of mounted directory names

guusw:
As mentioned in #18112 I decided to PR my changes since I would like to use upstream emscripten for development where possible.

Basically retrieval of the CWD for mounted directories (e.g. fetchfs) will not return the correct name for the mounted directory, for example `//` instead of `/http_fs/`


sbc100:
Can you add a test for this?   That might also help me understand the problem we are solving here.

guusw:
Added a test, seems like it can be reproduced with just memory backend too

4.0.1:
```
WasmFS Mounted Directory CWD Test
Initial CWD: /
CWD after chdir("/tfs"): /
Aborted(Assertion failed: strcmp(buf, "/tfs") == 0, at: /Users/bakje.coffee/Projects/emscripten/test/wasmfs/wasmfs_mounted_cwd.c,36,main)
/Users/bakje.coffee/Projects/emscripten/out/test/wasmfs_mounted_cwd.js:684
```
After PR works fine

sbc100:
I recently updated `test_unistd_unlink` in `test_core.py` such this it skips under wasmfs due to this bug.. can you remove the `skipTest` there as part of this PR?

tlively:
Part of the documented requirements for a backend opting into `maintainsFileIdentity` are that the backend needs to implement `getName` itself:

https://github.com/emscripten-core/emscripten/blob/340aa0f451c1680c3a0dfbe65a42651c8d187307/system/lib/wasmfs/file.h#L254-L263

It sounds like maybe we have backends that aren't doing this correctly? If so, it would be better to fix the backends to make sure the implementation is consistent with the documented intent.

guusw:
> It sounds like maybe we have backends that aren't doing this correctly? If so, it would be better to fix the backends to make sure the implementation is consistent with the documented intent.

Yeah memory backend does this:
```
std::string MemoryDirectory::getName(std::shared_ptr<File> file) {
  auto it =
    std::find_if(entries.begin(), entries.end(), [&](const auto& entry) {
      return entry.child == file;
    });
  if (it != entries.end()) {
    return it->name;
  }
  return "";
}
```
But the entries don't seem to contain the file that is requested in these paths, the mounted folder itself I assume?

PR #23759, opened 2025-02-25
Use JS spread operator over Object.assign. NFC

sbc100:
This saves a little bit of code size.


kripken (approved):
Nice... I guess that is the shortest code to do a shallow copy of an object.

PR #23760, opened 2025-02-25
Cleanup event loop tests. NFC

sbc100:
The browser tests were only being run in PROXY_TO_PTHREAD mode for some reason even though these tests should pass on the main thread.

In test_other the `PROXY_TO_PTHREAD` flags were being passed as args (rather then emcc_arggs).  Likely a copy/paste bug.


src/lib/libeventloop.js
```diff
@@ -21,6 +21,13 @@ LibraryJSEventLoop = {
     }, timeout);
   },
 
+  $safeClearTimeout: (id) => {
+    // TODO: Track outstanding timeout IDs so that we can avoid calling
+    // runtimeKeepalivePop trice for the same timeout.
```
kripken, line 26:
```suggestion
    // runtimeKeepalivePop twice for the same timeout.
```

src/lib/libeventloop.js
```diff
@@ -21,6 +21,13 @@ LibraryJSEventLoop = {
...
+    {{{ runtimeKeepalivePop() }}}
```
kripken, line 27: Without this tracking, can't we consider the program to have exited at an earlier time than it did? That sounds more dangerous than the state before this PR IIANM?

sbc100: Yes, but only programs that call `emscripten_clear_timeout` with the ID of a timeout that has already fired.

Without this change it is not possible for programs that clear timeouts to exit.

With this change programs that clear invalid timeout ID might exit earlier than they expect.

sbc100: Perhaps I'll open a bug and make this change NFC.

kripken: That sgtm. Early exits can be pretty dangerous, worse than not being able to exit, I'd say.

sbc100: Reverting this change and making this a test-only NFC

test/test_browser.py
```diff
@@ -1977,13 +1976,10 @@ def test_sdl_glshader2(self):
...
-  @parameterized({
-    '': ([],),
-    'pthreads': (['-pthread', '-sPROXY_TO_PTHREAD', '-sOFFSCREEN_FRAMEBUFFER'],),
-  })
   @requires_graphics_hardware
-  def test_gl_textures(self, args):
-    self.btest_exit('gl_textures.c', args=['-lGL', '-g', '-sSTACK_SIZE=1MB'] + args)
+  @also_with_proxy_to_pthread
+  def test_gl_textures(self):
+    self.btest_exit('gl_textures.c', args=['-lGL', '-g', '-sSTACK_SIZE=1MB', '-sOFFSCREEN_FRAMEBUFFER'])
```
kripken, line 1982: This did not always test with OFFSCREEN_FRAMEBUFFER before. Do we not error on that being used without pthreads?

test/test_other.py
```diff
@@ -13883,9 +13883,8 @@ def test_pthread_trap(self):
   def test_pthread_kill(self):
     self.do_run_in_out_file_test('pthread/test_pthread_kill.c')
 
-  @node_pthreads
   def test_emscripten_set_interval(self):
-    self.do_runf('emscripten_set_interval.c', args=['-pthread', '-sPROXY_TO_PTHREAD'])
+    self.do_runf('emscripten_set_interval.c')
```
kripken, line 13887: Previously this was tested with pthreads, but now it isn't, if I read this diff correctly?

test/test_browser.py
sbc100: No, I think think its harmlessly ignored.

sbc100: Maybe we should error ?  But we don't today.

test/test_other.py
kripken: I see. Should we fix the typo and test it with pthreads, or not?

kripken: I can't think of a reason to use pthreads on this myself. lgtm either way.

kripken (approved):

test/test_other.py
sbc100: We test with and without threads in the browser already.

PR #23761, opened 2025-02-25
Remove launcher scripts from git

sbc100:
For a while now we have been mandating that bootstrap is run before emscripten can be used.  This change removes most of the launcher scripts from git (since we can rely on bootstrap to create them).

This helps clean up the top level directory.

I've left emcc and em++ launchers in source control.  If anyone runs these without first running bootstrap it will ask them to first run bootstrap.


dschuff (approved):

sbc100:
I think I'm going to merge this with the larger change to of moving the entry points to `/bin`

PR #23767, opened 2025-02-26
Add support for TRUSTED_TYPES when using Wasm Workers

jriv01:
This change allows usage of Trusted Types compatible wrappers for Wasm Workers when `TRUSTED_TYPES` is defined.

Similar behavior has already been implemented in [libpthread.js](https://github.com/emscripten-core/emscripten/blob/main/src/lib/libpthread.js#L426), so the changes here are based off of & exhibit the same logic used for Trusted Types & pthread compatibility.


src/lib/libwasm_worker.js
```diff
@@ -170,21 +170,36 @@ if (ENVIRONMENT_IS_WASM_WORKER
...
+        {
+          createScriptURL: (ignored) => ww_js
+        }
+      );
+      worker = _wasmWorkers[_wasmWorkersID] = new Worker(p.createScriptURL('ignored'));
+    } else {
+      worker = _wasmWorkers[_wasmWorkersID] = new Worker(ww_js);
+    }
+#else
+    worker = _wasmWorkers[_wasmWorkersID] = new Worker(ww_js);
```
sbc100, line 201: I think bundlers cannot deal will with `new Worker` calls that don't have literal arguments.     

Perhaps you can do a trick like with `pthreadWorkerOptions` in `libpthread.js`

src/wasm_worker.js
```diff
@@ -58,6 +58,18 @@ if (ENVIRONMENT_IS_NODE) {
...
+#endif
+#if TRUSTED_TYPES
+  // Use Trusted Types compatible wrappers.
+  if (typeof trustedTypes != 'undefined' && trustedTypes.createPolicy) {
+    var p = trustedTypes.createPolicy(
+      'emscripten#scriptPolicy1',
+      {
+        createScriptURL: (ignored) => d.js
+      }
+    );
```
sbc100, line 70: Cam this all be on single line?

sbc100 (commented):
I guess we need some kind of test for this?

kripken:
> I guess we need some kind of test for this?

We do have an existing test for TRUSTED_TYPES in `test/test_other.py`. Can perhaps add a variation to one of the existing `WASM_WORKERS` tests there, that adds this.

src/lib/libwasm_worker.js
jriv01: Are nested macros unsupported? I tried a similar trick as `pthreadWorkerOptions` and it results in 
```
  locateFile('WASM_WORKER_FILE'

SyntaxError: Unexpected end of input
```

I was trying something like this:
```
{{{
  const wasmWorkerJs = `
#if WASM_WORKERS == 2
    _wasmWorkerBlobUrl
#elif MINIMAL_RUNTIME
    Module['$wb']
#else
    locateFile('{{{ WASM_WORKER_FILE }}}')
#endif
`;
  null;
}}}
```

jriv01: I think each set of opening brackets is matched to the closest set of closing brackets, based on [parseTools.mjs](https://github.com/emscripten-core/emscripten/blob/main/src/parseTools.mjs#L36). So the opening bracket of the outer macro is matching with the closing bracket of the nested macro & resulting in that error. 

Is there something extra I should be doing to nest these? Or another way to pass literals to `new Worker()`? I can do it by adding more preprocessor checks for both cases of `TRUSTED_TYPES` being defined or not but it's not the prettiest solution

sbc100: Actually looking at libpthread it looks like it avoids repeating itself by doing it like this: https://github.com/emscripten-core/emscripten/blob/961c40e58ba752b504f343e63d2e039f5b0e3a3c/src/lib/libpthread.js#L422-L438

sbc100: > I think each set of opening brackets is matched to the closest set of closing brackets, based on [parseTools.mjs](https://github.com/emscripten-core/emscripten/blob/main/src/parseTools.mjs#L36). So the opening bracket of the outer macro is matching with the closing bracket of the nested macro & resulting in that error.
> 
> Is there something extra I should be doing to nest these? Or another way to pass literals to `new Worker()`? I can do it by adding more preprocessor checks for both cases of `TRUSTED_TYPES` being defined or not but it's not the prettiest solution

Ah yes, I forgot to say, you don't need the extra `{{{` `}}}` when you are already inside a `{{{` block.  You can just reference `TARGET_JS_NAME` as a normal variable.

See: https://github.com/emscripten-core/emscripten/compare/main...sbc100:emscripten:example?expand=1

jriv01: I see, that seems to work. Thanks!

src/wasm_worker.js
jriv01: Done.

src/lib/libwasm_worker.js
```diff
@@ -170,21 +179,30 @@ if (ENVIRONMENT_IS_WASM_WORKER
...
+    // MINIMAL_RUNTIME has a structure where the .ww.js file is loaded from
+    // the main HTML file in parallel to all other files for best performance
+    let ww_js = Module['$wb']; // $wb="Wasm worker Blob", abbreviated since not DCEable
 #else
-      // default runtime loads the .ww.js file on demand.
-      locateFile('{{{ WASM_WORKER_FILE }}}')
-#endif
-    );
+    // default runtime loads the .ww.js file on demand.
+    let ww_js = locateFile('{{{ WASM_WORKER_FILE }}}');
```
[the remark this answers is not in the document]
jriv01: Thanks for pointing that out! Missed it

sbc100 (approved):

jriv01:
> > I guess we need some kind of test for this?
> 
> We do have an existing test for TRUSTED_TYPES in `test/test_other.py`. Can perhaps add a variation to one of the existing `WASM_WORKERS` tests there, that adds this.

When you say add a variation that matches the existing `TRUSTED_TYPES` test, did you mean using `@paramaterized`?

https://github.com/emscripten-core/emscripten/blob/88ffcb9898f606cce067773fc924241b0471a933/test/test_other.py#L13161-L13165

So something like
```
   @parameterized({ 
     '': [[]], 
     'trusted': [['-sTRUSTED_TYPES']] 
   }) 
   def test_wasm_worker_hello(self):
     ...
```

sbc100:
Yup that looks about right.

sbc100 (approved):

jriv01:
It seems like there's still a CI failing, any idea what may be causing that?

I see it says `Test expectations are out-of-date on the PR branch. You can run './tools/maint/rebaseline_tests.py' to create a commit updating the expectations.` but I already did that at commit c6305cf

sbc100:
I think maybe you should try reverting `test/other/codesize/test_codesize_files_wasmfs.size`.  I don't see how this change could have changed the size of that test.

sbc100 (approved):

jriv01:
There seems to be some new failures despite not changing anything since 43a04dc.

Any idea what might be happening here? I see that f391a0d addresses the `test_fs_js_api_wasmfs` failure under `test-wasm64-4gb` so syncing this branch should fix that, but there's two separate failing tests in `test-wasm64`.

PR #23773, opened 2025-02-26
[WasmFS] Limit max file descriptor number

sbc100:
This enabled wasmfs.test_unistd_dup to pass


kripken (approved):
This looks right, and it enables the test to pass, but it makes WasmFS less capable than before with the new limit.

I guess the limit might be fine, as it's higher than Linux's?

sbc100:
We can always up the limit or make it configurable via getrlimit/setrlimit.

sbc100:
According to AI: "The default RLIMIT_NOFILE resource limit for userspace processes on Linux is 1024 (soft) and 4096 (hard). This limit specifies the maximum number of file descriptors that a process can open. "

PR #23781, opened 2025-02-27
Make 'instantiateWasm' callback resolve the result return from 'recei…

gogogoghost:
In js output content. 

When ASYNCIFY enabled, wasmExports modified by **Asyncify.instrumentWasmExports**. [source](https://github.com/emscripten-core/emscripten/blob/4b7653130284439898345419f2990c1e14fa60a1/src/preamble.js#L871)

But when **Module['instantiateWasm']** used. The **createWasm** function will always resolve the original wasmExports. [source](https://github.com/emscripten-core/emscripten/blob/4b7653130284439898345419f2990c1e14fa60a1/src/preamble.js#L994).

It will case crash when async C++ function rewind.

So I make 'instantiateWasm' callback resolve the result return from 'receiveInstance'.

### DIFF
--- a/src/preamble.js
+++ b/src/preamble.js
@@ -990,8 +990,7 @@ function getWasmImports() {
       try {
 #endif
         Module['instantiateWasm'](info, (mod, inst) => {
-          receiveInstance(mod, inst);
-          resolve(mod.exports);
+          resolve(receiveInstance(mod, inst));
         });
 #if ASSERTIONS
       } catch(e) {


sbc100 (commented):
Tanks for the fix!

Do you think we can update one of the existing tests to cover this?   Maybe `test_instantiate_wasm` in `test_other.py`

gogogoghost:
I have just updated the test unit. And this issues only affects when EXPORT_ES6 enabled.

When EXPORT_ES6 disabled. **createWasm** called after **wasmExports** defined.
```js
var wasmExports;
createWasm();
```
**wasmExports** can be updated in createWasm.

When EXPORT_ES6 enabled. **wasmExports** assigned by the return value of **createWasm**.
```js
var wasmExports = await createWasm();
```

test/test_manual_wasm_instantiate.c
```diff
@@ -6,10 +6,15 @@
...
+});
 
 int main() {
-  printf("in main\n");
-  int result = EM_ASM_INT({return Module.testWasmInstantiationSucceeded;});
-  assert(result);
+  printf("main: before getResult\n");
+  int result = getResult();
+  printf("testWasmInstantiationSucceeded: %d\n",result);
   return 0;
```
sbc100, line 19: Are the changes in this file necessary?  It looks like you could revert this file probably.

test/test_other.py
```diff
@@ -15688,7 +15688,12 @@ def test_instantiate_wasm(self):
         });
         return {}; // Compiling asynchronously, no exports.
       }''')
-    self.do_runf('test_manual_wasm_instantiate.c', emcc_args=['--pre-js=pre.js'])
+    self.run_process([EMCC, test_file('test_manual_wasm_instantiate.c'), '--pre-js=pre.js','-sASYNCIFY','-sEXPORT_ES6=1','-o','manual_init_module.mjs'])
```
sbc100, line 15691: I don't think you need to add '-sASYNCIFY' here do you?

sbc100, line 15691: This test is already decorated with `also_with_modularize` so it should already run both with and without `-sMODULARIZE`.  

Do this bug depened on `-sEXPORT_ES6` , no just `-sMODULARIZE` somehow?

gogogoghost: Did I misunderstand it? I need to add a new test unit to cover this situation not edit existing one.

This bug just depened on **-sASYNCIFY** and **-sEXPORT_ES6**.

sbc100: Ah I see.. the bug only occurs with `-sEXPORT_ES6`.

I don't see how it depends on `-sASYNCIFY` though, maybe that flag can be removed?

sbc100: Sorry, I think I understand now.   `-sASYNCIFY` is needed because in that mode we wrap the wasm exports.

However, I think `-sMODULARIZE` should be enough (no need for `-sEXPORT_ES6`).   So I think if you run `test_instantiate_wasm_modularize` (the modularized version) with `-sASYNCIFY` that should be enough to repro the issue without creating `manual_init.mjs`.

Sorry for the confusion.

gogogoghost: [source](https://github.com/emscripten-core/emscripten/blob/4b7653130284439898345419f2990c1e14fa60a1/src/preamble.js#L994) here.
```js
Module['instantiateWasm'](info, (mod, inst) => {
    // When -sASYNCIFY. This function will wrap the original exports and return.
    receiveInstance(mod, inst);
    // But here always resolve the original exports.
    resolve(mod.exports);
  });
```
And when **EXPORT_ES6** enabled. wasmExports assigned by the return value of createWasm.
```js
var wasmExports = await createWasm();
```
So the wasmExports will always be the original exports object. Not the wrapped one.
It will case crash when a C++ function rewind.
If **-sASYNCIFY** removed. It will always success. because:
```js
var wasmExports;
// return value not used.
createWasm();
```
And [source](https://github.com/emscripten-core/emscripten/blob/4b7653130284439898345419f2990c1e14fa60a1/src/preamble.js#L871) here updated the global **wasmExports**.

test/test_other.py
```diff
@@ -15688,7 +15688,12 @@ def test_instantiate_wasm(self):
         });
         return {}; // Compiling asynchronously, no exports.
       }''')
```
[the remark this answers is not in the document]
gogogoghost: Sorry for my time zone I should go to bed, I will do this tomorrow

test/test_other.py
sbc100: > And when **EXPORT_ES6** enabled. wasmExports assigned by the return value of createWasm.

But this is also true for just `-sMODULARIZE`.   i.e. i think this bug will show up with either `-sMODULARIZE` or `-sEXPORT_ES6`.. so it should be enough to test with the existing `test_instantiate_wasm_modularize`

test/test_other.py
```diff
@@ -15688,7 +15688,9 @@ def test_instantiate_wasm(self):
...
-    self.do_runf('test_manual_wasm_instantiate.c', emcc_args=['--pre-js=pre.js'])
+    # Test with ASYNCIFY here to ensure that that wasmExports gets set to the wrapped version of the wasm exports.
```
sbc100, line 15691: `modularize_post_js.js` is automatically added by `@also_with_modularize` so I don't think you need to add it here..

You also don't need to specify the output filename and you should also be able to continue to use `self.do_runf` instead of splitting this into two lines like this.

sbc100: Basically, I think all you need add is `-sASYNCIFY`

gogogoghost: I figured out how to test the **also_with_modularize** function. Thanks for your patience

test/test_other.py
```diff
@@ -15705,7 +15705,8 @@ def test_instantiate_wasm(self):
...
+    self.do_runf(test_file('test_manual_wasm_instantiate.c'),emcc_args=['--pre-js=pre.js','-sASYNCIFY'])
```
sbc100, line 15709: Space after comms here

And remove the `test_file` function call

Otherwise LGTM now!

sbc100:
The test failures look like they are because `test_browser.py` run the same test and needs `-sASYNCIFY` too.

gogogoghost:
> The test failures look like they are because `test_browser.py` run the same test and needs `-sASYNCIFY` too.

When **-sASYNCIFY** added into **test_manual_wasm_instantiate** in **test_browser.py**.

 Then main function in **test_manual_wasm_instantiate.c** have to call **REPORT_RESULT(0);** and change **reporting=Reporting.JS_ONLY** to **reporting=Reporting.FULL** to report the result.

But **other.test_instantiate_wasm** doesn't link **report_result.c** that will case compile failed.

Please give me some advice. Do I need to create a separate .c file for **browser.test_manual_wasm_instantiate**?

caiiiycuk:
Hi, what is the status of this PR?

caiiiycuk:
Maybe adding a new test for this specific case wouldn’t be such a bad idea?

sbc100:
I think its almost ready.. just needs the tests to pass.

test/test_browser.py
```diff
@@ -4227,9 +4227,9 @@ def test_async_compile(self, opts, returncode):
   @also_with_asan
   def test_manual_wasm_instantiate(self):
     self.set_setting('EXIT_RUNTIME')
-    self.compile_btest('test_manual_wasm_instantiate.c', ['-o', 'manual_wasm_instantiate.js'], reporting=Reporting.JS_ONLY)
+    self.compile_btest('test_manual_wasm_instantiate.c', ['-o', 'manual_wasm_instantiate.js','-sASYNCIFY'])
     shutil.copy(test_file('test_manual_wasm_instantiate.html'), '.')
-    self.run_browser('test_manual_wasm_instantiate.html', '/report_result?exit:0')
+    self.run_browser('test_manual_wasm_instantiate.html', '/report_result?1')
```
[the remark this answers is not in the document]
gogogoghost: Because **test_manual_wasm_instantiate.c** using ASYNCIFY. So I think there has 3 ways to fix it:

1、Add **-sASYNCIFY** to **browser.test_manual_wasm_instantiate**. And use REPORT_RESULT macro to report result. (committed)

2、Modify **test_manual_wasm_instantiate.c** using macro to let it run with ASYNCIFY for **other.test_instantiate_wasm** and without ASYNCIFY for **browser.test_manual_wasm_instantiate**.

3、Create a new .c file for **browser.test_manual_wasm_instantiate**. But the code is too less. I think the first two are better.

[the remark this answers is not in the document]
sbc100: I'm actually curious about (1) too.  Why does adding `-sASYNCIFY` to  `browser.test_manual_wasm_instantiate` lead to needing to use `REPORT_RESULT`?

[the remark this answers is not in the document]
gogogoghost: > I'm actually curious about (1) too. Why does adding `-sASYNCIFY` to `browser.test_manual_wasm_instantiate` lead to needing to use `REPORT_RESULT`?

Calling async function from main will color main to be a async function. Then when main function exit. **runtimeKeepaliveCounter** will be 1 now. So **onExit** will not execute. The test unit will wait for the exit message until timeout. So I use **REPORT_RESULT**.

So the way (2) is modifying **test_manual_wasm_instantiate.c** to let main function using **ASM_JS_INT** instead of **EM_ASYNC_JS** to avoid this problem.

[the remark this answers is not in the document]
gogogoghost: But **other.test_instantiate_wasm_modularize** needs wasm call async function to test this bug. So I made the way (2) to  use macro to let wasm call **EM_ASYNC_JS** for **other.test_instantiate_wasm_modularize**. And call **ASM_JS_INT** for **browser.test_manual_wasm_instantiate**.

sbc100 (approved):

PR #23787, opened 2025-02-27
Avoid double-reporting the filename in JS library errors/warnings

sbc100:
As part of this I also avoid converting JS libraries path to absolute paths where possible.  This means that errors and warnings show the actual pathname passed by the users, and not some absolute path.  This makes writing tests for the double-reporting issue easier, and makes the errors/warnings easier to read (IMHO).


tools/link.py
```diff
@@ -218,7 +218,9 @@ def get_js_sym_info():
   input_files.extend(read_file(jslib) for jslib in sorted(jslibs))
   for jslib in settings.JS_LIBRARIES:
     if not os.path.isabs(jslib):
-      jslib = utils.path_from_root('src/lib', jslib)
+      system_lib = utils.path_from_root('src/lib', jslib)
+      if os.path.exists(system_lib):
+        jslib = system_lib
```
kripken, line 223: Does this condition mean that a js library with the same basename as a system one will behave differently than before?

sbc100: I think this actually fixes a bug here the lookup used by `get_js_sym_info` was different to the look used in `src/modules.mjs`:

The two codepaths now have the same logic, which is good.

sbc100: The JS always had `(!path.isAbsolute(filename) && fs.existsSync(path.join(systemLibdir, filename))` as the condition.

sbc100: Oh man, that would be a very subtle bug to write a fix for.    I can give it a go but I'm not sure its worth it.

sbc100: Actually I'm not sure its possible since all `--js-library` args were processed as `os.path.abspath` below, so I guess we can consider it an inconsistency/bug that was not previously exposed in any way.

sbc100: But maybe there is an edge case where for when the user specifies the name of a js-library that happens to match a system library.   I'll write test for that.

kripken: If it's complex to test than I agree it's not worth it. I was hoping it was simple...

sbc100:
OK, this PR is now much simpler since #23798 landed separately.

dschuff (approved):

PR #10457, opened 2020-02-15
Adjust to Closure compiler convention that parseInt() must always be …

juj:
…called with two input parameters.

Closure compiler has their own convention that the second radix argument to parseInt() is not optional, but then unfortunately Closure compiler also does not optimize out the second argument when one does pass it.

See https://github.com/google/closure-compiler/issues/3230 and https://github.com/google/closure-compiler/issues/3548

This PR adds two new functions jstoi_s and jstoi_q, similar to spirit with the C "atoi" function.

jstoi_s is not actually used, but I think it's good to have for completeness.

PR #23788, opened 2025-02-27
Remove unused JS library function: `jstoi_s`. NFC

sbc100:
This function was also mis-documented.  Its was added #10457 but even then it was unused only add because "good to have for completeness".

However, the behavior of the JS `Number` function is not as described in the comment.  Instead `Number("123abc")` returns NaN.

On balance I think its better to just remove it.


sbc100:
Closing a favor of larger change https://github.com/emscripten-core/emscripten/pull/23789

PR #23789, opened 2025-02-27
Remove usage of `jstoi_q/jstoi_s` JS library functions. NFC

sbc100:
The `jstoi_q` was introduced in #10457 because of issues with using `parseInt` directly and closure compiler.  However, its seems like simply using `Number` is smaller and simpler.  See
https://github.com/google/closure-compiler/issues/3230

The `jstoi_s` function was unused and documented incorrectly (`Number` never actually throws and always returns NaN)


kripken:
The API is different, though, IIANM?
```js
> parseInt('42foo')
42
> Number('42foo')
NaN
```
Do we never use it with prefixes like that?

sbc100:
> The API is different, though, IIANM?
> 
> ```js
> > parseInt('42foo')
> 42
> > Number('42foo')
> NaN
> ```
> 
> Do we never use it with prefixes like that?

I would hope not, but I will audit all the call sites as best I can.

sbc100:
Hmm, looks like we at least some of the libwebgl.js cases depend on that difference.

I'll split this up.

PR #23790, opened 2025-02-28
Remove no-longer-needed jstoi_q helper function. NFC

sbc100:
Less risky version #23789.  It seems that the original issue we had with closure compile is no longer a problem and we can simply use parseInt with a single argument now.

See  #10457 and https://github.com/google/closure-compiler/issues/3230


sbc100:
It looks like radix parameter is now optional in closure compiler:
https://github.com/google/closure-compiler/blob/71785a7e3e04be489510d4695870c7491af17c3f/externs/es6.js#L1942-L1953

sbc100:
Oops, I misinterpreted the closure defintion.. this is still causing warnings.

PR #23791, opened 2025-02-28
Remove unnecessary usage of jstoi_q. NFC

sbc100:
The `jstoi_q` was introduced in #10457 because of issues with using `parseInt` directly and closure compiler.  However, in many cases its simpler to just use `Number`.  See https://github.com/google/closure-compiler/issues/3230

Split out from #23789.

### DIFF
--- a/src/lib/libtime.js
+++ b/src/lib/libtime.js
@@ -257,7 +257,7 @@ addToLibrary({
   },
 
   strptime__deps: ['$isLeapYear', '$arraySum', '$addDays', '$MONTH_DAYS_REGULAR', '$MONTH_DAYS_LEAP',
-                   '$jstoi_q', '$intArrayFromString' ],
+                   '$intArrayFromString' ],
   strptime: (buf, format, tm) => {
     // char *strptime(const char *restrict buf, const char *restrict format, struct tm *restrict tm);
     // http://pubs.opengroup.org/onlinepubs/009695399/functions/strptime.html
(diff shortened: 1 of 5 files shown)


src/lib/libhtml5.js
```diff
@@ -1347,8 +1347,8 @@ var LibraryHTML5 = {
     }
 
     if (inCenteredWithoutScalingFullscreenMode) {
-      var t = (innerHeight - jstoi_q(canvas.style.height)) / 2;
-      var b = (innerWidth - jstoi_q(canvas.style.width)) / 2;
+      var t = (innerHeight - Number(canvas.style.height)) / 2;
+      var b = (innerWidth - Number(canvas.style.width)) / 2;
```
kripken, line 1351: where in the docs does it say this must be a string with only an integer? i couldn't find that on MDN.

src/lib/libtime.js
```diff
@@ -362,21 +362,21 @@ addToLibrary({
 
       // seconds
       if ((value=getMatch('S'))) {
-        date.sec = jstoi_q(value);
+        date.sec = Number(value);
```
kripken, line 365: The code above is fairly complex - how can I see that this is a pure number without a prefix?

src/lib/libwebgl.js
```diff
@@ -3205,7 +3205,7 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
 #if GL_DEBUG
       console.dir(match);
 #endif
-      explicitUniformLocations[match[5]] = jstoi_q(match[1]);
+      explicitUniformLocations[match[5]] = Number(match[1]);
```
kripken, line 3208: Hmm, is the 5th capture not `\w+`, which accepts more than digits? Am I computing that wrong, or perhaps does the GL API guarantee an integer here?

src/lib/libtime.js
sbc100: The `getMatch` function returns that result of the regex's above.

You can see that we only call `Number` on the following matches: S, M, H, I, Y, y, C, m, d, j, U, W.

You can look at the above table of regexes and see that each of them only ever match numbers.

We also have pretty extensive testing for strftime I believe.

src/lib/libwebgl.js
sbc100: We are using the first capture here aren't we?   That looks like `(-?\d+)` to me

kripken: Oh, somehow my eye read the `match[5]` on the left...

kripken (approved):

PR #23794, opened 2025-02-28
Add adjustWasmImports incoming Module setting

hoodmane:
This provides a way to add symbols to the `wasmImports` dynamically. This doesn't allow anything new compared to `instantiateWasm` except that it doesn't interfere with sourcemaps or ASAN and it doesn't require reproducing the somewhat tricky behavior in the standard `instantiateWasm` function.

As an alternative, we could expose something like `instantiateWasmDefault` to the user, and then they could do this like:
```js
Module['instantiateWasm'] = (imports, successCallback) => {
   adjustWasmImports(imports);
   instantiateWasmDefault(imports, successCallback);
}
```

### DIFF
--- a/src/settings.js
+++ b/src/settings.js
@@ -999,7 +999,7 @@ var INCOMING_MODULE_JS_API = [
   'SDL_numSimultaneouslyQueuedBuffers', 'INITIAL_MEMORY', 'wasmMemory', 'arguments',
   'buffer', 'canvas', 'doNotCaptureKeyboard', 'dynamicLibraries',
   'elementPointerLock', 'extraStackTrace', 'forcedAspectRatio',
-  'instantiateWasm', 'keyboardListeningElement', 'freePreloadedMediaOnUse',
+  'instantiateWasm', 'adjustWasmImports', 'keyboardListeningElement', 'freePreloadedMediaOnUse',
   'loadSplitModule', 'locateFile', 'logReadFiles', 'mainScriptUrlOrBlob', 'mem',
   'monitorRunDependencies', 'noExitRuntime', 'noInitialRun', 'onAbort',
   'onCustomMessage', 'onExit', 'onFree', 'onFullScreen', 'onMalloc',
--- a/src/preamble.js
+++ b/src/preamble.js
@@ -977,6 +977,12 @@ function getWasmImports() {
 
   var info = getWasmImports();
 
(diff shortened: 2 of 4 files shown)


src/settings.js
```diff
@@ -999,7 +999,7 @@ var INCOMING_MODULE_JS_API = [
   'SDL_numSimultaneouslyQueuedBuffers', 'INITIAL_MEMORY', 'wasmMemory', 'arguments',
   'buffer', 'canvas', 'doNotCaptureKeyboard', 'dynamicLibraries',
   'elementPointerLock', 'extraStackTrace', 'forcedAspectRatio',
-  'instantiateWasm', 'keyboardListeningElement', 'freePreloadedMediaOnUse',
```
sbc100, line 1002: New imports should instead get added to `EXTRA_INCOMING_JS_API` in `emcc.py` to avoid adding bloat to all programs by default.

hoodmane:
I originally added this in order to make JSPI work with emscripten js exceptions by replacing the `invoke_` functions with runtime generated wasm trampolines that use wasm try/catch like in #23619. But just a few days ago I ran into someone else who needed to dynamically inject functions for some of their imports and wanted to know how to do it. It could also be helpful to wrap certain imports for debugging purposes. I think there are a lot of possible use cases.

I would say that it doesn't make much sense to restrict it to adding new imports because `instantiateWasm` doesn't have that restriction and the point of this is to make it possible to do the same things you can do with `instantiateWasm` but without having to duplicate the logic for getting the actual wasm module.

sbc100:
The main problem I have adding new hooks like this it add yet more combinations of settings that we need to test and maintains. 

I wonder if we could find some other possible ways to solve these problems.  Its kind of a shame `--pre-js` comes before `wasmImports` is created and `--post-js` comes after its is used.  Perhaps we could consider moving `--post-js` to between `wasmImports` creation and usage?   Then you could do what you need here simply with `--post-js` I think

hoodmane:
What about my idea of exposing `instantiateWasmDefault()`?

sbc100:
> What about my idea of exposing `instantiateWasmDefault()`?

If we could do it at no codesize code that seems reasonable.  I suppose we would need to pass it as an augment to instantiateWasm somehow?   (Because the `instantiateWasm` might be declared out side the scope of the generated code).

hoodmane:
Yeah passing it as an argument would make sense to me.

Issue #23769, opened 2025-02-26
Regression: mainScriptUrlOrBlob ignored when USE_ES6_IMPORT_META is off


JoeOsborn:
Per https://github.com/emscripten-core/emscripten/pull/23171#issuecomment-2686032660

The recent removal of USE_ES6_IMPORT_META broke a use case like this:

1. I want to host an emscripten-compiled package somewhere (e.g., a CDN) that is compiled to use pthreads
2. I want to instantiate a module from those scripts on another origin domain
3. Browsers refuse to download cross-origin workers, so I can't create threads.

This used to be supported using either an importScripts() hack in the worker or by downloading the whole script as a blob, but now that's not possible since the fix in #23171 forces the use of a specific file name.

JoeOsborn:
For a bit more detail:  Using a blob as the actual script means that import.meta.url is a blob, so new URL(...) can't work.  Using a mainScriptBlobOrUrl is just ignored.  So both of the tricks I've used in the past for cross origin workers are not available.

sbc100:
Thanks Joe!   We are obviously lacking some test coverage here.

JoeOsborn:
I think a reasonable fix for this is to:

1. If mainScriptUrlOrBlob is set, create the worker as in the old non-use-import-meta case (or else deprecate that field under export_es6 and assert that mainScriptUrlOrBlob is not set)
2. If import.meta.url is a blob url, use it instead of creating a new URL.

sbc100:
I believe this was fixed in #23890 and can be closed?

PR #23804, opened 2025-02-28
Support blob URLs on ES6 with USE_ES6_IMPORT_META enabled

JoeOsborn:
Per the discussion in #23769, implement this policy:

- If mainScriptUrlOrBlob is set, create the worker as in the old non-use-import-meta case
- If import.meta.url is a blob url, use it for the worker script
- Otherwise, use new URL(TARGET_JS_NAME, import.meta.url)

I know I didn't add any automated tests; I'd appreciate pointers on how to implement tests for this.

### DIFF
--- a/src/lib/libpthread.js
+++ b/src/lib/libpthread.js
@@ -424,12 +424,16 @@ var LibraryPThread = {
         worker = new Worker(p.createScriptURL('ignored'), {{{ pthreadWorkerOptions }}});
       } else
 #endif
-      // We need to generate the URL with import.meta.url as the base URL of the JS file
-      // instead of just using new URL(import.meta.url) because bundler's only recognize
-      // the first case in their bundling step. The latter ends up producing an invalid
-      // URL to import from the server (e.g., for webpack the file:// path).
-      // See https://github.com/webpack/webpack/issues/12638
-      worker = new Worker(new URL('{{{ TARGET_JS_NAME }}}', import.meta.url), {{{ pthreadWorkerOptions }}});
+#if expectToReceiveOnModule('mainScriptUrlOrBlob')
+        if (Module['mainScriptUrlOrBlob']) {
+          var pthreadMainJs = Module['mainScriptUrlOrBlob'];
+          if (typeof pthreadMainJs != 'string') {
+            pthreadMainJs = URL.createObjectURL(pthreadMainJs);
+          }
+          worker = new Worker(new URL(pthreadMainJs, import.meta.url), {{{ pthreadWorkerOptions }}});
+        } else
+#endif
+          worker = new Worker(import.meta.url, {{{ pthreadWorkerOptions }}});
 #else // EXPORT_ES6
       var pthreadMainJs = _scriptName;
 #if expectToReceiveOnModule('mainScriptUrlOrBlob')


src/lib/libpthread.js
```diff
@@ -435,11 +435,29 @@ var LibraryPThread = {
...
-      // instead of just using new URL(import.meta.url) because bundler's only recognize
-      // the first case in their bundling step. The latter ends up producing an invalid
-      // URL to import from the server (e.g., for webpack the file:// path).
-      worker = new Worker(new URL('{{{ TARGET_JS_NAME }}}', import.meta.url), {{{ pthreadWorkerOptions }}});
+      {
+#if expectToReceiveOnModule('mainScriptUrlOrBlob')
+        if (Module['mainScriptUrlOrBlob']) {
+          var pthreadMainJs = Module['mainScriptUrlOrBlob'];
+          if (typeof pthreadMainJs != 'string') {
+            pthreadMainJs = URL.createObjectURL(pthreadMainJs);
```
sbc100, line 443: It looks like `pthreadMainJs` it not being used?

src/lib/libpthread.js
```diff
@@ -435,11 +435,29 @@ var LibraryPThread = {
...
+        {
+          if (import.meta.url.startsWith('blob:')) {
+            // If our import.meta.url is a blob, we should just use it.
+            worker = new Worker(import.meta.url, {{{ pthreadWorkerOptions }}});
+          } else {
+            // We need to generate the URL with import.meta.url as the base URL of the JS file
+            // instead of just using new URL(import.meta.url) because bundler's only recognize
+            // the first case in their bundling step. The latter ends up producing an invalid
+            // URL to import from the server (e.g., for webpack the file:// path).
+            worker = new Worker(new URL('{{{ TARGET_JS_NAME }}}', import.meta.url), {{{ pthreadWorkerOptions }}});
```
sbc100, line 457: Man.. this limitation of the bundlers really makes our generated code nasty.   I wonder if the bundlers can be fixed... its really a shame to have all this extra stuff just to make them happy.

src/lib/libpthread.js
JoeOsborn: Oops, I had a typo on the next line. Should have read `new Worker(pThreadMainJS)`.  I’ll submit a fix soon.

JoeOsborn: I’ve corrected it but I’m also ok to throw an error in case mainScriptUrlOrBlob is set under ES6_MODULES given -sASSERTIONS.

src/lib/libpthread.js
sbc100: I wonder if we could just fix webpack to support `new Worker(import.meta.url)` which is what we actually want to write here

src/lib/libpthread.js
```diff
@@ -435,11 +435,29 @@ var LibraryPThread = {
...
+          if (typeof pthreadMainJs != 'string') {
+            pthreadMainJs = URL.createObjectURL(pthreadMainJs);
+          }
+          worker = new Worker(pthreadMainJs, {{{ pthreadWorkerOptions }}});
+        } else
+#endif
+        {
+          if (import.meta.url.startsWith('blob:')) {
+            // If our import.meta.url is a blob, we should just use it.
+            worker = new Worker(import.meta.url, {{{ pthreadWorkerOptions }}});
```
sbc100, line 451: I think this syntax is not currently supported by webpack :(   Although it does look like it now is support by vite, which is good!

src/lib/libpthread.js
sbc100: I found a webpack bug from way back about this: https://github.com/webpack/webpack/issues/12638

schlawg:
While you're tinkering there, why isn't `mainScriptUrlOrBlob` honored in the `EXPORT_ES6` code path when it's just a string? It seems consistent and uncontroversial to do that.

src/lib/libpthread.js
schlawg: shouldn't there be some mechanism to make emscripten use `worker = new Worker(import.meta.url, ...)`?

i was excited to hear about USE_ES6_IMPORT_META but it seems that's the bundler hack with import.meta.url as the base.

src/lib/libpthread.js
JoeOsborn: My understanding from https://github.com/webpack/webpack/issues/12638#issuecomment-2698973729 is that this is fixed now, so what’s the move?  To keep the mainScriptUrlOrBlob path and default to new Worker(import.meta.url, options), eliminating the new URL calls/paths?

sbc100: Fantastic.   I think we should probably check with all the different bundlers, but assuming we can go with `new Worker(import.meta.ur)` that sounds like the best solution.

Re-adding support for mainScriptUrlOrBlob sounds good too.

src/lib/libpthread.js
sbc100: Also, can you remove the `USE_ES6_IMPORT_META` from the title (that setting was removed).

JoeOsborn: Yes, I can do this.

JoeOsborn:
Closing in favor of #23890 , #23889 .

PR #23805, opened 2025-03-02
Enable -rpath flag

ryanking13:
Updates `webassembly.py` to parse and store `WASM_DYLINK_RUNTIME_PATH` that were suggested in: https://github.com/WebAssembly/tool-conventions/pull/246


sbc100 (approved):
lgtm.. I guess we should add a test though?  Perhaps we need a test that adds and `-rpath` to the link comment?

hoodmane:
We also need to drop the code that filters out `-Wl,rpath=...` and warns that it is unsupported.

hoodmane:
We also need to teach the runtime function `getDylinkMetadata` about this:
https://github.com/emscripten-core/emscripten/blob/main/src%2Flib%2Flibdylink.js#L466

tools/link.py
```diff
@@ -72,7 +72,6 @@
     # wasm-ld doesn't support soname or other dynamic linking flags (yet).   Ignore them
     # in order to aid build systems that want to pass these flags.
     '-allow-shlib-undefined': False,
-    '-rpath': True,
```
hoodmane, line 75: I wonder if we should also add this to `SUPPORTED_LINKER_FLAGS` above?

hoodmane: Maybe not? @sbc100 it's not entirely clear to me from reading the code.

ryanking13: Thanks. I added `-rpath` there.

ryanking13:
Okay, I teached rpath to linker and libdylink.js, and added some tests that check the RUNTIME_PATH subsection.

The follow-up work would be actually utilizing this runtime path when locating the libraries.

test/test_other.py
```diff
@@ -2374,6 +2374,55 @@ def test_dylink_dependencies(self):
...
+    self.run_js('a.out.js')
+
+    def get_runtime_paths(path):
+      with webassembly.Module(path) as module:
+        dylink_section = module.parse_dylink_section()
+        return dylink_section.runtime_paths
+
+    self.assertEqual(get_runtime_paths('libside2.so') == [])
+    self.assertEqual(get_runtime_paths('libside1.so') == ['$ORIGIN'])
+    self.assertEqual(get_runtime_paths('a.out.wasm') == ['$ORIGIN'])
```
hoodmane, line 2424:
```suggestion
    self.assertEqual(get_runtime_paths('libside2.so'), [])
    self.assertEqual(get_runtime_paths('libside1.so'), ['$ORIGIN'])
    self.assertEqual(get_runtime_paths('a.out.wasm'), ['$ORIGIN'])
```

sbc100 (approved):

test/other/codesize/test_codesize_hello_dylink.gzsize
```diff
@@ -1 +1 @@
-5907
+5926
```
sbc100, line 1: Do you know why the codesize for this test changed?

I would imagine that only tests that explicitly use the rpath would see a code size change.

test/test_other.py
ryanking13: Thanks. I wasn't familiar with the Python unittest syntax 😅

test/other/codesize/test_codesize_hello_dylink.gzsize
ryanking13: I guess it is because I updated libdylink.js?

ryanking13:
Can I ask for you help to understand the `core2` failure?

```
test_dylink_raii_exceptions_reversed_wasm (test_core.core2) ... FAIL
test_dylink_raii_exceptions_reversed_wasm_legacy (test_core.core2) ... ok (7.02s)
test_dylink_raii_exceptions_reversed_emscripten (test_core.core2) ... ok (6.78s)
```

Among the three sibling tests, the one using wasm-eh fails (I guess), while other one passes. I am not sure how they are related to the flag.

test/other/codesize/test_codesize_hello_dylink.gzsize
sbc100: Ah yes, that makes sense.

src/lib/libdylink.js
```diff
@@ -503,6 +504,12 @@ var LibraryDylink = {
               customSection.weakImports.add(symname);
             }
           }
+        } else if (subsectionType === WASM_DYLINK_RUNTIME_PATH) {
+          var runtimePathsCount = getLEB();
+          for (var i = 0; i < runtimePathsCount; ++i) {
+            path = getString();
```
sbc100, line 510: Missing `var` here.

Also how about just `while (runtimePathsCount--)` here?

hoodmane (approved):
Thanks @ryanking13!

sbc100:
Thanks for working on this!

I wonder if we should add this to the ChangeLog.

PR #23807, opened 2025-03-02
Make include files deterministic in debug info

lvlte:
Include files (eg. /emscripten/cache/sysroot/include/c++/v1/complex) now have the deterministic prefix in debug info and source map.

### DIFF
--- a/emcc.py
+++ b/emcc.py
@@ -47,6 +47,7 @@ from tools import cache
 from tools.settings import default_setting, user_settings, settings, MEM_SIZE_SETTINGS, COMPILE_TIME_SETTINGS
 from tools.utils import read_file, removeprefix, memoize
 from tools import feature_matrix
+from tools.system_libs import DETERMINISTIC_PREFIX
 
 logger = logging.getLogger('emcc')
 
@@ -375,6 +376,9 @@ def get_target_flags():
 
 def get_clang_flags(user_args):
   flags = get_target_flags()
+
+  source_dir = utils.path_from_root()
+  flags.append(f'-ffile-prefix-map={source_dir}={DETERMINISTIC_PREFIX}')
 
   # if exception catching is disabled, we can prevent that code from being
   # generated in the frontend


emcc.py
```diff
@@ -376,6 +377,9 @@ def get_target_flags():
 def get_clang_flags(user_args):
   flags = get_target_flags()
 
+  source_dir = utils.path_from_root()
+  flags.append(f'-ffile-prefix-map={source_dir}={DETERMINISTIC_PREFIX}')
```
sbc100, line 381: I'd rather not inject this for all compile commands like this.  I'd rather limit the use of `-ffile-prefix-map` to when we are building system libraries (i.e. `tools/system_libs.py`)`

`get_base_cflags` in `tools/system_libs.py` already has this exact flag, so I guess that question is, which objects are you seeing without the deterministic prefix.

BTW, the deterministic prefix is not indented to apprear in user-generated object or libraries, only in system libraries (since they are built in CI and would contain paths that don't exist on most user's machines).  Any object files that are build on a user's machine can use real paths.  At least that was the idea.

sbc100:
Which object files (or libraries) are you seeing that don't have deterministic debug paths in them?

emcc.py
lvlte: The files that don't have the prefix are those in `cache/sysroot/include` (all of them I suppose), they are built on user's machine (shouldn't they also be pre-built ?) so I added the flag here. I get the idea but I thought it was a good thing to have all emscripten files use the same prefix.

sbc100: I'd rather not inject `-ffile-prefix-map` into build for normal user files.

The result is that I would expect the debug for a given program to contain two types of debug paths:

1. Paths that start with `/emsdk/` (any pre-built library files that emsdk builds)
2. Paths that start with `/real/path/to/emsdk` (any object or libraries build on a user machine where `/real/path/to/emsdk` does actually exist).

If users what to make their debug info portable across machines it would be up to them to use `-ffile-prefix-map` themselves just like they would for clang today.

sbc100: Another way of putting it: I think we are trying to make all the binaries in emsdk itself deterministic, but its not a goal to make use code/objects/libraries deterministic any more than normal clang does.

lvlte: Yes that makes sense, thanks for the explanation.

PR #23815, opened 2025-03-03
Fix babel invocation. NFC

sbc100:
Babel needs to be able to find the emscripten node modules no matter where the config file or current working directory is.

### DIFF
--- a/tools/building.py
+++ b/tools/building.py
@@ -512,6 +512,7 @@ def version_split(v):
 def transpile(filename):
   config = {
     'sourceType': 'script',
+    'presets': ['@babel/preset-env'],
     'targets': {}
   }
   if settings.MIN_CHROME_VERSION != UNSUPPORTED:
@@ -529,8 +530,13 @@ def transpile(filename):
   config_file = shared.get_temp_files().get('babel_config.json').name
   logger.debug(config_json)
   utils.write_file(config_file, config_json)
-  cmd = shared.get_npm_cmd('babel') + [filename, '-o', outfile, '--presets', '@babel/preset-env', '--config-file', config_file]
-  check_call(cmd, cwd=path_from_root())
+  cmd = shared.get_npm_cmd('babel') + [filename, '-o', outfile, '--config-file', config_file]
+  # Babel needs access to `node_modules` for things like `preset-env`, but the
+  # location of the config file (and the current working directory) might not be
+  # in the emscripten tree, so we explicitly set NODE_PATH here.
+  env = os.environ.copy()
+  env['NODE_PATH'] = path_from_root('node_modules')
+  check_call(cmd, env=env)
   return outfile
 
 


tools/building.py
```diff
@@ -529,8 +530,13 @@ def transpile(filename):
   config_file = shared.get_temp_files().get('babel_config.json').name
   logger.debug(config_json)
   utils.write_file(config_file, config_json)
-  cmd = shared.get_npm_cmd('babel') + [filename, '-o', outfile, '--presets', '@babel/preset-env', '--config-file', config_file]
-  check_call(cmd, cwd=path_from_root())
+  cmd = shared.get_npm_cmd('babel') + [filename, '-o', outfile, '--config-file', config_file]
+  # Babel needs access the node_modules for thingsl ike `preset-env`, but the
```
kripken, line 534:
```suggestion
  # Babel needs access to node_modules for things like `preset-env`, but the
```

tools/building.py
```diff
@@ -529,8 +530,13 @@ def transpile(filename):
...
+  # location of the config file (and the current working directory) might not be
+  # in the emscripten tree, so we explicitly set NODE_PATH here.
+  env = os.environ.copy()
+  env['NODE_PATH'] = path_from_root('node_modules')
+  check_call(cmd, env=env)
```
kripken, line 539: The env change fits the PR description, but how does using `babel/preset-env` fit into that? That looks like new functionality?

sbc100: We are using `babel/preset-env`, both before and after this change.

Prior to this change I could not figure out why it only worked when specified on the command line, and not via the config file.   Running with `NODE_PATH` fixes the issue and allows it to live in the config file (which is the normal/documented way to specify it I believe).   

My guess is the reason it worked on the command line is that the resolution of command line flags was don't relative to the babel command itself, but the resolution of the things in the config file was happening relative to the config file.   In this case the config file itself is the the `/tmp` diretory so it has no `node_modules` at all.

kripken: Thanks, now I see where it was used before...

tools/building.py
```diff
@@ -529,8 +530,13 @@ def transpile(filename):
...
   logger.debug(config_json)
   utils.write_file(config_file, config_json)
-  cmd = shared.get_npm_cmd('babel') + [filename, '-o', outfile, '--presets', '@babel/preset-env', '--config-file', config_file]
-  check_call(cmd, cwd=path_from_root())
+  cmd = shared.get_npm_cmd('babel') + [filename, '-o', outfile, '--config-file', config_file]
+  # Babel needs access to `node_modules` for things like `preset-env`, but the
+  # location of the config file (and the current working directory) might not be
+  # in the emscripten tree, so we explicitly set NODE_PATH here.
+  env = os.environ.copy()
+  env['NODE_PATH'] = path_from_root('node_modules')
```
kripken, line 538: Is this always the right path? This gives the `node_module` under the emscripten source tree, but in emsdk builds do we have the same structure?

sbc100: Yup, this is always where `node_modules` lives, alongside `package.json` at the top level.

kripken (approved):

PR #23816, opened 2025-03-04
Simplify worker creation for TRUSTED_TYPES + ESM. NFC

sbc100:
The trustedTypes.createPolicy call there is now subject to the same limitation as the `new Worker` call below.  Its can simply use `import.meta.url` directly I believe.

### DIFF
--- a/src/lib/libpthread.js
+++ b/src/lib/libpthread.js
@@ -422,12 +422,7 @@ var LibraryPThread = {
 #if TRUSTED_TYPES
       // Use Trusted Types compatible wrappers.
       if (typeof trustedTypes != 'undefined' && trustedTypes.createPolicy) {
-        var p = trustedTypes.createPolicy(
-          'emscripten#workerPolicy1',
-          {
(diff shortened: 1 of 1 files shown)


kripken (commented):
Looks right, though I don't know Trusted Types much. Do you know if the test would catch a problem here?

sbc100:
> Looks right, though I don't know Trusted Types much. Do you know if the test would catch a problem here?

No I don't think we have testing that could catch issues with this code.

sbc100:
The old code was using `new URL("{{{ TARGET_JS_NAME }}}", import.meta.url)` instead of just `import.meta.url` in order to be bundler-friendly, but the bundler-friendly requirement only exists for the literal arguments to `new Worker`

kripken:
Oh, then the test for TRUSTED_TYPES in other does not check the actual functionality? I admit I am a little fuzzy on what is verified there. Just building?

kripken (approved):
I see. I don't know if people usually use the two together. lgtm as is.

PR #23818, opened 2025-03-04
Embind: add explicit resource management support

RReverser:
Adds experimental support for the [explicit resource management proposal](https://github.com/tc39/proposal-explicit-resource-management) to Embind classes. It's currently only natively supported in latest Chrome (unflagged) and Node.js (flagged), but users can already enjoy it via TypeScript and Babel transpilation.

It's designed specifically for RAII objects that need to be freed when they go out of scope, which is a perfect fit for Embind (much better than FinalizationRegistry that we used as a loose fallback for forgotten deletes so far).

Before:

```js
var foo = new Module.MyClass();
foo.doSomething();
// somewhere later:
foo.delete();
```

With this support:

```js
using foo = new Module.MyClass();
foo.doSomething();
```

### DIFF
--- a/site/source/docs/porting/connecting_cpp_and_javascript/embind.rst
+++ b/site/source/docs/porting/connecting_cpp_and_javascript/embind.rst
@@ -243,11 +243,27 @@ a C++ object is no longer needed and can be deleted:
 Automatic memory management
 ---------------------------
 
-JavaScript only gained support for `finalizers`_ in ECMAScript 2021, or ECMA-262
-Edition 12. The new API is called `FinalizationRegistry`_ and it still does not
-offer any guarantees that the provided finalization callback will be called.
-Embind uses this for cleanup if available, but only for smart pointers,
-and only as a last resort.
+Embind integrates with the `Explicit Resource Management`_ proposal.
+
+It allows to automatically delete short-lived C++ objects at the end of the
+scope when they're declared with a `using` keyword:
+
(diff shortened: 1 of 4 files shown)


sbc100 (approved):
Nice!

kripken:
Looks like a closure error on CI,
```
building:ERROR: [31m/tmp/emtest_oarnty0w/emscripten_temp_0zte7hq4/embind_test.jso1.js:1844:33: WARNING - [JSC_INEXISTENT_PROPERTY] Property dispose never defined on Symbol
  1844|     const symbolDispose = Symbol.dispose;
                                         ^^^^^^^
```
I guess closure is not aware of this proposal yet?

RReverser:
Yeah I guess I need to wrap the property name in quotes. I'll take a look.

sbc100:
> Yeah I guess I need to wrap the property name in quotes. I'll take a look.

Or maybe it can be added to one of the files in `src/closure-externs/`

RReverser:
Yeah either one is fine.

kripken (approved):
Great! I didn't know JS was getting RAII... :smile:

RReverser:
> Great! I didn't know JS was getting RAII... 😄

Heh yeah long time coming.

Shame that it's only for scopes and won't work for e.g. properties stored in objects and such, but as a primitive to build upon, already better than nothing.

PR #23819, opened 2025-03-04
Display Overloading Method Name in WebIDL Assertion Errors

xpenatan:
When working with a large WebIDL file and a mistake occurs, it’s nearly impossible to determine where the overloading method is located. 
This small change simply adds the method name to the error message, making it easier to identify and resolve the issue:

`AssertionError: overloading must differentiate by # of arguments (cannot have two signatures that differ by types but not by length)`
image

By including the method identifier in the error message, it becomes much simpler to pinpoint the exact location to investigate, especially in WebIDL files exceeding 3,000 lines:

`AssertionError: OffsetCenterOfMassShapeSettings: overloading must differentiate by # of arguments (cannot have two signatures that differ by types but not by length)`
image

### DIFF
--- a/tools/webidl_binder.py
+++ b/tools/webidl_binder.py
@@ -768,7 +768,7 @@ for name in names:
         assert return_type == ret.name, 'overloads must have the same return type'
       for i in range(len(args) + 1):
         if i == len(args) or args[i].optional:
-          assert i not in sigs, 'overloading must differentiate by # of arguments (cannot have two signatures that differ by types but not by length)'
+          assert i not in sigs, f"{m.identifier.name}: overloading must differentiate by # of arguments (cannot have two signatures that differ by types but not by length)"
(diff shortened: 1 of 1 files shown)


tools/webidl_binder.py
```diff
@@ -768,7 +768,7 @@ def add_bounds_check_impl():
         assert return_type == ret.name, 'overloads must have the same return type'
       for i in range(len(args) + 1):
         if i == len(args) or args[i].optional:
-          assert i not in sigs, 'overloading must differentiate by # of arguments (cannot have two signatures that differ by types but not by length)'
+          assert i not in sigs, f"{m.identifier.name}: overloading must differentiate by # of arguments (cannot have two signatures that differ by types but not by length)"
```
sbc100, line 771: Can you keep the single quotes there.

PR #23828, opened 2025-03-04
[test] Remove test/wasmfs/get_backend.h. NFC

sbc100:
These days we have a better mechanism that uses  `setup_nodefs.js` as a `--pre-js`.

See `setup_nodefs_test` in `test/common.py`

The advantage of this new mechanism is that it doesn't require the test itself to know about wasmfs or any specific backend.


test/wasmfs/wasmfs_getdents.c
```diff
@@ -54,21 +52,23 @@ void print(const char* dir) {
...
-  int err = wasmfs_create_directory("/root", 0777, get_backend());
+  int err;
 
   // Set up test directories.
-  err = mkdir("/root/working", 0777);
+  err = mkdir("root", 0777);
+  assert(err != -1);
+  err = mkdir("root/working", 0777);
   assert(err != -1);
-  err = mkdir("/root/working/test", 0777);
```
tlively, line 62: Why change these files paths to be relative?

sbc100: See https://github.com/emscripten-core/emscripten/blob/4c91962ea50831ba0da77c751017866252e50916/test/setup_nodefs.js#L1-L6

tlively (approved):
Nice cleanup!

test/test_other.py
```diff
@@ -133,12 +138,17 @@ def also_with_wasmfs_all_backends(f):
...
+        self.setup_noderawfs_test()
       self.emcc_args.append(f'-D{backend}')
     f(self, *args, **kwargs)
 
   parameterize(metafunc, {'': (None,),
-                          'wasmfs': ('WASMFS_MEMORY_BACKEND',),
-                          'wasmfs_node': ('WASMFS_NODE_BACKEND',)})
+                          'wasmfs': ('memory',),
+                          'wasmfs_node': ('node',),
+                          'wasmfs_raw': ('raw',)})
```
sbc100, line 151: @tlively do you think it makes sense to test both nodefs (where nodefs is mounted as `/nodefs`) and noderawfs (where nodefs is mounted at the root) with wasmfs?  I guess maybe yes?

tlively: No strong opinion, but I would err on the side of testing both, yeah.

Issue #23685, opened 2025-02-16
emrun-related JS code is broken


oitel:
emrun cannot get output from a browser (tested for Firefox 135 and Chromium 133) for the apps compiled with emsdk v4.0.3 and v4.0.4:
```
Web server root directory: /tmp/em_test/v403
Starting web server: http://0.0.0.0:6931/
Starting browser: /usr/bin/xdg-open http://localhost:6931/main.html
/usr/bin/xdg-open
Searching for processes by full path name "/usr/bin/xdg-open".. found 0 entries
Launched browser process with pid=200059
Now listening at http://0.0.0.0:6931/
Entering web server loop.
First navigation occurred. Identifying currently running browser processes
Searching for processes by full path name "/usr/bin/xdg-open".. found 0 entries
Was unable to detect the browser process that was spawned by emrun. This may occur if the target page was opened in a tab on a browser process that already existed before emrun started up.
The html page you are running is not emrun-capable. Stdout, stderr and exit(returncode) capture will not work. Recompile the application with the --emrun linker flag to enable this, or pass --no-emrun-detect to emrun to hide this check.
```

For the apps compiled with emsdk v4.0.2, emrun works successfully:
```
Web server root directory: /tmp/em_test/v402
Starting web server: http://0.0.0.0:6931/
Starting browser: /usr/bin/xdg-open http://localhost:6931/main.html
/usr/bin/xdg-open
Searching for processes by full path name "/usr/bin/xdg-open".. found 0 entries
Launched browser process with pid=200392
Now listening at http://0.0.0.0:6931/
Entering web server loop.
First navigation occurred. Identifying currently running browser processes
Searching for processes by full path name "/usr/bin/xdg-open".. found 0 entries
Was unable to detect the browser process that was spawned by emrun. This may occur if the target page was opened in a tab on a browser process that already existed before emrun started up.
POST: "/stdio.html" (path: "/stdio.html", query: "")
POST: "/stdio.html" (path: "/stdio.html", query: "")
Hello, world!
POST: "/stdio.html" (path: "/stdio.html", query: "")
Shutting down because browser is no longer alive
Browser process has shut down, quitting web server.
Web page has quit with a call to exit() with return code 0. Shutting down web server. Pass --serve-after-exit to keep serving even after the page terminates with exit().
Web server loop done.
Closed web server.
emrun quitting with process exit code 0
```

[...]

PR #23830, opened 2025-03-05
emrun: Add missing library function inclusion

oitel:
Related to #23685 

This change fixes the runtime error which causes emrun to be unable to get output from a browser.


sbc100 (approved):

sbc100:
Our emrun tests are currently disabled in CI (we should fix that).. so landing this without CI testing for now,

PR #23839, opened 2025-03-05
Update test_wasmfs_getdents. NFC

sbc100:
Handle more file entry types.

Don't attempt to seek to position 1.  Instead seek back to a previously seen position.  I'm honestly not even sure this is valid but both old and new filesystems support this.

These changes are in preparation for running this test on all filesystem types.

### DIFF
--- a/test/wasmfs/wasmfs_getdents.c
+++ b/test/wasmfs/wasmfs_getdents.c
@@ -20,6 +20,17 @@
 #include <sys/syscall.h>
 #include <unistd.h>
 
+const char* type_to_string(char d_type) {
+  switch (d_type) {
+    case DT_REG: return "DT_REG";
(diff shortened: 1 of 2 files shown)


test/wasmfs/wasmfs_getdents.c
```diff
@@ -107,12 +112,17 @@ int main() {
     console.log();
   });
 
-  // Try to advance the offset of the directory.
-  // Expect that '.' will be skipped.
+  // Try to advance reset of the offset of the directory.
+  // Expect that '..' be printed a second time.
```
tlively, line 116: The wording of these sentences is a little confusing.

tlively (approved):

sbc100:
> Oops, didn't see that auto-merge was set.

I will update.

PR #23847, opened 2025-03-06
[WebGL] Move `$JSEvents` dependency to `libhtml5_webgl.js`. NFC

kleisauke:


src/lib/libhtml5_webgl.js
```diff
@@ -331,9 +331,12 @@ var LibraryHtml5WebGL = {
   },
 
   emscripten_webgl_destroy_context__proxy: 'sync_on_webgl_context_handle_thread',
-  emscripten_webgl_destroy_context__deps: ['free'],
+  emscripten_webgl_destroy_context__deps: ['free', '$JSEvents'],
   emscripten_webgl_destroy_context: (contextHandle) => {
     if (GL.currentContext == contextHandle) GL.currentContext = 0;
+    // Release all JS event handlers on the DOM element that the GL context is
```
sbc100, line 337: Should we keep the soft dependency in the form of `if (typeof JSEvents == 'object') {` rather than adding a hard dependency?

Or is the code in the tile all expected to to depend in libhtml5 stuff already?

sbc100 (commented):
Can you explain the rational for this change in the PR description?  Why do you think this code belongs more in the new location than the old?

src/lib/libhtml5_webgl.js
sbc100: sgtm

sbc100: Hmm.. but actually it looks like `emscripten_webgl_do_create_context` should not be depending on `JSEvents` .. I don't see any usage of it in that function.

kleisauke: Ah, you're right. That's weird. It looks like it's only used by `$registerWebGlEventCallback()` via the 
`emscripten_set_webglcontextlost_callback()` and `emscripten_set_webglcontextrestored_callback()` APIs.

kleisauke:
Closing in favor of #23850.

sbc100:
Might it not be good to move this code anyway?

kleisauke:
`GL.deleteContext()` is [also used by the EGL implementation](https://github.com/emscripten-core/emscripten/blob/44323dfc75a34427a5b47ce9a3d5e281b2cd1791/src/lib/libegl.js#L393), so this PR does have an impact if you use that API alongside `emscripten_set_webglcontext{lost,restored}_callback()`.

There isn't a test that specifically covers this, but I think one can reproduce that by rewriting [`test/browser/webgl_destroy_context.c`](https://github.com/emscripten-core/emscripten/blob/44323dfc75a34427a5b47ce9a3d5e281b2cd1791/test/browser/webgl_destroy_context.c) on top of EGL, so using `egl{Create,Destroy}Context()` instead of `emscripten_webgl_{create,destroy}_context()`.

sbc100:
> `GL.deleteContext()` is [also used by the EGL implementation](https://github.com/emscripten-core/emscripten/blob/44323dfc75a34427a5b47ce9a3d5e281b2cd1791/src/lib/libegl.js#L393), so this PR does have an impact if you use that API alongside `emscripten_set_webglcontext{lost,restored}_callback()`.
> 
> There isn't a test that specifically covers this, but I think one can reproduce that by rewriting [`test/browser/webgl_destroy_context.c`](https://github.com/emscripten-core/emscripten/blob/44323dfc75a34427a5b47ce9a3d5e281b2cd1791/test/browser/webgl_destroy_context.c) on top of EGL, so using `egl{Create,Destroy}Context()` instead of `emscripten_webgl_{create,destroy}_context()`.

I suppose the question is.. is it valid to use libhtml5 APIs in EGL programs or are those two APIs more-or-less mutually exclusive?

PR #23850, opened 2025-03-06
Remove unused `$JSEvents` dependencies. NFC

kleisauke:


sbc100 (approved):

sbc100:
You know what might be nices is a conservative debug-mode scanner that confirmed that a all dependencies were indeeed presend in a given function..  I guess it there will be exceptions and / or false positives.. but I might try to build it.

PR #23852, opened 2025-03-06
Fix fetchfs off by one error

JoeOsborn:
In #23669 , I tried to implement length clipping to prevent reads past the end of a file.  Unfortunately I had an extra, unnecessary +1 term in some places which led to issues including problems when the whole file fit into a single chunk.  This patch corrects the clipping.


src/lib/libwasmfs_fetch.js
```diff
@@ -134,7 +134,7 @@ addToLibrary({
           return failedResponse.status === 404 ? -{{{ cDefs.ENOENT }}} : -{{{ cDefs.EBADF }}};
         }
         var fileInfo = wasmFS$JSMemoryRanges[file];
-        length = Math.min(length, fileInfo.size-offset+1) | 0;
+        length = Math.min(length, fileInfo.size-offset) | 0;
```
sbc100, line 137: Do you need the `| 0` here or above?

test/wasmfs/wasmfs_fetch.c
```diff
@@ -263,6 +263,23 @@ void test_nonexistent() {
   assert(close(fd) == 0);
 }
 
+void test_big_chunks() {
+  printf("Running %s...\n", __FUNCTION__);
+
+  char expected[] = "hello";
+  size_t size = 5;
+
+  backend_t backend = wasmfs_create_fetch_backend("small.dat",16);
```
sbc100, line 272: Space after comma

sbc100 (approved):

src/lib/libwasmfs_fetch.js
JoeOsborn: Ah no, I guess they're only needed after divisions

JoeOsborn:
I saw a failure in test_setjmp_many_emscripten but I'm not sure how my code could have influenced that?

sbc100:
We do have some test flakiness sadly.   I will re-run.

PR #23853, opened 2025-03-06
[embind] Fix JSPI dynCalls.

### LINKED ISSUE #23834: "WebAssembly.promising(): Argument 0 must be a WebAssembly exported function" when using JSPI and MAXIMUM_MEMORY=4GB
Using Emsdk 4.0.2
Compiler line: em++ main.cpp -o main.html -sMAXIMUM_MEMORY=4GB -sJSPI --bind

Using JSPI and setting MAXIMUM_MEMORY to 4GB causes a runtime exception, most likely related to functions exported EMSCRIPTEN_BINDINGS using emscripten::async(). 

Setting MAXIMUM_MEMORY to 2GB and below works as expected. (As I understand it, MAXIMUM_MEMORY > 2GB triggers an extra binaryen pass over the generated JS code)

Unfortunately, I've not been able to reduce this to a minimal test case. It does not reproduce using a small example with a single EMSCRIPTEN_BINDINGS exported function.

Stack trace:
```
Uncaught (in promise) TypeError: WebAssembly.promising(): Argument 0 must be a WebAssembly exported function
    at Object.makeAsyncFunction (mandelbrot.js:5713:24)
    at craftInvokerFunction (mandelbrot.js:5743:31)
    at mandelbrot.js:5937:31
    at onComplete (mandelbrot.js:5247:28)
    at whenDependentTypesAreResolved (mandelbrot.js:5276:5)
    at __embind_register_function (mandelbrot.js:5935:3)
    at mandelbrot.wasm.void emscripten::function<void, emscripten::async>(char const*, void (*)(), emscripten::async) (http://localhost:8000/webassembly-examples/wasm-jspi/qmake/mandelbrot.wasm)
    at mandelbrot.wasm.embind_init_qtSuspendResumeControl() (http://localhost:8000/webassembly-examples/wasm-jspi/qmake/mandelbrot.wasm)
    at mandelbrot.wasm.emscripten::internal::InitFunc::InitFunc(void (*)()) (http://localhost:8000/webassembly-examples/wasm-jspi/qmake/mandelbrot.wasm)
    at mandelbrot.wasm.EmBindInit_qtSuspendResumeControl::EmBindInit_qtSuspendResumeControl() (http://localhost:8000/webassembly-examples/wasm-jspi/qmake/mandelbrot.wasm)

```

brendandahl:
WebAssembly.promising requires a wasm function argument, previously we were passing in a wrapper JS function that caused an error. Move wrapping the wasm function with WebAssembly.promising into
embind__requireFunction and also handle it in dynCall.

Fixes #23834

### DIFF
--- a/src/lib/libembind.js
+++ b/src/lib/libembind.js
@@ -774,13 +774,6 @@ var LibraryEmbind = {
 #if ASSERTIONS && ASYNCIFY != 2
     assert(!isAsync, 'Async bindings are only supported with JSPI.');
 #endif
-
-#if ASYNCIFY == 2
-    if (isAsync) {
-      cppInvokerFunc = Asyncify.makeAsyncFunction(cppInvokerFunc);
-    }
-#endif
-
     var isClassMethodFunc = (argTypes[1] !== null && classType !== null);
 
     // Free functions with signature "void function()" do not need an invoker that marshalls between wire types.
@@ -896,11 +889,15 @@ var LibraryEmbind = {
   },
 
   $embind__requireFunction__deps: ['$readLatin1String', '$throwBindingError'
-#if DYNCALLS || !WASM_BIGINT || MEMORY64
+#if DYNCALLS || !WASM_BIGINT || MEMORY64 || CAN_ADDRESS_2GB
     , '$getDynCaller'
 #endif
   ],
-  $embind__requireFunction: (signature, rawFunction) => {
+  $embind__requireFunction: (signature, rawFunction, isAsync = false) => {
+#if ASSERTIONS && ASYNCIFY != 2
+    assert(!isAsync, 'Async bindings are only supported with JSPI.');
+#endif
+
     signature = readLatin1String(signature);
 
     function makeDynCaller() {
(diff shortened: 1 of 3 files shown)


test/test_other.py
```diff
@@ -3255,13 +3255,14 @@ def test_embind_asyncify(self):
...
-    'no_dynamic': ['-sDYNAMIC_EXECUTION=0'],
+    '': [['-sDYNAMIC_EXECUTION=1']],
+    'no_dynamic': [['-sDYNAMIC_EXECUTION=0']],
+    'dyncall': [['-sALLOW_MEMORY_GROWTH', '-sMAXIMUM_MEMORY=4GB']],
   })
   @requires_jspi
   def test_embind_jspi(self, extra):
     self.emcc_args += ['-lembind', '-g']
-    self.emcc_args += [extra]
+    self.emcc_args += extra
```
sbc100, line 3265: Maybe just combine two lines into the do_runf call below:

```
self.do_runf('embind/embind_jspi_test.cpp', 'done', emcc_args=['-lembind', '-g'] + extra)
```

Also, the convention I think is to call the `extra` param `args`.

src/lib/libembind.js
```diff
@@ -914,10 +911,16 @@ var LibraryEmbind = {
 #endif
 #if MEMORY64 || CAN_ADDRESS_2GB
       if (signature.includes('p')) {
-        return getDynCaller(signature, rawFunction);
+        return getDynCaller(signature, rawFunction, isAsync);
+      }
+#endif
+      var rtn = getWasmTableEntry(rawFunction);
+#if JSPI
+      if (isAsync) {
```
sbc100, line 919: Should we just always use `getDynCaller` maybe?  Or always use it in JSPI mode?

sbc100 (approved):

src/lib/libembind.js
brendandahl: I'd like to avoid dynCall since it looks like it adds quite a bit of overhead. It creates a wrapper and then that wrapper calls into getWasmTable entry every time the function is called. In non dyncall mode, the current code just directly calls the function or the the promising wrapper.

sbc100: Fair enough

PR #23855, opened 2025-03-07
[WIP] Add dependency checking

sbc100:


src/jsifier.mjs
```diff
@@ -45,6 +45,10 @@ const addedLibraryItems = {};
 
 const extraLibraryFuncs = [];
 
+// Experimental feature to check for invalid __deps entries. Modify this line
+// locally to try it out.
+const CHECK_DEPS = true;
```
sbc100, line 50: I plan on setting this to false before landing it.

We can look into enabling more widely as a followup.

src/lib/libglemu.js
```diff
@@ -3616,7 +3615,7 @@ var LibraryGLEmulation = {
 
   // OpenGL Immediate Mode matrix routines.
   // Note that in the future we might make these available only in certain modes.
-  glMatrixMode__deps: ['$GL', '$GLImmediateSetup', '$GLEmulation'], // emulation is not strictly needed, this is a workaround
+  glMatrixMode__deps: ['$GL', '$GLImmediateSetup'], // emulation is not strictly needed, this is a workaround
```
kripken, line 3618: The comment suggests emulation was needed for some non-direct reason. Do you know what that is?

sbc100: Good point.. perhaps we should leave this in out of an abundance of caution.    I'll see if the git log can tell me why this was needed.

sbc100:
OK I think this PR is ready to land now.  Please take a final look.

src/jsifier.mjs
```diff
@@ -45,6 +45,10 @@ const addedLibraryItems = {};
 
 const extraLibraryFuncs = [];
 
+// Experimental feature to check for invalid __deps entries. Modify this line
+// locally to try it out.
```
kripken, line 49: Is the comment stale? Now it seems there is an env var, no need to modify the source line?

kripken (approved):

PR #23862, opened 2025-03-07
Rename browser test parameter `args` -> `emcc_args`. NFC

sbc100:
This is for consistency with the non-browser tests where `args` means program arguments and `emcc_args` means compiler arguments.


sbc100:
I could instead go the other way and use `args` in core and other tests and then rename the existing `args` to something like `program_args`?

kripken (approved):
This looks nicer than the other option I think.

PR #15987, opened 2022-01-12
SDL2 CMake config file

Daft-Freak:
This is something I've been wanting to look at for a while. SDL's configure/cmake builds generate a configuration file for CMake that allows you to do this:
```cmake
find_package(SDL2 REQUIRED)
...
target_link_libraries(blah SDL2::SDL2)
```

This PR adds a minimal config file that just adds `-s USE_SDL=2` to the compile/link flags and then the port build sorts everything else out.

(Using the old style `SDL2_INCLUDE_DIRS`/`_LIBRARIES` fails if the port isn't already downloaded as there's nowhere to insert `-s USE_SDL=2` into the compiler flags...)

PR #23869, opened 2025-03-08
Add SDL3 CMake config file

bobsayshilol:
This is the equivalent of, and heavily based on, #15987 but for SDL3.

Assuming that I've understood https://github.com/emscripten-core/emscripten/pull/19780#issuecomment-1622467552 correctly, none of the old style variables are expected to be set so we only need to provide targets. This has the benefit that it Just Works without needing to run `emcc -sUSE_SDL=3` first to download the port.


system/lib/cmake/SDL3/sdl3-config.cmake
```diff
@@ -0,0 +1,18 @@
+if(NOT TARGET SDL3::SDL3)
```
sbc100, line 1: The SDL2 version of this file sets SDL2_INCLUDE_DIRS and SDL2_LIBRARIES.. are those not needed in this case?

sbc100 (approved):

system/lib/cmake/SDL3/sdl3-config.cmake
bobsayshilol: Nope, since SDL3 is now target based and no longer uses the old style variables, so linking to one of the `SDL3::*` targets is enough to set it up correctly. See https://wiki.libsdl.org/SDL3/README/cmake where there's no mention of the old style `<name>_LIBRARIES`/`<name>_INCLUDE_DIRS`.

sbc100 (approved):

bobsayshilol:
@sbc100 The CI job seems stuck. Do I need to press `Update branch` again or is there something else that needs doing?

PR #23872, opened 2025-03-09
Respect runtime paths when loading shared libraries

ryanking13:
This updates `loadDynamicLibrary` to locate shared libraries from the file system if possible.

When locating a shared library with a relative path, first attempt to find it in the `LD_LIBRARY_PATH` directories. If it is not found there, search for it in the runtime path directories. If it is not found, fall back to the original `Module['locateFile']`.

Followup to #23805.


src/lib/libdylink.js
```diff
@@ -959,11 +1014,24 @@ var LibraryDylink = {
...
     dbg(`loadDynamicLibrary: ${libName} handle: ${handle}`);
     dbg(`existing: ${Object.keys(LDSO.loadedLibsByName)}`);
 #endif
+
+#if FILESYSTEM
+      var runtimePathsAbs = (rpath.paths || []).map((p) => replaceORIGIN(rpath.parentLibPath, p));
+      var searchDirs = getDefaultLibDirs().concat(runtimePathsAbs);
+      var [existsInFS, libNameAbs] = locateLibraryFromFS(libName, searchDirs);
+      if (existsInFS) {
+        libName = libNameAbs;
```
ryanking13, line 1028: Review point: This part will switch the `libName`  to the absolute path, which means the values stored in `LSDO.loadedLibsByName` will change to absolute path accordingly.

In Pyodide, we were already storing absolute path to prevent loading the same library twice ([code pointer](https://github.com/pyodide/pyodide/blob/a53c17fd8ef87e79c7d3bfbb1262ca7ff6718ef2/src/js/dynload.ts#L157-L168)), but I guess this can be a breaking change?

Also, the absolute path of the shared library is needed to calculate the `$ORIGIN` of the rpath.

src/lib/libdylink.js
```diff
@@ -48,6 +48,62 @@ var LibraryDylink = {
     registerWasmPlugin();
     `,
   $preloadedWasm: {},
+
+  $locateLibraryFromFS__deps: ['$FS', '$PATH'],
+  $locateLibraryFromFS: (filename, searchDirs) => {
+    // Find the library in the filesystem
+    // returns [true, abspath] if found, [false, null] otherwise.
+    if (!(typeof FS.findObject === 'function')) {
```
hoodmane, line 56: I think we can use `#ifdef` to stub out this function, but I'm not sure what the right test is to check if FS is included. Probably this check can replace the FS dependency.

src/lib/libdylink.js
```diff
@@ -48,6 +48,62 @@ var LibraryDylink = {
...
+      // FS may not have the findObject method if the filesystem is not
+      // used in the build. Skip the search in this case.
+      return [false, null];
+    }
+    var candidates = [];
+    if (PATH.isAbs(filename)) {
+      candidates.push(filename);
+    } else if (searchDirs) {
+      for (var i = 0; i < searchDirs.length; i++) {
+        candidates.push(PATH.join(searchDirs[i], filename));
```
hoodmane, line 66: I'm concerned that `PATH.join()` mishandles symlinks, I'll double check later.

src/lib/libdylink.js
```diff
@@ -48,6 +48,62 @@ var LibraryDylink = {
...
+      }
+    } else {
+      return [false, null];
+    }
+
+    for (var i = 0; i < candidates.length; i++) {
+      var path = candidates[i];
+      if (FS.findObject(path)) {
+        return [true, path];
+      }
```
hoodmane, line 76: `findObject` is semi deprecated I think. This also resolves symlinks and returns the fully resolved absolute path:
```suggestion
      try {
        var res = FS.lookupPath(path);
        return [true, res.path];
      } catch(e) {}
```

src/lib/libdylink.js
```diff
@@ -48,6 +48,62 @@ var LibraryDylink = {
...
+    return data;
+  },
+
+  $getDefaultLibDirs__deps: ['$ENV'],
+  $getDefaultLibDirs: () => {
+    var ldLibraryPath = ENV['LD_LIBRARY_PATH'];
+    if (ldLibraryPath) {
+      return ldLibraryPath.split(':');
+    }
+    return [];
```
hoodmane, line 94: One liner version:
```suggestion
    return ENV['LD_LIBRARY_PATH']?.split(':') ?? [];
```

src/lib/libdylink.js
```diff
@@ -48,6 +48,57 @@ var LibraryDylink = {
     registerWasmPlugin();
     `,
   $preloadedWasm: {},
+
+  $locateLibraryFromFS__deps: ['$FS', '$PATH'],
+  $locateLibraryFromFS: (filename, searchDirs) => {
+    // Find the library in the filesystem
+    // returns [true, abspath] if found, [false, null] otherwise.
```
hoodmane, line 55: I think just returning `abspath` or `null` is sufficient.

src/lib/libdylink.js
```diff
@@ -959,11 +1014,24 @@ var LibraryDylink = {
...
+  $loadDynamicLibrary: function(libName, flags = {global: true, nodelete: true}, localScope, rpath = {parentLibPath: '', paths: []}, handle) {
 #if DYLINK_DEBUG
     dbg(`loadDynamicLibrary: ${libName} handle: ${handle}`);
     dbg(`existing: ${Object.keys(LDSO.loadedLibsByName)}`);
 #endif
+
+#if FILESYSTEM
+      var runtimePathsAbs = (rpath.paths || []).map((p) => replaceORIGIN(rpath.parentLibPath, p));
+      var searchDirs = getDefaultLibDirs().concat(runtimePathsAbs);
+      var [existsInFS, libNameAbs] = locateLibraryFromFS(libName, searchDirs);
```
hoodmane, line 1031: If `locateLibraryFromFS` is changed to return just a single value, then the following replaces `libName` with the value returned from `locateLibraryFromFS` unless it was `null`:
```suggestion
      libName = locateLibraryFromFS(libName, searchDirs) ?? libName;
```

src/lib/libdylink.js
```diff
@@ -959,11 +1014,24 @@ var LibraryDylink = {
...
+      if (existsInFS) {
+        libName = libNameAbs;
+#if DYLINK_DEBUG
+        dbg(`found library from filesystem: ${libName}`);
+#endif
```
hoodmane: Maybe move this log into `locateLibraryFromFS`.

src/lib/libdylink.js
hoodmane: But I guess you need `existsInFS` later, so the result from `locateLibraryFromFS` needs to be stored to calculate that.

hoodmane: I think `readLibraryFromFS` and `locateLibraryFromFS` should be combined into one function. Then we can just do:
```
var libData = readLibraryFromFS(libName, searchDirs);
if (libData) {
   return flags.loadAsync ? Promise.resolve(libData) : libData;
}
```

test/test_other.py
```diff
@@ -2406,6 +2406,7 @@ def get_runtime_paths(path):
     self.assertEqual(get_runtime_paths('libside1.so'), ['$ORIGIN'])
     self.assertEqual(get_runtime_paths('a.out.wasm'), ['$ORIGIN'])
 
+
```
hoodmane: Revert whitespace change.

test/test_other.py
```diff
@@ -7662,6 +7662,137 @@ def test_ld_library_path(self, args):
     self.assertContained('Hello4', out)
     self.assertContained('Ok', out)
 
+  def test_ld_library_path_dependencies(self, args):
+    if args:
+      self.setup_node_pthreads()
```
hoodmane, line 7667:
```suggestion
  def test_ld_library_path_dependencies(self):
```

src/lib/libdylink.js
hoodmane: So the only case where it could be a problem is if they were defining `EMV['LD_LIBRARY_PATH']` and there is a file of the given name on the `LD_LIBRARY_PATH`? Otherwise there is no change? Doesn't seem like a problematic change to me.

src/lib/libdylink.js
sbc100: Kind of a shame we need to inject a try/crash here though.   @hoodmane do I remember talking with you about a version of `lookupPath` that doesn't throw?

src/lib/libdylink.js
ryanking13: It would also change the parameter passed to [`locateFile`](https://github.com/emscripten-core/emscripten/blob/f391a0d8462339a22ac335a0c8e00693614748de/src/lib/libdylink.js#L1027C21-L1027C31) which will be called as a fallback when loading from filesystem fails.

But in that case, two conditions should met: 1) the file is in the filesystem, 2) but it had failed to load from the filesystem for some reason, hence fallback to `locateFile`. So I don't think it would be a breaking change for users of `locateFile`.

src/lib/libdylink.js
ryanking13: I wanted to call `libname = libnameAbs` at the very first part of the function because other wise, different `libname` will be used in `loadedLibsByName`, `mergeLibSymbols`, `sharedModules`.

src/lib/libdylink.js
ryanking13: I haven't thought about symlink behavior in an emscripten environment, could it be an issue with a filesystem like NODEFS?

hoodmane: I've dealt with issues like this a lot lately:
* https://github.com/emscripten-core/emscripten/pull/23000
https://github.com/emscripten-core/emscripten/pull/23136
https://github.com/emscripten-core/emscripten/pull/23072
https://github.com/emscripten-core/emscripten/pull/23002
https://github.com/emscripten-core/emscripten/pull/23307

I think the appropriate code here is:
```suggestion
        candidates.push(searchDirs[i] + '/' + filename);
```

src/lib/libdylink.js
```diff
@@ -48,6 +48,59 @@ var LibraryDylink = {
...
+
+    for (var i = 0; i < candidates.length; i++) {
+      var path = candidates[i];
+      try {
+        var res = FS.lookupPath(path);
+        return res.path;
+      } catch(e) {
+        // do nothing is file is not found
+      }
+    }
```
hoodmane, line 80:
```suggestion
    for (var path of candidates) {
      try {
        var res = FS.lookupPath(path);
        return res.path;
      } catch(e) {
        // do nothing is file is not found
      }
    }
```

src/lib/libdylink.js
ryanking13: > I also removed the FS and PATH dependency as I think they are implied in the ifdef FILESYSTEM.

Never mind. It doesn't seem to work in some cases.

ryanking13:
The CI is green now, but two issues are remaining:

1. I wasn't able to remove `$FS` and `$PATH` dependency. `ifdef` hacks doesn't seem to work. And marking those dependencies seems to always link `libfs` and `libpath` (even when they are in the `ifdef FILESYSTEM` block), increasing the output size a lot for the `codesize_hello_dylink` test.

2. It seems like `wasmfs` doesn't yet implement `FS.lookupPath`. So I ended up checking `FS.lookupPath` existence and early return.

src/lib/libdylink.js
hoodmane: We added `noent_okay` for mkdir / O_CREAT so that it won't throw if the __last__ entry doesn't exist. Perhaps we should make a variant that returns `null` if an `ENOENT` occurs in `lookupNode`.

sbc100:
There is existing logic in system/lib/libc/dynlink.c which searches `LD_LIBRARY_PATH` so we would somehow avoid searching that path twice.

ryanking13:
> There is existing logic in system/lib/libc/dynlink.c which searches LD_LIBRARY_PATH so we would somehow avoid searching that path twice.

Yes, but that only happens in top-level libraries, I guess? Here is the code flow I understand:

1. A user call `dlopen("lib.so")` in their c codebase
2. LD_LIBRARY_PATH will be searched inside dlopen, replace the `lib.so` with the absolute path, and call `_emscripten_dlopen_js` js function with the resolved path.
3. `_emscripten_dlopen_js` will internally call `loadDynamicLibrary` with the absolute path (== `loadDynamicLibrary('/abs/path/of/lib.so')` )
4. Since the absolute path is given to `loadDynamicLibrary`, it avoids searching `LD_LIBRARY_PATH` and loads the file directly.
5. `loadDynamicLibrary` starts to load the dependencies of `lib.so`, e.g. `dep1.so` ( == `loadDynamicLibrary('dep1.so')`)
6. We don't know where to locate `dep1.so`. Therefore, search `LD_LIBRARY_PATH` and rpaths of `lib.so` to locate dep1.so.

hoodmane:
test_dylink_dso_needed is failing so maybe the JS approach is indeed needed:

```
Aborted(Assertion failed: native function `_emscripten_resolve_path` called before runtime initialization)
/tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:869
          ^

RuntimeError: Aborted(Assertion failed: native function `_emscripten_resolve_path` called before runtime initialization)
    at abort (/tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:869:11)
    at assert (/tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:270:5)
    at /tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:879:5
    at loadDynamicLibrary (/tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:4680:27)
    at /tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:4818:11
Thrown at:
    at abort (/tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:869:11)
    at assert (/tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:270:5)
    at /tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:879:5
    at loadDynamicLibrary (/tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:4680:27)
    at /tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:4818:11
```

sbc100:
> test_dylink_dso_needed is failing so maybe the JS approach is indeed needed:
> 
> ```
> Aborted(Assertion failed: native function `_emscripten_resolve_path` called before runtime initialization)
> /tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:869
>   var e = new WebAssembly.RuntimeError(what);
>           ^
> 
> RuntimeError: Aborted(Assertion failed: native function `_emscripten_resolve_path` called before runtime initialization)
>     at abort (/tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:869:11)
>     at assert (/tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:270:5)
>     at /tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:879:5
>     at loadDynamicLibrary (/tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:4680:27)
>     at /tmp/emtest_kxb8mepw/emscripten_test_lsan_10i7e2lr/main.js:4818:11
> ```

One problem is the for wasmfs the filesystem itself cannot be accessed before the runtime is initialized.  Its hard to see how its possible to load shared libraries out of the filesystem before the runtime is initialaized.

Perhaps we need to start initializing the runtime, and then have DSO loading triggered at the point when the filesystem is ready.

hoodmane:
Well calling `wasmExports["_emscripten_resolve_path"]` to explicitly work around the `runtimeInitialized` check seems to pass all the tests. But also, the JS implementation seems to pass all the tests, including the wasmfs test. So both approaches work and we just have to decide which we prefer. As far as I understand, the main thing we can't do before `runtimeInitialized` is call malloc?

hoodmane:
I guess we haven't applied relocations yet either which must add additional restrictions but I'm not sure exactly what they are.

sbc100:
> I guess we haven't applied relocations yet either which must add additional restrictions but I'm not sure exactly what they are in practice.

None of the C or C++ static constructors are guaranteed to have run until the runtime has been initialized.

We keep a list of known static initializers here:
https://github.com/emscripten-core/emscripten/blob/main/system/lib/README.md#static-constructor-ordering

It seems like if want load time DSO loading to work correctly for wasmfs it would need to come after wasmfs init.

So maybe we need a new static initializer calls `load_dso` that comes at the end of this list.

I'm not saying we have to block this work on the wasmfs stuff, but I think it might be a good idea, and its also a good argument against the pure-JS approach.

hoodmane:
Don't we currently allow a static initializer to call symbols from preloaded dynamic libraries? It seems like running the static initializers before preloading would be a breaking change.

hoodmane:
Also, `other.test_ld_library_path_dependencies_wasmfs` is currently passing. Is it possible that nothing is actually broken with wasmfs with the current diff? Or if there is a problem, how should we change the test to detect it?

sbc100:
> Don't we currently allow a static initializer to call symbols from preloaded dynamic libraries? It seems like running the static initializers before preloading would be a breaking change.

We would have to order the intializers so that DSO loading came before all user initializers.   In general I think the number 0-100 are reserved for system internalizera.

hoodmane:
Where are the initializers for wasmfs located?

test/test_other.py
```diff
@@ -7691,6 +7691,136 @@ def test_ld_library_path(self, args):
     self.assertContained('Hello4', out)
     self.assertContained('Ok', out)
 
+  @also_with_wasmfs
+  def test_ld_library_path_dependencies(self):
```
sbc100, line 7695: Can we call this `test_dlopen_rpath` maybe?   (at least it should start with something like `test_dlopen` or `test_dylink`).

Also, perhaps we have an existing tests which has shared libraries with dependencies that we can make this a variant of?

sbc100:
> Also, `other.test_ld_library_path_dependencies_wasmfs` is currently passing. Is it possible that nothing is actually broken with wasmfs with the current diff? Or if there is a problem, how should we change the test to detect it?

I think that is because the test uses runtime linking with `dlopen`.. which means that the logic in running after runtime init.   In order to really test the wasmfs we would need to have load time dynamic libraries that requires rpath stuff (e.g. we would need to link the main module itself with `-Wl,--rpath`.).   I'm not sure this is super important use case .. but we might want to make sure it at least fails with a good error message.

sbc100:
> I guess it's here: https://github.com/emscripten-core/emscripten/blob/main/system/lib/wasmfs/wasmfs.cpp?plain=1#L33 so if we dropped the wasmfs priority to 99 and then do the dynamic library preloading with 100 that would work.

Thats the kind of thing I was thinking we could try.

hoodmane:
Okay, I will add a test where the main module gets an rpath and a preloaded dynamic library and wasmfs so hopefully we can see it failing. Then see if I can work out this initializers approach.

hoodmane:
It seems that even the normal JS file system isn't set up when we preload dynamic libraries.

sbc100:
> It seems that even the normal JS file system isn't set up when we preload dynamic libraries.

I see.. so I guess that means that preloading of dynamic libraries cannot currently use `--embed-file`?

hoodmane:
@sbc100 Is it possible to do asynchronous work in static initializers? They are all called by `___wasm_call_ctors`? If so then we can't do anything asynchronous.

It seems like making this mechanism work for preloaded dynamic libraries with wasmfs requires either stack switching or some other way of splitting up `___wasm_call_ctors`. Making it work for the JS FS is less work but still nontrivial. So I think we should punt on these features. Especially since our use case doesn't require it.

hoodmane:
This seems to double the size of `test_codesize_hello_dylink`. Maybe we could add a setting to opt into this?

hoodmane:
@sbc100 would appreciate a new review on this.

src/settings.js
```diff
@@ -1135,6 +1135,10 @@ var MAIN_MODULE = 0;
 // [compile+link]
 var SIDE_MODULE = 0;
 
+// Include RPATH support.
+// [link]
+var SUPPORT_RPATH = 0;
```
sbc100, line 1140: I don't think we should need a new setting for this.  Dynamic linking is pretty expensive thing to do use... i can't imagine the marginal cost of rpath is very much, is it?

src/lib/libwasmfs.js
```diff
@@ -503,7 +503,7 @@ addToLibrary({
     return FS_mknod(path, mode, 0);
   },
 
-  $FS_writeFile__deps: ['_wasmfs_write_file', '$stackSave', '$stackRestore'],
+  $FS_writeFile__deps: ['_wasmfs_write_file', '$stackSave', '$stackRestore', 'malloc', 'free'],
```
sbc100, line 506: Is this related?   Maybe land this separately?

src/lib/libdylink.js
```diff
@@ -48,6 +48,17 @@ var LibraryDylink = {
     registerWasmPlugin();
     `,
   $preloadedWasm: {},
+
+  $replaceORIGIN__deps: ['$PATH'],
+  $replaceORIGIN: (parentLibPath, rpath) => {
```
sbc100, line 53: s/parentLibPath/parentLibName/ ? (path is sounds a little like it might be the directory in which that parent lives).

src/lib/libdylink.js
```diff
@@ -959,11 +980,46 @@ var LibraryDylink = {
      * @param {number=} handle
      * @param {Object=} localScope
      */`,
-  $loadDynamicLibrary: function(libName, flags = {global: true, nodelete: true}, localScope, handle) {
+  $loadDynamicLibrary: function(libName, flags = {global: true, nodelete: true}, localScope, rpath = {parentLibPath: '', paths: []}, handle) {
```
sbc100, line 983: Remove the default value here?   and maybe put `rpath` last, then maybe you don't need to change the loadDynamicLibrary calls below at all.

src/lib/libdylink.js
```diff
@@ -959,11 +980,46 @@ var LibraryDylink = {
...
+
+#if FILESYSTEM && SUPPORT_RPATH
+      var foundFile = false;
+      if (libName.startsWith("/")) {
+        try {
+          FS.lookupPath(libName);
+          foundFile = true;
+        } catch(e){}
+      } else if (runtimeInitialized) {
+        var runtimePathsAbs = (rpath.paths || []).map((p) => replaceORIGIN(rpath.parentLibPath, p));
```
sbc100, line 997: Write this whole block in `withStackSave(() => ..` rather then explicit save and restore?

src/lib/libdylink.js
```diff
@@ -959,11 +980,46 @@ var LibraryDylink = {
...
+        var origStack = stackSave();
+        var bufSize = 2*255 + 2;
```
sbc100, line 999: Where do these numbers come from?

src/lib/libdylink.js
```diff
@@ -959,11 +980,46 @@ var LibraryDylink = {
...
      */`,
-  $loadDynamicLibrary: function(libName, flags = {global: true, nodelete: true}, localScope, handle) {
+  $loadDynamicLibrary: function(libName, flags = {global: true, nodelete: true}, localScope, rpath = {parentLibPath: '', paths: []}, handle) {
 #if DYLINK_DEBUG
     dbg(`loadDynamicLibrary: ${libName} handle: ${handle}`);
     dbg(`existing: ${Object.keys(LDSO.loadedLibsByName)}`);
 #endif
+
+#if FILESYSTEM && SUPPORT_RPATH
+      var foundFile = false;
```
sbc100, line 990: Indentation looks wrong here.

Also, shouldn't the existing checks for `LDSO.loadedLibsByName[libName];` and / or `handle` come first?

src/lib/libdylink.js
```diff
@@ -1024,6 +1080,18 @@ var LibraryDylink = {
...
+#if FILESYSTEM && SUPPORT_RPATH
+      if (foundFile) {
+        var libData = FS.readFile(libName, {encoding: 'binary'});
+        if (libData) {
+#if DYLINK_DEBUG
+          dbg(`loaded library from filesystem: ${libName}`);
+#endif
+          return flags.loadAsync ? Promise.resolve(libData) : libData;
+        }
+      }
```
sbc100, line 1092: It seems a little odd to re-introduce this alternative method of loading libraries.

Also, this check for `foundFile` looks like its inside the `if (dso) ` block, but the settings of `foundFile` is outside?

src/settings.js
hoodmane: It doubles the size of the hello_dylink example. But I think in real cases the effect will be much smaller.

sbc100: Do you know where the size bloat is coming from?  Is the JS filesystem APIs?  

In any case i think we not add a new setting here.

hoodmane: I'll remove it and rebaseline and we can see the added symbols. I suspect it's from including more of the file system.

src/lib/libdylink.js
hoodmane: This change made closure angry. Any idea why?

<details><summary>Details</summary>
<p>

```
building:ERROR: Closure compiler completed with warnings and -Werror=closure enabled, aborting!

building:ERROR: /tmp/emscripten_temp_u9hrpjho/a.out.jso3.js:1232:80: WARNING - [JSC_WRONG_ARGUMENT_COUNT] Function loadDynamicLibrary: called with 5 argument(s). Function requires at least 1 argument(s) and no more than 4 argument(s).
  1232|     return metadata.neededDynlibs.reduce((chain, dynNeeded) => chain.then(() => loadDynamicLibrary(dynNeeded, flags, localScope, undefined, {
                                                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  1233|       parentLibPath: libName,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  1234|       paths: metadata.runtimePaths
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  1235|     })), Promise.resolve()).then(loadModule);
        ^^^^^^

/tmp/emscripten_temp_u9hrpjho/a.out.jso3.js:1237:43: WARNING - [JSC_WRONG_ARGUMENT_COUNT] Function loadDynamicLibrary: called with 5 argument(s). Function requires at least 1 argument(s) and no more than 4 argument(s).
  1237|   metadata.neededDynlibs.forEach(needed => loadDynamicLibrary(needed, flags, localScope, undefined, {
                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  1238|     parentLibPath: libName,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  1239|     paths: metadata.runtimePaths
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  1240|   }));
        ^^^^

/tmp/emscripten_temp_u9hrpjho/a.out.jso3.js:1452:10: WARNING - [JSC_OPTIONAL_ARG_AT_END] optional arguments must be at the end
  1452|        */ function loadDynamicLibrary(libName, flags = {
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  1453|   global: true,
        ^^^^^^^^^^^^^^^
...
  1572|   return true;
        ^^^^^^^^^^^^^^
  1573| }
        ^

0 error(s), 3 warning(s), 58.8% typed
```



</p>
</details>

src/lib/libwasmfs.js
hoodmane: https://github.com/emscripten-core/emscripten/pull/24036

src/lib/libdylink.js
hoodmane: Okay I fixed it by moving rpath into flags.

src/lib/libdylink.js
hoodmane: I added a comment:
```
// In dylink.c we use: `char buf[2*NAME_MAX+2];` and NAME_MAX is 255.
// So we use the same size here.
```

src/lib/libdylink.js
sbc100: You could also have fixed by updating the docstring I believe.

hoodmane:
@sbc100 CI is green again, would appreciate another review when you have a chance.

anutosh491:
Hey @hoodmane 

Haven't fully followed the work here (shall educate myself with the same soon). Curious if this ends up resolving the patch added on pyodide ? (I think the patch-1 was dealing with rpaths iirc)

ryanking13:
> Haven't fully followed the work here (shall educate myself with the same soon). Curious if this ends up resolving the patch added on pyodide ? (I think the patch-1 was dealing with rpaths iirc)

@anutosh491  Yes, we hope so

anutosh491:
Thanks @ryanking13 for the reply. Would be interested to know how (and possibly where) does pyodide put this to use. Could you maybe give me a brief use case for starters ?

ryanking13:
> Thanks @ryanking13 for the reply. Would be interested to know how (and possibly where) does pyodide put this to use. Could you maybe give me a brief use case for starters ?

@anutosh491 

We are using it [here](https://github.com/pyodide/pyodide/blob/aa8a92abe7772202460615fa98c996265a8582e0/src/js/dynload.ts#L141-L155) when loading shared libraries. There, we directly call `loadDynamicLibrary` API instead of calling dlopen or emscripten_dlopen, and pass the modified version of fs parameter to let emscripten locate the dependencies. After this PR is landed, we want to replace that call with `emscripten_dlopen` or something similar so we don't need to rely on the internal emscripten APIs.

hoodmane:
@sbc100 would appreciate another review on this.

sbc100:
@hoodmane is the PR description up-to-date?

hoodmane:
Not sure, will check.

system/lib/libc/dynlink.c
```diff
@@ -484,41 +484,55 @@ static void dlopen_onerror(struct dso* dso, void* user_data) {
...
+      s += l;
     }
-    s += l;
   }
+  return -1;
 }
 
 // Resolve filename using LD_LIBRARY_PATH
-static const char* resolve_path(char* buf, const char* file, size_t buflen) {
+const char* _emscripten_resolve_path(char* buf, const char* rpath, const char* file, size_t buflen) {
```
sbc100, line 530: Can we name this something dylink specific?  How about `_emscripten_find_dylib` or `_emscripten_resolve_dylib_path`?

test/test_other.py
```diff
@@ -7697,6 +7697,137 @@ def test_ld_library_path(self, args):
     self.assertContained('Hello4', out)
     self.assertContained('Ok', out)
 
+  @also_with_wasmfs
+  def test_dlopen_library_path(self):
+    create_file('hello1_dep.c', r'''
+#include<stdio.h>
```
sbc100, line 7703: Space after `include`

test/test_other.py
```diff
@@ -7697,6 +7697,137 @@ def test_ld_library_path(self, args):
...
+
+void hello1_2() {
+  printf ("Hello1_2\n");
```
sbc100, line 7706: No space after printf

sbc100: How about calling this function `hello_dep` .. 1_2 seems a little cryptic.

test/test_other.py
```diff
@@ -7697,6 +7697,137 @@ def test_ld_library_path(self, args):
...
+    self.run_process([EMCC, '-o', 'hello1.wasm', 'hello1.c', '-sSIDE_MODULE', 'subdir/libhello1_dep.so'])
+    self.run_process([EMCC, '--profiling-funcs', '-o', 'main.js', 'main.c', '-sMAIN_MODULE=2', '-sINITIAL_MEMORY=32Mb',
+                      '--embed-file', 'hello1.wasm@/lib/libhello1.wasm',
+                      '--embed-file', 'subdir/libhello1_dep.so@/usr/lib/libhello1_dep.so',
+                      'hello1.wasm', '-sNO_AUTOLOAD_DYLIBS',
+                      '-L./subdir', '-lhello1_dep', '--pre-js', 'pre.js'])
+    out = self.run_js('main.js')
+    self.assertContained('Hello1', out)
+    self.assertContained('Hello1_2', out)
+    self.assertContained('Ok', out)
```
sbc100, line 7761: How about just `self.assertContained('Hello1\nHello1_2\nOk\n', out)`?

Or maybe even use `self.do_runf('main.c', 'Hello1\nHello1_2\nOk\n', emcc_args=args)`

test/test_other.py
```diff
@@ -7697,6 +7697,137 @@ def test_ld_library_path(self, args):
...
+  def test_dlopen_rpath(self):
+    create_file('hello1_dep.c', r'''
+#include<stdio.h>
+
+void hello1_2() {
+  printf ("Hello1_2\n");
+  return;
+}
+''')
+    create_file('hello1.c', r'''
```
sbc100, line 7773: Ditto, lets just call this `hello`.   Or use `libfoo.c` which exports `foo`?

test/test_other.py
```diff
@@ -7697,6 +7697,137 @@ def test_ld_library_path(self, args):
     self.assertContained('Hello4', out)
     self.assertContained('Ok', out)
 
+  @also_with_wasmfs
+  def test_dlopen_library_path(self):
```
sbc100, line 7701: How is this different to the existing `test_ld_library_path`?

src/lib/libdylink.js
```diff
@@ -1029,6 +1055,56 @@ var LibraryDylink = {
...
+#if DYLINK_DEBUG
+      dbg(`checking filesystem: ${libName}: ${foundFile ? 'found' : 'not found'}`);
+#endif
+      if (foundFile) {
+        var libData = FS.readFile(libName, {encoding: 'binary'});
+        if (libData) {
+#if DYLINK_DEBUG
+          dbg(`loaded library from filesystem: ${libName}`);
+#endif
+          return flags.loadAsync ? Promise.resolve(libData) : libData;
```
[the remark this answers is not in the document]
hoodmane: Well in unchanged lines in the same function we do:
```js
      if (sharedMod) {
        return flags.loadAsync ? Promise.resolve(sharedMod) : sharedMod;
      }
```
presumably either both should change or neither.

src/lib/libdylink.js
```diff
@@ -890,6 +901,7 @@ var LibraryDylink = {
       return postInstantiation(module, instance);
     }
 
+    flags = Object.assign(flags, {rpath: { parentLibPath: libName, paths: metadata.runtimePaths }})
```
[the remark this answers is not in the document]
sbc100: Oh I see.  I in that case how about `flags = {...flags, rpath: { parentLibPath: libName, paths: metadata.runtimePaths }}`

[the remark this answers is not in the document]
sbc100: Maybe worth a comment too.. i.e. why copy and not mutate.

src/lib/libdylink.js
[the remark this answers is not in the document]
sbc100: Fair enough.  Leave it as is and we can consider changing later.

src/lib/libdylink.js
```diff
@@ -1029,6 +1055,56 @@ var LibraryDylink = {
...
+      if (foundFile) {
+        var libData = FS.readFile(libName, {encoding: 'binary'});
+        if (libData) {
+#if DYLINK_DEBUG
+          dbg(`loaded library from filesystem: ${libName}`);
+#endif
+          return flags.loadAsync ? Promise.resolve(libData) : libData;
+        }
+      }
+#endif
```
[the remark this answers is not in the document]
hoodmane: Yeah, this leads to several nice simplifications and two fewer layers of nested blocks in one spot.

test/test_other.py
hoodmane: I don't see any important differences, so I think we can remove this one.

hoodmane: Oh, the new behavior is that we now correctly handle dependencies between dynamic libraries -- before, if a lib depends on another one, the dlopened lib could be loaded from `LD_LIBRARY_PATH` but not the dependency.

sbc100: Can you extend the existing test maybe?

hoodmane: I combined the two tests into one.

src/lib/libdylink.js
```diff
@@ -1029,6 +1055,56 @@ var LibraryDylink = {
...
+#if WASMFS
+        var result = withStackSave(() => __wasmfs_identify(stringToUTF8OnStack(libName)));
+        foundFile = result === {{{ cDefs.EEXIST }}};
+#else
+        try {
+          FS.lookupPath(libName);
+          foundFile = true;
+        } catch (e) {}
+#endif
+      } else if (runtimeInitialized) {
```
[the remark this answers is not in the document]
hoodmane: Actually, I'm not sure that either path works. I think this guard should probably be moved up. I'll double check.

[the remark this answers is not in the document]
hoodmane: Indeed. I moved up the guard to cover both absolute and relative paths and added a comment.

system/lib/libc/dynlink.c
```diff
@@ -484,41 +484,55 @@ static void dlopen_onerror(struct dso* dso, void* user_data) {
 }
 
 // Modified version of path_open from musl/ldso/dynlink.c
-static int path_find(const char *name, const char *s, char *buf, size_t buf_size) {
+static int path_find(const char *name, int ncandidates, const char **candidates, char *buf, size_t buf_size) {
```
sbc100, line 487: Can we leave this function unmodified and instead write a wrapper function around it?

Also, I'm not sure about the name candidates, these are two different paths I guess?

Maybe just call `path_find` twice rather than being overly general like this?

hoodmane:
`test_ld_library_path_pthread` is broken because `FS.readFile()` doesn't work there. I guess we have to solve it with proxying.

sbc100:
> `test_ld_library_path_pthread` is broken because `FS.readFile()` doesn't work there. I guess we have to solve it with proxying.

Hmm.. yes maybe use `FS_readFile` and mark that with `__proxy` attribute.

src/lib/libdylink.js
```diff
@@ -932,6 +947,66 @@ var LibraryDylink = {
...
+#else
+      try {
+        FS.lookupPath(libName);
+        return libName;
+      } catch (e) {
+        return undefined;
+      }
+#endif
+    }
+    var runtimePathsAbs = (rpath?.paths || []).map((p) => replaceORIGIN(rpath?.parentLibPath, p));
```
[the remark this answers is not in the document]
hoodmane: Well we already do have a variable called `rpath`...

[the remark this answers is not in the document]
hoodmane: I'll go with `rpathResolved`.

src/lib/libdylink.js
```diff
@@ -932,6 +947,66 @@ var LibraryDylink = {
...
+      var buf = stackAlloc(bufSize);
+      var size = 0;
+      for (var str of runtimePathsAbs) {
+        size += lengthBytesUTF8(str) + 1;
+      }
+      var rpathC = stackAlloc(size);
+      var cur = rpathC
+      for (var str of runtimePathsAbs) {
+        cur += stringToUTF8(str, cur, size);
+        HEAP8[cur] = ':'.charCodeAt(0);
```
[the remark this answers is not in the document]
hoodmane: That is indeed quite a bit shorter.

system/lib/libc/dynlink.c
hoodmane: How's it look to you now?

src/lib/libdylink.js
```diff
@@ -932,6 +947,59 @@ var LibraryDylink = {
...
+    var rpathResolved = (rpath?.paths || []).map((p) => replaceORIGIN(rpath?.parentLibPath, p));
+    return withStackSave(() => {
+      // In dylink.c we use: `char buf[2*NAME_MAX+2];` and NAME_MAX is 255.
+      // So we use the same size here.
+      var bufSize = 2*255 + 2;
+      var buf = stackAlloc(bufSize);
+      var rpathC = stringToUTF8OnStack(rpathResolved.join(':'));
+      var libNameC = stringToUTF8OnStack(libName);
+      var resLibNameC = __emscripten_find_dylib(buf, rpathC, libNameC, bufSize);
+      return (resLibNameC !== libNameC) ? UTF8ToString(resLibNameC) : undefined;
```
sbc100, line 998: Should __emscripten_find_dylib return NULL when it can't find the file?   Then you don't need this pointer comparison?

hoodmane:
But the arguments to/return value of `FS_readFile` aren't pointers.

src/lib/libdylink.js
sbc100: Oh I guess its used internally so maybe we don't want to change the interface.

hoodmane: I'll have to check what the other callers of `_emscripten_find_dylib` expect.

test/test_other.py
```diff
@@ -7653,23 +7664,85 @@ def test_ld_library_path(self, args):
...
+  f = dlsym(h, "hello");
+  assert(f);
+  f();
+  dlclose(h);
+
+  printf("Ok\n");
+
+  return 0;
+}
+''')
```
sbc100, line 7728: Can you indent all this code 4 spaces?  (the test running will strip the leading whitespace when it creates the file on disc).

sbc100 (approved):
LGTM with some final comments.

Thanks for bearing with me on this one.. i know it was a lot of back and forth.

sbc100:
> > `test_ld_library_path_pthread` is broken because `FS.readFile()` doesn't work there. I guess we have to solve it with proxying.
> 
> Hmm.. yes maybe use `FS_readFile` and mark that with `__proxy` attribute.

Oh I see.. so we cannot proxy it :(

Maybe just skip the test for now?   And we can try to fix it as a followup?  My understanding was that all this loading should happen on the main thread anyway since he original dlopen is always proxied back, but maybe I'm misremembering.

src/lib/libdylink.js
hoodmane: Yeah I guess I can add a wrapper function for calling from C that keeps the current behavior so that JS can get what it wants and C can get what it wants too.

sbc100:
> > > `test_ld_library_path_pthread` is broken because `FS.readFile()` doesn't work there. I guess we have to solve it with proxying.
> > 
> > 
> > Hmm.. yes maybe use `FS_readFile` and mark that with `__proxy` attribute.
> 
> Oh I see.. so we cannot proxy it :(
> 
> Maybe just skip the test for now? And we can try to fix it as a followup? My understanding was that all this loading should happen on the main thread anyway since he original dlopen is always proxied back, but maybe I'm misremembering.

Yes, IIRC all the initial loading should happen on the main thread and then background threads should should be able mirror all the same loads but hopfully we can avoid re-resolving all the pathnames on the background threads since them main thread already resolved them.

src/lib/libdylink.js
hoodmane: It's a bit annoying that in C, `x || y` isn't the same as `x ? x : y` like it is in JS.

hoodmane:
Shouldn't the side threads hit the `if (handle) {` condition and never make it to this logic?

hoodmane:
Okay I skipped `test_ld_library_path` in pthreads and added a `WASMFS` variant to it.

ryanking13:
Thanks so much for finishing the PR @hoodmane @sbc100! I know it was a lot of work and I really appreciate it. While this change is not visible to the users right now, I believe this is going to be a big progress in Pyodide.

PR #23877, opened 2025-03-10
Remove magic `=== Body ===` tag. NFC

sbc100:


hoodmane (approved):
Not sure what `do_replace` does, but it looks like a readability improvement to me.

Issue #23875, opened 2025-03-10
`addOnPreRun` no longer works in EM_JS top level code after #23565


hoodmane:
#23565 moved the definition of `addOnPreRun` slightly lower in the the output file so it cannot be used by top level `EM_JS` code anymore. 

#### Version of emscripten/emsdk

This worked on version 4.0.2 or 39ca93535a83eb6d071a1c27390f032a8285414d. Does not work on 4.0.3 or a4f64a37229d06a810e27fe98f0b0db7e110a3eb.

#### Code

```C
#include "emscripten.h"


EM_JS(int, f, (void), {}
addOnPreRun(() => console.log("hi!"))
)

int main() {
    return 0;
}
```

Compile, link, and run:
```sh
emcc a.c && node a.out.js
```
Before: prints "hi" and exist with status 0
After: Fails with "ReferenceError: addOnPreRun is not defined"

hoodmane:
The real code where we are hitting this problem is here:
https://github.com/python/cpython/blob/main/Python/emscripten_trampoline.c#L169-L174

sbc100:
I think this was likely due to #23565.

Does adding `$addOnPreRun` to `DEFAULT_LIBRARY_FUNCS_TO_INCLUDE` fix the issue?

Even better I think you can do `EM_JS_DEPS(deps, '$stringToUTF8')` in the C source.

hoodmane:
I will try this, that would be great if it works.

hoodmane:
I get:
```
error: undefined symbol: $addOnPreRun (referenced by root reference (e.g. compiled C/C++ code))
warning: To disable errors for undefined symbols use `-sERROR_ON_UNDEFINED_SYMBOLS=0`
warning: addOnPreRun may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
Internal compiler error in src/compiler.mjs!
Please create a bug report at https://github.com/emscripten-core/emscripten/issues/
with a log of the build and the input files used to run. Exception message: "node:internal/assert/utils:281
    throw err;
    ^

AssertionError [ERR_ASSERTION]: runtimeElements contains library symbol: $addOnPreRun
    at exportRuntime (file:///home/rchatham/Documents/programming/emscripten/src/modules.mjs:508:7)
    at postamble.js:1:2
    at Script.runInContext (node:vm:149:12)
    at Module.runInContext (node:vm:301:6)
[...]

PR #23878, opened 2025-03-10
Emit EM_JS/EM_ASM code after JS library code

sbc100:
This change fixes an issue with trailing code in EM_JS macros.  See #23875 and #23884

The other advantage of this ordering is that it removes the needs for the magic macro here.


kripken (commented):
Is this change worth the risk? I agree the existing usages of this are unintended, but the benefit of this PR seems fairly minor. Keeping existing code working, even if odd, seems more important to me.

sbc100:
The problem is that we broke existing code already:  #23875.  So that only case we know of where the order matters, it is currently wrong.

kripken:
Oh, then does this PR not make it any more or less wrong? Is it NFC then? (or at least NFC modulo the broken cases?)

sbc100:
I believe there are two ways this problem could be visible to users:

1. Abusing EM_JS to inject top level code which refernces JS library funcs.
2. Using a `__postset` in a JS library func to refer to en EM_JS/EM_ASM function.

Its really hard to imagine (2) happening since `__postset` code cannot generally call native functions (since the module is not loaded yet).  EM_JS/EM_ASM and kind of like "native" functions, even though they are technically available before the module is loaded.

kripken:
> This PR fixes the issue in https://github.com/emscripten-core/emscripten/issues/23875.

The issue there, from the first sentence:

> https://github.com/emscripten-core/emscripten/pull/23565 moved the definition of addOnPreRun slightly lower in the the output file so it cannot be used by top level EM_JS code anymore.

But you are saying that top-level EM_JS code will *not* work even after this PR? Or did I misunderstand?

If this *does* fix top-level EM_JS code, should the PR add a test? (If not, should that other thing get a test?)

sbc100:
Added a new test for trailing EM_JS code, and a new bug.

kripken (approved):
Thanks! Now I see.

Yes, this seems best - we can consider removing this later, but having the test will prevent us from doing so by mistake at least.

hoodmane (approved):
Thanks @sbc100! This looks good to me.

It would be nice to either decide to officially support this or as @brendandahl suggested add a separate way to inject top level code like `EM_JS_POSTSET` and/or `EM_JS_TOPLEVEL` but for now this is working for us.

PR #23882, opened 2025-03-10
Remove pthread `alert` polyfill

sbc100:
I'm not sure why we were polyfilling this web API in all our workers. We don't generally do this for other APIs.


kripken:
This is for debugging, I believe - it allows you to write `alert('foo')` in a worker, and have it actually show a synchronous alert to the user. I remember it being useful in the past, but maybe rarely?

Perhaps we can keep it in ASSERTIONS builds?

sbc100:
> This is for debugging, I believe - it allows you to write `alert('foo')` in a worker, and have it actually show a synchronous alert to the user. I remember it being useful in the past, but maybe rarely?
>
> Perhaps we can keep it in ASSERTIONS builds?

Are you sure its worth keeping?

I've never used it myself in all my years of pthread development.   Normally `err(..)` and `dbg(..)` are way more useful.  `dbg()` will even log the worker ID and the thread ID these days.

sbc100:
> > This is for debugging, I believe - it allows you to write `alert('foo')` in a worker, and have it actually show a synchronous alert to the user. I remember it being useful in the past, but maybe rarely?
> > Perhaps we can keep it in ASSERTIONS builds?
> 
> Are you sure its worth keeping?
> 
> I've never used it myself in all my years of pthread development. Normally `err(..)` and `dbg(..)` are way more useful. `dbg()` will even log the worker ID and the thread ID these days.

Once problem with `alert('foo')` if you use the implementation here is that it does not block the progress of the worker.  Normally one would expect the worker to be blocked until the user clicks the alert.. but doing it this way is fire-and-forget and the worker keeps running.. which seems confusing.

sbc100:
> > > This is for debugging, I believe - it allows you to write `alert('foo')` in a worker, and have it actually show a synchronous alert to the user. I remember it being useful in the past, but maybe rarely?
> > > Perhaps we can keep it in ASSERTIONS builds?
> > 
> > 
> > Are you sure its worth keeping?
> > I've never used it myself in all my years of pthread development. Normally `err(..)` and `dbg(..)` are way more useful. `dbg()` will even log the worker ID and the thread ID these days.
> 
> Once problem with `alert('foo')` if you use the implementation here is that it does not block the progress of the worker. Normally one would expect the worker to be blocked until the user clicks the alert.. but doing it this way is fire-and-forget and the worker keeps running.. which seems confusing.

Maybe we should call it `asyncAlert` :)

kripken:
Heh, yeah...

We could use a busy-wait to make it fully sync... that would be better for debugging. Anyhow, if we want such a thing we can add it as an API somehow later. I agree it makes sense to remove this hackish thing.

kripken (approved):

PR #23887, opened 2025-03-11
Fill correct stack info in pthread_getattr_np stub

hoodmane:
See use case here:
https://github.com/python/cpython/pull/130398/files#diff-c22186367cbe20233e843261998dc027ae5f1f8c0d2e778abfa454ae74cc59deR365-R380

I guess for stack protection, this only enables introspecting whether we run out of the spill stack. I'm not sure whether to expect running out of spill stack or running out of true stack to be more common. But at least this unbreaks the code.


test/other/test_stub_pthread_getattr_np.c
```diff
@@ -0,0 +1,17 @@
+#define _GNU_SOURCE
+#include "pthread.h"
+#include "stdio.h"
+#include "assert.h"
+
+int main() {
+    pthread_attr_t attr;
```
sbc100, line 7: Two space indentation.

test/test_other.py
```diff
@@ -15804,3 +15804,6 @@ def test_rollup(self):
 
   def test_rlimit(self):
     self.do_other_test('test_rlimit.c', emcc_args=['-O1'])
+
+  def test_stub_pthread_getattr_np(self):
+    self.do_runf(test_file('other/test_stub_pthread_getattr_np.c'), 'stack_addr: 0, stack_size: 65536, guard_size: 8192')
```
sbc100, line 15809: No need for `test_file` here.

sbc100: Maybe use `.out` file alongside the `.c` file?

Then you would use `do_other_test`... and you can run with `--rebase` to generate the .out content.

sbc100: LGTM either way though.

sbc100 (approved):

test/test_other.py
hoodmane: `--rebase` is convenient.

sbc100 (approved):

system/lib/pthread/library_pthread_stub.c
```diff
@@ -277,6 +277,9 @@ int pthread_condattr_setpshared(pthread_condattr_t *attr, int shared) {
 }
 
 int pthread_getattr_np(pthread_t thread, pthread_attr_t *attr) {
+  attr->_a_stackaddr = emscripten_stack_get_base();
+  attr->_a_stacksize = emscripten_stack_get_base() - emscripten_stack_get_end();
+  attr->_a_guardsize = __default_guardsize;
```
sbc100, line 282: I wonder if we can just remove this stub and use `musl/src/thread/pthread_getattr_np.c` in the stub pthread library?

test/test_other.py
```diff
@@ -15807,6 +15807,9 @@ def test_rollup(self):
   def test_rlimit(self):
     self.do_other_test('test_rlimit.c', emcc_args=['-O1'])
 
+  def test_stub_pthread_getattr_np(self):
+    self.do_other_test('test_stub_pthread_getattr_np.c')
```
sbc100, line 15811: Actually can this simply be combined with the existing `test_pthread_stub` in this fvile?  (Or at least co-located with it).

system/lib/pthread/library_pthread_stub.c
hoodmane: Good idea, I did this.

hoodmane:
The closure tests are now failing for me locally with:
```

building:ERROR: /tmp/emtest_aihv0i03/emscripten_temp_z4b7qfms/a.out.js:121:4: ERROR - [JSC_UNDEFINED_VARIABLE] variable assert is undeclared
  121|     assert(typeof data == 'object');
```

system/lib/pthread/pthread_self_stub.c
```diff
@@ -28,4 +29,7 @@ __attribute__((constructor))
 static void init_pthread_self(void) {
   __main_pthread.locale = &libc.global_locale;
   __main_pthread.tid = getpid();
+  __main_pthread.stack = (void*)emscripten_stack_get_base();
+  __main_pthread.stack_size = emscripten_stack_get_base();
```
sbc100, line 33: Should this be `emscripten_stack_get_end() - emscripten_stack_get_base()`

hoodmane: Indeed...

sbc100: Can you see if using the linker generated symbol `&__stack_high`  and `&__stack_high - &__stack_low` help with the regression?   (Along with `extern int __stack_high; extern int __stack_low;`)

hoodmane: Using:
```
-  __main_pthread.stack = (void*)emscripten_stack_get_base();
-  __main_pthread.stack_size = emscripten_stack_get_base() - emscripten_stack_get_end();
+  __main_pthread.stack = (void*)&__stack_low;
+  __main_pthread.stack_size = &__stack_high - &__stack_low;
```
I get:
```
-stack_addr: 0, stack_size: 65536, guard_size: 8192
+stack_addr: 0xffffc000, stack_size: 16384, guard_size: 8192
```
So it seems slightly different? Looking at the code I can't understand why.

sbc100:
I experimented with using static struct initializer instead but it looks like it actually caused and overall regression: https://github.com/emscripten-core/emscripten/pull/23904

system/lib/pthread/pthread_self_stub.c
hoodmane: Also, for `test_pthread_stub`, when I use `emscripten_stack_get_base` and `emscripten_stack_get_end` it comes out 9 bytes smaller...

sbc100: That seems odd, how did you declare `__stack_low` and `__stack_high`?

hoodmane: Indeed, with `-Os` the `&__stack_high - &__stack_low;` code comes out 22 bytes smaller.

hoodmane: Okay got it working with these.

sbc100 (approved):
Nice!  LGTM

hoodmane:
I guess this is still a few bytes bigger than before because we added `pthread_getattr_np.c`/

test/other/codesize/test_codesize_cxx_ctors1.size
```diff
@@ -1 +1 @@
-129166
+129206
```
sbc100, line 1: Seems like a 40 byte increase is more than I would expect from this change.. I was hoping for more like 6 bytes (2 bytes each for the 3 extra store instructions).

hoodmane: Well we also added `pthread_getattr_np.c` which looks like:
```C
int pthread_getattr_np(pthread_t t, pthread_attr_t *a)
{
	*a = (pthread_attr_t){0};
	a->_a_detach = t->detach_state>=DT_DETACHED;
	a->_a_guardsize = t->guard_size;
	a->_a_stackaddr = (uintptr_t)t->stack;
	a->_a_stacksize = t->stack_size;
	return 0;
}
```
Presumably the other 34 bytes come from that?

sbc100: But that file should not linking into any of those tests.  It should only be linked into the (rare) programs that use it.

PR #23889, opened 2025-03-11
Use simple import.meta.url for worker creation

JoeOsborn:
This needs to be tested under various bundlers.  Spun out from #23804 .

### DIFF
--- a/src/lib/libpthread.js
+++ b/src/lib/libpthread.js
@@ -424,12 +424,7 @@ var LibraryPThread = {
         worker = new Worker(p.createScriptURL('ignored'), {{{ pthreadWorkerOptions }}});
       } else
 #endif
-      // We need to generate the URL with import.meta.url as the base URL of the JS file
-      // instead of just using new URL(import.meta.url) because bundler's only recognize
-      // the first case in their bundling step. The latter ends up producing an invalid
-      // URL to import from the server (e.g., for webpack the file:// path).
-      // See https://github.com/webpack/webpack/issues/12638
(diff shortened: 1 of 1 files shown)


kleisauke:
Looking at the test failures in `test-other`, I think it needs to be `new Worker(new URL(import.meta.url))`, at least on Node.js.
```
TypeError [ERR_WORKER_PATH]: The worker script or module filename must be an absolute path or a relative path starting with './' or '../'. Wrap file:// URLs with `new URL`. Received "file:///tmp/emtest_zpd2a426/emscripten_test_other_ybmo_i00/subdir/module.mjs"
```

We could perhaps use `new Worker(import.meta.url)` when `ENVIRONMENT_MAY_BE_WEB && !ENVIRONMENT_MAY_BE_NODE`, but this code section already has quite a few conditional expressions. :sweat_smile:

sbc100:
> Looking at the test failures in `test-other`, I think it needs to be `new Worker(new URL(import.meta.url))`, at least on Node.js.
> 
> ```
> TypeError [ERR_WORKER_PATH]: The worker script or module filename must be an absolute path or a relative path starting with './' or '../'. Wrap file:// URLs with `new URL`. Received "file:///tmp/emtest_zpd2a426/emscripten_test_other_ybmo_i00/subdir/module.mjs"
> ```
> 
> We could perhaps use `new Worker(import.meta.url)` when `ENVIRONMENT_MAY_BE_WEB && !ENVIRONMENT_MAY_BE_NODE`, but this code section already has quite a few conditional expressions. 😅

Or we could improve our `new Worker` polyfill on node so that it support string as a well as URLs.

PR #23890, opened 2025-03-11
Use mainScriptUrlOrBlob if present under EXPORT_ES6

JoeOsborn:
Spun out from #23804 .

This is a fix for #23769 which was introduced by #23171 , where the use of the bundler pattern hard-coding a script name was made to apply to all ES6 exports.

`mainScriptUrlOrBlob` is necessary under es6 exports to support running a threaded emscripten module from another origin, e.g. to distribute an emscripten module on a CDN (the script will load for the main thread but won’t be allowed to create workers). Using blobs is an effective workaround.


sbc100:
Can we add a test for this.

Otherwise lgtm.

sbc100:
Can you mention that is a fix for a regression that was introduced #23171?  (Assuming that is true?)

sbc100:
Could you also perhaps mention (in a comment) why/when this might be needed /used?

JoeOsborn:
Indeed this is a fix for #23769 which was introduced by #23171 , where the use of the bundler pattern hard-coding a script name was made to apply to all ES6 exports.

I’ll look into adding tests for vite and webpack.

JoeOsborn:
And it’s necessary to support running a threaded emscripten module from another origin, e.g. to distribute an emscripten module on a CDN (the script will load for the main thread but won’t be allowed to create workers).  Using blobs instead allows a workaround.

sbc100:
Can you put all those detail in the PR description?  

In terms of testing I think it should be fine to just extend the existing testing we have for `mainScriptUrlOrBlob`in test_browser.py (test_load_js_from_blob_with_pthreads).

sbc100:
I created a PR to increase the testing we have for mainScriptUrlOrBlob

JoeOsborn:
Yes, cool! I am working on tests for pthread as blob and for webpack/vite, but I’m encountering something weird in my vite test.  I want to instantiate the module twice (once with a string mainUrlOrBlob and once with a blob) within one test, but I’m getting two report_result outputs and the test harness doesn’t like it.  Is there a way I can skip the generation of one or both of those `report_result?exit=0` or expect to see two of them?

src/lib/libpthread.js
```diff
@@ -423,6 +423,15 @@ var LibraryPThread = {
...
         worker = new Worker(p.createScriptURL('ignored'), {{{ pthreadWorkerOptions }}});
       } else
+#endif
+#if expectToReceiveOnModule('mainScriptUrlOrBlob')
+        if (Module['mainScriptUrlOrBlob']) {
+          var pthreadMainJs = Module['mainScriptUrlOrBlob'];
+          if (typeof pthreadMainJs != 'string') {
+            pthreadMainJs = URL.createObjectURL(pthreadMainJs);
+          }
+          worker = new Worker(new URL(pthreadMainJs, import.meta.url), {{{ pthreadWorkerOptions }}});
```
sbc100, line 433: Can you just do `new Worker(pthreadMainJs, {{{ pthreadWorkerOptions }}})` like below?

JoeOsborn:
As a quick update, I’m pretty sure I can revise this PR with tests tomorrow.  There are some issues with the paths used by the web server giving 404s under some conditions but not others, but I’m confident I can resolve these.

sbc100:
Instead of creating a new browser test can you just update the new `test_mainScriptUrlOrBlob` in `test_other.py` ?  I think it will make testing much simpler.

test/test_browser.py
```diff
@@ -5536,19 +5547,30 @@ def test_error_reporting(self):
...
+    elif useblob:
+      copytree(test_file('webpack_with_blob'), '.')
+      self.emcc_args += ['-pthread', '-sPTHREAD_POOL_SIZE=1']
+      outfile = 'src/hello.js'
     else:
       copytree(test_file('webpack'), '.')
       outfile = 'src/hello.js'
     self.compile_btest('hello_world.c', ['-sEXIT_RUNTIME', '-sMODULARIZE', '-sENVIRONMENT=web,worker', '-o', outfile])
     self.run_process(shared.get_npm_cmd('webpack') + ['--mode=development', '--no-devtool'])
+    shutil.copy(outfile, 'dist/')
```
sbc100, line 5573: I think the idea is that the bundler should do this.. so you should not have to do this here.  In fact, if you have to do this that would seem like a bug maybe?   (Same with the .wasm file below I guess).

In any case, I think this bundler stuff is orthogonal the this change which I think we can test without modifying any bundler tests.

JoeOsborn: Sorry, this one wasn't meant to be included here.

JoeOsborn:
Sorry my commits got messy there, I accidentally ran a commit -a without noticing.

JoeOsborn:
> Instead of creating a new browser test can you just update the new `test_mainScriptUrlOrBlob` in `test_other.py` ? I think it will make testing much simpler.

I ended up modifying a different existing browser test specifically about pthread blob support.

sbc100:
> > Instead of creating a new browser test can you just update the new `test_mainScriptUrlOrBlob` in `test_other.py` ? I think it will make testing much simpler.
> 
> I ended up modifying a different existing browser test specifically about pthread blob support.

BTW, `test_mainScriptUrlOrBlob` is also specifically about pthread blob support, but it doesn't need the browser to run.   I think it should certainly be updated as part of this PR.

test/pthread/main_js_as_url_loader_es6.html
```diff
@@ -0,0 +1,158 @@
...
+    </script>
+
+    <script type="module">
+      Module['mainScriptUrlOrBlob'] = 'hello_thread_with_blob_url.mjs';
+      import('./'+Module['mainScriptUrlOrBlob']).then((fac) => fac.default(Module).then((instance) => {
+        console.log("success");
+      }))
+    </script>
+  </body>
+</html>
```
sbc100, line 158: It seems sad (hard to maintain anyway) to add three new almost-identical files like this.

How about we just use a single `.html` file with something like this at the end:

```
<!-- For any given test exactly one of these two scripts will exist -->
<script type="module" src="loader.mjs">
<script src="loader.js">
```

test/test_browser.py
```diff
@@ -4758,15 +4758,26 @@ def test_pthread_reltime(self):
...
   # Tests that it is possible to load the main .js file of the application manually via a Blob URL,
   # and still use pthreads.
-  def test_load_js_from_blob_with_pthreads(self):
+  @parameterized({
+    'blob': ('main_js_as_blob_loader.html',False),
+    'url': ('main_js_as_url_loader.html',False),
+    'blob_es6': ('main_js_as_blob_loader_es6.html',True),
+    'url_es6': ('main_js_as_url_loader_es6.html',True),
+  })
+  def test_load_js_from_blob_with_pthreads(self, which_html, es6):
```
sbc100, line 4767: I guess that `blob` part of this test name is no longer correct?

Should we rename this test to `test_mainScriptUrlOrBlob` like the `test_other.py` test?

test/pthread/main_js_as_url_loader_es6.html
JoeOsborn: I guess we'll need four scripts/script tags, but I see your point

test/test_browser.py
sbc100: Since `mainScriptUrlOrBlob` is really only a pthread feature I think you can just call it `test_mainScriptUrlOrBlob`, like in `test_other.py`.     (Or rename the one in test_other.py.. either way they probably should be called the same thing).

test/pthread/main_js_as_url_loader_es6.html
sbc100: I think you can probably just inline those in the test rather then creating 4 new files in the repo:

```
create_file('loader' + EXT, '''
Module['mainScriptUrlOrBlob'] = 'hello_thread_with_blob_url.mjs';
  import('hello_thread_with_blob_url.mjs').then((fac) => fac.default(Module).then((instance) => {
     console.log("success");
}))
''')
```

test/test_browser.py
```diff
@@ -4756,18 +4756,61 @@ def test_pthread_growth(self, emcc_args):
...
+  @parameterized({
+    'blob': (False, True),
+    'url': (False, False),
+    'blob_es6': (True, True),
+    'url_es6': (True, False),
+  })
+  def test_mainScriptUrlOrBlob(self, es6, use_blob):
     # TODO: enable this with wasm, currently pthreads/atomics have limitations
     self.set_setting('EXIT_RUNTIME')
+    js_name = 'hello_thread_with_loader.js'
```
sbc100, line 4770:
```
js_name = 'hello_thread_with_loader' + ext
```

(so you only need to set it once)

sbc100 (commented):
Nice!  Thanks for doing that.

Can you update the test in test_other too with an es6 module variant? 

Otherwise lgtm!

test/test_browser.py
```diff
@@ -4756,18 +4756,61 @@ def test_pthread_growth(self, emcc_args):
...
+      ext = '.mjs'
+      self.emcc_args += ['-sEXPORT_ES6']
+      js_name = 'hello_thread_with_loader.mjs'
+    if es6 and use_blob:
+      create_file('loader.mjs', '''
+        Module['locateFile'] = (path,_prefix) => path;
+        fetch('hello_thread_with_loader.mjs').then((rsp) => rsp.blob().then((blob) => {
+          Module['mainScriptUrlOrBlob'] = blob;
+          import(URL.createObjectURL(blob)).then((fac) => fac.default(Module))
+        }));
```
sbc100, line 4782: Since this is ES6 I guess you could use `await` there to make this a little more readable?   Doesn't make much difference I guess?

test/test_browser.py
```diff
@@ -4756,18 +4756,61 @@ def test_pthread_growth(self, emcc_args):
...
+        Module['mainScriptUrlOrBlob'] = blob;
+        var script = document.createElement('script');
+        script.src = URL.createObjectURL(blob);
+        document.body.appendChild(script);
+        }));
+      ''')
+    elif es6:
+      create_file('loader.mjs', '''
+        Module['mainScriptUrlOrBlob'] = 'hello_thread_with_loader.mjs';
+        import('./'+Module['mainScriptUrlOrBlob']).then((fac) => fac.default(Module))
```
sbc100, line 4797: Just `./hello_thread_with_loader.mjs` maybe?

test/pthread/main_js_with_loader.html
```diff
@@ -147,28 +147,8 @@
...
-    document.body.appendChild(script);
-  });
-}
-
-download('hello_thread_with_blob_url.js').then(addToDom);
-
-    </script>
+    <!-- Exactly one of these two scripts will exist -->
+    <script src="loader.js"></script>
+    <script type="module" src="loader.mjs"></script>
```
sbc100, line 152: Actually I wonder if that loader can always be an `mjs` file ..even when its loading a plain `js` file?   (i.e. can we remove that first of these two lines?

sbc100:
> Will do! It will only be for URLs though, not for blobs, since Node doesn’t seem to support blobs.

Yes, the existing test in other just does `Module = { mainScriptUrlOrBlob: "./foo.js" }`.   I don't think we need to change that.  Just run it in ES6 mode.

test/pthread/main_js_with_loader.html
JoeOsborn: Good idea!

test/test_browser.py
```diff
@@ -4756,18 +4756,61 @@ def test_pthread_growth(self, emcc_args):
...
+      create_file('loader.mjs', '''
+        Module['locateFile'] = (path,_prefix) => path;
+        fetch('hello_thread_with_loader.mjs').then((rsp) => rsp.blob().then((blob) => {
+          Module['mainScriptUrlOrBlob'] = blob;
+          import(URL.createObjectURL(blob)).then((fac) => fac.default(Module))
+        }));
+      ''')
+    elif use_blob:
+      create_file('loader.js', '''
+        Module['locateFile'] = (path,_prefix) => path;
```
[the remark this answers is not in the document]
JoeOsborn: Only for blobs under es6, I'll remove it here

test/test_browser.py
JoeOsborn: Sure

test/test_browser.py
JoeOsborn: Nice.

test/test_other.py
```diff
@@ -15805,15 +15805,31 @@ def test_rollup(self):
...
-    self.run_process([EMCC, test_file('hello_world.c'), '-sEXIT_RUNTIME', '-sPROXY_TO_PTHREAD', '-pthread', '--pre-js=pre.js'])
+    if es6:
+      self.emcc_args += ['-sEXPORT_ES6']
+      create_file('pre.js', '')
+      create_file('run.mjs', '(await import("./foo.mjs")).default({mainScriptUrlOrBlob:"./foo.mjs"});')
+      binfile = 'a.out.mjs'
+      real_binfile = 'foo.mjs'
+      runfile = 'run.mjs'
+    else:
+      create_file('pre.js', 'Module = { mainScriptUrlOrBlob: "./foo.js" }')
```
sbc100, line 15822: I think the `pre.js` file can be the same in both cases.

test/test_other.py
```diff
@@ -15805,15 +15805,31 @@ def test_rollup(self):
...
+    'es6': (True,),
+  })
+  def test_mainScriptUrlOrBlob(self, es6):
     # Use `foo.js` instead of the current script name when creating new threads
-    create_file('pre.js', 'Module = { mainScriptUrlOrBlob: "./foo.js" }')
-    self.run_process([EMCC, test_file('hello_world.c'), '-sEXIT_RUNTIME', '-sPROXY_TO_PTHREAD', '-pthread', '--pre-js=pre.js'])
+    if es6:
+      self.emcc_args += ['-sEXPORT_ES6']
+      create_file('pre.js', '')
+      create_file('run.mjs', '(await import("./foo.mjs")).default({mainScriptUrlOrBlob:"./foo.mjs"});')
```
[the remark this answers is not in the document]
sbc100: Then you don't need `runfile` at all.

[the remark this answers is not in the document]
JoeOsborn: Cool, I didn't know about that other extern-post-js and it let me simplify the test a lot

test/test_other.py
```diff
@@ -15805,10 +15805,18 @@ def test_rollup(self):
...
+  })
+  def test_mainScriptUrlOrBlob(self, es6):
     # Use `foo.js` instead of the current script name when creating new threads
+    if es6:
+      self.emcc_args += ['-sEXPORT_ES6', '--extern-post-js', test_file('modularize_post_js.js')]
+
     create_file('pre.js', 'Module = { mainScriptUrlOrBlob: "./foo.js" }')
-    self.run_process([EMCC, test_file('hello_world.c'), '-sEXIT_RUNTIME', '-sPROXY_TO_PTHREAD', '-pthread', '--pre-js=pre.js'])
+
+    self.run_process([EMCC, test_file('hello_world.c'), '-sEXIT_RUNTIME', '-sPROXY_TO_PTHREAD', '-pthread', '--pre-js=pre.js', '-o', 'a.out.js'])
```
sbc100, line 15819: I think you can remove the `'-o', 'a.out.js'` but you maybe need to add `self.get_emcc_args()` the array ( sorry I forgot this test wasn't using a helper that did this already.

sbc100: Oh wait maybe you do still the explicit output name (`'-o', 'a.out.js'`) if `EXPORT_ES6` is added.

JoeOsborn: I decided to keep it explicit just to be clear what file is being generated, since it’s important to the test that it has a particular name.

sbc100: I think you still need to add `self.get_emcc_args()` here in order to ensure that `-sEXPORT_ES6` is added.

Alternatively you should change the parameter `args` and do something like this:

```
@parameterized({
   '': ([],),
   'es6': ([-sEXPORT_ES6', '--extern-post-js', test_file('modularize_post_js.js')],),
})
def test_mainScriptUrlOrBlob(self, args):
    ...
     self.run_process([EMCC, test_file('hello_world.c'), '-sEXIT_RUNTIME', '-sPROXY_TO_PTHREAD', '-pthread', '--pre-js=pre.js', '-o', 'a.out.js'] + args)
```

JoeOsborn: I like that better, thanks!

JoeOsborn:
Rebaselined and resubmitted

test/other/codesize/test_codesize_cxx_ctors1.size
```diff
@@ -1 +1 @@
-129166
+129206
```
sbc100, line 1: I would not expect to see any code size changes with this change?   At least not for tests that don't use ES6

JoeOsborn: I'll rebase and try again.

sbc100 (approved):

sbc100 (commented):
Oops, looks like you accidentally included an update to test/third_party/googletest.  A assume that was not deliberate?

JoeOsborn:
Ah sorry, yeah that was something that bootstrap or some other script updated as far as I can tell. Do you need me to revert that and push another commit or will you handle it during merge?

sbc100:
Yup, can you revert and push again.

PR #23893, opened 2025-03-11
[CI] Upgrade to latest macOS image

sbc100:
This xcode version is that latest and corresponds to macOS 14.6.1.


dschuff (approved):

sbc100:
Looks like this change alone fixed the slow brew install phase!

PR #23898, opened 2025-03-11
[WasmFS] Fix asan.test_fs_js_api_wasmfs

sbc100:


src/lib/libwasmfs.js
```diff
@@ -195,13 +195,14 @@ addToLibrary({
...
         bytesRead = __wasmfs_read(stream.fd, dataBuffer, length);
       }
-      bytesRead = FS.handleError(bytesRead);
-
-      for (var i = 0; i < length; i++) {
-        buffer[offset + i] = {{{ makeGetValue('dataBuffer', 'i', 'i8')}}}
+      if (bytesRead > 0) {
+        for (var i = 0; i < bytesRead; i++) {
+          buffer[offset + i] = {{{ makeGetValue('dataBuffer', 'i', 'i8')}}}
+        }
```
kripken, line 201:
```suggestion
      for (var i = 0; i < bytesRead; i++) {
        buffer[offset + i] = {{{ makeGetValue('dataBuffer', 'i', 'i8')}}}
```
The `if` isn't needed, I think? The fix is the condition in the loop, With the new condition, it does the check in the `if` anyhow.

kripken (approved):

PR #23901, opened 2025-03-11
Cleanup test_fs_js_api. NFC

sbc100:
- Remove `cleanup` function (tests don't need to cleanup after themselves like this).
- Remove trailing whitespace
- Fix indentation

### DIFF
--- a/test/fs/test_fs_js_api.c
+++ b/test/fs/test_fs_js_api.c
@@ -14,453 +14,444 @@
 #include <fcntl.h>
 
 EM_JS(void, test_fs_open, (), {
-    FS.writeFile('testfile', 'a=1\nb=2\n');
-    var readStream = FS.open('testfile', 'r');
-    var writeStream = FS.open('testfile', 'w');
-    var writePlusStream = FS.open('testfile', 'w+');
-    var appendStream = FS.open('testfile', 'a');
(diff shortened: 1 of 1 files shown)


sbc100:
View with "hide whitespace"

test/fs/test_fs_js_api.c
```diff
@@ -14,453 +14,444 @@
...
-    test_fs_mmap();
+  // TODO: Fix legacy API FS.mmap bug involving emscripten_builtin_memalign
+  test_fs_mmap();
 #endif
-    test_fs_mkdirTree();
-    test_fs_utime();
-
-    cleanup();
+  test_fs_mkdirTree();
+  test_fs_utime();
```
kripken, line 464: Hmm, I wonder if we need the cleanup for things like IDBFS that persist... but perhaps we don't test this in that mode, which is fine. If we did we'd need to clear it manually.

kripken (approved):

test/fs/test_fs_js_api.c
sbc100: There could be some cases like this, but this is not one of them (IIUC).

In fact its preferable to leave the files around in the test directory afterwards so manual inspection can happen if needed.

PR #23902, opened 2025-03-12
Fix FS.mkdirTree with relative paths

sbc100:
Split out from #23897

### DIFF
--- a/src/lib/libfs.js
+++ b/src/lib/libfs.js
@@ -748,7 +748,8 @@ FS.staticInit();
       var d = '';
       for (var dir of dirs) {
         if (!dir) continue;
-        d += '/' + dir;
+        if (d || PATH.isAbs(path)) d += '/';
+        d += dir;
         try {
           FS.mkdir(d, mode);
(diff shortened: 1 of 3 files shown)


tlively (approved):

src/lib/libfs.js
```diff
@@ -745,10 +745,11 @@ FS.staticInit();
     // Creates a whole directory tree chain if it doesn't yet exist
     mkdirTree(path, mode) {
       var dirs = path.split('/');
-      var d = '';
+      var d = path.startsWith('/') ? '/' : '';
       for (var dir of dirs) {
         if (!dir) continue;
-        d += '/' + dir;
+        if (d) d += '/';
+        d += dir;
```
kripken, line 752: If `d` starts with `/`, say `/foo/bar`, then before the loop we end up with an initial value of `d = '/'`. Then in this first iteration we add `/' (since `d` is truthy) and end up with `//', after which we add `dir`, so we end up with `//foo` in the first iteration, which does not seem right?

PR #23903, opened 2025-03-12
fix 'set EM_PY' in batch files

fs-eire:
This fixes the following error:

```
  '"python"' is not recognized as an internal or external command,
```

This is not usually triggered because usually `EMSDK_PYTHON` is set.

### DIFF
--- a/bootstrap.bat
+++ b/bootstrap.bat
@@ -15,7 +15,7 @@
 @set _PYTHON_SYSCONFIGDATA_NAME=
 @set EM_PY=%EMSDK_PYTHON%
 @if "%EM_PY%"=="" (
-  set EM_PY=python
+  set EM_PY=python.exe
 )
 
 :: Work around Windows bug https://github.com/microsoft/terminal/issues/15212 : If this
--- a/em++.bat
+++ b/em++.bat
@@ -15,7 +15,7 @@
 @set _PYTHON_SYSCONFIGDATA_NAME=
 @set EM_PY=%EMSDK_PYTHON%
 @if "%EM_PY%"=="" (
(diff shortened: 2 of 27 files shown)


sbc100 (commented):
These files are all auto-generated by `create_entry_points.py`, so if you want to make changes that is the place to start.

Does cmd.exe not look for `python.exe` automatically when you type just `python`?

Isn't it better to use the bare `python` string since then the cmd.exe shell will also look for things like `python.bat` and `python.com`?

fs-eire:
> Does cmd.exe not look for `python.exe` automatically when you type just `python`?
> 
> Isn't it better to use the bare `python` string since then the cmd.exe shell will also look for things like `python.bat` and `python.com`?

I ran into this issue when trying to use vcpkg with emscripten in my project. I don't know much about this parts so not sure about the answer to your questions. The direct error I saw is emcc.bat failed because of the error.

sbc100:
> > Does cmd.exe not look for `python.exe` automatically when you type just `python`?
> > Isn't it better to use the bare `python` string since then the cmd.exe shell will also look for things like `python.bat` and `python.com`?
> 
> I ran into this issue when trying to use vcpkg with emscripten in my project. I don't know much about this parts so not sure about the answer to your questions. The direct error I saw is emcc.bat failed because of the error.
> 
> Let me do some further investigation.

How are you installing/using emscripten?  Are you not using emsdk?

Where is `python.exe` coming from on your system?   i.e what does `where python` say?

fs-eire:
Close this one. The root cause is that environment variables are not inherited when using vcpkg. It’s not something we should fix in emscripten.

PR #23904, opened 2025-03-12
Use static initializer for main pthread struct

sbc100:
Turns out this is a slight regression in code size because we end up with more data segments that cannot got in .bss.  I guess its not worth landing?


kripken:
Hmm, yeah, I guess it moves stuff to data, which then is no longer zeroes...

I tend to agree it isn't worth landing, unless there is a benefit I am missing.

sbc100:
No, I just wanted share these slightly surprising results.

PR #23910, opened 2025-03-12
[WasmFS] Use stack allocation for in FS.stat/lstat. NFC

sbc100:


kripken (approved):
Hmm, I guess it's small enough. Looks like it's 96 bytes, big for a struct, but ok even for a small stack.

sbc100:
> Hmm, I guess it's small enough. Looks like it's 96 bytes, big for a struct, but ok even for a small stack.

Yes 96 is (a) a fixed size and (b) likely often smaller than filename which we always put on the stack already.

PR #23916, opened 2025-03-13
Fix iteration of OpenAL src.bufQueue

JoeOsborn:
This failing test shows up under the interactive tests only, so I guess CI didn't catch it.  Caused by #23867 .

### DIFF
--- a/test/openal/test_openal_buffers.c
+++ b/test/openal/test_openal_buffers.c
@@ -89,7 +89,9 @@ void iter() {
   alGetSourcef(source, AL_PITCH, &testPitch);
   assert(pitch == testPitch);
 #endif
-
+#ifdef TEST_FAST
+  offset = size;
+#endif
   // Exit once we've processed the entire clip.
   if (offset >= size) {
 #if __EMSCRIPTEN__
@@ -192,7 +194,6 @@ int main(int argc, char* argv[]) {
   ALint srcLen = 0;
   alGetSourcei(source, 0x2009 /* AL_BYTE_LENGTH_SOFT */, &srcLen);
   assert(srcLen == NUM_BUFFERS * BUFFER_SIZE);
-
 #ifdef TEST_ANIMATED_PITCH
   printf("You should hear a clip of the 1902 piano song \"The Entertainer\" played back at a high pitch rate, and animated to slow down to half playback speed.\n");
 #else
--- a/test/openal/test_openal_buffers.c
+++ b/test/openal/test_openal_buffers.c
@@ -89,7 +89,7 @@ void iter() {
   alGetSourcef(source, AL_PITCH, &testPitch);
   assert(pitch == testPitch);
 #endif
-#ifdef TEST_FAST
+#ifndef TEST_INTERACTIVE
   offset = size;
 #endif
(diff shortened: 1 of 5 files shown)


JoeOsborn:
I skimmed that PR and it looks like this is the only case where .length wasn't deleted from the iterated sequence during the conversion to for..of.

sbc100 (approved):
Thanks!

I wonder if we could make one of the non-interactive tests cover this?

JoeOsborn:
I wonder---doesn't AudioContext stuff require a click?  I don't know how it acts or what it does in the absence of interaction.

sbc100:
> I wonder---doesn't AudioContext stuff require a click? I don't know how it acts or what it does in the absence of interaction.

The CI tests we recently updated to avoid that requirement.  See #23665 and #23701

JoeOsborn:
I've copied one or two tests and given a "fast" version of the test_openal_buffers to avoid having to cycle through the whole entertainer.wav in automated tests.

sbc100 (approved):
Nice!

test/openal/test_openal_buffers.c
```diff
@@ -89,7 +89,9 @@ void iter() {
   alGetSourcef(source, AL_PITCH, &testPitch);
   assert(pitch == testPitch);
 #endif
-
+#ifdef TEST_FAST
```
sbc100, line 92: I wonder if this should be called `TEST_INTERACTIVE` and invert that check?

JoeOsborn: Done.

test/test_browser.py
```diff
@@ -2323,6 +2323,17 @@ def test_openal_capture_sanity(self):
   def test_openal_extensions(self):
     self.btest_exit('openal/test_openal_extensions.c')
 
+  @parameterized({
+    '': ([],),
+    'proxy_to_pthread': (['-sPROXY_TO_PTHREAD', '-pthread'],),
+  })
+  def test_openal_playback(self, args):
+    shutil.copy(test_file('sounds/audio.wav'), '.')
+    self.btest('openal/test_openal_playback.c', '1', emcc_args=['-O2', '--preload-file', 'audio.wav'] + args)
```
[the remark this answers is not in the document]
JoeOsborn: I'm sure we could!  I just copied this from `test_other.py`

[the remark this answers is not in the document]
JoeOsborn: Actually, this test doesn't seem to work in browser under pthreads (the sound clip finishes but the test never terminates), but that's not really related to this PR.  Any ideas on why that might be happening?  Does the thread need to do some sleeps or something?

In the meantime I have removed the pthread parameterization of this test on test_browser.py.

sbc100 (approved):

JoeOsborn:
The rebaseline on this branch was a mistake, do you want me to undo it?

test/other/codesize/test_codesize_cxx_mangle.size
```diff
@@ -1 +1 @@
-232600
+232604
```
sbc100, line 1: This rebase just landed separately: https://github.com/emscripten-core/emscripten/pull/23918

Issue #23915, opened 2025-03-13
Docs: Mouse events in device orientation and device motion callbacks?


davidhiberbom:
https://github.com/emscripten-core/emscripten/blob/002962761ca7e370aed80680b61e5237b53510c0/site/source/docs/api_reference/html5.h.rst?plain=1#L799-L803

The docs suggest you need to use to pass mouse events into the device orientation callback. This is a typo, right?

Same thing with device motion:

https://github.com/emscripten-core/emscripten/blob/002962761ca7e370aed80680b61e5237b53510c0/site/source/docs/api_reference/html5.h.rst?plain=1#L891-L895

sbc100:
Looks like a typo yes.  Would you have time to send a PR?

PR #23919, opened 2025-03-13
Fix typo about devicemotion and deviceorientation events in html5.h.rst

davidhiberbom:
As suggested in #23915, the docs seem to say you need to pass mouse events into the devicemotion and deviceorientation callbacks, which looks like a typo.

### DIFF
--- a/site/source/docs/api_reference/html5.h.rst
+++ b/site/source/docs/api_reference/html5.h.rst
@@ -800,7 +800,7 @@ Functions
 
   Returns the most recently received ``deviceorientation`` event state.
 
-  Note that for this function call to succeed, :c:func:`emscripten_set_deviceorientation_callback` must have first been called with an :c:data:`EMSCRIPTEN_EVENT_DEVICEORIENTATION` event and a non-zero callback function pointer to enable the ``deviceorientation`` state capture.
+  Note that for this function call to succeed, :c:func:`emscripten_set_deviceorientation_callback` must have first been called with a non-zero callback function pointer to enable the ``deviceorientation`` state capture.
 
   :param orientationState: The most recently received ``deviceorientation`` event state.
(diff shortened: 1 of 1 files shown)


site/source/docs/api_reference/html5.h.rst
```diff
@@ -800,7 +800,7 @@ Functions
 
   Returns the most recently received ``deviceorientation`` event state.
 
-  Note that for this function call to succeed, :c:func:`emscripten_set_deviceorientation_callback` must have first been called with one of the mouse event types and a non-zero callback function pointer to enable the ``deviceorientation`` state capture.
+  Note that for this function call to succeed, :c:func:`emscripten_set_deviceorientation_callback` must have first been called with an :c:data:`EMSCRIPTEN_EVENT_DEVICEORIENTATION` event and a non-zero callback function pointer to enable the ``deviceorientation`` state capture.
```
davidhiberbom, line 803: Or perhaps you would prefer `deviceorientation` over 

```
:c:data:`EMSCRIPTEN_EVENT_DEVICEORIENTATION`
```

?

davidhiberbom: Actually, this still looks wrong. `emscripten_set_deviceorientation_callback` isn't the function that needs to have been called with an `EMSCRIPTEN_EVENT_DEVICEORIENTATION` event, but the callback passed to `emscripten_set_deviceorientation_callback`.

sbc100: I think maybe just skip that part of the sentence.

It should be enough to say `Note that for this function call to succeed, emscripten_set_deviceorientation_callback must have first been called with a non-zero callback function pointer....`

sbc100 (approved):

PR #23930, opened 2025-03-14
Fix flake8 errors/warnings. NFC

sbc100:
Turns out ruff still doesn't support several flake8 warnings.  I've enabled some more that are in preview-only mode.

See #23085


tools/maint/npm_update.py
```diff
@@ -23,7 +23,7 @@ def run(cmd, **args):
 def main():
   if run(['git', 'status', '-uno', '--porcelain']).strip():
     print('tree is not clean')
-    #return 1
+    return 1
```
sbc100, line 26: This was actually a bug that accidentally submitted this.

kripken (approved):

PR #23935, opened 2025-03-15
Reimplement `resolve_path` in JS NFC

ryanking13:
As discussed in #23872, this PR reimplements `resolve_path` function, which is used to locate the library using `LD_LIBRARY_PATH` in JavaScript, so that it can be reused inside `libdylink.js`.

Currently, this code doesn't work in thread.

### DIFF
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -1238,7 +1238,14 @@ var LibraryDylink = {
   _dylink_resolve_path_js__deps: ['$UTF8ToString', '$stringToUTF8', '$locateLibraryFromFS', '$getDefaultLibDirs'],
   _dylink_resolve_path_js__proxy: 'sync',
   _dylink_resolve_path_js: (cbuf, cfile, buflen) => {
-    var file = UTF8ToString(cfile);
+    var cfilePtr = cfile;
+
+#if MEMORY64
+    cfilePtr = Number(cfilePtr);
+#endif
+
+    var file = UTF8ToString(cfilePtr);
+
     if (file.startsWith("/")) {
       return cfile;
     }
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -1215,6 +1215,9 @@ var LibraryDylink = {
 #endif
         return res.path;
       } catch(e) {
+#if DYLINK_DEBUG
+        dbg(`locateLibraryFromFS: ${path} not found: ${e}`);
+#endif
         // do nothing is file is not found
       }
     }
@@ -1238,10 +1241,6 @@ var LibraryDylink = {
     if (file.startsWith("/")) {
       return cfile;
     }
-
-// #if DYLINK_DEBUG
-//     dbg("_dylink_resolve_path_js: entries in sharedModules: " + Object.keys(sharedModules));
-// #endif
 
     var res = locateLibraryFromFS(file, getDefaultLibDirs(), buflen - 1);
     if (!res) {
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -1239,9 +1239,9 @@ var LibraryDylink = {
       return cfile;
     }
 
-#if DYLINK_DEBUG
-    dbg("_dylink_resolve_path_js: entries in sharedModules: " + Object.keys(sharedModules));
-#endif
+// #if DYLINK_DEBUG
+//     dbg("_dylink_resolve_path_js: entries in sharedModules: " + Object.keys(sharedModules));
+// #endif
 
     var res = locateLibraryFromFS(file, getDefaultLibDirs(), buflen - 1);
     if (!res) {
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -656,7 +656,7 @@ var LibraryDylink = {
       }
 #if DYLINK_DEBUG
       dbg(`loadModule: memory[${memoryBase}:${memoryBase + metadata.memorySize}]` +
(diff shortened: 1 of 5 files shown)


ryanking13:
@hoodmane  @sbc100 This is a split off from https://github.com/emscripten-core/emscripten/pull/23872, but I am having trouble making the function work in thread. Can I get some help with this?

hoodmane:
Is there a ci failure for this problem? Or which test should I run? I only see the sigs job failing which needs an update to the expected output.

ryanking13:
This one fails

```
./test/runner other.test_ld_library_path_pthread
```

while

```
./test/runner other.test_ld_library_path
```

success

hoodmane:
Well applying this change, I can see that `stat` finds the file but `lookupPath` does not:
```
diff --git a/test/test_other.py b/test/test_other.py
index ad240f4a1..d33fdf0b4 100644
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -7605,12 +7605,29 @@ Module.preRun = () => {
 #include <stdlib.h>
 #include <string.h>
 #include <dlfcn.h>
+#include "emscripten.h"
+#include <sys/stat.h>
+
+EM_JS(int, check_path, (char* cpath), {
+  const path = UTF8ToString(cpath);
+  console.log("check_path", path);
+  try {
+    const res = FS.lookupPath(path);
+    console.log("res", res);
+  } catch(e) {
+    console.log(e);
+  }
+});
 
 int main() {
   void *h;
   void (*f)();
   double (*f2)(double);
 
+  check_path("/lib/libhello1.wasm");
+  struct stat statbuf;
+  printf("res: %d\n", stat("/lib/libhello1.wasm", &statbuf));
+
   h = dlopen("libhello1.wasm", RTLD_NOW);
   assert(h);
   f = dlsym(h, "hello1");
```

hoodmane:
Okay I think I fixed the problem in [b805542](https://github.com/emscripten-core/emscripten/pull/23935/commits/b80554258417f6637531f88d61f0b63261b62ba5).

ryanking13:
> Okay I think I fixed the problem in [b805542](https://github.com/emscripten-core/emscripten/pull/23935/commits/b80554258417f6637531f88d61f0b63261b62ba5).

Indeed. Thanks for the fix! Didn't know about the proxy parameter. The remaining issue now seems to be handling BigInt in a MEMORY64 environment.

hoodmane:
The memory64 test now passes for me locally.

ryanking13:
>The memory64 test now passes for me locally.

@hoodmane Awesome, thanks for helping me fix the issues.

src/lib/libdylink.js
```diff
@@ -1172,6 +1172,96 @@ var LibraryDylink = {
...
+
+  _dylink_resolve_path_js__deps: ['$UTF8ToString', '$stringToUTF8', '$locateLibraryFromFS', '$getDefaultLibDirs'],
+  _dylink_resolve_path_js__proxy: 'sync',
+  _dylink_resolve_path_js: (cbuf, cfile, buflen) => {
+    var cfilePtr = cfile;
+
+#if MEMORY64
+    cfilePtr = Number(cfilePtr);
+    buflen = Number(buflen)
+#endif
```
hoodmane, line 1246: So in the memory 64 case, we need the return value to be a `BigInt`? I wonder if the best thing to do would be:
```suggestion
#if MEMORY64
    cbuf = Number(cbuf);
    cfile = Number(cfile);
    buflen = Number(buflen)
#endif
```
and then explicitly convert back to `BigInt` while returning.

hoodmane: But surprisingly I can't trigger any more failures with the current implementation. So I guess it's probably fine? I'll add a wasm64 variant to the LD_LIBRARY_PATH test just to be sure.

src/lib/libdylink.js
```diff
@@ -1172,6 +1172,96 @@ var LibraryDylink = {
 #endif
   },
 
+  $locateLibraryFromFS__deps: ['$FS'],
+  $locateLibraryFromFS: (filename, searchDirs, maxLength = Infinity) => {
+    // Find the library in the filesystem.
+    // returns null if not found.
+    if (typeof FS.lookupPath !== 'function') {
```
hoodmane, line 1179: Slightly shorter:
```suggestion
    if (!FS.lookupPath) {
```

src/lib/libdylink.js
```diff
@@ -1172,6 +1172,96 @@ var LibraryDylink = {
...
+      // wasmfs does not implement FS.lookupPath
+#if DYLINK_DEBUG
+      dbg("locateLibraryFromFS: FS.lookupPath not implemented");
+#endif
+      return null;
+    }
+
+    var candidates = [];
+    if (filename.charAt(0) === '/') {  // abs path
```
hoodmane, line 1188: shorter:
```suggestion
    if (filename[0] === '/') {  // abs path
```

hoodmane (approved):
It looks good to me, thanks @ryanking13. Let's wait for @sbc100 to review as well.

hoodmane:
Hmm it seems the test I added breaks in CI. When I run it locally it works fine...

hoodmane:
I hit the merge button on this, but now I'm thinking it might have been premature. Ah well, we could either revert and reland or fix any problems in followups.

sbc100:
Yes, lets revert this pending some more discussion.

sbc100:
Can you re-open this PR? (or perhaps a new one is needed?)

ryanking13:
@hoodmane  @sbc100  Reopened it #23958

hoodmane:
Sorry about that!

Issue #23938, opened 2025-03-16
Size of base64 encoded WASM more than doubled going from 3.1.74 to 4.0.0


jrouwe:
I upgraded emcc from 3.1.74 to 4.0.0 and noticed that [my library](https://github.com/jrouwe/JoltPhysics.js) went from [3.8 MB to 9.0 MB](https://bundlephobia.com/package/jolt-physics@0.32.0) without any code changes. This only happens with the version that embeds WASM base64 encoded in the JS, not when WASM is stored in a separate file.

I've created a minimal reproduction case here:

[MRP.zip](https://github.com/user-attachments/files/19274747/MRP.zip)

This compiles a nearly empty file generated by `webidl_binder.py` using:

```
emcc glue.cpp -include JoltJS.h --closure=1 -s EXPORT_ES6=1 -s EXPORT_NAME=Jolt -s MODULARIZE=1 -O3 -s WASM=1 -s SINGLE_FILE=1 -o out.js
```

When running this on emcc 3.1.74 out.js is 14 KB, when running on 4.0.0 (or 4.0.5) the file is 34 KB. Both with this MRP and in my library the result is roughly 2.4x bigger so I hope that my MRP shows the same issue.

The thing that seems to trigger this is the `--closure=1` parameter. If I remove it, both versions become roughly 20 KB. It appears that closure does something with the base64 encoded WASM: it becomes much longer when the `--closure=1` argument is added. Without that parameter the base64 encoded WASM is comparable. This is the output for both runs, with and without closure:

[out_3_1_74.txt](https://github.com/user-attachments/files/19274824/out_3_1_74.txt)
[out_3_1_74_noclosure.txt](https://github.com/user-attachments/files/19274822/out_3_1_74_noclosure.txt)
[out_4_0_0.txt](https://github.com/user-attachments/files/19274823/out_4_0_0.txt)
[out_4_0_0_noclosure.txt](https://github.com/user-attachments/files/19274825/out_4_0_0_noclosure.txt)

If you diff the files you can see the difference in the base64 encoded block.

**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.5 (53b38d0c6f9fce1b62c55a8012bc6477f7a42711)
```

**Full link command and output with `-v` appended:**
```
[...]

PR #23967, opened 2025-03-21
Avoid repeating wasm data in SINGLE_FILE mode with closure. NFC

sbc100:
This change removes the `wasmBinaryFile` global in SINGLE_FILE mode. Instead the `wasmBinaryFile` is inlined into the `createWasm` function.

As part of this change we now also use the same embedding method as is used in `MINIMAL_RUNTIME`, avoiding the need for the `tryParseAsDataURI` helper.

Fixes: #23938


sbc100 (commented):
See the huge reduction in `test/other/codesize/test_codesize_hello_single_file.jssize`

src/preamble.js
```diff
@@ -603,14 +599,14 @@ function findWasmBinary() {
 #endif
 
 function getBinarySync(file) {
+#if SINGLE_FILE
+  if (ArrayBuffer.isView(file)) {
+    return file;
+  }
```
kripken, line 605: Isn't it always a view here, so can't we remove the `if`?

kripken (approved):
Nice!

src/preamble.js
sbc100: This code I believe can be used to load other wasm files, not just the main wasm file.   Unless I'm misremembering.

kripken: Oh, maybe in dynamic linking? sgtm.

sbc100: Actually, no it looks like only `emscripten_lazy_load_code` and `load_secondary_module` use these loading functions for other wasm file..   anyway, maybe that can be a diffident cleanup.

kripken: Sounds good.

PR #23970, opened 2025-03-21
Deprecate -sMAYBE_WASM2JS settings

sbc100:
Users of this settings will now see:

```
emcc: warning: MAYBE_WASM2JS is deprecated (no known users). Please open a bug if you have a continuing need for this setting [-Wdeprecated]
```

See #23969

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -8388,7 +8388,7 @@ Module.onRuntimeInitialized = () => {
       self.skipTest('redundant to test wasm2js in wasm2js* mode')
     self.set_setting('MAYBE_WASM2JS')
     # see that running as wasm works
-    self.do_core_test('test_hello_world.c')
+    self.do_core_test('test_hello_world.c', emcc_args=['-Wno-deprecated'])
     # run wasm2js, bundle the code, and use the wasm2js path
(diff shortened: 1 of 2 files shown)


kripken:
sgtm, though @juj was using this at some point - @juj do you still need it?

kripken:
Ah, I may have confused this with `WASM=2`.

Looks like wasm2js always enables MAYBE_WASM2JS. This may have been useful as a user-facing feature in the past for debugging purposes, but maybe now we can remove it from there (but keep it internally, or refactor).

sbc100:
> Ah, I may have confused this with `WASM=2`.
> 
> Looks like wasm2js always enables MAYBE_WASM2JS. This may have been useful as a user-facing feature in the past for debugging purposes, but maybe now we can remove it from there (but keep it internally, or refactor).

Yes, I was think we could keep `-sWASM=2` which I believe @juj uses, and either remove `-sMAYBE_WASM2JS` or at least make it internal.

kripken (approved):
Making it internal (or refactoring in some other internal way) sgtm. I'm not aware of MAYBE_WASM2JS usage without wasm2js.

juj:
Thanks for pinging: Yeah, we do not use -sMAYBE_WASM2JS, only the -sWASM=0/1/2 settings.

PR #23973, opened 2025-03-24
Make `__file__` work in config file

sbc100:
As of today folks who want to find stuff relative to the config file itself need to do `os.getenv('EM_CONFIG')` which is pretty hacky.


kripken (approved):

kripken:
This test seems to fail on windows:

https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket/8719400409790050369/+/u/Emscripten_testsuite__other_/stdout

PR #23981, opened 2025-03-25
Fix document for returnType and argTypes of cwrap

wakairo:
[`cwrap` function calls `ccall`](https://github.com/emscripten-core/emscripten/blob/4.0.5/src/lib/libccall.js#L144), which does [not accept "array" as a return type](https://github.com/emscripten-core/emscripten/blob/4.0.5/src/lib/libccall.js#L61).
Consequently, the current documentation for cwrap's returnType and argTypes parameters is inaccurate, and this PR updates it accordingly.

### DIFF
--- a/site/source/docs/api_reference/preamble.js.rst
+++ b/site/source/docs/api_reference/preamble.js.rst
@@ -114,8 +114,8 @@ Calling compiled C functions from JavaScript
         my_func(12)
 
   :param ident: The name of the C function to be called.
-  :param returnType: The return type of the function. This can be ``"number"``, ``"string"`` or ``"array"``, which correspond to the appropriate JavaScript types (use ``"number"`` for any C pointer, and ``"array"`` for JavaScript arrays and typed arrays; note that arrays are 8-bit), or for a void function it can be ``null`` (note: the JavaScript ``null`` value, not a string containing the word "null").
-  :param argTypes: An array of the types of arguments for the function (if there are no arguments, this can be omitted). Types are as in ``returnType``, except that ``array`` is not supported as there is no way for us to know the length of the array).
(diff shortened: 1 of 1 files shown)


site/source/docs/api_reference/preamble.js.rst
```diff
@@ -114,8 +114,8 @@ Calling compiled C functions from JavaScript
         my_func(12)
 
   :param ident: The name of the C function to be called.
-  :param returnType: The return type of the function. This can be ``"number"``, ``"string"`` or ``"array"``, which correspond to the appropriate JavaScript types (use ``"number"`` for any C pointer, and ``"array"`` for JavaScript arrays and typed arrays; note that arrays are 8-bit), or for a void function it can be ``null`` (note: the JavaScript ``null`` value, not a string containing the word "null").
-  :param argTypes: An array of the types of arguments for the function (if there are no arguments, this can be omitted). Types are as in ``returnType``, except that ``array`` is not supported as there is no way for us to know the length of the array).
+  :param returnType: The return type of the function. This can be ``"number"`` or ``"string"``, which correspond to the appropriate JavaScript types (use ``"number"`` for any C pointer), or for a void function it can be ``null`` (note: the JavaScript ``null`` value, not a string containing the word "null"). ``array`` is not supported as there is no way for us to know the length of the array.
```
sbc100, line 117: Looking at the code it looks like we support "string", "pointer" and "boolean".

I'm not sure we need to spell out where why `array` is not a valid return type, or rather I'm not sure exactly what the rational is so maybe better to leave it out.

wakairo: The documentation for `array` was simply carried over from what originally existed, so I don't think it's essential, and it can be removed.

That aside, the following description for ":param argTypes:" has disappeared; is this acceptable?

```
  :param argTypes: An array of the types of arguments for the function (if there are no arguments, this can be omitted). Types can be ``"number"``, ``"string"`` or ``"array"``, which correspond to the appropriate JavaScript types (use ``"number"`` for any C pointer, and ``"array"`` for JavaScript arrays and typed arrays; note that arrays are 8-bit)
```

sbc100: I think we should still have some `argTypes` documentation.   I did't realize the rational was part of the original docs here.

PR #23985, opened 2025-03-25
Initial experimental support for ESM integration

sbc100:
This change add a new experimental settings call `WASM_ESM_INTEGRATION`.

In this mode we can elide all of the wasm module loading and instantiation code, and instead rely on the module loader to both load and instantiate the wasm file.

Current limitations:

- No support for INCOMING_JS_MODULE_API / moduleArg.
- No support for closure
- Many features such as threads and dynamic linking do not work

### DIFF
--- a/src/preamble.js
+++ b/src/preamble.js
@@ -46,7 +46,9 @@ if (typeof WebAssembly != 'object') {
 
 // Wasm globals
 
+#if !WASM_ESM_INTEGRATION
 var wasmMemory;
+#endif
 
 #if SHARED_MEMORY
 // For sending to workers.
@@ -220,7 +222,11 @@ function initRuntime() {
   <<< ATINITS >>>
 
 #if hasExportedSymbol('__wasm_call_ctors')
+#if WASM_ESM_INTEGRATION
+  ___wasm_call_ctors();
+#else
   wasmExports['__wasm_call_ctors']();
+#endif
 #endif
 
   <<< ATPOSTCTORS >>>
@@ -577,7 +583,7 @@ function resetPrototype(constructor, attrs) {
 }
 #endif
 
-#if !SOURCE_PHASE_IMPORTS
+#if !SOURCE_PHASE_IMPORTS && !WASM_ESM_INTEGRATION
 var wasmBinaryFile;
 
 function findWasmBinary() {
@@ -819,6 +825,7 @@ async function instantiateAsync(binary, binaryFile, imports) {
 #endif // WASM_ASYNC_COMPILATION
 #endif // SOURCE_PHASE_IMPORTS
 
+#if !WASM_ESM_INTEGRATION
 function getWasmImports() {
 #if PTHREADS
   assignWasmImports();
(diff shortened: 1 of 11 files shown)


src/preamble.js
```diff
@@ -577,7 +583,7 @@ function resetPrototype(constructor, attrs) {
 }
 #endif
 
-#if !SOURCE_PHASE_IMPORTS
+#if !SOURCE_PHASE_IMPORTS && !WASM_ESM_INTEGRATION
```
kripken, line 586: Can these two settings not be merged into one? Are there use cases for one without the other?

tools/link.py
```diff
@@ -2111,6 +2118,27 @@ def create_worker_file(input_file, target_dir, output_file, options):
...
+  wrapper = []
+  if wasm_exports:
+    wrapper.append(f"export {{ {wasm_exports} }} from './{settings.WASM_BINARY_FILE}';")
+  support_url = f'./{os.path.basename(support_target)}'
+  wrapper.append(f"import init from '{support_url}';")
+  wrapper.append('export default init;')
+  write_file(wrapper_file, '\n'.join(wrapper) + '\n')
+
+  # FIXME(sbc): This is a huge hack to rename the imports in the
+  # wasm file.  Find a better way to do this.
```
kripken, line 2132: We can add a binaryen pass for this in time, I suppose.

kripken (approved):

src/preamble.js
sbc100: They are fairly different.  I think it make sense to keep them separate.   

I'm not big of something like `ESM=1` and `ESM=2` where the numbers don't really describe which is happening.

tools/link.py
sbc100: Yes, I guess binaryen would be the place to do this.. although is it possible to do this without parsing and validating and re-writing the whole file?

kripken: Ah yes, it would parse... I guess a binary-level rewrite could be more efficient, for compiles that never use binaryen at all.

src/preamble.js
kripken: Why would I was source phase imports but not wasm ESM, though?

brendandahl: What about trying to move all the settings that control output format into one? maybe `OUTPUT_FORMAT`?
then have:
 - esm
 - es
 - source-phase
 - cjs
 - none
 - ....

sbc100: > What about trying to move all the settings that control output format into one? maybe `OUTPUT_FORMAT`? then have:
> 
> * esm
> * es

Are you suggesting `es` for the existing JS module output?

The problem is that they are both ESM, right?  So we could need `esm` and `esm-with-wasm-integration` both I think.

sbc100: > Why would I want source phase imports but not wasm ESM, though?

Source phase imports is basically a different type of ESM integration.   Its more flexible and likely to be supported in browsers soon.  Full ESM integration is different, and further from browser adoption, but its available in node today with certain flags. 

The both allow is to remove loading code but full ESM integration allows to remove even more.

For now at least, I think they are separately useful.

sbc100:
OK, I think I will land this initial experiment and we can build on it over time.

PR #23991, opened 2025-03-26
shell.js: Remove moduleOverrides. NFC

sbc100:
This mechanism was added back in 80f517d78 so that shell code could unconditionally assign to the `Module` object without clobbering the incoming values there.

These days the shell code no longer assigns to the Module object so it is no longer necessary.


kripken (approved):
Nice! Yeah, I think you got rid of all the writes to Module in that code... very cool.

PR #23995, opened 2025-03-26
Move some Module object assertions. NFC

sbc100:
This moves these assertion alongside other similar checks.


brendandahl (approved):
r+ but update commit message since this moves two things.

kripken (approved):

kripken:
It looks like `other.test_emcc_4` failed on CI here, and is now failing on the rollers,

https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket/8719235225433347489/+/u/Emscripten_testsuite__other_/stdout

Maybe it looked like a flake but it wasn't?

PR #23998, opened 2025-03-27
Move `sharedModules` object to `runtime_pthread.js`

kleisauke:


kleisauke:
For some reason this fixes a regression after commit a14f47165630eb1f2ae6a9b3c60b141e4b5541fc. After this commit, my pet project started failing with:
```console
TypeError: Cannot read properties of undefined (reading 'vips-jxl.wasm')
    at e (file:///home/kleisauke/wasm-vips/lib/vips-node.mjs:44:61)
    at f (file:///home/kleisauke/wasm-vips/lib/vips-node.mjs:44:402)
    at zf (file:///home/kleisauke/wasm-vips/lib/vips-node.mjs:45:263)
    at file:///home/kleisauke/wasm-vips/lib/vips-node.mjs:46:305
```

At this line:
https://github.com/emscripten-core/emscripten/blob/a14f47165630eb1f2ae6a9b3c60b141e4b5541fc/src/lib/libdylink.js#L1008

This could probably(?) also be resolved by moving this object to `preamble.js`, but this approach felt neater.

(added this as a second comment since the PR description is nowadays used in the landing commit)

test/test_other.py
```diff
@@ -15265,26 +15265,34 @@ def test_preload_module(self, args):
...
+        }
+      #ifdef __EMSCRIPTEN_PTHREADS__
+        else {
           int found = EM_ASM_INT(
             err(sharedModules);
             return sharedModules['/library.so'] !== undefined;
           );
           assert(found);
         }
+      #endif
```
sbc100, line 15294: Was this change really needed?  I think if this code worked prior to this change it should work after it no?

i.e. the `sharedModules` symbol is available in the same builds as before, right?

sbc100:
So it makes sense that https://github.com/emscripten-core/emscripten/pull/23704 could have caused this kind of issue, because it makes means that postable code run only after the createWasm is resolved. 

However, I'd love to know why none of our tests caught this.   Maybe we don't have enough coverage of dynamic linking + modularize + pthreads?

sbc100:
I think maybe a better place for this would be in `src/runtime_pthread.js` at the top (around line 15).

test/test_other.py
kleisauke: Seems to be unnecessary, reverted with commit da0b4df48a9bc1f6d55909416fdc71e90591ff44.

kleisauke:
> Maybe we don't have enough coverage of dynamic linking + modularize + pthreads?

I think so. It might also be a Closure-specific issue since I'm using that as well.

> I think maybe a better place for this would be in `src/runtime_pthread.js` at the top (around line 15).

Done with ba18f68215e0a7b202c17871a6aad8d39424b16a. I confirm that works too.

sbc100 (approved):
LGTM although I'd love to also get a test for this.

Maybe `test_preload_module` could have `also_with_modularize` added?

sbc100:
> LGTM although I'd love to also get a test for this.
> 
> Maybe `test_preload_module` could have `also_with_modularize` added?

Actually maybe not that test.. but any test that just used load-time dynamic linking and pthreads.

kleisauke:
Added a test with commit 3854ac7192d6a9158c7851faf75a438c53ceb545. The `_modularize` variant of this test would previously fail with:
```console
/tmp/emtest_pg577vro/emscripten_test_other_kb_3mt0y/main.js:2546
        var sharedMod = sharedModules[libName];
                                     ^

TypeError: Cannot read properties of undefined (reading 'side.wasm')
    at loadLibData (/tmp/emtest_pg577vro/emscripten_test_other_kb_3mt0y/main.js:2546:38)
    at getExports (/tmp/emtest_pg577vro/emscripten_test_other_kb_3mt0y/main.js:2583:18)
    at loadDynamicLibrary (/tmp/emtest_pg577vro/emscripten_test_other_kb_3mt0y/main.js:2600:16)
    at /tmp/emtest_pg577vro/emscripten_test_other_kb_3mt0y/main.js:2641:11
Thrown at:
    at loadLibData (/tmp/emtest_pg577vro/emscripten_test_other_kb_3mt0y/main.js:2546:38)
    at getExports (/tmp/emtest_pg577vro/emscripten_test_other_kb_3mt0y/main.js:2583:18)
    at loadDynamicLibrary (/tmp/emtest_pg577vro/emscripten_test_other_kb_3mt0y/main.js:2600:16)
    at /tmp/emtest_pg577vro/emscripten_test_other_kb_3mt0y/main.js:2641:11
```

test/test_other.py
```diff
@@ -2175,6 +2175,29 @@ def test(link_flags, lib_suffix):
...
+
+      extern int answer();
+
+      int main() {
+        assert(answer() == 42);
+        printf("done\n");
+        return 0;
+      }
+    ''')
+    self.do_runf('main.c', 'done\n', emcc_args=['-sMAIN_MODULE=2', '-pthread', '-Wno-experimental', 'side.wasm'])
```
sbc100, line 2199: Do we need to write a whole new test here?  Is there not an existing dylink pthread test we could just add `@also_with_modularize` to?

sbc100 (commented):
Awesome, thanks for the fix!

LGTM % comment.

test/test_other.py
kleisauke: Ah, we could just merge it with the test below. Done with eea195e12fbe70d7ff5c06af6574e695d5d8f0d3.

sbc100 (approved):

sbc100:
If you merge / rebase with main the test_emcc_4 failure is now fixed.

PR #23999, opened 2025-03-27
Improve err/out implementation for node pthreads. NFC

sbc100:
The console overrides for node were not correctly formatting their arguments. i.e. all objects were showing up as `[object Object]`


test/other/codesize/test_codesize_minimal_pthreads.gzsize
```diff
@@ -1 +1 @@
-4037
+4045
```
kripken, line 1: Code size increase seems fine as this is just for node.js (and pthreads) logic.

kripken (approved):

test/other/codesize/test_codesize_minimal_pthreads.gzsize
sbc100: Indeed.

PR #24005, opened 2025-03-28
[esm-integration] Allow for module argument processing

sbc100:
This change delays module argument processing until the init function is run.  This means that at least some INCOMING_MODULE_JS_API properties can be supported with ESM integration.

 See #24060


src/parseTools.mjs
```diff
@@ -720,6 +720,14 @@ function makeEval(code) {
   return ret;
 }
 
+// Add code that runs before the wasm modules is loaded.  This is the first
+// point at which the global `Module` object is guaranteed to exist. This hooks
```
brendandahl, line 724:
```suggestion
// point at which the global `Module` object is guaranteed to exist. This hook
```

src/preamble.js
```diff
@@ -461,8 +457,8 @@ var FS = {
 
   ErrnoError() { FS.error() },
 };
-Module['FS_createDataFile'] = FS.createDataFile;
-Module['FS_createPreloadedFile'] = FS.createPreloadedFile;
+//Module['FS_createDataFile'] = FS.createDataFile;
```
brendandahl, line 460: Does this need to be moved or removed?

src/preamble.js
```diff
@@ -1086,3 +1082,68 @@ function getCompilerSetting(name) {
...
+  {{{ makeRemovedFSAssert('PROXYFS') }}}
+  {{{ makeRemovedFSAssert('WORKERFS') }}}
+  {{{ makeRemovedFSAssert('FETCHFS') }}}
+  {{{ makeRemovedFSAssert('ICASEFS') }}}
+  {{{ makeRemovedFSAssert('JSFILEFS') }}}
+  {{{ makeRemovedFSAssert('OPFS') }}}
+
+#if !NODERAWFS
+  {{{ makeRemovedFSAssert('NODEFS') }}}
+#endif
```
brendandahl, line 1141: These all turn into `var IDBFS = 'IDBFS is no longer included ...';` and will now be in the processModuleArgs function. If the user tries to use the variable it will undefined which will be different than the old behavior. Do we want to leave these outside of the function?

brendandahl (changes requested):

sbc100:
Updated.   PTAL

src/preamble.js
brendandahl: What about this one?

sbc100:
Done

brendandahl (approved):

sbc100:
I had to refactor again.  I ended up going back to something more like the naming used in your original PR @brendandahl.

Also, I had to introduce another JS file called `postlibrary.js` which gets inject after the JS library code but before the module creation code.

brendandahl (approved):

PR #24012, opened 2025-03-29
Fix order of execution of runtime callbacks

lvlte:
This makes the `addOn<X>` functions use `[].push()` instead of `[].unshift()` for enqueueing callbacks properly (hooks were processed in the correct order, but runtime callbacks in each hook were enqueued and executed in reverse compared to the order of registration).

The `AT<X>.unshift('callRuntimeCallbacks(on<X>);')` lines ensure runtime callbacks are always executed before compile time callbacks.

Related tests updated accordingly.


sbc100:
Seems reasonable to me.

The compile-time callback are already run in FIFO right?    So this change makes the runtime callbacks also be FIFO.   Was this something that changed recently with #23565 or were the runtime callback always LIFO historically?

sbc100:
Also can you update the PR description to something like `Change order of runtime callback from LIFO to FIFO`?  The word "Fix" seems like a maybe a stretch.

sbc100 (approved):
lgtm assuming @brendandahl is good with this change

brendandahl (approved):
Sounds reasonable to make it consistent.

test/test_other.py
```diff
@@ -2907,12 +2907,13 @@ def test_prepost2(self):
...
-      Module.preRun = [() => out('pre-run')];
+      Module.preRun = [() => out('pre-run-0'), () => out('pre-run-1')];
     ''')
     create_file('pre2.js', '''
-      Module.preRun.push(() => out('prepre'));
+      Module.preRun.push(() => out('pre-run-2'));
     ''')
-    self.do_runf('hello_world.c', 'prepre\npre-run\nhello, world!\n',
+    self.do_runf('hello_world.c',
+                 'pre-run-0\npre-run-1\npre-run-2\nhello, world!\n',
```
kripken, line 2916: It looks like this reverses the order of execution on this test. Is this not risking breaking user code as well?

kripken:
If we believe a breaking change is justified here, we should document that in the changelog.

lvlte:
Yes I added an entry in the changelog in a separate PR (#24022) as this one was already merged.

PR #24021, opened 2025-04-01
[dylink] Use an async iife to loadAsync loadWebAssemblyModule

hoodmane:
Split from #23619


src/lib/libdylink.js
```diff
@@ -871,13 +871,14 @@ var LibraryDylink = {
...
-        return WebAssembly.instantiate(binary, info).then(
-          (result) => postInstantiation(result.module, result.instance)
-        );
+        return (async function() {
+          if (binary instanceof WebAssembly.Module) {
+            var instance = new WebAssembly.Instance(binary, info);
+            return postInstantiation(binary, instance);
+          }
+          const result = await WebAssembly.instantiate(binary, info);
+          return postInstantiation(result.module, result.instance);
```
sbc100, line 880: Can we write this such that we only have one call to postInstantiation maybe?

Something like:

```
var instance;
if (binary instanceof WebAssembly.Module) {
  instance = new WebAssembly.Instance(binary, info);
} else {
  { binary, instance } = await WebAssembly.instantiate(binary, info);
}
return postInstantiation(binary, instance);
```

?

sbc100: Looks like you would need to add parans around the whole `{ binary, instance } = await WebAssembly.instantiate(binary, info)` statement BTW.. might warrant a comment :)

src/lib/libdylink.js
```diff
@@ -871,13 +871,18 @@ var LibraryDylink = {
...
 
       if (flags.loadAsync) {
-        if (binary instanceof WebAssembly.Module) {
-          var instance = new WebAssembly.Instance(binary, info);
-          return Promise.resolve(postInstantiation(binary, instance));
-        }
-        return WebAssembly.instantiate(binary, info).then(
-          (result) => postInstantiation(result.module, result.instance)
-        );
+        return (async function() {
```
sbc100, line 874: Use arrow function here?

sbc100 (approved):

PR #24023, opened 2025-04-01
Simplify `_scriptName` / `scriptDirectory` handling. NFC

kleisauke:
- Avoid defining `_scriptName` outside the wrapper function in `EXPORT_ES6` mode and for Node.js CommonJS builds.
- Simplify `scriptDirectory` resolution to `new URL('.', _scriptName).href` on the web.
- Add tests to verify that `scriptDirectory` is an absolute path across different build configurations.


sbc100:
Can you add some info to the PR description.

sbc100:
Presumably this comes with some code size deltas?

src/shell.js
```diff
@@ -137,27 +137,51 @@ var quit_ = (status, toThrow) => {
...
-#if SHARED_MEMORY && !MODULARIZE
-// In MODULARIZE mode _scriptName needs to be captured already at the very top of the page immediately when the page is parsed, so it is generated there
-// before the page load. In non-MODULARIZE modes generate it here.
-var _scriptName = (typeof document != 'undefined') ? document.currentScript?.src : undefined;
-
-#if ENVIRONMENT_MAY_BE_NODE
-if (ENVIRONMENT_IS_NODE) {
 #if EXPORT_ES6
-  _scriptName = typeof __filename != 'undefined' ? __filename : import.meta.url
+var _scriptName = import.meta.url;
```
sbc100, line 141: It looks like the declaration of this variable was complete elided previous unless `#if SHARED_MEMORY && !MODULARIZE`.   Can we not continue to default to not defining it at all?

kleisauke: The idea was to always ensure that `_scriptName` is defined, so that we can do our `scriptDirectory` logic on top of that (which is just `new URL('.', _scriptName).href` on the web). Note that `_scriptName` was also previously defined in `MODULARIZE` mode, but outside the wrapper function.

In `EXPORT_ES6` mode it was also previously defined outside the wrapper function, but I moved it to here since storing `import.meta.url` there isn't necessary (afaik).

kleisauke:
Marking as draft until I've written a PR description and rebasedlined the test expectations.

kleisauke:
Rebasedlined the test expectations and added a (short) PR description summarizing the changes. The changes to `shell_minimal.js` and `wasm_worker.js` are NFC and could probably be split into a separate PR.

This is ready for review now.

test/test_other.py
```diff
@@ -15895,3 +15895,48 @@ def test_mainScriptUrlOrBlob(self, es6):
...
+      Module['locateFile'] = (fileName, scriptDirectory) => {
+        assert(nodePath['isAbsolute'](scriptDirectory), `scriptDirectory (${scriptDirectory}) should be an absolute path`);
+        return scriptDirectory + fileName;
+      };
+      ''')
+    self.run_process([EMCC, '-o', 'hello_world.mjs',
+                      '--pre-js', 'pre.js',
+                      '--extern-post-js', test_file('modularize_post_js.js'),
+                      test_file('hello_world.c')] + args)
+    self.assertContained('hello, world!', self.run_js('hello_world.mjs'))
```
sbc100, line 15929: Can you use `do_runf` here to?

test/test_other.py
```diff
@@ -15895,3 +15895,48 @@ def test_mainScriptUrlOrBlob(self, es6):
...
+
+  @node_pthreads
+  def test_locate_file_abspath_pthread(self):
+    # Verify that `scriptDirectory` is an absolute path when `ENVIRONMENT_IS_WORKER`
+    self.set_setting('PROXY_TO_PTHREAD')
+    self.set_setting('EXIT_RUNTIME')
```
sbc100, line 15935: Can this just be another parameterization of `test_locate_file_abspath` above?

sbc100 (commented):
Great!

Out of interest, do any/all of the new tests fail without this change?

src/shell.js
```diff
@@ -139,24 +139,32 @@ var quit_ = (status, toThrow) => {
   throw toThrow;
 };
 
-#if SHARED_MEMORY && !MODULARIZE
+#if EXPORT_ES6
+var _scriptName = import.meta.url;
+#else
+#if ENVIRONMENT_MAY_BE_WEB
```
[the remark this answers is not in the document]
kleisauke: This `#else` block also contains logic for non-web environments, see e.g.:
https://github.com/emscripten-core/emscripten/blob/65b2a41c78f197468e6d8db5c2e48d893723a1c5/src/shell.js#L155-L167

So, it can't be combined to `#elif ENVIRONMENT_MAY_BE_WEB`.

test/test_other.py
kleisauke: Combined these test cases with commit a2d8b09f25e631764ca9374e80c3f3866465b9fc.

kleisauke:
BTW, do you know if `-sMINIMAL_RUNTIME` and `-sWASM_WORKERS` support the `-sEXPORT_ES6` setting? It seems to always use `__filename`, which won't work in ESM modules.

test/test_other.py
sbc100: It does yes.  You can enable it either by passing `-sEXPORT_ES6` explicitly in which case the underlying `build` function will choose as filename that ends in `.mjs`, or you can add `output_suffix='.mjs'`.  Either way should work.

sbc100 (approved):

sbc100:
> BTW, do you know if `-sMINIMAL_RUNTIME` and `-sWASM_WORKERS` support the `-sEXPORT_ES6` setting? It seems to always use `__filename`, which won't work in ESM modules.

It seems unlikely that was ever tested.  I think its unlike that `MINIMAL_RUNTIME` and `EXPORT_ES6` are used together.

test/test_other.py
kleisauke: Great! Simplified to `do_runf()` with commit 46dd3f7c668e7886cb288360448f4383d9fc5290.

sbc100:
> > Out of interest, do any/all of the new tests fail without this change?
> 
> The new tests also seem to pass on the main branch. IIRC, the `other.test_locate_file_abspath_pthread` test used to fail, but it was likely fixed when the pthread `*.worker.js` file was inlined into the main output file.
> 
> Given this, this PR is (likely) a non-functional change now. I'll add the NFC keyword to the title.

Can you add this info the to the PR?    i.e. explain why the new tests are useful, even though they pass before and after this PR?

sbc100 (approved):

kleisauke:
> Can you add this info the to the PR?

Sure, I've just added another bullet point with this info to the PR description.

> I love this change BTW. I've tried to do something like this myself several times in the past but always given up. Thanks!

No problem! :)

PR #24025, opened 2025-04-01
Allow non-ASCII environment variables.

sbc100:
Fixes: #24024


src/lib/libwasi.js
```diff
@@ -84,27 +84,31 @@ var WasiLibrary = {
...
     var bufSize = 0;
-    getEnvStrings().forEach((string, i) => {
+    var envp = 0;
+    for (var string of getEnvStrings()) {
       var ptr = environ_buf + bufSize;
-      {{{ makeSetValue('__environ', `i*${POINTER_SIZE}`, 'ptr', POINTER_TYPE) }}};
-      stringToAscii(string, ptr);
+      {{{ makeSetValue('__environ', 'envp', 'ptr', POINTER_TYPE) }}};
+      stringToUTF8(string, ptr, Infinity);
       bufSize += string.length + 1;
```
kripken, line 109: Isn't this using the length of the JS string, not the encoded C string? In non-ascii those can diverge iiuc

test/test_other.py
```diff
@@ -8728,7 +8728,7 @@ def test_override_js_execution_environment(self):
 
   def test_override_c_environ(self):
     create_file('pre.js', r'''
-      Module.preRun = () => { ENV.hello = 'world'; ENV.LANG = undefined; }
+      Module.preRun = () => { ENV.hello = '💩 world'; ENV.LANG = undefined; }
```
kripken, line 8731: :laughing:

src/lib/libwasi.js
sbc100: Oooops good catch.

kripken (approved):

PR #24028, opened 2025-04-01
Improve performance of linker symbol cache under high contention. NFC

sbc100:
This drastically improves the performance of running the test suite even though most users probably won't see any effect.

Prior to this change the symbol generation process itself always happened with the symbol list lock held.  Now in the most common cases the symbol list is generated before we grab the lock.

We really should look into a better way to do this without a global symbol list lock, but this is a good interim step:

before:

- hot: 0m25.259s
- cold: 1m36.121s

after:

- hot: 0m20.220s
- cold: 0m22.324s

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -220,19 +220,32 @@ def get_js_sym_info():
     input_files.append(read_file(jslib))
   content = '\n'.join(input_files)
   content_hash = hashlib.sha1(content.encode('utf-8')).hexdigest()
+  library_syms = None
+
+  cache_file = f'symbol_lists/{content_hash}.json'
+  # When we are racing with other processes the existence of the file
+  # at this point in time only means its likely a cache hit, since another
+  # process could be performing a cache cleanup.
+  cache_miss_likely = not os.path.exists(cache.get_path(cache_file))
+
+  if cache_miss_likely:
+    library_syms = generate_js_sym_info()
+  else:
+    library_syms = None
 
   def build_symbol_list(filename):
     """Only called when there is no existing symbol list for a given content hash.
     """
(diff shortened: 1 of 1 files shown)


tools/link.py
```diff
@@ -220,19 +220,32 @@ def get_js_sym_info():
     input_files.append(read_file(jslib))
   content = '\n'.join(input_files)
   content_hash = hashlib.sha1(content.encode('utf-8')).hexdigest()
+  library_syms = None
+
+  cache_file = f'symbol_lists/{content_hash}.json'
+  # When we are racing with other processes the existence of the file
+  # at this point in time only means its likely a cache hit, since another
```
kripken, line 227:
```suggestion
  # at this point in time only means it's likely a cache hit, since another
```

tools/link.py
```diff
@@ -220,19 +220,32 @@ def get_js_sym_info():
...
+  # process could be performing a cache cleanup.
```
kripken, line 228: I don't follow the sentence after `, since`. Is the idea "if the file exists now, we will likely still have it later, but we do need to be careful of a cache clearing in the middle" - ?

kripken: If so, that sentence might work, plus "but if the file does not exist now, a cache miss is likely (unless we get lucky and another process generates it), and in that case we generate the info now rather than later, as it is more efficient to do so now before we take the cache lock".

sbc100: Yes, exactly there could be another process performing a cache cleanup between our time-of-check and time-of-use of the file.

sbc100:
Actually I decide to just rewrite this without using `cache.get` at all: #24029

PR #24029, opened 2025-04-01
Improve performance of linker symbol cache under high contention. NFC

### LINKED ISSUE #18622: Constant "generating system asset: symbol_lists/<hash>" in build
After updating recently, I keep getting these prints
```
cache:INFO: generating system asset: symbol_lists/4b1c671fb19fde79292783da0799997fdd1b96c4.txt... (this will be cached in "C:\emsdk\emscripten\main\cache\symbol_lists\4b1c671fb19fde79292783da0799997fdd1b96c4.txt" for subsequent builds)
cache:INFO:  - ok
```
in my builds, which looks like unnecessary noise. Can those be avoided?

sbc100:
This drastically improves the performance of running the test suite even though most users probably won't see any effect.

Prior to this change the symbol generation process itself always happened with a global symbol list lock held.   Now we use a per-symbol-list lock which means that only linker processes with identical linker flags with contend for the lock.

before:

- hot: 0m25.259s
- cold: 1m36.121s

after:

- hot: 0m21.371s
- cold: 0m20.350s

This also happens to fix #18622 since it avoids using `cache.get()` which does logging.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -220,35 +220,37 @@ def get_js_sym_info():
     input_files.append(read_file(jslib))
   content = '\n'.join(input_files)
   content_hash = hashlib.sha1(content.encode('utf-8')).hexdigest()
+  library_syms = None
 
-  def build_symbol_list(filename):
-    """Only called when there is no existing symbol list for a given content hash.
-    """
-    library_syms = generate_js_sym_info()
+  cache_file = str(cache.get_path(f'symbol_lists/{content_hash}.json'))
 
-    write_file(filename, json.dumps(library_syms, separators=(',', ':'), indent=2))
+  utils.safe_ensure_dirs(cache.get_path('symbol_lists'))
+  with filelock.FileLock(cache_file + '.lock'):
+    if os.path.exists(cache_file):
+      # Cache hit, read the file
+      library_syms = json.loads(read_file(cache_file))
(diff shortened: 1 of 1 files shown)


tools/link.py
```diff
@@ -220,35 +220,37 @@ def get_js_sym_info():
...
+  cache_file = str(cache.get_path(f'symbol_lists/{content_hash}.json'))
+
+  utils.safe_ensure_dirs(cache.get_path('symbol_lists'))
+  with filelock.FileLock(cache_file + '.lock'):
+    if os.path.exists(cache_file):
+      # Cache hit, read the file
+      library_syms = json.loads(read_file(cache_file))
+    else:
+      # Cache miss, generate the symbol list and write the file
+      library_syms = generate_js_sym_info()
```
kripken, line 234: The description says

> Prior to this change the symbol generation process itself always happened with the symbol list lock held. Now in the most common cases the symbol list is generated before we grab the lock.

But this line is run under the lock from 228?

It looks like only the cache limit check was moved out of the lock - was that really so slow?

sbc100: But now we use a per-file lock, not a global lock, so only processes with the exact same linker flags will be in contention here.

sbc100: I update the description.

PR #24033, opened 2025-04-02
Elaborate on what "runtime callbacks" means in  changelog

lvlte:
See #24012.


sbc100 (approved):

ChangeLog.md
```diff
@@ -23,8 +23,11 @@ See docs/process.md for more on how version tagging works.
 - Added experimental support for Wasm ESM integration with
   `-sWASM_ESM_INTEGRATION`. This is currently only supported in node behind a
   flag and not in any browsers. (#23985)
-- Runtime callbacks are now enqueued and executed FIFO (historically was LIFO),
-  which matches the behavior of compile-time callbacks. (#24012)
+- Runtime callbacks (ie. registered in `Module['preInit']`, `Module['preRun']`,
+  or `Module['postRun']`, or using `addOnPreRun()`, `addOnInit()`,
+  `addOnPostCtor()`, `addOnPreMain()`, `addOnExit()`, `addOnPostRun()`) are now
+  enqueued and executed FIFO (historically was LIFO), which matches the behavior
+  of compile-time callbacks. (#24012)
```
kripken, line 30: What does "LIFO/FIFO" mean for `Module['preInit']` etc., which is an array?

I missed that we reversed the order of processing of those arrays. Can we revert that part? It seems separable from how we handle the calls. And I worry the arrays are used more, so there is a greater risk of users breaking if we flip things. Also, I am not sure there is a "natural" order to the handling of the arrays (I can see arguments both ways, at least).

sbc100: There was no change the processing of `Module['preInit']` since those get invoked directly, with the last one running first:  https://github.com/emscripten-core/emscripten/blob/e014359fcb41dcf901765ec51cc3e0a236310453/src/postamble.js#L291-L296

Here you can see that callbacks are run by calling `shift()` on the array so the first elements get run first.

For `preRun` and `postRun` we use `addOnPreRun` and `addOnPostRun`, so I think we might be need actually change the callsites to preserve the previous behaviour:  


https://github.com/emscripten-core/emscripten/blob/e014359fcb41dcf901765ec51cc3e0a236310453/src/preamble.js#L176-L181

https://github.com/emscripten-core/emscripten/blob/e014359fcb41dcf901765ec51cc3e0a236310453/src/preamble.js#L278-L283

sbc100: Actually I think we do want to document that the order of `preRun` and `postRun` did change.  Using FIFO here actually matches the `preInit` behaviour.

kripken: Can you clarify what you mean by "FIFO" on an array? IIANM that is ambiguous.

In other words, `[1,2,3]` - what is the FIFO item to take next? It depends if we built up the array by pushing to the end `[1] ; [1,2] ; [1,2,3]` or unshifting to the start `[3] ; [2,3] ; [1,2,3]`. In the former case FIFO is 1, in the latter it is 3. The final state of the array `[1,2,3]` does not tell us which item was "first".

kripken: I do agree that consistent behavior on `Module.preInit/preRun` etc. makes sense, and can justify a breaking change.

kripken: > Perhaps its better to say the callback in arrays are run from first to last, or left to right?

Yes, that would be unambiguous.

@lvlte Please clarify this in detail in the comment. This seems like a breaking change that can trip up our users, so we want to explain it in the best way we can.

Explaining the justification for the breakage is also important so that they do not read this and respond "wait why are they breaking my code for no reason".

lvlte: No problem, I'm going to add some details. By FIFO for an array of functions I mean first registered first executed. The `preInit` callbacks were already FIFO (ie.). `preRun` and `postRun` are using the corresponding `addOn<X>` function, which historically has enqueued callbacks in reverse (last in first executed).

ChangeLog.md
```diff
@@ -23,8 +23,14 @@ See docs/process.md for more on how version tagging works.
...
-- Runtime callbacks are now enqueued and executed FIFO (historically was LIFO),
-  which matches the behavior of compile-time callbacks. (#24012)
+- Runtime callbacks registered in `Module['preRun']` or `Module['postRun']`, or
+  using `addOnPreRun()`, `addOnInit()`, `addOnPostCtor()`, `addOnPreMain()`,
+  `addOnExit()`, or `addOnPostRun()`, are now enqueued and executed following
+  the order of registration (was historically in reverse). While this might be
+  a breaking change for some users, the intention is to be more consistent by
+  making those callbacks match the behavior of `Module['preInit']` and compile
+  time callbacks (rather than the contrary, as we generally expect an array of
+  functions to be executed left to right). (#24012)
```
kripken, line 33: Much better, thanks, and this looks good aside from the arrays. I still cannot understand from this text how the arrays are handled. That is:

`"Runtime calledbacks registered in Module.preRun are enqueued and executed following the order of registration"` => what does that mean? If I, the user, provide `Module.preRun = [a,b,c]` then what is the "order of registration"? I did not register `a,b,c` myself, I provided an array.

I could guess that the order is the natural order in the array, `[a,b,c]`, but it would be good to clarify that. Perhaps adding something like this:

"For example, for `Module['preRun'] = [a, b]`, the previous behavior was to run b and then a. The new behavior is to run a and then b." 

A user reading that should have an easy time figuring out exactly what changed, I think.

ChangeLog.md
```diff
@@ -23,8 +23,16 @@ See docs/process.md for more on how version tagging works.
...
   `-sWASM_ESM_INTEGRATION`. This is currently only supported in node behind a
   flag and not in any browsers. (#23985)
-- Runtime callbacks are now enqueued and executed FIFO (historically was LIFO),
-  which matches the behavior of compile-time callbacks. (#24012)
+- Runtime callbacks registered in `Module['preRun']` or `Module['postRun']`, or
+  using `addOnPreRun()`, `addOnInit()`, `addOnPostCtor()`, `addOnPreMain()`,
+  `addOnExit()`, or `addOnPostRun()`, are now enqueued and executed following
+  the order of registration (ie. `Module['preRun'] = [a, b]`, or equivalently
+  `addOnPreRun(a); addOnPreRun(b);` will run `a` then `b`; the previous behavior
+  was to run `b` then `a`. While this might be a breaking change for some users,
```
```suggestion
  was to run `b` then `a`). While this might be a breaking change for some users,
```

ChangeLog.md
```diff
@@ -23,8 +23,16 @@ See docs/process.md for more on how version tagging works.
 - Added experimental support for Wasm ESM integration with
   `-sWASM_ESM_INTEGRATION`. This is currently only supported in node behind a
   flag and not in any browsers. (#23985)
-- Runtime callbacks are now enqueued and executed FIFO (historically was LIFO),
-  which matches the behavior of compile-time callbacks. (#24012)
+- Runtime callbacks registered in `Module['preRun']` or `Module['postRun']`, or
+  using `addOnPreRun()`, `addOnInit()`, `addOnPostCtor()`, `addOnPreMain()`,
+  `addOnExit()`, or `addOnPostRun()`, are now enqueued and executed following
+  the order of registration (ie. `Module['preRun'] = [a, b]`, or equivalently
```
kripken, line 29:
```suggestion
  the order of registration (i.e. `Module['preRun'] = [a, b]`, or equivalently
```

PR #24034, opened 2025-04-02
Fix FindOpenGL.cmake when included more than once

sbc100:
Fixes a bug introduced in #23705.


test/cmake/find_modules/CMakeLists.txt
```diff
@@ -3,16 +3,19 @@ project(find_modules)
 
 add_executable(test_prog test.c)
 
+find_package(OpenGL REQUIRED)
 find_package(OpenGL REQUIRED)
```
kripken, line 7: Why do we need these duplicate lines here and bwlow?

sbc100: Just to test that its safe to find the package more than once.

test/cmake/find_modules/CMakeLists.txt
```diff
@@ -3,16 +3,19 @@ project(find_modules)
 
 add_executable(test_prog test.c)
 
+find_package(OpenGL REQUIRED)
```
kripken, line 6:
```suggestion
# find the package twice to check we do not error
find_package(OpenGL REQUIRED)
```

test/cmake/find_modules/CMakeLists.txt
kripken: Oh, maybe add a comment then. Otherwise the test code looks like it might be a copy-paste error I think.

kripken (approved):

kripken (approved):

PR #24035, opened 2025-04-02
Cleanup lock files in UNIX

sbc100:
See https://github.com/benediktschmitt/py-filelock/issues/31 and https://stackoverflow.com/a/51070775.

Without this change the `cache/symbol_lists` directory ends up full of many lock files that might never get cleaned up (At least since #24029 landed).

I'm also sending this change upstream: https://github.com/tox-dev/filelock/pull/408


kripken (approved):
Not an expert on this but it seems right and upstream approved it.

PR #24039, opened 2025-04-02
Group JS library exports together, along with JS runtime exports. NFC

sbc100:
This is in preparation for #24005 which delays the processing and creation of Module object itself until later in some cases.


src/modules.mjs
```diff
@@ -540,8 +545,25 @@ function exportRuntime() {
   return results.join('\n') + '\n';
 }
 
+function exportLibrarySymbols() {
+  const results = ['// Begin JS library exports'];
+  for (const ident of librarySymbols) {
+    if (!isDecorator(ident) && !isInternalSymbol(ident)) {
+      if (EXPORT_ALL || EXPORTED_FUNCTIONS.has(ident)) {
```
brendandahl, line 552: This can be moved into the `if` above.

src/modules.mjs
```diff
@@ -509,6 +513,7 @@ function exportRuntime() {
   }
 
   const exports = runtimeElements.map(maybeExport);
+  exports.unshift('// Begin runtime exports')
```
brendandahl, line 516:
```suggestion
  exports.unshift('// Begin runtime exports');
```

brendandahl (approved):
2 nits.

PR #24041, opened 2025-04-02
Mark more `other` tests as being slow. NFC

sbc100:
These tests were all taking over 20 seconds on my cloudtop. With this change I can now run the whole other test suite with --skip-slow in under 1 minute.


sbc100:
Both these tests were taking over 20 seconds on my cloudtop.

kripken (approved):

PR #24053, opened 2025-04-05
[embind] Fix misindexed template parameters which prevented policies from being applied to class functions or constructors

canislupaster:
Hi, trying to `allow_raw_pointers()` in a class method bound with embind when a custom signature was given as a template argument to `.class_function` failed. It's easy to see why by comparing the definition of either `RegisterClassMethod` or `RegisterClassConstructor` with their usage:

```cpp
template<typename ReturnType, typename ThisType, typename... Args>
struct RegisterClassMethod<ReturnType (ThisType, Args...)> {

    template <typename ClassType, typename Callable, typename... Policies>
    static void invoke(const char* methodName,
...
```

and its invocation in `class_function`:

```cpp
using invoker = internal::RegisterClassConstructor<
            typename std::conditional<std::is_same<Signature, internal::DeduceArgumentsTag>::value,
                                      Callable,
                                      Signature>::type>;

invoker::template invoke<ClassType, Policies...>(callable);
```

which will skip the `Callable` template argument in some specializations (and `Callable` will become the return value policy). To fix this, I used the same approach used by the getters/setter `PropertyTag`s (c.f. diff). Please let me know if anything is off.

### DIFF
--- a/system/include/emscripten/bind.h
+++ b/system/include/emscripten/bind.h
@@ -820,8 +820,8 @@ inline T* getContext(const T& t) {
     return ret;
 }
 
-template<typename Accessor, typename ValueType>
-struct PropertyTag {};
+template<typename Func, typename ValueTypeOrSignature>
+struct FunctionTag {};
 
 template<typename T>
 struct GetterPolicy;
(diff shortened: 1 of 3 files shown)


canislupaster:
I'm not sure if something with my environment is different from CircleCI's (MacOS / Apple Silicon, latest emsdk, no warnings about version mismatches for clang/binaryen/etc), but I get the old/incorrect code size locally. This seems like a terrible solution, but I can substitute the "correct" sizes from https://github.com/emscripten-core/emscripten/pull/24055 .

sbc100:
> I'm not sure if something with my environment is different from CircleCI's (MacOS / Apple Silicon, latest emsdk, no warnings about version mismatches for clang/binaryen/etc), but I get the old/incorrect code size locally. This seems like a terrible solution, but I can substitute the "correct" sizes from #24055 .

Did you do `emsdk instal tot` to get the tip-of-tree release?

brendandahl:
Sorry I thought this merged. It looks like you can drop the rebaseline changes and this should pass.

PR #24061, opened 2025-04-07
[esm-integration] Allow JS library symbols to be exported

sbc100:
See #24060

### DIFF
--- a/src/modules.mjs
+++ b/src/modules.mjs
@@ -558,6 +558,9 @@ function exportLibrarySymbols() {
 }
 
 function exportJSSymbols() {
+  // In WASM_ESM_INTEGRATION mode JS library symbols are marked with `export`
+  // at the point of declaration.
+  if (WASM_ESM_INTEGRATION) return '';
   return exportRuntimeSymbols() + '  ' + exportLibrarySymbols();
 }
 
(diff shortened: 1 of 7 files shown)


src/modules.mjs
```diff
@@ -558,6 +558,7 @@ function exportLibrarySymbols() {
 }
 
 function exportJSSymbols() {
+  if (WASM_ESM_INTEGRATION) return '';
```
brendandahl, line 561: Comment why would be helpful here.

tools/link.py
```diff
@@ -2113,12 +2113,17 @@ def create_worker_file(input_file, target_dir, output_file, options):
...
+  js_exports = []
   for f in settings.USER_EXPORTS:
     if f == '_main' and '__main_argc_argv' in settings.WASM_EXPORTS:
       wasm_exports.append('__main_argc_argv as main')
     else:
-      wasm_exports.append(shared.demangle_c_symbol_name(f))
+      if f in settings.WASM_EXPORTS:
+        wasm_exports.append(shared.demangle_c_symbol_name(f))
+      else:
+        js_exports.append(f)
```
brendandahl, line 2124: This could be "outdented"  above to an `elif` and `else`

sbc100: Nice

brendandahl (approved):

PR #24069, opened 2025-04-08
Rename come JS libraries to match their native counterparts. NFC

sbc100:


kripken:
Does this not affect e.g. `./emcc test/hello_world.c -lxlib.js`? That works now but after this change it would need to be `./emcc test/hello_world.c -lX11.js`

sbc100:
> Does this not affect e.g. `./emcc test/hello_world.c -lxlib.js`? That works now but after this change it would need to be `./emcc test/hello_world.c -lX11.js`

Yes, that is true.   However, the standard way that this library was includes has always been via `-lX11` (without the `.js` suffix).   I dont think that folks are generally including system libraries via `-lxlib.js`, but I guess its worth considering more and at least adding a changelog entry.

PR #24074, opened 2025-04-08
Move test_modularize_instance into test_core.py. NFC

sbc100:
Move this test alongside its peer (test_esm_integration).  These are the only tests we have for these modes and its useful to be able to test other modes such as wasm64.


brendandahl (approved):
I thought there were issues with O3, but if it's working great!

PR #24089, opened 2025-04-10
Make sure postRun is called after main() if main suspends

### LINKED ISSUE #12402: Asyncify and postRun()
I noticed that when using Asyncify, the `Module.postRun()` only runs once after the first call of `emscripten_sleep()`. Is this behaviour expected? Ideally we want the `postRun()` to run after then `main()` function returns and the programme exits, not when an async function called. Otherwise I don't know when my programme has exited to do my clean-up routines. Thanks.
Amazing work with Emscripten + Asyncify 👍

msorvig:
await callMain(), which is an async function if JSPI is enabled.

Fixes #12402.

### DIFF
--- a/src/postamble.js
+++ b/src/postamble.js
@@ -169,7 +169,7 @@ function run() {
     return;
   }
 
-  function doRun() {
+  {{{ asyncIf(ASYNCIFY == 2) }}}function doRun() {
     // run may have just been called through dependencies being fulfilled just in this very frame,
     // or while the async setStatus time below was happening
 #if ASSERTIONS
@@ -199,9 +199,9 @@ function run() {
 #if HAS_MAIN
     var noInitialRun = {{{ makeModuleReceiveExpr('noInitialRun', !INVOKE_RUN) }}};
 #if MAIN_READS_PARAMS
-    if (!noInitialRun) callMain(args);
+    if (!noInitialRun) {{{ awaitIf(ASYNCIFY == 2) }}}callMain(args);
 #else
-    if (!noInitialRun) callMain();
+    if (!noInitialRun) {{{ awaitIf(ASYNCIFY == 2) }}}callMain();
 #endif
 #else
 #if ASSERTIONS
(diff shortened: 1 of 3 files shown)


msorvig:
This possibly requires more fine-grained async usage with ASYNCIFY == 2 checks. 

It should be possible to await callMain() in all cases (as I understand it await forwards the value if is not a Promise), but can we make doRun() async unconditionally?

sbc100:
I think we should probably add a test for this.

src/postamble.js
```diff
@@ -169,7 +169,7 @@ function run() {
     return;
   }
 
-  function doRun() {
+  async function doRun() {
```
brendandahl, line 172: You'll want to do the same thing as `callMain` higher up in the file with `{{{ asyncIf(ASYNCIFY == 2) }}}`
and `{{{ awaitIf(ASYNCIFY == 2) }}}`.

msorvig: Thanks - awaitIf was the missing piece of the puzzle.

msorvig:
Regarding tests I was planning to extend or parameterize test_prepost2 in test_other.py, but I'm having some trouble running it in an environment where JSPI is enabled. What would the most straightforward way to do this be?

brendandahl:
You'll need a canary version of node. The testing infrastructure uses https://nodejs.org/download/v8-canary/v24.0.0-v8-canary20250303cbbb63ede3/node-v24.0.0-v8-canary20250303cbbb63ede3-linux-x64.tar.xz

Then point `NODE_JS` in your `.emscripten` config in the node binary.

There are few different version if you're not on linux:
https://nodejs.org/download/v8-canary/v24.0.0-v8-canary20250303cbbb63ede3/

test/hello_world_suspend.c
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright 2011 The Emscripten Authors.  All rights reserved.
```
sbc100, line 2: 2025

test/hello_world_suspend.c
```diff
@@ -0,0 +1,15 @@
+/*
```
sbc100, line 1: Can you move this file to `test/other/`

sbc100 (approved):

Issue #24086, opened 2025-04-10
Exception handling helpers do not work with ASSERTIONS=1


canislupaster:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.7-git (a53b1a34f0b31f05f2045f2d4aa1343f75180693)
Copyright (C) 2025 the Emscripten authors (see AUTHORS.txt)
This is free and open source software under the MIT license.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

**Failing command line in full:**

`em++ ./main.cpp -o main.js -fwasm-exceptions -sEXPORT_EXCEPTION_HANDLING_HELPERS=1`

(Any C++ source file will work.)

This outputs a `main.js` where `getExceptionMessage` is in `unexportedSymbols` due to the implicit `-sASSERTIONS=1`. Importing `main.js` always causes an error since `ASSERTIONS` overrides the getter of anything in `unexportedSymbols` to throw an error: https://github.com/emscripten-core/emscripten/blob/c2d94c86c9e977b2a772e8364783ebf158914322/src/modules.mjs#L531

The test in `test_core.py`, however, explicitly disables ASSERTIONS: https://github.com/emscripten-core/emscripten/blob/c2d94c86c9e977b2a772e8364783ebf158914322/test/test_core.py#L1380

**I don't think disabling ASSERTIONS should be required to use exception helpers?** If it needs to be, then you can probably dismiss this.

This stems from https://github.com/emscripten-core/emscripten/blob/c2d94c86c9e977b2a772e8364783ebf158914322/tools/link.py#L1839 where the helper functions are added to `EXPORTED_FUNCTIONS` instead of `EXPORTED_RUNTIME_METHODS`, and changing this lets me import the file.

However, now `getExceptionMessage` with `ASSERTIONS` and Emscripten exceptions somehow triggers a `WebAssembly.RuntimeError` for `memory access out of bounds`. The `__get_exception_message` in `system/lib/libcxxabi/src/cxa_exception_js_utils.cpp` used from `src/lib/libcore.js` apparently dies while returning. I have no idea when this error is emitted or how `ASSERTIONS` could impact that.

Please let me know if I'm missing anything. Thanks for reading.

sbc100:
[...]

PR #24091, opened 2025-04-10
Fix unexported symbol check for library symbols exported with EXPORTED_FUNCTIONS

sbc100:
This bug meant that library symbol exported via `EXPORTED_FUNCTIONS` were incorrectly being reported as not exported using unexportedRuntimeSymbol

See #24086


brendandahl (approved):
r+ w/ nit

PR #24092, opened 2025-04-10
Update test_esm_integration expectations

sbc100:
This was not updated as part of #24079 but we didn't notice because we were not running the test as part of CI.


sbc100:
Landing TBR

PR #24103, opened 2025-04-14
Allow __indirect_function_table to be DCE'd

sbc100:
Prior this this change `__indirect_function_table` would be seen as an unexpected user export and therefore added to `user_requested_exports` which in turn would mean it would be marked as metaDCE root.


kripken (approved):
Nice!

PR #24107, opened 2025-04-15
Add support for passing a response file to `file_packager`

iscgar:
Because calls to `file_packager` go through the shell wrapper, this can on Windows due to a "command line is too long" error (e.g. with a long list of `--preload-file`), even if a response file was provided to `emcc` in order avoid the issue.

Fix it by supporting response files in `file_packager`, and changing `link` so that it uses a response file where needed.

----
#### Original PR description (superseded, kept for posterity)

Going through the batch scripts on Windows can end up failing due to a "command line is too long" error, even if a response file was provided to `emcc` in order avoid the issue.

Since this code already executes in a Python interpreter, there's no reason not to use that interpreter to execute the target script directly instead of going through the batch file, thus allowing the use of longer command lines.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -3739,6 +3739,22 @@ More info: https://emscripten.org
 
     self.assertTextDataIdentical(clean(proc.stdout), clean(proc2.stdout))
 
+  def test_file_packager_response_file(self):
+    filenames = [f'foo.{i:032}' for i in range(4096)]
+
+    create_file('src.c', 'int main() { return 0; }\n')
+    create_file('data.txt', '')
+
+    response_data = '\n'.join(f'--preload-file data.txt@{f}' for f in filenames)
+    self.assertGreater(len(response_data), (1 << 16))
+    create_file('data.rsp', response_data)
+
+    self.run_process([EMCC, 'src.c', '@data.rsp'])
+    data = read_file('a.out.js')
+
+    for f in filenames:
+      self.assertTrue(f'"/{f}"' in data)
+
   def test_file_packager_separate_metadata(self):
     # verify '--separate-metadata' option produces separate metadata file
     ensure_dir('subdir')
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -84,6 +84,7 @@ __rootdir__ = os.path.dirname(__scriptdir__)
 sys.path.insert(0, __rootdir__)
 
 from tools import shared, utils, js_manipulation
+from tools.response_file import substitute_response_files
 
 
 DEBUG = os.environ.get('EMCC_DEBUG')
@@ -368,12 +369,18 @@ def main():  # noqa: C901, PLR0912, PLR0915
   See the source for more details.''')
     return 1
 
-  data_target = sys.argv[1]
+  # read response files very early on
+  try:
+    args = substitute_response_files(sys.argv[1:])
+  except IOError as e:
+    shared.exit_with_error(e)
+
+  data_target = args[0]
   data_files = []
   plugins = []
   leading = ''
 
-  for arg in sys.argv[2:]:
+  for arg in args[1:]:
     if arg == '--preload':
       leading = 'preload'
     elif arg == '--embed':
--- a/tools/link.py
+++ b/tools/link.py
@@ -3070,7 +3070,8 @@ def package_files(options, target):
     file_args += ['--obj-output=' + object_file]
     rtn.append(object_file)
 
-  cmd = [shared.FILE_PACKAGER, shared.replace_suffix(target, '.data')] + file_args
+  cmd = building.get_command_with_possible_response_file(
+    [shared.FILE_PACKAGER, shared.replace_suffix(target, '.data')] + file_args)
   if options.preload_files:
     # Preloading files uses --pre-js code that runs before the module is loaded.
     file_code = shared.check_call(cmd, stdout=PIPE).stdout


sbc100:
> Going through the batch scripts on Windows can end up failing due to a "command line is too long" error, even if a response file was provided to `emcc` in order avoid the issue.

Can you explain under what circumstances this can happen?  i.e. can you give a specific example of when emcc calls another .bat file without using a response file?

tools/shared.py
```diff
@@ -127,6 +127,17 @@ def run_process(cmd, check=True, input=None, *args, **kw):
...
+  if cmd and cmd[0].endswith('.py'):
+    cmd = [sys.executable, '-E'] + cmd
+    # $PYTHON -E will not ignore _PYTHON_SYSCONFIGDATA_NAME, which is an internal
+    # of cpython used in cross compilation via setup.py.
+    environ = kw.get('env')
+    if environ is None:
+      environ = os.environ.copy()
+    if '_PYTHON_SYSCONFIGDATA_NAME' in environ:
+      del environ['_PYTHON_SYSCONFIGDATA_NAME']
+      kw['env'] = environ
```
sbc100, line 140: I would prefer not to modify this function, but instead replace `bat_suffix` with something like `make_py_cmd`

tools/shared.py
```diff
@@ -127,6 +127,17 @@ def run_process(cmd, check=True, input=None, *args, **kw):
...
+  # Handle calls to the Python scripts by going through the python executable
+  if cmd and cmd[0].endswith('.py'):
+    cmd = [sys.executable, '-E'] + cmd
+    # $PYTHON -E will not ignore _PYTHON_SYSCONFIGDATA_NAME, which is an internal
+    # of cpython used in cross compilation via setup.py.
+    environ = kw.get('env')
+    if environ is None:
+      environ = os.environ.copy()
+    if '_PYTHON_SYSCONFIGDATA_NAME' in environ:
+      del environ['_PYTHON_SYSCONFIGDATA_NAME']
```
sbc100, line 139: How about if we simple did `del os.environ['_PYTHON_SYSCONFIGDATA_NAME']` somewhere at the top level instead of duplicating the environment here?

iscgar:
When `--preload-file` is given at the linking step, `emcc` calls `file_packager` through the batch file. This is an issue that was [encountered in the WASM build of SolveSpace](https://github.com/solvespace/solvespace/pull/1566#issuecomment-2805676416), and we had to resort to using relative paths as a means to shorten the command line and work around this issue.

tools/shared.py
iscgar: The problem with such a function is that it doesn't handle the removal of `_PYTHON_SYSCONFIGDATA_NAME` from the environment.

sbc100:
> When `--preload-file` is given at the linking step, `emcc` calls `file_packager` through the batch file. This is an issue that was [encountered in the WASM build of SolveSpace](https://github.com/solvespace/solvespace/pull/1566#issuecomment-2805676416), and we had to resort to using relative paths as a means to shorten the command line and work around this issue.

Ah, I see.   Perhaps we can modify the calling of file_packager to use a response file when needed.   We already have a helper called `get_command_with_possible_response_file` that can do this.

sbc100:
Do you have a long list of files that you want to pass to `--preload-file`?   I also wonder if it might help if `--preload-file` could itself take a response file?   Like you can already do with `-sEXPORTED_FUNCTION=@filename`?

tools/shared.py
sbc100: Ah, but assuming that a .bat file was used to launch to initial process I think we can maybe assume that `_PYTHON_SYSCONFIGDATA_NAME` was cleared already?

tools/shared.py
iscgar: Because users never call the Python scripts directly, we're not supposed to have `_PYTHON_SYSCONFIGDATA_NAME` in the environment anyway (since the batch wrappers unset it), but I was trying to be careful and handle the case of a potential call that uses a custom environment and accidentally sets it. I didn't find such a case in the code, but I wanted to handle it just in case I missed something. If that's not a case that should be handled, I'm finding with doing the removal at the top level, but I would appreciate some guidance as to best place for this code.

sbc100: I think the `get_command_with_possible_response_file` approach when calling the file_packager might actually be the best/simplest solution for now.

tools/shared.py
```diff
@@ -127,6 +127,17 @@ def run_process(cmd, check=True, input=None, *args, **kw):
   sys.stderr.flush()
```
sbc100, line 127: Perhaps we could add an assertion here. e.g `assert len(shlex_join(cmd)) <= MAX_CMDLINE_LEN`.

iscgar: We'll get a failure from `subprocess.run()` in such a case, so I don't see the point in having to define a maximum command line length for each supported platform and adding an assertion here.

tools/shared.py
sbc100: Lets just assume `_PYTHON_SYSCONFIGDATA_NAME` was removed by the outer batch file.  The removal of `_PYTHON_SYSCONFIGDATA_NAME` is a pretty obscure/strange thing we need to do IIRC its only needed when building CPython itself.

tools/shared.py
sbc100: Fair enough.  As long as the OS message to clear enough.

tools/shared.py
iscgar: Indeed, there's still a limit even with a direct process call without going through the batch file, but it is much higher (32KiB vs. 8KiB for Windows and CMD, respectively) so it will also help in the general case, and `file_packager` can additionally gain support for response files in order to work with even longer command lines.

tools/shared.py
iscgar: Another issue with a `make_py_cmd` approach is that we no longer have a single string, but a list of strings (because we need to add `[sys.executable, '-E']` in front of it), which means some tool constants at the end of `shared.py` would be plain string while others lists, causing confusion.

tools/shared.py
iscgar: At least on Windows, the returned error is `ERROR_FILENAME_EXCED_RANGE`, which results in the following Python exception: `FileNotFoundError: [WinError 206] The filename or extension is too long`.

sbc100:
> That would probably be a good solution in this case, although there may be other cases which I just haven't encountered that may fail similarly, and with this change we can at least lift the command line length limit from the one imposed by CMD to the larger one supported by the Windows process API, especially considering the fact that going through the batch files is redundant anyway.

Still, its seems like a simple and obvious change that we can make to fix your problem now.   That way we can consider wider changes separately/later.

iscgar:
> Still, its seems like a simple and obvious change that we can make to fix your problem now. That way we can consider wider changes separately/later.

Fair. I reverted my changed and pushed a commit on top that implements response file support for `file_packager`. If this is what you had in mind, I'll squash and update the description of the PR.

iscgar:
Squashed and updated.

sbc100 (approved):
I wonder if its worth adding a test?  Maybe tricky to construct?

lgtm either way

iscgar:
> I wonder if its worth adding a test? Maybe tricky to construct?

In theory we just need to compile an empty `int main() { return 0; }` with a huge list of `--preload-file` using a response file and ensure that it doesn't fail, and that the resulting data file contains the names of the preloaded files. I'm not really sure though where to put such a test, so I would appreciate some guidance.

sbc100:
> > I wonder if its worth adding a test? Maybe tricky to construct?
> 
> In theory we just need to compile an empty `int main() { return 0; }` with a huge list of `--preload-file` using a response file and ensure that it doesn't fail, and that the resulting data file contains the names of the preloaded files. I'm not really sure though where to put such a test, so I would appreciate some guidance.

Such a test would live alongside `test_file_packager` in `test_other.py`.   You can run that test using `./test/runner other.test_file_packager`.  However I think landing this as is is also reasonable, if you would prefer to do that.

test/test_other.py
```diff
@@ -3739,6 +3740,25 @@ def clean(txt):
...
+      filenames.add(str(uuid.uuid4()))
+
+    create_file('src.c', 'int main() { return 0; }\n')
+    create_file('data.txt', '')
+
+    response_data = '\n'.join(f'--preload-file data.txt@{f}' for f in filenames)
+    self.assertGreater(len(response_data), (1 << 16))
+    create_file('data.rsp', response_data)
+
+    self.run_process([EMCC, 'src.c', '-o', 'src.js', '-sSINGLE_FILE=1', '@data.rsp'])
```
sbc100, line 3755: Is `-sSINGLE_FILE=1` needed here?

test/test_other.py
```diff
@@ -3739,6 +3740,25 @@ def clean(txt):
...
+    with open('src.js') as f:
+      data = f.read()
```
sbc100, line 3757: You can just do `data = read_file('a.out.js')`

test/test_other.py
sbc100, line 3755: Just drop `-o src.js` and assume the output is `a.out.js`.

test/test_other.py
```diff
@@ -3739,6 +3740,25 @@ def clean(txt):
 
     self.assertTextDataIdentical(clean(proc.stdout), clean(proc2.stdout))
 
+  def test_file_packager_response_file(self):
+    filenames = set()
+    while len(filenames) < 4096:
+      filenames.add(str(uuid.uuid4()))
```
sbc100, line 3746: Maybe nice if these were deterministic but not a big deal.

sbc100 (commented):
Nice!  Thanks for adding the test.  Very cool.

test/test_other.py
```diff
@@ -3739,6 +3740,25 @@ def clean(txt):
 
     self.assertTextDataIdentical(clean(proc.stdout), clean(proc2.stdout))
 
+  def test_file_packager_response_file(self):
+    filenames = set()
```
sbc100, line 3744: Just use a list here?   Keeps the order deterministic and I don't think we use any set features below?

sbc100: Oh wait, just in case they collide I guess.. seems fine in that case.

test/test_other.py
iscgar: Nope, just a force of habit, I suppose.

iscgar:
PR comments addressed.

test/test_other.py
```diff
@@ -3739,6 +3739,22 @@ def clean(txt):
 
     self.assertTextDataIdentical(clean(proc.stdout), clean(proc2.stdout))
 
+  def test_file_packager_response_file(self):
+    filenames = [f'foo.{i:032}' for i in range(4096)]
```
sbc100, line 3743: Nice!

sbc100 (approved):

sbc100:
Thanks for working on this

Issue #24083, opened 2025-04-09
EMSCRIPTEN_KEEPALIVE not working with -sWASM_ESM_INTEGRATION


brendandahl:
`emcc -sWASM_ESM_INTEGRATION -g main.c -o a.mjs`

```c
// main.c
#include <stdio.h>
#include <emscripten.h>

EMSCRIPTEN_KEEPALIVE void foo() {}

int main() {
  
}
```

a.mjs is missing the foo export:
```js
import * as unused from './a.wasm';
export { default, HEAPF32, HEAPF64, HEAP8, HEAPU8, HEAP16, HEAPU16, HEAP32, HEAPU32, HEAP64, HEAPU64 } from './a.support.mjs';
```

brendandahl:
FWIW, it was working at some point.

PR #24108, opened 2025-04-15
[wasm-esm-integration] Fix EMSCRIPETEN_KEEPALIVE

sbc100:
As part of this change we re-export all wasm symbols from the support JS file.  Going forward this also means we can support things like JS wrappers.

Fixes: #24083

### DIFF
--- a/test/core/test_esm_integration.expected.mjs
+++ b/test/core/test_esm_integration.expected.mjs
@@ -1,2 +1,4 @@
-export { __main_argc_argv as main } from './hello_world.wasm';
-export { default, err, stringToNewUTF8 } from './hello_world.support.mjs';
+// The wasm module must be imported here first before the support file
+// in order to avoid issues with circular dependencies.
+import * as unused from './hello_world.wasm';
+export { default, _foo, _main, err, stringToNewUTF8 } from './hello_world.support.mjs';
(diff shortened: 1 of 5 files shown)


tools/link.py
```diff
@@ -2116,28 +2116,13 @@ def create_worker_file(input_file, target_dir, output_file, options):
...
-    wrapper.append(f"export {{ {wasm_exports} }} from './{settings.WASM_BINARY_FILE}';")
-    # Because of the wasm ESM integration worker we need to make sure we import
-    # the wasm module here, before we import the support.js file.
-    # So, if there are no other exports we at least export the `memory`.
-  else:
-    wrapper.append('// The wasm module must be imported here first before the support file')
-    wrapper.append('// in order to avoid issues with circullr dependencies.')
-    wrapper.append(f"import * as unused from './{settings.WASM_BINARY_FILE}';")
+  wrapper.append('// The wasm module must be imported here first before the support file')
+  wrapper.append('// in order to avoid issues with circullr dependencies.')
```
brendandahl, line 2124:
```suggestion
  wrapper.append('// in order to avoid issues with circular dependencies.')
```

test/core/test_esm_integration.expected.mjs
```diff
@@ -1,2 +1,4 @@
-export { __main_argc_argv as main } from './hello_world.wasm';
-export { default, err, stringToNewUTF8 } from './hello_world.support.mjs';
+// The wasm module must be imported here first before the support file
+// in order to avoid issues with circullr dependencies.
```
brendandahl, line 2:
```suggestion
// in order to avoid issues with circular dependencies.
```

brendandahl (approved):
r+ w/ nits

Thanks for fixing!

PR #24111, opened 2025-04-15
Use `shlex.join`. NFC

sbc100:
We bumped the minimum python version to 3.8 in #23417.  I guess I overlooked this change at the time.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -88,7 +88,7 @@ EMMAKE = shared.bat_suffix(path_from_root('emmake'))
 EMCMAKE = shared.bat_suffix(path_from_root('emcmake'))
 EMCONFIGURE = shared.bat_suffix(path_from_root('emconfigure'))
 EMRUN = shared.bat_suffix(shared.path_from_root('emrun'))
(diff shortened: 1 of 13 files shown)


test/common.py
```diff
@@ -1400,7 +1400,7 @@ def build(self, filename, libraries=None, includes=None, force_c=False, emcc_arg
       output = output_basename + output_suffix
     else:
       output = shared.unsuffixed_basename(filename) + output_suffix
-    cmd = compiler + [filename, '-o', output] + self.get_emcc_args(main_file=True)
+    cmd = compiler + [str(filename), '-o', output] + self.get_emcc_args(main_file=True)
```
kripken, line 1403: Why is this change needed?

kripken (approved):
lgtm % question

test/common.py
sbc100: The official shlex.join only accepts strings and not `Path` objects.  Our version had the call to `os.fspath` in it.

kripken (approved):

PR #24116, opened 2025-04-15
Use library aliases for FS_<symbol> functions.

brendandahl:
Removes the custom code for exporting the FS_<symbol> with EXPORTED_RUNTIME_FUNCTIONS. Using library symbols makes these exports behave like all the other library exports.


src/lib/libfs.js
```diff
@@ -1935,4 +1933,17 @@ FS.staticInit();
 
   $FS_createLazyFile__deps: ['$FS'],
   $FS_createLazyFile: 'FS.createLazyFile',
-});
+
+};
+
+// Add library aliases for all the FS.<symbol> as FS_<symbol>.
```
brendandahl, line 1939: Alternatively, we could make a static list of these like FS_createLazyFile above. There are currently ~121 properties on the FS object though.

sbc100: I think we can probably come up with a shortlist.  I think just the ones that wamfs.js implements should be enough.  

I'd be OK with landing this version first if you prefer.

sbc100: Seems like we should delete the ones above that do the same thing (if we do land this version)

src/lib/libfs.js
```diff
@@ -1943,7 +1931,7 @@ for (let key in LibraryFS.$FS) {
   if (LibraryFS[alias]) {
     continue;
   }
-  LibraryFS[alias] = `FS.${key}`;
+  LibraryFS[alias] = `(...args) => FS.${key}(...args)`;
```
sbc100, line 1934: Is the forwarding function needed?  (Rather than just the string?)

src/lib/libfs.js
```diff
@@ -1915,25 +1915,13 @@ FS.staticInit();`;
...
-  $FS_unlink__deps: ['$FS'],
-  $FS_unlink: (path) => FS.unlink(path),
-
   $FS_mkdirTree__docs: `
   /**
    * @param {number=} mode Optionally, the mode to create in. Uses mkdir's
    *                       default if not set.
    */`,
   $FS_mkdirTree__deps: ['$FS'],
   $FS_mkdirTree: (path, mode) => FS.mkdirTree(path, mode),
```
sbc100, line 1924: Can we remove these two lines and just leave the `__docs`?   Or is there an assertion we would run into?

src/lib/libfs.js
brendandahl: Yeah, depending on the ordering of the output some of these functions are output before `FS` is defined.

sbc100: Hmm.. I wonder how that is possible give the `__deps` entry?  IIRC deps should always come first.

I guess its not a big deal.

brendandahl: I was wondering about that too. It's when I remove `FS_createDataFile` `other.test_exported_runtime_methods` fails.

sbc100 (approved):

PR #24117, opened 2025-04-16
Add `--help` flag to file_packager.py

sbc100:


tools/file_packager.py
```diff
@@ -21,7 +21,7 @@
 
 Usage:
 
-  file_packager TARGET [--preload A [B..]] [--embed C [D..]] [--exclude E [F..]]] [--js-output=OUTPUT.js] [--no-force] [--use-preload-cache] [--indexedDB-name=EM_PRELOAD_CACHE] [--separate-metadata] [--lz4] [--use-preload-plugins] [--no-node]
+  file_packager TARGET [--preload A [B..]] [--embed C [D..]] [--exclude E] [--js-output=OUTPUT.js] [--no-force] [--use-preload-cache] [--indexedDB-name=EM_PRELOAD_CACHE] [--separate-metadata] [--lz4] [--use-preload-plugins] [--no-node]
```
kripken, line 24: Why change `--exclude` here? Doesn't it work like `--embed` right before it?

sbc100: Hmm I have no idea :)

It does look like there was a stray bracket so I removed that.

I also made the long help message consistent.

PR #24131, opened 2025-04-16
Warn about unknown entries in INCOMING_MODULE_JS_API

sbc100:


tools/link.py
```diff
@@ -971,6 +971,10 @@ def phase_linker_setup(options, linker_args):  # noqa: C901, PLR0912, PLR0915
     else:
       default_setting('INCOMING_MODULE_JS_API', [])
 
+  for prop in settings.INCOMING_MODULE_JS_API:
+    if prop not in settings.ALL_INCOMING_MODULE_JS_API:
+      exit_with_error(f'invalid entry in INCOMING_MODULE_JS_API: {prop}')
```
kripken, line 976: Should this be a warning perhaps, and only an error when all warnings are errors?

sbc100: Why?

kripken: One thing that comes to mind is if the user builds with different versions, and an older version lacks an API, or a newer one deprecated it. It might be nice to let them disable the warning in that case, rather than work around it in other ways. We do have `diagnostics.warning('deprecated'` for that, though I admit it is not exactly the same thing.

I don't feel strongly here, though.

kripken: Anyhow, either way is good for me.

kripken (approved):

tools/link.py
sbc100: Done

kripken (approved):
Thanks, this does feel a little less risky.

PR #24132, opened 2025-04-16
Move jstoi_s to liblegacy.js

sbc100:
This function was added alongside `jstoi_q` in #10457, but its documentation is wrong. `Number("123abc")` does not throw, it returns NaN.

This function never had any internal usage within emscripten.


src/lib/liblegacy.js
```diff
@@ -116,6 +116,11 @@ legacyFuncs = {
   // Legacy names for runtime `out`/`err` symbols.
   $print: 'out',
   $printErr: 'err',
+
+  // Converts a JS string to an integer base-10, with signaling error
+  // handling (throws a JS exception on error). E.g. jstoi_s("123abc")
+  // throws an exception.
```
kripken, line 122:
```suggestion
  // Converts a JS string to an integer base-10. Despite _s, which
  // suggests signaling error handling, this returns NaN on error.
  // (This was a mistake in the original implementation, and kept
  // to avoid breakage.)
```
Might as well fix the docs too?

kripken (approved):

src/lib/liblegacy.js
sbc100: Oops I tried to commit your suggestion (normally I prefer to make the changes locally so I'm not used that UI.. I must have done something wrong).

kripken: Hmm, maybe it commited ok but your next force-push overwrote it... force-push doesn't warn on that iirc.

kripken (approved):

PR #24134, opened 2025-04-17
Avoid redundantly specifying opt level in ports

sbc100:
The default is -O2


kripken (approved):
I am slightly worried there was a reason for the `-O3`, and we just forgot to document it, but can't remember. Maybe if we had benchmarks for them that would be worth checking, but I don't see any, so lgtm.

sbc100:
I'll revert the -O3 ones.

PR #24144, opened 2025-04-18
Under report download size when compression involved

vjyanand:
When a server uses compression (e.g., Brotli (`br`) or Gzip (`gzip`)), the Content-Length header in the HTTP response reflects the compressed size of the data, not the uncompressed size. However, the `fetchRemotePackage` function tracks the uncompressed size of the downloaded data (via the `value.length` of each chunk after decompression), leading to a discrepancy where the reported `total` downloaded (uncompressed) can exceed the Content-Length (compressed). 

This causes inaccurate progress reporting, such as `totalLoaded` appearing larger than `total`. Unfortunately, the Fetch API doesn’t directly expose the compressed stream’s size for each chunk, as decompression happens transparently. 

This fix ensures that the reported loaded and total values align with the compressed size (from Content-Length) rather than the decompressed size

### DIFF
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -995,11 +995,24 @@ def generate_js(data_target, data_files, metadata):
             const headers = response.headers;
             const total = Number(headers.get('Content-Length') ?? packageSize);
             let loaded = 0;
+            // Track both network size and actual decompressed size
+            let decompressedLoaded = 0;
+            const contentEncoding = headers.get('Content-Encoding');
+            const isCompressed = contentEncoding === 'gzip' || contentEncoding === 'br';
 
             const handleChunk = ({done, value}) => {
               if (!done) {
                 chunks.push(value);
-                loaded += value.length;
+                decompressedLoaded += value.length;
+                if (isCompressed) {
+                  // Estimate progress proportionally, assuming 3:1 compression ration
+                  loaded = Math.min(total, 
+                  total * (decompressedLoaded / (packageSize || total * 3)));
+                } else {
+                  // When not compressed, they're the same
+                  loaded = decompressedLoaded;
+                }
+
                 Module['dataFileDownloads'][packageName] = {loaded, total};
 
                 let totalLoaded = 0;


tools/file_packager.py
```diff
@@ -995,11 +995,24 @@ def generate_js(data_target, data_files, metadata):
...
 
             const handleChunk = ({done, value}) => {
               if (!done) {
                 chunks.push(value);
-                loaded += value.length;
+                decompressedLoaded += value.length;
+                if (isCompressed) {
+                  // Estimate progress proportionally, assuming 3:1 compression ration
+                  loaded = Math.min(total, 
+                  total * (decompressedLoaded / (packageSize || total * 3)));
```
sbc100, line 1010: This seems a little too much like guesswork.  Is there really no better way?

vjyanand: @sbc100 how about now?

tools/file_packager.py
```diff
@@ -993,7 +993,9 @@ def generate_js(data_target, data_files, metadata):
 
             const chunks = [];
             const headers = response.headers;
-            const total = Number(headers.get('Content-Length') ?? packageSize);
+            const contentEncoding = headers.get('Content-Encoding');
+            const isCompressed = contentEncoding === 'gzip' || contentEncoding === 'br'; 
+            const total = Number(isCompressed ? packageSize ?? headers.get('Content-Length') : headers.get('Content-Length') ?? packageSize);
```
sbc100, line 998: The combination of nullish coalescing and ternary operator makes this pretty hard to read.

How about this instead (assuming this is what you are trying to express.

```
const total = packageSize;
// Ignore Content-Length when compression is used since that reports the compressed size
if (!isCompressed && headers.get('Content-Length')) {
  total = Number(headers.get('Content-Length'));
}
``

Comment (vjyanand) [reply tools/file_packager.py:998]: @sbc100 Yes close, may be below when packageSize is null

```
const total = packageSize;
if (!total || (!isCompressed && headers.get('Content-Length'))) {
  total = Number(headers.get('Content-Length'));
}

```

sbc100: Can you re-write in that form ?

vjyanand: Done, Thanks

sbc100:
Can you update the PR description and title now that we have changed the PR content?

sbc100:
Do you know why we cannot just use `packageSize` in all cases?  Are there some cases where `packageSize` is not known?

vjyanand:
> Do you know why we cannot just use `packageSize` in all cases? Are there some cases where `packageSize` is not known?

just my guess, `packageSize` is calculated while doing packaging, one can still change the file with different size on the server without doing packaging. in that case, content-length will be correct value,

sbc100:
It looks like someone else just ran into this in #24529.   Any chance you could update this PR and we can land it?

PR #24146, opened 2025-04-18
Mark certain JS library functions as `__noleakcheck`

sbc100:
The functions all allocate on first use and cache the results, never freeing.  Marking them as `__noleakcheck` avoids lsan false positives.


src/lib/libegl.js
```diff
@@ -507,6 +507,8 @@ var LibraryEGL = {
   eglGetError: () => EGL.errorCode,
 
   // EGLAPI const char * EGLAPIENTRY eglQueryString(EGLDisplay dpy, EGLint name);
+  // The allocations in glGetString are cached and never freed.
```
kripken, line 510:
```suggestion
  // The allocated strings are cached and never freed.
```
(or fix the function name)

src/lib/libhtml5.js
```diff
@@ -249,6 +249,7 @@ var LibraryHTML5 = {
     },
   },
 
+  $registerKeyEventCallback__noleakcheck: true,
```
kripken, line 252: Is this intentional in these functions? Are they not allocating events the user must deallocate?

tools/ports/sdl2.py
```diff
@@ -70,7 +70,7 @@ def create(final):
     srcs += ['thread/%s/%s' % (thread_backend, s) for s in thread_srcs]
 
     srcs = [os.path.join(src_dir, 'src', s) for s in srcs]
-    flags = ['-sUSE_SDL=0']
+    flags = ['-sUSE_SDL=0', '-fsanitize=address']
```
kripken, line 73: local debugging?

src/lib/libhtml5.js
sbc100: They are all allocating the event struct on first use an never free'ing it.

kripken (approved):

PR #24138, opened 2025-04-17
Ignore EXPORT_NAME in EXPORT_ES6 mode. NFC

sbc100:
The internal name of the factory function is not exposed to the outside world in `EXPORT_ES6` mode, so there is no point in setting this setting.

PR #24149, opened 2025-04-18
Revert "Ignore EXPORT_NAME in EXPORT_ES6 mode. NFC"

sbc100:
Reverts emscripten-core/emscripten#24138

This change broke the binaryen JS build.


kripken (approved):

sbc100:
Actually I have a fix we could land instead : https://github.com/emscripten-core/emscripten/pull/24150

sbc100:
Actually binaryen looks like it depends on being able to use EXPORT_NAME in ES6 mode.   So I guess we should revert, at least for now.

kripken:
Do you mean we need to remove that flag from binaryen? Or something else?

The error binaryen gets isn't the one the PR has (which is a good error for users), so I assumed it was more than just removing the flag?

sbc100:
Binaryen is wrapping the generated ES6 module in other JS code and assuming that a factory function called `EXPORT_NAME` (AKA Binaryen) exists.   I suppose this is reasonable.. although I was kind of hoping it was an internal detail of the ES6 module.

kripken:
Hmm, yeah, there is some complex history with how we generate the JS there... related to common.js for npm iirc, but I don't think I ever knew all the details. Might be worth cleaning up if there's a better way.

PR #24150, opened 2025-04-18
Fix crash on use of EXPORT_NAME with EXPORT_ES6

sbc100:
Followup to #24138 which introduced the issue.


kripken (approved):
lgtm depending on whether you decide to land the reversion.

kripken:
I confirmed this gets us past that error on Binaryen.

sbc100:
> I confirmed this gets us past that error on Binaryen.

Only the first build error sadly.   It fails when it tries to run tests.

sbc100:
Closing this the original change was reverted

PR #21345, opened 2024-02-14
Add external ports + options support to embuilder

ypujante:
This is a follow up of PR #21316 as [discussed](https://github.com/emscripten-core/emscripten/pull/21316#issuecomment-1942188916)

Now `embuilder` supports external ports with the simple syntax:

```sh
# building an external port
embuilder build test/other/ports/simple.py
# building with options
embuilder build sdl2_image:formats=png,jpg
```

Prior to this PR, it was possible to build `sdl2_image` or `sdl2_image-png` or `sdl2_image-jpg` but it was not possible to build one that contains both. This PR allow to do this!

Part of this PR is that I have changed the `handle_options` api (ports) to take an error handler. I think it is a lot cleaner than having each port call `utils.exit_with_error` which is clearly an internal API. That way it is possible to change the `error_handler` without changing the ports code.

PR #24158, opened 2025-04-21
Support `formats` option in sdl2 mixer

ryanking13:
This PR adds `handle_options` function in the SDL2_mixer port so one can build SDL2_mixer with specific formats with embuilder, such as

```
embuilder build sdl2_mixer:formats:mid,ogg,mp3
```

I found this syntax was added in https://github.com/emscripten-core/emscripten/pull/21345, but it was only applied to SDL2_image.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -2470,7 +2470,7 @@ F1 -> ''
   def test_sdl2_mixer_wav(self):
     self.emcc(test_file('browser/test_sdl2_mixer_wav.c'), ['-sUSE_SDL_MIXER=2'], output_filename='a.out.js')
     self.emcc(test_file('browser/test_sdl2_mixer_wav.c'), ['--use-port=sdl2_mixer'], output_filename='a.out.js')
-    self.emcc(test_file('browser/test_sdl2_mixer_wav.c'), ['--use-port=sdl2_mixer:formats=mp3'], output_filename='a.out.js')
+    self.emcc(test_file('browser/test_sdl2_mixer_wav.c'), ['--use-port=sdl2_mixer:formats=ogg'], output_filename='a.out.js')
 
   def test_sdl2_linkable(self):
     # Ensure that SDL2 can be built with LINKABLE.  This implies there are no undefined


tools/ports/sdl2_mixer.py
```diff
@@ -16,14 +17,28 @@
...
 
+OPTIONS = {
+  'formats': 'A comma separated list of formats (ex: --use-port=sdl2_mixer:formats=ogg,mp3)'
+}
+
+SUPPORTED_FORMATS = {'ogg', 'mp3', 'mod', 'mid'}
+
+# user options (from --use-port)
+opts: Dict[str, Set] = {
+  'formats': set()
```
sbc100, line 28: Maybe just make this a list, since you are going to uniq the list anyway?

tools/ports/sdl2_mixer.py
```diff
@@ -16,14 +17,28 @@
...
+}
+
 
 def needed(settings):
   return settings.USE_SDL_MIXER == 2
 
 
+def get_formats(settings):
+  return set(settings.SDL2_MIXER_FORMATS).union(opts['formats'])
```
sbc100, line 37: I think you can just write this as `opts['formats'].union(settings.SDL2_MIXER_FORMATS)` (i.e. `union` can take a list)

ryanking13: Thanks. I changed it to `opts['formats'].union(settings.SDL2_MIXER_FORMATS)` (and update the similar code in sdl2_image as well).

ryanking13:
I added some tests, but it seems like FROZEN_CACHE is not happy with sdl2_mixer variants, hmm...

.circleci/config.yml
```diff
@@ -900,6 +900,10 @@ jobs:
           # some native-dependent tests fail because of the lack of native
           # headers on emsdk-bundled clang
           test_targets: "other skip:other.test_native_link_error_message"
+          # There are tests that using libraries
+          # that are not included by "./embuilder build ALL".  For example the
+          # different variant of libSDL2_mixer is used by test_sdl2_mixer_wav
+          frozen_cache: false
```
sbc100, line 906: Better not to do this.    Instead you can add the variant you are testing the variants list in sdl2_mixer.py.

sbc100 (approved):

.circleci/config.yml
ryanking13: Thanks I removed this and used `ogg` format for a test case.

test/test_other.py
```diff
@@ -2470,6 +2470,7 @@ def test_sdl_get_key_name(self):
   def test_sdl2_mixer_wav(self):
     self.emcc(test_file('browser/test_sdl2_mixer_wav.c'), ['-sUSE_SDL_MIXER=2'], output_filename='a.out.js')
     self.emcc(test_file('browser/test_sdl2_mixer_wav.c'), ['--use-port=sdl2_mixer'], output_filename='a.out.js')
+    self.emcc(test_file('browser/test_sdl2_mixer_wav.c'), ['--use-port=sdl2_mixer:formats=ogg'], output_filename='a.out.js')
```
sbc100, line 2473: How does this test (which plays a wav file) work if only ogg is selected?  Is this because `wav` is simply always available maybe?

ryanking13: Yeah, I think so. wav is already available:

https://github.com/emscripten-core/emscripten/blob/2328a81304238c2d69ce8d1662e3f3505fe88b85/tools/ports/sdl2_mixer.py#L44-L47

PR #24163, opened 2025-04-21
[WASM_WORKERS] Inline `.ww.js` file into the main output `.js` file

sbc100:
Similar to what we did for pthreads in #21701.

This is win for both code size and complexity.  Importantly it avoid the issue how to package and deploy the extra `.ww.js` file.

This also completely removes the need for the `-sWASM_WORKERS=2` mode.


RReverser:
Very happy to see this happening, thanks for working on it!

I think conceptually .aw.js are also very similar, I gave a stab to trying to inline them before, but the amount of changes was getting out of hand (now I see why from this diff, it does require quite a bit of work). Do you think they could be also covered in or after this PR?

sbc100:
> Very happy to see this happening, thanks for working on it!
> 
> I think conceptually .aw.js are also very similar, I gave a stab to trying to inline them before, but the amount of changes was getting out of hand (now I see why from this diff, it does require quite a bit of work). Do you think they could be also covered in or after this PR?

I think may be possible, but would likely be a followup.

src/audio_worklet.js
```diff
@@ -144,6 +148,7 @@ class BootstrapMessages extends AudioWorkletProcessor {
     // Wasm Module and Memory objects.  After this we are ready to load in the
     // main application JS script, which the main thread will addModule()
     // to this scope.
+    globalThis.wasmWorkerArgs = arg['processorOptions'];
     globalThis.Module = arg['processorOptions'];
```
kripken, line 152: These have the same value, is one of them not enough?

src/lib/libcore.js
```diff
@@ -2043,7 +2043,7 @@ addToLibrary({
   $runtimeKeepalivePush: () => {
     runtimeKeepaliveCounter += 1;
 #if RUNTIME_DEBUG
-    dbg(`runtimeKeepalivePush -> counter=${runtimeKeepaliveCounter}`);
+    //dbg(`runtimeKeepalivePush -> counter=${runtimeKeepaliveCounter}`);
```
kripken: (local debugging?)

src/lib/libwasm_worker.js
```diff
@@ -34,18 +25,28 @@
...
 #else
-    locateFile('${WASM_WORKER_FILE}')
+    locateFile('${TARGET_JS_NAME}')
 #endif
 `;
+  const wasmWorkerOptions = `{
+#if ENVIRONMENT_MAY_BE_NODE
+        // This is the way that we signal to the node worker that it is hosting
+        // a wasm worker.
+        'workerData': 'em-ww',
```
kripken, line 42: indentation looks too deep here?

src/wasm_worker.js
```diff
@@ -28,52 +27,44 @@ if (ENVIRONMENT_IS_NODE) {
...
+#if MINIMAL_RUNTIME
+  Module['wasm'] = d['wasm'];
+  loadModule()
+#else
+  wasmModuleReceived(d['wasm']);
 #endif
   // Drop now unneeded references to from the Module object in this Worker,
   // these are not needed anymore.
-  d.wasm = d.mem = d.js = 0;
+  d['wasm'] = d['mem'] = 0;
```
kripken, line 67: Are these quotings bugfixes, or necessary only now for some reason?

test/code_size/hello_wasm_worker_wasm.js
```diff
@@ -1,50 +1,77 @@
...
+        y = a.g;
+        z = a.i;
+        m = a.h;
+        c ? (a = f, b.$ww = a.$ww, z(a.sb, a.sz), removeEventListener("message", p), k = k.forEach(n), 
+        addEventListener("message", n)) : a.f();
+        c || y();
+    }));
+}
+
+c || h();
```
kripken, line 77: This looks larger than before, is that because of the inlined code?

sbc100: They was not needed before since `wasm_worker.js` was generated as a separate file that was not run through closure.

Now this code is part of the main output which closure sees so quotes are needed both on the sending and receiving size.

Once we do the followup of also inlining the AUDIO_WORKLET file I think we can completly remove these quotes and let closure minify these names (like we did in the pthread case) which will further reduce code size.

src/wasm_worker.js
sbc100: They was not needed before since `wasm_worker.js` was generated as a separate file that was not run through closure.

Now this code is part of the main output which closure sees so quotes are needed both on the sending and receiving size.

Once we do the followup of also inlining the AUDIO_WORKLET file I think we can completly remove these quotes and let closure minify these names (like we did in the pthread case) which will further reduce code size.

src/audio_worklet.js
sbc100: This is the simplest way I could figure to make it work for now both in the audio worklet case and in the wasm worker case.

Audio worklet are just like wasm workers but they have this laternate creation path that currently lives in the `aw.js` file.   As a followup to this PR I hope to remove the aw.js file too, so this code here should no longer be needed once that lands.

src/wasm_worker.js
```diff
@@ -28,52 +27,44 @@ if (ENVIRONMENT_IS_NODE) {
...
     self: global,
-    require,
-#if !EXPORT_ES6
-    // `vm.runInThisContext` global scope lacks `__filename` and `__dirname`
-    __filename,
-    __dirname,
-#endif
-    Worker: nodeWorkerThreads.Worker,
-    importScripts: (f) => vm.runInThisContext(fs.readFileSync(f, 'utf8'), {filename: f}),
     postMessage: (msg) => parentPort.postMessage(msg),
```
RReverser, line 32: I think these polyfills could be shared with runtime_pthread too: https://github.com/emscripten-core/emscripten/blob/081ac97000e8c609db809df50672b8d5a04e545e/src/runtime_pthread.js#L29-L35 + https://github.com/emscripten-core/emscripten/blob/f163dc219468cf98873a054319be19aacfec5a79/src/wasm_worker.js#L17

sbc100: True. Refactored shared code.

RReverser (approved):
I think there is further cleanup that could be made here (eg replacing `onmessage =` with `addEventListener` / `removeEventListener` to merge more polyfills together) but those can be left for future PRs.

LGTM % CI fixes.

RReverser:
@juj The worker is the main JS file itself, so the origin and CSP properties should remain the same, since you needed to load main JS before this PR as well (without it nothing works).

kripken (approved):

sbc100:
> What is the origin that the Worker gets loaded from after this change? Is it in the same origin? Also, does this retain csp-eval safety?

I reverted the part of this change related to how the main js file is found in MINIMAL_RUNTIME mode.   Now we continue to use `Module['js']` which I think is important for you use case?  Is the right @juj?

cwoffenden:
> @cwoffenden I think you might be happy to see this change.

I'll try to take a look next week, I've had little time for this since the last PR a few months back.

PR #24164, opened 2025-04-21
Simplify AUDIO_WORKLET checks. NFC

sbc100:
I assume the idea here was that there would be an AUDIO_WORKLET=2 mode but that was never added.

### DIFF
--- a/src/settings_internal.js
+++ b/src/settings_internal.js
@@ -137,9 +137,6 @@ var WASM_BINARY_FILE = '';
 // name of the file containing the Wasm Worker *.ww.js, if relevant
 var WASM_WORKER_FILE = '';
 
-// name of the file containing the Audio Worklet *.aw.js, if relevant
-var AUDIO_WORKLET_FILE = '';
-
 // Base URL the source mapfile, if relevant
 var SOURCE_MAP_BASE = '';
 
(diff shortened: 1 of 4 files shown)


sbc100:
@cwoffenden

src/settings_internal.js
```diff
@@ -137,9 +137,6 @@ var WASM_BINARY_FILE = '';
 // name of the file containing the Wasm Worker *.ww.js, if relevant
 var WASM_WORKER_FILE = '';
 
-// name of the file containing the Audio Worklet *.aw.js, if relevant
-var AUDIO_WORKLET_FILE = '';
```
kripken, line 141: Is this a separate change?

sbc100: No, its just an internal setting that is no longer needed after this change.

cwoffenden:
> I assume the idea here was that there would be an AUDIO_WORKLET=2 mode but that was never added.

I did use`AUDIO_WORKLET=2` for the updated API I had in development, but I think it better to run the two side-by-side marking parts with `__attribute__((deprecated))`.

Removing the need for a custom `AUDIO_WORKLET_FILE ` makes sense.

juj:
I recall I would have originally intended `AUDIO_WORKLET=2` to be used for an embedded mode where the file `a.aw.js` would be emdedded into the main JS (like with Wasm Workers). It is a tradeoff between CSP-eval policy safety to choose whether to embed the file, or have it separate.

But this change LGTM, I presume we can still add a `AUDIO_WORKLET=2` mode later if we wanted to.

sbc100:
Can someone lgtm?

kripken (approved):

PR #24180, opened 2025-04-23
Browser logging by default

sbc100:


sbc100:
Assuming this works do we think its good idea?  I wonder how much it would slow down tests with a lof stdio?

Since browser tests are already run iserially it think it should actually make tests failures much more easy to debug.

If we choose not to do this I guess we could instead have a mechanism for enabling this on a per-test (or per-test-run) basis?

sbc100:
Looks like we raced to comment :)

kripken:
Heh, yeah, great minds and all that...

The debuggability is really nice, so if you can't measure a regression here, it's probably worthwhile?

PR #24187, opened 2025-04-25
Fix issue where `dest` would be called while undef

### LINKED ISSUE #23046: library_sockfs.js can access properties of undefined `dest` (logic flaw)
> [!NOTE]
> This issue was unearthed by us due to what seems an issue on our side. My colleague @Faless explains it on the [Godot Dev Chat](https://chat.godotengine.org/channel/web?msg=oMN3SjGjhpyzZqbw6). Though, the issue underlined is still a logic flaw.

Regression introduced by #22630

https://github.com/emscripten-core/emscripten/blob/06cebfca51f908d3bc6897113330e45a77010db2/src/library_sockfs.js#L632-L647

If `dest == null` and `sock.type !== {{{ cDefs.SOCK_DGRAM }}}` (see code above), then the following error occurs:

Capture d’écran, le 2024-12-02 à 11 27 03

adamscott:
One-char fix!

Fixes #23046.


src/library_sockfs.js
```diff
@@ -642,7 +642,7 @@ addToLibrary({
 #if SOCKET_DEBUG
           dbg(`websocket: queuing (${length} bytes): ${new Uint8Array(data)}`);
 #endif
-          dest.msg_send_queue.push(data);
+          dest?.msg_send_queue.push(data);
           return length;
```
sbc100, line 646: I'm not sure about this fix because it contradicts the comment maybe? "queue the message to be sent upon connect, and lie, saying the data was sent now.".  Here we are lying but not queueing.

Do you have a repro case for this?  Perhaps we can add a test case to test_sockets.py?

adamscott: Yeah, you're right. I'm gonna try to add a test case.

adamscott:
@sbc100 I completely rewrote my PR from the ground up. I added a test, changed the fix, and I updated the PR description.

adamscott:
@sbc100 ci/circleci seems to break because `emcc: error: LLVM version for clang executable "/root/emsdk/upstream/bin/clang" appears incorrect (seeing "21.0", expected "20") [-Wversion-check] [-Werror]`

src/library_sockfs.js
```diff
@@ -634,7 +634,8 @@ addToLibrary({
         // connect, and lie, saying the data was sent now.
         if (!dest || dest.socket.readyState !== dest.socket.OPEN) {
           // if we're not connected, open a new connection
-          if (sock.type === {{{ cDefs.SOCK_DGRAM }}}) {
+          if (!dest || sock.type === {{{ cDefs.SOCK_DGRAM }}}) {
+            sock.type = {{{ cDefs.SOCK_DGRAM }}};
```
sbc100, line 638: It seem lik you could get here with `!dest` and with `sock.type = {{{ cDefs.SOCK_STREAM }}}`.. in which case this line would change the type of the socket, no?

adamscott: Yes, but the check was specific for `sock.type === {{{ cDefs.SOCK_DGRAM }}}` only, that's why I forced it. Maybe it's ok to keep the socket type, then.

test/sockets/test_sockets_sendmsg_to_invalid_dest.c
```diff
@@ -0,0 +1,67 @@
+// Copyright 2013 The Emscripten Authors.  All rights reserved.
```
sbc100, line 1: 2025

test/sockets/test_sockets_sendmsg_to_invalid_dest.c
```diff
@@ -0,0 +1,67 @@
...
+#include <emscripten.h>
+#include <emscripten/eventloop.h>
+#include <ifaddrs.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+int sock;
+
+EM_BOOL wait_for_recv(double d, void* u) {
```
sbc100, line 16: We use bool/true/false these days.

test/sockets/test_sockets_sendmsg_to_invalid_dest.c
adamscott: Sorry! Forgot to update that.

sbc100:
In the PR description I assume you meant to write `sendmsg` and not `setmsg`?

sbc100:
I wonder if the problem can instead be fixed at a higher level.  It seems that `getifaddrs` in musl is using `__rtnetlink_enumerate` which uses `PF_NETLINK` protocol family.. which emscripten does not support.

Do i'm pretty sure `getifaddrs` does not work at all right?  I don't how it fails?   Presumably this crash is just the tip if the iceberg.

sbc100:
I think I found a better solution which is simply not allow the AF_NETLINK socket to be created: https://github.com/emscripten-core/emscripten/pull/24364

PR #24190, opened 2025-04-25
[AUDIO_WORKLET] Inline `.aw.js` file into the main output `.js` file

sbc100:
This is similar to what we already did for pthreads (#24163) and normal wasm workers (#24163).  By integrating this file we can actually share more startup code with normal wasm workers.

### DIFF
--- a/src/audio_worklet.js
+++ b/src/audio_worklet.js
@@ -12,24 +12,16 @@
 // the node constructor's "processorOptions" field, we can share the necessary
 // bootstrap information from the main thread to the AudioWorkletGlobalScope.
 
+if (ENVIRONMENT_IS_AUDIO_WORKLET) {
+
 function createWasmAudioWorkletProcessor(audioParams) {
   class WasmAudioWorkletProcessor extends AudioWorkletProcessor {
     constructor(args) {
       super();
 
-      // Copy needed stack allocation functions from the Module object
-      // to global scope, these will be accessed in hot paths, so maybe
-      // they'll be a bit faster to access directly, rather than referencing
-      // them as properties of the Module object.
-      globalThis.stackAlloc = Module['stackAlloc'];
-      globalThis.stackSave = Module['stackSave'];
-      globalThis.stackRestore = Module['stackRestore'];
-      globalThis.HEAPU32 = Module['HEAPU32'];
-      globalThis.HEAPF32 = Module['HEAPF32'];
-
       // Capture the Wasm function callback to invoke.
       let opts = args.processorOptions;
-      this.callbackFunction = Module['wasmTable'].get(opts['cb']);
+      this.callbackFunction = wasmTable.get(opts['cb']);
       this.userData = opts['ud'];
       // Then the samples per channel to process, fixed for the lifetime of the
       // context that created this processor. Note for when moving to Web Audio
@@ -43,6 +35,9 @@ function createWasmAudioWorkletProcessor(audioParams) {
       return audioParams;
     }
 
+    /**
+     * @param {Object} parameters
+     */
     process(inputList, outputList, parameters) {
       // Marshal all inputs and parameters to the Wasm memory on the thread stack,
       // then perform the wasm audio worklet call,
@@ -99,6 +94,7 @@ function createWasmAudioWorkletProcessor(audioParams) {
       paramsPtr = dataPtr;
       k = paramsPtr >> 2;
       dataPtr += numParams * {{{ C_STRUCTS.AudioParamFrame.__size__ }}};
+
       for (i = 0; paramArray = parameters[i++];) {
         // Write the AudioParamFrame struct instance
         HEAPU32[k + {{{ C_STRUCTS.AudioParamFrame.length / 4 }}}] = paramArray.length;
@@ -134,28 +130,15 @@ function createWasmAudioWorkletProcessor(audioParams) {
   return WasmAudioWorkletProcessor;
 }
 
+var messagePort;
+
 // Specify a worklet processor that will be used to receive messages to this
 // AudioWorkletGlobalScope.  We never connect this initial AudioWorkletProcessor
 // to the audio graph to do any audio processing.
 class BootstrapMessages extends AudioWorkletProcessor {
   constructor(arg) {
     super();
-    // Audio worklets need to show up as wasm workers.  This is the way we signal
-    // that.
-    globalThis.name = 'em-ww';
-    // Initialize the global Emscripten Module object that contains e.g. the
-    // Wasm Module and Memory objects.  After this we are ready to load in the
-    // main application JS script, which the main thread will addModule()
-    // to this scope.
-    globalThis.Module = arg['processorOptions'];
-#if !MINIMAL_RUNTIME
-    // Default runtime relies on an injected instantiateWasm() function to
-    // initialize the Wasm Module.
-    globalThis.Module['instantiateWasm'] = (info, receiveInstance) => {
-      var instance = new WebAssembly.Instance(Module['wasm'], info);
-      return receiveInstance(instance, Module['wasm']);
-    };
-#endif
+    startWasmWorker(arg['processorOptions'])
 #if WEBAUDIO_DEBUG
     console.log('AudioWorklet global scope looks like this:');
     console.dir(globalThis);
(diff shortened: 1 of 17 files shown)


cwoffenden:
This is awesome, thanks for taking the time to do this. I'll try to look at it next week.

src/audio_worklet.js
```diff
@@ -164,6 +145,7 @@ class BootstrapMessages extends AudioWorkletProcessor {
     // scope to create the real AudioWorkletProcessors that call out to Wasm to
     // do audio processing.
     let p = globalThis['messagePort'] = this.port;
+    /** @suppress {checkTypes} */
     p.onmessage = async (msg) => {
       let d = msg.data;
       if (d['_wpn']) {
```
cwoffenden, line 151: Now that this is part of the main JS, we can eventually go with expressive names that get minimised, right?

sbc100: I think so yes.

src/audio_worklet.js
```diff
@@ -99,6 +94,7 @@ function createWasmAudioWorkletProcessor(audioParams) {
       paramsPtr = dataPtr;
       k = paramsPtr >> 2;
       dataPtr += numParams * {{{ C_STRUCTS.AudioParamFrame.__size__ }}};
+
```
cwoffenden, line 97: Nit, but not of the other sections have this superfluous space.

test/webaudio/audioworklet.c
```diff
@@ -110,6 +111,7 @@ void AudioWorkletProcessorCreated(EMSCRIPTEN_WEBAUDIO_T audioContext, bool succe
 // Processors.
 void WebAudioWorkletThreadInitialized(EMSCRIPTEN_WEBAUDIO_T audioContext, bool success, void *userData) {
   assert(success && "Stopped in WebAudioWorkletThreadInitialized");
+  printf("audio worklet initialzied\n");
```
cwoffenden, line 114: Spelling of 'initialzied' (and you've no idea the finger pain I went through to write that with a zed...).

src/audio_worklet.js
```diff
@@ -164,23 +145,12 @@ class BootstrapMessages extends AudioWorkletProcessor {
     // scope to create the real AudioWorkletProcessors that call out to Wasm to
     // do audio processing.
     let p = globalThis['messagePort'] = this.port;
```
RReverser, line 147: Now that it's all in the same file, we should be able to replace this `globalThis[...]` property here and its usages in libwebaudio.js with a simple variable, which should also compress better.

src/parseTools.mjs
```diff
@@ -79,6 +79,13 @@ export function preprocess(filename) {
       .replace(/\bimport\.meta\b/g, 'EMSCRIPTEN$IMPORT$META')
       .replace(/\bawait import\b/g, 'EMSCRIPTEN$AWAIT$IMPORT');
   }
+  if (USE_CLOSURE_COMPILER && AUDIO_WORKLET) {
+    // Closure compiler doesn't currently handle static getters.
+    text = text.replace(
```
RReverser, line 84: As far as I can tell, that property is not used anywhere, so a simpler fix might be to just remove it (which also gives a tiny code size win).

src/wasm_worker.js
```diff
@@ -1,10 +1,38 @@
...
+  wasmMemory = props['mem'];
+  updateMemoryViews();
+#if MINIMAL_RUNTIME
+  loadModule()
+#else
+  wasmModuleReceived(props['wasm']);
+#endif
+  // Drop now unneeded references to from the Module object in this Worker,
+  // these are not needed anymore.
+  props['wasm'] = props['mem'] = 0;
```
RReverser, line 26: You mentioned in the previous PR:

> Once we do the followup of also inlining the AUDIO_WORKLET file I think we can completly remove these quotes and let closure minify these names (like we did in the pthread case) which will further reduce code size.

Should we do this now or left as a follow-up?

tools/link.py
```diff
@@ -2098,34 +2089,20 @@ def phase_source_transforms(options):
...
   src = read_file(js_file)
-  write_file(js_file, src
-             .replace('EMSCRIPTEN$IMPORT$META', 'import.meta')
-             .replace('EMSCRIPTEN$AWAIT$IMPORT', 'await import')
-             .replace('EMSCRIPTEN$AWAIT(createWasm())', 'await createWasm()')
-             .replace('EMSCRIPTEN$AWAIT(', 'await ('))
+  if settings.MODULARIZE:
+    src = (src.replace('EMSCRIPTEN$IMPORT$META', 'import.meta')
+              .replace('EMSCRIPTEN$AWAIT$IMPORT', 'await import')
+              .replace('EMSCRIPTEN$AWAIT(createWasm())', 'await createWasm()')
```
RReverser, line 2098: Wait not related to this PR as I see it was added earlier, but why are these necessary, does Closure not support `await` at all? IIRC, the replacements above were necessary specifically for unsupported `import.meta` and dynamic `import()`, but in other contexts it should work as-is.

src/audio_worklet.js
```diff
@@ -203,7 +173,7 @@ class BootstrapMessages extends AudioWorkletProcessor {
 #else
         var ptr = d['_wsc'];
```
RReverser, line 174: Nit (unrelated to this PR) but should this just use the `to64` helper?

cwoffenden: I wrote a whole wasm64 conversion waiting to be simplified, and it wasn't always a simple as to which of the helpers were needed:

https://github.com/cwoffenden/emscripten/blob/6b3f01eb7327c2ee7f8056a4849bdddf536fb5c1/src/audio_worklet.js#L278

src/audio_worklet.js
```diff
@@ -219,5 +189,11 @@ class BootstrapMessages extends AudioWorkletProcessor {
   }
 };
 
+
 // Register the dummy processor that will just receive messages.
+#if WEBAUDIO_DEBUG
+console.log('new audio worklet, calling registerProcessor');
+#endif
 registerProcessor('message', BootstrapMessages);
```
RReverser, line 197: Something I changed in my earlier attempt to inline aw.js - let's rename it to something like `em-boostrap`. This is a globally registered name, and may conflict with user-defined ones if any.

src/audio_worklet.js
sbc100: Resolving as unrelated.

src/audio_worklet.js
sbc100: Lets do that kind of orthogonal improvement as a followup.

tools/link.py
sbc100: The `await createWasm()` is inside the closure / factory function when running in MODULARIZE mode.

However because closure is only run on the inner code it think its looking at a top-level await and it errors out.  At least that is my memory of why this is needed.

Running closure on the full/final output would be better, but I failed to land that change last time I tried because it caused a size regression (closure wan't able to optimize quite a a well for some reason).

src/wasm_worker.js
sbc100: Lets do that as a followup yes.

src/parseTools.mjs
sbc100: This property is part of the audio worklet API.. the system reads it.  without this fix the audio worklet tests fail.

src/audio_worklet.js
RReverser: I don't think it's really orthogonal, because now the main JS itself is used as audio worklet instead of having a separate file, so the naming conflicts if any become more likely.

RReverser: Basically: before this if you loaded Emscripten-generated JS in your own custom audio worklet, it's fine, no conflict would occur because the audio worklet itself lives in a separate file anyway, but now if you load it, the `registerProcessor` function will be invoked and it might conflict name-wise, so it's more important to keep it under unique prefix.

Name like `message` is just way too generic.

sbc100: Ideally we the generated code would know that its running in an AW (like we do with wasm workers and pthreads via the `em-ww` and `em-pthread` globals).. but you are right, the way AW seems to be started there is not way to inject any kind of options/settings.    It seems rather odd that nothing like is available when calling `addModule` (https://developer.mozilla.org/en-US/docs/Web/API/Worklet/addModule#options).  For `new Worker` we use the name field: https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker#name.  Maybe we can ask them to add that.

In the mean time I will make this change :)

RReverser (approved):

sbc100:
Ok, I think this is ready to land now.   Hopefully the tests will pass..

PR #24191, opened 2025-04-25
[embind] Initialize embind on wasm workers.

### LINKED ISSUE #24189: Expose Embindings in wasm worker
Hello.

I have a use-case where I need to run some EM_JS code in a WASM_WORKER. The function executed will be asynchronous (listen to a message) and needs to call back into WASM again. When testing this in the main thread, `Module` in the scope of `EM_JS` exist and does contain all bindings, but they do no exist when run in the Worker context.

```js
EM_JS(void, log_from_console, (), {
  self.addEventListener("message", (msg) => {
     Module.myMethod(msg.data.someData);
   });
});
```

Can an option to enable the bindings in the WASM_WORKER Module instance be created? Or that the binding are exported by default.

brendandahl:
Fixes #24189


tools/link.py
```diff
@@ -1405,6 +1405,8 @@ def limit_incoming_module_api():
 
   if settings.WASM_WORKERS:
     settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$_wasmWorkerInitializeRuntime']
+    if settings.EMBIND:
+      settings.REQUIRED_EXPORTS.append('_embind_initialize_bindings')
```
sbc100, line 1409: These two lines are duplicated in setup_pthreads.

How about moving them to the `if EMBIND` section on line 1383?

sbc100:
I spoke to @lindell about #24189 and it sounds like pthreads should work find for their use case.  But no harm fixing this I guess?

sbc100 (commented):
lgtm % comment

kripken (approved):

sbc100 (approved):

PR #24196, opened 2025-04-28
Add typings for explicit resource management

RReverser:
This is a follow-up to https://github.com/emscripten-core/emscripten/pull/23818.

When targeting any existing ES version, TypeScript transpiles `using ...` and doesn't require any typings.

However, when targeting ESNext, it expects `using ...` to be a native feature, and then looks up `[Symbol.dispose]` in the type and fails with a type validation error if it doesn't exist.

Since there's no way to do conditional compilation based on target version in TS, I'm adding this property to type declarations but marking with `ts-ignore` so that `Symbol.dispose` not existing is not an error in versions below ESNext.


brendandahl (approved):

RReverser:
Did Emscripten CI get a lot flakier lately? I've rerun this a dozen times now, and each time it fails with different, clearly unrelated errors...

PR #24208, opened 2025-04-28
Enable ruff COM rule and fix resulting warnings. NFC

sbc100:
This change was automatically generated using `ruff fix`.


kripken:
Is this convention common in the Python world? I see some benefit to it, but it also seems slightly more verbose in general.

sbc100:
> Is this convention common in the Python world? I see some benefit to it, but it also seems slightly more verbose in general.

I always follow this convention myself.  I like how to keeps diffs small when you add/remove list elements.

Note that there are two types of changes as part of this file (some of the changes remove verbosity. e.g. `(a, b,)` on one line is not `(a, b)`

kripken (approved):
Sounds good, I don't feel strongly, lgtm.

PR #24212, opened 2025-04-29
Fix manual triggering in test_audioworklet_emscripten_locks

sbc100:
At least under chrome it seems that the main thread doesn't run at all until the first few frames have been produced.

This fixes the test failures I've been seeing while working on #24190

### DIFF
--- a/test/webaudio/audioworklet_emscripten_locks.c
+++ b/test/webaudio/audioworklet_emscripten_locks.c
@@ -46,6 +46,18 @@ double startTime = 0;
 
 bool ProcessAudio(int numInputs, const AudioSampleFrame *inputs, int numOutputs, AudioSampleFrame *outputs, int numParams, const AudioParamFrame *params, void *userData) {
   assert(emscripten_current_thread_is_audio_worklet());
+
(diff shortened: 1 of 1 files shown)


sbc100:
@cwoffenden

dschuff (approved):
LGTM as a workaround to make the test work. Maybe we should file a bug to track investigating whether this is the desired behavior?

cwoffenden:
I guess all the audio tests have this problem, though none interact with the main thread (until they're finished).

It's a changed behaviour in Chrome, so certainly worth investigating why.

sbc100:
@juj, when you designed the audio worklet API did you imagine this kind of thing being possible?   Is it even valid (in the audio worklet API) to block on the main thread in this way?

cwoffenden:
> Are you still unable to reproduce ?

I ran it easily 10+ times today in each browser without problem, and countless times during development on many browser/OS combinations.

> Perhaps this test is simply not valid.

[snip]

Interesting point from the spec, and perhaps instead of the main thread a worker should be used? For our use, we don't interact with the main thread like this, it's AW and other threads.

PR #24218, opened 2025-04-29
Skip wasm worker tests in sanitizer configurations. NFC

sbc100:
This configuration was recently marked as unsupported (#24171)


sbc100:
This should fix the current waterfall breakage.

PR #24220, opened 2025-04-30
Deprecate -sUSE_WEBGPU in favor of --use-port=emdawnwebgpu

### LINKED ISSUE #23432: Emscripten's webgpu.h is behind the standard and Dawn implementations
Is it possible to update to a recent version of webgpu_cpp.h?

I am specifically interested in missing templated async functions.

like in the Buffer class.
```cpp
template <typename F, typename T,
              typename Cb = BufferMapCallback<T>,
              typename CbChar = void (MapAsyncStatus status, const char* message, T userdata),
              typename = std::enable_if_t<std::is_convertible_v<F, Cb*> || std::is_convertible_v<F, CbChar*>>>
Future MapAsync(MapMode mode, size_t offset, size_t size, CallbackMode callbackMode, F callback, T userdata) const;
template <typename L,
              typename Cb = BufferMapCallback<>,
              typename CbChar = std::function<void(MapAsyncStatus status, const char* message)>,
              typename = std::enable_if_t<std::is_convertible_v<L, Cb> || std::is_convertible_v<L, CbChar>>>
Future MapAsync(MapMode mode, size_t offset, size_t size, CallbackMode callbackMode, L callback) const;
```

kainino0x:
Add a linker warning for using `-sUSE_WEBGPU` pointing to `--use-port=emdawnwebgpu`, and update documentation.

Moves the port from `contrib.emdawnwebgpu` to just `emdawnwebgpu` since we'll keep maintaining it as we work on Emdawnwebgpu.

> Emscripten's implementation of webgpu.h is unmaintained. Dawn now maintains "emdawnwebgpu", a fork of Emscripten's implementation that is up to date, resolving [#23432](https://github.com/emscripten-core/emscripten/issues/23432).
> 
> We want people to move to emdawnwebgpu. However, the API has changed substantially, so it will take people time to switch over. Plus, there might be new bugs in that implementation. So we need a decently long deprecation period before removing the old implementation.

Closes #23432
Starts fixing #24265


tools/link.py
```diff
@@ -1257,6 +1257,9 @@ def limit_incoming_module_api():
   if not settings.GL_SUPPORT_SIMPLE_ENABLE_EXTENSIONS and settings.GL_SUPPORT_AUTOMATIC_ENABLE_EXTENSIONS:
     exit_with_error('-sGL_SUPPORT_SIMPLE_ENABLE_EXTENSIONS=0 only makes sense with -sGL_SUPPORT_AUTOMATIC_ENABLE_EXTENSIONS=0!')
 
+  if settings.USE_WEBGPU:
+    diagnostics.warning('deprecated', 'USE_WEBGPU is deprecated. Please try migrating to Dawn\'s "emdawnwebgpu" implementation, which implements a newer, more stable version of the standardized interface, and file feedback with Dawn (see package README). https://github.com/google/dawn/releases')
```
sbc100, line 1261: You can just add an entry to `DEPRECATED_SETTINGS` in `tools/settings.py`

xpenatan:
Hi, what about firefox or browsers that dont use dawn?

kainino0x:
emdawnwebgpu is independent of browser (except it supports a few experimental features that are only in Dawn). They're actually just a fork of the bindings currently in Emscripten, the only difference is we're maintaining them in the Dawn repository instead of here.

I should clarify this in the deprecation message and/or in the emdawnwebgpu docs.

sbc100:
I think we should maybe wait until we have a way for folks to do `--use-port=contrib/emdawn.py` (or something like that) to use the new version. i.e. lets fix https://github.com/emscripten-core/emscripten/issues/24260 first?

ypujante:
> I think we should maybe wait until we have a way for folks to do `--use-port=contrib/emdawn.py` (or something like that) to use the new version. i.e. lets fix #24260 first?

I agree 100%. It would be better to deprecate `-sUSE_WEBGPU` while offering an alternative that is as painless as possible. And using the `--user-port=contrib.emdawn` syntax would be the least painless possible...

kainino0x:
Seems reasonable.

sbc100 (approved):

sbc100 (commented):
Can you add a ChangeLog entry?

test/test_browser.py
ypujante, line 1: I would argue that since you and @sbc100 have made the decision to have emdawnwebgpu an internal port, then the port should be tested and as a result these tests should be either ported to use `emdawnwebgpu` or new tests for it should be added.

sbc100: Yes indeed, we will certainly want some emscripten-side test for this port.  That doesn't' need to be part of this PR necessarily though.

kainino0x: I've added a basic compile+link test. Thanks for the nudge.

kainino0x:
@sbc100 I think this is ready again. I've done a bunch of stuff since your last full review, so could you review again?

test/emdawnwebgpu_link_test.cpp
```diff
@@ -0,0 +1,235 @@
+// Copyright 2021 The Emscripten Authors.  All rights reserved.
```
sbc100, line 1: Can you name this file starting with `test_` to match the name of the test in python?

sbc100 (approved):
lgtm % comments

kainino0x:
The `dylink` variants failed with:
`Exception: FROZEN_CACHE is set, but cache file is missing: "sysroot/lib/wasm32-emscripten/pic/lib_emdawnwebgpu-8eb10848-O0.a" (in cache root path "/root/cache")`

I've removed those for now but how would I fix that?

sbc100:
Also the name of the library looks a little odd.   Libraries don't normally have an underscore after the `lib` part.  e.g we call them `libc` and `libz` and not `lib_c` and `lib_z`

kainino0x:
> If you want a port to be available in PIC mode here I think you would need to add it to `MINIMAL_PIC_TASKS` in `./embuilder.py`.

Thanks for the tip! It's probably not necessary, I think I'd rather not add a port to that list.

kainino0x:
> Also the name of the library looks a little odd. Libraries don't normally have an underscore after the `lib` part. e.g we call them `libc` and `libz` and not `lib_c` and `lib_z`

This name comes from the upstream port, so I'll take note and consider changing it at some point while I'm doing something else (doesn't seem important enough to make a whole CL to change).

sbc100:
> > If you want a port to be available in PIC mode here I think you would need to add it to `MINIMAL_PIC_TASKS` in `./embuilder.py`.
> 
> Thanks for the tip! It's probably not necessary, I think I'd rather not add a port to that list.

The point of that list is purely to enable this kind of testing.   `giflib` (a port) is already in the list.

Users won't be using that list.  And also `./embuilder ALL` will already include your new port so its already in embuillder in that sense.

sbc100:
> > Also the name of the library looks a little odd. Libraries don't normally have an underscore after the `lib` part. e.g we call them `libc` and `libz` and not `lib_c` and `lib_z`
> 
> This name comes from the upstream port, so I'll take note and consider changing it at some point while I'm doing something else (doesn't seem important enough to make a whole CL to change).

Indeed, not a blocker.

kainino0x:
> The point of that list is purely to enable this kind of testing. `giflib` (a port) is already in the list.
> 
> Users won't be using that list. And also `./embuilder ALL` will already include your new port so its already in embuillder in that sense.

Ah, then it wouldn't be so bad. Though thinking about it, it looks like that's a list of filenames, and since we have a filename that changes constantly it would introduce friction to the roll (unless maybe we detect that we're being downloaded in "remote port" configuration and remove the hash). It's fine, we want the more rigorous testing in Dawn anyway so we catch issues before landing them.

Issue #24162, opened 2025-04-21
wasmfs opfs rename directory fails with busy


rickg-hcl:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.6 (1ddaae4d2d6dfbb678ecc193bc988820d1fc4633)
clang version 21.0.0git (https:/github.com/llvm/llvm-project 4775e6d9099467df9363e1a3cd5950cc3d2fde05)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /Users/richardgillaspy/GitHubOpen/emsdk/upstream/bin
**Failing command line in full:**

**Full link command and output with `-v` appended:**
emcc -g  -v -s WASMFS -pthread -s PROXY_TO_PTHREAD=1 -sFORCE_FILESYSTEM  -sEXPORTED_FUNCTIONS='_malloc,_free' wasmfs_opefs_rename.c  -o test.html
 /Users/richardgillaspy/GitHubOpen/emsdk/upstream/bin/clang -target wasm32-unknown-emscripten -fignore-exceptions -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr --sysroot=/Users/richardgillaspy/GitHubOpen/emsdk/upstream/emscripten/cache/sysroot -D__EMSCRIPTEN_SHARED_MEMORY__=1 -DEMSCRIPTEN -Xclang -iwithsysroot/include/fakesdl -Xclang -iwithsysroot/include/compat -g3 -v -pthread -c wasmfs_opefs_rename.c -o /var/folders/d8/08bv2b6s2b50q04__ln3cmbh0000gp/T/emscripten_temp_ckhixeop/wasmfs_opefs_rename_0.o
clang version 21.0.0git (https:/github.com/llvm/llvm-project 4775e6d9099467df9363e1a3cd5950cc3d2fde05)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /Users/richardgillaspy/GitHubOpen/emsdk/upstream/bin
 (in-process)
[...]

PR #24227, opened 2025-04-30
21462 - work around directory rename

rickg-hcl:


sbc100:
Could you update the PR title and description with something a little more descriptive?  Is there an open bug for this fix?

rickg-hcl:
I updated the description.  This is a "fix" for https://github.com/emscripten-core/emscripten/issues/24162 in that it works around the missing directory rename functionality in opfs.

tlively:
Can you please update the PR description to describe the problem (e.g. the reason the backend does not already support directory renames) and the solution (e.g. what you are doing in the workaround and what limitations the approach might have)? That will help contextualize the code for review.

rickg-hcl:
updated description and refreshed branch

rickg-hcl:
@tlively Is there any chance of this merging ever?  Should we just keep this for our private use until OPFS supports directory rename?

steffanhalv:
I think it is also possible to use the FileSystemHandle.move() for directories, instead of copy and unlink now .. Or, it is not supported by chrome just yet: https://caniuse.com/mdn-api_filesystemhandle_move

PR #24229, opened 2025-04-30
[AUDIO_WORKLET] Use meaningful/unquoted properties messages. NFC

sbc100:
Now that audio worklet code is part of the main JS it all gets compiled together so this minification and quoting to make closure happy is no longer needed.


kripken (approved):
Might be worth verifying the relevant interactive checks pass manually (do they use closure?)

sbc100:
Confirmed.. I hear output when I run `browser.test_audio_worklet_params_mixing_minimal_with_closure`

PR #24232, opened 2025-05-01
Use normal unquoted names in wasm worker messages. NFC

sbc100:
Now that the wasm worker code is always in the same file as the main JS code there is no need to quote or minify these names.  See #24163.

### DIFF
--- a/src/lib/libwasm_worker.js
+++ b/src/lib/libwasm_worker.js
@@ -91,14 +91,14 @@ addToLibrary({
 #endif
   ],
   $_wasmWorkerInitializeRuntime: () => {
-    let m = Module;
 #if ASSERTIONS
-    assert(m && m['$ww']);
-    assert(m['sb'] % 16 == 0);
-    assert(m['sz'] % 16 == 0);
+    assert(wwParams);
+    assert(wwParams.wwID);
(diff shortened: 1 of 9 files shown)


RReverser (approved):

src/lib/libwasm_worker.js
```diff
@@ -113,11 +113,12 @@ addToLibrary({
     // already exists".  So for now, invoke this function from JS side. TODO:
     // remove this in the future.  Note that this call is not exactly correct,
     // since this limit will include the TLS slot, that will be part of the
-    // region between m['sb'] and m['sz'], so we need to fix up the call below.
-    ___set_stack_limits(m['sb'] + m['sz'], m['sb']);
+    // region between wwParams.stackLowAddress and wwParams.stackSize, so we
+    // need to fix up the call below.
+    ___set_stack_limits(wwParams.stackLowAddress + wwParams.stackSize, wwParams.stackLowestAddress);
```
kripken, line 118: Above we have `stackLowestAddress` and here `stackLowAddress` - ?

sbc100: I guess that explains why the tests didn't pass :)

kripken (approved):

PR #24234, opened 2025-05-01
Fix rpath calculation in nested dependencies

ryanking13:
This fixes a bug in rpath calculation in nested dependencies.

Basically, it fixes a case when a relative path is passed to the `loadDynamicLibrary`.

Think of the following scenario:

0. libhello.wasm depends on libhello_dep.so depends on libhello_nested_dep.so
1. [main.c] calls dlopen('/usr/lib/libhello.wasm')
2. [dynlink.c] loads '/usr/lib/libhello.wasm' and locates the absolute path of `libhello_dep.so` (`/usr/lib/libhello_dep.so`), and calls `loadDynamicLibrary('/usr/lib/libhello_dep.so')` (_dlopen_js ==> dlopenInternal ==> loadDynamicLibrary)
3. [libdylink.js] While loading `/usr/lib/libhello_dep.so`, it loads its dependency 'libhello_nested_dep.so' first. So it calls `loadDynamicLibrary('libhello_nested_dep.so')`.

In step 3, the relative path to the library is passed to the `loadDynamicLibrary`, and it wasn't correctly handled during RPATH calculation.

### DIFF
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -904,7 +904,8 @@ var LibraryDylink = {
     // We need to set rpath in flags based on the current library's rpath.
     // We can't mutate flags or else if a depends on b and c and b depends on d,
     // then c will be loaded with b's rpath instead of a's.
-    flags = {...flags, rpath: { parentLibPath: libName, paths: metadata.runtimePaths }}
+    var dso = LDSO.loadedLibsByName[libName];
+    flags = {...flags, rpath: { parentLibPath: dso.path, paths: metadata.runtimePaths }}
     // now load needed libraries and the module itself.
     if (flags.loadAsync) {
       return metadata.neededDynlibs
@@ -937,6 +938,7 @@ var LibraryDylink = {
     var dso = {
       refcount: Infinity,
       name,
+      path: name, // full path to the library, updated when the library is resolved in the filesystem.
       exports: syms,
       global: true,
     };
@@ -1105,6 +1107,7 @@ var LibraryDylink = {
 #endif
       if (f) {
         var libData = FS.readFile(f, {encoding: 'binary'});
+        dso.path = f;
         return flags.loadAsync ? Promise.resolve(libData) : libData;
       }
 #endif
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -7701,11 +7701,22 @@ int main() {
 
   @also_with_wasmfs
   def test_dlopen_rpath(self):
+    create_file('hello_nested_dep.c', r'''
+    #include <stdio.h>
+
+    void hello_nested_dep() {
+      printf("Hello_nested_dep\n");
+      return;
+    }
+    ''')
     create_file('hello_dep.c', r'''
     #include <stdio.h>
 
+    void hello_nested_dep();
+
     void hello_dep() {
       printf("Hello_dep\n");
+      hello_nested_dep();
       return;
     }
     ''')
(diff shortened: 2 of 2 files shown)


src/lib/libdylink.js
```diff
@@ -937,6 +939,7 @@ var LibraryDylink = {
     var dso = {
       refcount: Infinity,
       name,
```
sbc100, line 941: So does that mean that `name` is always just the basename / so_name?  Is is name sometimes absolute too here?

test/test_other.py
```diff
@@ -7747,20 +7758,22 @@ def test_dlopen_rpath(self):
     os.mkdir('subdir')
 
     def _build(rpath_flag, expected, **kwds):
-      self.run_process([EMCC, '-o', 'subdir/libhello_dep.so', 'hello_dep.c', '-sSIDE_MODULE'])
+      self.run_process([EMCC, '-o', 'subdir/libhello_nested_dep.so', 'hello_nested_dep.c', '-sSIDE_MODULE'])
+      self.run_process([EMCC, '-o', 'subdir/libhello_dep.so', 'hello_dep.c', '-sSIDE_MODULE', 'subdir/libhello_nested_dep.so'] + rpath_flag)
```
sbc100, line 7762: If you want to use the `.so` extensions in this test we should also rename `libhello.wasm` to `libhello.so` above.   Or use the `.wasm` extension here.  I don't really mind which.

src/lib/libdylink.js
```diff
@@ -904,7 +904,9 @@ var LibraryDylink = {
     // We need to set rpath in flags based on the current library's rpath.
     // We can't mutate flags or else if a depends on b and c and b depends on d,
     // then c will be loaded with b's rpath instead of a's.
-    flags = {...flags, rpath: { parentLibPath: libName, paths: metadata.runtimePaths }}
+    var dso = LDSO.loadedLibsByName[libName];
```
sbc100, line 907: If the module is already loaded (i.e. already in loadedLibsByName) why/how would end up in this function which is loading it?

src/lib/libdylink.js
```diff
@@ -904,7 +904,9 @@ var LibraryDylink = {
...
+    var libPath = dso?.path ?? libName;
+    flags = {...flags, rpath: { parentLibPath: libPath, paths: metadata.runtimePaths }}
```
sbc100, line 909: Should `parentLibPath` be just the dirname part of the parent?   i.e. is there any point in passing a value if its just a basename?

src/lib/libdylink.js
ryanking13: No, only the dependencies of the libraries will have only the so_name in `name`. The parent library called from `dlopen` would have absolute path in the `name`.

src/lib/libdylink.js
ryanking13: The module is not fully loaded yet here. The `exports` section would still have 'loading' value in it.

The call graph is:

- `loadDynamicLibrary`
  - adds `dso` field in `LSDO.loadedLibsByName` with exports field not filled out.
  - calls `loadWebAssemblyModule`
     - `loadWebAssemblyModule`
       - [the codepointer in this comment] locates the library from the file system
       - loads the WASM module
       - return with exports field inside the module
   - fills out the exports field in the dso object (the module is fully loaded now).
   - return

So... I don't think `dso` would be the best place to put this information, but I wasn't sure what other object to use.

src/lib/libdylink.js
ryanking13: I think it is possible to calculate the dirname here.

For now, the dirname is calculated inside [replaceORIGIN](https://github.com/emscripten-core/emscripten/blob/8bc8fff4397781587b2075c17e02e5984cef74e9/src/lib/libdylink.js#L52-L58) function, and the reason I did that was to isolate the code that depends on `FS` or `PATH` as much as possible.

If we calculate the dirname here, I think it can be something like:

```
    var dso = LDSO.loadedLibsByName[libName];
#if FILESYSTEM
    var libPath = PATH.dirname(dso?.path ?? libName);
#endif
    flags = {...flags, rpath: { parentLibPath: libPath, paths: metadata.runtimePaths }}
```

src/lib/libdylink.js
ryanking13: > No, only the dependencies of the libraries will have only the so_name in name.

So, I think the alternative approach would be ensuring `loadDynamicLibrary` always take absolute path as an input. It requires changes here:

https://github.com/emscripten-core/emscripten/blob/38e16464cffc8f886364fe4919a712131a4c9456/src/lib/libdylink.js#L909-L917

instead of passing `dynNeeded` directly to the `loadDynamicLibrary`, we can locate the library first, and pass the absolute path.

test/test_other.py
ryanking13: Makes sense. I renamed it to `.so` extension.

ryanking13:
@sbc100, May I ask for your review again when you have time?

sbc100:
@ryanking13 I'm curious about the different between this patch and #25694.   Do they both solve the same issue?   Do you have a preference for one over the other?

sbc100:
Also, could you rebase (or merge) this ?

ryanking13:
> @ryanking13 I'm curious about the different between this patch and https://github.com/emscripten-core/emscripten/pull/25694. Do they both solve the same issue? Do you have a preference for one over the other?

@sbc100 Thanks for the review!

I think both approaches somehow solve this nested dependency issue, but I prefer mine because I think this is more straightforward.  https://github.com/emscripten-core/emscripten/pull/25694 modifies the `libName` variable in multiple places which increases the cognitive load. This PR always resolves the `libname` before running `loadDynamicLibrary` soi I think it is better.

ryanking13:
@sbc100 

> Would this also fix https://github.com/emscripten-core/emscripten/issues/25814?

Unfortunately no, I tested the worker case but I get fiile not found error. Does `--embed-file`  parameter work differently in worker?

sbc100 (approved):

sbc100:
To get the codesize expectations to match you need to run `toosl/main/rebaseline_test.py` after first doing `emsdk install tot`.  If you have having trouble doing that let me know and I can push the expected results.

ryanking13:
@sbc100 

> To get the codesize expectations to match you need to run toosl/main/rebaseline_test.py after first doing emsdk install tot. If you have having trouble doing that let me know and I can push the expected results.

Thanks, updated the codesize. I think where was some code courrupted in my cache.

Issue #24236, opened 2025-05-01
Improve browser test shell


brendandahl:
With small browser windows the spinner, progress bar, and status text shift around causing the canvas below them to be at a different location from when the test starts. Ideally these things wouldn't cause the layout shift around. Some options:

1) Use a test only shell.html that has the minimal amount elements.
2) Change shell.html to have a fixed size top bar that doesn't affect the content below.

PR #24237, opened 2025-05-01
Fix flaky browser.test_sdl_mouse in Chrome.

brendandahl:
In small windows the spinner, progress bar, and status text shift around causing the canvas below them to be at a different location from when the test starts.

I've filed #24236 about better ways to fix this. For now, this seems like the smallest change.


kripken (approved):
Maybe add a comment here explaining why we have this flag?

sbc100 (approved):

sbc100:
Looks like some kind of circleci infra failure?  Maybe have to restart test run from the beggining?

PR #18285, opened 2022-12-01
Extend standalone support

jerbob92:
- Implement _emscripten_throw_longjmp that aborts the program
- Add nice messages to __cxa_throw and __cxa_allocate_exception
- Implement getentropy using WASI
- Implement WASI FS for syscalls lstat64, stat64, newfstatat, getdents64, fstat64 and openat

This is far from complete if you look at the WASI FS spec and the syscalls, but at least it gets basic operations like directory listing, file/directory opening and file/directory stat working. It was enough for me to get pdfium working in standalone mode.

I didn't really know what I should do with AT_FDCWD or what the runtime should do with it when passing AT_FDCWD. AT_FDCWD is also negative while I think `fd` in WASI is unsigned?

Most of the syscall code was taken from or based on wasi-libc.

PR #24246, opened 2025-05-02
Implement `open`/`stat`/`mkdir`/`symlink` for standalone WASI mode

jiixyj:
Based on <https://github.com/emscripten-core/emscripten/pull/18285>, implement just enough syscalls to make the `test_fcntl_open` test from `test_core` work in standalone WASI mode.

Compared to #18285 I moved the preopen handling into their own `paths.{h,c}` component. The idea is that this component will also be responsible for cwd handling in a follow up PR.

Some Notes:

- The preopen handling is done before `main` runs. Therefore, I removed the locking functions.
- To enable the `test_fcntl_open` test I had to extend the `@also_with_standalone_wasm` decorator a bit. The test doesn't run on node and wasmer. Wasmer crashes when resolving symlinks. However, both wasmtime and toywasm successfully run the test.
- Standalone WASM engines now get a temporary `fs` folder inside the `out/test` folder to put temporary files in. This folder is cleaned after each WASM engine is done. I couldn't make this work for wasmer, but for wasmtime and toywasm it works very nicely.

### DIFF
--- a/system/lib/standalone/paths.c
+++ b/system/lib/standalone/paths.c
@@ -1,0 +1,228 @@
+#define _GNU_SOURCE
+#include "paths.h"
+
+#include <assert.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sysexits.h>
+
+/// A name and file descriptor pair.
+typedef struct preopen {
+  /// The path prefix associated with the file descriptor.
+  const char* prefix;
+
+  /// The file descriptor.
+  __wasi_fd_t fd;
+} preopen;
+
+/// A simple growable array of `preopen`.
+static preopen* preopens;
+static size_t num_preopens;
+static size_t preopen_capacity;
+
+#ifdef NDEBUG
+#define assert_invariants() // assertions disabled
+#else
+static void assert_invariants(void) {
+  assert(num_preopens <= preopen_capacity);
+  assert(preopen_capacity == 0 || preopens != NULL);
+  assert(preopen_capacity == 0 ||
+         preopen_capacity * sizeof(preopen) > preopen_capacity);
+
+  for (size_t i = 0; i < num_preopens; ++i) {
+    const preopen* pre = &preopens[i];
+    assert(pre->prefix != NULL);
+    assert(pre->fd != (__wasi_fd_t)-1);
+#ifdef __wasm__
+    assert((uintptr_t)pre->prefix <
+           (__uint128_t)__builtin_wasm_memory_size(0) * PAGESIZE);
+#endif
+  }
+}
+#endif
+
+/// Allocate space for more preopens. Returns 0 on success and -1 on failure.
+static bool resize_preopens(void) {
+  size_t start_capacity = 4;
+  size_t old_capacity = preopen_capacity;
+  size_t new_capacity = old_capacity == 0 ? start_capacity : old_capacity * 2;
+
+  preopen* old_preopens = preopens;
+  preopen* new_preopens = calloc(sizeof(preopen), new_capacity);
+  if (new_preopens == NULL) {
+    return false;
+  }
+
+  memcpy(new_preopens, old_preopens, num_preopens * sizeof(preopen));
+  preopens = new_preopens;
+  preopen_capacity = new_capacity;
+  free(old_preopens);
+
+  assert_invariants();
+  return true;
+}
+
+// Normalize an absolute path. Removes leading `/` and leading `./`, so the
+// first character is the start of a directory name. This works because our
+// process always starts with a working directory of `/`. Additionally translate
+// `.` to the empty string.
+static const char* strip_prefixes(const char* path) {
+  while (1) {
+    if (path[0] == '/') {
+      path++;
+    } else if (path[0] == '.' && path[1] == '/') {
+      path += 2;
+    } else if (path[0] == '.' && path[1] == 0) {
+      path++;
+    } else {
+      break;
+    }
+  }
+
+  return path;
+}
+
+/// Register the given preopened file descriptor under the given path.
+///
+/// This function takes ownership of `prefix`.
+static bool register_preopened_fd(__wasi_fd_t fd, const char* relprefix) {
+  // Check preconditions.
+  assert_invariants();
+  assert(fd != AT_FDCWD);
+  assert(fd != -1);
+  assert(relprefix != NULL);
+
+  if (num_preopens == preopen_capacity && !resize_preopens()) {
+    return false;
+  }
+
+  char* prefix = strdup(strip_prefixes(relprefix));
+  if (prefix == NULL) {
+    return false;
+  }
+  preopens[num_preopens++] = (preopen){
+    prefix,
+    fd,
+  };
+
+  assert_invariants();
+  return true;
+}
+
+/// Are the `prefix_len` bytes pointed to by `prefix` a prefix of `path`?
+static bool
+prefix_matches(const char* prefix, size_t prefix_len, const char* path) {
+  // Allow an empty string as a prefix of any relative path.
+  if (path[0] != '/' && prefix_len == 0)
+    return true;
+
+  // Check whether any bytes of the prefix differ.
+  if (memcmp(path, prefix, prefix_len) != 0)
+    return false;
+
+  // Ignore trailing slashes in directory names.
+  size_t i = prefix_len;
+  while (i > 0 && prefix[i - 1] == '/') {
+    --i;
+  }
+
+  // Match only complete path components.
+  char last = path[i];
+  return last == '/' || last == '\0';
+}
+
+bool __paths_resolve_path(int* resolved_dirfd, const char** path_ptr) {
+  const char* path = *path_ptr;
+
+  if (*resolved_dirfd != AT_FDCWD && path[0] != '/') {
+    return true;
+  }
+
+  // Strip leading `/` characters, the prefixes we're mataching won't have
+  // them.
+  while (*path == '/')
+    path++;
+  // Search through the preopens table. Iterate in reverse so that more
+  // recently added preopens take precedence over less recently addded ones.
+  size_t match_len = 0;
+  int fd = -1;
+  for (size_t i = num_preopens; i > 0; --i) {
+    const preopen* pre = &preopens[i - 1];
+    const char* prefix = pre->prefix;
+    size_t len = strlen(prefix);
+
+    // If we haven't had a match yet, or the candidate path is longer than
+    // our current best match's path, and the candidate path is a prefix of
+    // the requested path, take that as the new best path.
+    if ((fd == -1 || len > match_len) && prefix_matches(prefix, len, path)) {
+      fd = pre->fd;
+      match_len = len;
+    }
+  }
+
+  if (fd == -1) {
+    return false;
+  }
+
+  // The relative path is the substring after the portion that was matched.
+  const char* computed = path + match_len;
+
+  // Omit leading slashes in the relative path.
+  while (*computed == '/')
+    ++computed;
+
+  // *at syscalls don't accept empty relative paths, so use "." instead.
+  if (*computed == '\0')
+    computed = ".";
+
+  *resolved_dirfd = fd;
+  *path_ptr = computed;
+  return true;
+}
+
+// Populate WASI preopens.
+__attribute__((constructor(100))) // construct this before user code
+static void _standalone_populate_preopens(void) {
+  // Skip stdin, stdout, and stderr, and count up until we reach an invalid
+  // file descriptor.
+  for (__wasi_fd_t fd = 3; fd != 0; ++fd) {
+    __wasi_prestat_t prestat;
+    __wasi_errno_t ret = __wasi_fd_prestat_get(fd, &prestat);
+    if (ret == __WASI_ERRNO_BADF)
+      break;
+    if (ret != __WASI_ERRNO_SUCCESS)
+      goto oserr;
+    switch (prestat.pr_type) {
+      case __WASI_PREOPENTYPE_DIR: {
+        char* prefix = malloc(prestat.u.dir.pr_name_len + 1);
+        if (prefix == NULL)
+          goto software;
+
+        // TODO: Remove the cast on `path` once the witx is updated with
+        // char8 support.
+        ret = __wasi_fd_prestat_dir_name(
+          fd, (uint8_t*)prefix, prestat.u.dir.pr_name_len);
+        if (ret != __WASI_ERRNO_SUCCESS)
+          goto oserr;
+        prefix[prestat.u.dir.pr_name_len] = '\0';
+
+        if (!register_preopened_fd(fd, prefix))
+          goto software;
+        free(prefix);
+
+        break;
+      }
+      default:
+        break;
+    }
+  }
+
+  return;
+oserr:
+  _Exit(EX_OSERR);
+software:
+  _Exit(EX_SOFTWARE);
+}
--- a/system/lib/standalone/standalone.c
+++ b/system/lib/standalone/standalone.c
@@ -12,10 +12,12 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
+#include <sys/stat.h>
 #include <malloc.h>
 #include <syscall_arch.h>
 #include <time.h>
 #include <unistd.h>
+#include <fcntl.h>
 
 #include <emscripten.h>
 #include <emscripten/heap.h>
@@ -25,6 +27,7 @@
 
 #include "lock.h"
 #include "emscripten_internal.h"
+#include "paths.h"
 
 /*
  * WASI support code. These are compiled with the program, and call out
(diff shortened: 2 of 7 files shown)


jiixyj:
You can run the tests locally by adding some WASM engines to `WASM_ENGINES` in `.emscripten` and then:

```sh
./test/runner -v test_hello_argc_standalone
# or
./test/runner -v test_fcntl_open_standalone
```

jiixyj:
Oh, and I have patched wasmer a bit, so behavior compared to a release version may differ:

```patch
diff --git a/lib/wasix/src/syscalls/wasix/path_open2.rs b/lib/wasix/src/syscalls/wasix/path_open2.rs
index 6305b8bf59..c2297e05dc 100644
--- a/lib/wasix/src/syscalls/wasix/path_open2.rs
+++ b/lib/wasix/src/syscalls/wasix/path_open2.rs
@@ -148,7 +148,7 @@ pub(crate) fn path_open_internal(
     //
     // Maximum rights: should be the working dir rights
     // Minimum rights: whatever rights are provided
-    let adjusted_rights = /*fs_rights_base &*/ working_dir_rights_inheriting;
+    let adjusted_rights = fs_rights_base & working_dir_rights_inheriting;
     let mut open_options = state.fs_new_open_options();
 
     let target_rights = match maybe_inode {
@@ -276,7 +276,10 @@ pub(crate) fn path_open_internal(
                 }
             }
             Kind::Dir { .. } => {
-                if fs_rights_base.contains(Rights::FD_WRITE) {
+                if fs_rights_base.contains(Rights::FD_WRITE)
+                    || o_flags.contains(Oflags::TRUNC)
+                    || o_flags.contains(Oflags::CREATE)
+                {
                     return Ok(Err(Errno::Isdir));
                 }
             }
```

jiixyj:
The `other.test_standalone_whole_archive` test complains about missing `_fd_prestat_dir_name` and `_fd_prestat_get` symbols:

```
error: undefined symbol: fd_prestat_dir_name (referenced by root reference (e.g. compiled C/C++ code))                                                                                          
warning: _fd_prestat_dir_name may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
error: undefined symbol: fd_prestat_get (referenced by root reference (e.g. compiled C/C++ code))                                                                                               
warning: _fd_prestat_get may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
```

Those are used in the preopen functions that run before main. Does anyone know how to fix this?

sbc100:
fd_prestat_dir_name and fd_prestat_get are WASI syscalls which are implemented in libwasi.js.. seems strange that they would be reported as undefined symbols.

test/common.py
```diff
@@ -1562,6 +1584,9 @@ def run_js(self, filename, engine=None, args=None,
       else:
         self.fail('JS subprocess failed (%s): %s (expected=%s).  Output:\n%s' % (error.cmd, error.returncode, assert_returncode, ret))
 
+    if run_in_tmpdir:
+      force_delete_contents(self.in_dir('fs'))
```
sbc100, line 1588: In general tests do not need to (indeed should not since helps with debugging) cleanup after themselves like this.

test/common.py
```diff
@@ -1505,7 +1522,11 @@ def cleanup(line):
   def run_js(self, filename, engine=None, args=None,
              assert_returncode=0,
              interleaved_output=True,
-             input=None):
+             input=None,
+             run_in_tmpdir=False):
```
sbc100, line 1526: I wonder if we can avoid adding this extra logic to `run_js` which is use in a lot of different places.

Could we instead allow just add `cwd=None` here and allow higher level functions to pass CWD?

Is there some fundamental reason why the tests need to run in a subdirectory?

test/common.py
```diff
@@ -981,14 +997,14 @@ def get_nodejs(self):
       return None
     return config.NODE_JS_TEST
 
-  def require_node(self):
+  def require_node(self, allow_wasm_engines=False):
```
sbc100, line 1000: It seems odd to mark a test as `require_node` but also allow wasm engines.   Maybe we can just remove `require_node` from the tests in question instead?

sbc100 (approved):

test/common.py
jiixyj: Yeah, I wonder about this, too. The original reason I did it like this comes from this line in `also_with_standalone_wasm`: https://github.com/emscripten-core/emscripten/blob/1ae691e9c61efe2d19fa2ab5e7cdd92f17596bf2/test/common.py#L639

Eventually this ends up in `require_engine()` where `self.wasm_engines` is set to an empty list. So what happened is that none of the WASM engines were actually used for any of the `also_with_standalone_wasm` tests!

Alternatively I could stash the `self.wasm_engines` in a local variable, then call `self.require_node()`, and then repopulate `self.wasm_engines`?

test/common.py
jiixyj: That's a good point! While there already is some existing cleanup logic it only ran after each "full" test, not in between each WASM engine run when iterating over them. This made the test fail because each test expects a blank directory.

What do you think about a subdir per each engine? So `out/temp/wasmer`, `out/temp/wasmtime`, etc.? Then the existing cleanup code should work just as is.

jiixyj: > What do you think about a subdir per each engine? So out/temp/wasmer, out/temp/wasmtime, etc.? Then the existing cleanup code should work just as is.

I implemented this and it works very well!

test/common.py
jiixyj: > Good idea! I'll try to implement this.

Done!

test/common.py
jiixyj: > Alternatively I could stash the self.wasm_engines in a local variable, then call self.require_node(), and then repopulate self.wasm_engines?

I've implemented this, avoiding this `allow_wasm_engines` parameter.

jiixyj:
> fd_prestat_dir_name and fd_prestat_get are WASI syscalls which are implemented in libwasi.js.. seems strange that they would be reported as undefined symbols.

Ah, they were gated behind `PURE_WASI`! I could fix this by gating the new preopen functions in the .c code behind `EMSCRIPTEN_PURE_WASI` as well.

jiixyj:
To fix some failing CI tests I had to do some additional things:

- Updated wasmtime to v32.0.0
- Added stubs for some WASI syscalls to `libsigs.js`/`libwasi.js` and to scripts under `tools`

I also added attribution to wasi-libc to `LICENSE`. Does that look OK to you?

The `test-browser-chrome-2gb` failed in the latest run, but I think this may be just a bit flaky?

system/lib/standalone/paths.c
```diff
@@ -0,0 +1,233 @@
...
+
+#include <assert.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sysexits.h>
+
+/// A name and file descriptor pair.
```
sbc100, line 12: How much of the code in this file is actually derived from the wasi-libc code?  If its actually derived from that codebase we should probably say so in the licence comment at the top.

Does wasi-libc use the triple-slash (///) for comments?   If not I would probably stick to normal C comments.

system/lib/standalone/paths.c
```diff
@@ -0,0 +1,233 @@
...
+/// A simple growable array of `preopen`.
+static preopen* preopens;
+static size_t num_preopens;
+
+/// Are the `prefix_len` bytes pointed to by `prefix` a prefix of `path`?
+static bool
+prefix_matches(const char* prefix, size_t prefix_len, const char* path) {
+  // Allow an empty string as a prefix of any relative path.
+  if (path[0] != '/' && prefix_len == 0)
+    return true;
```
sbc100, line 30: IIRC out coding style uses braces around the body of single line `if` blocks (unless this is copied code in which case we should mention why the style is different in a comment up top.

sbc100 (commented):
Thanks for working on this!

I like the direction of this PR.  I've not yet had a chance for fully grok the code in paths.c/h but the rest is looking good

system/lib/standalone/paths.c
jiixyj: Yeah, this is wasi-libc derived code. I changed the brace style to match Emscripten's, as they aren't consistent about it anyway. Does this work for you?

src/lib/libwasi.js
```diff
@@ -608,6 +608,21 @@ var WasiLibrary = {
     randomFill(HEAPU8.subarray(buffer, buffer + size));
     return 0;
   },
+
+  path_filestat_get__nothrow: true,
```
sbc100, line 612: Wrap these in `#if ALLOW_UNIMPLEMENTED_SYSCALLS` maybe?

jiixyj: Sounds good! Let's see what the CI says...

test/common.py
```diff
@@ -623,6 +623,9 @@ def metafunc(self, standalone):
       if not standalone:
         func(self)
       else:
+        nonlocal exclude_engines
+        if exclude_engines is None:
+          exclude_engines = []
```
sbc100, line 628: I think you just put these two lines at the top of the function (before the `def decorated`) and avoid the nonlocal thing.

You could also just write `exclude_engines=[]` in the signature and add `# noqa` to silence the warning.

test/common.py
```diff
@@ -635,9 +638,19 @@ def metafunc(self, standalone):
         self.emcc_args.append('-Wno-unused-command-line-argument')
         # if we are impure, disallow all wasm engines
         if impure:
-          self.wasm_engines = []
-        nodejs = self.require_node()
-        self.node_args += shared.node_bigint_flags(nodejs)
+          wasm_engines = []
+        else:
+          wasm_engines = [engine for engine in self.wasm_engines
+            if all(excluded not in os.path.basename(engine[0]) for excluded in exclude_engines)]
```
sbc100, line 644: I find these double nested comprehension quite hard to understand.  Can we expand into a loop or split in to two statements perhaps?

How about:

```
wasm_engines = []
if not impure:
  for engine in self.wasm_engines:
     basename = os.path.basename(engine[0])
     if not any(pattern in basename for pattern in excluded_engines):
       wasm_engines.append(engine)
```

test/common.py
```diff
@@ -614,7 +614,7 @@ def can_do_standalone(self, impure=False):
 
 # Impure means a test that cannot run in a wasm VM yet, as it is not 100%
 # standalone. We can still run them with the JS code though.
-def also_with_standalone_wasm(impure=False):
+def also_with_standalone_wasm(impure=False, exclude_engines=None):
```
sbc100, line 617: Can you convert add some info about `exclude_engines` to the above comment?

Something like: "exclude_engines" is a list of engine names on which the test cannot run.  These are pattern that are matched against the basename of the engine executable.

test/common.py
```diff
@@ -635,9 +638,19 @@ def metafunc(self, standalone):
...
+          wasm_engines = [engine for engine in self.wasm_engines
+            if all(excluded not in os.path.basename(engine[0]) for excluded in exclude_engines)]
+        if 'node' in exclude_engines:
+          self.js_engines = []
+          if not wasm_engines:
+            self.skipTest('no WASM engines available for this test')
+        else:
+          nodejs = self.require_node()
+          self.node_args += shared.node_bigint_flags(nodejs)
+        # `self.require_node()` clears `self.wasm_engines`, so set them here.
```
sbc100, line 652: I don't love this, but we can refactor later I suppose.

test/common.py
```diff
@@ -1966,11 +1981,15 @@ def _build_and_run(self, filename, expected_output, args=None,
...
         logger.warning('no wasm engine was found to run the standalone part of this test')
-      engines += self.wasm_engines
+      engines += [(engine, os.path.basename(engine[0])) for engine in self.wasm_engines]
     if len(engines) == 0:
       self.fail('No JS engine present to run this test with. Check %s and the paths therein.' % config.EM_CONFIG)
-    for engine in engines:
+    for engine, subdir in engines:
+      if subdir:
+        subdir = self.in_dir(subdir)
+        ensure_dir(subdir)
```
sbc100, line 1990: So it looks for wasm engines only your want to run the JS in a new directory with the name of the engine?

I guess that will work for some tests, but don't some tests need to be run in alongside the files in the test directory?

Why is this needed? What happens if you don't do the chdir here?

system/lib/standalone/standalone.c
```diff
@@ -65,21 +104,106 @@ weak int _munmap_js(
...
+        max |= __WASI_RIGHTS_FD_DATASYNC | __WASI_RIGHTS_FD_WRITE |
+               __WASI_RIGHTS_FD_ALLOCATE |
+               __WASI_RIGHTS_FD_FILESTAT_SET_SIZE;
+      }
+    } else if (accmode == O_EXEC || accmode == O_SEARCH) {
+      // Do nothing.
+    } else {
+      return -EINVAL;
+    }
+  }
```
sbc100, line 147: Is there some reason to add this extra scope here?

system/lib/standalone/paths.c
```diff
@@ -0,0 +1,254 @@
...
+      default:
+        break;
+    }
+  }
+
+  return;
+oserr:
+  _Exit(EX_OSERR);
+software:
+  _Exit(EX_SOFTWARE);
```
sbc100, line 251: Why not just `abort()` above instead of these goto + exit stuff?

test/common.py
jiixyj: done! `exclude_engines` isn't modified so doing it in the signature sounds good.

test/common.py
jiixyj: Agreed, this is much more readable!

sbc100:
This change looks like its almost ready to land.

Before we do though I wanted to check what your motivation for trying push this forward is?   Is there some use case where you want to use emscripten to target non-web engines?  Is there some reason why you need to use emscripten and not just wasi-sdk?

test/common.py
jiixyj: When I don't run inside a clean directory, the test for the second wasm engine would fail when trying to create `test-file` [here](<https://github.com/emscripten-core/emscripten/blob/f3d9eb42da8e7703fd319c9ca585b56ddad35732/test/fcntl/test_fcntl_open.c#L33>), as it already exists from the first wasm engine.

The current cleanup logic does not clean up the tests working directory after each engine, but only once in `setUp` for the whole thing.

> I guess that will work for some tests, but don't some tests need to be run in alongside the files in the test directory?

Yep, I have to admin that I don't have a good overview of the tests. But I thought that since the `wasm_engines` didn't have access to the filesystem before this PR, I at least can't break anything with this approach. The downside is of course that adding `@also_with_standalone_wasm` to a test like this won't work...

system/lib/standalone/standalone.c
jiixyj: Yes, it is there to limit `accmode`'s scope. I don't feel strongly about this, though.

test/common.py
sbc100: I see.. so the problem only exists when running a single test with mutliple engines at the same time?

This is not something that gets a lot testing.. I for one basically never do this.  

I kind of wonder if we should remove this feature?   For now could we just run in the first wasm engine instead of in all of them?  

Actually the `EMTEST_USE_ALL_ENGINES` setting is false so enless you enable it we should only be running in a single engine.  Perhaps we should fix that?

system/lib/standalone/paths.c
jiixyj: Good question -- this seems to be a wasi-libc thing. The code in `__main_void` (also derived from wasi-libc) uses `EX_OSERR` as well: <https://github.com/emscripten-core/emscripten/blob/f3d9eb42da8e7703fd319c9ca585b56ddad35732/system/lib/standalone/__main_void.c#L36>

test/common.py
sbc100: How about this solution:

- Leave `engines` as a simple list.
- Add a `self.directory_per_engine` test field which defaults to `false`
- Set this field to `true` in the tests that need it.
- Add this code here in the loop:

```
for engine, subdir in engines:
  cwd = None
  if self.directory_per_engine:
    # Some tests require a fresh directory for each engine, for example, because
    # they create file and are sensitive to existing files.
    dirname = os.path.basename(engine[0])
    ensure_dir(dirname)
    cwd = dirname
```

jiixyj:
> Before we do though I wanted to check what your motivation for trying push this forward is? Is there some use case where you want to use emscripten to target non-web engines? Is there some reason why you need to use emscripten and not just wasi-sdk?

For the fun of it I'm trying to port a hobby project of mine (CLI tool written in C++, using exceptions for error handling) to WASI. I actually tried wasi-sdk a few months ago but couldn't quite get exceptions to work. Then a few weeks ago I saw that Emscripten supported both a `STANDALONE_WASM`/`PURE_WASI` mode and even out of the box support for the new "exnref" exceptions (using `-sWASM_LEGACY_EXCEPTIONS=0`). So in the end it was faster for me to get to a working `.wasm` blob using Emscripten. But yeah, "on paper", wasi-sdk is more suitable for my use case...

test/common.py
jiixyj: Good idea, I'll try to implement this.

system/lib/standalone/paths.c
```diff
@@ -0,0 +1,254 @@
...
+  assert(num_preopens <= preopen_capacity);
+  assert(preopen_capacity == 0 || preopens != NULL);
+  assert(preopen_capacity == 0 ||
+         preopen_capacity * sizeof(preopen) > preopen_capacity);
+
+  for (size_t i = 0; i < num_preopens; ++i) {
+    const preopen* pre = &preopens[i];
+    assert(pre->prefix != NULL);
+    assert(pre->fd != (__wasi_fd_t)-1);
+#ifdef __wasm__
```
sbc100, line 129: Why `ifdef __wasm__` here?     IIUC this code only compiles and run on wasm

system/lib/standalone/paths.c
```diff
@@ -0,0 +1,254 @@
...
+  if (new_preopens == NULL) {
+    return false;
+  }
+
+  memcpy(new_preopens, old_preopens, num_preopens * sizeof(preopen));
+  preopens = new_preopens;
+  preopen_capacity = new_capacity;
+  free(old_preopens);
+
+  assert_invariants();
```
sbc100, line 154: Why not just `preopens = realloc(preopen, new_capacity)` here?

system/lib/standalone/paths.c
```diff
@@ -0,0 +1,254 @@
...
+// first character is the start of a directory name. This works because our
+// process always starts with a working directory of `/`. Additionally translate
+// `.` to the empty string.
+static const char* strip_prefixes(const char* path) {
+  while (1) {
+    if (path[0] == '/') {
+      path++;
+    } else if (path[0] == '.' && path[1] == '/') {
+      path += 2;
+    } else if (path[0] == '.' && path[1] == 0) {
```
sbc100, line 168: use '`\0`' here for the terminator?

This case looks like its trying to match `.` on its own and then return the empty string.. is that right?

system/lib/standalone/paths.c
```diff
@@ -0,0 +1,254 @@
...
+          fd, (uint8_t*)prefix, prestat.u.dir.pr_name_len);
+        if (ret != __WASI_ERRNO_SUCCESS) {
+          goto oserr;
+        }
+        prefix[prestat.u.dir.pr_name_len] = '\0';
+
+        if (!register_preopened_fd(fd, prefix)) {
+          goto software;
+        }
+        free(prefix);
```
sbc100, line 238: Given that `register_preopened_fd` is going to store and use the prefix, why not pass owership?  i.e. don't call `free()` here or `strdup` in `register_preopened_fd`?

kripken (commented):
(otherwise no comments aside from @sbc100 's)

PR #24254, opened 2025-05-05
[contrib.glfw3] Added support for closure compiler

ypujante:
With the support of closure args in ports, I updated contrib.glfw3 to support the closure compiler (reducing the JavaScript size quite significantly, ~42% decrease in size)

Ex:

```text
# previous version (no closure compiler)
-rw-r--r-- 1 runner docker 59959 Mar  5 13:08 index.js
# new version (with closure compiler)
-rw-r--r-- 1 runner docker 24994 May  5 16:21 index.js
```

### DIFF
--- a/tools/ports/contrib/glfw3.py
+++ b/tools/ports/contrib/glfw3.py
@@ -6,8 +6,8 @@
 import os
 from typing import Union, Dict
 
-TAG = '3.4.0.20250305'
-HASH = 'd5d1496cf1e05a640c182584a09ed0b43c8d6c42386513e5c2b1e05a2289854a0a0abe085e0a8b64c3a484836a8e63dc5ddb683ce4d2899fccbbf89536d58ee1'
+TAG = '3.4.0.20250505'
+HASH = 'f2e52b0b7489ec63cc32e4cadbc85faa9b5f09ae3f59a8335cb958706411de749f36cbe02df083b681340c0c2b8f6f3e6b631d2ac4589512f732c9593a5c10ef'
 
 # contrib port information (required)
 URL = 'https://github.com/pongasoft/emscripten-glfw'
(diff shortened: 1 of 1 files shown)


ypujante:
I found an issue in another project:

```text

building:ERROR: /usr/local/emsdk/upstream/emscripten/cache/ports/emscripten-glfw3/src/js/emscripten-glfw3-externs.js:3:0: WARNING - [JSC_UNDEFINED_EXTERN_VAR_ERROR] name Module is not defined in the externs.
  3| Module.requestFullscreen = function() {};
     ^^^^^^
```

Unclear why it works in some other project...

Closing this pull request until I can figure it out.

sbc100:
For Module properties we tend to use `Module['requestFullscreen']` rather than adding `requestFullscreen` to the externs file.  These two techniques have roughly the same effect.

sbc100:
Also we actually do want closure to minify the `Module` object name so we don't want to declare `Module` itself as an exten.

ypujante:
@sbc100 Thanks for the feedback

Do you know why the functions defined in the js_library are not minified:

Example:
```js
// index.js ran through the closure compiler
emscripten_glfw3_context_init:(a,b,c,d,e,f,g,h)=>{Wc["Module['canvas']"]=...
```

These functions are invoked only from cpp and defined like this in the js library file:

```js
  //! emscripten_glfw3_context_init
  emscripten_glfw3_context_init__deps: ['$specialHTMLTargets'],
  emscripten_glfw3_context_init__proxy: 'sync',
  emscripten_glfw3_context_init__sig: 'vpfpppppp',
  emscripten_glfw3_context_init: (context, scale, scaleChangeCallback, windowResizeCallback, keyboardCallback, clipboardCallback, requestFullscreen, errorHandler) => { ...}
```

sbc100:
Can you share a link to the original JS library code?

sbc100:
Are you asking why the name `emscripten_glfw3_context_init` is not minified?  It certainly should be.

ypujante:
This is the link to the [library code](https://github.com/pongasoft/emscripten-glfw/blob/master/src/js/lib_emscripten_glfw3.js). It gets included in the port with the code:

```python
  root_path = os.path.join(ports.get_dir(), port_name)
  source_js_path = os.path.join(root_path, 'src', 'js')
  settings.JS_LIBRARIES += [os.path.join(source_js_path, 'lib_emscripten_glfw3.js')]
```

And yes I am saying that the name `emscripten_glfw3_context_init` is not minified and appears in the `index.js` file (all functions in the library are not minified)

sbc100:
Oh, I think the issue is that wasm imports and export names are not minified at `-O2`.  If you use `-Os` or `-Oz` or `-O3` you should see those names get minified.

This is controlled by an internal setting called ` MINIFY_WASM_IMPORTS_AND_EXPORTS`.

PR #24255, opened 2025-05-05
Change definition of true and false macros to be js friendly

hoodmane:


system/lib/libc/musl/include/stdbool.h
```diff
@@ -3,8 +3,11 @@
 
 #ifndef __cplusplus
 
-#define true 1
-#define false 0
+// These work for both C and javascript.
+// In C !!0 ==> 0 and in javascript !!0 ==> false
+// In C !!1 ==> 1 and in javascript !!1 ==> true
+#define true (!!1)
+#define false (!!0)
```
sbc100, line 10: When does this come into play though?  Do these macros expand inside of EM_JS/EM_ASM macros?  

I'd like to see a test that fails without this change if possible.

hoodmane: They don't expand inside of them by default, but they do if you define another macro wrapping `EM_JS`. I can add a test.

hoodmane:
@sbc100 I added a test but curiously it does not pass because `stdbool.h` is included from `upstream/lib/clang/21/include/stdbool.h` and not from `cache/sysroot/include/stdbool.h`.

sbc100:
> @sbc100 I added a test but curiously it does not pass because `stdbool.h` is included from `upstream/lib/clang/21/include/stdbool.h` and not from `cache/sysroot/include/stdbool.h`.

But you've seen this fail locally for you?

hoodmane:
Yes, I've been using this in Pyodide for years:
https://github.com/pyodide/pyodide/blob/main/src/core/types.h#L12-L17

I just had to copy paste it into another project and figured it would be nice to upstream.

sbc100:
Another solution would be to using `-std=c17` (or C++) in which case these macros won't get defined.

hoodmane:
Yeah, I guess the system search path looks like:
```
.../emsdk-dev/upstream/lib/clang/21/include
.../emscripten/cache/sysroot/include
```
so the clang include files are preferred over the emscripten files.

sbc100:
We could potentially inject a fix like your into our `em_asm/em_js` headers?   I would certainly want to guard the fix behind `if defined(true) && defined(false)` though so it was a no-op in C++ / C-17

hoodmane:
Makes sense to me, I did that.

system/include/emscripten/em_js.h
```diff
@@ -72,3 +72,14 @@
 
 #define EM_ASYNC_JS(ret, name, params, ...) _EM_JS(ret, name, __asyncjs__##name, params,          \
   "{ return Asyncify.handleAsync(async () => " #__VA_ARGS__ "); }")
+
+
```
sbc100, line 76: Can we add a comment here as to why these are needed?

hoodmane:
Okay I added the comments you suggested.

emcc.py
```diff
@@ -969,6 +969,7 @@ def compile_source_file(input_file):
       # driver to perform linking which would be big change.
       cmd += ['-Xclang', '-split-dwarf-file', '-Xclang', unsuffixed_basename(input_file) + '.dwo']
       cmd += ['-Xclang', '-split-dwarf-output', '-Xclang', unsuffixed_basename(input_file) + '.dwo']
+    print(" ".join(cmd))
```
sbc100, line 972: Debugging?

test/test_other.py
```diff
@@ -16060,3 +16060,29 @@ def test_getentropy_d8(self):
...
+        }
+      })
+
+      int main() {
+        check_bool_type();
+        return 0;
+      }
+    ''')
+    self.do_runf('main.c')
+
```
sbc100, line 16088: Remove trailing newline

emcc.py
hoodmane: Oops fixed.

sbc100 (approved):
I wonder if its worth adding this without also adding something like `EM_JS_MACROS`.

I would imagine the best place for something like this would be alongside  something like `EM_JS_MACROS`.

Is the idea that there could be other users out there who try to write `EM_JS_MACROS` and this will help them avoid the pain you went through trying to debug stuff?

hoodmane:
Yeah. A few months ago, someone was asking how to make `EM_JS_MACROS` in an issue and I told them that this was the one tricky thing. I think it would make sense to add an `EM_JS_MACROS` macro at the same time. It is super useful if you have to do struct accesses:
```C
struct MyStruct {
   void* field1;
   void* field2;
}

#define MyStruct_field1_offset 0
#define MyStruct_field2_offset 4

_Static_assert(offsetof(MyStruct, field1) == MyStruct_field1_offset);
_Static_assert(offsetof(MyStruct, field2) == MyStruct_field2_offset);

#define MyStruct_field1(x) HEAPU32[x/4 + MyStruct_field1_offset]
#define MyStruct_field2(x) HEAPU32[x/4 + MyStruct_field2_offset]

EM_JS_MACROS(void, f, (MyStruct* arg), {
   const field1 = MyStruct_field1(x);
   const field2 = MyStruct_field2(x);
});
```

Another common use case is for sharing bitflags between C and JS:
```C
#define FLAG1 (1 << 0)
#define FLAG2 (1 << 1)
...

EMSCRIPTEN_KEEPALIVE int get_flags() {
    return FLAG1 | FLAG2;
}

EM_JS_MACROS(void, process_flags, (int flags), {
   if (flags & FLAG1) {
     // handle FLAG1
   }
   if (flags & FLAG2) {
     // handle FLAG2
   }
});
```

But the biggest reason for me is to inject standard error handling everywhere so that it's easier to implement a C calling convention in JS:
```C
#define EM_JS_REF(ret, func_name, args, body...)                               \
  EM_JS_MACROS(ret WARN_UNUSED, func_name, args, {                             \
    try    /* intentionally no braces, body already has them */                \
      body /* <== body of func */                                              \
    catch (e) {                                                                \
        LOG_EM_JS_ERROR(func_name, e);                                         \
        Module.handle_js_error(e);                                             \
        return 0;                                                              \
    }                                                                          \
    errNoRet();                                                                \
  })
```

hoodmane:
Opened #24272.

PR #24256, opened 2025-05-05
Remove no-op branch from acorn-optimizer. NFC

RReverser:
This line is equivalent to `(!fun.type) === 'FunctionDeclaration'` which is equivalent to `false === 'FunctionDeclaration'` which is equivalent to just `if (false)`.

The intent was obviously `!==` so fixing to reflect that.

### DIFF
--- a/tools/acorn-optimizer.mjs
+++ b/tools/acorn-optimizer.mjs
@@ -1731,6 +1731,9 @@ function minifyLocals(ast) {
   assert(extraInfo && extraInfo.globals);
 
   for (const fun of ast.body) {
+    if (fun.type !== 'FunctionDeclaration') {
+      continue;
+    }
     // Find the list of local names, including params.
     const localNames = new Set();
     for (const param of fun.params) {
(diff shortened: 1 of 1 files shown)


tools/acorn-optimizer.mjs
```diff
@@ -1731,9 +1731,6 @@ function minifyLocals(ast) {
   assert(extraInfo && extraInfo.globals);
 
   for (const fun of ast.body) {
-    if (!fun.type === 'FunctionDeclaration') {
```
sbc100, line 1734: How about `assert(fun.type == 'FunctionDeclaration')` just to be sure?

sbc100 (approved):

tools/acorn-optimizer.mjs
RReverser: Hmm, but that's not guaranteed at all? But I can see your point, the code below does seem to assume it's operating on function declarations. In that case `!==` is the correct fix after all.

sbc100 (approved):

PR #24259, opened 2025-05-05
[embind] Do not export InternalError and BindingError by default.

brendandahl:
These exports are likely not needed by default. They also use a non standard way of exporting on the Module object by manually assigning, which causes issues in the future with MODULARIZE=instance.


sbc100 (approved):
Nice!

I'm pretty sure I've written this change myself at some point too :)

brendandahl:
I added one more `count_emval_handles`

kripken (approved):

PR #24271, opened 2025-05-06
Fix `computed` checks in acorn-optimizer. NFC

RReverser:
`.computed` is a property of `MemberExpression`, but in a few cases it was incorrectly checked on `CallExpression`s.

This doesn't cause issues in existing tests, mostly noticed as a drive-by fix.

### DIFF
--- a/tools/acorn-optimizer.mjs
+++ b/tools/acorn-optimizer.mjs
@@ -421,9 +421,8 @@ function JSDCE(ast, aggressive) {
       },
       MemberExpression(node, c) {
         c(node.object);
-        // Ignore a property identifier (a.X), but notice a[X] (computed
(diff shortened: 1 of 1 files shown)


sbc100 (approved):

sbc100:
How hard would it be to write a test for this?

RReverser:
> How hard would it be to write a test for this?

Hard enough that I don't even want to start thinking about it 😅

sbc100:
> > How hard would it be to write a test for this?
> 
> Hard enough that I don't even want to start thinking about it 😅

Because constructing some JS to tickle this would be hard or because of unfamiliarity with the unit tests we have for JSDCE?

RReverser:
> Because constructing some JS to tickle this would be hard

This one.

PR #24272, opened 2025-05-06
Add EM_JS_MACROS

hoodmane:


system/include/emscripten/em_js.h
```diff
@@ -73,6 +73,9 @@
 #define EM_ASYNC_JS(ret, name, params, ...) _EM_JS(ret, name, __asyncjs__##name, params,          \
   "{ return Asyncify.handleAsync(async () => " #__VA_ARGS__ "); }")
 
+// Macros enabled version of EM_JS. Requires an extra layer of indirection.
+#define EM_JS_MACROS(ret, func_name, args, body...)                            \
```
sbc100, line 77: I wonder if we can come up with better name?   `EM_JS_WITH_MACROS` ?     `EM_JS_PREPROCESSED` ?   I don't really love them either :)

PR #24282, opened 2025-05-07
Make exported __cpp_exception into a global. NFC

sbc100:
This still is not quite a generic as I would like so as a followup we can probably make this work for all tags, and perhaps unify the updating of the globals when the exports are available.

### DIFF
--- a/src/lib/libexceptions.js
+++ b/src/lib/libexceptions.js
@@ -283,16 +283,11 @@ var LibraryExceptions = {
   },
 #endif
 #if WASM_EXCEPTIONS
-  $getCppExceptionTag: () =>
-    // In static linking, tags are defined within the wasm module and are
-    // exported, whereas in dynamic linking, tags are defined in library.js in
-    // JS code and wasm modules import them.
-#if RELOCATABLE
-    ___cpp_exception // defined in library.js
-#else
-    wasmExports['__cpp_exception']
-#endif
-  ,
+  $getCppExceptionTag__deps: ['__cpp_exception'],
+  // In static linking, tags are defined within the wasm module and are
+  // exported, whereas in dynamic linking, tags are defined in libcore.js in
+  // JS code and wasm modules import them.
+  $getCppExceptionTag: () => ___cpp_exception,
 
 #if EXCEPTION_STACK_TRACES
   // Throw a WebAssembly.Exception object with the C++ tag with a stack trace
--- a/src/preamble.js
+++ b/src/preamble.js
@@ -924,6 +924,11 @@ function getWasmImports() {
 #endif
 #endif
 
+#if hasExportedSymbol('__cpp_exception') && !RELOCATABLE
+    ___cpp_exception = wasmExports['__cpp_exception'];
+    {{{ receivedSymbol('___cpp_exception') }}};
+#endif
+
 #if hasExportedSymbol('__wasm_apply_data_relocs')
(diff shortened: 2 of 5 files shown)


src/lib/libexceptions.js
```diff
@@ -283,16 +283,11 @@ var LibraryExceptions = {
...
-    ___cpp_exception // defined in library.js
-#else
-    wasmExports['__cpp_exception']
-#endif
-  ,
+  $getCppExceptionTag__deps: ['__cpp_exception'],
+  // In static linking, tags are defined within the wasm module and are
+  // exported, whereas in dynamic linking, tags are defined in libcore.js in
+  // JS code and wasm modules import them.
+  $getCppExceptionTag: () => ___cpp_exception,
```
kripken, line 290: It looks like this splits the code from this file to two places, here and preamble.js. That seems less simple at first glance - are there other benefits?

sbc100: The removal of the dependency on `wasmExports` is the main benefit.   

That currently doesn't work in `WASM_ESM_INTEGRATION` mode.  We could make it work I suppose by adding an `import * as wasmExports` statement.. but it seems less precise that just importing what is used.

In the long run I think we should have a generic way to handle the non-function, non-global exports of a wasm module.

As of today we have special cases in a lot of places for `memory`, `__indirect_function_table` and `__cpp_exception` but I'd like to eliminate those if possible and replace them with generic export handling.

kripken: Now I see, thanks.

src/preamble.js
```diff
@@ -924,6 +924,11 @@ function getWasmImports() {
 #endif
 #endif
 
+#if hasExportedSymbol('__cpp_exception') && !RELOCATABLE
```
kripken, line 927: Why do we check for not relocatable?

sbc100: Because in relocatable mode symbols like this (along with the wasm table and memory) are not exported by the wasm but created in JS and *imported* by each wasm module.

kripken: In that case wouldn't `hasExportedSymbol` be false when relocatable?

sbc100: Yes we expect hasExportedSymbol to be false in relocatable mode and true for static builds.  Here are are checking for static builds we check for hasExportedSymbol to be true.

kripken: But then is
```
A = hasExportedSymbol('__cpp_exception') && !RELOCATABLE
```
not equivalent to
```
B = hasExportedSymbol('__cpp_exception')
```
? Looking at the two cases,

* If relocatable, then hasExportedSymbol is false, `A = false && false => false`, and `B = false`.
* If not relocatable, hasExportedSymbol may be true or false, but `!relocatable` is always true, so it does not cause a difference between `A` and `B`.

So can't we do this?
```suggestion
#if hasExportedSymbol('__cpp_exception')
```

kripken (approved):
lgtm regardless - if you think adding `!RELOCATABLE` is helpful to clarify, I don't object.

(or maybe I just misunderstood something)

src/preamble.js
sbc100: Sure you could argue these are redundant.  I'm mostly copying the existing logic for `__indirect_function_table` above.

sbc100:
I think I might wait on landing this until I can come up with a more generic solution,..

sbc100:
This did actually turn out to be necessary for my ESM integration work.

Issue #20354, opened 2023-09-28
Embind doesn't handle 64-bit types as expected in wasm64 mode


RReverser:
See https://github.com/emscripten-core/emscripten/pull/20353 for failing tests.

Issue #13902, opened 2021-04-15
unint64_t is passed through to JS as a signed value instead


cuberoot:
While trying to pass `std::numeric_limits<uint64_t>::max()` through to JS in bigint mode, I hit an issue where I get a -1 on the JS side. I would expect the value to come through correctly as bigint can represent a number that large. 

I can reproduce this with a cpp file with a JS library as follows:

Put this in `main.cpp`
```c++
#include <iostream>
#include <emscripten/bind.h>
#include <emscripten/emscripten.h>

extern "C" {
  extern void my_js(uint64_t value);
}

int main() {
    std::cout << std::numeric_limits<uint64_t>::max() << std::endl;
    my_js(std::numeric_limits<uint64_t>::max());

    return 0;
}
```

Put this in `main.js`:
```javascript
mergeInto(LibraryManager.library, {
    my_js: function(value) {
        console.log('value', value);
    },
});
```

Then compile with this command:
```shell
emcc --js-library main.js -s WASM_BIGINT main.cpp -o test.html
```

This gives me the following output when I load test.html:
```
18446744073709551615
test.js:4286 value -1n
```

mmarczell-graphisoft:
@juj @kripken @sbc100 I would be interested in working on this, but I have no idea where to start. Any ideas?

sbc100:
The issue I believe is that at the boundary of JS and Wasm we only have i32 and i64 and interpreting those types as either signed or unsigned is up to the embedder (in this case JS).   

In this example the JS library function doesn't actually know the signature of the native function, and neither the wasm module since the type was lowered to i64 during compilation.

I'm not even sure how to go about reinterpreting the BigInt here as an unisigned i64.   Perhaps one could write it do a DataView via `.setBigInt64()` and then read it back via `.getBigUint64()` but that seems like overkill.

juj:
> I hit an issue where I get a -1 on the JS side. I would expect the value to come through correctly as bigint can represent a number that large.

When JS side code receives values from Wasm, they will all come out as signed Numbers, or now as BigInts when supported and enabled. Wasm does not really have any machinery to know the intended interpretation of the signed vs unsigned aspect, so there will always be a need for the JS side to "fix up" the received numbers it gets.

https://github.com/emscripten-core/emscripten/pull/15433 aims to resolve issues like this. Also, there exists the library `src/library_int53.js` today, which can help non-BigInt uses.

[...]

PR #23990, opened 2025-03-26
[FS] Fix resolvePath handling of .. up from fs mountpoint

hoodmane:
node.parent doesn't actually give us the parent if the node is a FS root. In that case, instead of using `node.parent` we should start over from the beginning.

Resolves #23983.

PR #24285, opened 2025-05-08
Fix handling of uint64_t in Embind

RReverser:
Before this fix, `uint64_t` would be returned as a signed integer from Embind (e.g. if you return `UINT64_MAX`, it gets returned as `-1`).

I fixed that behaviour so that unsigned integers are correctly "fixed up" like they already are for `uint32_t`, and added tests for 64-bit integer limits to `test_i64_binding`.

In the process I had to also delete the invalid test `other.test_embind_long_long` - it had incorrect expectations (see commit message for more details) and is now superseded by the correct test mentioned above.

 - Fixes https://github.com/emscripten-core/emscripten/issues/20354. Closes https://github.com/emscripten-core/emscripten/issues/20353.
 - Fixes https://github.com/emscripten-core/emscripten/issues/13902.


test/embind/test_i64_binding.cpp
```diff
@@ -13,79 +13,76 @@
...
+
   register_vector<int64_t>("Int64Vector");
   register_vector<uint64_t>("UInt64Vector");
 }
 
 extern "C" void ensure_js_throws_with_assertions_enabled(const char* js_code, const char* error_type);
 
+// Checks that the given value has correctly preserved value and sign.
+template <typename T>
+void assert_bigint_preserved(T value) {
```
sbc100, line 41: Is this used anywhere?

sbc100, line 41: Is this used anywhere?

test/embind/test_i64_binding.cpp
```diff
@@ -13,79 +13,76 @@
 using namespace emscripten;
 using namespace std;
 
-#define assert_js(X) assert(run_js(X))
+#define assert_js_eq(X, Y) run_js(string("const x = ") + X + ", y = " + Y + "; assert(x === y, `" + X + ": actual = ${typeof x} ${x}, expected = ${typeof y} ${y}`);")
 
 void test(string message)
 {
```
sbc100, line 19: If you are gonna clean up this test maybe move all the open parens onto the declaration line?

test/embind/test_i64_binding.cpp
```diff
@@ -13,79 +13,76 @@
...
   cout << "test:\n" << message << "\n";
```
sbc100, line 20: How about just `printf("test: %s\n", message.c_str());`   (note that the first \n is removed so the output will need updating too).

src/lib/libembind.js
```diff
@@ -318,59 +318,50 @@ var LibraryEmbind = {
     }
   },
 
+#if ASSERTIONS
+  $assertIntegerRange: (name, value, minRange, maxRange) => {
```
sbc100, line 322: name => typeName?

src/lib/libembind.js
```diff
@@ -318,59 +318,50 @@ var LibraryEmbind = {
...
-    // out as 'i32 -1'. Always treat those as max u32.
-    if (maxRange === -1) {
-      maxRange = 4294967295;
-    }
 
-    var fromWireType = (value) => value;
+    const isUnsignedType = maxRange < minRange;
 
-    if (minRange === 0) {
+    let fromWireType = (value) => value;
```
sbc100, line 344: We are still using `var` in most places in the JS library code (mostly for code size reasons).   Any reason we need to use `let` here?

src/lib/libembind.js
```diff
@@ -318,59 +318,50 @@ var LibraryEmbind = {
...
   _embind_register_integer: (primitiveType, name, size, minRange, maxRange) => {
     name = readLatin1String(name);
-    // LLVM doesn't have signed and unsigned 32-bit types, so u32 literals come
-    // out as 'i32 -1'. Always treat those as max u32.
-    if (maxRange === -1) {
-      maxRange = 4294967295;
-    }
 
-    var fromWireType = (value) => value;
+    const isUnsignedType = maxRange < minRange;
```
sbc100, line 342: Does this work for all the integer types?  I guess so if the tests pass.

Perhaps comment here.  Something like "When minRange is passed as a negative value from the native side it will show up here is a larger unsigned value.".  

Also, is the optimization of `isUnsignedType` something that is separable perhaps?  Seems like it could land in isolation.. but maybe not worth it?

src/lib/libembind.js
RReverser: Embind already uses `let`/`const` in quite a few places. Just code style thing, no particular reason.

src/lib/libembind.js
RReverser: It's not (just) an optimisation, it actually doesn't work correctly sometimes since it relied on name of the type passed from C++ side and I changed how types are registered somewhat. This check is more reliable, and happens to be more optimized as well.

test/embind/test_i64_binding.cpp
RReverser: I don't mind, but that code was here before, seems out of scope of the PR?

test/embind/test_i64_binding.cpp
RReverser: Not anymore, good catch.

src/lib/libembind.js
RReverser: > When minRange is passed as a negative value from the native side it will show up here is a larger unsigned value.

FWIW it's the other way around. I'll add a comment.

test/embind/test_i64_binding.cpp
sbc100: Fair enough.. I guess I couldn't tell if you were generally cleaning up this test or just updating it.

For example the `run_js` improvement below?  Is that needed or just a cleanup?  I'd be find with a cleanup PR landing first (which is why I normally do when I work on tests are crufty)

RReverser: Yeah, also fair... I just noticed that two functions were doing exactly same thing, so I removed one of them, but then, hey, it doesn't make sense for body to be that complicated and "accidentally" cleaned that up too.

I can change this line as well, np.

test/code_size/embind_hello_wasm.json
```diff
@@ -1,10 +1,10 @@
 {
   "a.html": 552,
   "a.html.gz": 380,
-  "a.js": 9005,
```
RReverser, line 4: I mentioned this in Discord, I'm going to need some help here.

For a few tests `--rebaseline` generates wildly different sizes on my WIndows machine than those in the repo or on CI for some reason.

src/lib/libembind.js
RReverser: Actually I can simplify this to just `minRange === 0` and `minRange === 0n` correspondingly.

sbc100: Sweeeet!

system/lib/embind/bind.cpp
```diff
@@ -134,13 +134,6 @@ EMSCRIPTEN_BINDINGS(builtin) {
   register_integer<unsigned short>("unsigned short");
   register_integer<signed int>("int");
   register_integer<unsigned int>("unsigned int");
-#if __wasm64__
```
brendandahl: Why are these removed? Bindings for long will no longer work. 

Somewhat of an aside: I have had it on my todo to get rid of the `register_bigint<int64_t>("int64_t");` and unsigned version since that is actually not a "primitive" type. They're really `long long int` and `unsigned long long int`

RReverser: Yeah locally it worked regardless for some reason, but on CI it failed so I reverted.

RReverser: > I have had it on my todo to get rid of the register_bigint<int64_t>("int64_t"); and unsigned version since that is actually not a "primitive" type. They're really long long int and unsigned long long int

I agree, just wasn't sure that I should do it in this PR. Might as well I guess.

brendandahl: I think you need at least node 23 to run in 64 bit mode.

sbc100: I think node 24 ended up getting memory64 support.

test/code_size/embind_hello_wasm.json
kleisauke: You might also be interested in this Dockerfile:
https://github.com/emscripten-core/emscripten/pull/23990#issuecomment-2770475864

RReverser: Omg that's cursed. Should we maybe just use Node for gzipping or something? (anything where our version is pinned)

sbc100: That could maybe work yes, as long as it not dynamically linked against the system zlib.

RReverser: What's even more interesting, now that I'm looking at the logs deeper, it's the Wasm size that differs, not JS. I'm definitely on latest `tot`, I cleared the cache, how can binary be different :/

sbc100 (commented):
lgtm, but will defer to @brendandahl for final approval

RReverser:
Will fix test failures a bit later.

brendandahl (approved):

PR #24289, opened 2025-05-09
Add ChangeLog entries for 64-bit integer changes

RReverser:

PR #24283, opened 2025-05-08
Support 64-bit integers in acorn passes

RReverser:
I noticed that acorn-optimizer transforms for ASan, SAFE_HEAP and GROWABLE_HEAP didn't cover the 64-bit integer support enabled by `WASM_BIGINT` (`HEAP64` and `HEAPU64` memory views).

This PR adds demos of failing transforms in one commit, and actual support in the next one so that it's easier to see the diff between incorrectly generated code and the fixed one.

I've only tested on these snapshots for now, so waiting for CI to see if I missed any other tests that need to be updated.

PR #24291, opened 2025-05-09
Simplify and fix SAFE_HEAP

RReverser:
This streamlines the `SAFE_HEAP` transform in acorn-optimizer. I was going to send this as a separate cleanup PR as part of my other work on acorn-optimizer, but it happens to also be the easy fix for a regression mentioned in https://github.com/emscripten-core/emscripten/pull/24289 and caused by https://github.com/emscripten-core/emscripten/pull/24283 which added support for 64-bit integers to acorn-optimizer passes.

I could fix that issue separately by tweaking `unSign`, but this way we don't need to bother - the values with correct sign are now automatically retrieved by accessing the correct `HEAP*` array.

I've added a regression test for the linked issue as well, which fails before this PR and passes after.


test/test_core.py
```diff
@@ -7493,9 +7493,14 @@ def test_embind_dynamic_initialization(self):
...
   @no_wasm2js('wasm_bigint')
-  def test_embind_i64_val(self):
+  @parameterized({
+    '': ([],),
+    'safe_heap': (['-sSAFE_HEAP'],),
+  })
+  def test_embind_i64_val(self, extra_args):
     self.set_setting('WASM_BIGINT')
     self.emcc_args += ['-lembind']
+    self.emcc_args += extra_args
```
sbc100, line 7503: Maybe just do this one line `self.emcc_args += ['-lembind'] extra_args`.

I think the more common pattern is to call the param just `args`

test/other/codesize/test_codesize_hello_single_file.jssize
```diff
@@ -1 +1 @@
-6690
```
sbc100, line 1: Do you know why code size changes are part of this PR ?   It looks like it should only effect `-sSAFE_HEAP` programs?

src/runtime_safe_heap.js
```diff
@@ -8,91 +8,39 @@
...
-  }
-}
-
 #if SAFE_HEAP_LOG
 var SAFE_HEAP_COUNTER = 0;
 #endif
 
-/** @param {number|boolean=} isFloat */
-function SAFE_HEAP_STORE(dest, value, bytes, isFloat) {
+function SAFE_HEAP_INDEX(arr, idx, log) {
```
sbc100, line 15: Maybe call `log` something like `action` since its also just used the abort calls=?

src/runtime_safe_heap.js
```diff
@@ -8,91 +8,39 @@
...
 #endif
+  if (idx <= 0) abort(`segmentation fault ${log} ${bytes} bytes from address ${dest}`);
 #if EXIT_RUNTIME
   if (runtimeInitialized && !runtimeExited) {
 #else
   if (runtimeInitialized) {
 #endif
     var brk = _sbrk(0);
-    if (dest + bytes > brk) abort(`segmentation fault, exceeded the top of the available dynamic heap when storing ${bytes} bytes to address ${dest}. DYNAMICTOP=${brk}`);
+    if (dest + bytes > brk) abort(`segmentation fault, exceeded the top of the available dynamic heap when ${log} ${bytes} bytes from address ${dest}. DYNAMICTOP=${brk}`);
```
sbc100, line 31: I guess this can be rephrased now to something like `when attempting to ${log} ${bytes} bytes`

sbc100: (otherwise the grammer looks wrong I think)

sbc100: (sorry ignore that.. I see that actions do are of the `ing` form).

tools/acorn-optimizer.mjs
```diff
@@ -1667,81 +1617,10 @@ function safeHeap(ast) {
...
-              createLiteral(0),
-            ]);
-            break;
-          }
-        }
+        makeCallExpression(node, 'SAFE_HEAP_LOAD', [
+          node.object,
+          node.property
+        ]);
       }
```
sbc100, line 1624: Wow, this is way simpler.  Nice!

sbc100 (commented):
Nice!

test/other/codesize/test_codesize_hello_single_file.jssize
RReverser: I was surprised by it too but I long gave up trying to understand why code sizes are behaving the way they are by now 😅

RReverser: Looks like CI disagrees with my rebaseline, so I guess I need to update tot yet again.

Is this something we can simplify? Why do we rely on developer having latest tot rather than just use local Emscripten checkout we are in?

RReverser: Ah I guess it's because local checkout doesn't have binary deps? Can test runner download those separately into local checkout maybe?

sbc100: Isn't that basically that `emsdk install tot` does?   

Its pretty simple to use those binaryies, from your emscripten checkout, my preferred way is to just do `export EM_CONFIG=/path/to/emsdk/.emscripten` and then run `./test/runner` and it will use the emsdk binaries.

One followup I was intending to do is have the CI action be able to post the updated results back to the PR so you can just "accept" the results of the CI machine if you prefer.  Not sure how hard that would be though.

sbc100:
This seems like a clear improvement so I'm tempted to land it once that tests pass, but maybe we should wait for an approval from @kripken since he knows this code better then me.

sbc100 (approved):

test/other/codesize/test_codesize_hello_single_file.jssize
RReverser: > Isn't that basically that emsdk install tot does?

Yeah I just meant to have it 1) be done automatically and 2) isolated from my "normal" Emscripten setup as I kinda want stable Emscripten release when not contributing but just using it.

> One followup I was intending to do is have the CI action be able to post the updated results back to the PR so you can just "accept" the results of the CI machine if you prefer. Not sure how hard that would be though.

That would be amazing tbh.

RReverser:
@kripken Can you take a look please - this fixes a CI regression in emscripten-releases.

sbc100:
Is this the first in your series of patches to land?   And ready to go?

RReverser:
Either order is fine, they're pretty independent from each other. It's ready to go, I'm just waiting for @kripken's review since you suggested we should wait for his approval above.

RReverser:
I'm merging since this is a blocker. I'm happy to address follow-up comments if any in a separate PR.

src/runtime_safe_heap.js
```diff
@@ -8,91 +8,39 @@
...
+  const bytes = arr.BYTES_PER_ELEMENT;
+  const dest = idx * bytes;
 #if SAFE_HEAP_LOG
-  dbg('SAFE_HEAP store: ' + [dest, value, bytes, isFloat, SAFE_HEAP_COUNTER++]);
-#endif
-  if (dest <= 0) abort(`segmentation fault storing ${bytes} bytes to address ${dest}`);
-#if SAFE_HEAP == 1
-  if (dest % bytes !== 0) abort(`alignment error storing to address ${dest}, which was expected to be aligned to a multiple of ${bytes}`);
-#else
-  if (dest % bytes !== 0) warnOnce(`alignment error in a memory store operation, alignment was a multiple of ${(((dest ^ (dest-1)) >> 1) + 1)}, but was was expected to be aligned to a multiple of ${bytes}`);
```
kripken, line 38: What happened to the alignment checks?

RReverser: No longer necessary because we pass in an index in the corresponding array instead of pre-multiplied pointer.

kripken: Good point. I was also concerned about directly-emitted `SAFE_HEAP_*` calls from parseTools (not from the acorn transform), but I see we removed those a while ago, which I forgot...

kripken: I agree, yeah, best to just remove that bit, good find.

RReverser: Hm this section https://emscripten.org/docs/debugging/Sanitizers.html#comparison-to-safe-heap also seems outdated:

> The specific things SAFE_HEAP errors on include:
> 
> [...]
> 
> Unaligned reads or writes. These work in WebAssembly, but on some platforms an incorrectly-aligned read or write may be much slower, and with wasm2js (WASM=0) it will be incorrect, as JavaScript Typed Arrays do not allow unaligned operations.

Are unaligned reads/writes no longer supported on wasm2js or are they handled differently now?

kripken: No, that still seems accurate to me? Reading a 32-bit number from unaligned address 2 is `HEAP32[2 >> 2] = HEAP32[0]` which is wrong.

RReverser: `0` seems like easy case, but where would something like `HEAP32[123 >> 2]` -> `HEAP32[30]` be caught and handled? It seems at the moment the remainder part of the division is always ignored?

kripken: Normally it is silently ignored and behaves wrongly, but SAFE_HEAP in wasm checks it here:

https://github.com/WebAssembly/binaryen/blob/23a9e69f12afd749f35308dc504c8f6a89fc75f1/src/passes/SafeHeap.cpp#L376-L391

RReverser: I see. So, just to confirm, the comment above refers specifically to Wasm (and its wasm2js output) and it's not a bug that `SAFE_HEAP` JS transform doesn't perform a separate check for unaligned reads/writes?

kripken: No, I think it is a bug that SAFE_HEAP in JS doesn't do alignment checks. We should in principle check the same things in both places (say someone moves code from JS to wasm or back - we should check safety in the same way). But, I think we saw that that JS-side bug is long-standing and never caused any issues, so it doesn't feel urgent to fix.

sbc100: If the input code contains `HEAP32[x]` then by construction `x` must be 4-byte aligned, no?

kripken: Yes, but what about this?
```js
// a JS function that is called from wasm, and given a (maybe unaligned) pointer
function jsHelper(ptr) {
  return HEAP32[ptr >> 2];
}
```

RReverser: > But, I think we saw that that JS-side bug is long-standing and never caused any issues, so it doesn't feel urgent to fix.

Fair enough. I was mainly curious in context of those no-op checks we are commenting on, as I assume they used to do something useful in the long past, so technically a regression happened at some point... but idk when. As you say, it's probably not urgent.

> If the input code contains `HEAP32[x]` then by construction `x` must be 4-byte aligned, no?

Problem I'm thinking about, a lot of code has form like `HEAP32[somePtr >> 2]` and such, whether handcrafted or generated via `makeGetValue`, and in those cases there is no guarantee `somePtr` is already correctly aligned.

JS transform could detect at least this common form of access and check the alignment. Or maybe at least `makeGetValue`/`getHeapOffset` itself could insert those checks when `SAFE_HEAP` is enabled.

sbc100: > Yes, but what about this?
> 
> ```js
> // a JS function that is called from wasm, and given a (maybe unaligned) pointer
> function jsHelper(ptr) {
>   return HEAP32[ptr >> 2];
> }
> ```

But I don't think the old code handled this did it?  It just checked the value of `ptr >> 2` at runtime, right?   So there is no regression here is here?

sbc100: > JS transform could detect at least this common form of access and check the alignment. Or maybe at least `makeGetValue`/`getHeapOffset` itself could insert those checks when `SAFE_HEAP` is enabled.

To be clear neither the old new the version the JS transform do this though do they?

RReverser: > To be clear neither the old new the version the JS transform do this though do they?

If you mean before/after this PR, then yeah.

But judging by the removed code, it used to in the long past. And, well, even if it didn't, just saying it would be a useful addition.

Issue #18589, opened 2023-01-24
Replace GROWABLE_HEAPS with native functions


arsnyder16:
I was playing around with your benchmark, and i think i came up with an option to reduce the overhead of `warning: USE_PTHREADS + ALLOW_MEMORY_GROWTH may run non-wasm code slowly`

Instead of always relying on the array views, have a native function to call into which can cast the memory address, since this is on the wasm side no need to check if the underlying buffer has changed.

<details>
 <summary>Here is the new benchmark.cpp</summary>

```cpp
#include <iostream>
#include <string>
#include <chrono>
#include <emscripten.h>

extern "C" {
EMSCRIPTEN_KEEPALIVE uint8_t view_uint8(void* address) { return reinterpret_cast<uint8_t*>(address)[0]; }
}

int main() {
  auto start = std::chrono::steady_clock::now();
  size_t total = 0;
    for (const char* str :  { "hello", "world", "!", "test", "ing", "strings", "many", "of", "them" }) {
      total += EM_ASM_INT({
          var total = 0;
          var value;
          for (var i = 0; i < 1000000; ++i) {
              value = AsciiToString($0);
            total += value.length;
          }
          console.log(value);
          return total;
        }, str);
    }
  auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start).count() ;
  std::cout << total << " in " <<  ms << " ms " <<  total/ms << " /ms \n";
}
```

</details>

<details>
<summary>NO_ALLOW_MEMORY_GROWTH</summary>

```bash
~/emscripten/benchmark ((3.1.7))# em++ -Os -pthread -o -sNO_ALLOW_MEMORY_GROWTH -o benchmark.js benchmark.cpp 
~/emscripten/benchmark ((3.1.7))# node --experimental-wasm-threads --experimental-wasm-bulk-memory benchmark.js
hello
world
!
test
ing
strings
many
of
them
35000000 in 293 ms 119453 /ms 
```

</details>

<details>
<summary>ALLOW_MEMORY_GROWTH</summary>

```bash
~/emscripten/benchmark ((3.1.7))# em++ -Os -pthread -sALLOW_MEMORY_GROWTH -o benchmark.js benchmark.cpp 
em++: warning: USE_PTHREADS + ALLOW_MEMORY_GROWTH may run non-wasm code slowly, see https://github.com/WebAssembly/design/issues/1271 [-Wpthreads-mem-growth]
[...]

PR #24295, opened 2025-05-09
Simplify GROWABLE_HEAP transform

RReverser:
Similar to https://github.com/emscripten-core/emscripten/pull/24291, but for the transform handling `-pthread` + `-sALLOW_MEMORY_GROWTH`.

In the first commit I added a codesize variation to commit current sizes, and in the second you can see that this transform actually saves the total size in addition to making acorn-optimizer simpler.


test/test_other.py
```diff
@@ -10542,9 +10542,9 @@ def test_pthreads_growth_and_unsigned(self):
                       '-sMAXIMUM_MEMORY=4GB', '-sALLOW_MEMORY_GROWTH'])
     # growable-heap must not interfere with heap unsigning, and vice versa:
     # we must have both applied, that is
-    #   - GROWABLE_HEAP_I8() replaces HEAP8
+    #   - GROWABLE_HEAP() wraps HEAP8
```
sbc100, line 10545: `GROWABLE_HEAP(HEAP8) wraps HEAP8`  ?

src/modules.mjs
```diff
@@ -456,16 +456,7 @@ function exportRuntimeSymbols() {
...
-      'GROWABLE_HEAP_I16',
-      'GROWABLE_HEAP_U16',
-      'GROWABLE_HEAP_I32',
-      'GROWABLE_HEAP_U32',
-      'GROWABLE_HEAP_I64',
-      'GROWABLE_HEAP_U64',
-      'GROWABLE_HEAP_F32',
-      'GROWABLE_HEAP_F64',
+      'GROWABLE_HEAP',
     );
```
sbc100, line 460: I suppose technically this is a change in the public API.. so maybe we continue to include these aliases?  e.g. `GROWABLE_HEAP_I8  = () => GROWABLE_HEAP(HEAP8)`?

sbc100 (commented):
Nice!

src/modules.mjs
RReverser: I thought those are just internal helpers?

sbc100: I assume they are .. but `runtimeElements` here (which these are being added to) is a list of things that you can include in `-sEXPORTED_RUNTIME_METHODS`.

Also, internal helpers are still callable from user code.. so I guess it depends if we think its likely anyone is actually calling these?   Maybe no?

sbc100: If they really are internal we can probably remove this whole block here instead of updating it.

test/test_other.py
sbc100: SGTM

src/modules.mjs
RReverser: > Also, internal helpers are still callable from user code.. so I guess it depends if we think its likely anyone is actually calling these? Maybe no?

Yeah that's a good question... I only found discussions like https://github.com/emscripten-core/emscripten/issues/18589 where you suggested those helpers will be gone altogether, which sounds like it means they should be internal - if not already then at least for future-proofing.

They're not documented anywhere, so that seems like a good sign but I'm not sure if there's anything else we can do to make them "truly" internal?

sbc100: I'm up for simply deleting this block then.   Lets check with @kripken before landing though.

RReverser: > I assume they are .. but `runtimeElements` here (which these are being added to) is a list of things that you can include in `-sEXPORTED_RUNTIME_METHODS`.

Ugh that's actually worrying. I was going to do some simplifications to `LE_HEAP_LOAD`/`LE_HEAP_STORE` as well - those are slightly different from those 2 PRs, but there's still some code reduction & size savings to be had there, but those would be breaking changes too if we're not allowed to assume those helpers are internal...

sbc100: I think little endian has very very few users.. if we document the change I think its pretty safe.

RReverser: > I think little endian has very very few users.. 

I have to say, this phrase sounds really funny out of context.

fivemtebex:
The documentation in this repository is excellent! Very clear and comprehensive.

fivemeyestore:
Great implementation of this feature. I particularly like the approach to error handling.

RReverser:
Ugh AI bots are getting annoying. I don't have the rights to block users in this repo it seems, will leave that to others.

fivemhostingserver:
Great implementation of this feature. I particularly like the approach to error handling.

scriptforfivem:
Impressive project structure! The architecture really demonstrates solid design principles.

fivemserverhostings:
Excellent contribution to open source! The community will benefit greatly from this.

RReverser:
Converted to draft because I thought of some cases where this change can be problematic without further work, even though those cases are not yet covered by tests.

SAFE_HEAP one should be still good to go.

RReverser:
> Converted to draft because I thought of some cases where this change can be problematic without further work, even though those cases are not yet covered by tests.

Converted to ready for review because turns out it's not a regression, the edge case I was thinking of (memory growth + ASAN/SAFE_HEAP) is already not working as expected and will be addressed separately in #24309.

sbc100 (approved):

RReverser:
I want to do a bit more testing here just in case, so holding off merging for now.

RReverser:
Yessss. After investigating #24309, I realised another issue with this refactoring, and I knew this should be failing in some cases, but couldn't make it break.

Turns out, it's tricky to trigger that edge case due to pthreads involved and due to Emscripten accessing the heap in a lot of functions (and, thus, updating it even before it's actually needed), but finally managed to craft a test that breaks after this transform.

One of those rare feelings when things work but they really shouldn't 😅 I'll submit a new test either in this PR or separately.

RReverser:
Ok the tests look pretty different now because majority of Emscripten APIs that are implemented in JS touch HEAP in one way or another, and so even logging via `printf`/`puts` or macro like `EM_ASM` would hide the issue by updating the heap bindings too early.

I had to drop all the way down to `EM_JS` + atomics, which successfully broke all those tests as expected (showing that memory view is not updated on the first access), and subsequently fixed them by changing transform from `GROWABLE_HEAP(HEAPxx)` to `(GROWABLE_HEAP(), HEAPxx)` which ensures that the heap value is always retrieved _after_ updating the views, and not evaluated before that.

@kripken @sbc100 I'm waiting for the CI to see if this uncovered anything else, but otherwise should be ready for another look.

RReverser:
> I don't understand why you can't use use the return value of `GROWABLE_HEAP(HEAPxx)`.. can you explain a little more? Is the problem that incoming argument to `GROWABLE_HEAP` would somehow be wrong?

Arguments are evaluated before the function itself, which is what I was trying to demonstrate & catch with the updated tests.

If you do

```js
let gotHeap = GROWABLE_HEAP(HEAP8);
```

then it's equivalent to

```js
let gotHeap = (function GROWABLE_HEAP(arr) {
  // inside GROWABLE_HEAP:
  HEAP8 = new Uint8Array(...);
  HEAP16 = ...;
  ...;
  return arr;
})(HEAP8);
```

which in turn is like

```js
let arr = HEAP8;
HEAP8 = new Uint8Array(...);
HEAP16 = ...;
...;
let gotHeap = arr; // the old value of HEAP8, not the one from assignment above
```

sbc100:
I see.   In that case perhaps we can rename `GROWABLE_HEAP()` to something more meaningful?  "MAYBE_UPDATE_HEAP_VIEWS" (seems a bit lot).  `CHECK_MEM_GROWTH` (better?)?

RReverser:
> In that case perhaps we can rename `GROWABLE_HEAP()` to something more meaningful? "MAYBE_UPDATE_HEAP_VIEWS" (seems a bit lot). `CHECK_MEM_GROWTH` (better?)?

Yeah, either is fine by me. I was thinking of `maybeUpdateMemoryViews()` to be in line with `updateMemoryViews()` but wasn't sure if worth it.

RReverser:
> > In that case perhaps we can rename `GROWABLE_HEAP()` to something more meaningful? "MAYBE_UPDATE_HEAP_VIEWS" (seems a bit lot). `CHECK_MEM_GROWTH` (better?)?
> 
> Yeah, either is fine by me. I was thinking of `maybeUpdateMemoryViews()` to be in line with `updateMemoryViews()` but wasn't sure if worth it.

I'll just do this, especially since this function is no longer exposed as runtime symbol and doesn't need to have the uppercase name.

sbc100:
Yeah I guess there is no reason to do with all caps here.

sbc100:
Since this symbol will appear a *lot* in unminified output is it worth picking a short-ish name?  Maybe not?

RReverser:
Eh, personally I'd prioritise debuggability over code size for unminified JS.

src/growableHeap.js
```diff
@@ -6,7 +6,7 @@
 
 // Support for growable heap + pthreads, where the buffer may change, so JS views
 // must be updated.
-function GROWABLE_HEAP() {
+function maybeUpdateMemoryViews() {
```
sbc100, line 9: I think I prefer `checkMemGrowth` myself, since this function is only needed in the memory growth case.

RReverser: I mildly dislike it because "check" sounds synonymous with "assert" (check that some invariant holds) and doesn't convey that it actually is a side-effectful function that might update things, but I also don't want to bikeshed too long and happy to change if you disagree :)

sbc100: Hmm.. I see what you mean.  I would also accept `handleMemGrowth` but I think `checkMemGrowth` is slightly more meaningful to me.

sbc100: SGTM

RReverser:
CI looks good and I feel more confident in this change now 🤞

test/pthread/test_pthread_memory_growth.c
```diff
@@ -4,42 +4,64 @@
...
+// Checking this correctly is somewhat tricky because a lot of Emscripten APIs access heap on their own, so the memory views might be updated
+// by them before we explicitly check the state in our own JS routines below, which leads to false positives (test passing even though in
+// isolation the heap access in our own JS is not the one updating the memory views).
+//
+// To test it in isolation, we need to use only EM_JS - which is as close to a pure JS call as we can get - and not EM_ASM, pthread_join, proxying
+// etc., not even `puts`/`printf` for logging - as all of those access the heap from JS on their own and might update the memory views too early.
+// Not using standard proxying mechanisms also means we need to drop down all the way to raw atomics.
+
+EM_JS(void, assert_initial_heap_state, (bool isWorker), {
+  console.log(`Checking initial heap state on the ${isWorker ? 'worker' : 'main'} thread`);
```
sbc100, line 23: Can you use `dbg` here .. `console.log` is not the best thing on node workers and `dbg` will go direct to stdout.

test/pthread/test_pthread_memory_growth.c
```diff
@@ -4,42 +4,64 @@
...
+
+// We want to test that JavaScript access in the main thread automatically updates the memory views if the heap has grown from the pthread meanwhile.
+//
+// Checking this correctly is somewhat tricky because a lot of Emscripten APIs access heap on their own, so the memory views might be updated
+// by them before we explicitly check the state in our own JS routines below, which leads to false positives (test passing even though in
+// isolation the heap access in our own JS is not the one updating the memory views).
+//
+// To test it in isolation, we need to use only EM_JS - which is as close to a pure JS call as we can get - and not EM_ASM, pthread_join, proxying
+// etc., not even `puts`/`printf` for logging - as all of those access the heap from JS on their own and might update the memory views too early.
+// Not using standard proxying mechanisms also means we need to drop down all the way to raw atomics.
```
sbc100, line 20: Can you format this comment 80 columns

test/pthread/test_pthread_memory_growth.c
```diff
@@ -4,42 +4,64 @@
...
+  return NULL;
 }
 
 int main()
 {
-  printf("prep\n");
+  assert_initial_heap_state(false);
 
   pthread_t thr;
+  const char *_Atomic buffer = NULL;
```
sbc100, line 54: Maybe just make this a global atomic to avoid passing it via void* and the extra cast in thread_start?

test/pthread/test_pthread_memory_growth_mainthread.c
```diff
@@ -4,51 +4,95 @@
...
+EM_JS(void, js_assert_initial_heap_state, (bool isWorker), {
+  console.log(`Checking initial heap state on the ${isWorker ? 'worker' : 'main'} thread`);
+  assert(HEAP8.length === 32 * 1024 * 1024, "start at 32MB");
+});
+
+EM_JS(void, js_assert_final_heap_state, (bool isWorker, const char* buffer, int finalHeapSize), {
+  console.log(`Checking final heap state on the ${isWorker ? 'worker' : 'main'} thread`);
+  assert(HEAP8.length > finalHeapSize, "end with >64MB");
+  assert(HEAP8[buffer] === 42, "readable from JS");
+});
```
sbc100, line 31: Seems like a lot of code duplication between these two tests.. but I guess that is preexisting :(

test/test_browser.py
```diff
@@ -4726,27 +4726,27 @@ def test_minimal_runtime_hello_thread(self, opts):
...
-    'proxy': (['-sPROXY_TO_PTHREAD'],),
+    'proxy': (['-sPROXY_TO_PTHREAD'], 2),
     'minimal': (['-sMINIMAL_RUNTIME', '-sMODULARIZE', '-sEXPORT_NAME=MyModule'],),
   })
   @no_2gb('uses INITIAL_MEMORY')
   @no_4gb('uses INITIAL_MEMORY')
-  def test_pthread_growth(self, emcc_args):
-    self.emcc_args.remove('-Werror')
-    self.btest_exit('pthread/test_pthread_memory_growth.c', emcc_args=['-pthread', '-sPTHREAD_POOL_SIZE=2', '-sALLOW_MEMORY_GROWTH', '-sINITIAL_MEMORY=32MB', '-sMAXIMUM_MEMORY=256MB', '-g'] + emcc_args)
+  def test_pthread_growth(self, emcc_args, pthread_pool_size = 1):
```
sbc100, line 4747: Looks like this argument is never used.  Maybe just leave it hardcoded below?

RReverser: It is - it's overridden to "2" in the proxy variant, just default to "1" otherwise.

test/pthread/test_pthread_memory_growth.c
RReverser: Yeah could do, either is fine by me.

test/pthread/test_pthread_memory_growth.c
sbc100: See https://github.com/emscripten-core/emscripten/issues/14804

sbc100 (approved):
lgtm.  @kripken can you also approve this?

src/lib/libpthread.js
```diff
@@ -979,11 +979,6 @@ var LibraryPThread = {
   $establishStackSpace__internal: true,
   $establishStackSpace__deps: ['$stackRestore', 'emscripten_stack_set_limits'],
   $establishStackSpace: (pthread_ptr) => {
-#if ALLOW_MEMORY_GROWTH
-    // If memory growth is enabled, the memory views may have gotten out of date,
-    // so resync them before accessing the pthread ptr below.
-    updateMemoryViews();
-#endif
```
kripken, line 986: Was this a special case that was never needed?

test/pthread/test_pthread_memory_growth_mainthread.c
```diff
@@ -4,51 +4,101 @@
...
+
+EM_JS(void, js_assert_initial_heap_state, (bool isWorker), {
+  console.log(`Checking initial heap state on the ${isWorker ? 'worker' : 'main'} thread`);
+  assert(HEAP8.length === 32 * 1024 * 1024, "start at 32MB");
+});
+
+EM_JS(void, js_assert_final_heap_state, (bool isWorker, const char* buffer, int finalHeapSize), {
+  console.log(`Checking final heap state on the ${isWorker ? 'worker' : 'main'} thread`);
+  assert(HEAP8.length > finalHeapSize, "end with >64MB");
+  assert(HEAP8[buffer] === 42, "readable from JS");
```
kripken, line 36: Perhaps stash a reference to HEAP8 initially, then compare it later and see it changed?

tools/acorn-optimizer.mjs
```diff
@@ -1288,28 +1288,27 @@ function littleEndianHeap(ast) {
...
+          (node.id.name === 'growMemViews' || node.id.name === 'LE_HEAP_UPDATE')
         )
       ) {
         c(node.body);
       }
     },
     AssignmentExpression: (node) => {
-      if (node.left.type === 'Identifier' && isEmscriptenHEAP(node.left.name)) {
+      if (node.left.type !== 'Identifier') {
         // Don't transform initial setup of the arrays.
```
kripken, line 1309: This comment looks stale.

kripken:
(lgtm otherwise)

test/pthread/test_pthread_memory_growth.c
RReverser: Oof TIL, that might explain some odd console.log behaviour I saw in the past. 

> I tested and this remains broken in node 14 but is fixed in node 15+. 

Sounds like it should be fixed now? Either way, yeah, happy to change.

src/lib/libpthread.js
RReverser: Yeah: https://github.com/emscripten-core/emscripten/pull/24295/commits/333c9f4cc3cb1e8c0f48408e0ee8da41d386f154

test/pthread/test_pthread_memory_growth_mainthread.c
RReverser: Could do, sure.

Although the length checks already ensure growth and should be somewhat more reliable if/when the resizable ArrayBuffer comes to Wasm, as in that case the heap reference will remain the same but the size should still change.

tools/acorn-optimizer.mjs
RReverser: It's still relevant - that's the main usecase we want to cover here by skipping the lhs - but I can clarify the comment.

test/pthread/test_pthread_memory_growth_mainthread.c
RReverser: I think I'm going to keep it as-is for reasons above. Stashing a reference doesn't add new value but does make test potentially less reliable in the future.

RReverser:
FAIL [0.002s]: test_async_hello_v8 (test_core.instance) looks like a broken test on main branch.

kripken:
Trying to read the changes since my last review, I have a problem that I've seen in @sbc100 's PRs as well with force-pushes... how can I see only the last changes? The last "compare" button shows what appears to be a merge commit from main, and nothing else, and otherwise I see 13 force-pushed commits but I can't tell which are different from before and which (if any) are new compared to before.

In general I think using normal merge commits, not force-pushes, avoids such problems. Is there another way to read the diff here?

sbc100:
> Trying to read the changes since my last review, I have a problem that I've seen in @sbc100 's PRs as well with force-pushes... how can I see only the last changes? The last "compare" button shows what appears to be a merge commit from main, and nothing else, and otherwise I see 13 force-pushed commits but I can't tell which are different from before and which (if any) are new compared to before.
> 
> In general I think using normal merge commits, not force-pushes, avoids such problems. Is there another way to read the diff here?

It looks like he did squash the commits so if you want to do an incremental review you could just look at the later commits.   

Obviously its not ideal because (a) you have to trust the earlier commits are not updated and (b) you need to keep track or remember which commit you last reviewed.   But as somehow a very much a rebase workflow person, I'm hesitant to force folks into a merge workflow.

sbc100:
Personally I always like to review the whole change each time, just to remind myself what its doing as a whole.

RReverser:
Hm I don't think I did a squash, just the "update branch with rebase" via Github. Those are annoying too on Github, but the actual commits should be still the same as before.

@kripken this last commit is the only change since last review: https://github.com/emscripten-core/emscripten/pull/24295/commits/54031be14c7a79a33639abffbf877cc0cef36e89

kripken:
Ok, thanks, last commit looks good.

Issue #24260, opened 2025-05-05
Implement a different syntax for contrib ports


ypujante:
In order to stop hijacking the #23432 issue, I am creating a new entry to track this effort.

From @sbc100:

> I've been thinking about how we can make out-of-tree ports work better.
> 
> How about something like this:
> 
> ```
> $ cat tools/ports/contrib/emdawn.py
> EXTERNAL_PORT=http://foo.bar/releases/release.zip
> SHA1=8407a6393c858ad5e7c7f23f7579d524c403be97
> PORT_FILE=myport.py
> ```

ypujante:
@sbc100 I have started investigating these changes.

The fact that ports are loaded when the system starts (and not on demand) makes this quite challenging: fetching the port (= zip file aka artifact) currently happens when the port is being used, the rest is just static data (like dependencies, options, etc...). But with this new design a remote port only contains information to fetch the artifact and the actual port is inside that artifact.

I don't think we want to download the remote artifact (even if it happens only once) if the port is never used just so that we can build the list of ports up front...

At some point (way back when), I had a design where ports would be loaded on demand and that was dropped because the "cost" was deemed negligible. I feel like we need to implement something like that. I am not entirely sure how to merge the 2 concepts otherwise...

Certainly open for guidance/feedback

kainino0x:
It would be very nice to have first class support for things like this, though earlier I think we had a "hack" in mind that would just load the port and effectively re-export it. I've tested this locally by dropping it into `emscripten/tools/ports/contrib` and it works:

```py
# Copyright 2025 The Emscripten Authors.  All rights reserved.
# Emscripten is available under two separate licenses, the MIT license and the
# University of Illinois/NCSA Open Source License.  Both these licenses can be
# found in the LICENSE file.

# Contrib port wrapper for emdawnwebgpu, Dawn's implementation of webgpu.h
# for Emscripten (on top of the WebGPU JS API).
# To use this: --use-port=contrib.emdawnwebgpu-remote

import importlib.util
import os
import sys

URL = 'https://github.com/google/dawn/releases'
DESCRIPTION = "emdawnwebgpu, Dawn's implementation of webgpu.h for Emscripten"
LICENSE = 'mixed licenses, see license files'

_port_name = 'emdawnwebgpu'
_url = 'https://github.com/google/dawn/releases/download/v20250512.200721/emdawnwebgpu_pkg-v20250512.200721.zip'
[...]

PR #24303, opened 2025-05-10
Allow externally hosted contrib ports

ypujante:
This PR is implementing #24260.

The logic that this PR implements is the following:

*  the external port is loaded like the other ports and "flagged" as being external (`port.is_external = hasattr(port, 'EXTERNAL_PORT'`)
* if the port is never needed, then nothing else happens
* if the port is needed (because the user wants to use it via `--use-port=contrib.emdawn`, or it is a dependency of another port), then the code fetches the artifact and uses the port file included in the artifact to load the actual port file as if it was a local port, and replaces it in the ports array and `ports_by_name` map.

=> as a result, all calls using `ports_by_name` have been changed to call `get_port_by_name` to detect whether the port is remote or not and act appropriately

An external port looks like this: it only contains info about where the port is located, the actual logic is coming from `PORT_FILE`

```
TAG = 'v20250509.171557'

EXTERNAL_PORT = f'https://github.com/google/dawn/releases/download/{TAG}/emdawnwebgpu_pkg-{TAG}.zip'
SHA512 = '4b66bf0f64b9616a6420abdad636b3ecefab892dde8f67cd941147bfddf7920f5523ff10160c9a563ef377a0f88b2dfc033527591b2d0753d531de5cbbabde59'
PORT_FILE = 'emdawnwebgpu_pkg/emdawnwebgpu.port.py'

# contrib port information (required)
URL = 'https://dawn.googlesource.com/dawn'
DESCRIPTION = 'Dawn is an open-source and cross-platform implementation of the WebGPU standard'
LICENSE = 'BSD 3-Clause License'
```


ypujante:
The logic I tried to implement is the following:

* the external port is loaded like the other port and "flagged" as being external (`port.is_external = hasattr(port, 'EXTERNAL_PORT')`
* if the port is never needed, then nothing happens
* if the port is needed (because the user wants to use it via `--use-port=contrib.emdawn`, or it is a dependency of another port), then the code fetches the artifact and uses the port file included in the artifact to load the actual port file as if it was a local port, and replaces it in the `ports` array and `ports_by_name` map.

=> as a result, all calls using `ports_by_name` have been changed to call `get_port_by_name` to detect whether the port is remote or not and act appropriately

ypujante:
After investigating the code further, I may have found a solution for the cache issue: I added `cache.ensure` right before calling `cache.withlock` which fixes the cache issue.

Please let me know what you think of theses changes.

I don't understand why `mypy` is failing on something that is **not** changed in this PR

```text
tools/ports/__init__.py:29: error: Need type annotation for "ports_by_name" (hint: "ports_by_name: Dict[<type>, <type>] = ...")  [var-annotated]
Found 1 error in 1 file (checked 131 source files)
```

ypujante:
@sbc100 Note that it is no longer a draft pull request as the implementation is working.

The test added works:

```text
  test_remote_ports (test_other.other) ... ok (0.000s)
```

The test failures don't seem to be related, for example:

```text
FAIL [1.000s]: test_webgl_multi_draw_elements_instanced (test_browser.browser)
  File "/root/project/test/common.py", line 2357, in run_browser
  File "/root/project/test/common.py", line 1651, in assertContained
AssertionError: Expected to find '/report_result?0
' in '/report_result?exception:Cannot read properties of undefined (reading 'GLctx') / TypeError: Cannot read properties of undefined (reading 'GLctx')
    at _emscripten_webgl_enable_WEBGL_multi_draw (http://localhost:8888/test.js:1833:107)
    at http://localhost:8888/test.wasm:wasm-function[31]:0x894
    at http://localhost:8888/test.wasm:wasm-function[32]:0xbeb
    at http://localhost:8888/test.js:896:12
    at callMain (http://localhost:8888/test.js:2687:15)
    at doRun (http://localhost:8888/test.js:2740:24)
    at http://localhost:8888/test.js:2749:7
```

This test passes locally.

Of course the `mypy` test is failing due to missing type annotation but I don't understand why since I did not change that part of the code.

So at this stage I am awaiting your feedback to see if this is the right direction and what you want me to do to move forward with this PR.

I will obviously update documentation/Changelog once you are satisfied with the changes

tools/ports/__init__.py
```diff
@@ -71,6 +88,30 @@ def init_port(name, port):
   validate_port(port)
 
 
+def load_external_port(external_port):
+  need_clear_project_build = Ports.fetch_project(external_port.name, external_port.EXTERNAL_PORT, external_port.SHA512, True)
+  port_file = os.path.join(Ports.get_dir(), external_port.name, external_port.PORT_FILE)
+  module_name = f'tools.ports.external.{external_port.name}'
+  spec = importlib.util.spec_from_file_location(module_name, port_file)
+  local_port = importlib.util.module_from_spec(spec)
+  spec.loader.exec_module(local_port)
```
sbc100, line 97: Some of this code (these 3 lines at least) are shared with `load_port` .. perhaps we can factor out?

tools/ports/__init__.py
```diff
@@ -71,6 +88,30 @@ def init_port(name, port):
...
+  ports.remove(external_port)
+  for a in ['URL', 'DESCRIPTION', 'LICENSE']:
+    if not hasattr(local_port, a):
+      setattr(local_port, a, getattr(external_port, a))
+  init_port(external_port.name, local_port)
+  if need_clear_project_build:
+    Ports.clear_project_build(external_port.name)
```
sbc100, line 104: Why does this need to be delayed like this?

tools/ports/__init__.py
```diff
@@ -343,16 +384,17 @@ def up_to_date():
 
     # before acquiring the lock we have an early out if the port already exists
     if up_to_date():
-      return
+      return False
 
     # main logic. do this under a cache lock, since we don't want multiple jobs to
     # retrieve the same port at once
+    cache.ensure()
```
sbc100, line 391: This seems a little arbitrary to add this here.  Hopefully we can find a better place for it.

sbc100 (commented):
This approach seem reasonable yes.

Can update the PR description with some more details.

tools/ports/__init__.py
ypujante: The problem is that this function (`load_external_port`) calls `Ports.fetch_project` which then can calls`Ports.clear_project_build` which then calls `get_port_by_name` which then calls `load_external_port` again.

Meaning it is a recursive call... 

I ran into this issue after clearing the cache manually. By delaying the call until the end, it fixes this issue.

tools/ports/__init__.py
ypujante: That is where I was blocked and that is how I was able to move forward. If you can tell me where to put that logic which would make more sense I will gladly do it.

tools/ports/__init__.py
ypujante: I will factor it out, no problem.

ypujante: Completed

ypujante:
@sbc100 I am good with landing this internal-only. Once this is released, I can then also turn contrib.glfw as a remote port for another real-life test case and release it publicly with this change.

kainino0x:
> @kainino0x is there a version of Dawn/emdawn that is more "stable" than others or should I just pick the latest (nightly) version?

Latest nightly [v20250514.194204](https://github.com/google/dawn/releases/tag/v20250514.194204) should be the best one we have. It has the most bugfixes.

tools/ports/contrib/emdawn.py
```diff
@@ -0,0 +1,16 @@
...
+TAG = 'v20250509.171557'
+
+EXTERNAL_PORT = f'https://github.com/google/dawn/releases/download/{TAG}/emdawnwebgpu_pkg-{TAG}.zip'
+SHA512 = '4b66bf0f64b9616a6420abdad636b3ecefab892dde8f67cd941147bfddf7920f5523ff10160c9a563ef377a0f88b2dfc033527591b2d0753d531de5cbbabde59'
+PORT_FILE = 'emdawnwebgpu_pkg/emdawnwebgpu.port.py'
+
+# contrib port information (required)
+URL = 'https://dawn.googlesource.com/dawn'
+DESCRIPTION = 'Dawn is an open-source and cross-platform implementation of the WebGPU standard'
+LICENSE = 'BSD 3-Clause License'
```
kainino0x, line 16: Unfortunately the license of the port is mixed. It contains files from both Dawn/webgpu-headers (BSD 3-Clause) and Emscripten (MIT / UI/NCSA dual license)

ypujante: That is fine with me. Can you tell me what I should use?

kainino0x: I don't know how to properly condense that into a string. I guess it doesn't need to be that short.
`"Some files: BSD 3-Clause License. Other files: Emscripten's license (available under both MIT License and University of Illinois/NCSA Open Source License."`

sbc100: Can we just drop these 3 fields for now?

sbc100: Since they exist in the upstream port?

ypujante: They don't exist in the upstream port...

tools/ports/__init__.py
```diff
@@ -71,15 +88,41 @@ def init_port(name, port):
...
+  if not up_to_date:
+    Ports.clear_project_build(name)
+
+
+def init_external_port(name, port):
+  expected_attrs = ['SHA512', 'PORT_FILE', 'URL', 'DESCRIPTION', 'LICENSE']
+  for a in expected_attrs:
+    assert hasattr(port, a), 'port %s is missing %s' % (port, a)
+  port.needed = lambda s: name in ports_needed
+  port.show = lambda: f'{port.name} (--use-port={port.name}; {port.LICENSE})'
```
sbc100, line 117: We could simplify and just drop `; {port.LICENSE})` here?

tools/ports/contrib/emdawn.py
ypujante: And that allows to do `emcc --show-ports` without having to download anything and it would require a code change to make it work... I feel like that is too much for getting rid of a string...

kainino0x: I have LICENSE but not URL/DESCRIPTION. And LICENSE just says "look at the license files" which it can do because they're given out together.

kainino0x: And I think it is fine to include them here. Otherwise we would need to include basically the same info in some other form (comments?) anyway.

ypujante:
@sbc100 @kainino0x Alright, so I just updated the PR with the recommended emdawn version (`v20250514.194204`) and LICENSE string.

As per our prior conversation, let's move forward with pushing this PR as-is (no docs / internal-only). Once this is merged, @kainino0x can deprecate `-sUSE_WEBGPU` and recommend switching to `--use-port=contrib.emdawn`.

Once the next version of Emscripten is released with these changes, I can then work on making `contrib.glfw3` a remote port as well so that we have another test case and then make another PR to update it and make these changes more official (docs , etc..)

sbc100:
> @sbc100 @kainino0x Alright, so I just updated the PR with the recommended emdawn version (`v20250514.194204`) and LICENSE string.
> 
> As per our prior conversation, let's move forward with pushing this PR as-is (no docs / internal-only). Once this is merged, @kainino0x can deprecate `-sUSE_WEBGPU` and recommend switching to `--use-port=contrib.emdawn`.
> 
> Once the next version of Emscripten is released with these changes, I can then work on making `contrib.glfw3` a remote port as well so that we have another test case and then make another PR to update it and make these changes more official (docs , etc..)

I wonder if we can avoid asking folks to make actual releases for repos that just contains a single port (like `contrib.glfw3`)?  Perhaps we can just use the archive URL for a given GIT commit?

sbc100:
> > @sbc100 @kainino0x Alright, so I just updated the PR with the recommended emdawn version (`v20250514.194204`) and LICENSE string.
> > As per our prior conversation, let's move forward with pushing this PR as-is (no docs / internal-only). Once this is merged, @kainino0x can deprecate `-sUSE_WEBGPU` and recommend switching to `--use-port=contrib.emdawn`.
> > Once the next version of Emscripten is released with these changes, I can then work on making `contrib.glfw3` a remote port as well so that we have another test case and then make another PR to update it and make these changes more official (docs , etc..)
> 
> I wonder if we can avoid asking folks to make actual releases for repos that just contains a single port (like `contrib.glfw3`)? Perhaps we can just use the archive URL for a given GIT commit?

Although I do wonder if the SHA first such a URL might change over time.. I think github used to have issues with this but I I also think they might have fixed that.

ypujante:
> I wonder if we can avoid asking folks to make actual releases for repos that just contains a single port (like `contrib.glfw3`)? Perhaps we can just use the archive URL for a given GIT commit?

I honestly do not understand what you mean by this. Can you elaborate?

sbc100:
> > I wonder if we can avoid asking folks to make actual releases for repos that just contains a single port (like `contrib.glfw3`)? Perhaps we can just use the archive URL for a given GIT commit?
> 
> I honestly do not understand what you mean by this. Can you elaborate?

For any tag in any repo you can get a zip archive by creating a URL that looks something like: https://github.com/emscripten-core/emscripten/archive/refs/tags/4.0.8.zip.  I think you can do this for any commit as well but I can't remember the URL syntax.

ypujante:
> For any tag in any repo you can get a zip archive by creating a URL that looks something like: https://github.com/emscripten-core/emscripten/archive/refs/tags/4.0.8.zip. I think you can do this for any commit as well but I can't remember the URL syntax.

I see what you are talking about now. For my project, the link would be

`https://github.com/pongasoft/emscripten-glfw/archive/refs/tags/v3.4.0.20250305.zip`

That being said, this link would download the entire source code of the repo (which at the moment is 216K and contains all the examples and tests) vs the artifact I generate specifically which only contains the necessary files and is 112K `https://github.com/pongasoft/emscripten-glfw/releases/download/v3.4.0.20250305/emscripten-glfw3-3.4.0.20250305.zip`

kleisauke:
> I think you can do this for any commit as well but I can't remember the URL syntax.

The easiest way is to replace `commit` with `tarball` or `zipball` in the URL. For example:
https://github.com/emscripten-core/emscripten/tarball/f224d23c74000cb74f59e4060a88fad8590a5e93
https://github.com/emscripten-core/emscripten/zipball/f224d23c74000cb74f59e4060a88fad8590a5e93

> That being said, this link would download the entire source code of the repo (which at the moment is 216K and contains all the examples and tests) vs the artifact I generate specifically which only contains the necessary files and is 112K

FWIW, you can reduce the size of these archives by marking unnecessary files or directories with the `export-ignore` attribute in the repo's `.gitattributes` file. For example:
```gitattributes
test/ export-ignore
examples/ export-ignore
```

ypujante:
I merged main and the tests reran. There are 2 failures (which were not present before the merge) which don't seem to be related to this PR. So I believe this is ready to be merged.

```
ERROR [0.003s]: test_pthread_dlopen_many (test_core.core0)
  File "/usr/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib/python3.8/unittest/case.py", line 676, in run
  File "/usr/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/root/project/test/test_core.py", line 121, in decorated
    return func(self, *args, **kwargs)
  File "/root/project/test/common.py", line 349, in decorated
    f(self, *args, **kwargs)
  File "/root/project/test/test_core.py", line 9296, in test_pthread_dlopen_many
    self.do_runf('core/pthread/test_pthread_dlopen_many.c',
  File "/root/project/test/common.py", line 1926, in do_runf
    return self._build_and_run(filename, expected_output, **kwargs)
  File "/root/project/test/common.py", line 1971, in _build_and_run
  File "/root/project/test/common.py", line 1555, in run_js
    raise timeout_error
  File "/root/project/test/common.py", line 1527, in run_js
    jsrun.run_js(filename, engine, args,
  File "/root/project/test/jsrun.py", line 103, in run_js
    proc = subprocess.run(
  File "/usr/lib/python3.8/subprocess.py", line 495, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
  File "/usr/lib/python3.8/subprocess.py", line 1028, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib/python3.8/subprocess.py", line 1894, in _communicate
    self.wait(timeout=self._remaining_time(endtime))
  File "/usr/lib/python3.8/subprocess.py", line 1083, in wait
    return self._wait(timeout=timeout)
  File "/usr/lib/python3.8/subprocess.py", line 1798, in _wait
    raise TimeoutExpired(self.args, timeout)
subprocess.TimeoutExpired: Command '['/root/emsdk/node/20.18.0_64bit/bin/node', '--stack-trace-limit=50', '--trace-uncaught', '/tmp/emtest_h9ywpgy3/emscripten_test_core0_bcxwcaqa/test_pthread_dlopen_many.js']' timed out after 299.9999436500002 seconds

FAIL [0.002s]: test_async_hello_v8 (test_core.instance)
  File "/usr/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib/python3.8/unittest/case.py", line 676, in run
  File "/usr/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/root/project/test/common.py", line 316, in decorated
    return func(self, *args, **kwargs)
  File "/root/project/test/test_core.py", line 8073, in test_async_hello_v8
    self.test_async_hello()
  File "/root/project/test/common.py", line 895, in resulting_test
    return func(self, *args)
  File "/root/project/test/test_core.py", line 197, in metafunc
    f(self, *args, **kwargs)
  File "/root/project/test/test_core.py", line 8054, in test_async_hello
    self.do_runf('main.c', 'HelloWorld!99')
  File "/root/project/test/common.py", line 1926, in do_runf
    return self._build_and_run(filename, expected_output, **kwargs)
  File "/root/project/test/common.py", line 1955, in _build_and_run
    js_file = self.build(filename, **kwargs)
  File "/root/project/test/common.py", line 1414, in build
    self.run_process(cmd, stderr=self.stderr_redirect if not DEBUG else None)
  File "/root/project/test/common.py", line 1757, in run_process
    self.fail(f'subprocess exited with non-zero return code({e.returncode}): `{shlex.join(cmd)}`')
AssertionError: subprocess exited with non-zero return code(1): `/root/project/emcc main.c -o main.mjs -sNO_DEFAULT_TO_CXX -sMODULARIZE=instance -sASYNCIFY -sEXIT_RUNTIME -Wclosure -Werror -Wno-limited-postlink-optimizations -Wno-experimental -Wno-unused-command-line-argument -sENVIRONMENT=shell````

**Comment by ypujante** [comment]
@kainino0x I don't think it is a showstopper for releasing these changes, but version v20250514.194204 generates an issue with the closure compiler:

```
// rebuilding my example: see https://github.com/pongasoft/emscripten-ports/tree/master/examples/Dawn
// /usr/local/emsdk/emscripten/main/emcc -sASYNCIFY=1 --closure=1 -O2 --use-port=contrib.emdawn main.cpp -o /tmp/dawn/index.html

building:ERROR: /var/folders/wz/zqhny4ks031bqlr6r27bfkzm0000gn/T/emscripten_temp_48uj9d0y/index.jso1.js:1152:17: ERROR - [JSC_UNDEFINED_VARIABLE] variable stringToNewUTF8 is undeclared
  1152|     var strPtr = stringToNewUTF8(strs);
                         ^^^^^^^^^^^^^^^

1 error(s), 0 warning(s)
```

**Comment by ypujante** [comment]
This is the snippet of code that triggers the error:

```js
// line 579 | library_webgpu.js
...
      }
    },

    fillAdapterInfoStruct: (info, infoStruct) => {
      {{{ gpu.makeCheckDescriptor('infoStruct') }}}

      // Populate subgroup limits.
      {{{ makeSetValue('infoStruct', C_STRUCTS.WGPUAdapterInfo.subgroupMinSize, 'info.subgroupMinSize', 'i32') }}};
      {{{ makeSetValue('infoStruct', C_STRUCTS.WGPUAdapterInfo.subgroupMaxSize, 'info.subgroupMaxSize', 'i32') }}};

      // Append all the strings together to condense into a single malloc.
      var strs = info.vendor + info.architecture + info.device + info.description;
      var strPtr = stringToNewUTF8(strs);
...
```

it is definitely missing the `fillAdapterInfoStruct__deps` section

kainino0x:
Oops, that was a recent regression (from [here](https://dawn-review.googlesource.com/c/dawn/+/242274)). We haven't quite got automated testing up and running yet. Will fix.

ypujante:
@kainino0x it's funny because it is my own [CI](https://github.com/pongasoft/emscripten-ports/blob/785c94fbd88cc3958213aecf8917e2e78b13d5c6/.github/workflows/main.yml#L51) that caught the problem :)

kainino0x:
Heh, any build is better than none! I just added a basic compile test but it's not running anywhere, I should add that to the github release so it does a last check before releasing.

Should be fixed today with https://dawn-review.googlesource.com/c/dawn/+/242734. Thanks for the report.

kainino0x:
Fixed release here. https://github.com/google/dawn/releases/tag/v20250516.124039

tools/ports/contrib/emdawn.py
kainino0x, line 1: I would prefer to call this `emdawnwebgpu` for consistency with the project name (the reason it has that weird name is to make it very easily keyword searchable)

ypujante:
* I bumped the version to v20250516.124039 which fixes the closure compiler issue I had noticed
* I renamed the port per @kainino0x request

The tests are passing minus one which seems to be irrelevant `ERROR: could not find the following tests: esm-integration`

ypujante:
I merged with main yesterday and there was no test failures.

I merged again this morning and there are 2 test failures which are unrelated to these changes:

```
FAIL [0.005s]: test_dyncall_pointers_legacy (test_core.core3)

ERROR [0.005s]: test_fcntl_open_rawfs (test_core.esm_integration)
```

@sbc100 Is there anything else that I need to do for this PR? Just as an FYI, I will be traveling in a couple of days so I will be less available for changes. So please let me know if there is anything that need to be changed.

sbc100:
Both of those failures are known issues.   The former has now been fixed and the later is just a flake.

sbc100 (approved):

test/test_other.py
```diff
@@ -2592,6 +2592,10 @@ def test_contrib_ports(self):
     # with a different contrib port when there is another one
     self.emcc(test_file('other/test_contrib_ports.cpp'), ['--use-port=contrib.glfw3'])
 
+  @requires_network
+  def test_remote_ports(self):
+    self.emcc(test_file('hello_world.c'), ['--use-port=contrib.emdawnwebgpu'])
```
sbc100, line 2597: @kainino0x the name `emdawnwebgpu` seems rather long and redundant to me.

How about simple `emdawn` or even just `dawn`.   The `em` prefix and the `webgpu` suffix both seem redundant in this context but maybe I'm missing something?

In any case we can land this and bikeshed later.

ypujante: FWIW,  I agree 100% with @sbc100. I had named it originally `dawn` because it is the name of the project and it is following the pattern of all the other ports (`sdl2`, `glfw3`, etc... which are not named `emsdl2`, `emglfw3`). I only renamed it because @kainino0x asked me to.

kainino0x:
🎉

test/test_other.py
kainino0x: Emdawnwebgpu is the name of the project, it's used for the package releases and in the Dawn source tree. I chose it to be easily searchable and unambiguous (from USE_WEBGPU, any other potential/future ways of using Dawn in Wasm, Tint in Wasm, etc.)

sbc100: Fair enough!

Also, we could drip the `contrib.` here if you like since we do care internally about maintain this port if good working order.   When we could just have `--use-port=emdawnwebg`

kainino0x: If the distinction is how much we maintain it, then making it a non-contrib port makes sense to me. Will probably roll that change into #24220.

sbc100: Yeah, kind of.  Its not super well defined.  Basically I wanted to allow folks to create new ports in emscripten without adding more work for emscripten developers to make sure they keep working in all out different build configurations.  I think we webgpu is key enough that we should consider it core and make sure it works over time.

That said, we should at some point add a few more emscripten-side tests to ensure we don't break things on our end.

PR #24306, opened 2025-05-12
Install SDL2/SDL3 cmake config only when installing the ports

sbc100:
The cmake config files are logically part of the ports so install them then the port is installed

This doesn't fix #19780 completely, since trying to build an SDL2 or SDL3 cmake project will still require some kind of install step, but it makes it less confusing.


sbc100:
I'll probably land https://github.com/emscripten-core/emscripten/pull/24308 first we we can move these files alongside the config headers.

tools/system_libs.py
```diff
@@ -2493,10 +2493,6 @@ def install_system_headers(stamp):
   bin_dest = cache.get_sysroot_dir('bin')
   copytree_exist_ok(bin_src, bin_dest)
 
-  cmake_src = utils.path_from_root('system/lib/cmake')
-  cmake_dest = cache.get_sysroot_dir('lib/cmake')
-  copytree_exist_ok(cmake_src, cmake_dest)
```
kripken, line 2498: Why is this no longer needed?

sbc100: This directory contains on the SDL2 and SDL3 cmake files.  They are now installed when the respective port is installed instead.

kripken: This is in generic code, though, not SDL - does this not affect any other port?

sbc100:
Ok, now that we are storing these files alongside the ports this change might make more sense.

kripken (approved):

PR #24309, opened 2025-05-12
Test and fix for incompatible HEAP transforms in acorn-optimizer

RReverser:
While working on #24295, I noticed I could accidentally prevent ASan and SAFE_HEAP from finding any `HEAPxx[n]` accesses from JS when `ALLOW_MEMORY_GROWTH` is enabled, and no tests would catch it, because no tests verify that acorn-optimizer passes which change `HEAPxx[n]` into something else can work in conjunction with each other.

After adding the test, I found that an existing `SUPPORT_BIG_ENDIAN` feature was already broken in the same way, because it transforms `HEAPxx[n]` accesses into DataView methods, and since it was running before ASan and SAFE_HEAP, those latter transforms became no-ops. I fixed that in the same PR by moving the big-endian transform to the end.

In the future we might want to consolidate those transforms in a more reliable way (e.g. only run a single code transform that does `HEAPxx[n]` -> `__loadHeap`/`__storeHeap` where those functions do all the work using `#if`s), but at least for now this PR fixes an immediate problem and catches potential future regressions.


RReverser:
Actually, if I add `-pthread`, there are even more combos that currently unexpectedly succeed... changing into draft until I fix them too.

RReverser:
Ah yeah, looks like `ALLOW_MEMORY_GROWTH` doesn't work correctly with ASan / SAFE_HEAP for the same reasons even before my change in #24295.

test/test_other.py
```diff
@@ -12226,6 +12226,28 @@ def test_asan_strncpy(self):
...
+  })
+  def test_null_deref_via_js(self, expected_output, *args):
+    # Multiple JS transforms look for pattern like `HEAPxx[...]` and transform it.
+    # This test ensures that one of the transforms doesn't produce a pattern that
+    # another pass can't find anymore, that is that features can work in conjunction.
+    self.do_runf(
+      'other/test_null_deref_via_js.c',
+      emcc_args=args,
+      assert_returncode=NON_ZERO,
+      expected_output=[expected_output])
```
kripken, line 12249: Perhaps a single test with all 3 features enabled is enough?

kripken:
I guess it's not surprising some of these combinations don't work, given the possible interactions, and that we don't have exhaustive testing...

I'd say that only memory growth is a high priority, of the three (asan is run locally, and devs can work around feature combinations, and big-endian likely has very few users).

test/test_other.py
RReverser: It would be enough to reproduce the issue, but I found that having separate tests helps to narrow down which particular transform is at fault.

RReverser:
> I'd say that only memory growth is a high priority

Yeah I agree, combo of memory growth + ASan is the first one I noticed and then thought it's worth checking other combinations too.

I'd like to merge https://github.com/emscripten-core/emscripten/pull/24291 + https://github.com/emscripten-core/emscripten/pull/24295 + one more upcoming PR for simplifying ASan first, and then I'll fix these combo interactions as it will be somewhat easier to do on top of simpler transforms.

RReverser:
Ok after all the simplifications in acorn-optimizer transforms the fix here was a trivial reordering of passes, so I added it to the same PR.

This PR should be ready for review now.

test/other/test_null_deref_via_js.c
```diff
@@ -0,0 +1,7 @@
+#include <emscripten.h>
+
+int main() {
+  EM_ASM({
+    HEAP32[0] = 0;
```
sbc100, line 5: How about `return HEAP32[0]`  since that would be more like a null deref?

Also, do we not have tests for this already elsewhere?

sbc100 (approved):

test/other/test_null_deref_via_js.c
RReverser: > How about `return HEAP32[0]` since that would be more like a null deref?

Well one is load, another is store, but both result in null deref. Either is fine by me.

> Also, do we not have tests for this already elsewhere?

Apparently not because nothing caught this before.

RReverser:
Okay looks like reordering introduces its own edge cases. I'll take another look.

test/other/test_null_deref_via_js.c
sbc100: lgtm as is then

RReverser: Ah found test_safe_heap_user_js.c that I can reuse.

src/lib/libpthread.js
```diff
@@ -978,7 +978,7 @@ var LibraryPThread = {
 
   $establishStackSpace__internal: true,
   $establishStackSpace__deps: ['$stackRestore', 'emscripten_stack_set_limits'],
-  $establishStackSpace: (pthread_ptr) => {
+  $establishStackSpace: function (pthread_ptr) {
```
kripken, line 981: why did this  change?

tools/acorn-optimizer.mjs
```diff
@@ -1372,7 +1372,10 @@ function isHEAPAccess(node) {
 function asanify(ast) {
   recursiveWalk(ast, {
     FunctionDeclaration(node, c) {
-      if (node.id.type === 'Identifier' && node.id.name.startsWith('_asan_js_')) {
+      if (
+        node.id.type === 'Identifier' &&
+        (node.id.name.startsWith('_asan_js_') || node.id.name === 'establishStackSpace')
```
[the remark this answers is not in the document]
RReverser: Without fixes in this PR some heap patterns (in particular, in memory growth + asan tests) were never asanified.

That's what the PR was testing & fixing after all, and once fixed, it started hitting new code paths.

src/lib/libpthread.js
RReverser: It was the easy way to add another `node.id.name` check to skip the `establishStackSpace` helper too, as we skip specific function declarations.

I could instead add a match to find variable declarator with a specific name and with arrow function expression initializer, but it would look somewhat more complex.

PR #24314, opened 2025-05-12
Simplify ASan JS transform

RReverser:
Leverages ASan's C API for arbitrary loads/stores instead of custom C code with separate variants for each possible type.

Similar to https://github.com/emscripten-core/emscripten/pull/24295 and https://github.com/emscripten-core/emscripten/pull/24291.


src/runtime_asan.js
```diff
@@ -8,111 +8,22 @@
...
-  return HEAPF32[ptr];
-}
-/** @suppress{duplicate} */
-function _asan_js_load_d(ptr) {
-  if (runtimeInitialized) return __asan_c_load_d(ptr);
-  return HEAPF64[ptr];
+function _asan_js_load(arr, index) {
+  if (runtimeInitialized) {
+    const elemSize = arr.BYTES_PER_ELEMENT;
+    ___asan_loadN(index * elemSize, elemSize);
```
kripken, line 18:
```suggestion
    return ___asan_loadN(index * elemSize, elemSize);
```

src/runtime_asan.js
```diff
@@ -8,111 +8,22 @@
...
+  }
```
kripken, line 19: How does this handle signed/unsigned differences?

src/runtime_asan.js
RReverser: It's a void function.

src/runtime_asan.js
RReverser: See above, asan_loadN is a `void` function that only needs to check if pointer is valid and aborts if not. The beauty of this approach is that it doesn't need to care about signed/unsigned differences - instead, we just retrieve value with correct sign by retrieving it from the provided array (`HEAP32` / `HEAPU32` / whatever).

kripken: Oh nice, now I see. Yes, very simple.

kripken (approved):
Nice!

tools/link.py
```diff
@@ -1608,7 +1608,7 @@ def limit_incoming_module_api():
       # do anything (as the user's list won't contain these functions), and if
       # we did add them, the pass would assert on incompatible lists, hence the
       # condition in the above if.
-      settings.ASYNCIFY_REMOVE += emscripten.ASAN_C_HELPERS
+      settings.ASYNCIFY_REMOVE.append("__asan_*")
```
sbc100, line 1611: Is this change needed?

RReverser: Yeah, ASan's C API calls into other C APIs and then I was getting test failures from Asyncify trying to transform more of ASan internals.

It's easier and more reliable to cover all of ASan APIs at once.

sbc100: sgtm

brendandahl:
This appears to be causing a new failure on the [rollers](https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket/8715012528599384993/+/u/Emscripten_testsuite__ASan_/stdout) `FAIL: test_stack (test_core.asan)`

RReverser:
Sigh. I wish regular CI would cover all those checks. I'll look into it.

sbc100:
I found a fix: https://github.com/emscripten-core/emscripten/pull/24317

sbc100:
I only verified that one test.   Which other ones are failing?

sbc100:
Yup looks like it fixes all of them.  The common theme was they all used `-sSTACK_OVERFLOW_CHECK=2` (or `-sASSERTIONS=2` which implies it)

PR #24316, opened 2025-05-12
[jspi] Fix returning pointers from async dyncalls.

### LINKED ISSUE #24307: JSPI: Unable to return emscripten::val from an embind function when CAN_ADDRESS_2GB is true
This bug is due to dynCaller thinking that the returned promise is a pointer in https://github.com/emscripten-core/emscripten/blob/main/src/lib/libcore.js#L1815C1-L1828C7. It shifts the promise which makes it return 0 instead of a promise.

```js
#if JSPI
    if (promising) {
      func = WebAssembly.promising(func);
    }
#endif
    var rtn = func(...args);
#endif
#if MEMORY64
    return sig[0] == 'p' ? Number(rtn) : rtn;
#elif CAN_ADDRESS_2GB
    return sig[0] == 'p' ? rtn >>> 0 : rtn;
#else
    return rtn;
#endif
```

I'm unsure how it should handle promises.

```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.8 (70404efec4458b60b953bc8f1529f2fa112cdfd1)
clang version 21.0.0git (https:/github.com/llvm/llvm-project 23e3cbb2e82b62586266116c8ab77ce68e412cf8)
Thread model: posix
```

This snippet uses fetch to show the issue.

# fetchBinding.cpp
```cpp
#include <emscripten>
#include <emscripten/bind.h>
#include <emscripten/val.h>

namespace
{
emscripten::val fetch(const char* url) {
  return emscripten::val::global("fetch")(std::string(url));
}
}

EMSCRIPTEN_BINDINGS(fetch_bindings)
{
  emscripten::function("fetchAsync", &fetch, emscripten::async());
}
```

# testFetchAsync.js
```js
let fetchAsyncModule = await import("./fetchAsync.mjs");
let wasmModule = await fetchAsyncModule.default();
let resp = await wasmModule.fetchAsync("https://vtk.org/files/ExternalData/SHA512/477f47ed5f941950bd8f0f2f5a37516572184bf5c218b105443c45104591cc6edaa105b2ee7dddd92ad51e40318feaaf0eaf272d5b0e904c722cb93f510fff9b");
console.log(await resp.json()); // should print json
```

what works:
```sh
emcc fetchBinding.cpp -sJSPI -sALLOW_MEMORY_GROWTH=1 -lembind -o fetchAsync.mjs
node --experimental-wasm-jspi testFetchAsync.js
```

what fails:
```sh
emcc fetchBinding.cpp -sJSPI -sALLOW_MEMORY_GROWTH=1 -sMAXIMUM_MEMORY=4GB -lembind -o fetchAsync.mjs
node --experimental-wasm-jspi testFetchAsync.js
<anonymous_script>:13
return rv.then(onDone);}
          ^

TypeError: rv.then is not a function
    at Object.fetchAsync (eval at craftInvokerFunction (file:///C:/Users/jaswant.panchumarti/dev/fetchAsync.mjs:1284:19), <anonymous>:13:11)
    at file:///C:/Users/jaswant.panchumarti/dev/main.js:3:29

Node.js v24.0.1
```

brendandahl:
When dyncall calls an async function it should wait for the result of the promise to then convert the pointer.

Fixes #24307

### DIFF
--- a/src/lib/libcore.js
+++ b/src/lib/libcore.js
@@ -1818,14 +1818,23 @@ addToLibrary({
     }
 #endif
     var rtn = func(...args);
-#endif
+    function convert(rtn) {
 #if MEMORY64
-    return sig[0] == 'p' ? Number(rtn) : rtn;
+      return sig[0] == 'p' ? Number(rtn) : rtn;
 #elif CAN_ADDRESS_2GB
-    return sig[0] == 'p' ? rtn >>> 0 : rtn;
+      return sig[0] == 'p' ? rtn >>> 0 : rtn;
 #else
-    return rtn;
+      return rtn;
 #endif
+    }
+#endif
+
+#if JSPI
+    if (promising) {
+      return rtn.then(convert);
+    }
+#endif
+    return convert(rtn);
   },
 
   $callRuntimeCallbacks__internal: true,
--- a/src/lib/libcore.js
+++ b/src/lib/libcore.js
@@ -1818,6 +1818,8 @@ addToLibrary({
     }
 #endif
     var rtn = func(...args);
+#endif // DYNCALLS
+
     function convert(rtn) {
 #if MEMORY64
       return sig[0] == 'p' ? Number(rtn) : rtn;
@@ -1827,7 +1829,6 @@ addToLibrary({
       return rtn;
 #endif
     }
-#endif
 
 #if JSPI
     if (promising) {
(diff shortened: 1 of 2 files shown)


sbc100 (approved):

jspanchu:
Thanks for working on it! Will this fix go into 4.0.9?

sbc100:
Yup, it will be in 4.0.9.   Emscripten releases are always cut from main so all changes that land on main are by definition part of the next release.

jspanchu:
awesome

PR #24320, opened 2025-05-13
Default to `a.out.mjs` in EXPORT_ES6 mode

sbc100:


ChangeLog.md
```diff
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 4.0.9 (in development)
 ----------------------
+- When building with `-sEXPORT_ES6` the default name of the output of the
+  compiler is now `a.out.mjs` rather than `a.out.js`.  This only applies in the
+  absence of an explicit `-o` flag. (#24320)
```
kripken, line 25: What are the benefits of this breaking change?

Though if we don't think many people are using the flag, maybe the breakage is minimal?

sbc100: I can't imagine many folks being effected.  How many people (other than our test suite) are not using a `-o` flag?   I would imagine very few.  Still fewer are going to be using `-sEXPORT_ES6`.

I had a whole class of tests that in the test suite that were fixed by this... but I can't remember now exaclyt which ones.  Let me do a little more investigation.

We could also limit this to just `-sWASM_ESM_INTEGRATION` perhaps?

brendandahl: Not as clean, but one alternative is we could emit a package.json with 
```json
{
  "type": "module"
}
```

sbc100: Yeah, but that is kind of gross... and it effects all other files in the same directory and below.

kripken: Seems like a reasonable breaking change. My one concern is maybe automatic tests, with stuff like `emcc test.cpp && nodejs a.out.js`, so it compiles and runs using the assumed default name. Some of those might use EXPORT_ES6, but maybe it's rare.

sbc100: But `emcc test.cpp && nodejs a.out.js` already never works in `EXPORT_ES6` mode because it implies `MODULARIZE` which means you need to write a wrapper around it and instantiate to get it to run.

`-sMODULARIZE=instance` is different and might work but that has not been announced and is still experimental.

Anyway I think I found and alternative solution: https://github.com/emscripten-core/emscripten/pull/24321

sbc100:
Closing in favor of https://github.com/emscripten-core/emscripten/pull/24321

PR #24321, opened 2025-05-13
[esm-integration] Allow modules to be runnable even without `main`

sbc100:
Also, ensure the output format defaults to `MJS` in esm integration and modularize=instance modes.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -749,6 +749,11 @@ def phase_linker_setup(options, linker_args):  # noqa: C901, PLR0912, PLR0915
     if s in user_settings:
       diagnostics.warning('deprecated', f'{s} is deprecated ({reason}). Please open a bug if you have a continuing need for this setting')
 
+  # Set the EXPORT_ES6 default early since it affects the setting of the
+  # default oformat below.
+  if settings.WASM_ESM_INTEGRATION or settings.MODULARIZE == 'instance':
+    default_setting('EXPORT_ES6', 1)
+
   # If no output format was specified we try to deduce the format based on
   # the output filename extension
   if not options.oformat and (options.relocatable or (options.shared and not settings.SIDE_MODULE)):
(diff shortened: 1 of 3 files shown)


tools/link.py
```diff
@@ -749,6 +749,11 @@ def phase_linker_setup(options, linker_args):  # noqa: C901, PLR0912, PLR0915
     if s in user_settings:
       diagnostics.warning('deprecated', f'{s} is deprecated ({reason}). Please open a bug if you have a continuing need for this setting')
 
+  # Set the EXPORT_ES6 default early since it effects the setting of the
```
kripken, line 752:
```suggestion
  # Set the EXPORT_ES6 default early since it affects the setting of the
```
affects/effects is truly one of the most bizarre things in English... each has multiple interweaving meanings with the other

tools/link.py
```diff
@@ -786,10 +791,11 @@ def phase_linker_setup(options, linker_args):  # noqa: C901, PLR0912, PLR0915
 
   if settings.WASM_ESM_INTEGRATION:
     diagnostics.warning('experimental', '-sWASM_ESM_INTEGRATION is still experimental and not yet supported in browsers')
-    default_setting('EXPORT_ES6', 1)
     default_setting('MODULARIZE', 'instance')
-    if not settings.EXPORT_ES6 or settings.MODULARIZE != 'instance':
-      exit_with_error('WASM_ESM_INTEGRATION requires EXPORT_ES6 and MODULARIZE=instance')
+    if options.oformat != OFormat.MJS:
+      exit_with_error('WASM_ESM_INTEGRATION is only compatible with .mjs output files')
```
kripken, line 796: Is this a new restriction?

kripken (approved):

tools/link.py
sbc100: Its not new I'm just trying to be consistent with the same error message for `-sMODULARIZE=instance` below.  I'm rephrased both of them.

PR #24322, opened 2025-05-13
[embind] Initialize embind on workers after main thread ctors.

### LINKED ISSUE #19849: embind cannot be initialized correctly in a thread created in static constructor
I found this bug because I created a thread in `wasmfs_before_preload()`, which has `__attribute__((init_priority(100)))`.

When the thread runs, embind has not been fully initialized in main thread. So `Module['__embind_initialize_bindings']` did not register the types for me, but set `initializedJS` to true.

I did not use embind in thread created in `wasmfs_before_preload`, but the worker is reused in another thread, and in that thread embind does not work, caused a crash in my app.

I hope we can have a way to ensure `wasm_init_ctors` are completed, or embind inited, when `spawnthread` get called.

brendandahl:
Previously, when a statically constructed thread was created it could potentially run before all of the embind static constructors on the main thread had run. This caused the workers to miss the embind bindings.

To fix this, I've added a `std::atomic` so that each worker waits for embind to be ready before running the binding code.

Fixes #19849

### DIFF
--- a/src/lib/libembind.js
+++ b/src/lib/libembind.js
@@ -31,6 +31,10 @@ var LibraryEmbind = {
   $UnboundTypeError: class extends Error {},
   $PureVirtualError: class extends Error {},
   $GenericWireTypeSize: {{{ 2 * POINTER_SIZE }}},
+#if PTHREADS || WASM_WORKERS
+  $postCtors: null, // Unused symbol stub for injecting post ctors call.
+  $postCtors__postset: () => { addAtPostCtor('__embind_post_ctors()'); },
+#endif
 #if EMBIND_AOT
   $InvokerFunctions: '<<< EMBIND_AOT_INVOKERS >>>',
 #endif
@@ -2319,5 +2323,9 @@ var LibraryEmbind = {
     });
   },
 };
+
+#if PTHREADS || WASM_WORKERS
+extraLibraryFuncs.push('$postCtors');
+#endif
 
 addToLibrary(LibraryEmbind);
--- a/src/lib/libwasm_worker.js
+++ b/src/lib/libwasm_worker.js
@@ -138,7 +138,7 @@ addToLibrary({
 
 #if EMBIND
     // Embind must initialize itself on all threads, as it generates support JS.
-    __embind_initialize_bindings();
+    __embind_initialize_worker_bindings();
 #endif
 
 #if AUDIO_WORKLET
--- a/src/runtime_pthread.js
+++ b/src/runtime_pthread.js
@@ -147,7 +147,7 @@ if (ENVIRONMENT_IS_PTHREAD) {
 #endif
           // Embind must initialize itself on all threads, as it generates support JS.
           // We only do this once per worker since they get reused
-          __embind_initialize_bindings();
+          __embind_initialize_worker_bindings();
 #endif // EMBIND
           initializedJS = true;
         }
--- a/system/lib/embind/bind.cpp
+++ b/system/lib/embind/bind.cpp
@@ -19,6 +19,8 @@
 using namespace emscripten;
 using namespace internal;
 
+static std::atomic<bool> embind_initialized(false);
+
 extern "C" {
 
 const char* __getTypeName(const std::type_info* ti) {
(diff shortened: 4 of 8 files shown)


src/lib/libembind.js
```diff
@@ -31,6 +31,10 @@ var LibraryEmbind = {
   $UnboundTypeError: class extends Error {},
   $PureVirtualError: class extends Error {},
   $GenericWireTypeSize: {{{ 2 * POINTER_SIZE }}},
+#if PTHREADS || WASM_WORKERS
+  $postCtors: null, // Unused symbol stub for injecting post ctors call.
+  $postCtors__postset: () => { addAtPostCtor('__embind_post_ctors()'); },
+#endif
```
sbc100, line 37: Can't we just make this is `__postset` of some function that a core embind dependnecy?   Adding this new symbol and forcing it to be included seems a more complex than it needs to be.

sbc100 (approved):
lgtm % comment

kripken:
Did the priority option not work out? Either

1. Adjust the embind constructor priority to be before all user constructors. You said this might not work since it needs user stuff to run in some cases...
2. Document the issue and tell users to adjust the priority of their own constructors to be later. This seems safer to me than worker waiting, would could deadlock IIUC?

sbc100:
> Did the priority option not work out? Either
> 
> 1. Adjust the embind constructor priority to be before all user constructors. You said this might not work since it needs user stuff to run in some cases...

This doesn't work, at least last time we tried it it didn't work because there are (apparently) cases there the EMSCIRPTEN_BINDINGS block must be run *after* the static initializers that precede it in the same file.

Obviously we can look into those cases and see if we can change things around, but last time we tried to do this it got reverted.

> 2. Document the issue and tell users to adjust the priority of their own constructors to be later. This seems safer to me than workers waiting on the main thread, which could deadlock IIUC?

This doesn't work because unfortunately the default for C++ constructors is the highest possible priority so its not possible to make things to run after default C++ constructors.

sbc100:
I think the possible deadlock risk here is very low because it would mean writing code that both creates a thread and expects to be up and running and interact-able-with all before main even runs.   Doing that kind of thing in static constructors goes against most folks C/C++ sensibilities I think.

We could emit a warning in debug builds when thread end up waiting here?

brendandahl:
> Did the priority option not work out? Either
> 
> 1. Adjust the embind constructor priority to be before all user constructors. You said this might not work since it needs user stuff to run in some cases...

The following is supported in embind:
```c++
const std::string global_string = "global string";

EMSCRIPTEN_BINDINGS(constants) {
  emscripten::constant("global_string", global_string);
}
```

Which doesn't work if we run the binding code before user's static constructors. I think this is a relatively recent change (~2yrs), so we could potentially disallow it or tell users that they need to set the priority on the `global_string` to be higher than the embind code.

kripken:
I see, thanks @sbc100 @brendandahl 

This PR sgtm then. Is there a way we can improve debugging here? E.g. in a debug build, perhaps show a warning when we end up waiting for that lock?

sbc100:
> I see, thanks @sbc100 @brendandahl
> 
> This PR sgtm then. Is there a way we can improve debugging here? E.g. in a debug build, perhaps show a warning when we end up waiting for that lock?

Yes that seem reasonable.

brendandahl:
A further thought I had on dropping support for binding static properties. C++20 supports `constexpr std::string`, so maybe in the future we could drop support for binding static strings, but only allow constexpr strings (or other objects).  `constexpr` strings should be defined before any static initializers are run. We could then move the embind code back to running before static constructors and get rid of this atomic.

sbc100:
> A further thought I had on dropping support for binding static properties. C++20 supports `constexpr std::string`, so maybe in the future we could drop support for binding static strings, but only allow constexpr strings (or other objects). `constexpr` strings should be defined before any static initializers are run. We could then move the embind code back to running before static constructors and get rid of this atomic.

That sounds great!

sbc100:
Would it be worth still landing this anyway, just so we can be sure we get the order right?  Perhaps it could just be an assertion in debug builds?

PR #24328, opened 2025-05-14
[MODULARIZE=instance] Allow Module argument in `--pre-js` code.

sbc100:
Previously we were clobbering any existing Module arguments with the arguments passed to the `init()` function.

Also, don't limit INCOMING_MODULE_JS_API in MODULARIZE=instance mode. At least for now I think we should go for compatibility by default.

The test I've enabled here sets the `dynamicLibraries` Module property during in a pre-js script.

### DIFF
--- a/src/postamble.js
+++ b/src/postamble.js
@@ -303,7 +303,7 @@ function preInit() {
 // In MODULARIZE=instance mode we delay most of the initialization work until
 // the `init` function is called.
 export default async function init(moduleArg = {}) {
-  Module = moduleArg;
+  Object.assign(Module, moduleArg);
   processModuleArgs();
 #if WASM_ESM_INTEGRATION
   updateMemoryViews();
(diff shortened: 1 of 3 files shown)


src/postamble.js
```diff
@@ -303,7 +303,7 @@ function preInit() {
 // In MODULARIZE=instance mode we delay most of the initialization work until
 // the `init` function is called.
 export default async function init(moduleArg = {}) {
-  Module = moduleArg;
+  Object.assign(Module, moduleArg);
```
brendandahl, line 306: I was really hoping we could move towards separating the arguments from the `Module` object.  I missed the change where we started assigning `var Module = {};`  in instance mode, but having it undefined helped point out places that were reading from the Module object before everything was setup.

sbc100: yeah I agree, but we have so much go that expects to be able to `Module[xxx]` in `--pre-js` code.  We would have to update all those tests if want to drop that.

I suppose we would make it conditional and only do this if there exists some pre-js code?

brendandahl (approved):

PR #24332, opened 2025-05-15
[ASYNCIFY] Add some assertions. NFC

sbc100:


kripken (approved):

sbc100:
@kripken what do you think about the assertion firing in `core0.test_asyncify_side_module_reversed`: `Aborted(Assertion failed: export not found: main)` .. is that OK?  Should we fail there?  Should we fix the test somehow?

kripken:
Hmm, that sounds like a bug... why is it trying to create a rewind function for a non-existing export? I'd expect it to create them when it actually needs them.

Perhaps somewhere in the code has a hardcoded assumption about main always existing?

sbc100:
Ah, I had the assertion in slightly the wrong place.  All good now.

PR #24335, opened 2025-05-15
Fix UTF16ToString in TEXTDECODER=2 and MAXIMUM_MEMORY>=2GB

RReverser:
I accidentally noticed that Embind UTF-16 support was producing incorrect results with MAXIMUM_MEMORY over 2GB.

Turns out, the `TextDecoder` implementation of `UTF16ToString` itself was broken in this mode due to bitwise ops, so I've added new tests and fixed this bug, slightly simplifying code in the process.


RReverser:
What's up with Windows CI? Seems to be failing in download stage on several PRs.

test/test_core.py
```diff
@@ -335,6 +335,19 @@ def decorated(self, *args, **kwargs):
   return outer_decorator
 
 
+def maybe_with_textdecoder(f):
```
sbc100, line 338: How about `with_both_text_decoder`

src/lib/libstrings.js
```diff
@@ -322,23 +322,22 @@ addToLibrary({
...
+    var endIdx = idx;
     // If maxBytesToRead is not passed explicitly, it will be undefined, and this
     // will always evaluate to true. This saves on code size.
-    while (!(idx >= maxIdx) && HEAPU16[idx]) ++idx;
-    endPtr = idx << 1;
+    while (!(endIdx >= maxIdx) && HEAPU16[endIdx]) ++endIdx;
 
 #if TEXTDECODER != 2
-    if (endPtr - ptr > 32 && UTF16Decoder)
+    if (endIdx - idx > 16 && UTF16Decoder)
```
sbc100, line 338: Is the 16 to 32 change here because the diff here is between two 16-bit indexes now, but was between bytes index before?

Can we add a comment here?  Something like: `For short (<16 code points) strings the call to `UTF16Decoder.decode is not worth it?` ... assuming that is what is going on here?

sbc100 (approved):

sbc100:
> What's up with Windows CI? Seems to be failing in download stage on several PRs.

Looks like infra flake with the chocolaty package manager.   I think its fairly rare.  About as common as npm infra flake I would say.

src/lib/libstrings.js
RReverser: Yeah, bytes. On my phone now but I think there was an existing comment to that extent above?

RReverser: Ah no, that was just in the UTF-8 decoder. Will do.

sbc100 (approved):

RReverser:
FAIL [0.002s]: test_async_hello_v8 (test_core.instance) looks like a broken test on main branch.

sbc100:
Yup, test_async_hello_v8 is unrelated.  Sorry about that.

PR #24336, opened 2025-05-15
Honor Nodejs environment variables

turran:
Be able to pass Node.js environment variables through the command line and make them available to Emscripten

### DIFF
--- a/src/lib/libwasi.js
+++ b/src/lib/libwasi.js
@@ -67,6 +67,10 @@ var WasiLibrary = {
         '_': getExecutableName()
 #endif
       };
+      // Forward the Node ones, if any.
+      if (ENVIRONMENT_IS_NODE) {
+        env = {...env, ...process.env};
+      }
       // Apply the user-provided values, if any.
       for (var x in ENV) {
         // x is a key in ENV; if ENV[x] is undefined, that means it was


kleisauke:
This is a breaking change, see e.g. the discussion in PR #18820.

turran:
Thanks @kleisauke I wasn't aware of that PR.
I don't understand why such a change can be a breaking change. Not honoring node env is a missing feature as there is no way to pass environment variables other than manually modifying the .js file, which is impractical from the user PoV.

sbc100:
> Thanks @kleisauke I wasn't aware of that PR. I don't understand why such a change can be a breaking change. Not honoring node env is a missing feature as there is no way to pass environment variables other than manually modifying the .js file, which is impractical from the user PoV.

Its a little more subtle than that.  Emscripten tends to present a kind of virtual environment.   By default for example, we don't expose the user's actual filesystem, even under node.  You have to opt into that behaviour.  See  https://github.com/emscripten-core/emscripten/pull/18820/files#r1114735936

I do think we want to enable this behaviour somehow, but it will likely need to be behind some kind of flag.

In any case this does appear to be a duplicate of #18820, no?

turran:
Duplicates https://github.com/emscripten-core/emscripten/pull/18820

PR #24343, opened 2025-05-15
[MODULARIZE=instance] Allow incoming wasmMemory

sbc100:


dschuff (commented):
should this be entitled 'disallow incoming wasmMemory'?

sbc100:
> should this be entitled 'disallow incoming wasmMemory'?

No, I actually noticed that we were disallowing it even though it is supported.

dschuff (approved):

PR #24346, opened 2025-05-16
Update libcxx and libcxxabi to LLVM 20.1.4

aheejin:
This updates libcxx and libcxxabi to LLVM 20.1.4:
https://github.com/llvm/llvm-project/releases/tag/llvmorg-20.1.4

Before going into each additional change of the PR, these are some noteworthy changes from this LLVM release:

- Freezing C++03 headers
  - RFC: https://discourse.llvm.org/t/rfc-freezing-c-03-headers-in-libc/77319
  - PRs:
    - https://github.com/llvm/llvm-project/pull/108999
    - https://github.com/llvm/llvm-project/pull/109000
    - https://github.com/llvm/llvm-project/pull/109001
    - https://github.com/llvm/llvm-project/pull/109002
    - 
  This copies libc++ headers of the last LLVM 19 release into a separate directory (https://github.com/llvm/llvm-project/tree/main/libcxx/include/__cxx03) and redirects all C++03 workflow there. The motivation is not to fix C++03 related changes unless they are critical bugs, and simplifies the main headers. So the main headers are like
  ```
  #if __cplusplus < 201103L && defined(_LIBCPP_USE_FROZEN_CXX03_HEADERS)
  #  include <__cxx03/algorithm>
  #else
  ... main header content ...
  ```

  As you can see it looks like we can avoid opting into this by not defining `_LIBCPP_USE_FROZEN_CXX03_HEADERS` at the moment. I think their eventual goal is to remove C++03 support from the main headers but it hasn't seem to have happened yet and I don't know what their timeline for that is.

  Adding that  `__cxx03` directory increases the header size by 10MB. We can get away not using them by not defining `_LIBCPP_USE_FROZEN_CXX03_HEADERS` in this release, so this update does not include that directory.

- Sharing of headers between libc++ and libc (Project Hand in Hand)
  - RFC: https://discourse.llvm.org/t/rfc-project-hand-in-hand-llvm-libc-libc-code-sharing/77701
  - PR: https://github.com/llvm/llvm-project/pull/91651 (More are likely to come)

  This tries to share some libc headers from libcxx. libcxx's source files can depend on headers from `libc/shared`, `libc/src/__support`, `libc/include/llvm-libc-macros`, and `libc/include/llvm-libc-types`. And it turns out libcxx can also depend on `libc/hdr`, even though the directory is not in the diagram in the RFC.

  These headers can be only included from `libcxx/src` and not `libcxx/include`, to prevent libcxx's API from changing. So these headers don't need to be copied into `cache/`.

- Locale API reimplementation
  It doesn't seem to have an RFC, but the PRs are:
  - https://github.com/llvm/llvm-project/pull/113737
  - https://github.com/llvm/llvm-project/pull/115176
  - https://github.com/llvm/llvm-project/pull/115752
  - https://github.com/llvm/llvm-project/pull/122489

  It looks this aims provide a new way to define locale API for each platform. Currently Apple, FreeBSD, MSVCRT, and Fuschia are using the new API and the others are still using the old one: https://github.com/llvm/llvm-project/blob/ec28b8f9cc7f2ac187d8a617a6d08d5e56f9120e/libcxx/include/__locale_dir/locale_base_api.h#L116-L138
  For our purpose, adding `if defined(__EMSCRIPTEN__)` entry to the list of "old" list seems to work for the moment, but we may need to move to the new way eventually later.

---

Additional changes:

- Copy `vendor/llvm/default_assertion_handler.in` to `__assertion_handler`: https://github.com/emscripten-core/emscripten/pull/24346/commits/aa536481df5f205e1828a734769c37d3f919208c
  Our previous `__assertion_handler` was copied from `libcxx/vendor/llvm/default_assertion_handler.in`. This updates our `__assertion_handler` to the new `libcxx/vendor/llvm/default_assertion_handler.in`. For what this file is, see the PR description of #22994, with which the file was added.

- Remove `libcxx/include/__cxx03` directory: https://github.com/emscripten-core/emscripten/pull/24346/commits/d646f6b8b872d31add619fa87ae8854cad425798
  As I described in "Freezing C++ headers" above, C++03 headers were copied to `include/cxx03` to be "frozen". But by not defining `_LIBCPP_USE_FROZEN_CXX03_HEADERS` we can still use the main headers. This new `__cxx03` header directory is almost 10M and we are not using it in this update, this deletes it.

- Define more variables in `__config_site`: https://github.com/emscripten-core/emscripten/pull/24346/commits/9912236587f0440afabf0bd67ea0cc4991c91aa0
  libcxx decides to almost all configuration macros to be defined. So before it tested whether a macro was defined/undefined, and now it assumes it is defined and tests whether its value is 1/0.

  Before https://github.com/llvm/llvm-project/pull/112094's `_config_site.in` used to use `#cmakedefine`, which only defined  variables when they were enabled, but now it uses `#cmakedefine01`, which always defines variables and assigns 1 or 0 depending on whether the feature is enabled.

  So this change adds `#define` to each variable that `_config_site.in` has an entry of. The value assigned is 1 if it was defined in our previous Emscripten environment and 0 if not.

- Fix Emscripten's locale: https://github.com/emscripten-core/emscripten/pull/24346/commits/2ae01b0026879c7b478e144ce293d99c57dad338
  Before, the code was like https://github.com/emscripten-core/emscripten/blob/dc1abd514b1bade135a01a4453a9ff6def0793b6/system/lib/libcxx/include/__locale_dir/locale_base_api.h#L12-L30
  But now they are divided into two parts, one using the new API and the other using old. See "Locale API reimplementation" above.
  https://github.com/llvm/llvm-project/blob/ec28b8f9cc7f2ac187d8a617a6d08d5e56f9120e/libcxx/include/__locale_dir/locale_base_api.h#L116-L138
  This adds Emscripten in the beginning of the "old" API list. (This has to be the beginning; see #23414)

- Import libc headers used by libcxx: https://github.com/emscripten-core/emscripten/pull/24346/commits/12a4ee468a0faecab4888343eb39d31ee93849f8, https://github.com/emscripten-core/emscripten/pull/24346/commits/12a4ee468a0faecab4888343eb39d31ee93849f8 and https://github.com/emscripten-core/emscripten/pull/24346/commits/43c8ce4ee7992b28594fdd234311a60bb0353231
  This imports a part of libc headers into `system/lib/llvm-libc`. The imported directories are:
  - `libc/shared`
  - `libc/src/__support`
  - `libc/include/llvm-libc-macros`
  - `libc/include/llvm-libc-types`
  - `libc/hdr`
  See "sharing of headers between libc+ and libc" above for details.
  This also applies https://github.com/llvm/llvm-project/pull/133999, which is a  bugfix that has not be backported, which fixes the bug of including from a wrong directory.

- `std::uncaught_exception` -> `std::uncaught_exceptions`: https://github.com/emscripten-core/emscripten/pull/24346/commits/1bf4e78943f5d833c8c9acb5b7c38107aecfdf3d
  `std::uncaught_exception` has been deprecated in C++17, so it generates a warning (which we treat as an error). Replaced it with `std::uncaught_exceptions`, which returns the number of uncaught exceptions (but can still be used as a boolean in the test).

- Remove `ryu_constants.h` and `ryu_long_double_constants.h` from libc: https://github.com/emscripten-core/emscripten/pull/24346/commits/5767ac49739ea94122296d96c4b185512b91e6ac
  These are not used from libcxx and `ryu_long_double_constants.h` is huge (12M).


aheejin:
The only failing test is `instance.test_async_hello_v8` and this fails for me locally even in the main branch, saying
```
emcc: error: MODULARIZE=instance is not compatible with -sASYNCIFY=1
```
So I think this error is not related.

sbc100:
Indeed `instance.test_async_hello_v8` is unrelated.

sbc100 (approved):
Do we need a new update script or add llvm-libc to system/lib/update_libcxx.py?

sbc100 (approved):

sbc100:
After this we still have the compiler-rt changes right?  Do you think its worth holding off making a release until that is done?  (I was hoping to make a release soon).

aheejin:
> After this we still have the compiler-rt changes right? Do you think its worth holding off making a release until that is done? (I was hoping to make a release soon).

I think maybe even for this PR it's better to wait for at least several days to see whether the waterfall is stable, because this is also a big change. How about doing the release now (or in a few days) and I land this (and compiler-rt) later?

sbc100:
Yes, I actually think it might be a good to get a release out once I land https://github.com/emscripten-core/emscripten/pull/24288.. then we can land this after the release.

sbc100:
The release is now done if you want to land this?

(Does it matter if compiler-rt lands before or after this change?)

aheejin:
It turned out a file was missing from `llvm-libc/src/__support` (which is one of the dirs we import). I re-added it: https://github.com/emscripten-core/emscripten/pull/24346/commits/ac7ead3f441b08f53222b4e284f06a0e5c12fcc9
That file is a list of constants and a 12M text file. This file will not be copied to headers though. Would that be fine?

sbc100 (approved):

PR #24349, opened 2025-05-16
Tweak condition for TEXTDECODER=2 default

RReverser:
- When using pthreads, our `TextDecoder` usage already correctly handles shared memory through the `getUnsharedTextDecoderView` helper, so the check and the comment are outdated.
 - When targeting Audio Worklets, we already have assertion that TEXTDECODER!=2 as it's not compatible with that environment, so opt-out from the size optimisation there instead.


sbc100:
Is see the pthreads build uses `getUnsharedTextDecoderView`... I assume that applies to wasm workers too.  Maybe mention `getUnsharedTextDecoderView`?

sbc100 (approved):

RReverser:
> Maybe mention `getUnsharedTextDecoderView`?

You mean in the PR description or in a comment in the code?

sbc100:
> > Maybe mention `getUnsharedTextDecoderView`?
> 
> You mean in the PR description or in a comment in the code?

The PR description

RReverser:
Done.

PR #24355, opened 2025-05-17
Use AsciiToString in Embind. NFC

RReverser:
Embind had its own read readLatin1String, which is functionally identical to the AsciiToString helper we have in the shared libstrings library.

The only difference is that It had extra logic for caching characters by code, which are likely not even an optimisation in modern JIT engines, and if we do want to do that in the future for some reason, it's best to do so in centralised helpers.

For now, a simple search-replace gives an easy win in terms of code size.

### DIFF
--- a/src/lib/libembind.js
+++ b/src/lib/libembind.js
@@ -8,7 +8,7 @@
 /*global Module, asm*/
 /*global _malloc, _free, _memcpy*/
 /*global FUNCTION_TABLE, HEAP8, HEAPU8, HEAP16, HEAPU16, HEAP32, HEAPU32, HEAPF32, HEAPF64*/
-/*global readLatin1String*/
+/*global AsciiToString*/
 /*global Emval, emval_handle_array, __emval_decref*/
 /*jslint sub:true*/ /* The symbols 'fromWireType' and 'toWireType' must be accessed via array notation to be closure-safe since craftInvokerFunction crafts functions as strings that can't be closured. */
 
@@ -229,9 +229,9 @@ var LibraryEmbind = {
     return sharedRegisterType(rawType, registeredInstance, options);
   },
 
-  _embind_register_void__deps: ['$readLatin1String', '$registerType'],
+  _embind_register_void__deps: ['$AsciiToString', '$registerType'],
   _embind_register_void: (rawType, name) => {
-    name = readLatin1String(name);
+    name = AsciiToString(name);
     registerType(rawType, {
       isVoid: true, // void return values can be optimized out sometimes
       name,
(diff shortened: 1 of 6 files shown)


src/lib/libembind.js
```diff
@@ -8,7 +8,7 @@
 /*global Module, asm*/
 /*global _malloc, _free, _memcpy*/
 /*global FUNCTION_TABLE, HEAP8, HEAPU8, HEAP16, HEAPU16, HEAP32, HEAPU32, HEAPF32, HEAPF64*/
-/*global readLatin1String*/
+/*global AsciiToString*/
```
sbc100, line 11: I don't actually know that these comments are event here for.. do you ?

src/lib/libemval.js
```diff
@@ -42,16 +42,16 @@ var LibraryEmVal = {
...
+    emval_symbols[address] = AsciiToString(address);
   },
 
-  $getStringOrSymbol__deps: ['$emval_symbols', '$readLatin1String'],
+  $getStringOrSymbol__deps: ['$emval_symbols', '$AsciiToString'],
   $getStringOrSymbol: (address) => {
     var symbol = emval_symbols[address];
     if (symbol === undefined) {
-      return readLatin1String(address);
+      return AsciiToString(address);
```
sbc100, line 54: Why does this function start with a capital letter .. so strange.    Should we add make a lower case version and deprate this one?

src/lib/libembind.js
RReverser: I think they were meant to be Eslint comments, but they are likely entirely useless in library code due to our custom syntax.

src/lib/libemval.js
RReverser: Idk, but it was always like that, I believe it existed from early days of Emscripten.

src/lib/libembind.js
RReverser: We should be able to remove them in a separate PR, I don't think it's a blocker for this one.

sbc100: sgtm

sbc100 (approved):

RReverser:
The only concern I have (eternal with the way libraries are built in Emscripten) is whether `readLatin1String` is considered a public API that might be used by other users or an internal one.

Do we just need a changelog line to advise users if any to switch to the other function? Or do we need to preserve this name as an alias ~forever?

sbc100:
I'm fairly confident that we can consider this function internal only.  @brendandahl WDYT?

Might be worth adding a changelog entry yes.

If we are concerned then its also trivial to add an alias using `oldname: "newname",` in the library library file.

brendandahl:
I'd consider it internal and am fine with it going away.

RReverser:
@sbc100 I already restarted firefox CI, but it failed again. I see we don't configure `EMTEST_RETRY_FLAKY` for Firefox & Chrome CI tasks, only for core tests - we probably should?

sbc100:
> @sbc100 I already restarted firefox CI, but it failed again. I see we don't configure `EMTEST_RETRY_FLAKY` for Firefox & Chrome CI tasks, only for core tests - we probably should?

Browser tests have (annoyingly) a different retry mechanism which applies to all tests.  They are all retried on failure.

PR #24356, opened 2025-05-17
Use fromCodePoint / codePointAt APIs

RReverser:
These APIs were added to JS strings long ago now and are supported in all engines specified in our `OLDEST_*` browser versions.

They save some code size by avoiding manual UTF-16 <-> UTF-32 conversion, although we still need to worry about correct iteration.


sbc100 (commented):
Nice find!

Thanks for all the recent improvements ingvar!

@juj will be glad to see this too I'm sure

sbc100 (approved):

sbc100:
I fixed the test-esm-integration on main if you want to re-base (or we can just land this)

I might want to wait for @juj or @kripken in case they know of any reason not to go this direction.

RReverser:
> I fixed the test-esm-integration on main if you want to re-base (or we can just land this)

Thanks, all green now.

PR #24358, opened 2025-05-17
Fix missing parenthesis

adamscott:
Discovered that issue when trying to find a way to run eslint on library files.


hoodmane (approved):
Thanks!

PR #24359, opened 2025-05-17
Don't check exports for being valid C/C++ identifiers in side modules

hoodmane:
This unbreaks Rust side modules, which have been broken since https://github.com/emscripten-core/emscripten/pull/23563 was merged.


hoodmane:
@sbc100 I think we discussed a while back the possibility of adding a Rust smoke test. I think such a smoke test would have picked up this regression. Would it be okay if I added such a smoke test?

sbc100:
Sure we have `test_rust_integration_basics` in `test_other.py`  if you can find a way to write a similar test that will fail without this change that would great.

hoodmane:
Closed as fixed by #24427.

PR #24364, opened 2025-05-19
Fix crash in getifaddrs

sbc100:
We don't support `AF_NETLINK` sockets so just fail early here.

Replaces #24187


src/lib/libsockfs.js
```diff
@@ -44,7 +44,15 @@ addToLibrary({
       return FS.createNode(null, '/', {{{ cDefs.S_IFDIR | 0o777 }}}, 0);
     },
     createSocket(family, type, protocol) {
+      // We only support AD_INET
```
kripken, line 47:
```suggestion
      // We only support AF_INET
```

kripken (approved):

PR #24371, opened 2025-05-19
Unify legacy dyncall mechanisms

sbc100:
With this change we use the MINIMAL_RUNTIME technique in all cases.

That is, dyncall functions are stored in a global `dynCalls` object that maps signatures to functions.  Previously we were relying on the `dynCall_xx` helpers existing on the `Module` object.

This allows `-sMODULARIZE=instance` to work with `-sDYNCALL` and by extension with `-sASYNCIFY=1`.


kripken (approved):
Nice!

PR #24376, opened 2025-05-20
Set MIN_x_VERSION as unsupported when not targeting ENVIRONMENT

### LINKED ISSUE #20561: MIN_XXX_VERSION should take ENVIRONMENT into account
`MIN_XXX_VERSION` currently can drag supported feature set down even if `ENVIRONMENT` clearly excludes `XXX`.

Extreme example:

```
> ./emcc temp.cpp -o temp.js -s MIN_CHROME_VERSION=1 -s ENVIRONMENT=node
emcc: warning: enabling transpilation via closure due to browser version settings.  This warning can be suppressed by passing `--closure=1` or `--closure=0` to opt into our explicitly. [-Wtranspile]
```

In this case having `MIN_CHROME_VERSION` causes Emscripten to perform transpilation, even though target `ENVIRONMENT` is clearly specified to only care about Node.js.

Same can happen in the other direction with `-s ENVIRONMENT=web` being dragged down by `MIN_NODE_VERSION`, and it can happen even when user doesn't specify any custom version overrides, merely with built-in defaults.

As a more realistic example: our `MIN_NODE_VERSION` default is set to Node.js 16.0, so building with `-s ENVIRONMENT=node` should enable bulk memory by default, but currently doesn't because minimum Firefox version defaults to 68 and minimum Safari version defaults to 14.1, both of which didn't have bulk memory. So Node.js projects don't get bulk memory even though they don't care about those browsers. cc @sbc100 @kripken

RReverser:
If `ENVIRONMENT` doesn't contain `node`, treat `MIN_NODE_VERSION` as unsupported, and vice versa - if `ENVIRONMENT` only contains `node`, treat all browser `MIN_*_VERSION` as unsupported.

Will show a warning if user specified their own `MIN_*_VERSION` despite corresponding environment not being targeted.

Fixes #20561.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -16099,3 +16099,23 @@ addToLibrary({
 
   def test_getifaddrs(self):
     self.do_other_test('test_getifaddrs.c')
+
+  @parameterized({
+    'web': ('web',),
+    'node': ('node',),
+  })
+  def test_unsupported_min_version_when_unsupported_env(self, env):
+    create_file('pre.js', '''
+      #preprocess
+      var MIN_NODE_VERSION = {{{ MIN_NODE_VERSION }}};
+      var MIN_CHROME_VERSION = {{{ MIN_CHROME_VERSION }}};
+      var MIN_SAFARI_VERSION = {{{ MIN_SAFARI_VERSION }}};
+      var MIN_FIREFOX_VERSION = {{{ MIN_FIREFOX_VERSION }}};
+    ''')
+    self.emcc(test_file('hello_world.c'), [f'-sENVIRONMENT={env}', '--pre-js=pre.js'])
+    src = read_file('a.out.js')
+    unsupported = 0x7FFFFFFF
+    self.assertContainedIf(f'var MIN_NODE_VERSION = {unsupported};', src, env == 'web')
+    self.assertContainedIf(f'var MIN_CHROME_VERSION = {unsupported};', src, env == 'node')
+    self.assertContainedIf(f'var MIN_SAFARI_VERSION = {unsupported};', src, env == 'node')
+    self.assertContainedIf(f'var MIN_FIREFOX_VERSION = {unsupported};', src, env == 'node')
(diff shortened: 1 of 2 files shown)


test/test_other.py
```diff
@@ -16095,3 +16095,23 @@ def test_em_js_bool_macro_expansion(self):
...
+    'web': ('web',),
+    'node': ('node',),
+  })
+  def test_unsupported_min_version_when_unsupported_env(self, env):
+    create_file('pre.js', '''
+      #preprocess
+      var MIN_NODE_VERSION = {{{ MIN_NODE_VERSION }}};
+      var MIN_CHROME_VERSION = {{{ MIN_CHROME_VERSION }}};
+      var MIN_SAFARI_VERSION = {{{ MIN_SAFARI_VERSION }}};
+      var MIN_FIREFOX_VERSION = {{{ MIN_FIREFOX_VERSION }}};
```
sbc100, line 16109: Can we write this test by putting this in the pre.js instead:

```
#if !ENVIRONMENT_MAY_BE_NODE
assert(MIN_NODE_VERSION == 0x7FFFFFFF)
#endif

#if !ENVIRONMENT_MAY_BE_WEB
assert(MIN_CHROME_VERSION == 0x7FFFFFFF)
assert(MIN_SAFARI_VERSION == 0x7FFFFFFF)
assert(MIN_FIREFOX_VERSION == 0x7FFFFFFF)
#endif
```

I don't love trying read/understand `assertContainedIf` type things.

sbc100 (approved):

test/test_other.py
RReverser: I tried something similar first, but unfortunately that doesn't work because when the test is built in `-sENVIRONMENT=web` mode, it complains about not being able to run in Node.js (which is our default test runner).

RReverser: If you have a better idea, I'm happy to switch though.

sbc100: > I tried something similar first, but unfortunately that doesn't work because when the test is built in `-sENVIRONMENT=web` mode, it complains about not being able to run in Node.js (which is our default test runner).

Ah yes of course.

sbc100 (approved):

RReverser:
I see CI is broken, so need to wait for https://github.com/emscripten-core/emscripten/pull/24377.

sbc100:
Should be fixed now.  Feel free to land without re-testing though.

RReverser:
It's fine, I'm in no rush. (also I don't have rights to merge bypassing checks anyway)

PR #24382, opened 2025-05-20
[ASYNCIFY] Avoid dependency on the names of exports

sbc100:
This change moves from using export names to function identity to track the functions in the resumption stack.

The upside of this is that we can wrap non-exports, such as table entire, which we currently avoid by enabling `-sDYNCALLS` whenever `ASYNCIFY=1` is used.

It also means we can potentially make `ASYNCIFY=1` work with the `wasmExports` global existing (i.e. in `WASM_ESM_INTEGRATION` mode.

### DIFF
--- a/src/lib/libasync.js
+++ b/src/lib/libasync.js
@@ -35,7 +35,7 @@ addToLibrary({
     // Asyncify code that is shared between mode 1 (original) and mode 2 (JSPI).
     //
 #if ASYNCIFY == 1 && MEMORY64
-    rewindArguments: {},
+    rewindArguments: new Map(),
 #endif
     instrumentWasmImports(imports) {
 #if EMBIND_GEN_MODE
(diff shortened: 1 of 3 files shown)


sbc100:
@kripken .. the one test that fails with this change is `test_emscripten_lazy_load_code`.

This is because `test_emscripten_lazy_load_code` relies on the asyncify resume stack containing names that are then looked in up the dynamically updates wasmExports map.

I can think of some ways I might be able to make ASYNCIFY_LAZY_LOAD_CODE work in with this patch.. but I'm also tempted to simple remove it?  WDYT?

kripken (approved):

sbc100:
I managed to make `ASYNCIFY_LAZY_LOAD_CODE` in a way that wasn't too offensive I think

kripken (approved):

PR #24384, opened 2025-05-21
Avoid generating JS thunks for Wasm exports

sbc100:
This change unifies the way we handle Wasm exports between various different modes.

Now, we no longer generate thunks for Wasm exports before the module is instantiated.  Instead, in debug mode we generate trapping functions that let users know that they used a function reference that was taken too early.

All modes (except DECLARE_ASM_MODULE_EXPORTS=0) now generate a function called `assignWasmExports` that gets called once the module is instantiated and update the global symbols as the Module exports.

This is code size win overall.

This change also simplifies the emitDCEGraph and applyDCEGraphRemovals because it reduces the number of ways the wasmExports are used.


RReverser:
Could be useful to have some small committed `.js` in the repo (for one of the minimal codesize tests maybe?) that would show difference before/after.

sbc100:
> Could be useful to have some small committed `.js` in the repo (for one of the minimal codesize tests maybe?) that would show difference before/after.
 
Yes I agree.

sbc100:
Awesome, the new test/other/codesize/test_codesize_minimal_O0.expected.js file makes the effect of this change much easier to grok.   Great idea.

sbc100:
I'm pretty sure I remember @hoodmane asking about these trampolines/thunks, but I can't seem to find the issue now.

tools/emscripten.py
```diff
@@ -989,29 +950,58 @@ def create_receiving(function_exports, tag_exports):
...
+  # existing in top level JS scope, i.e.
+  # var _main;
+  # function assignWasmExports(wasmExport) {
+  #   _main = wasmExports["_main"];
+  generate_dyncall_assignment = settings.MINIMAL_RUNTIME and settings.DYNCALLS and '$dynCall' in settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE
+  exports = {name: sig for name, sig in function_exports.items() if name != building.WASM_CALL_CTORS}
+
+  if settings.ASSERTIONS:
+    # In debug builds we generate trapping function in case
+    # folks try to call use a reference that was taked before the
```
kripken, line 963:
```suggestion
    # folks try to call/use a reference that was taken before the
```

kripken (approved):

tools/emscripten.py
```diff
@@ -989,29 +950,58 @@ def create_receiving(function_exports, tag_exports):
...
+    # wasm module is available.
+    for sym in exports:
+      mangled = asmjs_mangle(sym)
+      if settings.MODULARIZE == 'instance' and should_export(mangled):
+        maybe_export = 'export '
+      else:
+        maybe_export = ''
+      receiving.append(f"{maybe_export}var {mangled} = makeInvalidEarlyAccess('{mangled}');")
+  elif settings.MODULARIZE == 'instance':
+    # In MODULARIZE == 'instance' mode we declare each export individually
```
RReverser, line 973: Alternatively, you could use `var a, b, ...` unconditionally in both branches, and for ESM integration mode emit `export {a, b};` in a separate step.

It would be functionally equivalent, but a bit simpler implementation-wise.

sbc100: Great idea! That simplifies that code nicely.

RReverser (approved):

hoodmane:
I don't remember the context but I do remember talking about these. The shorter code size seems nice.

PR #24386, opened 2025-05-21
Extract common memory declarations. NFC

RReverser:
I noticed that regular preamble had types for Closure for all the memory arrays, whereas preamble_minimal didn't but probably should.

OTOH, preamble also had unused `HEAP` variable, which preamble_minimal correctly didn't.

Those vars + runtime{Initialized,Exited} seem easy enough to extract into runtime_shared.js so that they stay consistent and in one place.


RReverser:
FWIW while I'm not having any issues with other codesize tests (after doing clear cache + tot), the `test_codesize_hello_single_file` has been consistently producing different results than CI on 2 different Windows machines for a while now, so I'm very suspect it's actually platform-dependant somehow.

src/runtime_shared.js
```diff
@@ -61,6 +61,53 @@ var wasmOffsetConverter;
...
+#endif
+
+#if SUPPORT_BIG_ENDIAN
+/** @type {!DataView} */
+var HEAP_DATA_VIEW;
+#endif
+
+#if !MINIMAL_RUNTIME || ASSERTIONS || SAFE_HEAP || USE_ASAN
+var runtimeInitialized = false;
+#endif
```
sbc100, line 105: Maybe leave this in the non-minimal file rather than moving it and also adding `!MINIMAL_RUNTIME`

sbc100 (approved):
lgtm % comment.

(I think we should rename `runtime_shared.js` to `runtime_common.js` to avoid conflating with the meaning of shared in "shared memory").

src/runtime_shared.js
sbc100: Perhaps put it right along side the `initRuntime` function that uses it?

RReverser: > Perhaps put it right along side the `initRuntime` function that uses it?

You mean in both preamble and preamble_minimal?

RReverser:
> (I think we should rename `runtime_shared.js` to `runtime_common.js` to avoid conflating with the meaning of shared in "shared memory").

Agreed, I almost wrote a commit message talking about "shared memory" and caught myself thinking it's rather confusing.

src/runtime_shared.js
sbc100: It doesn't need to go in preamble_minimal, right?  Since this is not a variable that exists under `-sMINIMUMAL_RUNTIME`

RReverser: It does if those other conditions are met. It existed in both files, I just merged the conditions when extracting.

sbc100: Oops I misunderstood.

sbc100:
> Every time I rebaseline, by the time CI finishes, I get new conflicts 😂
> 
> Automated rebaselining via CI would be a huge time-saver.

Sorry about that...strange that I haven't experienced those issues (or if I have its normally just a question of an out-of-date binaryen or llvm build).  

If you can figure out the root cause that would be really interesting, since, for example, if we have x-platform differences that would be a major bug.

RReverser:
Looks like I was just unlucky as there were couple of PRs / commits over the last 24 hours or so like LLVM upgrade, JS thunks one, maybe few others, so every time I came back to the PR to rebase, I saw new conflicts and had to rebaseline all over again.

Hopefully won't be a very common occurence, but it can get mildly annoying when it happens few times in a row.

RReverser:
Ok down to actual exceptions.

>+/report_result?exception:_emscripten_stack_get_current is not a function

Looks like a legitimate one, although I'm not yet sure why / how this PR affected it.

I'll try to take a quick look but I'm out tomorrow and next week, so it might have to wait.

PR #5919, opened 2017-12-08
Meta-DCE for JS+WASM

kripken:
This starts to use binaryen's `wasm-metadce` tool in `-Os` and `-Oz`, which lets us remove unused code along the boundary of wasm and JS. That is, it can remove things which doing DCE separately on each can't, like cycles between the two worlds.

For example, if wasm imported a JS function, then if the wasm optimizer manages to remove all uses of it, this lets us remove it not just from wasm but also from the JS side too. And the same works the other way as well.

So far, in `-Os` this shrinks hello world's `.js` by 3% and `.wasm` by 18%. This is less effective on large programs, since the removable runtime overhead is smaller - e.g. when using C++ iostreams, it shrinks by 7% / 1%. But where this really shines is on small programs like the testcase in #5836, that just do a little pure computation and don't need a runtime at all: this shrinks that `.js` by 5% and the `.wasm` by 84% (mostly thanks to getting rid of malloc/free!). The absolute numbers are interesting for the wasm, it goes from 10,729 bytes to just 1,740, which is getting us most of the way to emitting a really minimal wasm for such inputs. This reason this doesn't get us all the way is because:

 * We need to represent JS's full graph of reachability for wasm-dce to be maximally effective. Currently this just represents imports and exports by parsing them directly. So interesting cycles on the JS side can't be collected yet.
 * We still export a lot of things by default, which prevents this dead code elimination from eliminating them. So as we improve that (#5836 etc.) this will get more powerful.

On the other hand, this does increase compile times noticeably, mostly on tiny files: 47% for hello world and 23% for C++ iostreams. That's why I think this makes sense to do in `-Os` and `-Oz`, but not other modes.

PR #24387, opened 2025-05-21
Cleanup run_codesize_test. NFC

sbc100:
This change removes the `expected_exists`, `expected_not_exists` arguments which are no longer used.

We had a few places where `waka` was being passed to `expected_not_exists` but there are no code size tests that use a symbols called `waka`.  Perhaps there once was and this was just cargo-culted?

Also, remove the now-unnecessary use of `*args` which was making it hard to understand exactly which arguments the tests were using.


sbc100:
It looks like the presence of `waka` in the `expected_not_exists` goes all the way back to when these metadce tests were added in #5919.   But I can't see why it was part of that list.

These lists are not redundant since we check instead with golden files.

test/other/codesize/hello_libcxx.cpp
```diff
@@ -5,8 +5,7 @@
 
 #include <iostream>
 
-int main()
-{
+int main() {
```
RReverser, line 8: We have clang-format config, is it not applied to tests or not enforced in general?

Could be nice if it was enforced the same way prettier is.

RReverser (approved):
Thank you! I was wondering what these unused arguments were for but wasn't brave enough to touch them.

PR #24390, opened 2025-05-21
Add full expected JS output test_codesize_minimal_O0                     

sbc100:
This file will be quite volatile since it will change on almost any          
change to the code JS code we generate.  However it will make is much        
easier to see the effect of changes like #24384.                             
                                                                             
Also, it should not be any more of less volatile than the existing code      
size expectations files and, like those, this file is updated                
automatically using the `--rebaseline` test flag.

### DIFF
--- a/test/other/codesize/test_codesize_minimal_O0.expected.js
+++ b/test/other/codesize/test_codesize_minimal_O0.expected.js
@@ -1,0 +1,1392 @@
+// include: shell.js
+"use strict";
+
+// The Module object: Our interface to the outside world. We import
+// and export values on it. There are various ways Module can be used:
+// 1. Not defined. We create it here
+// 2. A function parameter, function(moduleArg) => Promise<Module>
+// 3. pre-run appended it, var Module = {}; ..generated code..
+// 4. External script tag defines var Module.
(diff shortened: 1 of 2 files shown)


sbc100:
This change is built on top of #24387 .. please just review the second commit.

RReverser:
> This file will be quite volatile since it will change on almost any
> change to the code JS code we generate

In other codesize tests we commit JS for optimised files only, so that at least any DCE'd code does not appear in the diffs.

Do you think it's worthwhile to do the same here or is O0 intentional?

sbc100:
> > This file will be quite volatile since it will change on almost any
> > change to the code JS code we generate
> 
> In other codesize tests we commit JS for optimised files only, so that at least any DCE'd code does not appear in the diffs.
> 
> Do you think it's worthwhile to do the same here or is O0 intentional?

I'm doing -O0 intentionally so that the JS is readable.  

If figure if we want to see the effect on the generated code and review it that O0 build would be best.

We can extend this it future.

RReverser:
Okay, fine by me, but I'll leave the review to @kripken for one more opinion.

kripken (approved):

PR #24391, opened 2025-05-21
Refactor browser.test_runtime_misuse. NFC

sbc100:
This test is still quite hard to understand, but at least it doesn't have any loops.


test/test_runtime_misuse_2.c
```diff
@@ -1,21 +0,0 @@
...
-EMSCRIPTEN_KEEPALIVE char* note(int n) {
-  EM_ASM({ Module.noted = Number($0); out("set noted " + Module.noted) }, &noted);
-  EM_ASM({ out([$0, $1]) }, n, noted);
-  noted += n;
-  EM_ASM({ out(['noted is now', $0]) }, noted);
-  return (char*)"silly-string";
-}
-
-void free(void* ptr) { // free is valid to call even after the runtime closes, so useful as a hack here for this test
-  EM_ASM({ out(['reporting', $0]) }, noted);
```
sbc100, line 20: This function was never used.. and the rest of this file is that same is as  
test/test_runtime_misuse.c (except without main)

sbc100: Looks like call to free was removed in 3ddc9f28726bcb209829aa83792e8ebbcb99b80c but not the function itself.

kripken (approved):

PR #24392, opened 2025-05-21
Update update_libcxx.py to handle recent changes

aheejin:
`update_libcxx.py` updates
  - Now it handles the new `llvm-libc` directory
  - It excludes unused `ryu_(long_double)constants.h` from libc
  - It deletes `libcxx/include/__cxx03` for now because we are not using it.
  - Make `clean_dir` and `copy_tree` handle the case when the local dir does not exist, which can happen when new directories are added
  - Make `copy_tree` handle removing files in `excludes` when they are in subdirectories (In libc, many `CMakeLists.txt` are in subdirectories)

This also adds `readme.txt` to `llvm-libc` directory.


system/lib/update_libcxx.py
```diff
@@ -60,6 +70,32 @@ def main():
...
+  assert os.path.exists(libc_upstream_dir)
+  libc_local_dir = os.path.join(script_dir, 'llvm-libc')
+
+  copy_dirs = [
+      ('hdr',),
+      ('include', 'llvm-libc-macros'),
+      ('include', 'llvm-libc-types'),
+      ('shared',),
+      ('src', '__support'),
+  ]
```
sbc100, line 87: Should these go at the top level with the the file lists like preserve_files and excludes?

sbc100 (approved):

system/lib/update_libcxx.py
aheejin: Done (Changed the variable name to `libc_copy_dirs`)

aheejin:
`test-mac-arm64` and `test-windows` fail with
```
Running 'git status --short'

M system/lib/llvm-libc/README.txt

Checkout is not clean.  See above for list of dirty/untracked files.

Exited with code exit status 1
```

Not sure what happened, but is there a way to clean the working directory?

sbc100:
> `test-mac-arm64` and `test-windows` fail with
> 
> ```
> Running 'git status --short'
> 
> M system/lib/llvm-libc/README.txt
> 
> Checkout is not clean.  See above for list of dirty/untracked files.
> 
> Exited with code exit status 1
> ```
> 
> Not sure what happened, but is there a way to clean the working directory?

Somehow the README file is being updated during the build... that seems odd?  Each test run starts off clean

sbc100:
You try `run-run with ssh` on the  test-mac-arm64 runner.  Then do `git diff` in the `project` directory to see how the readme is reported to have changed.

sbc100:
Oh interesting you are trying to add a `readme.txt` and there is an existing `README.txt`. That won't work on mac or windows where the filesystem is case sensitive.

aheejin:
> Oh interesting you are trying to add a `readme.txt` and there is an existing `README.txt`. That won't work on mac or windows where the filesystem is case insensitive.

Ah thanks!

PR #24397, opened 2025-05-22
Optimize *.test_sse4_1 test runtime.  NFC

juj:
LLVM struggles with compilation time on large functions, and counterproductively inlines functions to make things worse for itself. Outline each SSE function to be tested in separate noninlineable function.

Before: Total core time: 487.852s. Wallclock time: 323.379s. Parallelization: 1.51x.
After: Total core time: 243.685s. Wallclock time: 120.836s. Parallelization: 2.02x.

### DIFF
--- a/test/sse/test_sse.h
+++ b/test/sse/test_sse.h
@@ -27,6 +27,14 @@
 #define align1_double double
 #endif
 
+#ifdef __GNUC__
+#define NOINLINE __attribute__((noinline))
+#elif defined(_MSC_VER)
+#define NOINLINE __declspec(noinline)
+#else
+#define NOINLINE
+#endif
+
 // Recasts floating point representation of f to an integer.
 uint32_t fcastu(float f) { return *(uint32_t*)&f; }
 uint64_t dcastu(double f) { return *(uint64_t*)&f; }
--- a/test/sse/test_sse4_1.cpp
+++ b/test/sse/test_sse4_1.cpp
@@ -19,19 +19,113 @@ int numInterestingInts = sizeof(interesting_ints_)/sizeof(interesting_ints_[0]);
 double *interesting_doubles = get_interesting_doubles();
 int numInterestingDoubles = sizeof(interesting_doubles_)/sizeof(interesting_doubles_[0]);
 
-void test_round() {
-  Ret_M128d(__m128d, _mm_ceil_pd);
-  Ret_M128(__m128, _mm_ceil_ps);
-  Ret_M128d_M128d(__m128d, _mm_ceil_sd);
-  Ret_M128_M128(__m128, _mm_ceil_ss);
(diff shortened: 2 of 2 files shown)


test/sse/test_sse.h
```diff
@@ -27,6 +27,14 @@
 #define align1_double double
 #endif
 
+#ifdef __GNUC__
+#define NOINLINE __attribute__((noinline))
+#elif defined(_MSC_VER)
+#define NOINLINE __declspec(noinline)
+#else
+#define NOINLINE
+#endif
```
sbc100, line 36: IIUC this code is only ever run via emcc so I think you can just do `#define NOINLINE __attribute__((noinline))`

test/sse/test_sse4_1.cpp
```diff
@@ -19,94 +19,177 @@ int numInterestingInts = sizeof(interesting_ints_)/sizeof(interesting_ints_[0]);
 double *interesting_doubles = get_interesting_doubles();
 int numInterestingDoubles = sizeof(interesting_doubles_)/sizeof(interesting_doubles_[0]);
```
sbc100, line 20: I wonder if there is some `#pragma` to turn on inlining for the whole file?  Maybe not?

test/sse/test_sse4_1.cpp
```diff
@@ -19,94 +19,177 @@ int numInterestingInts = sizeof(interesting_ints_)/sizeof(interesting_ints_[0]);
...
-  Ret_M128i(__m128i, _mm_cvtepu32_epi64);
-  Ret_M128i(__m128i, _mm_cvtepu8_epi16);
-  Ret_M128i(__m128i, _mm_cvtepu8_epi32);
-  Ret_M128i(__m128i, _mm_cvtepu8_epi64);
+void NOINLINE test_dp_pd() {
+  bool oldTestNaNBits = testNaNBits;
   testNaNBits = false;
   Ret_M128d_M128d_Tint(__m128d, _mm_dp_pd);
+  testNaNBits = oldTestNaNBits;
+}
```
sbc100, line 80: Blank line here between functions.

sbc100 (approved):
Thanks for looking into this!

I didn't verify 100% that each function appears in the new version.  I assume you did this semi-programmatically so as not to leave anything out?

test/sse/test_sse.h
juj: I have run the test_sse files against Visual Studio in the past. (Where I know that VS and Clang have different semantics on some), so this is kind of looking to preserve that possibility to triage through such things.

test/sse/test_sse4_1.cpp
juj: The only other possibility I suppose would be to build with `-fno-inline-functions` though that might affect more things.

sbc100 (approved):
lgtm.

PR #24399, opened 2025-05-22
reland: Unify legacy dyncall mechanisms

sbc100:
This change was originally landed in #24371 but got reverted.

With this change we use the MINIMAL_RUNTIME technique in all cases.

That is, dyncall functions are stored in a global `dynCalls` object that maps signatures to functions. Previously we were relying on the `dynCall_xx` helpers existing on the `Module` object.

This allows `-sMODULARIZE=instance` to work with `-sDYNCALL` and by extension with `-sASYNCIFY=1`.

### DIFF
--- a/site/source/docs/compiling/Modularized-Output.rst
+++ b/site/source/docs/compiling/Modularized-Output.rst
@@ -123,10 +123,6 @@ fix in future releses.  Current limitations include:
 
 * `ccall`/`cwrap` are not supported (depends on the ``Module`` global).
 
-* :ref:`dyncalls` is not supported (depends on the ``Module`` global)
-
-* :ref:`asyncify` is not supported (depends on :ref:`dyncalls`)
-
 * :ref:`asyncify_lazy_load_code` is not supported (depends on ``wasmExports``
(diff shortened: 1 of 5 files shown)


kripken (commented):
What forced the revert? (and how is it fixed here?)

sbc100:
When the change landed this change the `core3.test_dyncall_pointers_legacy` was broken.

I make the mistake of not waiting until the CI was complete so that change broke CI for emscripten until I reverted.

My initial fix was https://github.com/emscripten-core/emscripten/pull/24377 and there is some background there about why the change broke.   Basically it related to the `applyDCEGraphRemovals` not being updated to handle the new shape the wrapper function we generate for wasm exports. 

However I abandoned that fix and decided instead to try to simplify things by just removing the wrapper functions completely:  #24384.  After this change the bug won't happen because we only have one to look for a single pattern of wasm export use now and that is the pattern that was already used by MINIMAL_RUNTIME all along.

kripken:
I see, thanks!

kripken (approved):

PR #24402, opened 2025-05-23
Fix test runner warning caused by previous commit, and rename calculateDuration() to calculateElapsed() to avoid confusing/overloading similar terms.

juj:
I removed the odd stub `addDuration()` function in the previous PR: https://github.com/emscripten-core/emscripten/pull/24396/files#diff-fd5a5983b044aa841287bf197f2b5e46781162bf0e6a0ca8d1ccaddf1cc1c969L110-L111

turns out that Python unit test runner starts to warn if this function is not present.

```
test_sse4 (test_core.strict_js.test_sse4) ... ok (1.14s)
C:\python\Lib\unittest\case.py:580: RuntimeWarning: TestResult has no addDuration method
  warnings.warn("TestResult has no addDuration method",
test_sse4 (test_core.bigint.test_sse4) ... ok (1.10s)
C:\python\Lib\unittest\case.py:580: RuntimeWarning: TestResult has no addDuration method
  warnings.warn("TestResult has no addDuration method",
test_sse4 (test_core.instance.test_sse4) ... ok (1.13s)
C:\python\Lib\unittest\case.py:580: RuntimeWarning: TestResult has no addDuration method
  warnings.warn("TestResult has no addDuration method",
test_sse4 (test_core.core_2gb.test_sse4) ... ok (1.25s)
```

Add it back in, and use it to register the elapsed time in the test (looks like Python internally tracks it).

Rename the `calculateDuration()` function to a `calculateElapsed()` method so it doesn't look too similar to the `addDuration()` one.

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -109,9 +109,11 @@ class BufferedParallelTestResult:
   def test(self):
     return self.buffered_result.test
 
-  def calculateDuration(self):
-    self.test_duration = time.perf_counter() - self.start_time
-    return self.test_duration
+  def addDuration(self, test, elapsed):
+    self.test_duration = elapsed
+
+  def calculateElapsed(self):
+    return time.perf_counter() - self.start_time
 
   def updateResult(self, result):
     result.startTest(self.test)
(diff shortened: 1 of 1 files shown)


sbc100 (approved):

test/parallel_testsuite.py
```diff
@@ -130,17 +132,17 @@ def stopTest(self, test):
 
   def addSuccess(self, test):
     if hasattr(time, 'perf_counter'):
-      print(test, '... ok (%.2fs)' % (self.calculateDuration()), file=sys.stderr)
+      print(test, '... ok (%.2fs)' % (self.calculateElapsed()), file=sys.stderr)
```
sbc100, line 135: What is this different to reporting `self.test_duration` ?

juj: It is an processing order thing with Python's test harness. `self.test_duration` has not been calculated yet at this point, i.e. `BufferedParallelTestResult.addDuration()` has not yet been called by the time this function is called, so `self.test_duration` would still report zero.

Issue #24404, opened 2025-05-23
When EM_CACHE is overridden, Permission Denied crash occurs when installing system headers


Technius:
When emscripten is installed to a location with read-only permissions (such as when installed with Nix) and the `EM_CACHE` environment variable is set to a folder with read-write-execute permissions, the `emcc` command still crashes when installing the system headers in the `EM_CACHE`.

---

**Version of emscripten/emsdk:**
Please include the output `emcc -v` here
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.8-git
clang version 20.1.4
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /nix/store/cf6j1f9f0z1syny3dgbickzc42zgcxhf-clang-20.1.4/bin
```

This bug does not occur with `emcc 3.1.73`

**Failing command line in full:**
If this is compile or link-time failure please include the full failing command
along with its entire output.

<details>
<summary>Log</summary>

Using an `example.cpp` with the contents:
```
#include <iostream>

int main(int argc, char** argv) {
    return 0;
}
```

Running `emcc`:
```
% EM_CACHE="$PWD"/emscriptencache emcc example.cpp
cache:INFO: generating system headers: sysroot_install.stamp... (this will be cached in "/private/tmp/example/emscriptencache/sysroot_install.stamp" for subsequent builds)
Traceback (most recent call last):
  File "/nix/store/bzq9l9651iqfk63glxdd3z7lxl10lilz-emscripten-4.0.8/share/emscripten/emcc.py", line 1597, in <module>
    sys.exit(main(sys.argv))
             ^^^^^^^^^^^^^^
  File "/nix/store/v0vy1c59f0vf71h64caw17hzymvg0zms-python3-3.12.10/lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/nix/store/bzq9l9651iqfk63glxdd3z7lxl10lilz-emscripten-4.0.8/share/emscripten/emcc.py", line 1590, in main
    ret = run(args)
          ^^^^^^^^^
  File "/nix/store/bzq9l9651iqfk63glxdd3z7lxl10lilz-emscripten-4.0.8/share/emscripten/emcc.py", line 670, in run
    linker_args = phase_compile_inputs(options, state, newargs)
[...]

PR #24405, opened 2025-05-23
Fix sysroot installation when emscripten itself is read-only

sbc100:
The resulting header tree need to be at least user-writable so that we can incrementally copy files into it, adding new files and also updating previously copied ones.

Sadly, python's shutil doesn't seem to have any options for disabling the copying of mode bits.

This was originally fixed in #15386 but then regressed in #23417.

Fixes: #24404

### DIFF
--- a/test/test_sanity.py
+++ b/test/test_sanity.py
@@ -7,6 +7,7 @@ import glob
 import os
 import platform
 import shutil
+import stat
 import time
 import re
 import tempfile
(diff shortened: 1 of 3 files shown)


sbc100:
It seems like constructing a test for this would be tricky... do you think we can land this without a test?   I verified it works locally with a read-only root-owned copy of emscripten.

kripken (approved):
A test sounds good.

kripken (approved):

Technius:
@sbc100 Thank you for the prompt fix!

PR #24414, opened 2025-05-27
updates url if redirected during fetch

dipidoo:
during redirect, url should be updated to new redirected url


src/Fetch.js
```diff
@@ -345,6 +345,16 @@ function fetchXHR(fetch, onsuccess, onerror, onprogress, onreadystatechange) {
     {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.readyState, 'xhr.readyState', 'i16') }}}
     {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.status, 'xhr.status', 'i16') }}}
     if (xhr.statusText) stringToUTF8(xhr.statusText, fetch + {{{ C_STRUCTS.emscripten_fetch_t.statusText }}}, 64);
+    // update url if redirect happens
+    if (xhr.url_ !== xhr.responseURL) {
+      var urlPtr = stringToNewUTF8(xhr.responseURL);
+      if (urlPtr) {
+        if (fetch + {{{ C_STRUCTS.emscripten_fetch_t.url }}}) {
+          _free(fetch + {{{ C_STRUCTS.emscripten_fetch_t.url }}});
```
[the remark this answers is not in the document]
dipidoo: I will remove this one. I will keep the other section's free null check to limit scope.

ChangeLog.md
```diff
@@ -36,6 +36,8 @@ See docs/process.md for more on how version tagging works.
   nightly release](https://github.com/google/dawn/releases) and file feedback
   with Dawn. (Emdawnwebgpu is maintained as part of Dawn, the open-source
   WebGPU implementation used by Chromium, but it is still cross-browser.)
+- Fetch C APIs will now update the `url` field in `emscripten_fetch_t` if a
+  redirect occurred, in line with JS `XHR` and `fetch` behavior.
```
sbc100, line 40: Can you add `(#24414)` here.

src/Fetch.js
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2016 The Emscripten Authors
+ * Copyright 2025 The Emscripten Authors
```
sbc100, line 3: We don't update copyright years in existing files.

sbc100 (approved):
Lgtm with comments addressed

src/Fetch.js
dipidoo: Sorry I misunderstood your comment, I will update them.

src/Fetch.js
```diff
@@ -345,6 +345,16 @@ function fetchXHR(fetch, onsuccess, onerror, onprogress, onreadystatechange) {
     {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.readyState, 'xhr.readyState', 'i16') }}}
     {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.status, 'xhr.status', 'i16') }}}
     if (xhr.statusText) stringToUTF8(xhr.statusText, fetch + {{{ C_STRUCTS.emscripten_fetch_t.statusText }}}, 64);
+    // update url if redirect happens
+    if (xhr.url_ !== xhr.responseURL) {
+      var urlPtr = stringToNewUTF8(xhr.responseURL);
+      if (urlPtr) {
+        if (fetch + {{{ C_STRUCTS.emscripten_fetch_t.url }}}) {
```
sbc100, line 352: Were you going to remove this `if` check?

sbc100 (approved):

src/Fetch.js
```diff
@@ -190,7 +190,7 @@ function fetchLoadCachedData(db, fetch, onsuccess, onerror) {
   }
 }
 
-function fetchCacheData(/** @type {IDBDatabase} */ db, fetch, data, onsuccess, onerror) {
+function fetchCacheData(/** @type {IDBDatabase} */ db, fetch, dest, data, onsuccess, onerror) {
```
dipidoo, line 193: @sbc100 I restored the IndexedDB key to use the request url instead of the response one that was the effect of updating the url field. This was the unit tests failure root cause and now they are passing. Please let me know if this is acceptable approach.

src/Fetch.js
```diff
@@ -535,7 +546,7 @@ function startFetch(fetch, successcb, errorcb, progresscb, readystatechangecb) {
         else successcb?.(fetch);
       });
     };
-    fetchCacheData(Fetch.dbInstance, fetch, xhr.response, storeSuccess, storeError);
+    fetchCacheData(Fetch.dbInstance, fetch, xhr.url_, xhr.response, storeSuccess, storeError);
```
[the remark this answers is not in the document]
dipidoo: I added an argument to the internal function `fetchCacheData` to accept the verbatim request destination url that will be used as a retrieval key for the cache, and that is the value to be used in this context.

test/fetch/test_fetch_from_thread.cpp
```diff
@@ -22,7 +26,7 @@ void* thread_main(void* arg) {
     assert(fetch);
     printf("Finished downloading %llu bytes\n", fetch->numBytes);
     assert(fetch->url);
-    assert(!strcmp(fetch->url, "gears.png"));
+    assert(!strcmp(fetch->url, ORIGIN_URL "gears.png"));
```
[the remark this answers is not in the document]
dipidoo: It is not a redirect but a resolution. This will make it consistent with XMLHttpRequest, but I am aware while implementing it that some compromises are needed in a few solutions that I can think of, given the current behavior:

- Always resolve it to make it consistent with XHR API (current). This changes the behavior for non-redirect case as well.
- Strictly limit the XHR behavior alignment to redirect case. This requires accessing `globalThis.location` to obtain the host part to determine if redirect occurred.
- Limit the XHR behavior alignment change with a flag: one of the above, but with compile flag required to limit the breaking change.

Please let me know which one you prefer, or hopefully a better alternative than all above.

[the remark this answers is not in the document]
sbc100: How about adding a new field so that you can have `fetch->url` (the original requested URL) and `fetch->resolvedUrl` (the resolved/redireced URL)?  Then hopefully no old code would need to be changed?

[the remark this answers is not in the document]
dipidoo: I think that is a great idea, thank you. Do you mind if I set the name to responseUrl to align with the JS XHR API?

src/Fetch.js
```diff
@@ -345,6 +345,15 @@ function fetchXHR(fetch, onsuccess, onerror, onprogress, onreadystatechange) {
     {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.readyState, 'xhr.readyState', 'i16') }}}
     {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.status, 'xhr.status', 'i16') }}}
     if (xhr.statusText) stringToUTF8(xhr.statusText, fetch + {{{ C_STRUCTS.emscripten_fetch_t.statusText }}}, 64);
+    if (fetchAttrSynchronous)
+      // The response url pointer malloc()ed here has the same lifetime as the emscripten_fetch_t structure itself has, and is
+      // freed when emscripten_fetch_close() is called.
+      var ruPtr = stringToNewUTF8(xhr.responseURL);
+      if (ruPtr) {
+        _free(fetch + {{{ C_STRUCTS.emscripten_fetch_t.responseUrl }}});
```
sbc100, line 353: I don't think you need to free here do you since the pointer will always be null when a request starts?

sbc100: Also I think you can skip this `if` check and just assign to responseUrl unconditionally (here and below).

sbc100:
Can you update the PR title?

src/Fetch.js
```diff
@@ -345,14 +345,12 @@ function fetchXHR(fetch, onsuccess, onerror, onprogress, onreadystatechange) {
...
-    if (fetchAttrSynchronous)
+    if (fetchAttrSynchronous) {
       // The response url pointer malloc()ed here has the same lifetime as the emscripten_fetch_t structure itself has, and is
       // freed when emscripten_fetch_close() is called.
       var ruPtr = stringToNewUTF8(xhr.responseURL);
-      if (ruPtr) {
-        _free(fetch + {{{ C_STRUCTS.emscripten_fetch_t.responseUrl }}});
-        {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.responseUrl, 'ruPtr', '*') }}}
-      }
+      _free(fetch + {{{ C_STRUCTS.emscripten_fetch_t.responseUrl }}});
```
sbc100, line 352: I don't think you need this free call either.. since its only assigned once isn't it?

dipidoo: You're right. I did not do it on previous commit because I was still checking the standard. But now I have normative reference that in whatwg/xhr section 3.5.6 step 11.9 that this will only happen exactly once unless an error occurred before reaching this step when it will not be triggered at all. I will remove the free's.

sbc100 (approved):

PR #21428, opened 2024-02-26
Multiple fixes to EmscriptenOrientationChangeEvent

cwoffenden:
This fixes multiple issues with `EmscriptenOrientationChangeEvent`. Buckle in!

- The existing implementation doesn't work at all and is only useful to notify the orientation did change. It's incorrectly processed as a string from the start but is a `ScreenOrientation` object.
https://github.com/emscripten-core/emscripten/blob/b5b7fedda835bdf8f172a700726109a4a3899909/src/library_html5.js#L940
The lookup should be from `screen.orientation.type`:
ScreenOrientation object

- Since it will always fail the string lookup it only ever results in `-1`, which then gets stored in the `EmscriptenOrientationChangeEvent` as `1 << orientation`, not matching any of the known values:
https://github.com/emscripten-core/emscripten/blob/b5b7fedda835bdf8f172a700726109a4a3899909/src/library_html5.js#L946

- The orientation angle is never correct, since it's the same  `-1` as above:
https://github.com/emscripten-core/emscripten/blob/b5b7fedda835bdf8f172a700726109a4a3899909/src/library_html5.js#L947

- Building with `-s ASSERTIONS=2 -s SAFE_HEAP=1` always asserts due to the same issue:
```
Aborted(Assertion failed: value (-2147483648) too small to write as 32-bit value)
```

- The docs mention `window.orientation` as an Emscripten extension but the code doesn't contain any of this:
https://github.com/emscripten-core/emscripten/blob/b5b7fedda835bdf8f172a700726109a4a3899909/site/source/docs/api_reference/html5.h.rst#L961

A repro case is to build the existing `test_html5_core.c` with `SAFE_HEAP` to trigger all the cases above:
```
emcc -O0 -g3 -s ASSERTIONS=2 -s SAFE_HEAP=1 -s USE_WEBGL2=1 -lGL test_html5_core.c -o index.html
```

This will assert on opening (even on desktop), then any devices that generate orientation changes will be always filled with the same data.

This fix:
- Rewrites the JS to pull the values out of the `ScreenOrientation` object.
- Correctly sets the `EM_ORIENTATION_PORTRAIT_xxx` type.
- Correctly sets the angle.
- For older browsers that's don't support `screen.orientation` (Safari before 2023) fallback to the `window.orientation` mentioned in the docs for the angle.

PR #24423, opened 2025-05-28
Remove/fix usage of MIN_SAFARI_VERSION

sbc100:
We don't support targeting safari version older than 10.10.00 so two of these conditions were always true.

The third check was added in #21428 and looks to me like it was always wrong.  The window.orientation API was added in safari 16.4 as the comment says so I updated the check to match.  I'm not sure where the old value of 0x100400 came from or what it means.

### DIFF
--- a/src/lib/libemval.js
+++ b/src/lib/libemval.js
@@ -303,7 +303,7 @@ var LibraryEmVal = {
     return id;
   },
 
-#if MIN_CHROME_VERSION < 49 || MIN_FIREFOX_VERSION < 42 || MIN_SAFARI_VERSION < 100101
+#if MIN_CHROME_VERSION < 49 || MIN_FIREFOX_VERSION < 42
   $reflectConstruct: null,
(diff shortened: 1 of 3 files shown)


sbc100:
CC @cwoffenden who authors the check with the hex value.

cwoffenden:
> CC @cwoffenden who authors the check with the hex value.

That is odd, but since it’s passed on the command-line it probably worked during testing. It should be 160400, at a glance.

sbc100:
> > CC @cwoffenden who authors the check with the hex value.
> 
> That is odd, but since it’s passed on the command-line it probably worked during testing. It should be 160400, at a glance.

Right, but presumably not 0x160400 which would be 1442816.   Looks like some confusion somewhere (I hope).

cwoffenden:
> Looks like some confusion somewhere (I hope).

The docs say MMmmVV (something like that, from memory), it’s pretty clear, I don’t know where I made up the hex from.

PR #24424, opened 2025-05-28
Fix bootstrap.py script to not depend on tools/shared.py

juj:
Fix bootstrap.py script to not depend on tools/shared.py so that it can be run during emsdk installation phase when no .emscripten config yet exists. Fixes https://github.com/emscripten-core/emsdk/issues/1545

### DIFF
--- a/test/test_sanity.py
+++ b/test/test_sanity.py
@@ -822,13 +822,9 @@ fi
   # cause an exception. (A first time run scenario is before .emscripten, env.
   # vars nor PATH has been configured)
   def test_bootstrap_without_em_config(self):
-    # Make sure we don't have an EM_CONFIG file.
-    if os.path.isfile(EM_CONFIG):
-      os.remove(EM_CONFIG)
-
     # Remove all environment variables that might help config.py to locate Emscripten tools.
     env = os.environ.copy()
-    for e in ['LLVM_ROOT', 'EMSDK_NODE', 'EMSDK_PYTHON', 'EMSDK', 'EMSCRIPTEN', 'BINARYEN_ROOT', 'EMCC_SKIP_SANITY_CHECK']:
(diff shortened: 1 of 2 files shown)


test/test_sanity.py
```diff
@@ -817,3 +817,28 @@ def test_bootstrap(self):
 
     # Now the compiler should work again
     self.run_process([EMCC, test_file('hello_world.c')])
+
+  # Verify that running bootstrap.py in a first-time run scenario should not
+  # cause an exception. (A first time run scenario is before .emscripten, env.
+  # vars nor PATH has been configured)
+  def test_bootstrap_without_em_config(self):
```
sbc100, line 824: Can you just do `@with_env_modify({'EM_CONFIG': "/path/does/not/exist"})` maybe?

juj: Sure - folded that to the other env. modify code.

sbc100: But can you literally just write this test as:

```
@with_env_modify({'EM_CONFIG': "/path/does/not/exist"})
 def test_bootstrap_without_em_config(self):
   # Verify that bootstrap works even without a valid config file.
   self.run_process([shared.bat_suffix(shared.path_from_root('bootstrap'))])
```

juj: Oh I see, the behavior of `from tools import shared` is different with

```
@with_env_modify({'EM_CONFIG': "/path/does/not/exist"})
```
versus
```
@with_env_modify({'EM_CONFIG': None})
```

It is better to test the scenario that first-time emsdk install also exercises, when there are no env. vars yet set. This way the test matches what happens on a new installation, and won't rely on EM_CONFIG pointing to a nonexistent file (emsdk doesn't do that either on first-time install)

sbc100: Fair enough.  I think its probably overkill but I get that there could be difference.

sbc100 (approved):

Issue #24413, opened 2025-05-26
Add option to allow dashes "-" in export names for WASM components


peterhirn:
Hi,

I'm experimenting with building WASM components with emscripten. I'm using [wit-bindgen](https://github.com/bytecodealliance/wit-bindgen) to generate C code from a `wit` file, then compile everything with `emcc`.

After that I run `wasm-tools component new emcc-output.wasm -o component.wasm` to create the component and finally `wasmtime run --invoke 'error-message(-6)' component.wasm` to run the code.

The problem is that [WIT functions](https://component-model.bytecodealliance.org/design/wit.html#functions) are kebab-case and this is currently not allowed in emscripten, see https://github.com/emscripten-core/emscripten/blob/main/tools/emscripten.py#L575

Using this `.wit`

```wit
world root {
  export error-message: func(error-code: s32) -> string;
}
```

produces the error

```
emcc: error: invalid export name: cabi_post_error-message
```

For testing I patched out the check and everything seems to be working. Maybe you could allow dashes in export names for `STANDALONE_WASM` or `PURE_WASI`?

sbc100:
The problem is that we create corresponding JS function for each of these function exports and JS symbols cannot contains hyphens like this.

How are JS users supposed to invoke these symbols?   I see a couple of options but I'm not sure either of then is great:

1. Use the mostly-internal-and-perhaps-deprecated soon `wasmExports` map that allows string access to all wasm exports.  I've been hoping to remove the use of this global with the work I've been doing on `WASM_ESM_INTEGRATION` since using this kind of prevents a lot of DCE efforts (since it doesn't express dependencies via the module graph).   
2. Use some kind of name mangling rule.  So `cabi_post_error-message` would be exposed as `cabi_post_error_message`.

I'm not sure I like either of these options.  

I suppose in the ESM integration world we would do something kind of name mangled internally and then re-export via the original name.  e.g.

```
import { "error-message" as _internal_error_message} from "./themodule.wasm";

... code that uses `_internal_error_message`..

export { _internal_error_message as "error-message" };
```

This would at least give the user-faces JS module the same shaped symbols as the internal wasm module, but it still has the problem that is not very consume via normal `import` statements.

sbc100:
Ah yes, if we are not generating JS I don't think we need this check.  Good point.

PR #24427, opened 2025-05-28
Allow arbitrary export names when not generating JS output

sbc100:
Fixes: #24413


sbc100:
@hoodmane I think this might be enough to fix your rust issue too?  (Since we don't generate JS for side modules).

kripken (approved):

hoodmane:
Sounds like it.

PR #24429, opened 2025-05-28
[MODULARIZE=instance] Fix growableHeap pass

sbc100:
This was causing `asan.test_modularize_instance_pthreads` to fail.


sbc100:
This should make that emscripten-releases waterfall green again.

kripken (approved):

PR #19938, opened 2023-07-31
Allow certain command line settings to be additive

sbc100:
For certain command line settings (specifically most settings that take a list) it is useful to be able to add to the existing setting rather than replace it.

For example, the `-sEXPORTED_FUNCTIONS` argument.  In a build system where different components can inject link flags it makes sense that each component be able add to the list of exported functions without clobbering the current list.

We have had this feature requested several times in the past. For example, from the bazel toolchain folks.

PR #24431, opened 2025-05-29
Fix strict test suite

juj:
Enable `test/runner strict` to pass (tested on Windows)

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -1147,7 +1147,7 @@ class RunnerCore(unittest.TestCase, metaclass=RunnerMeta):
     if self.get_setting('WASMFS'):
       # without this the JS setup code in setup_nodefs.js doesn't work
       self.set_setting('FORCE_FILESYSTEM')
-    self.emcc_args += ['-DNODEFS', '-lnodefs.js', '--pre-js', test_file('setup_nodefs.js')]
+    self.emcc_args += ['-DNODEFS', '-lnodefs.js', '--pre-js', test_file('setup_nodefs.js'), '-sINCOMING_MODULE_JS_API=[onRuntimeInitialized]']
 
   def setup_noderawfs_test(self):
     self.require_node()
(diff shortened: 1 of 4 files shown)


test/common.py
```diff
@@ -1336,17 +1336,17 @@ def in_dir(self, *pathelems):
   def add_pre_run(self, code):
     assert not self.get_setting('MINIMAL_RUNTIME')
     create_file('prerun.js', 'Module.preRun = function() { %s }\n' % code)
-    self.emcc_args += ['--pre-js', 'prerun.js']
+    self.emcc_args += ['--pre-js', 'prerun.js', '-sINCOMING_MODULE_JS_API=[preRun]']
```
sbc100, line 1339: The problem with doing this here is that then the test doesn't have any way to add more to `INCOMING_MODULE_JS_API`.

We currently don't have any way to make settings like `INCOMING_MODULE_JS_API` additive.

I've tried to make them additive by default (which seems like the right behaviour in almost all cases I can think of): https://github.com/emscripten-core/emscripten/pull/19938.   But that PR got hung up with disagreement about how / if to change the default behaviour.

sbc100 (approved):
lgtm % concern about INCOMING_MODULE_JS_API additive-ness

test/common.py
juj: That is, either that specific test can be written to duplicate the `-sINCOMING_MODULE_JS_API=[preRun,theothersymbol]`, or `-sINCOMING_MODULE_JS_API=[preRun]` be hoisted from `add_pre_run()` to all the tests.

test/test_core.py
```diff
@@ -9414,7 +9414,11 @@ def test_pthread_dylink_longjmp(self):
   @needs_dylink
   @node_pthreads
   def test_pthread_dylink_main_module_1(self):
-    self.emcc_args += ['-Wno-experimental', '-pthread', '-lhtml5']
+    # TODO: For some reason, -lhtml5 must be passed in -sSTRICT mode, but can NOT
+    # be passed when not compiling in -sSTRICT mode. That does not seem intentional?
```
sbc100, line 9418: This certainly sounds like a bug. Could you open one an link it here (and below).

PR #24432, opened 2025-05-29
Skip Intel x86 SSE tests on ARM64 platforms that cannot natively compile SSE code.

juj:
Also remove one case of @requires_native_clang that wasn't needed.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -140,13 +140,13 @@ def needs_dylink(func):
   return decorated
 
 
-def requires_x86_cpu(func):
+def requires_x64_cpu(func):
   assert callable(func)
 
   @wraps(func)
   def decorated(self, *args, **kwargs):
-    if platform.machine().lower() not in ['i686', 'x86_64', 'amd64']:
-      return self.skipTest(f'This test requires a native x86 CPU. Current CPU is {platform.machine()}.')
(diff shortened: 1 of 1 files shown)


test/common.py
```diff
@@ -252,6 +253,18 @@ def decorated(self, *args, **kwargs):
...
+def requires_x64_cpu(func):
+  assert callable(func)
+
+  @wraps(func)
+  def decorated(self, *args, **kwargs):
+    if platform.machine().lower() not in ['i686', 'x86_64', 'amd64']:
+      return self.skipTest(f'This test requires a native x86-64 CPU. Current CPU is {platform.machine()}.')
+    return func(self, *args, **kwargs)
+
+  return decorated
```
sbc100, line 265: If this is only used in core.py let's just put it in core.py

test/common.py
```diff
@@ -252,6 +253,18 @@ def decorated(self, *args, **kwargs):
   return decorated
 
 
+def requires_x64_cpu(func):
+  assert callable(func)
+
+  @wraps(func)
+  def decorated(self, *args, **kwargs):
+    if platform.machine().lower() not in ['i686', 'x86_64', 'amd64']:
```
sbc100, line 261: If we are including i686 I guess we should call this function requires_x86.  However emscripten doesn't run on x86-32 so maybe drop the i686?

PR #24439, opened 2025-05-29
Fix test_emsize under windows. NFC

sbc100:
Split out from #24437


dschuff (approved):

juj:
Great. I observe on my Windows box that pulling to branch that has this fix does not unfortunately yet fix the issue, but the test `other.test_emsize` still fails.

Running
```
del test\other\test_emsize.js
git checkout -- test\other\test_emsize.js
```
performs a re-checkout of the file, and after that the test starts to pass. Thought that's ok I think.

Issue #24438, opened 2025-05-29
FetchFS: `getFileRange` issues out-of-bounds chunk requests


foxtacles:
**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.9 (bbf1caa6e24f64fca9eb6a13a9e02d3f42123e77)
clang version 21.0.0git (https:/github.com/llvm/llvm-project 2f05451198e2f222ec66cec4892ada0509519290)
Target: wasm32-unknown-emscripten
Thread model: posix
```

The [`getFileRange` function](https://github.com/emscripten-core/emscripten/blob/1d35dfbd0a48a48c203da01bb5c1f8ec3a6e2047/src/lib/libwasmfs_fetch.js#L23) in FetchFS currently allows requests for chunks outside the file boundaries. The server will then respond with `416 Range Not Satisfiable`, which is properly handled in the code but I think we should save ourselves the request in the first place.

I've applied the attached patch to resolve this problem. I'd submit a pull request but I can't find instructions on how to run tests / get started with a dev environment.

CC @JoeOsborn 

[bounds.patch](https://github.com/user-attachments/files/20512114/bounds.patch)

sbc100:
I opened https://github.com/emscripten-core/emscripten/pull/24442 with this patch.  But we really should write a test too.

Do either of you know how to re-produce the issue and make a test?

sbc100:
If you are referring to test server using by the emscripten test suite that seems like something we should be able to fix (re generating 416 error).

The fact that you cannot make it happen using normal read calls is more worrying.  If you are able to reproduce in your full app can you capture a full backtrace so we can see where call comes from.

PR #24442, opened 2025-05-29
[wasmfs] FetchFS: fix getFileRange out-of-bounds issues

sbc100:
This change is authored by @foxtacles (See #24438)

Fixes: #24438

### DIFF
--- a/src/lib/libwasmfs_fetch.js
+++ b/src/lib/libwasmfs_fetch.js
@@ -90,6 +90,12 @@ addToLibrary({
       // one request for all the chunks we need, rather than one
       // request per chunk.
       var start = firstChunk * chunkSize;
+
+      // Out of bounds. No request necessary.
+      if (start >= wasmFS$JSMemoryRanges[file].size) {
+        return Promise.resolve();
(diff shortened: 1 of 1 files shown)


src/lib/libwasmfs_fetch.js
```diff
@@ -90,6 +90,12 @@ addToLibrary({
       // one request for all the chunks we need, rather than one
       // request per chunk.
       var start = firstChunk * chunkSize;
+
+      // Out of bounds. No request necessary.
+      if (start >= wasmFS$JSMemoryRanges[file].size) {
+        return Promise.resolve();
```
sbc100, line 96: I think we can just use `return` here actually.  See https://github.com/emscripten-core/emscripten/pull/24443

kripken (approved):

foxtacles:
More context: https://github.com/emscripten-core/emscripten/issues/24438#issuecomment-2920971818

I'd suggest holding off merging since until there's been more investigation. I'll update you as soon as I had the chance to dig further.

PR #24443, opened 2025-05-29
[wasmfs] FetchFS: Remove unnecessary Promise.resolve() calls. NFC

sbc100:


tlively (approved):

JoeOsborn:
Sorry I missed this. The calls were there to match what I thought the JS in-memory FS implementation was doing. If they’re not load-bearing then they should be removed.

PR #24444, opened 2025-05-30
Fix test_browser_language_detection after emsdk node update

sbc100:
The new emsdk version of node seems to report `en_US` in `navigator.languages` (even though the CI machines only have C and POSIX locales installed).


sbc100:
The other failure he is fixed in separate PR: #24447 .

Landing this TBR to green the tree.

PR #24449, opened 2025-05-30
Use node v24 for testing wasm64 and esm integration. NFC

sbc100:


brendandahl:
Alternatively, instead of pinning it at 24, we could call this `node-current` and keep bumping it?

sbc100:
Done.

sbc100:
OK, #24471 and #24465 and #24463 are now all landed separately so this PR should now be much simpler.

.circleci/config.yml
```diff
@@ -1070,6 +1077,7 @@ jobs:
       EMTEST_SKIP_SCONS: "1"
       EMTEST_SKIP_RUST: "1"
       EMTEST_SKIP_NODE_CANARY: "1"
+      EMTEST_SKIP_ESM_INTEGRATION: "1"
```
kripken, line 1080: Where is this not skipped, after this PR?

sbc100: Anywhere that has node 24 or above installed

sbc100: We could replace this with `EMTEST_SKIP_SKIP_NODE24` I suppose but the general approach so far seems to be per-feature flags.

sbc100:
Actually it looks like we can simplify things by landing this first: https://github.com/emscripten-core/emscripten/pull/24478

PR #24451, opened 2025-05-30
Fix test other.test_abspaths

juj:
Python 3.13 changed behavior so that `os.path.isabs('/foo/bar')` now returns `False` when run on Windows.


emcc.py
```diff
@@ -1295,7 +1295,7 @@ def consume_arg_file():
       exit_with_error('cannot --separate-asm with the wasm backend, since not emitting asm.js')
     elif arg.startswith(('-I', '-L')):
       path_name = arg[2:]
-      if os.path.isabs(path_name) and not is_valid_abspath(options, path_name):
+      if (path_name.startswith('/') or os.path.isabs(path_name)) and not is_valid_abspath(options, path_name):
```
sbc100, line 1298: Maybe worth a comment here mentioning the python change/weirdness.

sbc100 (approved):

PR #24383, opened 2025-05-20
Mark ASYNCIFY_LAZY_LOAD_CODE as deprecated

sbc100:
This setting was added as an experiment a long time ago and as far we know has no active users.

PR #24460, opened 2025-06-02
Reduce creating wrappers when instrumenting exported functions from wasm

Mintyboi:
I'm using JSPI in my application and currently, a javascript wrapper is created for all exports from each module (main / side) in `instrumentWasmExports()`.
Hence, whenever we make calls across the code boundaries, we will go through wasm-to-js.

Maybe we can create wrappers only for the required exports instead of for all?

### DIFF
--- a/src/lib/libasync.js
+++ b/src/lib/libasync.js
@@ -150,7 +150,7 @@ addToLibrary({
     instrumentWasmExports(exports) {
 #if EMBIND_GEN_MODE
       // Instrumenting is not needed when generating code.
-      return;
+      return exports;
 #endif
 #if ASYNCIFY_DEBUG
       dbg('asyncify instrumenting exports');
@@ -159,24 +159,25 @@ addToLibrary({
       var exportPattern = {{{ new RegExp(`^(${ASYNCIFY_EXPORTS.join('|').replace(/\*/g, '.*')})$`) }}};
       Asyncify.asyncExports = new Set();
 #endif
+      var ret = {};
       for (let [x, original] of Object.entries(exports)) {
+        ret[x] = original;
+
         if (typeof original == 'function') {
-          var wrapper = Asyncify.instrumentFunction(original);
 #if ASYNCIFY == 2
           // Wrap all exports with a promising WebAssembly function.
           let isAsyncifyExport = exportPattern.test(x);
           if (isAsyncifyExport) {
             Asyncify.asyncExports.add(original);
             original = Asyncify.makeAsyncFunction(original);
-            exports[x] = wrapper;
+            ret[x] = Asyncify.instrumentFunction(original);
           }
 #else
-          exports[x] = wrapper;
-       } else {
-          exports[x] = original;
-#endif
+          ret[x] = Asyncify.instrumentFunction(original);
         }
+#endif
       }
+      return ret;
     },
 
 #if ASYNCIFY == 1
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -769,7 +769,7 @@ var LibraryDylink = {
         updateTableMap(tableBase, metadata.tableSize);
         moduleExports = relocateExports(instance.exports, memoryBase);
 #if ASYNCIFY
-        Asyncify.instrumentWasmExports(moduleExports);
+        moduleExports = Asyncify.instrumentWasmExports(moduleExports);
 #endif
         if (!flags.allowUndefined) {
           reportUndefinedSymbols();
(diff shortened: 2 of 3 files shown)


src/lib/libasync.js
```diff
@@ -159,31 +159,24 @@ addToLibrary({
...
-#if ASYNCIFY_LAZY_LOAD_CODE
-          original.exportName = x;
-#endif
-          ret[x] = wrapper;
-
+#else
+          exports[x] = wrapper;
        } else {
-          ret[x] = original;
+          exports[x] = original;
```
sbc100: If we are modifying the exports list in-place then I would imagine this line is not needed.

Mintyboi:
> When calls are made from one wasm module to another I don't think we use the wrappers.
> 
> Each wrapper has `wrapper.orig = original;` attached to it and libdylink.js does symbol resolution is always uses the `.orig` versions when supplying symbols to native code.

You're right, I do see the symbol resolution happening in `dlsym_js` when we're calling a function in the side module from the main module.
However when we're making a call into main from side, we pass the wasmExport object (which contains the wrapped export) when doing a `loadDynamicLibrary`. I guess that's where the wrappers will be used.

> In other words, while we create JS wrappers for all function those wrappers should only ever get called when calling into wasm from JS. Are you seeing behaviour different to this?

I am using v3.1.61 and am seeing this behavior even up till 4.0.3.
But interestingly, I do not see this happening on latest anymore, even thou the code remains the same.

I still hope to get this in my patched 3.1.61 as it improves my application's performance a little. Do you see any other workflows that I should test out?

sbc100:
I see a lot of test failures that looks like they probably related to this PR.  Are you able to reproduce them locally?

src/lib/libdylink.js
```diff
@@ -769,7 +769,7 @@ var LibraryDylink = {
         updateTableMap(tableBase, metadata.tableSize);
         moduleExports = relocateExports(instance.exports, memoryBase);
 #if ASYNCIFY
-        moduleExports = Asyncify.instrumentWasmExports(moduleExports);
+        Asyncify.instrumentWasmExports(moduleExports);
```
sbc100: Is converting `instrumentWasmExports` to mutate-in-place an intrinsic part of this change?   Or is that a separate optimization you are attempting?  

I think there could be issue with that because I'm not sure that actual export object of a wasm module is mutable like that.

Mintyboi: It does seem to replace it just fine when I stepped through the code. However from https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/Instance/exports, it sounds like it should be a read-only object.

To be on the safe side, let me create a new object just like the previous code.

Mintyboi:
> I see a lot of test failures that looks like they probably related to this PR. Are you able to reproduce them locally?

Yes it should be due to my changes, I'll be sure to look into it once we feel that these changes are safe to land.
wdyt?

sbc100:
It sounds like the problem is that we are wrapping all the exports, but that we are using the wrappers when importing directly into wasm.   

I wonder if there is some way we can show that this happens in practice and when it happens, then we can work on *not* using the wrappers in that case.

src/lib/libasync.js
```diff
@@ -161,28 +161,22 @@ addToLibrary({
...
-#endif
-          ret[x] = wrapper;
-
-       } else {
-          ret[x] = original;
+#else
+          ret[x] = Asyncify.instrumentFunction(original);
         }
-      }
-#if ASYNCIFY_LAZY_LOAD_CODE
```
sbc100, line 183: It looks like you are deleting a bunch of stuff relating to `ASYNCIFY_LAZY_LOAD_CODE` ?   Does the lazy_load_code test still pass?

BTW, I'm planning on completely removing `ASYNCIFY_LAZY_LOAD_CODE` soon so maybe this will be simlper once that happens?

Mintyboi:
> It sounds like the problem is that we are wrapping all the exports, but that we are using the wrappers when importing directly into wasm.
> 
> I wonder if there is some way we can show that this happens in practice and when it happens, then we can work on _not_ using the wrappers in that case.

I agree. I was looking through the code to understand why it doesn't happen anymore and I see this scenario getting fixed in 
https://github.com/emscripten-core/emscripten/pull/23581.
The proxyhandler now uses the .orig versions when supplying symbols to native code as you mentioned, so this is not an issue anymore moving forward.

I'll be closing this now and thank you for your involvement in this discussion @sbc100.

src/lib/libasync.js
Mintyboi: Yep, I saw your [PR](https://github.com/emscripten-core/emscripten/pull/24383) a few days back and decided that it was too cumbersome to tackle it in this discussion 😅

PR #24461, opened 2025-06-02
[dylink] Fix execution order in emscripten_dlopen_promise

ryanking13:
This PR changes the order of promise execution in `emscripten_dlopen_promise` to prevent the promise from being destroyed when resolving happens.

Current behavior of `emscripten_dlopen_promise` is:

1. Call `emscripten_dlopen` with callbacks (with promise `p`)
1.a. These callbacks destroy the promise `p` when the dlopen is finished.
2. Call `emscripten_promise_resolve` with another promise `ret`, which resolves to `p`.

The problem happens when 1.a. is called before 2. The success (or error) callback will destroy `p`, and the `emscripten_promise_resolve` failes to resolve the promise for `p` because it is already destroyed.

This PR fixes it by calling `emscripten_promise_resolve` then `emscripten_dlopen` to ensure the promise for `p` always exists when calling `emscripten_promise_resolve`

I am not sure how to test this behavior, as this it related to how JS schedules the promise, but in Pyodide, we experienced this issue (https://github.com/pyodide/pyodide/pull/5610#issuecomment-2926336938), and this patch fixes it.

### DIFF
--- a/system/lib/libc/dynlink.c
+++ b/system/lib/libc/dynlink.c
@@ -656,12 +656,14 @@ em_promise_t emscripten_dlopen_promise(const char* filename, int flags) {
   // Create a promise that is resolved (and destroyed) once the operation
   // succeeds.
   em_promise_t p = emscripten_promise_create();
-  emscripten_dlopen(filename, flags, p, promise_onsuccess, promise_onerror);
-
   // Create a second promise bound the first one to return the caller.  It's
   // then up to the caller to destroy this promise.
   em_promise_t ret = emscripten_promise_create();
+  
+  // The order matters here. Calling emscripten_dlopen before resolving the second promise
+  // may destroy the first promise before resolving the value.
   emscripten_promise_resolve(ret, EM_PROMISE_MATCH, p);
+  emscripten_dlopen(filename, flags, p, promise_onsuccess, promise_onerror);
   return ret;
 }
 


hoodmane:
Could you test this by calling `emscripten_dlopen_promise` on an already loaded library?

system/lib/libc/dynlink.c
```diff
@@ -656,12 +656,14 @@ em_promise_t emscripten_dlopen_promise(const char* filename, int flags) {
...
   em_promise_t p = emscripten_promise_create();
-  emscripten_dlopen(filename, flags, p, promise_onsuccess, promise_onerror);
-
   // Create a second promise bound the first one to return the caller.  It's
   // then up to the caller to destroy this promise.
   em_promise_t ret = emscripten_promise_create();
+  
+  // The order matters here. Calling emscripten_dlopen before resolving the second promise
+  // may destroy the first promise before resolving the value.
   emscripten_promise_resolve(ret, EM_PROMISE_MATCH, p);
```
sbc100, line 665: When this bug occurs and `emscripten_dlopen` wins the race, IIRC it means that `p` is already resolved by the time we call `emscripten_promise_resolve(ret, EM_PROMISE_MATCH, p)`?

Are you saying that `p` is no longer a valid promise handle at the point?   i.e. the entry in the HandleAllocator map on the JS side is no longer valid?

Or are you saying that `p` is valid but because it already been resolve its somehow too late to link `ret` to it?

sbc100: Oh I see `emscripten_promise_destroy` is called by the `onsuccess` / `onerror` callback.

In that case does the error manifest in crash in `emscripten_promise_resolve` here because `p` is not valid?

hoodmane: Well the problem is that `promise_onsuccess` and `promise_onerror` destroy the promise when they are done. If `emscripten_dlopen()` completes synchronously, then it destroys the promise right away. Then `emscripten_promise_resolve()` is too late.

sbc100:
Hm, is the problem that `onsuccess` and `onerror` can sometimes be called before `emscripten_dlopen` returns?  I think we should maybe fix that instead.  Is bad practice to have callbacks that sometimes fire synchronously.

ryanking13:
>Hm, is the problem that onsuccess and onerror can sometimes be called before emscripten_dlopen returns? I think we should maybe fix that instead. Is bad practice to have callbacks that sometimes fire synchronously.

Thanks @sbc100! I'll close this for now, then.

PR #24463, opened 2025-06-02
[test] Extract `try_require_node_version` helper function. NFC

sbc100:


sbc100:
Split out from #24449

test/common.py
```diff
@@ -1038,12 +1038,8 @@ def require_wasm64(self):
     if self.is_browser_test():
       return
 
-    nodejs = self.get_nodejs()
-    if nodejs:
-      version = shared.get_node_version(nodejs)
-      if version >= (24, 0, 0):
-        self.js_engines = [nodejs]
-        return
+    if self.try_require_node_version((24, 0, 0)):
```
kripken: Maybe using `*args` could allow writing `(24, 0, 0)` instead of `((24, 0, 0))`?

sbc100: Good idea

PR #24465, opened 2025-06-02
[test] Remove redundant calls to `require_node_canary`. NFC

sbc100:
The wasm64 tests modes already require this.

Split out from #24449


kripken:
> The wasm64 tests modes already require this.

Where can I find that? I looked at `require_wasm64` and other places but can't see it.

sbc100:
`require_wasm64` ensures node 24 (which until recently was what canary way).

There are no wasm64 feature that are in canary (25) that are not in node 24 (that I know of)

kripken:
Thanks, now I understand. I was looking for canary specifically.

PR #24478, opened 2025-06-03
Remove node canary test decorators. NFC

sbc100:
All these features actually work on the official node 24 these days.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -524,7 +524,7 @@ jobs:
   test-core0:
     executor: focal
     environment:
-      EMTEST_SKIP_NODE_CANARY: "1"
+      EMTEST_SKIP_NODE24: "1"
     steps:
       - run-tests-linux:
           test_targets: "core0"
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -524,15 +524,15 @@ jobs:
   test-core0:
     executor: focal
     environment:
-      EMTEST_SKIP_NODE_CANARY: "1"
+      EMTEST_SKIP_NODE24: "1"
```
kripken, line 527: Perhaps 24 => latest or such, so we can update it without changing all these places?

sbc100: The problem is that we don't want tests marked as `require_node_version(24)` to be disabled by `EMTEST_SKIP_NODE_LATEST` once latest becomes 25 or 26.

Unless you think we should also replace `require_node_version(24)` with just `require_node_latest`?

The approach here seems to me like the least complicated approach for now.   We can revisit later.

kripken: Fair enough, this does seem simple for now.

PR #24480, opened 2025-06-03
[test] Remove test_memory64_proxies. NFC

sbc100:
This tests is not actually testing anything anymore and it should have been removed as part of #20725.


kripken (approved):
lgtm, but do we have other memory64 tests that use >4gb memory? i can only find `browser64_4gb` (though maybe that is enough?)

sbc100:
> lgtm, but do we have other memory64 tests that use >4gb memory? i can only find `browser64_4gb` (though maybe that is enough?)

Yes we have the wasm64_4gb test mode which runs the whole suite with 4gm + memory.

PR #24486, opened 2025-06-04
Use higher-level string helpers where possible. NFC

RReverser:
I guess one functional change is in implementation of `emscripten_fetch_get_response_headers_length`.

The docs say that it just returns the length, and you must use `length + 1` for buffer size passed to `emscripten_fetch_get_response_headers`, and that's what tests do, but we were actually returning `length + 1` already. This seems like a bugfix, but happy to revert if deemed too risky.


src/lib/libwasmfs.js
```diff
@@ -507,19 +506,17 @@ addToLibrary({
...
-    var dataBuffer = _malloc(data.length);
+    var len = typeof data == 'string' ? lengthBytesUTF8(data) + 1 : data.length;
+    var dataBuffer = _malloc(len);
 #if ASSERTIONS
     assert(dataBuffer);
 #endif
-    for (var i = 0; i < data.length; i++) {
-      {{{ makeSetValue('dataBuffer', 'i', 'data[i]', 'i8') }}};
+    if (typeof data == 'string') {
+      len = stringToUTF8(data, dataBuffer, len);
```
sbc100, line 515: Does `len` need to be re-assigned here?  it looks like it will remain the same.

test/wasmfs/wasmfs_fetch.js
```diff
@@ -2,10 +2,7 @@ addToLibrary({
   getUrlOrigin__sig: 'vpp',
   getUrlOrigin: function (ptr, len) {
     try {
-      var orig = self.location.origin;
-      var nb = lengthBytesUTF8(orig) + 1;
-      if (nb <= len)
-        stringToUTF8(orig, ptr, len);
```
sbc100, line 8: Looks like the is a slight behavior change here., but probably fine?  (old code never writes truncated origin)

src/lib/libwasmfs.js
```diff
@@ -141,8 +141,7 @@ addToLibrary({
 
       // Default return type is binary.
       // The buffer contents exist 8 bytes after the returned pointer.
-      var ret = new Uint8Array(HEAPU8.subarray(buf, buf + length));
-      return opts.encoding === 'utf8' ? UTF8ArrayToString(ret) : ret;
+      return opts.encoding === 'utf8' ? UTF8ToString(buf, length) : HEAPU8.slice(buf, buf + length);
```
sbc100, line 144: Is `.slice` the same ias `new Uint8Array(HEAPU8.subarray...)`?

src/lib/libsdl.js
```diff
@@ -2904,8 +2901,8 @@ var LibrarySDL = {
       audio.frequency = info.audio.frequency;
     }
     audio['onended'] = function() { // TODO: cache these
-      if (channelInfo.audio === this || channelInfo.audio.webAudioNode === this) { 
-        channelInfo.audio.paused = true; channelInfo.audio = null; 
+      if (channelInfo.audio === this || channelInfo.audio.webAudioNode === this) {
+        channelInfo.audio.paused = true; channelInfo.audio = null;
```
sbc100, line 2905: I find these types of cleanups, when included in other changes, to be distracting during review, so in general I'd rather see them as separate PRs (no need to split up this one though).

BTW, I'm sure this is old new to you, but I recommend configuring a keyboard shortcut to trim whitespace in the whole file, rather than doing to automatically on all files you save.. that way you don't end up creating changes like these when working on projects that don't have good whitespace hygiene.

src/lib/libsdl.js
```diff
@@ -964,10 +964,7 @@ var LibrarySDL = {
         case 'keypress': {
           {{{ makeSetValue('ptr', C_STRUCTS.SDL_TextInputEvent.type, 'SDL.DOMEventToSDLEvent[event.type]', 'i32') }}};
           // Not filling in windowID for now
-          var cStr = intArrayFromString(String.fromCharCode(event.charCode));
-          for (var i = 0; i < cStr.length; ++i) {
-            {{{ makeSetValue('ptr', C_STRUCTS.SDL_TextInputEvent.text + ' + i', 'cStr[i]', 'i8') }}};
-          }
+          stringToUTF8(String.fromCharCode(event.charCode), ptr + {{{ C_STRUCTS.SDL_TextInputEvent.text }}}, 4);
```
sbc100, line 967: Nice!

src/Fetch.js
```diff
@@ -584,14 +584,12 @@ function startFetch(fetch, successcb, errorcb, progresscb, readystatechangecb) {
 }
 
 function fetchGetResponseHeadersLength(id) {
-  return lengthBytesUTF8(Fetch.xhrs.get(id).getAllResponseHeaders()) + 1;
+  return lengthBytesUTF8(Fetch.xhrs.get(id).getAllResponseHeaders());
```
sbc100, line 587: Doesn't this change the semantics of the function?  i.e. don't we want to include the null terminator in the size of the string here?

src/Fetch.js
```diff
@@ -584,14 +584,12 @@ function startFetch(fetch, successcb, errorcb, progresscb, readystatechangecb) {
...
 }
 
 function fetchGetResponseHeaders(id, dst, dstSizeBytes) {
   var responseHeaders = Fetch.xhrs.get(id).getAllResponseHeaders();
-  var lengthBytes = lengthBytesUTF8(responseHeaders) + 1;
-  stringToUTF8(responseHeaders, dst, dstSizeBytes);
-  return Math.min(lengthBytes, dstSizeBytes);
+  return stringToUTF8(responseHeaders, dst, dstSizeBytes) + 1;
```
sbc100, line 592: Especially since here we do include it, right?

src/lib/libwasmfs.js
RReverser: It will be `len - 1` as before the call it's length with nul character, but afterwards we just want the length of the string itself.

So it's either reassignment or doing `len--` manually, I figured reassignment is cleaner.

src/lib/libwasmfs.js
RReverser: Yup!

test/wasmfs/wasmfs_fetch.js
RReverser: Yeah, besides it's just a test anyway, it doesn't care about truncated inputs.

src/lib/libsdl.js
RReverser: > I'm sure this is old new to you, but I recommend configuring a keyboard shortcut to trim whitespace in the whole file, rather than doing to automatically on all files you save

Heh I remember we talked about that a while back. It's just extremely rare for me to encounter files with trailing spaces, so having it done automatically on save helps more than hurts, as it usually just removes spaces I introduced myself, but then there are exceptions like this one.

FWIW I almost always review Github diffs with "ignore whitespace" checked, it helps with the distraction aspect. But if you want, I can revert this one.

src/Fetch.js
RReverser: We don't, but yeah it does - I mentioned it in the PR description.

src/Fetch.js
RReverser: Same, see PR description. Judging by docs and tests, the intended behaviour was always for the Length function to just return the actual length, and it's responsibility of whoever calls `fetchGetResponseHeaders` to pass enough size for both length and the null terminator.

sbc100 (approved):
lgtm with a ChangeLog entry mentioning that `emscripten_fetch_get_response_headers_length` change (to match the documented and intended behaviour).

Issue #8805, opened 2019-06-15
UTF8String ends early if the string contains a NUL


d3x0r:
I have this simple utility function that takes a char* and length and make it a referencable JS object (puts it in an array, and returns the index of that element).  When converted from C U8HEAP to JS, if the string contains a `\0` character, the conversion is terminated, ignoring the length specified.

Also, related, but not critical, if the code contains a string with a \0 in it, (maybe I need to \\0 it?) the compilation fails (see end)

Sample C code...
``` C
#include <stdio.h>
#include <emscripten.h>


int makeString( char const*string, int stringlen ) EMSCRIPTEN_KEEPALIVE;
int makeString( char const*string, int stringlen ) {
	printf( "String: %5.5s\\0%5.5s  len:%d\n", string, string + 6, stringlen );
	int x = EM_ASM_INT( {
		const string = UTF8ToString( $0, $1 );
		return Module.objects.push( string )-1;
	},string, stringlen);
	return x;
}


int main( void ) {
	return EM_ASM_INT ( (
	        Module.objects = [];
        	var s = "Hello" + String.fromCodePoint( 0 ) + "World";

        	        // If the following line is used, the compile fails, which
        	        // is why the above uses fromCodePoint.
        	//var s = "Hello\0World";

        	console.log( "String:", s, s.length );
		var sa;
		var si = allocate( sa= intArrayFromString(s), 'i8', ALLOC_NORMAL);
                var index = Module._makeString( si, sa.length-1 );  // pass length of string, not byte size

		console.log( "Resulting string:", Module.objects[index], Module.objects[index].length );

                return 0;
        ) );
}

```

Compile with simple `emcc  -o ./nultest.js nultest.c`

Generated output, the error is that the result should also be 11 characters long.
(The string received 
```
String: Hello World 11
String: Hello\0World  len:11
Resulting string: Hello 5
```

Expected output... the output should be the same as the input.
```
String: Hello World 11
String: Hello\0World  len:11 
Resulting string: Hello World 11
```

---

```

M:\sack\amalgamate\fs_vfs\nultest>call emcc  -g -o ./nultest.js nultest.c 
emscripten:ERROR: emscript: failure to parse metadata output from compiler backend. raw output is: 

{
"staticBump": 3984,
[...]

PR #24487, opened 2025-06-04
Add NUL-ignoring mode to UTF*ToString functions

### LINKED ISSUE #12517: performance issue with UTF8ToString
My code frequently passes a string from C++ to JS. To do this I used the `UTF8ToString` built-in function.

Something like this:
```C++

EM_JS(void, consumeAString(const char* str, size_t length), {
  const string = UTF8ToString(str, length);
  typeof string === 'string';
  // do with string;
});

void myFunc() {
  std::string str;
  consumeAString(str.c_str(), str.size());
}

```

However, I found this code executes slower than expected (chrome profile shows 85% of total execution time). So I looked into the code.

call chain: `UTF8ToString` -> `UTF8ArrayToString`

And `UTF8ArrayToString` basically has such logic:

```js
function UTF8ArrayToString(heap, idx, maxBytesToRead) {
  var endIdx = idx + maxBytesToRead;
  var endPtr = idx;

  // (1)  find the '\0' terminator -- this is slow
  while (heap[endPtr] && !(endPtr >= endIdx)) ++endPtr;

  if (endPtr - idx > 16 && heap.subarray && UTF8Decoder) {
    // (2) do decode
    return UTF8Decoder.decode(heap.subarray(idx, endPtr));

...
```

Turns out that step (1) spend as much as (or even more than) step (2).

### So I wrote a simple test code.

<details><summary>test code</summary>
<p>

```js
const testIteration = 100;
const decoder = new TextDecoder('utf-8');

function test(length) {
    const buffer = new Int8Array(length);
    for (let i = 0; i < length; i++) {
        buffer[i] = 65; // ascii 'A'
    }

    function decode() {
        decoder.decode(buffer);
    }

    function iterate() {
        let i = 0;
        while(i < length && buffer[i]) i++;
    }

    let start = Date.now();
    for (let i = 0; i < testIteration; i++) {
        decode();
    }
    const decodeTime = (Date.now() - start) / testIteration;

    start = Date.now();
    for (let i = 0; i < testIteration; i++) {
        iterate();
    }
    const iterateTime = (Date.now() - start) / testIteration;

    console.log("TypedArray length:[" + length + "] decode:" + decodeTime + "ms" + " iterate:" + iterateTime + "ms");
}

test(1024);

test(1024 * 1024);

test(10 * 1024 * 1024);

```

</p>
</details>

[Online test code on repl.it](https://repl.it/talk/share/Emscripten-UTF8ArrayToString-performance-test/57886)

On my Chrome 85 @macOs 10.15 i9-9900K environment, it shows:

```
TypedArray length:[1024] decode:0ms iterate:0.01ms
TypedArray length:[1048576] decode:1.11ms iterate:0.91ms
TypedArray length:[10485760] decode:11.54ms iterate:8.96ms
```

### Plus: I also tested `strlen` performance in C (Wasm) and JS

<details>
<summary> StrlenTest.cc </summary>

<p>

```cpp

#include <iostream>
#include <chrono>

using namespace std;
using namespace std::chrono;

EM_JS(void, strlenTestJS, (const char* str), {
    function strlen(str) {
        let end = str;
        while (Module.HEAP8[end]) end++;
        return end - str;
    }

    for (let i = 0; i < 100; i++) {
        strlen(str);
    }
});

void strlenTestC(const char* str) {
    for (int i = 0; i < 100; i++) {
         std::strlen(str);
    }
}

void test(size_t length) {
    std::string str;
    str.resize(length, 'A');

    auto start = steady_clock::now().time_since_epoch();
    strlenTestC(str.c_str());
    auto c = steady_clock::now().time_since_epoch() - start;

    start = steady_clock::now().time_since_epoch();
    strlenTestJS(str.c_str());
    auto js = steady_clock::now().time_since_epoch() - start;

    cout << "length:" << length
        << " C:" << duration_cast<milliseconds>(c).count() / 100 << "ms"
        << " JS:" << duration_cast<milliseconds>(js).count() / 100 << "ms"
        << std::endl;
}

int main() {
    test(1024);
    test(1024 * 1024);
    test(10 * 1024 * 1024);
    test(50 * 1024 * 1024);
}


```
</p>
</details>

run with node.js
```
# em++ test.cpp -sALLOW_MEMORY_GROWTH=1
# Build with `-O0`
length:1024 C:0ms JS:0ms
length:1048576 C:0ms JS:0ms
length:10485760 C:1ms JS:6ms
length:52428800 C:9ms JS:37ms

# em++ test.cpp -O3 -sALLOW_MEMORY_GROWTH=1
# Build with `-O3`
length:1024 C:0ms JS:0ms
length:1048576 C:0ms JS:0ms
length:10485760 C:0ms JS:6ms
length:52428800 C:0ms JS:37ms
```

---
## Known facts:

1. use js to iterate TypeArray is slow
2. use C/Wasm to calculate `strlen` is really fast

---
## Discuss
`UTF8ToString` already takes a `maxBytesToRead` argument, but it is not really the string length, so, inside it, string length is still needed to be calculated.

C/C++ code already knows (or can be really fast to know) the actual length of a string.

So, a simple solution is to pass string length to `UTF8ToString`.

---
## Possible solutions:

Solutions are simple.

1. change the  `maxBytesToRead` semantic to `stringLength`
this breaks backward compatibility.

2. add a third argument named `stringLength`

PS: there is also `UTF16ToString`, `UTF32ToString`.

RReverser:
- Significantly simplifies Embind string implementation (and reduces code size) by avoiding the manual "read chunks between NUL characters" hackery.
- Closes #12517.
- Also relevant old issue: #8805.


site/source/docs/api_reference/preamble.js.rst
```diff
@@ -159,12 +159,13 @@ Accessing memory
 Conversion functions — strings, pointers and arrays
 ===================================================
 
-.. js:function:: UTF8ToString(ptr[, maxBytesToRead])
+.. js:function:: UTF8ToString(ptr[, maxBytesToRead], [ignoreNul])
```
sbc100, line 162: Null normally has two l's doesn't it?  

Did you consider the inverse?  i.e. `stopOnNull = true`?

site/source/docs/api_reference/preamble.js.rst
```diff
@@ -202,11 +205,13 @@ Conversion functions — strings, pointers and arrays
 
 
 
-.. js:function:: UTF32ToString(ptr)
+.. js:function:: UTF32ToString(ptr[, maxBytesToRead], [ignoreNul])
 
   Given a pointer ``ptr`` to a null-terminated UTF32LE-encoded string in the Emscripten HEAP, returns a copy of that string as a JavaScript ``String`` object.
 
   :param ptr: A pointer to a null-terminated UTF32LE-encoded string in the Emscripten HEAP.
+  :param maxBytesToRead: An optional length that specifies the maximum number of bytes to read. You can omit this parameter to scan the string until the first \0 byte. If maxBytesToRead is passed, and the string at ``[ptr, ptr+maxBytesToReadr)`` contains a null byte in the middle, then the string will cut short at that byte index (i.e. maxBytesToRead will not produce a string of exact length ``[ptr, ptr+maxBytesToRead)``) N.B. mixing frequent uses of ``UTF8ToString()`` with and without maxBytesToRead may throw JS JIT optimizations off, so it is worth to consider consistently using one style or the other.
```
sbc100: The comment mentions `UTF8ToString` here which I assume should be changed to `UTF32ToString` here?

Maybe we can rephrase to say "this function" instead?

Also maybe remove the ("i.e. maxBytesToRead will not produce a string of exact length ``[ptr, ptr+maxBytesToRead)``) since it seems obvious from the proceeding sentence?

Also, is the last sentence still accurate, and worth mentioning?  Maybe just remove that too?

src/lib/libembind.js
```diff
@@ -475,33 +475,17 @@ var LibraryEmbind = {
       // For some method names we use string keys here since they are part of
       // the public/external API and/or used by the runtime-generated code.
       'fromWireType'(value) {
-        var length = {{{ makeGetValue('value', '0', SIZE_TYPE) }}};
+        var length = {{{ makeGetValue('value', '0', '*') }}};
```
sbc100, line 478: Why change this?   I think the two are exactly the equivalent, right?

src/lib/libembind.js
```diff
@@ -475,33 +475,17 @@ var LibraryEmbind = {
...
-              decodeStartPtr = currentBytePtr + 1;
-            }
-          }
+          str = UTF8ToString(payload, length, true);
         } else {
-          var a = new Array(length);
+          str = '';
           for (var i = 0; i < length; ++i) {
-            a[i] = String.fromCharCode(HEAPU8[payload + i]);
+            str += String.fromCharCode(HEAPU8[payload + i]);
```
sbc100, line 487: The old code looks like it was trying to optimize for doing a lot of single char `+=`?  Are you sure thats not useful anymore?  Seems like of separate to the rest of this change.

src/lib/libembind.js
```diff
@@ -567,41 +551,22 @@ var LibraryEmbind = {
...
       decodeString = UTF16ToString;
       encodeString = stringToUTF16;
       lengthBytesUTF = lengthBytesUTF16;
-      readCharAt = (pointer) => {{{ makeGetValue('pointer', 0, 'u16') }}};
     } else if (charSize === 4) {
       decodeString = UTF32ToString;
       encodeString = stringToUTF32;
       lengthBytesUTF = lengthBytesUTF32;
-      readCharAt = (pointer) => {{{ makeGetValue('pointer', 0, 'u32') }}};
     }
```
sbc100, line 566: Looks like this `else if` can/should be replaced with `else` + `assert(charSize === 4)`?

src/lib/libstrings.js
```diff
@@ -398,16 +418,17 @@ addToLibrary({
...
-  $UTF32ToString: (ptr, maxBytesToRead) => {
+  $UTF32ToString: (ptr, maxBytesToRead, ignoreNul) => {
 #if ASSERTIONS
     assert(ptr % 4 == 0, 'Pointer passed to UTF32ToString must be aligned to four bytes!');
 #endif
     var str = '';
+    var startIdx = {{{ getHeapOffset('ptr', 'u32') }}};
     // If maxBytesToRead is not passed explicitly, it will be undefined, and this
     // will always evaluate to true. This saves on code size.
     for (var i = 0; !(i >= maxBytesToRead / 4); i++) {
```
sbc100, line 426: If you are doing to extract `startIdx` from the loop body here maybe also extract the last index calculation (maxBytesToRead / 4)?

Do you know why this is writtend as `!(i >= maxBytesToRead / 4)` rather than `i < maxBytesToRead`?  Something to do with how undefined works maybe?

sbc100: Maybe split this optimization out too?

site/source/docs/api_reference/preamble.js.rst
RReverser: > Null normally has two l's doesn't it?

I've often seen the character specifically called `NUL`, Wikipedia seems to agree:

> In [software documentation](https://en.wikipedia.org/wiki/Software_documentation), the null character is often represented with the text NUL (or NULL although that may mean the [null pointer](https://en.wikipedia.org/wiki/Null_pointer)).

(c) https://en.wikipedia.org/wiki/Null_character

RReverser: > Did you consider the inverse? i.e. `stopOnNull = true`?

Yeah, just wasn't sure if it's okay, since it then requires removing the hacky code size saving for `maxBytesToRead` as it will also need an explicit default and will add few more bytes.

IMO it would be still worth it, but I know we sometimes bikeshed even over a few bytes so tried to avoid it.

site/source/docs/api_reference/preamble.js.rst
RReverser: > The comment mentions `UTF8ToString` here which I assume should be changed to `UTF32ToString` here?

Oops, yeah, copy-paste error.

src/lib/libembind.js
RReverser: Seems not, one produces a number and another a bigint and without this change I'd get an error about mixing them up when trying to add `idx + length`. We already do the same in `embind_register_std_wstring` and some other places, so I assume this one exception not using `*` was the wrong one.

src/lib/libembind.js
RReverser: Yeah more of a drive-by change, but I'm pretty sure it's not worth it. Produces extra garbage for GC as well as requires a bit more code.

src/lib/libstrings.js
RReverser: > Do you know why this is writtend as `!(i >= maxBytesToRead / 4)` rather than `i < maxBytesToRead`? Something to do with how undefined works maybe?

Yeah this optimisation is all over the place. I'm not a huge fan of it for both code readability reasons and for "confusing JIT by comparing integers with undefined" reason, and would rather replace with a proper default, but as the comment says it saves a couple of bytes so idk.

Maybe let's remove in a separate PR if you're on board with it.

site/source/docs/api_reference/preamble.js.rst
RReverser: > as it will also need an explicit default

Ah no, I got confused with semantics of other languages. In JS the default value can be in any position, not just in last params. Huh.

RReverser: I pushed a new commit that inverts the param. It adds a few more bytes - let me know if you think it's worth it and, if not, I can easily revert it.

src/lib/libstrings.js
sbc100: Yes, lets leave that code as is for this PR.

src/lib/libembind.js
sbc100: Hmm.. that seems like a bug.  I would hope that `SIZE_TYPE`, `POITNER_TYPE` and `*` all produce the same thing but you are right it looks the first two return bigint on wasm64 :(

sbc100: I am confused why this isn't already broken though?  I would that though this would break with any wasm64 caller?

site/source/docs/api_reference/preamble.js.rst
sbc100: Regarding Nul, so we requite just a single null byte in call cases or are we checking for a 16-bit and 32-bit nul in the case of 16 and 32 bit encodings?

src/lib/libembind.js
RReverser: > I am confused why this isn't already broken though?

It's a subtle difference, but it's because before the code was doing `for (var i = 0; i < length;...)` and you're allowed to compare numbers (i) and bigints (length) but now we're doing `maxIdx = idx + length` first, and, unlike relation ops, arithmetic operations are not allowed to mix those types.

RReverser: > Hmm.. that seems like a bug. I would hope that `SIZE_TYPE`, `POITNER_TYPE` and `*` all produce the same thing

I thought it was intentional that we have separate types for "give me bigint if on wasm64" and "give me a number regardless of actual integer size", but yeah, I agree it's confusing.

To make matters more confusing, we also have i53/u53, which for some reason have a separate implementation from the `*` type on wasm64 even though IIUC they have exactly same semantics  ("64-bit integer represented as a JS number")?

site/source/docs/api_reference/preamble.js.rst
RReverser: > are we checking for a 16-bit and 32-bit nul in the case of 16 and 32 bit encodings?

We are, yes. No change from previous behaviour.

src/lib/libembind.js
sbc100: SIZE_TYPE, POINTER_TYPE and * should certainly all do the same thing on wasm64 which is give you Number and not a bigint.

site/source/docs/api_reference/preamble.js.rst
RReverser: > I pushed a new commit that inverts the param. It adds a few more bytes - let me know if you think it's worth it and, if not, I can easily revert it.

@sbc100 Thoughts on this?

sbc100: I don't think it makes much a huge difference.   Lets maybe just leave it how your originally had it with  [ignoreNul] (without the default param).

RReverser: Okay, I'll revert.

src/lib/libembind.js
RReverser: Could be, although that would be another drive-by change unrelated to the PR.

src/lib/libembind.js
RReverser: Yeah, figured by now from discussions on the general POINTER_TYPE issue. I'm marking this discussion as resolved then.

src/lib/libembind.js
RReverser: Done.

RReverser:
Other than rebaselining again, anything left to do here?

site/source/docs/api_reference/preamble.js.rst
```diff
@@ -159,12 +159,13 @@ Accessing memory
...
-  :param maxBytesToRead: An optional length that specifies the maximum number of bytes to read. You can omit this parameter to scan the string until the first \0 byte. If maxBytesToRead is passed, and the string at ``[ptr, ptr+maxBytesToReadr)`` contains a null byte in the middle, then the string will cut short at that byte index (i.e. maxBytesToRead will not produce a string of exact length ``[ptr, ptr+maxBytesToRead)``) N.B. mixing frequent uses of ``UTF8ToString()`` with and without maxBytesToRead may throw JS JIT optimizations off, so it is worth to consider consistently using one style or the other.
+  :param maxBytesToRead: An optional length that specifies the maximum number of bytes to read. You can omit this parameter to scan the string until the first \0 byte. If maxBytesToRead is passed, and the string at ``[ptr, ptr+maxBytesToReadr)`` contains a null byte in the middle, then the string will cut short at that byte index. N.B. mixing frequent uses of this function with and without maxBytesToRead may throw JS JIT optimizations off, so it is worth to consider consistently using one style or the other.
```
sbc100, line 167: Should we just drop the last sentence here?

It doesn't seem very actionable, since most of the usage of this function is internal anyway (and outside of the user's control).

I'm also not sure how true it is with engines that can do polymorphic/megamorphic stuff.

sbc100 (approved):
lgtm!

site/source/docs/api_reference/preamble.js.rst
RReverser: I agree, didn't want to touch it in this PR but also think it's a bit useless.

sbc100: Follow is fine too.  lgtm either way.

RReverser:
Those browser flakes... 🥵 I swear, they've gotten a lot worse lately...

sbc100:
> Those browser flakes... 🥵 I swear, they've gotten a lot worse lately...

The firefox audio ones are certainly new (if those are the ones you are seeing) since we only recently started running audio tests in headless mode.

RReverser:
No, usually pthread ones + redirect test that relies on httpbin.

sbc100:
redirect test is very new.. we should look into that one for sure.

RReverser:
What's going on here with `emsdk` not being found by the script on CI? https://github.com/emscripten-core/emscripten/actions/runs/15496117161/job/43633140840?pr=24487

````
The following (1) test expectation files were updated by

```
test/browser/test_small_js_flags.js.size updated
```

Test expectations are out-of-date on the PR branch.
You can run './tools/maint/rebaseline_tests.py' to
create a commit updating the expectations.
/home/runner/work/_temp/14e35ba4-6c49-4527-ac84-c4ae688a32b4.sh: line 11: emsdk: command not found
````

(aside from the fact I can't reproduce the size issue for that file locally, at least yet)

**Comment by RReverser** [comment]
> (aside from the fact I can't reproduce the size issue for that file locally, at least yet)

Found why, was just another code size test without `--output-eol`. Hope you don't mind, I fixed it in the same PR, I just want to get this landed.

File: src/lib/libstrings.js [hunk 8]
```diff
@@ -12,6 +12,18 @@ addToLibrary({
...
+  $findStringEnd: (heapOrArray, idx, maxBytesToRead, ignoreNul) => {
+    var maxIdx = idx + maxBytesToRead;
+    if (ignoreNul) return maxIdx;
+    // TextDecoder needs to know the byte length in advance, it doesn't stop on
+    // null terminator by itself.
+    // As a tiny code save trick, compare idx against maxIdx using a negation,
+    // so that maxBytesToRead=undefined/NaN means Infinity.
+    while (heapOrArray[idx] && !(idx >= maxIdx)) ++idx;
+    return idx;
+  },
```
MaxGraey, line 24: How about this?

```js
$findStringEnd: (heapOrArray, idx, maxBytesToRead, ignoreNul) => {
  var maxIdx = idx + (maxBytesToRead || heapOrArray.length);
  if (ignoreNul) return maxIdx;

  idx = heapOrArray.indexOf(0, idx);
  return idx !== -1 && idx < maxIdx ? idx : maxIdx;
},
```
I guess `indexOf` should be much faster than while loop with two checks

RReverser: Problem is, `indexOf` doesn't take an end index, and most of our strings are short enough that 1) they wouldn't benefit much from proper `indexOf` and 2) adding `.subarray()` to limit boundaries would likely make it even slower due to GC pressure.

MaxGraey: Now I see, thanks!

Then another suggestion. How about to use some trick from Hacker’s Delight and process 4 bytes at time? Or here we dealing with really, really short strings?

```js
// I guess emscripten already have something like this:
const U32 = new Uint32Array(heapOrArray.buffer);

$findStringEnd: (heapOrArray, idx, maxBytesToRead = heapOrArray.length, ignoreNul) => {
  const maxIdx = idx + maxBytesToRead;
  if (ignoreNul) return maxIdx;

  // round down to mul of 4
  const endWord = maxIdx - ((maxIdx - i) & 3);
  const U8 = heapOrArray;

  // 4-bytes scan
  while (idx < endWord) {
    const word = U32[idx];
    if (((word - 0x01010101) & ~w & 0x80808080) !== 0) {
      for (let i = 0; i < 4; ++i) {
        if (U8[idx + i] === 0) {
          return idx + i;
        }
      }
    }
    idx += 4;
  }

  // 1-byte tail scan
  while (U8[idx] && !(idx >= maxIdx)) ++idx;
  return idx;
}

```

RReverser: Usually very short ones, yeah.

src/lib/libembind.js
brendandahl: FWIW, It's been a few years since I've tested this, but string concatenation was usually faster, but used much more memory. Somewhat counterintuitively, concatenation generates a lot more garbage since strings are immutable and each concatenation creates a new string. In practice that's usually mitigated by JS engines using string ropes or some other mechanism, but still not as memory efficient as a join.

sbc100: Hmm.. maybe we should have landed this separately.  Oh well.

RReverser: I believe that was true in very, very distant paths, before modern JITs, but not nowadays. Easy to check eg on https://jsperf.app/string-concat-vs-array-join-10000/73:

image

Obviously will depend on the engine etc, but either way we use all those functions on tiny strings and microperf like this is unlikely to matter until you hit megabytes of strings.

MaxGraey: Such tests also do not take into account that flattening for rope structures happens lazily, you can try to force it out like `Number(string)` at the end, but it doesn't always work https://www.npmjs.com/package/flatstr

PR #24489, opened 2025-06-04
Refactor gen_struct_info to emit JSON directly. NFC

RReverser:
I wanted to add something to gen_struct_info functionality, but found it difficult to comprehend the custom intermediate format of the C program's output.

I changed it to emit the desired result directly as JSON instead, which results in the same output but makes code easier to reason about and debug, as well as possibly tiny bit faster as well.

### DIFF
--- a/src/lib/libbootstrap.js
+++ b/src/lib/libbootstrap.js
@@ -33,28 +33,19 @@ addToLibrary({
     quit_(1, e);
   },
 
-  // printf/puts implementations for when musl is not pulled in - very
-  // partial, but enough for bootstrapping structInfo
-  printf__deps: ['$formatString', '$intArrayToString'],
-  printf__sig: 'ipp',
-  printf: (format, varargs) => {
-    // int printf(const char *restrict format, ...);
-    // http://pubs.opengroup.org/onlinepubs/000095399/functions/printf.html
-    // extra effort to support printf, even without a filesystem. very partial, very hackish
-    var result = formatString(format, varargs);
-    var string = intArrayToString(result);
-    if (string.endsWith('\n')) string = string.slice(0, -1); // remove a final \n, as Module.print will do that
-    out(string);
-    return result.length;
-  },
-
-  puts__sig: 'ip',
-  puts: (s) => {
-    // extra effort to support puts, even without a filesystem. very partial, very hackish
-    var result = UTF8ToString(s);
-    var string = result;
-    if (string.endsWith('\n')) string = string.slice(0, -1); // remove a final \n, as Module.print will do that
-    out(string);
-    return result.length;
+  fd_write__sig: 'iippp',
+  fd_write: (fd, iov, iovcnt, pnum) => {
+    // implementation almost copied from libwasi.js one for SYSCALLS_REQUIRE_FILESYSTEM=0
+    // (the only difference is that we can't use C_STRUCTS here)
+    var num = 0;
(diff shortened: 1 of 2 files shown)


tools/gen_struct_info.py
```diff
@@ -127,80 +128,64 @@ def show(msg):
 #     },
 #     'outer': 0.999
 #   }
```
sbc100: Looks like this comment needs updating.

tools/gen_struct_info.py
```diff
@@ -127,80 +128,64 @@ def show(msg):
...
+        # scope.set('__offset__', '%zu', 'offsetof(' + prefix + path[0] + ', ' + '.'.join(path[1:]) + ')')
 
-  c_ascent(code)
+      for field in struct:
+        if isinstance(field, dict):
+          # We have to recurse to inspect the nested dict.
+          fname = list(field.keys())[0]
+          self.gen_inspect_code(path + [fname], field[fname])
+        else:
+          scope.set(field, '%zu', 'offsetof(' + prefix + path[0] + ', ' + '.'.join(path[1:] + [field]) + ')')
```
sbc100: Maybe use string interpolation in this function?

sbc100 (approved):
Seems like a reasonable refactoring.

tools/gen_struct_info.py
RReverser: Ah yeah, thanks.

src/lib/libbootstrap.js
```diff
@@ -33,28 +33,19 @@ addToLibrary({
...
+    var num = 0;
+    for (var i = 0; i < iovcnt; i++) {
+      var ptr = {{{ makeGetValue('iov', 0, '*') }}};
+      var len = {{{ makeGetValue('iov', POINTER_SIZE, '*') }}};
+      iov += {{{ POINTER_SIZE }}} * 2;
+      process.stdout.write(HEAPU8.subarray(ptr, ptr + len));
+      num += len;
+    }
+    {{{ makeSetValue('pnum', 0, 'num', '*') }}};
+    return 0;
```
sbc100, line 49: I can't remember exactly why this is the way it is.. but I'm surprised there is not sime kind of chichen and egg situation going on here.  

I guess maybe in the past it was not possible to build libc and link the boostrap program against it without have the strict info json file already in place.

I guess it fine to do it this way...

tools/gen_struct_info.py
```diff
@@ -254,7 +219,6 @@ def generate_cmd(js_file_path, src_file_path, cflags):
                                '-O0',
                                '-Werror',
                                '-Wno-format',
-                               '-nolibc',
```
sbc100, line 257: This seems like quite a big change, but I guess if works its fine.

RReverser: Oh, sorry, I thought it was a fairly small fixup to get unbroken JSON.

src/lib/libbootstrap.js
RReverser: Maybe just because of C_STRUCTS usage in the "proper" implementation? (as the comment mentions) Or maybe it was pre-wasi when bindings were different / more complicated than fd_write?

Either way, it seemed like a no-brainer and works fine, plus allows to easily use more APIs like `putc` etc without implementing each separately.

sbc100: Yup that sounds right.  Thanks for doing this!

Issue #24488, opened 2025-06-04
SINGLE_FILE doesn't respect output-eol


RReverser:
`test_codesize_hello_single_file` has been consistently producing different size on my Windows machine vs that committed in the repo and expected on CI, despite cache clean, tot updates and so on.

I finally cloned Emscripten in WSL and checked the outputs, and turns out it really does produce different output when it comes to line endings, despite the codesize test passing `--output-eol=linux`.

I now checked manually on a simpler example, and it does look like `-s SINGLE_FILE` results in `--output-eol` being ignored.

RReverser:
@sbc100 Tentatively assigning you since in one of the previous discussions you asked to raise an issue if I ever get to the bottom and can confirm the output is different across OS.

sbc100:
Thanks!

I wonder why this doesn't fail in our CI..

sbc100:
Ah.. its not marked as @crossplatform, and we deliberately skip code size tests on the emscripten-releases (integration) waterfall.

sbc100:
So according the CI results from #24492 that results are the same, at least when running on our CI windows machines.

Can you check with the latest emsdk changes that update windows python to 3.13.3?

RReverser:
Sure, I'll give it a try.

> I think you should be able to reproduce this in the opposite direction too - set `--output-eol=windows` and compile any hello world sample with `-s SINGLE_FILE` as well. If I'm right, it won't have the expected Windows line endings.

Can you try this from my other comment meanwhile please? As in, compile any file with `--output-eol=windows -s SINGLE_FILE` and check whether generated JS has `\r\n`.

If the issue is really `SINGLE_FILE` not respecting `output-eol` somehow, then you should be able to reproduce this in the opposite direction from mine too (since for you it will still generate default Linux line endings instead of the requested WIndows ones).

sbc100:
> Sure, I'll give it a try.
> 
> > I think you should be able to reproduce this in the opposite direction too - set `--output-eol=windows` and compile any hello world sample with `-s SINGLE_FILE` as well. If I'm right, it won't have the expected Windows line endings.
> 
> Can you try this from my other comment meanwhile please? As in, compile any file with `--output-eol=windows -s SINGLE_FILE` and check whether generated JS has `\r\n`.
> 
[...]

PR #24492, opened 2025-06-04
Mark test_codesize_hello as @crossplatform

sbc100:
See #24488

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -9363,6 +9363,7 @@ int main() {
     # lead to different inlining decisions which add or remove a function
     self.run_codesize_test('hello_libcxx.cpp', args, check_funcs=False)
 
+  @crossplatform
   @parameterized({
     'O0': ([],),
     'O1': (['-O1'],),


.circleci/config.yml
```diff
@@ -734,7 +734,6 @@ jobs:
             wasm64l.test_bigswitch
             wasm64l.test_module_wasm_memory
             wasm64l.test_longjmp2_emscripten
-            other.test_memory64_proxies
```
RReverser: Is this relevant?

RReverser:
Looks like still passed on windows for some reason?

.circleci/config.yml
sbc100: Oops that was from #24480

sbc100:
> Looks like still passed on windows for some reason?

Yup I see test_codesize_hello_single_file passing just fine on windows.. very strange.

What version of python are you using?   Can run the test on mac and on windows and compare the results byte for byte?

sbc100:
Can you lgtm this change in any case?

RReverser:
> Can run the test on mac and on windows and compare the results byte for byte?

I did it on Linux and Windows and compared byte by byte, that's how I found the difference. I'm just using Python from emsdk - 3.9.2.

I think you should be able to reproduce this in the opposite direction too - set `--output-eol=windows` and compile any hello world sample with `-s SINGLE_FILE` as well. If I'm right, it won't have the expected Windows line endings.

RReverser (approved):

sbc100:
> > Can run the test on mac and on windows and compare the results byte for byte?
> 
> I did it on Linux and Windows and compared byte by byte, that's how I found the difference. I'm just using Python from emsdk - 3.9.2.
> 
> I think you should be able to reproduce this in the opposite direction too - set `--output-eol=windows` and compile any hello world sample with `-s SINGLE_FILE` as well. If I'm right, it won't have the expected Windows line endings.

But all the codesize tests run with `--output-eol=linux` for this reason so that results should be identical.    And that fact that it passes CI makes me think this is working, at least on our CI machine.

If the test was not running with `--output-eol=linux` then the expectations would not match for any of these tests, regardless of the `-sSINGLE_FILE` setting.

sbc100:
Lets continue the discussion back on #24488

PR #24496, opened 2025-06-05
[acorn-opt] Use parseArgs. NFC

RReverser:


tools/acorn-optimizer.mjs
```diff
@@ -1750,16 +1751,27 @@ function reattachComments(ast, commentsMap) {
...
-}
+const {
+  values: {
+    'closure-friendly': closureFriendly,
+    'export-es6': exportES6,
+    verbose,
+    'no-print': noPrint,
+    'minify-whitespace': minifyWhitespace,
+    outfile,
+  },
```
sbc100, line 1762: Is this complex destructuring expression how folks normally use this or just a way to make this change with minimal diff?

If its the latter perhaps we could followup with another change?

RReverser: > Is this complex destructuring expression how folks normally use this

Idk. I personally like to do it this way to have clear mirroring between LHS / RHS parts, but I don't know if it's a shared sentiment.

sbc100: It looks like the docs do it like this:

```
const { parseArgs } = require('node:util');
const args = ['-f', '--bar', 'b'];
const options = {
  foo: {
    type: 'boolean',
    short: 'f',
  },
  bar: {
    type: 'string',
  },
};
const {
  values,
  positionals,
} = parseArgs({ args, options });
console.log(values, positionals);
```

Where all your options env up on the `values` object rather than in globals.  I think that could make sense as a followup perhaps.   I have complex destructuring a little non-obvious to read.. but maybe I just need to catch up.

PR #24497, opened 2025-06-05
[acorn-opt] Simplify skipping inner scopes. NFC

RReverser:
Instead of traversing the AST and changing scope nodes to empty ones before the main traversal and then traversing yet again to restore them back, just allow the `pre` callback to say when a node should be skipped.

### DIFF
--- a/tools/acorn-optimizer.mjs
+++ b/tools/acorn-optimizer.mjs
@@ -169,6 +169,7 @@ function hasSideEffects(node) {
       }
     },
     (node) =>
+      // Ignore inner scopes.
       !['FunctionDeclaration', 'FunctionExpression', 'ArrowFunctionExpression'].includes(node.type),
   );
   return has;
--- a/tools/acorn-optimizer.mjs
+++ b/tools/acorn-optimizer.mjs
@@ -60,13 +60,13 @@ function simpleWalk(node, cs) {
 }
 
 // Full post-order walk, calling a single function for all types. If |pre| is
-// provided, it is called in pre-order (before children).
+// provided, it is called in pre-order (before children). If |pre| returns
+// `false`, the node and its children will be skipped.
 function fullWalk(node, c, pre) {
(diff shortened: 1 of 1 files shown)


tools/acorn-optimizer.mjs
```diff
@@ -99,111 +99,77 @@ function dump(node) {
...
-      }
-      default: {
-        has = true;
       }
-    }
-  });
-  restoreInnerScopes(node, map);
+    },
+    (node) =>
+      !['FunctionDeclaration', 'FunctionExpression', 'ArrowFunctionExpression'].includes(node.type),
```
kripken, line 173:
```suggestion
      // Ignore inner scopes.
      !['FunctionDeclaration', 'FunctionExpression', 'ArrowFunctionExpression'].includes(node.type),
```

tools/acorn-optimizer.mjs
```diff
@@ -59,13 +59,13 @@ function simpleWalk(node, cs) {
 }
 
 // Full post-order walk, calling a single function for all types. If |pre| is
-// provided, it is called in pre-order (before children).
+// provided, it is called in pre-order (before children). If |pre| returns
+// `false`, the node and its children will be skipped.
```
sbc100, line 64: We should probably convert these comments to JSDoc style at some point..

tools/acorn-optimizer.mjs
```diff
@@ -99,111 +99,77 @@ function dump(node) {
...
-        break;
-      }
-      default: {
-        has = true;
       }
-    }
-  });
-  restoreInnerScopes(node, map);
+    },
+    (node) =>
```
sbc100, line 171: Shame you can't do `pre = ..` here to explicitly name the param (that doesn't work in JS right?)

RReverser: Yeah it doesn't.

PR #24498, opened 2025-06-05
[acorn-opt] Use Set for reaches and names. NFC

RReverser:
Those are already as sets semantically, this just makes it a bit clearer.

### DIFF
--- a/tools/acorn-optimizer.mjs
+++ b/tools/acorn-optimizer.mjs
@@ -898,7 +898,7 @@ function emitDCEGraph(ast) {
   // Final work: print out the graph
   // sort for determinism
   const graph = Object.entries(infos)
-    .sort(([name1], [name2]) => name1 > name2 ? 1 : -1)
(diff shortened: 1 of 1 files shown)


tools/acorn-optimizer.mjs
```diff
@@ -891,26 +890,19 @@ function emitDCEGraph(ast) {
...
-    names.sort();
-    return names;
-  }
-  sortedNamesFromMap(infos).forEach((name) => {
-    const info = infos[name];
-    info.reaches = sortedNamesFromMap(info.reaches);
-    graph.push(info);
-  });
+  const graph = Object.entries(infos)
+    .sort(([name1], [name2]) => name1 > name2 ? 1 : -1)
```
sbc100: That kind of looks like a default implementation of sort, does JS not provide that?

RReverser: Note we get arrays here instead of the names directly, that's why we need a custom callback. Otherwise on a simple array of strings just `.sort()` would also work, yeah.

PR #24499, opened 2025-06-05
[acorn-opt] Add walkPattern helper. NFC

RReverser:
Instead of implementing custom recursive functions, use recursiveWalk from a shared helper. It will automatically invoke one callback for nested expressions (e.g. computed properties or defaults) and another callback for any bound names.

### DIFF
--- a/tools/acorn-optimizer.mjs
+++ b/tools/acorn-optimizer.mjs
@@ -134,7 +134,7 @@ function restoreInnerScopes(node, map) {
   });
 }
 
-// Traverse a pattern node invoking onExpr on any nested expressions and onBoundIdent on any bound identifiers.
+// Traverse a pattern node (identifier, object/array pattern, etc) invoking onExpr on any nested expressions and onBoundIdent on any bound identifiers.
 function walkPattern(node, onExpr, onBoundIdent) {
   recursiveWalk(node, {
     AssignmentPattern(node, c) {
--- a/tools/acorn-optimizer.mjs
+++ b/tools/acorn-optimizer.mjs
@@ -134,6 +134,25 @@ function restoreInnerScopes(node, map) {
   });
 }
 
+// Traverse a pattern node invoking onExpr on any nested expressions and onBoundIdent on any bound identifiers.
+function walkPattern(node, onExpr, onBoundIdent) {
+  recursiveWalk(node, {
+    AssignmentPattern(node, c) {
+      c(node.left);
+      onExpr(node.right);
+    },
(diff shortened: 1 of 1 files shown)


tools/acorn-optimizer.mjs
```diff
@@ -133,6 +133,25 @@ function restoreInnerScopes(node, map) {
   });
 }
 
+// Traverse a pattern node invoking onExpr on any nested expressions and onBoundIdent on any bound identifiers.
```
kripken: Maybe give an example of what a "pattern node" is?

tools/acorn-optimizer.mjs
```diff
@@ -133,6 +133,25 @@ function restoreInnerScopes(node, map) {
...
+function walkPattern(node, onExpr, onBoundIdent) {
+  recursiveWalk(node, {
+    AssignmentPattern(node, c) {
+      c(node.left);
+      onExpr(node.right);
+    },
+    Property(node, c) {
```
kripken, line 110: Do we not need to worry about `Property` that is not part of a pattern?

kripken (approved):

kripken:
(lgtm % questions)

tools/acorn-optimizer.mjs
RReverser: No, because this function is only called on patterns and for places that might contain expressions we invoke the `onExpr` instead of descending into them.

kripken: Makes sense, thanks.

tools/acorn-optimizer.mjs
```diff
@@ -133,6 +133,25 @@ function restoreInnerScopes(node, map) {
   });
 }
 
+// Traverse a pattern node (identifier, object/array pattern, etc) invoking onExpr on any nested expressions and onBoundIdent on any bound identifiers.
```
sbc100, line 103: maybe mention what a pattern node is?   I assume its a destructuring thing that lives on the LHS of an assignment expression?

RReverser: Heh I already updated the comment in response to @kripken's comment above:

> Maybe give an example of what a "pattern node" is?

So I listed what nodes are pattern nodes (as per ecmascript & estree terminology). Does it need more clarification?

sbc100: Sounds good, i just don't grok the terminology well enough I guess.

sbc100 (approved):

PR #24505, opened 2025-06-05
Honor output-eol, even in SINGLE_FILE mode

sbc100:
Fixes: #24488


sbc100:
@juj do you remember the rational for this from back in 4c669e4?  Was it the fear the EOL conversion would not play nice with embedded base64 binary?

test/test_other.py
```diff
@@ -9001,14 +9001,20 @@ def test_disable_inlining(self):
...
   def test_output_eol(self, params):
     for eol in ('windows', 'linux'):
       self.clear()
       print('checking eol: ', eol)
-      self.run_process([EMCC, test_file('hello_world.c'), '-o', 'a.html', '--output-eol', eol] + params)
-      for f in ['a.html', 'a.js']:
+      if '-sSINGLE_FILE' not in params:
+        params.append('-oa.out.html')
+      self.run_process([EMCC, test_file('hello_world.c'), '--output-eol', eol] + params)
+      for f in ['a.out.html', 'a.out.js']:
```
RReverser: Nit but I'd move the single file check out of the loop, like

```py
out_files = ['a.out.js']
html = 'a.out.html'

if '-sSINGLE_FILE' in params:
  self.assertNotExists(html)
else:
  out_files.append(html)

for f in out_files:
```

sbc100:
> @juj do you remember the rational for this from back in [4c669e4](https://github.com/emscripten-core/emscripten/commit/4c669e4cd950c4b9ded5786a52120b1ed4ed4532)? Was it the fear the EOL conversion would not play nice with embedded base64 binary?

Ah, I think the reason was that in `SINGLE_FILE` (html) mode the JS file no longer existed at this point.  I moved the code up so that it happens before any html embedding.

RReverser:
Thank you!

test/test_other.py
```diff
@@ -9001,14 +9001,23 @@ def test_disable_inlining(self):
...
       self.clear()
       print('checking eol: ', eol)
-      self.run_process([EMCC, test_file('hello_world.c'), '-o', 'a.html', '--output-eol', eol] + params)
-      for f in ['a.html', 'a.js']:
+      if '-sSINGLE_FILE' not in params:
+        params.append('-oa.out.html')
+      self.run_process([EMCC, test_file('hello_world.c'), '--output-eol', eol] + params)
+      out_files = ['a.out.js']
+      html_file = 'a.out.html'
+      if '-sSINGLE_FILE' in params:
```
RReverser, line 9016: I guess this `if` could also be merged with the other one above, but too late and doesn't matter anyway.

PR #24508, opened 2025-06-06
Remove libformatString.js

RReverser:
After a recent change in https://github.com/emscripten-core/emscripten/pull/24489 we no longer need `formatString` for bootstrapping, and there is only one place remaining that uses this helper.

As far as I could track, it was added in very early days of Emscripten when `printf`, `fprintf` and all other friends were implemented directly in JS, and so needed a JS-based implementation of this formatting as well. Nowadays, we use native implementation from C, so usage of `formatString` can be easily replaced with `vasprintf` (or any other relevant `sprintf` variation).

After replacing it in the implementation of `emscripten_log`, the only other function that was used from this file was `unSign` in a test code, but that was no-op (it was unsigning items in a Uint8Array which is already unsigned by definition), so this entire file can be removed now.

My only concern is, as usual, whether these functions can be considered internal - they were never documented, so I hope so?


src/lib/libcore.js
```diff
@@ -1417,11 +1417,18 @@ addToLibrary({
...
+    var bufPtrPtr = stackAlloc({{{ POINTER_SIZE }}});
+    var len = _vasprintf(bufPtrPtr, format, varargs);
+    assert(len > 0);
+    var bufPtr = {{{ makeGetValue('bufPtrPtr', 0, '*') }}};
+    try {
+      var str = UTF8ToString(bufPtr, len);
+      emscriptenLog(flags, str);
+    } finally {
+      _free(bufPtr);
+    }
```
sbc100: It could be argued that converting `emscripten_log` to have more native dependencies a bad idea... but we already have  emscripten_outf / emscripten_errf / emscripten_dbgf that all depend on native formatting so maybe its fine.

A better fix here might be have `emscripten_log` implemented in native code and then only out to JS once the string is formatting.  However, IIRC emscripten_log itself is not much used at all, and the lower level emscirpten_out emscripten_err, emscripten_dbg, emscripten_console_log, and emscripten_console_err should be preferred.

test/test_core.py
```diff
@@ -6850,7 +6850,7 @@ def test_poppler(self):
     };
     Module.postRun = () => {
       var FileData = Array.from(MEMFS.getFileDataAsTypedArray(FS.root.contents['filename-1.ppm']));
-      out("Data: " + JSON.stringify(FileData.map(function(x) { return unSign(x, 8) })));
+      out("Data: " + JSON.stringify(FileData));
```
sbc100, line 6853: Can you add `Data: ` to the second arg to `do_run` below just to confirm this is the string being matched.

sbc100 (approved):

src/lib/libcore.js
sbc100: Actually, yes I think it would make sense to move this code to `system/lib/libc/emscripten_console.c` and then have the JS version take the fully formatted string.

src/lib/libcore.js
```diff
@@ -1417,11 +1417,18 @@ addToLibrary({
     }
   },
 
-  emscripten_log__deps: ['$formatString', '$emscriptenLog'],
+  emscripten_log__deps: ['$stackAlloc', 'vasprintf', '$emscriptenLog', 'free'],
   emscripten_log: (flags, format, varargs) => {
```
sbc100: I'm not even sure how this works as written because `emscripten_log` doesn't take `va_list` but `...` so I would expect to see a `va_start` here to injest the `...` args.  I guess `va_start` on wasm is a no-op, or this wouldn't work?

sbc100 (commented):
I'm loving all these cleanup PR @RReverser!  Please keep them coming.

src/lib/libcore.js
RReverser: > A better fix here might be have `emscripten_log` implemented in native code

Yeah I thought of / started doing that first, but then it seemed it blows the scope of PR a bit too much, so reverted and figured it's better to ship just the removal of libformatString first.

src/lib/libcore.js
RReverser: I might be misunderstanding it, but AFAIK va_start is just a macro that allows you to get a pointer to otherwise unnamed `...` varargs binding?

RReverser: Ah, no, because they are on stack. That's a good question, I'll double check that it works as expected.

Or maybe moving to native in the same PR is not a bad idea after all.

system/lib/libc/emscripten_console.c
```diff
@@ -77,3 +78,22 @@ void emscripten_dbg_backtracef(const char* fmt, ...) {
   va_end(ap);
 }
 #endif
+
+void _emscripten_log_formatted(int flags, const char* str);
```
sbc100: You might need to add this to `emscripten_internal.h`

RReverser: Is that for gen_sig_info or?

sbc100: Yup, it for gen_sig_info (it allows us to auto-generate the signature info for all JS function)

PR #24510, opened 2025-06-06
Allow using Closure Compiler with debug flags

RReverser:
Closure Compiler has a `--debug` mode that generates debug names that can be easily matched to original ones, while still performing all the usual advanced optimisations - constant folding, DCE, and so on.

Even though it's not strictly preserving original function names 1:1, such debug output is much more useful for debugging Closure-optimised code than disabling Closure altogether when `-g` is requested.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -12805,27 +12805,27 @@ int main(void) {
 
   def test_warning_flags(self):
     self.run_process([EMCC, '-c', '-o', 'hello.o', test_file('hello_world.c')])
-    cmd = [EMCC, 'hello.o', '-o', 'a.js', '-g', '--closure=1']
+    cmd = [EMCC, 'hello.o', '-o', 'a.js', '-g', '--llvm-opts=""']
 
     # warning that is enabled by default
     stderr = self.run_process(cmd, stderr=PIPE).stderr
-    self.assertContained('emcc: warning: disabling closure because debug info was requested [-Wemcc]', stderr)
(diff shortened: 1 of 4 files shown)


test/test_other.py
```diff
@@ -12821,11 +12821,11 @@ def test_warning_flags(self):
 
     # with -Werror should fail
     stderr = self.expect_fail(cmd + ['-Werror'])
-    self.assertContained('error: disabling closure because debug info was requested [-Wemcc] [-Werror]', stderr)
+    self.assertContained('error: use of legacy setting: TOTAL_MEMORY (setting renamed to INITIAL_MEMORY) [-Wlegacy-settings] [-Werror]', stderr)
 
     # with -Werror + -Wno-error=<type> should only warn
     stderr = self.run_process(cmd + ['-Werror', '-Wno-error=emcc'], stderr=PIPE).stderr
-    self.assertContained('emcc: warning: disabling closure because debug info was requested [-Wemcc]', stderr)
+    self.assertContained('error: use of legacy setting: TOTAL_MEMORY (setting renamed to INITIAL_MEMORY) [-Wlegacy-settings]', stderr)
```
sbc100: This check is supposed to be for a warning.   You need to update the line above too I think.

sbc100 (approved):

test/test_other.py
RReverser: Yeah, stupid copy-paste on my side. At least CI caught it too.

PR #24511, opened 2025-06-06
fix compilation of `test/gl_in_mainthread_after_pthread.c`

esqu1:
The test was missing an Emscripten header so wasn't compiling. Should be working now!

Compiled with:
```
emcc gl_in_mainthread_after_pthread.c -o hello.html -pthread -sOFFSCREEN_FRAMEBUFFER 
```

### DIFF
--- a/test/gl_in_mainthread_after_pthread.c
+++ b/test/gl_in_mainthread_after_pthread.c
@@ -11,6 +11,7 @@
 #include <assert.h>
 #include <emscripten/emscripten.h>
 #include <emscripten/html5.h>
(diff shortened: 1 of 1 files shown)


sbc100:
Hmm, this test is currently disabled: https://github.com/emscripten-core/emscripten/blob/4218f8946b7a760a5fae02baafc628f6cf31504c/test/test_browser.py#L4311-L4315.

I wonder if we can get it working?  If not its probably best just to delete it.  How did you come across this test BTW?

PR #24518, opened 2025-06-07
[contrib.glfw3] Use new external port feature

ypujante:
Now that 4.0.10 has been released and contains the new feature to make a port external, I went ahead and applied it to `contrib.glfw3`. In the end I really like this model because it keeps all the implementation details of the port outside Emscripten. Incidentally it makes reading/parsing the port file faster and also reduces the size that gets shipped with Emscripten, especially if you never need the port...


ypujante:
The only test failures after merging with main is:

```text
FAIL [7.210s]: test_webgl_with_closure (test_browser.browser64_4gb)
```

which is unrelated to my changes. The tests were all passing prior to the merge.

PR #24520, opened 2025-06-07
[embind] Remove ancient jshint comments. NFC

RReverser:
These comments were added together with Embind long, long time ago, but we don't use JSHint and our special library syntax wouldn't be compatible with it anyway, so the only thing they do is confuse devs who search for function usages.


sbc100 (approved):

sbc100:
I don't even know what jshint is!

RReverser:
JSHint what we had before ESLint, a tool by the same Douglas Crockford who created JSON. Long, long time ago now.

PR #24525, opened 2025-06-09
Automatically enable worker ENVIRONMENT when needed

### LINKED ISSUE #14162: Auto-include required code when using `-pthread -s ENVIRONMENT=web`
Tracking issue for @kripken's idea in #14151:

> Allow just setting ENVIRONMENT=node in combination with pthreads, as
that is unambiguous, and should not be broken.

> Perhaps we should do the same with pthreads outside of node, however
because of the node support code it is trivial to do it in node. I suspect it
would be more work otherwise.

Instead of showing an error telling used to use `-s ENVIRONMENT=web,worker` instead of `-s ENVIRONMENT=web`, we can just do the right thing and include the required code when `-pthread` is provided anyway.

RReverser:
- Similar to #14151 but for more combos.
 - Fixes #14162.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -21,6 +21,9 @@ See docs/process.md for more on how version tagging works.
 4.0.11 (in development)
 -----------------------
 
+----------------------
+- The `ENVIRONMENT` setting will now be automatically updated to include worker
+  if multi-threading is enabled. (#24525)
 
 4.0.10 - 06/07/25
 -----------------
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,8 +20,6 @@ See docs/process.md for more on how version tagging works.
 
 4.0.11 (in development)
 -----------------------
-
-----------------------
 - The `ENVIRONMENT` setting will now be automatically updated to include
   `worker` if multi-threading is enabled. (#24525)
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -22,8 +22,8 @@ See docs/process.md for more on how version tagging works.
 -----------------------
 
 ----------------------
-- The `ENVIRONMENT` setting will now be automatically updated to include worker
-  if multi-threading is enabled. (#24525)
+- The `ENVIRONMENT` setting will now be automatically updated to include
+  `worker` if multi-threading is enabled. (#24525)
 
 4.0.10 - 06/07/25
 -----------------
--- a/site/source/docs/tools_reference/settings_reference.rst
+++ b/site/source/docs/tools_reference/settings_reference.rst
@@ -998,7 +998,7 @@ are:
 - 'shell'   - a JS shell like d8, js, or jsc.
 
 This setting can be a comma-separated list of these environments, e.g.,
-"web,worker". If this is the empty string, then all environments are
+"web,node". If this is the empty string, then all environments are
 supported.
 
 Note that the set of environments recognized here is not identical to the
--- a/site/source/docs/tools_reference/settings_reference.rst
+++ b/site/source/docs/tools_reference/settings_reference.rst
@@ -998,7 +998,7 @@ are:
 - 'shell'   - a JS shell like d8, js, or jsc.
 
 This setting can be a comma-separated list of these environments, e.g.,
-"web,node". If this is the empty string, then all environments are
+"web,worker". If this is the empty string, then all environments are
 supported.
 
 Note that the set of environments recognized here is not identical to the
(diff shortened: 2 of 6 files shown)


sbc100:
Is seems a little of to me to automatically add environment support when the user explicitly sets the supported environment. 

Its somewhat contradictory for example if the user only wants the `shell` environment to then inject `worker` too.   

Maybe its good for convenience but perhaps it at least warrents a warning.

e.g. "You explictly specified `-sENVIRONMENT` but you didn't include `worker`.   Automatically adding `worker` to this list due to use of `-pthread` and/or `-sWASM_WORKERS`".?

RReverser:
Idk, in terms of DX I'm personally a big fan of automatically doing the right thing vs erroring out (or even warning).

If user already specified `-pthread -s ENVIRONMENT=web`, what do we gain from them having to also spell out `...,worker`? They already opted in to `-pthread`, so it seems pretty obvious that by "Web" they mean "Web and whatever implementation details you need for the pthread API".

We don't do this for other feature-specific implementation details, eg we don't have granularity for `-s ENVIRONMENT=audioworklet` (which is different from workers, but we simply detect it automatically), so why is `worker` special?

Plus, there is already a precedent with Node.js already automatically handling workers linked above.

After this PR the only reason to ever explicitly set `-s ENVIRONMENT=worker` is if the user wants to run Emscripten's main thread inside a Web Worker, but I don't think we should require it or warn on it when it's only necessary as an implementation detail for Emscripten's own workers.

RReverser:
> After this PR the only reason to ever explicitly set `-s ENVIRONMENT=worker` is if the user wants to run Emscripten's main thread inside a Web Worker, but I don't think we should require it or warn on it when it's only necessary as an implementation detail for Emscripten's own workers.

Although, come to think of it, we could auto-detect that case in exactly the same way Audio Worklets are, and then we could deprecate separate `-s ENVIRONMENT=worker` altogether.

sbc100:
> -s ENVIRONMENT=worker

Yes, I'd be up for having `-sENVIRONMENT=worker` *just* mean that you want to run the main runtime in a worker.

What do you mean by `auto-detect that case in exactly the same way Audio Worklets are`.   I assume there is some extra cost/complexity to supporting running the main runtime in worker?

In any case this change LGTM, but can you also remove `worker` from the default `-sENVIRONMENT` in that case?   (and from any multi-pthreading test cases that specify `worker` explicitly).  i.e. lets from all the explicit usage if we are making it implicit.

RReverser:
> What do you mean by `auto-detect that case in exactly the same way Audio Worklets are`. I assume there is some extra cost/complexity to supporting running the main runtime in worker?

Ah, that's a good question, might need to check how much of that complexity is for main runtime in worker vs for supporting shared memory. Either way, that would be outside of this PR.

> In any case this change LGTM, but can you also remove `worker` from the default `-sENVIRONMENT` in that case?

Do you mean from https://github.com/emscripten-core/emscripten/blob/8527d8dc8459f4350c62248a6eeaf1cffcd7dc9e/src/settings.js#L667? But then it would be a breaking change for users who do run their Emscripten-generated code in a Web Worker?

> (and from any multi-pthreading test cases that specify `worker` explicitly)

This one makes sense, yeah.

RReverser:
> Ah, that's a good question, might need to check how much of that complexity is for main runtime in worker vs for supporting shared memory. Either way, that would be outside of this PR.

Looks like there are only 3 (?) usages of `ENVIRONMENT_IS_WORKER` that aren't guarded by features like pthreads, so while it's not zero, the extra cost is pretty low. Either way, we can decide that later.

RReverser:
> Can you remove `worker` from the `var ENVIRONMENT = 'web,webview,worker,node';` line in settings.js ?

As I mentioned above, unlike rest of this PR so far, that would be a breaking change. I'd rather keep it separate if we decide to do that as well.

ChangeLog.md
```diff
@@ -20,10 +20,12 @@ See docs/process.md for more on how version tagging works.
 
 4.0.11 (in development)
 ----------------------
-
+- The `ENVIRONMENT` setting will now be automatically updated to include worker
```
sbc100: Backticks around "worker"?

sbc100:
> > Can you remove `worker` from the `var ENVIRONMENT = 'web,webview,worker,node';` line in settings.js ?
> 
> As I mentioned above, unlike rest of this PR so far, that would be a breaking change. I'd rather keep it separate if we decide to do that as well.

sgtm

ChangeLog.md
RReverser: I guess, yeah.

RReverser:
> and from any multi-pthreading test cases that specify `worker` explicitly

That was a good call, one of them actually caught an edge case I didn't consider.

Issue #24523, opened 2025-06-08
FS.staticInit throws MEMFS.doesNotExistError before it's assigned


RReverser:
I was trying to understand why I'm seeing `undefined` thrown from FS.staticInit when going at it with a debugger + "pause on caught exceptions", and traced it down to after which generated code in non-ASSERTIONS mode looks like:

```js
...
PThread.init();;

  FS.createPreloadedFile = FS_createPreloadedFile;
  FS.staticInit();;

      // This error may happen quite a bit. To avoid overhead we reuse it (and
      // suffer a lack of stack info).
      MEMFS.doesNotExistError = new FS.ErrnoError(44);
      /** @suppress {checkTypes} */
      MEMFS.doesNotExistError.stack = '<generic error, no stack>';
      ;
if (ENVIRONMENT_IS_NODE) { NODEFS.staticInit(); };
...
```

Problem is, FS.staticInit is doing lots of lookups for top-level folders and frequently ends up in

```js
  lookup(parent, name) {
          throw MEMFS.doesNotExistError;
        },
```

hence throwing it before it's even assigned. cc @sbc100

sbc100:
Maybe lets move the creation of `MEMFS.doesNotExistError` into `lookup` which is the only place it is used?

PR #24527, opened 2025-06-09
Fix memfs init racing with FS.staticInit. NFC

sbc100:
Fixes: #24523


src/lib/libmemfs.js
```diff
@@ -183,6 +174,13 @@ addToLibrary({
 #if ASSERTIONS
         throw new FS.ErrnoError({{{ cDefs.ENOENT }}});
 #else
+        // This error may happen quite a bit. To avoid overhead we reuse it (and
+        // suffer a lack of stack info).
+        if (!MEMFS.doesNotExistError) {
+          MEMFS.doesNotExistError = new FS.ErrnoError({{{ cDefs.ENOENT }}});
```
RReverser, line 180: Does it have to be exposed on MEMFS (is this public API)? If not, maybe just move it to its own variable like

```js
$MEMFS_doesNotExistError: 'new FS.ErrnoError({{{ cDefs.ENOENT }}})',
$MEMFS_doesNotExistError__postset: '...',
```

and depend on it? Then you don't need to do this lazy check on each call and should compress tiny bit better as well.

RReverser: Hm, or would that race too?

sbc100: No, certainly is not public.

sbc100: I think anything that uses `__postset` can race with `FS.staticInit()`.

sbc100: This solution seems like the most obvious and simple one to me.   Its a few more bytes in some builds and few less in others, but its in the noise compared to the cost of the memfs filesystem.

RReverser: Okay, yeah, I'm just not a huge fan of lazily-added properties, but I can see how it's simpler here.

RReverser: Although I guess you could also simply add it alongside `lookup`, right above the method?

Like

```js
doesNotExistError: Object.assign(new FS.ErrnoError({{{ cDefs.ENOENT }}}), { stack: '...' }),
lookup() { ... }
```

sbc100: I tried this but ran into a dependency cycle "FS -> MEMFS -> MEMFS_doesNotExistError -> FS" which means that FS is not defined prior to MEMFS_doesNotExistError

PR #24533, opened 2025-06-09
Move modularization code into JS file. NFC

sbc100:
Its much easier to generate JS code using out JS template system than emitting directly from python.

Split out from #23261


test/other/codesize/test_codesize_minimal_esm.gzsize
```diff
@@ -1 +1 @@
-1277
+1316
```
kripken: Why did code size increase?

sbc100: Oops, it was just comments/whitespace in the generated code.

I'll add a another `--minify-whitespace`... maybe I'll split that out.

sbc100:
No codesize changes now..

kripken (approved):

PR #24535, opened 2025-06-09
Cleanup code for capturing `document.currentScript` in MODULARIZE mode.

sbc100:
This refactor makes it clear exactly if/when we need the extra closure.  It also elides the closure completely when one is not needed.


kleisauke (approved):
LGTM!

PR #24539, opened 2025-06-10
[do not merge] Test suite with mimalloc

RReverser:
I noticed that mimalloc hangs when built into some 3rd-party projects, so want to try and see if any Emscripten tests can help narrow it down by changing the default.

### DIFF
--- a/src/settings.js
+++ b/src/settings.js
@@ -129,7 +129,7 @@ var STACK_SIZE = 64*1024;
 // the extra security checks it does (such as noticing metadata corruption in
 // its internal data structures, which emmalloc does not do).
 // [link]
-var MALLOC = "dlmalloc";
+var MALLOC = "mimalloc";
 
 // If 1, then when malloc would fail we abort(). This is nonstandard behavior,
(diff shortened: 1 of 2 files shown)


kleisauke:
FWIW, you might find success using mimalloc v3, see e.g. comment https://github.com/emscripten-core/emscripten/discussions/20936#discussioncomment-13116908.

RReverser:
Perhaps, but in my case I'm not seeing the "ran out of memory" issue mentioned at that link but an actual deadlock somewhere.

kleisauke:
That doesn't sound good. I've just opened PR #24542 to update mimalloc to v2.2.4, which includes several "important bug fixes" noted in the changelog since v2.1.7.

A changeset for the v3.1.4 release (still in beta) is also available here:
https://github.com/emscripten-core/emscripten/compare/HEAD...kleisauke:mimalloc-update-3.1.4

kleisauke:
PR #26696 updates mimalloc to v3.3.0 (latest recommended release).

PR #24542, opened 2025-06-10
Update mimalloc to 2.2.4

kleisauke:


kleisauke:
This got me thinking, perhaps we should migrate this to a submodule? I'm not sure it's feasible for the `system/` directory though, since it would complicate cloning (you'd need to use `--recurse-submodules`). I'm also not sure if emsdk already handles that.

sbc100:
> This got me thinking, perhaps we should migrate this to a submodule? I'm not sure it's feasible for the `system/` directory though, since it would complicate cloning (you'd need to use `--recurse-submodules`). I'm also not sure if emsdk already handles that.

Emscripten already uses a couple of submodules (today they are only for developers / testing) but it should be fine to start using them more.

However, we already include copies of musl / libc++ / compiler-rt and libc++abi, so converting this tiny library to a submodule would not make much difference to amount of verbatim copying we have.

sbc100 (commented):
lgtm if the tests pass.

Is this the latest official release?

kleisauke:
> Is this the latest official release?

Yes, this is the latest stable (non-beta) release. See e.g. https://github.com/microsoft/mimalloc/tags ([v3.1.4 is still in beta](https://github.com/microsoft/mimalloc/blob/v3.1.4/readme.md?plain=1#L15)).

system/lib/mimalloc/src/prim/emscripten/prim.c
```diff
@@ -196,7 +201,7 @@ bool _mi_prim_random_buf(void* buf, size_t buf_len) {
 // Thread init/done
 //----------------------------------------------------------------
 
-#ifdef __EMSCRIPTEN_SHARED_MEMORY__
+#if defined(MI_USE_PTHREADS)
```
kleisauke, line 204: Note: this was changed in https://github.com/microsoft/mimalloc/commit/0b3cd5124999efc673afb26bab3f5a1c8eff4c22#diff-ceeeab7fe916469cce6ab04ba551f6a2411d093f5cb3374cf95f0a1f9bd674aaR203 and https://github.com/microsoft/mimalloc/commit/8f874555d5d42c4e1006bfc78f6cadfb167b1e30#diff-dc7c5d6afe67633e8647bc21fb12adebb183b49dd43e6bd47e1b7a822587fbe0R17.

kleisauke:
Closing in favor of #26696.

PR #24545, opened 2025-06-10
Treat ENVIRONMENT setting as a list. NFC

sbc100:
This should not be a user-visible change I hope since the existing bespoke format of comma-separated-list will parse just fine as a real list too.


tools/link.py
```diff
@@ -166,18 +166,17 @@ def setup_environment_settings():
...
-  if any(x for x in environments if x not in VALID_ENVIRONMENTS):
+  if any(x for x in settings.ENVIRONMENT if x not in VALID_ENVIRONMENTS):
     exit_with_error(f'Invalid environment specified in "ENVIRONMENT": {settings.ENVIRONMENT}. Should be one of: {",".join(VALID_ENVIRONMENTS)}')
 
-  settings.ENVIRONMENT_MAY_BE_WEB = not settings.ENVIRONMENT or 'web' in environments
-  settings.ENVIRONMENT_MAY_BE_WEBVIEW = not settings.ENVIRONMENT or 'webview' in environments
-  settings.ENVIRONMENT_MAY_BE_NODE = not settings.ENVIRONMENT or 'node' in environments
-  settings.ENVIRONMENT_MAY_BE_SHELL = not settings.ENVIRONMENT or 'shell' in environments
-  settings.ENVIRONMENT_MAY_BE_WORKER = not settings.ENVIRONMENT or 'worker' in environments
+  settings.ENVIRONMENT_MAY_BE_WEB = not settings.ENVIRONMENT or 'web' in settings.ENVIRONMENT
```
RReverser, line 175: Btw, I didn't want to touch those, but... do we actually need all those `not settings.ENVIRONMENT`? I assume it existed to handle an empty default, but our default list is already non-empty...

tools/maint/update_settings_docs.py
```diff
@@ -101,7 +101,7 @@ def write_file(f):
       # Format:
       #   var NAME = DEFAULT;
       # Split it and strip the final ';'.
-      _, setting_name, _, setting_default = line.strip(';').split()
+      _, setting_name, _, setting_default = line.strip(';').split(None, 3)
```
RReverser, line 104: What is this for?

RReverser (approved):
Nice!

tools/maint/update_settings_docs.py
sbc100: This code parses the settings.js file and emits the `.rst` docs.

Without this change it cannot parse a list on single line because it has extra whitespace chars.  Adding 3 there means we never get more than 4 results from the split, which is what we want.

tools/link.py
sbc100: Setting this to the empty list explictly means "all environments".  This is currently how to handle that it seems.

tools/link.py
```diff
@@ -166,18 +166,17 @@ def setup_environment_settings():
   # Note: we need to actually modify ENVIRONMENTS variable here before the parsing,
   # because some JS code reads it back so modifying parsed info alone is not sufficient.
   if settings.SHARED_MEMORY and settings.ENVIRONMENT:
-    settings.ENVIRONMENT += ',worker'
+    settings.ENVIRONMENT.append('worker')
 
   # Environment setting based on user input
-  environments = settings.ENVIRONMENT.split(',')
-  if any(x for x in environments if x not in VALID_ENVIRONMENTS):
+  if any(x for x in settings.ENVIRONMENT if x not in VALID_ENVIRONMENTS):
```
RReverser, line 172: I wonder if we can make `ENVIRONMENT` a `set()` instead of a `list()`. Then you could just do `if settings.ENVIRONMENT - VALID_ENVIRONMENTS:`.

tools/link.py
RReverser: > Setting this to the empty list explictly means "all environments".

Ah, as in "even more than the default"? Feels a bit obscure, but yeah, then it would be a breaking change, nvm.

sbc100: Sadly that is what the docs say today.

tools/link.py
sbc100: There is some precedent for using `Set` for settings on the JS side for settings in python today are only strings, numbers or lists.

This change seems simpler as is, we can followup up if we decide it makes a difference.

RReverser: Yeah just took a moment to read, and thought at least `if set(settings.ENVIRONMENT) - VALID_ENVIRONMENTS:` would be simple enough if you don't want to change the setting itself.

But fine either way.

PR #24550, opened 2025-06-11
[JSPI] Do not automatically wrap `async` library functions.

### LINKED ISSUE #20289: Async JS library funcs should not be marked for Asyncify transform

brendandahl:
In a number of places library functions  (e.g. `emscripten_async_wget_data`) use an `async` JS functions, but they do not actually want to use JSPI wrappers. Change it so functions must explicitly use `foo__async: true` to enable JSPI wrappers.

Fixes #20289

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -3489,28 +3489,26 @@ More info: https://emscripten.org
     # Make sure async library functions are not automatically JSPI'd.
     create_file('lib.js', r'''
       addToLibrary({
-       foo: async function(f) { await Promise.resolve(); },
+        foo: async function(f) { await Promise.resolve(); },
       });
-      ''')
+    ''')
     create_file('main.c', r'''
-        #include <emscripten.h>
-        extern void foo();
-        EMSCRIPTEN_KEEPALIVE void test() {
-          foo();
-        }
-      ''')
+      #include <emscripten.h>
+      extern void foo();
+      EMSCRIPTEN_KEEPALIVE void test() {
+        foo();
+      }
+    ''')
     create_file('post.js', r'''
-        Module.onRuntimeInitialized = () => {
-          _test()
-          console.log('done');
-        };
-      ''')
-    self.emcc_args += [
-      '-sJSPI',
-      '--js-library=lib.js',
-      '-Wno-experimental',
-      '--post-js=post.js']
-    self.do_runf('main.c', 'done')
+      Module.onRuntimeInitialized = () => {
+        _test()
+        console.log('done');
+      };
+    ''')
+    self.do_runf('main.c', 'done', emcc_args=['-sJSPI',
+                                              '--js-library=lib.js',
+                                              '-Wno-experimental',
+                                              '--post-js=post.js'])
 
   @requires_dev_dependency('typescript')
   @parameterized({
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -3484,6 +3484,34 @@ More info: https://emscripten.org
       '-Wno-experimental']
     self.do_runf('other/test_jspi_add_function.c', 'done')
 
+  @requires_jspi
+  def test_jspi_async_function(self):
+    # Make sure async library functions are not automatically JSPI'd.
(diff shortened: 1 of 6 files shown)


RReverser (approved):
Yay!

test/test_other.py
```diff
@@ -3484,6 +3484,34 @@ def test_jspi_add_function(self):
...
+        Module.onRuntimeInitialized = () => {
+          _test()
+          console.log('done');
+        };
+      ''')
+    self.emcc_args += [
+      '-sJSPI',
+      '--js-library=lib.js',
+      '-Wno-experimental',
+      '--post-js=post.js']
```
sbc100: Maybe put this on just one line?  Or even inline it into the `do_runf` call?

test/test_other.py
```diff
@@ -3484,6 +3484,34 @@ def test_jspi_add_function(self):
...
+        EMSCRIPTEN_KEEPALIVE void test() {
+          foo();
+        }
+      ''')
+    create_file('post.js', r'''
+        Module.onRuntimeInitialized = () => {
+          _test()
+          console.log('done');
+        };
+      ''')
```
sbc100: The indentaion for lib.js looks different to the other two files here.  Maybe make them consistent?

Also, what do you think about make the closing `''')` on indent match the initial indent of `create_file`?

test/test_other.py
```diff
@@ -3484,6 +3484,34 @@ def test_jspi_add_function(self):
       '-Wno-experimental']
     self.do_runf('other/test_jspi_add_function.c', 'done')
 
+  @requires_jspi
+  def test_jspi_async_function(self):
+    # Make sure async library functions are not automatically JSPI'd.
```
RReverser, line 3489: I'm actually a bit confused as to how this test verifies it. Wouldn't it silently pass w/o the change in this PR too?

brendandahl: Without the change this test will error with `SuspendError: trying to suspend without WebAssembly.promising`.

RReverser: Ah, I didn't realise JSPI checks that. Sgtm.

PR #24552, opened 2025-06-12
Allow to set custom websocket factory

ololoken:
In some cases, especially when application uses more then one connection, it will be handy to define custom websocket factory.

### DIFF
--- a/src/lib/libsockfs.js
+++ b/src/lib/libsockfs.js
@@ -172,6 +172,19 @@ addToLibrary({
             addr = result[1];
             port = parseInt(result[2], 10);
           }
+        } else if (typeof SOCKFS.websocketArgs["factory"] === "function") {
+          try {
+            ws = SOCKFS.websocketArgs["factory"]({
+              addr,
+              port,
+              type: sock.type,
+              family: sock.family,
+              protocol: sock.protocol
+            });
+            ws.binaryType = 'arraybuffer';
+          } catch (e) {
+            throw new FS.ErrnoError({{{ cDefs.EHOSTUNREACH }}});
+          }
         } else {
           // create the actual websocket object and connect
           try {
--- a/site/source/docs/porting/networking.rst
+++ b/site/source/docs/porting/networking.rst
@@ -51,6 +51,9 @@ This is the default build mode for Emscripten. Use the linker flag
 to connect to, and the linker flag ``-sWEBSOCKET_SUBPROTOCOL`` or
 ``Module['websocket']['subprotocol']`` to control the connection type
 (``'binary'`` or ``'text'``).
+Or define custom WebSocket factory using ``Module['websocket']['factory']``. Factory
+is called with options param ``{addr, port, type, family, protocol}`` and
+must return WebSocket instance.
 
 Full POSIX Sockets over WebSocket Proxy Server
 ==============================================


src/lib/libsockfs.js
```diff
@@ -172,6 +172,19 @@ addToLibrary({
             addr = result[1];
             port = parseInt(result[2], 10);
           }
+        } else if (typeof SOCKFS.websocketArgs["factory"] === "function") {
```
sbc100: Wrap this whole block in `#if expectToReceiveOnModule('websocket')`

src/lib/libsockfs.js
```diff
@@ -172,6 +172,19 @@ addToLibrary({
...
+              addr,
+              port,
+              type: sock.type,
+              family: sock.family,
+              protocol: sock.protocol
+            });
+            ws.binaryType = 'arraybuffer';
+          } catch (e) {
+            throw new FS.ErrnoError({{{ cDefs.EHOSTUNREACH }}});
+          }
```
sbc100, line 189: Would it be enough to just replace `WebSocketConstructor` below?

Then the rest of the code in that block could be shared too?

Or do you want to be able all the logic to do with `url` and `subprotocol` below?

Perhaps you could give more details about why / when these might be needed? Perhaps via and example and/or a test?

src/lib/libsockfs.js
ololoken: done

src/lib/libsockfs.js
ololoken: I think when I define ws factory, then I know url/subprotocol/etc
Main goal is to have option to configure server ws proxy with for example ws connection url searchQuery. Or wrap ws object into Proxy for additional data filtering/management.

sbc100: OK, perhaps you can add a motivating example usage in the form of a test.

Are we sure this cannot be achieved without modifying emscripten by monkey-patching the WebSocket global perhaps?

ololoken: Not sure a test will test anything related to sockets functionality. The change is aimed for post-emscripten module integration. 
Real life example: I have an app which performs ping of hosts(udp), list of hosts is fetched inside module with simple tcp connection. Using this factory I'm able to use single server ws proxy written in nodejs. Also right now I wrap ws instance's send and onmessage in proxy for additional logging, but it also can be used for any type of data manipulation.

sbc100: Well, we certainly need some kind of test for this new functionality.

How about adding a test that shows that you can inject logging this way?

@juj what do you think of this change?   Can / should we leave this up to normal monkey patching mechanisms or should we provide this hook?

src/lib/libsockfs.js
```diff
@@ -172,7 +172,24 @@ addToLibrary({
...
+        else if (typeof SOCKFS.websocketArgs["factory"] === "function") {
+          try {
+            ws = SOCKFS.websocketArgs["factory"]({
+              addr,
+              port,
+              type: sock.type,
+              family: sock.family,
+              protocol: sock.protocol
+            });
+            ws.binaryType = 'arraybuffer';
```
juj, line 186: Is there a reason that this side of code is overwriting the .binaryType? Could the factory side do it before returning?

src/lib/libsockfs.js
juj: To me this looks ok. It would indeed allow runtime resolution of websocket connections, as opposed to needing to know at compile time how to connect.

Definitely would be great to have a test in sockets for this functionality, to ensure that the hook/dispatch mechanism is kept functional.

> Are we sure this cannot be achieved without modifying emscripten by monkey-patching the WebSocket global perhaps?

Patching WebSocket global is a bit tricky because it could then leak to other Emscripten WebSocket backends, altering their functionality.

src/lib/libsockfs.js
ololoken: https://github.com/emscripten-core/emscripten/blob/bf5d63c2c0cf244a5a8c31f2261c0d01c49d8f7b/src/lib/libsockfs.js#L226
I guess it's required for TextEncoder/Decoder and logic in recv shim

src/lib/libsockfs.js
ololoken: TBH I'm not sure I can add proper test. Maybe you can advice some existing test, which I can alter/modify?

PR #24553, opened 2025-06-12
[acorn-opt] Use walkPattern for variable declarators too

RReverser:


tools/acorn-optimizer.mjs
```diff
@@ -243,17 +243,16 @@ function JSDCE(ast, aggressive) {
           let removedHere = 0;
           node.declarations = node.declarations.filter((node) => {
             assert(node.type === 'VariableDeclarator');
-            const id = node.id;
-            if (id.type === 'ObjectPattern' || id.type === 'ArrayPattern') {
-              // TODO: DCE into object patterns, that is, things like
```
sbc100, line 248: Does this fix the TODO?  Perhaps update one of the existing tests to show that this works now?

RReverser: It does, yeah, and I can see from CI there's a existing test_js_optimizer_JSDCE-objectPattern test that can be updated now.

PR #24554, opened 2025-06-12
Fix test failures on Windows when path contains spaces or current codepage is Unicode

juj:
Fix two tests to pass on Windows when the Emscripten SDK directory contains spaces, or current codepage is Unicode.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -6314,9 +6314,12 @@ PORT: 3979
     # are having the desired effect.
     # This means that files open()'d by emscripten without an explicit encoding will
     # cause this test to file, hopefully catching any places where we forget to do this.
-    create_file('expect_fail.py', 'print(len(open(r"%s").read()))' % test_file('unicode_library.js'))
-    err = self.expect_fail([PYTHON, 'expect_fail.py'], expect_traceback=True)
-    self.assertContained('UnicodeDecodeError', err)
+
+    # On Windows when Unicode support is enabled, this test code does not fail.
+    if not (WINDOWS and self.run_process(['chcp'], stdout=PIPE, shell=True).stdout.strip() == 'Active code page: 65001'):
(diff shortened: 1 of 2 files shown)


test/test_core.py
```diff
@@ -6314,9 +6314,12 @@ def test_unicode_js_library(self):
     # are having the desired effect.
     # This means that files open()'d by emscripten without an explicit encoding will
     # cause this test to file, hopefully catching any places where we forget to do this.
-    create_file('expect_fail.py', 'print(len(open(r"%s").read()))' % test_file('unicode_library.js'))
-    err = self.expect_fail([PYTHON, 'expect_fail.py'], expect_traceback=True)
-    self.assertContained('UnicodeDecodeError', err)
+
+    # On Windows when Unicode support is enabled, this test code does not fail.
+    if not (WINDOWS and self.run_process(['chcp'], stdout=PIPE, shell=True).stdout.strip() == 'Active code page: 65001'):
```
sbc100, line 6319: What is 65001?  

Do you know why this doesn't fail?  Is the issue that `PYTHONUTF8` and/or `PYTHONCOERCECLOCALE` are not being honored under windows?

juj: 65001 is Unicode codepage in Windows command prompt.

I don't know why this test fails.

But I observe that if I enable Unicode support by ticking this checkbox:

Screenshot 2025-06-12 234126

( Which is required to successfully build LLVM/Clang in a subdirectory that contains Unicode chars: https://github.com/llvm/llvm-project/issues/143968 )

Then this test starts to fail in this check.

The effect of that checkbox is to change the default codepage in all opened command prompts to 65001.

sbc100: Fair enough, I guess the PYTHONUTF8 trick we use here doesn't work on windows.

sbc100 (approved):

PR #24555, opened 2025-06-12
[esm-integration] Add pthread support

sbc100:
Under ESM integration all dependencies must be satisfied at import time, but pthreads requires that we have supply the memory via postMessage, so the memory is, by definition, not available at import time.

On order to work around this issue we create a smaller pthread stub/loader file that delays the import of the main program until the initial `postMessage` has been received.  Once the memory is received we load main program using a dynamic `import` statement.

### DIFF
--- a/src/postamble.js
+++ b/src/postamble.js
@@ -306,6 +306,9 @@ export default async function init(moduleArg = {}) {
   Object.assign(Module, moduleArg);
   processModuleArgs();
 #if WASM_ESM_INTEGRATION
+#if PTHREADS
+  registerTLSInit(__emscripten_tls_init);
+#endif
   updateMemoryViews();
 #if DYNCALLS && '$dynCalls' in addedLibraryItems
 
@@ -318,7 +321,7 @@ export default async function init(moduleArg = {}) {
   run();
 }
 
-#if PTHREADS || WASM_WORKERS
+#if (WASM_WORKERS || PTHREADS) && !WASM_ESM_INTEGRATION
 // When run as a worker thread run `init` immediately.
 if ({{{ ENVIRONMENT_IS_WORKER_THREAD() }}}) await init()
 #endif
--- a/src/pthread_esm_startup.mjs
+++ b/src/pthread_esm_startup.mjs
@@ -1,0 +1,53 @@
+/**
+ * @license
+ * Copyright 2025 The Emscripten Authors
+ * SPDX-License-Identifier: MIT
+ */
+
+// This file is used as the initial script loaded into pthread workers when
+// running in WASM_ESM_INTEGRATION mode.
+// Tyhe point of this file is to delay the loading of the main program module
+// until the wasm memory has been received via postMessage.
+
+#if RUNTIME_DEBUG
+console.log("Running pthread_esm_startup");
+#endif
+
+#if ENVIRONMENT_MAY_BE_NODE
+// Create as web-worker-like an environment as we can.
+var worker_threads = await import('worker_threads');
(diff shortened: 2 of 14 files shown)


sbc100:
I found a way to make it work!  Not pretty, but it works!

sbc100:
This change has now been simplified a lot.  I think its good to go now.  PTAL

src/postamble.js
```diff
@@ -318,9 +321,9 @@ export default async function init(moduleArg = {}) {
   run();
 }
 
-#if PTHREADS || WASM_WORKERS
+#if (WASM_WORKERS || PTHREADS) && !WASM_ESM_INTEGRATION
 // When run as a worker thread run `init` immediately.
-if ({{{ ENVIRONMENT_IS_WORKER_THREAD() }}}) await init()
+if ({{{ ENVIRONMENT_IS_WORKER_THREAD() }}}) init()
```
```suggestion
if ({{{ ENVIRONMENT_IS_WORKER_THREAD() }}}) init();
```

kripken: (pre-existing, but still)

src/pthread_esm_startup.mjs
```diff
@@ -0,0 +1,48 @@
+/**
+ * @license
+ * Copyright 2025 The Emscripten Authors
+ * SPDX-License-Identifier: MIT
+ */
+
+// This file is used as the initial script loaded into pthread workers when
+// running in WASM_ESM_INTEGRATION mode.
+// The point of this file is to delay the loading the main program module
```
kripken:
```suggestion
// The point of this file is to delay loading of the main program module
```

src/pthread_esm_startup.mjs
```diff
@@ -0,0 +1,48 @@
...
+#endif
+
+self.onmessage = async (msg) => {
+#if RUNTIME_DEBUG
+  console.log('pthread_esm_startup', msg.data.cmd);
+#endif
+  if (msg.data.cmd == 'load') {
+    // Until we initialize the runtime, queue up any further incoming messages.
+    let messageQueue = [msg];
+    self.onmessage = (e) => messageQueue.push(e);
```
kripken, line 38: I'm confused.

1. Isn't this queuing also needed *before* "load" arrives? It seems like it applies after load, for things in between load and some future event..?
2. It looks like we process this queue right after the await - to be sure I follow, is the queue only for events arriving during the await?
3. Do we not need to undo the trampling of `self.onmessage` at the end?

src/postamble.js
sbc100: Reverted this line.

src/pthread_esm_startup.mjs
sbc100: I added some more comments that should explain what is doing on there more precisely.

src/pthread_esm_startup.mjs
```diff
@@ -0,0 +1,53 @@
...
+    // is complete.
+    let messageQueue = [msg];
+    self.onmessage = (e) => messageQueue.push(e);
+
+    // Now that we have the wasmMemory we can import the main program
+    globalThis.wasmMemory = msg.data.wasmMemory;
+    const prog = await import('./{{{ TARGET_JS_NAME }}}');
+
+    // Now that the import is completely the main program will have installed
+    // its `onmessage` handler and replaced out handler.
```
kripken:
```suggestion
    // Now that the import is complete, the main program will have installed
    // its `onmessage` handler and replaced our handler.
```

kripken (approved):
Now I see, thanks. Tricky but nice how you got it to work.

PR #24560, opened 2025-06-13
Fix -gsource-map with UTF-8 chars in directory and file names in the path.

### LINKED ISSUE #24557: -gsource-map does not work on Windows if username contains non-ASCII Unicode characters.
If username has non-ASCII Unicode characters, then on Windows the `TEMP` environment variable will point to a path that has Unicode chars in it (e.g. `TEMP=C:\Users\Matti Meikäläinen\AppData\Local\Temp`)

In such a mode, using `-gsource-map` command line flag fails on Windows.

e.g.

```
C:\ä ö\emsdk\emscripten\main>set TEMP
TEMP=C:\ä ö\temp

C:\ä ö\emsdk\emscripten\main>test\runner other.test_wasm_sourcemap_relative_paths
Running test_other: (1 tests)
test_wasm_sourcemap_relative_paths (test_other.other.test_wasm_sourcemap_relative_paths) ... a.cpp a.cpp
   a.cpp
     same CWD for compiler and linker
FAIL

FAIL: test_wasm_sourcemap_relative_paths (test_other.other.test_wasm_sourcemap_relative_paths)
  File "C:\ä ö\emsdk\emscripten\main\test\test_other.py", line 10781, in test_wasm_sourcemap_relative_paths
    test('a.cpp')
    ~~~~^^^^^^^^^
  File "C:\ä ö\emsdk\emscripten\main\test\test_other.py", line 10774, in test
    self.assertIn('"%s"' % expected_source_map_path, read_file('a.out.wasm.map'))
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: '"a.cpp"' not found in '{"version":3,"sources":["../../../../../../303/244 /303/266/emsdk/emscripten/main/out/test/a.cpp","/emsdk/emscripten/system/lib/compiler-rt/stack_ops.S","/emsdk/emscripten/system/lib/compiler-rt/stack_limits.S","/emsdk/emscripten/system/lib/pthread/library_pthread_stub.c","/emsdk/emscripten/system/lib/libc/musl/src/stdio/ofl.c","/emsdk/emscripten/system/lib/libc/musl/src/stdio/__lockfile.c","/emsdk/emscripten/system/lib/libc/musl/src/stdio/fflush.c"],"sourcesContent":[],"names":[],"mappings":"6XAOA,iCACE,QAAA,kBCQF,GACA,MACA,KAKA,MAEA,EAEA,CAEA,EACA,CACA,EACA,MACA,EACA,EAIA,OACA,ECWA,OAEA,MAMA,MAGA,EACA,CACA,EACA,CACA,MAEA,EAYA,OACA,MACA,CACA,EAnDA,OACA,EAIA,OACA,GCmVwB,GAEE,GChXzB,YACA,MAAA,GAKA,YACD,ECbA,GAaC,GASD,MCjBK,UAEC,cAA2B,SAAP,aACpB,UAA2B,SAAP,MAAH,MAEb,QAAV,GAAE,KACC,oBAAA,GAAA,QAEA,IADO,OAAW,KAAN,GAAkB,QAAH,MAC3B,eAHH,KAAE,MAKA,MAyBF,IApBC,kBAAA,GAAA,QAkBA,IAfO,WAAW,KAAN,SACX,iBACQ,KAAJ,MACH,QAMK,OAAW,OAAN,KAA8B,OAAP,GAAX,0BAGL,OAAU,OACrB,GAER,cAED,EAAA"}'

Ran 1 test in 0.533s

FAILED (failures=1)

C:\ä ö\emsdk\emscripten\main>
```

It looks like the source map generation has generated an incorrect source path: 

```
"../../../../../../303/244 /303/266/emsdk/emscripten/main/out/test/a.cpp"
```
when it was expected to have resolved to just `a.cpp`.

If the `TEMP` directory points to a path without Unicode chars, then the test passes on Windows.

juj:
Fix -gsource-map with UTF-8 chars in directory and file names in the path. Fixes #24557.

### DIFF
--- a/tools/wasm-sourcemap.py
+++ b/tools/wasm-sourcemap.py
@@ -197,6 +197,25 @@ def remove_dead_entries(entries):
     block_start = cur_entry
 
 
+# Given a string that has non-ASCII UTF-8 bytes 128-255 stored as octal sequences (\200 - \377), decode
+# the sequences back to UTF-8. E.g. "C:\\\303\244 \303\266\\emsdk\\emscripten\\main" -> "C:\\ä ö\\emsdk\\emscripten\\main"
+def decode_octal_encoded_utf8(str):
+  out = bytearray(len(str))
+  i = 0
+  o = 0
+  final_length = len(str)
+  while i < len(str):
+    if str[i] == '\\' and (str[i+1] == '2' or str[i+1] == '3'):
+      out[o] = int(str[i+1:i+4], 8)
+      i += 4
+      final_length -= 3
+    else:
+      out[o] = ord(str[i])
+      i += 1
+    o += 1
+  return out[:final_length].decode('utf-8')
+
+
 def extract_comp_dir_map(text):
   map_stmt_list_to_comp_dir = {}
   chunks = re.split(r"0x[0-9a-f]*: DW_TAG_compile_unit", text)
@@ -205,7 +224,7 @@ def extract_comp_dir_map(text):
     if stmt_list_match is not None:
       stmt_list = stmt_list_match.group(1)
       comp_dir_match = re.search(r"DW_AT_comp_dir\s+\(\"([^\"]+)\"\)", chunk)
-      comp_dir = comp_dir_match.group(1) if comp_dir_match is not None else ''
+      comp_dir = decode_octal_encoded_utf8(comp_dir_match.group(1)) if comp_dir_match is not None else ''
       map_stmt_list_to_comp_dir[stmt_list] = comp_dir
   return map_stmt_list_to_comp_dir
 
@@ -251,12 +270,12 @@ def read_dwarf_entries(wasm, options):
 
     include_directories = {'0': comp_dir}
     for dir in re.finditer(r"include_directories\[\s*(\d+)\] = \"([^\"]*)", line_chunk):
-      include_directories[dir.group(1)] = os.path.join(comp_dir, dir.group(2))
+      include_directories[dir.group(1)] = os.path.join(comp_dir, decode_octal_encoded_utf8(dir.group(2)))
 
     files = {}
     for file in re.finditer(r"file_names\[\s*(\d+)\]:\s+name: \"([^\"]*)\"\s+dir_index: (\d+)", line_chunk):
       dir = include_directories[file.group(3)]
-      file_path = os.path.join(dir, file.group(2))
+      file_path = os.path.join(dir, decode_octal_encoded_utf8(file.group(2)))
       files[file.group(1)] = file_path
 
     for line in re.finditer(r"\n0x([0-9a-f]+)\s+(\d+)\s+(\d+)\s+(\d+)(.*?end_sequence)?", line_chunk):
@@ -352,8 +371,8 @@ def main():
 
   logger.debug('Saving to %s' % options.output)
   map = build_sourcemap(entries, code_section_offset, options)
-  with open(options.output, 'w') as outfile:
-    json.dump(map, outfile, separators=(',', ':'))
+  with open(options.output, 'w', encoding='utf-8') as outfile:
+    json.dump(map, outfile, separators=(',', ':'), ensure_ascii=False)
 
   if options.strip:
     wasm = strip_debug_sections(wasm)
--- a/tools/wasm-sourcemap.py
+++ b/tools/wasm-sourcemap.py
@@ -205,8 +205,8 @@ def decode_octal_encoded_utf8(str):
   o = 0
   final_length = len(str)
   while i < len(str):
-    if str[i] == '\\' and (str[i+1] == '2' or str[i+1] == '3'):
-      out[o] = int(str[i+1:i+4], 8)
+    if str[i] == '\\' and (str[i + 1] == '2' or str[i + 1] == '3'):
+      out[o] = int(str[i + 1:i + 4], 8)
       i += 4
       final_length -= 3
     else:
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -10757,7 +10757,7 @@ int main() {
     ensure_dir('build')
 
     def test(infile, source_map_added_dir=''):
-      expected_source_map_path = 'a.cpp'
+      expected_source_map_path = 'A ä☃ö Z.cpp'
       if source_map_added_dir:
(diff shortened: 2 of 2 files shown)


tools/wasm-sourcemap.py
```diff
@@ -352,8 +371,8 @@ def main():
 
   logger.debug('Saving to %s' % options.output)
   map = build_sourcemap(entries, code_section_offset, options)
-  with open(options.output, 'w') as outfile:
-    json.dump(map, outfile, separators=(',', ':'))
+  with open(options.output, 'w', encoding='utf-8') as outfile:
+    json.dump(map, outfile, separators=(',', ':'), ensure_ascii=False)
```
sbc100, line 375: So `ensure_ascii=False` means that non-ascii chars are not escaped as they normally would be.    Is that what we want for source maps?

sbc100 (commented):
Have you verified that the resulting source maps actually work correctly in browsers?  (as opposed to just having the test pass).

juj:
Yes, it does look like UTF-8 in source maps work:

Screenshot 2025-06-13 211102

tools/wasm-sourcemap.py
juj: I think that is probably the simplest. JSON readily supports UTF-8 characters in content.

dschuff:
This one seems to be failing on our windows emscripten-releases bot:  https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket/8712146499624631425/+/u/Emscripten_testsuite__other_/stdout

```
ERROR: test_wasm_sourcemap_relative_paths (test_other.other)
  File "C:\Users\chrome-bot\AppData\Local\vpython-root.0\store\cpython+dunqimq8p43kdruh73ujf8fpl8\contents\bin\Lib\unittest\case.py", line 60, in testPartExecutor
    yield
  File "C:\Users\chrome-bot\AppData\Local\vpython-root.0\store\cpython+dunqimq8p43kdruh73ujf8fpl8\contents\bin\Lib\unittest\case.py", line 676, in run
  File "C:\Users\chrome-bot\AppData\Local\vpython-root.0\store\cpython+dunqimq8p43kdruh73ujf8fpl8\contents\bin\Lib\unittest\case.py", line 633, in _callTestMethod
    method()
  File "C:\b\s\w\ir\x\w\install\emscripten\test\test_other.py", line 10781, in test_wasm_sourcemap_relative_paths
    test('A �\u2603� Z.cpp')
  File "C:\b\s\w\ir\x\w\install\emscripten\test\test_other.py", line 10763, in test
    print(infile, expected_source_map_path)
  File "C:\Users\chrome-bot\AppData\Local\vpython-root.0\store\python_venv-n0ssd7qim3vjmkh5j9ljrgnplo\contents\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u2603' in position 3: character maps to <undefined>
```

Looks like it's trying to print, expecting a cp1252 encoding but it's actually UTF-8?

juj:
That's interesting. I tried to reproduce with the following setups:

- Windows 11, keeping Unicode support enabled in system Regions menu, resulting in `chcp` printing codepage 65001.
- Windows 11, disabling Unicode support in system Regions menu, resulting in `chcp` printing codepage 437.
- Windows 11, disabling Unicode support in system Regions menu, explicitly running `chcp 1252` before test.

and also on another Windows 10 PC, repeating the same three cases. I thought for sure setting codepage to 1252 and/or disabling Unicode support would reproduce the problem, but that doesn't seem to help.

Reading the line

```
EMSDK_PYTHON = C:\Users\chrome-bot\AppData\Local\vpython-root.0\store\python_venv-n0ssd7qim3vjmkh5j9ljrgnplo\contents\Scripts\python3.exe
```

in the log suggests that Python version might be different on the bot vs what I am running with from Emsdk.

I updated Emsdk recently to have Python 3.13.3.

So I wonder if this might be a Python version thing that older version had problems printing Unicode characters? If an older Python version is used there, I wonder if it would be feasible to update? Otherwise I think skipping that one test on the bot for now could be a simple approach as well.

dschuff:
Thanks for digging into this. It wouldn't surprise me if there's an issue with the particular version and configuration of Python and/or the underlying OS on these builders, as they are fairly customized for Chromium's CI, and we are just along for the ride. That also means that it's probably not worth trying to get them to update it; but since it's working on GitHub, and if you're pretty confident that everything is WAI after your investigation here, then I'm comfortable just disabling this test on the Chromium CI.

juj:
@dschuff  Would you be able to run `python --version` on the failing Chromium CI to verify which version it has? I can try to locally install the same version to see if that reproduces the problem.

dschuff:
According to the top line of the log in https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket/8712146499624631425/+/u/Emscripten_testsuite__other_/stdout
the version is
`Python version 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64 bit (AMD64)]`

dschuff:
which... actually I thought they were planning on updating all the builders to 3.11, so maybe ours got left behind somehow. I'll look into that. But maybe it's worth trying to reproduce on whatever the oldest version of Python we want to support in emscripten is. The reason Chrome infra is updating is because 3.8 is EOL, so maybe we want to update our min supported version as well.

dschuff:
I tried a CL (https://chromium-review.googlesource.com/c/emscripten-releases/+/6649504) updating our python runtime on the Chromium CI to 3.11, but the error (https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket/8711863587810698449/+/u/Emscripten_testsuite__other_/stdout) looks the same.

dschuff:
I made a CL to skip this test on Chromium CI for now (https://chromium-review.googlesource.com/c/emscripten-releases/+/6655657), to unblock the rollers, but it would be good to understand this if we can.

PR #24563, opened 2025-06-13
Fix decode_octal_encoded_utf8()

juj:
Fix decode_octal_encoded_utf8() to work in scenario where a directory name consists of exactly two digits, i.e. '\\foo\\21\\bar'

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -10780,8 +10780,8 @@ int main() {
 
     test('A ä☃ö Z.cpp')
 
-    ensure_dir('inner Z ö☃ä A')
-    test('inner Z ö☃ä A/A ä☃ö Z.cpp', 'inner Z ö☃ä A')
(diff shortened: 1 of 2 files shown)


test/test_other.py
```diff
@@ -10780,8 +10780,8 @@ def test(infile, source_map_added_dir=''):
 
     test('A ä☃ö Z.cpp')
 
-    ensure_dir('inner Z ö☃ä A')
-    test('inner Z ö☃ä A/A ä☃ö Z.cpp', 'inner Z ö☃ä A')
+    ensure_dir('inner Z ö☃ä A/21')
```
sbc100, line 10785: Maybe add comment here as to why we need strange paths in the test?

sbc100 (approved):

test/test_other.py
juj: Good idea

PR #24565, opened 2025-06-13
Don't execute the proxy queue while adding to it from same thread.

brendandahl:
test_pthread_cancel was intermittently failing with a deadlock. From the backtrace, I found we were adding to the queue which can sometimes call malloc->emscripten_yield->emscripten_proxy_execute_queue which also tries to lock the same queue on the same thread.


brendandahl:
Still working on a more reliable test, but I wanted to try this out...

system/lib/pthread/proxying.c
```diff
@@ -123,6 +127,16 @@ void emscripten_proxy_execute_queue(em_proxying_queue* q) {
     executing_system_queue = 1;
   }
 
+  // When the current thread is adding tasks it locks the queue, but we can
+  // potentially try to execute the queue during the add (from
+  // emscripten_yield). This will deadlock the thread, so only try to take the
+  // lock if the current thread is not adding to the queue. We then hope the
+  // queue is executed later when it is unlocked.
+  // XXX: This could leave to starvation if we never process the queue.
+  if (q->active_thread == pthread_self()) {
```
sbc100: How about `if (q->active_thread  && q->active_thread == pthread_self()) {` to avoid calling pthread_self at all in the non-contended case?

sbc100: I wonder if we could combine with the above `executing_system_queue` check?

At least we could use the same technique of using a TLS variable instead of calling pthread_self?   I guess that only works if this issue only applies to the system queue (which I think it does).

sbc100: At least maybe wrap this whole thing in the `is_system_queue` check?

brendandahl: I think there's a few different ways to do this and simplify it. I just wanted to try this out with minimal changes first to make sure it's going down the right path.

tlively: +1 to reusing the existing thread-local variable once this is confirmed to be the right direction. It can be hoisted to the top level for use from multiple functions if necessary.

system/lib/pthread/proxying.c
```diff
@@ -35,6 +35,8 @@ static em_proxying_queue system_proxying_queue = {
   .capacity = 0,
 };
 
+static _Thread_local int system_queue_in_use = 0;
```
sbc100: We could use `bool` here along with `true/false` but maybe thats a bigger/unnecessary change.

test/code_size/random_printf_wasm2js.json
```diff
@@ -1,6 +1,6 @@
 {
-  "a.html": 17233,
-  "a.html.gz": 7547,
-  "total": 17233,
-  "total_gz": 7547
+  "a.html": 17199,
+  "a.html.gz": 7525,
+  "total": 17199,
+  "total_gz": 7525
```
sbc100: How is this is a codesize win!?

Is it perhaps move the TLS init code into the global scope?  If so that is kind of crazy.

sbc100 (commented):
Nice work! 

Hopefully we can also figure out some way to improve the testing ?

test/code_size/random_printf_wasm2js.json
brendandahl: Strange it looks like the code size test is still failing. Maybe this isn't actually a code size win. Will re-try again.

system/lib/pthread/proxying.c
tlively: For whatever reason I decided to use `int` instead of `bool` when I wrote all this, so this is at least consistent with the rest of the file.

system/lib/pthread/proxying.c
```diff
@@ -110,17 +112,24 @@ void emscripten_proxy_execute_queue(em_proxying_queue* q) {
...
+  // Below is a recursion and deadlock guard:
+  // The recursion guard is to avoid infinite recursion when we arrive here from
+  // the pthread_lock call below that executes the system queue. The
+  // per-task_queue recursion lock can't catch these recursions because it can
+  // only be checked after the lock has been acquired.
+  // 
+  // This also guards against deadlocks when adding to the system queue. When
+  // the current thread is adding tasks, it locks the queue, but we can
+  // potentially try to execute the queue during the add (from
+  // emscripten_yield). This will deadlock the thread, so only try to take the
```
tlively:
```suggestion
  // potentially try to execute the queue during the add (from
  // emscripten_yield when malloc takes a lock). This will deadlock the thread, so only try to take the
```

test/code_size/random_printf_wasm2js.json
tlively: Makes sense to me. Function-scope TLS has to do all sorts of implicitly thread-safe lazy initialization that global-scope TLS does not have to do.

test/other/codesize/test_codesize_minimal_pthreads.funcs
```diff
@@ -72,6 +72,7 @@ $emscripten_futex_wake
 $emscripten_stack_get_current
 $emscripten_stack_set_limits
 $free_ctx
+$get_or_add_tasks_for_thread
```
tlively, line 75: I wonder what happened here.

tlively:
Solution LGTM, but I forgot about the testing.

brendandahl:
Changing to using a thread local causes failures in `test_pthread_run_on_main_thread_minimal_runtime`. I didn't get very far debugging this. I won't be able to work on this until June ~25th, so if anyone wants to pick it up feel free to.

brendandahl:
Unfortunately, I still haven't found a reliable way to reproduce this in a test case. Are we fine to land without?

system/lib/pthread/proxying.c
sbc100: This was switch to bool in #24596, so this can be updated now I think.

system/lib/pthread/proxying.c
```diff
@@ -106,22 +108,28 @@ static em_task_queue* get_or_add_tasks_for_thread(em_proxying_queue* q,
...
+  // catch these recursions because it can only be checked after the lock has
+  // been acquired.
+  //
+  // This also guards against deadlocks when adding to the system queue. When
+  // the current thread is adding tasks, it locks the queue, but we can
+  // potentially try to execute the queue during the add (from emscripten_yield
+  // when malloc takes a lock). This will deadlock the thread, so only try to
+  // take the lock if the current thread is not using the queue. We then hope
+  // the queue is executed later when it is unlocked.
+  int is_system_queue = q == &system_proxying_queue;
```
sbc100: bool

system/lib/pthread/proxying.c
```diff
@@ -134,14 +142,21 @@ void emscripten_proxy_execute_queue(em_proxying_queue* q) {
...
   if (is_system_queue) {
-    executing_system_queue = false;
+    system_queue_in_use = false;
   }
 }
 
 static int do_proxy(em_proxying_queue* q, pthread_t target_thread, task t) {
   assert(q != NULL);
   pthread_mutex_lock(&q->mutex);
+  int is_system_queue = q == &system_proxying_queue;
```
sbc100: bool

system/lib/pthread/proxying.c
```diff
@@ -134,14 +142,21 @@ void emscripten_proxy_execute_queue(em_proxying_queue* q) {
...
+  if (is_system_queue) {
+    system_queue_in_use = 1;
+  }
   em_task_queue* tasks = get_or_add_tasks_for_thread(q, target_thread);
+  if (is_system_queue) {
+    system_queue_in_use = 0;
```
sbc100: Use true and false here too

tlively:
Does this test sketch sound plausible?

 1. Main thread spawns thread 1 and spins
 2. Thread 1 asynchronously sends a set-flag message to the main thread
   - This allocates and fills a task_queues buffer of size 1.
 3. Thread 1 signals main thread to continue
 4. Main thread stops spinning and asynchronously sends a nop task to thread 1.
   - This allocates and fills a task_queues buffer of size 2.
 5. Main thread does not deadlock on the previous step
 6. Main thread asserts that the flag is not set, showing that the task queue has not been executed.

Maybe we can also hook the allocator to set flags as well and prove that the allocations happened at the expected times during the test.  We do something similar when testing frees here: https://github.com/emscripten-core/emscripten/blob/main/test/pthread/test_pthread_proxying_refcount.c#L25-L28.

If that doesn't sound plausible, then testing by observing fewer flakes SGTM.

sbc100:
@tlively (I added some numbers to your steps above).  Is the expectation that we could make step 4 deterministically fail without this PR?

tlively:
Yes, that's what I was aiming for. I haven't tried it, though :)

test/pthread/test_pthread_proxy_deadlock.c
```diff
@@ -0,0 +1,47 @@
...
+
+bool should_quit = false;
+pthread_t looper;
+
+// In the actual implementation of malloc the system queue may be executed
+// non-deterministically if malloc is waiting on a mutex. This wraps malloc and
+// executes the system queue during every allocation to make the behavior
+// deterministic.
+void *malloc(size_t size) {
+  if (emscripten_proxy_get_system_queue() && emscripten_is_main_runtime_thread()) {
```
sbc100: emscripten_proxy_get_system_queue() cannot ever return NULL since it returns that address of some static data.

If you want to delay this hook maybe set a variable in main and only start doing this once main is running?

test/pthread/test_pthread_proxy_deadlock.c
```diff
@@ -0,0 +1,47 @@
...
+  while (!should_quit) {
+    emscripten_proxy_execute_queue(emscripten_proxy_get_system_queue());
+    sched_yield();
+  }
+  return NULL;
+}
+
+int main() {
+  pthread_create(&looper, NULL, looper_main, NULL);
+  emscripten_proxy_async(emscripten_proxy_get_system_queue(), looper, run_on_looper, NULL);
```
sbc100: Could this ever set should_quit before looper_main actually starts running?

Can you add some comments somewhere in this file about exactly that case we are testing for.

sbc100 (approved):
lgtm % test comments

test/pthread/test_pthread_proxy_deadlock.c
brendandahl: This was leftover from a previous version where I wasn't using the system queue.

test/pthread/test_pthread_proxy_deadlock.c
brendandahl: I kind of assumed the thread start would be guaranteed to run before any proxy messages, but now I'm not sure. Even if it does run before, we'd still hit a deadlock since the main thread will be locked during task creation and trying to run the queue and lock again. It doesn't really matter what the other thread is doing.

tlively: You can have `looper_main` set a flag and have `main` wait for that flag to be set before doing the proxying to be sure. +1 to adding comments, especially about how the deadlock would occur without the bug fix.

brendandahl: I changed the test a bit more and removed the looper thread and just sends the task to the same thread. This makes the test a little less realistic, but highlights it doesn't really matter what the thread is doing.

test/pthread/test_pthread_proxy_deadlock.c
```diff
@@ -0,0 +1,40 @@
...
+
+int main() {
+  // Tests for a deadlock scenario that can occur when sending a task.
+  // The sequence of events is:
+  // 1. Sending a task locks the queue.
+  // 2. Allocating a new task queue calls malloc.
+  // 3. Malloc then attempts to execute and lock the already-locked queue,
+  //    causing a deadlock.
+  // This test ensures our implementation prevents this re-entrant lock.
+  emscripten_proxy_async(emscripten_proxy_get_system_queue(), pthread_self(), task, NULL);
```
sbc100, line 38: Oh nice, so you don't even need a second thread!  Makes sense I guess since this is a double-lock/self-deadlock issue.

Is it valid to send work to yourself?

brendandahl: test_pthread_proxying.c has a test for proxing to itself, so I assume so.

tlively: Yep, but only asynchronous self-proxying is allowed, since synchronous self-proxying would obviously deadlock.

test/other/codesize/test_codesize_minimal_pthreads.funcs
sbc100: I guess this was previously inlined in do_proxy?

PR #24567, opened 2025-06-13
Fix `makeRemovedFSAssert()` resolved path

adamscott:
This fixes an issue where `path.resolve()` resolves using the CWD (which is usually the target project path) instead of the directory of the script itself.

The current behavior [causes issues with the Closure compiler](https://github.com/godotengine/godot/pull/107460#issuecomment-2970895256), as stubs are erroneously created instead of being skipped in `src/shell.js`. Closure then complains that variables are declared twice.

The issue was caused by #23348 which changed the logic of the search.

### DIFF
--- a/src/parseTools.mjs
+++ b/src/parseTools.mjs
@@ -19,6 +19,7 @@ import {
   runInMacroContext,
   pushCurrentFile,
   popCurrentFile,
+  localFile,
   warn,
   srcDir,
 } from './utility.mjs';
@@ -921,7 +922,7 @@ function makeModuleReceiveWithVar(localName, moduleName, defaultValue) {
 function makeRemovedFSAssert(fsName) {
   assert(ASSERTIONS);
   const lower = fsName.toLowerCase();
-  if (JS_LIBRARIES.includes(path.resolve(path.join('lib', `lib${lower}.js`)))) return '';
+  if (JS_LIBRARIES.includes(localFile(path.join('lib', `lib${lower}.js`)))) return '';
   return `var ${fsName} = '${fsName} is no longer included by default; build with -l${lower}.js';`;
 }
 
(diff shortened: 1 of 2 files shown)


src/parseTools.mjs
```diff
@@ -921,7 +922,12 @@ function makeModuleReceiveWithVar(localName, moduleName, defaultValue) {
 function makeRemovedFSAssert(fsName) {
   assert(ASSERTIONS);
   const lower = fsName.toLowerCase();
-  if (JS_LIBRARIES.includes(path.resolve(path.join('lib', `lib${lower}.js`)))) return '';
+  if (
+    JS_LIBRARIES.includes(
+      path.join(path.dirname(fileURLToPath(import.meta.url)), 'lib', `lib${lower}.js`),
```
sbc100: I think you can do `localFile(path.join('lib', ``lib${lower}.js``))` perhaps.

Can we add a test case maybe, to be sure its fixed?

sbc100: For a test I think you perhaps do:

1. Build hello world.
2. Check that `IDBFS is no longer included by default` is in the generated file
3. Build again with `-lidbfs`
4. Check that `IDBFS is no longer included by default` is in not in the generated file.

adamscott:
```
emscripten on  resolve-actual-lib [$!?] via  v22.15.0 via 🐍 v3.12.8
❯ test/runner other.test_parsetools_make_removed_fs_assert
Running test_other: (1 tests)
test_parsetools_make_removed_fs_assert (test_other.other.test_parsetools_make_removed_fs_assert)
This tests that parseTools.mjs `makeRemovedFSAssert()` works as intended, ... ok

Ran 1 test in 0.631s

OK

```

sbc100 (approved):

PR #24575, opened 2025-06-15
Fix asynchronous GLUT main loop utilizing tests that raced against reftest XHR.

juj:


test/float_tex.c
```diff
@@ -160,6 +166,11 @@ int main(int argc, char *argv[]) {
     /* Set up glut callback functions */
     glutDisplayFunc(glut_draw_callback);
     gl_init();
+#ifdef __EMSCRIPTEN__
+    // This test kicks off an asynchronous glutMainLoop(), so do not perform a synchronous
+    // reftest immediately after falling out from main.
+    EM_ASM({ delete Module['postRun']; });
```
sbc100: It looks like we have a reftestBlock / reftestUnblock API for this.  See #22672

juj: Done

test/float_tex.c
```diff
@@ -160,6 +166,11 @@ int main(int argc, char *argv[]) {
     /* Set up glut callback functions */
     glutDisplayFunc(glut_draw_callback);
     gl_init();
+#ifdef __EMSCRIPTEN__
+    // This test kicks off an asynchronous glutMainLoop(), so do not perform a synchronous
+    // reftest immediately after falling out from main.
+    EM_ASM({reftestBlock()});
+#endif
```
sbc100, line 173: Is this change is needed then how/why is this test passing today?    I would assume the reftest is triggered by the first swapbuffers or something like that.. but I can't remember now?

juj: The reftest is triggered in Module['postRun']().

It probably passes because rAF() takes only 16 msecs, and XHRs from disk are likely slower than that, so one or more of rAF()s squeeze in to finish before the XHR completes.

sbc100: Perhaps reftest its triggers both by `postRun` and also by intercepting `window.requestAnimationFrame`.  Perhaps we should have some kind of way to tell the reftest code to wait for RAF instead of also running as part of `postRun`?

Or perhaps we can/should replace `postRun` with `onExit` (which wont' be called when main returns but only when the whole program exits)?

PR #24576, opened 2025-06-15
Fix race condition in browser.test_clientside_vertex_arrays_es3

juj:
Without this, browser would run the reftest as a Module['postRun'] trigger immediately after finishing main().

But because reftest is asynchronous, as it has to load an image asynchronously, the requestAnimationFrame() timers might race to squeeze in the needed test tick function call to render the expected image, or they might not, if the XHR of the reftest image is fast.

### DIFF
--- a/test/clientside_vertex_arrays_es3.c
+++ b/test/clientside_vertex_arrays_es3.c
@@ -19,7 +19,7 @@ void onDraw(void* arg) {
   glfwSwapBuffers(window);
   glfwPollEvents();
 #ifdef __EMSCRIPTEN__
-  EM_ASM(doReftest()); // All done, perform the JS side image comparison reftest.
+  EM_ASM({reftestUnblock()}); // All done, perform the JS side image comparison reftest.
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


test/clientside_vertex_arrays_es3.c
```diff
@@ -112,6 +115,10 @@ int main() {
   }
 
 #ifdef __EMSCRIPTEN__
+  // This test kicks off an asynchronous main loop, so do not perform a synchronous
+  // reftest immediately after falling out from main.
+  EM_ASM({ delete Module['postRun']; });
```
sbc100: Use reftestBlock/reftestUnlock API

juj: Updated

sbc100:
Can you remind me what the race condition is here exactly?  (Perhaps you can give a brief description in the PR description).

juj:
ok, updated description.

sbc100:
Hmm.. I wonder if we can/should instead have the reftest be more smart here.. so that we can avoid changing each of these tests.

I'm happy to see this change land as-is, but I feel like there should be a cleaner way to do this.

sbc100 (approved):

juj:
If there is a better design, would be happy for it - I don't unfortunately have time for larger refactors, but I need to work the CI to show up green in order to start merging our Emscripten fork PRs to upstream.

PR #20383, opened 2023-10-03
[emval] Reuse method calling optimisation in regular calls

RReverser:
Method, function and constructor calls are very similar in nature but used different code paths: method calls have an optimisation where an optimised caller trampoline is generated once and reused for each function signature, whereas the other two serialized arguments and types every time.

There is no reason for them to diverge, so I made function calls and constructors reuse the same optimisation.

Results on a silly microbenchmark running a side-effect-ful JS function with 100 integer args in a loop:

#### with dynamic execution:

as method: 224.504ns
as function: 5367.9ns

#### without dynamic execution:

as method: 845.044ns
as function: 5350.96ns

#### with dynamic execution:

as method: 262.74ns
as function: 246.593ns

#### without dynamic execution:

as method: 859.534ns
as function: 823.498ns

In real world the difference is unlikely to be perceptible, but at the very least this makes the code more consistent and deletes a bunch of duplication.

PR #13889, opened 2021-04-13
Add int64_t support to embind using bigint support

cuberoot:
This PR adds int64_t and uint64_t support to embind when the `-s WASM_BIGINT` flag is used (see issue 11726).

Emscripten has support for handling `int64_t` and `uint64_t` as JS bigint values, but embind does not know anything about those values.

This PR is incomplete. I need to add tests and work out how to add guards so that this change does not cause issues when the `-s WASM_BIGINT` is not used.

PR #24577, opened 2025-06-16
[embind] Use a single invoker mechanism

### LINKED ISSUE #24547: val.call<T>() can't return 64-bit integers
Simple demo (can be reproduced in `test_i64_val.cpp`):

```cpp
val::global().call<int64_t>("BigInt", max_int64_t);
```

fails with:

```
TypeError: Cannot convert a BigInt value to a number
    at wasm://wasm/000d1786:wasm-function[165]:0x66f9
    at wasm://wasm/000d1786:wasm-function[85]:0x44b2
    at wasm://wasm/000d1786:wasm-function[55]:0x381d
    at wasm://wasm/000d1786:wasm-function[41]:0x2810
    at wasm://wasm/000d1786:wasm-function[180]:0x6ad0
    at D:\emscripten\out\test\test_i64_val.js:629:12
    at callMain (D:\emscripten\out\test\test_i64_val.js:5377:15)
    at doRun (D:\emscripten\out\test\test_i64_val.js:5430:24)
    at run (D:\emscripten\out\test\test_i64_val.js:5443:5)
    at removeRunDependency (D:\emscripten\out\test\test_i64_val.js:584:7)
    at receiveInstance (D:\emscripten\out\test\test_i64_val.js:734:5)
    at receiveInstantiationResult (D:\emscripten\out\test\test_i64_val.js:752:12)
    at createWasm (D:\emscripten\out\test\test_i64_val.js:778:17)
Thrown at:
    at $func165 (wasm://wasm/000d1786:1:26362)
    at $func85 (wasm://wasm/000d1786:1:17587)
    at $func55 (wasm://wasm/000d1786:1:14366)
    at $func41 (wasm://wasm/000d1786:1:10257)
    at $main (wasm://wasm/000d1786:1:27345)
    at D:\emscripten\out\test\test_i64_val.js:629:12
    at callMain (D:\emscripten\out\test\test_i64_val.js:5377:15)
    at doRun (D:\emscripten\out\test\test_i64_val.js:5430:24)
    at run (D:\emscripten\out\test\test_i64_val.js:5443:5)
    at removeRunDependency (D:\emscripten\out\test\test_i64_val.js:584:7)
```

This is due to generic wiretype being a `double`. I imagine the easiest fix would be changing so that the return value is stored on the stack instead of being actually returned from the function, so that you can store bigints via `HEAP64` and all other numbers via `HEAPF64` in the same ABI. cc @brendandahl

RReverser:
This is the next step in refactorings I started back in #20383 to merge all Embind generic method callers into a single mechanism. I never got around to landing this PR because I kept trying to split it into even smaller changes / steps, never found a good way to do so due to how entangled they were, got frustrated and left it alone.

However, aside from being a nice-to-have optimisation or cleanup refactoring, it's also an easy way to deal with issues like #24547 which makes it a bit more pressing. Without this PR, we'd have to duplicate the `emval_as`+`emval_as_int64`+`emval_as_uint64` fix from #13889 for `emval_call` and `emval_call_method` as well, but after this PR everything goes through the same centralised implementation where we can deal with all Embind types in a consistent manner.

There are some more things that I'd like to clean up here (including some of the complexity added by this very own PR), but in order to try and keep review scope smaller, I'm going to submit them separately.

Fixes #24547.


src/lib/libemval.js
```diff
@@ -305,26 +271,38 @@ var LibraryEmVal = {
 
 #if !DYNAMIC_EXECUTION
     var argN = new Array(argCount);
-    var invokerFunction = (obj, func, destructorsRef, args) => {
+    var invokerFunction = (handle, methodName, destructorsRef, args) => {
```
sbc100, line 274: Would it make more sense to call the first argument `receiver`?  Or is `handle` (the type of the receiver I guess?) more descriptive here?

test/code_size/embind_val_wasm.json
```diff
@@ -1,10 +1,10 @@
...
-  "a.wasm": 9097,
-  "a.wasm.gz": 4696,
-  "total": 15334,
-  "total_gz": 7614
+  "a.js": 5748,
+  "a.js.gz": 2563,
+  "a.wasm": 9101,
+  "a.wasm.gz": 4698,
+  "total": 15401,
+  "total_gz": 7641
```
sbc100, line 9: Nice to see a code size saving here!

test/embind/test_i64_val.cpp
```diff
@@ -43,33 +57,16 @@ int main() {
...
-  assert(val::global()["a"].as<int64_t>() == max_int64_t);
-
-  val::global().set("a", val(min_int64_t));
-  ensure_js(compare_a_64_js(min_int64_t));
-  assert(val::global()["a"].as<int64_t>() == min_int64_t);
+  test_value(int64_t(1234));
+  test_value(int64_t(-4321));
+  test_value(int64_t(0x12345678aabbccddL));
+  test_value(min_int64_t);
+  test_value(max_int64_t);
```
sbc100, line 69: Does this test change need to happen as part of this change?

What I'm kind of aksing here is, does the unmodified/old test no longer pass/work after this change?   Maybe this test improvement can land separately / first?

sbc100 (commented):
I can't say I understand all of this but look like a good direction. I leave @brendandahl to do a full review.

test/code_size/embind_val_wasm.json
RReverser: You probably misread, it's the other way around unfortunately - slight increase :) Kinda expected when you extend to support more types / fix bugs though.

src/lib/libemval.js
RReverser: `handle` as in generally `EM_VAL`. If the invoker kind is "method" then it's a handle of the receiver, if it's a "function" or a "constructor" then it's a handle of the function itself, and if it's a cast, then it's just null... A handle seemed like the most generic name.

test/embind/test_i64_val.cpp
RReverser: The extended tests are a regression test for https://github.com/emscripten-core/emscripten/issues/24547 which this PR fixes as per description. If I'd land it separately, it would fail without the corresponding fix.

Grouping them into a function is just the easiest way to add those "function call" and "method call" tests to both int64 and uint64 and all the tested values at once.

sbc100: At to be clear the old test prior to this PR still passes with the changes in this PR?

RReverser: Yeah, it just got extended to also test equivalent of `BigInt(value)` and `globalThis.BigInt(value)` in addition to pre-existing `a = value;`.

system/include/emscripten/val.h
```diff
@@ -651,19 +568,37 @@ class EMBIND_VISIBILITY_DEFAULT val {
...
     WireTypePack<Args...> argv(std::forward<Args>(args)...);
     EM_DESTRUCTORS destructors = nullptr;
-    EM_GENERIC_WIRE_TYPE result = impl(
-      Signature<Kind, Ret, Args...>::get_method_caller(),
-      as_handle(),
-      &destructors,
-      argv);
+
+    RetWire result;
+    if constexpr (std::is_integral<RetWire>::value && sizeof(RetWire) == 8) {
```
brendandahl, line 584: Hmm...do we not test val with c++11? The rest of embind is c+11 compatible, though we've talked about bumping it to c++17.

brendandahl (changes requested):

system/include/emscripten/val.h
RReverser: We do, and I had to fix this line a few times to make it work for C++11.

brendandahl: Isn't `if constexpr`c++17 only?

RReverser: Hmm I see what you mean, apparently `if constexpr ...` is not really C++11? But it does pass the test, and we already used it in e.g. `val::array` implementation.

Maybe Clang chose to implement it earlier as an extension or something?

RReverser: Yeah apparently it's just a warning `-Wc++17-extensions` in `-std=c++11` and, like Clang itself, we disable those extension warnings for system libs.

brendandahl: Cool, didn't realize that it was a warning only. I've definitely wanted to use that in a few other places in embind.

RReverser:
The wasm2js failure looks a bit suspicious:

> error: undefined symbol: _emscripten_tempret_set (referenced by $setTempRet0, referenced by setTempRet0, referenced by _emval_invoke_i64, referenced by root reference (e.g. compiled C/C++ code))
warning: __emscripten_tempret_set may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
em++: error: '/root/emsdk/node/22.16.0_64bit/bin/node /root/project/tools/compiler.mjs -' failed (returned 1)

I guess it comes from `makeReturn64`, but then if the usage is injected automatically, shouldn't the dependency be added automatically as well by jsify? @sbc100

system/include/emscripten/val.h
RReverser: Yeah that was confusing for a moment. So are you okay with me leaving it as-is? I can refactor to use pure C++11 SFINAE I guess, but if it works, it works...

brendandahl: I'm fine with you leaving it. Especially since we already have one that you pointed out already.

sbc100:
Yes, `_emscripten_tempret_set` is listed as a native dependency of `$setTempRet`.   

If you build with `EMCC_DEBUG=1` should should see something like `/tmp/emscripten_temp/tmpz6ttjgoflibemscripten_js_symbols.so` being passed to the linker which is a map of JS dependencies.  

However this only handles native->JS dependencies, but what we have here is a case of the JS compiler itself generating calls to a JS function (effectively a JS->JS dependency), but I guess its being injected somehow too late (and the native link has already been done).

If you can share a repro case I can take a look.

RReverser:
> If you can share a repro case I can take a look.

I don't have a smaller one, I've only guessed what's happening by searching for keywords from the error message for now.

sbc100:
Does it fix the issue if you make the condition on line 536 of jsifier.js match the one on 314.  It looks like they should be the same condition but they are not.

RReverser:
No, still the same. I even tried making that dep unconditional just in case, but nothing:

> .\test\runner wasm2js1.test_embind_unsigned
Running test_core: (1 tests)
test_embind_unsigned (test_core.wasm2js1) ... error: undefined symbol: _emscripten_tempret_set (referenced by $setTempRet0, referenced by setTempRet0, referenced by $ExitStatus, referenced by root reference (e.g. compiled C/C++ code))

Worth noting that `setTempRet0` itself is discovered fine, so that condition shouldn't matter - the linker is complaining specifically about `_emscripten_tempret_set`.

Looks like that one is implemented in assembly, maybe it needs to be somehow exposed to wasm2js conversion?

system/include/emscripten/val.h
RReverser: Ok thanks.

RReverser:
It's all the weirder that `wasm2js0.test_embind_unsigned` and `wasm2js2.test_embind_unsigned` pass fine but `wasm2js1.test_embind_unsigned` fails with said linker error.

Aren't those digits at the end just optimisation levels?

sbc100:
yes, you can see the setTempRet0 dependency is supposed to be added on line 530 (`deps.push('setTempRet0')`).   Not that this needs to happen before the native link happens and the end result is that the native dependencies of `setTempRet0` should be passed to wasm-ld with `--export-if-defined=XX` (for each native JS dependency).

RReverser:
Sorry, I still don't understand.

> yes, you can see the setTempRet0 dependency is supposed to be added on line 530 (`deps.push('setTempRet0')`).

But then why even making it unconditional doesn't fix the issue? And why is `setTempRet0` the problem if it's added successfully already, just its own native dep isn't? And why does this not happen in `wasm2js0` and `wasm2js2`?

I don't understand the mechanism here at all, so I have more questions than answers 😅

sbc100:
> Sorry, I still don't understand.
> 
> > yes, you can see the setTempRet0 dependency is supposed to be added on line 530 (`deps.push('setTempRet0')`).
> 
> But then why even making it unconditional doesn't fix the issue? And why is `setTempRet0` the problem if it's added successfully already, just its own native dep isn't? 

Answering just this part:  We process the JS libraries files twice.  Once in `--symbol-only` mode where we extract dependencies.   We use this dependency information for figure out which native symbols we might need to `--export-if-defined`. 

I think this issue is that while `setTempRet0` is added to the deps here, it never actually ends up in `DEFAULT_LIBRARY_FUNCS_TO_INCLUDE` out on the python side.  I think this is likely we use when searching for which native symbols to explicitly export with `--export-if-defined`

The code for this is here: https://github.com/emscripten-core/emscripten/blob/542dc42ab75c98f41712208311c77f0e2864129a/tools/link.py#L3104-L3128

Specifically I think we are missing a call to `add_js_deps` for `getTempRet0`.

> And why does this not happen in `wasm2js0` and `wasm2js2`?

I imagine this is because at O0 getTempRet0 and/or its dependencies are exported for other reasons.

RReverser:
> I imagine this is because at O0 getTempRet0 and/or its dependencies are exported for other reasons.

That would make sense, but then why does O2 pass? Shouldn't it be more restrictive than O1 basically always?

Could you help me fix this one please? I still don't understand what the fix should look like.

RReverser:
Wait it passed after a rebase, wat? Nothing relevant changed upstream that I can see.

RReverser:
The CI failure `other.test_jspi_async_function` seems unrelated, likely a bad combination of https://github.com/emscripten-core/emscripten/pull/24578 + https://github.com/emscripten-core/emscripten/pull/24550 - I guess they raced on CI. @brendandahl

brendandahl (approved):

RReverser:
@brendandahl Can you please fix CI? I don't think I can merge until then (unless someone with force-merge rights wants to do it).

RReverser:
> Can you please fix CI?

For #24550 I mean.

sbc100:
> > Can you please fix CI?
> 
> For #24550 I mean.

https://github.com/emscripten-core/emscripten/pull/24590

RReverser:
> FAIL [1.119s]: test_fetch_redirect (test_browser.browser64_4gb)

Given how often it flakes lately, it really should use a local server instead of relying on httpbin.org... @sbc100 can you merge manually please? I already did rerun and it takes forever only to run into another flake.

sbc100:
> > FAIL [1.119s]: test_fetch_redirect (test_browser.browser64_4gb)
> 
> Given how often it flakes lately, it really should use a local server instead of relying on httpbin.org... @sbc100 can you merge manually please? I already did rerun and it takes forever only to run into another flake.

Agreed.  Would you open a bug for that?

RReverser:
Thanks, and done: https://github.com/emscripten-core/emscripten/issues/24597

PR #24160, opened 2025-04-21
Add support for -fcoverage-mapping support

arsnyder16:
Main introduction is porting `llvm-project/compiler_rt/lib/profile` `into system/lib/compiler_rt/lib/profile`

Related
https://github.com/emscripten-core/emscripten/pull/13072
https://github.com/llvm/llvm-project/pull/111332

No modifications were required for what was ported. llvm-cov worked fine for me with simple project. 

* `-g` is required to produce a properly instrumented binary `error: failed to load coverage: 'my.wasm': no coverage data found`

PR #24580, opened 2025-06-16
[compiler-rt] Build with -DCOMPILER_RT_HAS_ATOMICS

sbc100:
It looks like without this the fallback functions are not actually thread safe:

https://github.com/llvm/llvm-project/blob/63b80dd01dafc92104ee43e4f0f5296d644c25ec/compiler-rt/lib/profile/InstrProfilingUtil.c#L102-L119

Followup to #24160

### DIFF
--- a/tools/system_libs.py
+++ b/tools/system_libs.py
@@ -939,7 +939,7 @@ class libcompiler_rt(MTLibrary, SjLjLibrary):
   # restriction soon: https://reviews.llvm.org/D71738
   force_object_files = True
 
-  cflags = ['-fno-builtin', '-DNDEBUG', '-DCOMPILER_RT_HAS_UNAME=1']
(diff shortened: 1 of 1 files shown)


sbc100:
is it me or are these fallback functions kind of dangerous / wrong?

tlively:
What looks wrong about them? They look like reasonable single-threaded implementations of compare-and-swap and fetch-add.

sbc100:
Right, but as we have shown by the need for this PR the `COMPILER_RT_HAS_ATOMICS` might not correspond 1-to-1 with single-threaded-only.

In this case we were building without any macros and it was assuming incorrectly that single-theaded version could be substituted (even on multi-threaded builds).  Perhaps it would be safer to invert the macro and have COMPILER_RT_SINGLE_THREADED_ONLY or some such.

Issue #24579, opened 2025-06-16
🐛 The `--emit-tsd` option generated TypeScript type for the `args` parameter of the `ccall` function should use `IArguments` instead of `Arguments`


alpinebuster:
**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.8 (70404efec4458b60b953bc8f1529f2fa112cdfd1)
clang version 21.0.0git (https:/github.com/llvm/llvm-project 23e3cbb2e82b62586266116c8ab77ce68e412cf8)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: ~/emsdk/upstream/bin

**Failing command line in full:**
Nan

**Full link command and output with `-v` appended:**
The compilation process works fine, and the function can be called normally in JavaScript. However, when wrapping it in TypeScript as npm package, there's a type error with the use of `Arguments`.

```cmake
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -lembind -s EXPORT_ES6=1 -s MODULARIZE=1 -s EXPORT_NAME=createModule -s SINGLE_FILE=0 -s 'EXPORTED_FUNCTIONS=[\"_malloc\", \"_free\"]' -s 'EXPORTED_RUNTIME_METHODS=[\"ccall\", \"HEAPU8\", \"FS\"]' --emit-tsd ${PROJECT_NAME}.d.ts")
```

The generated `.d.ts` file is as follows:

```ts
// TypeScript bindings for emscripten-generated code.  Automatically generated at compile time.
declare namespace RuntimeExports {
    /**
     * @param {string|null=} returnType
     * @param {Array=} argTypes
     * @param {Arguments|Array=} args
     * @param {Object=} opts
     */
    function ccall(ident: any, returnType?: (string | null) | undefined, argTypes?: any[] | undefined, args?: (Arguments | any[]) | undefined, opts?: any | undefined): any;
    let HEAPU8: any;

    ...
}
```

The error message is:

```txt
Cannot find name 'Arguments'. Did you mean 'IArguments'?ts(2552)
```

PR #10525, opened 2020-02-21
Fix more Closure warnings

juj:
Getting closer to Closure clean.

PR #21331, opened 2024-02-13
[embind] Minor refactor to use more modern JS features

sbc100:

PR #24583, opened 2025-06-16
Remove `Arguments` from jsdoc

sbc100:
Looks like this was originally added in #10525.

Perhaps closure compiler used to have a distinct type for `arguments` builtin in JS?  I can't see any mention of it in closure compiler documentation today though.

Fixes: #24579

### DIFF
--- a/src/lib/libccall.js
+++ b/src/lib/libccall.js
@@ -24,7 +24,7 @@ addToLibrary({
   /**
    * @param {string|null=} returnType
    * @param {Array=} argTypes
-   * @param {Arguments|Array=} args
+   * @param {Array=} args
    * @param {Object=} opts
    */`,
   $ccall: (ident, returnType, argTypes, args, opts) => {


RReverser:
> Perhaps closure compiler used to have a distinct type for `arguments` builtin in JS?

It would make sense, since `arguments` object is historically special and isn't actually an array (eg it lacks all the array methods, and you would want typechecker like Closure to catch you using them by accident).

RReverser:
> I can't see any mention of it in closure compiler documentation today though.

Not documentation, but here's the declaration of extern: https://github.com/google/closure-compiler/blob/71975c675e458b9bfed8fe0e737d388a82b71918/externs/es3.js#L230

sbc100:
> > Perhaps closure compiler used to have a distinct type for `arguments` builtin in JS?
> 
> It would make sense, since `arguments` object is historically special and isn't actually an array (eg it lacks all the array methods, and you would want typechecker like Closure to catch you using them by accident).

But we don't use `arguments` any more in the code base right?  So this should no longer be needed.   Especially since it seems to confuse TSC.

RReverser:
> But we don't use `arguments` any more in the code base right?

I don't know, haven't checked, just saw the PR description and wanted to clarify they're not the same. If it's not used, go for it of course.

sbc100:
> > But we don't use `arguments` any more in the code base right?
> 
> I don't know, haven't checked, just saw the PR description and wanted to clarify they're not the same. If it's not used, go for it of course.

I'm pretty sure it was needed for the cwrap implemention which was converted to the rest operator in #21331

PR #24584, opened 2025-06-16
Fix Ninja support

### LINKED ISSUE #24393: Ninja support for Windows is broken
I used Ninja support on Windows successfuly in the past, but it seems it got broken by #24111 and now results in lots of errors like:

```
FAILED: D:/emscripten/cache/build/libGL-getprocaddr/gl.o
D:\emscripten\emcc.bat -MD -MF D:/emscripten/cache/build/libGL-getprocaddr/gl.o.d -O2 -Wall -fno-unroll-loops -Oz -fno-inline-functions -g -sSTRICT -Werror -sRELOCATABLE -DEMSCRIPTEN_DYNAMIC_LINKING '-ffile-prefix-map=D:\emscripten=/emsdk/emscripten' '-ffile-prefix-map=..\..\..=/emsdk/emscripten' -fdebug-compilation-dir=/emsdk/emscripten -DMAX_WEBGL_VERSION=1 -DGL_ENABLE_GET_PROC_ADDRESS=1 -c D:/emscripten/system/lib/gl/gl.c -o D:/emscripten/cache/build/libGL-getprocaddr/gl.o
clang: error: no such file or directory: ''-ffile-prefix-map=D:\emscripten=/emsdk/emscripten''; did you mean '-ffile-prefix-map=D:\emscripten=/emsdk/emscripten''?
clang: error: no such file or directory: ''-ffile-prefix-map=..\..\..=/emsdk/emscripten''; did you mean '-ffile-prefix-map=..\..\..=/emsdk/emscripten''?
```

`shlex.join` uses `shlex.quote` under the hood, and as the warning [on `shlex.quote` docs](https://docs.python.org/3/library/shlex.html#shlex.quote) says, it is only suitable for Unix shells:

> Warning The shlex module is only designed for Unix shells.
The [quote()](https://docs.python.org/3/library/shlex.html#shlex.quote) function is not guaranteed to be correct on non-POSIX compliant shells or shells from other operating systems such as Windows. Executing commands quoted by this module on such shells can open up the possibility of a command injection vulnerability.

We worked around that issue by having a custom `shlex_quote` implementation before, which worked correctly across different platforms by manually quoting arguments with double-quotes, and TBH I think we should just revert to doing that unless there is a simpler cross-platform solution.

RReverser:
Fixes #24393 by using subprocess.list2cmdline instead of shlex.join.

Unfortunately, subprocess.list2cmdline is undocumented but apparently has been available for well over a decade (eg see https://stackoverflow.com/questions/12130163/how-to-get-resulting-subprocess-command-string) and, judging by discussions like https://discuss.python.org/t/why-is-subprocess-list2cmdline-not-documented/25044 is unlikely to be ever removed since it's been long used in the wild.

It does exactly what we want - quote & join a command by Windows rules.


tools/system_libs.py
```diff
@@ -183,13 +184,15 @@ def create_ninja_file(input_files, filename, libname, cflags, asflags=None, cust
   if asflags is None:
     asflags = []
 
+  join_cmd = shlex.join if os.name == 'posix' else subprocess.list2cmdline
```
sbc100, line 187: Can we just use `subprocess.list2cmdline` in all cases?

sbc100: Oh I see its a windows-specific function.  I guess this will have to do

sbc100: Maybe we should re-add `shlex_join` to `utils.py` and use that instead of `shlex.join` everywhere.   Effectivly reverting #24111 but with a cleaner `shlex_join` impl?

RReverser: Maybe, but not sure.

As I said in https://github.com/emscripten-core/emscripten/issues/24393#issuecomment-2899562539, it does look like all other usages are for debug purposes only so syntax doesn't matter as much, and for command invocations we normally use `subprocess` module which does the right thing via native APIs automatically, so Ninja usecase - where we have to stringify a command with a valid OS-specific syntax - seems somewhat special in that regard.

If nothing else pops up, I wouldn't bother generalising it for now.

sbc100: OK sgtm

PR #24586, opened 2025-06-17
Reduce usage of POINTER_TYPE in JS library code

RReverser:
Same as https://github.com/emscripten-core/emscripten/pull/24490 but for the 17 usages of SIZE_TYPE which should always be the same as `*`.


sbc100:
I was thinking I would prefer to keep SIZE_TYPE, but instead make be semantically the same as `*`.  I have a PR to that effect but it was running into issues.

RReverser:
> I was thinking I would prefer to keep SIZE_TYPE, but instead make be semantically the same as `*`.

Maybe we could add an actual string literal to reflect it then? Both `usize` (for current `SIZE_TYPE`) and `isize` (for current `LONG_TYPE`) could be useful.

RReverser:
Closing since this is not how we want to proceed.

Issue #24587, opened 2025-06-17
Deprecate WasmOffsetConverter?


RReverser:
Judging by description in https://github.com/emscripten-core/emscripten/blob/39d93bea3c3abb6c6cbcafe75a97c7bfd0dc3527/src/wasm_offset_converter.js#L13-L19, it was created to work around a V8 bug that has now been long fixed, so seems like a good candidate for cleanup.

sbc100:
It looks like we use wasmOffsetConverter both in `convertFrameToPC` and also in `emscripten_pc_get_function`.   While the former looks like its no longer needed in modern browser I don't quite see how that latter usage could be replaced?    Am I missing something?

RReverser:
> I don't quite see how that latter usage could be replaced?

From a quick glance, Wasm functions should be able to use the UNWIND_CACHE mechanism already used by JS functions. It's populated by `jsStackTrace()` + `convertFrameToPC`, and the former will return the raw stacktrace for both JS and Wasm frames, while the latter, like you said, can be updated to handle Wasm w/o offset converter.

RReverser:
Even ignoring caching, here's what the state looks like in `emscripten_pc_get_function`:

![Image](https://github.com/user-attachments/assets/26afa330-f532-4670-8b75-ad9939668775)

That is, it's just a matter of splitting `jsStackTrace()` into lines like we do in other places and searching for the line with that `586` hex address, which in this case would be `at test_return_address.wasm.main (wasm://wasm/test_return_address.wasm-000a1cc2:wasm-function[10]:0x586)`.

From that you get the expected Wasm name `main` (in debug mode).

RReverser:
In fact, if I forcibly capture the unwind cache via `emscripten_stack_snapshot()` (IIUC `emscripten_pc_get_function` is meant to be used only with it anyway), I can see that it already collects Wasm entries too:

![Image](https://github.com/user-attachments/assets/6c2b7219-a199-4fe9-b050-f8a3d3d037b0)

So it's just a matter of extracting Wasm function names with separate regex but otherwise similarly to how JS names are already extracted.

Or even leaving the same regex, you'll just get slightly more verbose names like `test_return_address.wasm.main` but maybe that's fine.

PR #24588, opened 2025-06-17
Update wasm_offset_converter. NFC

sbc100:
- Convert to ES class
- Use string templates

### DIFF
--- a/src/source_map_support.js
+++ b/src/source_map_support.js
@@ -4,96 +4,95 @@
  * SPDX-License-Identifier: MIT
  */
 
-/**
- * @constructor
- */
-function WasmSourceMap(sourceMap) {
-  this.version = sourceMap.version;
-  this.sources = sourceMap.sources;
-  this.names = sourceMap.names;
+class WasmSourceMap {
+  mapping = {};
+  offsets = [];
 
-  this.mapping = {};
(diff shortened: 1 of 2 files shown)


sbc100:
View with "hide whitespace"

RReverser:
But is it still necessary at all? (see https://github.com/emscripten-core/emscripten/issues/24587, I think it can be just removed nowadays?)

sbc100:
Yes, I was inspired to create this while trying to figure out if this is still needed.

My guess is that it is still needed for things ASAN, but I will do some more investigation.

RReverser:
If you're referring to usage in https://github.com/emscripten-core/emscripten/blob/39d93bea3c3abb6c6cbcafe75a97c7bfd0dc3527/src/lib/libstack_trace.js#L112-L120, then first branch should be sufficient nowadays as stacktrace format was standardised as part of [Web API](https://www.w3.org/TR/wasm-web-api-1/#conventions).

sbc100:
You may well the correct yes, but can we land this change anyway, and attempt to delete this code as a followup?

sbc100:
> It does feel a bit like unnecessary work if it turns out we can delete it right after, but sure, why not :)

I think I agree, but since I already did it ...

PR #24591, opened 2025-06-17
JS FS: Don't return ELOOP on a path with a large number of ..'s

hoodmane:
Decrement nlinks before doing `continue linkloop;` due to a .. at a file system loop. We're not actually traversing a symlink here.

Caught by the Python test suite.


src/lib/libfs.js
```diff
@@ -199,6 +199,9 @@ FS.staticInit();`;
             current_path = PATH.dirname(current_path);
             if (FS.isRoot(current)) {
               path = current_path + '/' + parts.slice(i + 1).join('/');
+              // We're making progress here, don't let many consecutive ..'s
+              // lead to ELOOP
+              nlinks --;
```
sbc100: I don't think we normally use space before `--` or `++`

test/test_other.py
```diff
@@ -16084,3 +16084,19 @@ def test_em_js_bool_macro_expansion(self):
...
+    path = "/".join([".."] * 75)
+    create_file('main.c', '''
+      #include <stdio.h>
+      #include <errno.h>
+      #include <sys/stat.h>
+
+      int main() {
+        struct stat path_stat;
+      '''
+        f'return stat("{path}", &path_stat);'
```
sbc100: You can start the multiline string with `f'''` and then use `{{` and `}}` to open and close the main function.

Or you can just do normal string interpolation with `% path at the end?   I think either one is nicer than combining 3 strings like this.

test/test_other.py
```diff
@@ -16084,3 +16084,19 @@ def test_em_js_bool_macro_expansion(self):
       }
     ''')
     self.do_runf('main.c')
+
+  def test_stat_many_dotdot(self):
```
sbc100: Can we try adding `@with_all_fs` here?

Also, are there other fs/stat tests that this one can be placed alongside in the this file?

hoodmane: There is one other stat test, I'll put it next to it.

hoodmane:
Okay updated it.

test/test_other.py
```diff
@@ -6009,6 +6009,21 @@ def test_symlink_silly(self):
     self.assertContained(r'ok', self.run_js('a.out.js', args=['123', 'abc']))
     self.assertContained(r'Failed to symlink paths: abc, ; errno=44', self.run_js('a.out.js', args=['abc', '']))
 
+  def test_stat_many_dotdot(self):
```
sbc100, line 6074: How about `@with_all_fs`?

hoodmane: Oops sorry missed that suggestion.

hoodmane:
Now let's see if I can convince `rebaseline_tests` to work correctly. I reran `./emsdk install tot` and then `tools/maint/rebaseline_tests.py` but the size diffs looked way too big for me, so I tried `./emcc --clear-cache` and am running again. I get a couple of closure tests tests failing with roughly this error:
```
test_codesize_minimal_O1 (test_other.other) ... FAIL
Running codesize test: mem_no_argv.c: ['-O3', '-sSTANDALONE_WASM'] True False

building:ERROR: /tmp/emtest_y5rraz1l/emscripten_temp_2bt05snf/a.out.jso4.js:43:262: ERROR - [JSC_UNDEFINED_VARIABLE] variable assert is undeclared
  43| if(typeof module!="undefined"){module["exports"]=Module}quit_=(status,toThrow)=>{process.exitCode=status;throw toThrow}}else if(ENVIRONMENT_IS_SHELL){readBinary=f=>{if(typeof readbuffer=="function"){return new Uint8Array(readbuffer(f))}let data=read(f,"binary");assert(typeof data=="object");return data};readAsync=async f=>readBinary(f);globalThis.clearTimeout??=id=>{};// spidermonkey lacks setTimeout but we use it above in readAsync.
                                                                                                                                                                                                                                                                            ^^^^^^

1 error(s), 0 warning(s)
```

hoodmane:
The jssizes have all gone up by ~600 bytes and the gzsizes by ~1300 bytes. Seems too big for this change.

sbc100:
If you want I can try to generate the codesize change for this PR?

hoodmane:
That would be helpful. It would be good to eventually figure out how to generate these myself. Maybe we could run the rebasline_tests.py in a docker image like @kleisauke suggested.

sbc100:
Done

hoodmane:
All the tests passed but something went wrong with the status report for test-wasm-js and GitHub shows it as in progress while circle ci shows it as passed. I guess I can push an empty commit and retrigger.

PR #24593, opened 2025-06-17
Looking up ip for localhost should return 127.0.0.1

hoodmane:
Causing a failure in the Python test suite.


sbc100:
I think I tried to make this same patch once and ran into issue.. I can't remember exactly where now..

PR #24594, opened 2025-06-17
Optimise code size of addFunction. NFC

RReverser:
I think there is more that could be done here - in particular, `setWasmTableEntry` is only used from `libaddfunction` and likely could be merged into it to unify data structures and updates - but for now just tackled some low-hanging fruits.


src/lib/libaddfunction.js
```diff
@@ -49,49 +48,36 @@ addToLibrary({
     return type;
   },
 #endif
-  $generateFuncType__deps: ['$uleb128Encode'],
-  $generateFuncType: (sig, target) => {
-    var sigRet = sig.slice(0, 1);
-    var sigParam = sig.slice(1);
-    var typeCodes = {
-      'i': 0x7f, // i32
+  $wasmTypeCodes: {
```
sbc100: Can you tag all of these as `__internal: true`?

src/lib/libaddfunction.js
```diff
@@ -49,49 +48,36 @@ addToLibrary({
...
-    // Parameters, length + signatures
-    target.push(0x60 /* form: func */);
-    uleb128Encode(sigParam.length, target);
-    for (var paramType of sigParam) {
+  $generateTypePack__deps: ['$uleb128EncodeWithLen', '$wasmTypeCodes'],
+  $generateTypePack: (types) => uleb128EncodeWithLen(Array.from(types, type => {
+    var code = wasmTypeCodes[type];
 #if ASSERTIONS
-      assert(paramType in typeCodes, `invalid signature char: ${paramType}`);
+  assert(code, `invalid signature char: ${type}`);
```
sbc100: Indentation looks wrong here.

src/lib/libaddfunction.js
```diff
@@ -49,49 +48,36 @@ addToLibrary({
...
+    'd': 0x7c, // f64
+    'e': 0x6f, // externref
+  },
 
-    // Parameters, length + signatures
-    target.push(0x60 /* form: func */);
-    uleb128Encode(sigParam.length, target);
-    for (var paramType of sigParam) {
+  $generateTypePack__deps: ['$uleb128EncodeWithLen', '$wasmTypeCodes'],
+  $generateTypePack: (types) => uleb128EncodeWithLen(Array.from(types, type => {
```
sbc100: Out style always puts braces around that arguments to lambda functions

src/lib/libaddfunction.js
```diff
@@ -158,17 +142,19 @@ addToLibrary({
     if (freeTableIndexes.length) {
       return freeTableIndexes.pop();
     }
+#if ASSERTIONS
     // Grow the table
```
sbc100: Move this comment down to line 149?

src/lib/libexports.js
```diff
@@ -13,11 +13,8 @@ addToLibrary({
     if (name[0] == '_') name = name.slice(1);
     var exportedFunc = wasmExports[name];
     if (exportedFunc) {
-      // Record the created function pointer to each function object,
-      // so that if the same function pointer is obtained several times,
-      // the same address will be returned.
-      exportedFunc.ptr ||= addFunction(exportedFunc);
-      return exportedFunc.ptr;
+      // Note: addFunction automatically caches the created function pointer.
+      return addFunction(exportedFunc);
```
sbc100, line 17: Nice

sbc100 (approved):

RReverser:
Huh. Looks like in core2 Closure mangles

```js
$wasmTypeCodes: {
  'i': 0x7f, // i32
#if MEMORY64
  'p': 0x7e, // i64
#else
  'p': 0x7f, // i32
#endif
  'j': 0x7e, // i64
  'f': 0x7d, // f32
  'd': 0x7c, // f64
  'e': 0x6f, // externref
},
```

into

```js
wb = {qg: 127, p: 127, rg: 126, f: 125, d: 124, e: 111},
```

I thought quoting was supposed to prevent that...

RReverser:
Ahh it's because of how jsify generates the variables. It generates them as unquoted when using object literal...

sbc100:
> Ahh it's because of how jsify generates the variables. It generates them as unquoted when using object literal...

Yes, sadly I think the round-tripping that happens when the JS library is processed will remove the quotes.

You can either:
(1) move the object declaration back inside the functions body;
(2) Put the whole thing in a multi-line string.

PR #24596, opened 2025-06-17
Use global TLS variable in proxying.c. NFC

sbc100:
There was some suspicion that this might lead to a code size change but it doesn't seem to (I suppose because bool/int doesn't have any code in its constructor).

Also use `bool` rather than `int` to better convey intent.

Split out from #24565

### DIFF
--- a/system/lib/pthread/proxying.c
+++ b/system/lib/pthread/proxying.c
@@ -106,6 +106,8 @@ static em_task_queue* get_or_add_tasks_for_thread(em_proxying_queue* q,
   return tasks;
 }
 
+static _Thread_local bool executing_system_queue = false;
+
 void emscripten_proxy_execute_queue(em_proxying_queue* q) {
   assert(q != NULL);
   assert(pthread_self());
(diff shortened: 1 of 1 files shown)


tlively (commented):
Are you planning to update the rest of the file to use bools? It seems best to be consistent one way or the other.

sbc100:
There is already existing use of stdbool, for example:

`.owned = false,  `

And:

`assert(false && "emscripten_proxy_sync failed");`

I don't see many more opportunities to use `bool` without changing the public interfaces, do you?

sbc100:
I found one more :)

tlively:
Lines 141 and 410 are the other candidates I see.

sbc100:
> Lines 141 and 410 are the other candidates I see.

This don't work so well as they are basically underlying parts of public interfaces that return `int`.

system/lib/pthread/proxying.c
```diff
@@ -607,7 +608,7 @@ double _emscripten_run_on_main_thread_js(int func_index,
                                          void* em_asm_addr,
                                          int num_args,
                                          double* buffer,
-                                         int sync) {
+                                         bool sync) {
```
tlively, line 611: Do callers need to be updated, I wonder?

sbc100: Its only ever called from JS with either 0 or 1

PR #24599, opened 2025-06-17
Make `wasmOffsetConvert` usage more localized in libstack_trace.js. NFC

sbc100:
I noticed this while investigating #24587.


src/lib/libstack_trace.js
```diff
@@ -101,29 +101,29 @@ var LibraryStackTrace = {
...
+      // Older versions of v8 give function index and offset in the function,
       // so we try using the offset converter. If that doesn't work,
       // we pack index and offset into a "return address"
+#if !USE_OFFSET_CONVERTER
+      abort('Legacy backtrace format detected but -sUSE_OFFSET_CONVERTER not present.')
+#else
+#if ASSERTIONS
+      assert(wasmOffsetConverter, 'wasmOffsetConverter global not found');
+#endif
       return wasmOffsetConverter.convert(+match[1], +match[2]);
```
kripken, line 119:
```suggestion
      return wasmOffsetConverter.convert(+match[1], +match[2]);
#endif
```
May as well only emit it when it is enabled.

kripken (approved):

src/lib/libstack_trace.js
sbc100: That is the case already... this code is already part of the #else block right?

kripken: Oh sorry, I missed the if for ASSERTIONS... lgtm as is

Issue #13652, opened 2021-03-12
Use OffscreenCanvas instead of `--proxy-to-worker`


janpaul123:
I was trying to figure out how to get WORKERFS to work so I can read from `File` objects directly, and for that I needed my SDL application to run in a web worker. In order to do that I came across `--proxy-to-worker`, but ran into its limitations (e.g. it crashes on `createVertexArray` because [`getExtension('OES_vertex_array_object')` is not implemented](https://github.com/kripken/webgl-worker/blob/e64b0c6910cf8bb96b03517deab2f1f0f9377fbe/webGLWorker.js#L534)).

I was then wondering if `--proxy-to-worker` should be deprecated altogether, since we can now render to a WebGL OffscreenCanvas in a WebWorker directly. I'm still trying to figure out how to do this exactly, so maybe it's implemented already? In that case it might be good to update the documentation accordingly (and perhaps add a note to `--proxy-to-worker` that you shouldn't use it, and what the alternative is?). Thanks!!

sbc100:
I think `--proxy-to-worker` (AKA `PROXY_TO_WORKER`) has largely been replaced by `PROXY_TO_PTHREAD` these days.

We do still have support for it though and there is still some testing.   The last time this came up I think perhaps @juj still had some valid use cases for it?

janpaul123:
Yeah, I spent some more time trying to figure out `PROXY_TO_PTHREAD`, and it looks like it hasn't been fully implemented for SDL yet, see e.g. https://github.com/emscripten-ports/SDL2/pull/127 which is still open. So I don't think we can tell people to switch over from `--proxy-to-worker` quite yet.

It still might be worth adding a note in the docs at `--proxy-to-worker`, that `PROXY_TO_PTHREAD` might be preferable whenever possible, even if `--proxy-to-worker` doesn't get outright deprecated (if that's the direction things are going in).

juj:
>  The last time this came up I think perhaps @juj still had some valid use cases for it?

I have actually never used --proxy-to-worker for anything. When I authored --proxy-to-pthread, I thought that would obsolete --proxy-to-worker, and on several occassions I have proposed for --proxy-to-worker to be removed, though @kripken has held on to it iiuc as he has dealt with users that have used --proxy-to-worker.

kripken:
`proxy-to-worker` also does not require pthreads. It just proxies with `postMessage` basically, and it can proxy a single-threaded app, which can be useful.

[...]

PR #24603, opened 2025-06-18
Remove references to `mozRequestPointerLock` and friends. NFC

sbc100:
The official `requestPointerLock` API is supported in all the browsers that we support.

See https://developer.mozilla.org/en-US/docs/Web/API/Element/requestPointerLock

### DIFF
--- a/src/lib/libhtml5.js
+++ b/src/lib/libhtml5.js
@@ -1543,7 +1543,7 @@ var LibraryHTML5 = {
 
   $fillPointerlockChangeEventData__deps: ['$JSEvents', '$stringToUTF8'],
   $fillPointerlockChangeEventData: (eventStruct) => {
-    var pointerLockElement = document.pointerLockElement || document.mozPointerLockElement || document.webkitPointerLockElement || document.msPointerLockElement;
+    var pointerLockElement = document.pointerLockElement;
     var isPointerlocked = !!pointerLockElement;
 #if !SAFE_HEAP
     // Assigning a boolean to HEAP32 with expected type coercion.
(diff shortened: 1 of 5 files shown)


src/lib/libhtml5.js
```diff
@@ -1595,10 +1595,9 @@ var LibraryHTML5 = {
     '$specialHTMLTargets'
 #endif
   ],
-  emscripten_set_pointerlockchange_callback_on_thread__docs: '/** @suppress {missingProperties} */', // Closure does not see document.body.mozRequestPointerLock etc.
   emscripten_set_pointerlockchange_callback_on_thread: (target, userData, useCapture, callbackfunc, targetThread) => {
     // TODO: Currently not supported in pthreads or in --proxy-to-worker mode. (In pthreads mode, document object is not defined)
-    if (!document || !document.body || (!document.body.requestPointerLock && !document.body.mozRequestPointerLock && !document.body.webkitRequestPointerLock && !document.body.msRequestPointerLock)) {
+    if (!document || !document.body?.requestPointerLock) {
```
RReverser: `!document || !document.body?.requestPointerLock` is logically equivalent to `!(document && document.body?.requestPointerLock)`, and in that form it's easy to see it's equivalent to `!document?.body?.requestPointerLock`

RReverser: Although I'm not sure why these conditions are guarding against `document` in the first place, if it doesn't exist, this check would fail anyway with ReferenceError.

sbc100: True, and also all these functions are marked as `__proxy: sync` so document itself should always exist.

sbc100: Fixed

RReverser (approved):

RReverser:
In case you decide to go on a rampage, this regex should help find other potentially outdated fallbacks: https://github.com/search?q=repo%3Aemscripten-core%2Femscripten+%2F%28%3F-i%29%28%5Cb%28moz%7Cwebkit%7Cms%29%5BA-Z%5D%7C%5C-%28moz%7Cwebkit%7Cms%7Co%29-%29%2F+&type=code

src/lib/libhtml5.js
RReverser: > all these functions are marked as `__proxy: sync` so document itself should always exist.

Does that apply to `--proxy-to-worker` too? (do we even support that mode still?)

sbc100: No, in that mode we provide a fake/dummy document object in the worker.

And yes we do still test this mode (i.e. it is supported, even though I don't know who would find it useful).

RReverser: > And yes we do still test this mode (i.e. it is supported, even though I don't know who would find it useful).

Oh yeah I know we still test it I just remembered comment threads like https://github.com/emscripten-core/emscripten/issues/13652#issuecomment-797761826 and not sure if it's "actively/officially" supported going forward.

That's outside of this issue though.

sbc100:
I tested locally with `browser.test_html5_core` and `interactive.test_glfw_pointerlock`

PR #24604, opened 2025-06-18
[llvm-libc] Import the already imported part of llvm-libc from llvm main

google-yfyang:
Importing llvm-libc from main instead of llvm-20.

Add some basic file filtering and patch applying logic in the import script `system/lib/update_llvm_libc.py`.

Tested:
The same number of tests failing at HEAD when running `./test/runner llvmlibc`. 
#24493


system/lib/update_llvm_libc.py
```diff
@@ -0,0 +1,119 @@
...
+    print("STDERR:", e.stderr)
+
+def main():
+  if len(sys.argv) > 1:
+    llvm_dir = os.path.abspath(sys.argv[1])
+  else:
+    llvm_dir = default_llvm_dir
+  libcxx_dir = os.path.join(llvm_dir, 'libcxx')
+
+  # libcxx includes headers from LLVM's libc
```
sbc100: No need for this comment anymore

system/lib/update_llvm_libc.py
```diff
@@ -0,0 +1,119 @@
...
+        command,
+        capture_output=True,
+        text=True,
+        check=True
+    )
+
+  except subprocess.CalledProcessError as e:
+    print(f"Error applying patch: {e}")
+    print("STDOUT:", e.stdout)
+    print("STDERR:", e.stderr)
```
sbc100: If you don't capture the output is the error message not clear enough on its own?   I.e. maybe jsut remove the whole try catch and the `capture_output` arg?

system/lib/update_llvm_libc.py
```diff
@@ -0,0 +1,119 @@
...
+import os
+import sys
+import shutil
+import glob
+import subprocess
+
+script_dir = os.path.abspath(os.path.dirname(__file__))
+emscripten_root = os.path.dirname(os.path.dirname(script_dir))
+default_llvm_dir = os.path.join(os.path.dirname(emscripten_root), 'llvm-project')
+emscripten_patches = os.path.join(script_dir, "llvm-libc", "patches", "*.patch")
```
sbc100: We don't tend to check in patches like this, at least not for the other libraries here.  Instead we push our changes out to https://github.com/emscripten-core/llvm-project and maintain the patches in git.

I'm not totally opposed to this approach here, but it would be new thing for us.

sbc100 (commented):
Nice work!

Lets see if all the tests pass (i.e. lets see if libc++ is effected by this change).

Also, can you remove the llvm-libc stuff from system/lib/update_libcxx.py

sbc100 (commented):
Also, can you update the readme file in the llvm-libc directory?

system/lib/update_llvm_libc.py
google-yfyang: I suppose those patches are not submitted to the upstream?

If I understand correctly, you want me to submit the content of this patch to emscripten's fork of llvm?

system/lib/update_llvm_libc.py
google-yfyang: Done.

google-yfyang:
> Also, can you update the readme file in the llvm-libc directory?

Done.



> Nice work!
> 
> Lets see if all the tests pass (i.e. lets see if libc++ is effected by this change).
> 
> Also, can you remove the llvm-libc stuff from system/lib/update_libcxx.py

Tests are failing due to some differences in the expected code size. Can you give me a pointer on how those are updated?

system/lib/update_llvm_libc.py
```diff
@@ -0,0 +1,96 @@
...
+    if os.path.isdir(full):
+      shutil.copytree(full, os.path.join(local_dir, f))
+    elif f not in excludes:
+      shutil.copy2(full, os.path.join(local_dir, f))
+  for root, dirs, files in os.walk(local_dir):
+    for f in files:
+      if f in excludes:
+        full = os.path.join(root, f)
+        os.remove(full)
+
```
sbc100, line 68: Two blank lines between functions.

system/lib/update_llvm_libc.py
```diff
@@ -0,0 +1,96 @@
...
+def main():
+  if len(sys.argv) > 1:
+    llvm_dir = os.path.abspath(sys.argv[1])
+  else:
+    llvm_dir = default_llvm_dir
+  libcxx_dir = os.path.join(llvm_dir, 'libcxx')
```
sbc100: Unused?

Issue #24570, opened 2025-06-14
Pthread task queue deadlock race condition problem (with pthread_cancel()?)


juj:
There is a race condition problem that results in a deadlock, which seems to take its way somewhere back to the `em_proxying_queue` / `em_task_queue` rewrite of the proxying mechanism that @tlively landed in #15737.

It can be observed on the CI as a flaky `browser.test_pthread_cancel` test, for example by adding a Python script that repeatedly reruns the test. Running a short ad hoc python test script that repeats `test/runner browser.test_pthread_cancel` a hundred times, I get 19/100 failures, or 19.0%.

When the race condition occurs, the browser will hang with the main thread deadlocked into the following callstack:

```
Script terminated by timeout at:
_emscripten_get_now
test.wasm.futex_wait_main_browser_thread
test.wasm.emscripten_futex_wait
test.wasm.__timedwait_cp
test.wasm.__timedwait
test.wasm.__pthread_mutex_timedlock
test.wasm.__pthread_mutex_lock // 6
test.wasm.emscripten_proxy_execute_queue
test.wasm.emscripten_current_thread_process_queued_calls
test.wasm.emscripten_main_thread_process_queued_calls // 5
test.wasm._emscripten_yield
test.wasm.emscripten_futex_wait
test.wasm.__timedwait_cp
test.wasm.__timedwait
test.wasm.__pthread_mutex_timedlock
test.wasm.__pthread_mutex_lock // 4
test.wasm.emscripten_builtin_malloc // 3
test.wasm.em_task_queue_create  // 2
test.wasm.get_or_add_tasks_for_thread
test.wasm.do_proxy
test.wasm.emscripten_proxy_async
test.wasm.pthread_kill
test.wasm.pthread_cancel        // 1
test.wasm.__original_main
test.wasm.main
callMain
doRun
run/<
```

I.e.
1. main thread has decided to `pthread_cancel()` the pthread in the test.
2. nested inside `pthread_cancel()`, main thread decides to allocate a new proxying queue for the pthread.
3. it calls `malloc()`
4. that takes the malloc lock
5. inside malloc lock, the main thread decides to run operations proxied to it
6. running proxied operations involves taking a lock again, which never returns - the main thread is hung.

I do not understand exactly why the main thread is hanging on attempting to acquire the lock at step 6.

Two hypotheses come to mind:
[...]

PR #24607, opened 2025-06-18
Add pthread mutex deadlock detection for debug builds

sbc100:
This would really have helped in at least detecting the deadlock that was found in #24570 / #24565

### DIFF
--- a/system/lib/libc/musl/src/thread/pthread_mutex_trylock.c
+++ b/system/lib/libc/musl/src/thread/pthread_mutex_trylock.c
@@ -55,7 +55,7 @@ success:
 
 #if defined(__EMSCRIPTEN__) || !defined(NDEBUG)
 	// We can get here for normal mutexes too, but only in debug builds
-	// (where we track ownership purely for debug purposed).
+	// (where we track ownership purely for debug purposes).
 	if ((type & 15) == PTHREAD_MUTEX_NORMAL) return 0;
 #endif
 


system/lib/libc/musl/src/thread/pthread_mutex_lock.c
```diff
@@ -2,9 +2,12 @@
 
 int __pthread_mutex_lock(pthread_mutex_t *m)
 {
+#if !defined(__EMSCRIPTEN__) || defined(NDEBUG)
+	/* XXX EMSCRIPTEN always take the slow path in debug builds so we can trap rather than deadlock */
 	if ((m->_m_type&15) == PTHREAD_MUTEX_NORMAL
 	    && !a_cas(&m->_m_lock, 0, EBUSY))
 		return 0;
+#endif
```
kripken, line 10: Should upstream do the same for all platforms? I'm just trying to understand if this is a general improvement or if it works around a specific problem in wasm.

sbc100: I think its a general improvement for debug builds yes.  I guess most libc's don't ship debug builds like we do (especially automatically selecting debug libs based on command line flags, which emcc does is quite unusual).

sbc100: Upstream musl doesn't really have any/many assertions AFAICT, so its not something they tend to consider I suppose?

sbc100: Indeed, it looks like upstream musl only has assertion in 2 very specific places:

```
src/regex/regcomp.c
src/malloc/mallocng/..
```

No other assertions in whole of libc.

kripken: I see, thanks. Makes sense.

system/lib/libc/musl/src/thread/pthread_mutex_timedlock.c
```diff
@@ -79,6 +86,9 @@ int __pthread_mutex_timedlock(pthread_mutex_t *restrict m, const struct timespec
 		int own = r & 0x3fffffff;
 		if (!own && (!r || (type&4)))
 			continue;
+#if defined(__EMSCRIPTEN__) && !defined(NDEBUG)
+		assert(own != __pthread_self()->tid && "pthread mutex deadlock detected");
+#endif
 		if ((type&3) == PTHREAD_MUTEX_ERRORCHECK
```
kripken, line 89: Can we perhaps enable this `PTHREAD_MUTEX_ERRORCHECK` globally somehow, rather than add these three lines? (It wouldn't assert though, just return EDEADLK)

system/lib/pthread/library_pthread.c
```diff
@@ -157,6 +153,11 @@ void _emscripten_init_main_thread(void) {
   __main_pthread.next = __main_pthread.prev = &__main_pthread;
   __main_pthread.tsd = (void **)__pthread_tsd_main;
 
+  _emscripten_init_main_thread_js(&__main_pthread);
+
+  __main_pthread.stack = (void*)emscripten_stack_get_base();
+  __main_pthread.stack_size = emscripten_stack_get_base() - emscripten_stack_get_end();
```
kripken, line 159: Shouldn't the stack setup be as early as possible? Why move it down?

system/lib/libc/musl/src/thread/pthread_mutex_lock.c
RReverser: Would that change with llvm-libc or is it the same in regards to assertions?

system/lib/libc/musl/src/thread/pthread_mutex_timedlock.c
sbc100: `PTHREAD_MUTEX_ERRORCHECK` gives different semantics to the user, which is something a little different to just asserting. 

I don't think most users will be checking the return code from pthread_mutex_lock, at least not unless they are explicitly using `PTHREAD_MUTEX_ERRORCHECK`.. so I don't think it would be as useful and could lead to some really broken code that think it holds a lock when it doesn't.

system/lib/pthread/library_pthread.c
sbc100: `_emscripten_init_main_thread_js` is what sets up the stack regions, so it has to happen before `emscripten_stack_get_base` can work.

`_emscripten_init_main_thread_js` can end up calling `_emscripten_thread_profiler_init` which can call `malloc` which can take a lock which requires the `.tid` field to be set, so it has to happen after the `__main_pthread.tid` assignment.

system/lib/libc/musl/src/thread/pthread_mutex_lock.c
sbc100: @RReverser, I'm not sure I understand the question.   If we switch to llvm-libc fully we would likely end up using their version of the pthread API implementation.     This means we would likely need to make similar change there, unless they already have debug asserts like this.

RReverser: I was referring to "Upstream musl doesn't really have any/many assertions AFAICT" in general, the question was whether llvm-libc tends to do assertions for debug builds more than musl does.

sbc100: Ah... no idea I'm afraid.   I will find out.

test/other/test_pthread_mutex_deadlock.c
```diff
@@ -0,0 +1,20 @@
+#include <assert.h>
+#include <pthread.h>
+#include <stdbool.h>
+#include <stdio.h>
+
+pthread_mutex_t m = PTHREAD_MUTEX_INITIALIZER;
+
+int main() {
+  printf("in main\n");
```
kripken, line 9:
```suggestion
  puts("in main");
```

test/other/test_pthread_mutex_deadlock.c
```diff
@@ -0,0 +1,20 @@
...
+  int rtn = pthread_mutex_lock(&m);
+  assert(rtn == 0);
+
+  // Attempt to lock a second time.  In debug builds this should
+  // hit an assertion.  In release builds this will deadlock and
+  // never return.
+  pthread_mutex_lock(&m);
+  printf("should never get here\n");
```
kripken, line 17:
```suggestion
  puts("should never get here");
```

kripken (approved):
(nits optional)

brendandahl:
I tried out this change and it does cause an assert in `test_pthread_cancel.c`, though that test is only run in `-O3` currently.

test/other/test_pthread_mutex_deadlock.c
sbc100: We normally don't shy away from using printf for this kind of thing in test code.

system/lib/libc/musl/src/thread/pthread_mutex_lock.c
sbc100: Apparently they (llvm libc folks) are open to adding this kind of asserts.

RReverser: That sounds promising, being able to upstream these things makes your argument about the switch from musl to llvm-libc even more compelling.

system/lib/libc/musl/src/thread/pthread_mutex_trylock.c
```diff
@@ -53,6 +53,12 @@ int __pthread_mutex_trylock_owner(pthread_mutex_t *m)
 	}
 #endif
 
+#if defined(__EMSCRIPTEN__) || !defined(NDEBUG)
+	// We can get here for normal mutexes too, but only in debug builds
+	// (where we track ownership purely for debug purposed).
```
kripken:
```suggestion
	// (where we track ownership purely for debug purposes).
```

system/lib/libc/musl/src/thread/pthread_mutex_trylock.c
```diff
@@ -53,6 +53,12 @@ int __pthread_mutex_trylock_owner(pthread_mutex_t *m)
...
+	if ((type & 15) == PTHREAD_MUTEX_NORMAL) return 0;
+#endif
```
kripken, line 60: How can we get here? Is it because of one of the changes in this PR?

sbc100: Yes, prior to this change normal mutex's never make it here since they don't need to track ownership.

PR #24610, opened 2025-06-19
[embind] Allow Closure to compress registerType methods

RReverser:
The reason 'fromWireType', 'toWireType' and 'readValueFromPointer' were always protected from Closure is because we dynamically generate JS functions, and property mangling would make it impossible for the dynamically generated code to access properties declared statically.

However, we can easily work around that by binding corresponding methods outside of the generated lambdas, and passing just the methods instead of the whole type object.

### DIFF
--- a/src/lib/libembind.js
+++ b/src/lib/libembind.js
@@ -17,13 +17,13 @@ var LibraryEmbind = {
   $EmValType__deps: ['_emval_decref', '$Emval', '$readPointer'],
   $EmValType: `{
     name: 'emscripten::val',
-    'fromWireType': (handle) => {
+    fromWireType: (handle) => {
       var rv = Emval.toValue(handle);
       __emval_decref(handle);
       return rv;
     },
-    'toWireType': (destructors, value) => Emval.toHandle(value),
-    'readValueFromPointer': readPointer,
+    toWireType: (destructors, value) => Emval.toHandle(value),
+    readValueFromPointer: readPointer,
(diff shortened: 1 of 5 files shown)


src/lib/libembind.js
```diff
@@ -17,13 +17,13 @@ var LibraryEmbind = {
   $EmValType__deps: ['_emval_decref', '$Emval', '$readPointer'],
   $EmValType: `{
     name: 'emscripten::val',
-    'fromWireType': (handle) => {
+    fromWireType: (handle) => {
```
sbc100, line 20: Maybe convert to method syntax?

sbc100 (commented):
Great!

Are you sure these property names are not required as part of the public/external API?

RReverser:
> Are you sure these property names are not required as part of the public/external API?

Fairly sure, as they are only needed for type registration, but I'll wait for @brendandahl to review as well.

src/lib/libembind.js
RReverser: There is quite a lot more that could be done here, yeah, but knowing how it went in the past, I'm really trying not to get nerd-sniped and not mix too many unrelated changes into the same PR (like you asked :P).

sbc100: fair enough!

RReverser:
Actually, yeah, I rechecked, this JavaScript API is not part of our public API anywhere.

PR #24611, opened 2025-06-19
[embind] Reuse signature codes from em_asm

RReverser:
Aside from a minor discrepancy, there is no point in maintaining two separate sets of type pack to signature conversions, especially now that they've grown with wasm64 conditionals. Keeping them in one place reduces complexity in bind.h and ensures the signatures stay in sync.

In the process, I also removed the custom SignatureTranslator, since what we really want here is a signature of a function with (...argWireTypes) => retWireType shape, and we can get that via BindingType<T>::WireType directly instead of custom template mappings.


system/include/emscripten/bind.h
```diff
@@ -565,97 +565,35 @@ struct FunctorInvoker<ReturnPolicy, FunctorType, void, Args...> {
...
-template<>
-struct SignatureCode<float> {
-    static constexpr char get() {
-        return 'f';
-    }
-};
+template<typename T>
+struct SignatureCode<T, decltype(__em_asm_sig<T>::value)> : __em_asm_sig<T> {};
 
+// TODO: should we add this override to em_asm?
```
RReverser, line 576: @sbc100 Would love your thoughts on this one. In other places we represent size_t aka unsigned long as number, even though it's 32 bit on wasm32 and 64 on wasm64, so it's lossy on wasm64.

Should we change EM_ASM behaviour to match that, or should we change those other places to prevent precision loss for unsigned long, or is it expected that they handle it differently?

RReverser: Hm maybe "most" is an exaggeration. Embind itself also encodes size_t aka unsigned long as a BigInt on wasm64 and number otherwise.

Shame that you can't distinguish between size_t and "plain" longs.

RReverser:
@sbc100 @brendandahl Ping.

system/include/emscripten/bind.h
sbc100: (Without looking at this specific instance yet) I think the only times we want to use i53 numbers for 64-bit values is:

1. When we know that thing is a pointer type (e.g. when `p` is used in signature of a JS function)
2. When there is an explicit opt-in (i.e. `__i53abi` tag on JS functions).

Otherwise 64-bit values should be preserved either via bigint (or pair-of-numbers)

sbc100:
Sorry, I've been away on vacation.

RReverser:
I know, all good. Just pinged to get some attention now that you're back :)

system/include/emscripten/bind.h
RReverser: Makes sense, but in that case sounds like we should instead remove this override from Embind, as it will currently use i53 for any `unsigned long`.

Does that sound right / should I do that in the same PR?

sbc100: If this PR is NFC than lets keep that way,.. is it?

RReverser: For now it should be, yeah. I just wasn't sure if I'm going to add functional change here as well.

brendandahl: Changing the behavior in another PR to be consistent sounds good to me. FWIW, I thought size_t was already a bigint on memory64, but I'm guessing we left it int32 to avoid users having to deal with BigInt.

sbc100: We handle size_t like we handle pointers yes.  They have both have pointer-like behaviour in that they have a different size on wasm32 and wasm64.

Other types such as uint64 have the same size on both wasm32 and wasm64.

RReverser: > They have both have pointer-like behaviour in that they have a different size on wasm32 and wasm64.

But for pointers we translate them to int53 aka plain numbers in JS code, including in Embind, whereas size_t is different, as it will become bigint on wasm64.

RReverser:
Ok changed the title to NFC as per discussion, please review as-is.

system/include/emscripten/bind.h
sbc100: This is, I suppose, a limitation of type system.   I assume its not possible to distinguish between `unsigned long` and `size_t`?   If it is possible we should treat size_t like we do pointers.

RReverser: > This is, I suppose, a limitation of type system. I assume its not possible to distinguish between `unsigned long` and `size_t`?

Indeed. That's what I said above too:

> Shame that you can't distinguish between size_t and "plain" longs.

Issue #24540, opened 2025-06-10
EMBIND_AOT should AOT method callers


RReverser:
We have `emval_methodCallers` mechanism used for `foo()` and `foo.call<...>(...)` on `emscripten::val` instances here: https://github.com/emscripten-core/emscripten/blob/2fe9aeb84fb2b3f4462ecd6a8d7f156c06349312/src/lib/libemval.js#L284-L352

Right now it either constructs a JS function from strings on the fly, or uses a slower mechanism with looking up all the types on each invocation.

As far as I can tell, it should be fairly straightforward to add it to the `EMBIND_AOT` processing so that specialized function is generated during compile-time and then reused. cc @brendandahl

RReverser:
Although perhaps "straightforward" is an exaggeration since it can't be easily converted to a top-level static and at the moment method callers are registered lazily...

brendandahl:
Yeah, discovering which method callers are used is the hard part. One option that would be a bigger change is we could make a "typed" version of `val`, where you'd define a C++ interface to a JS object. The project [typescripten](https://github.com/think-cell/typescripten) does something similar to this, but behind the scenes it is still built on on `val`. I imagine we could change it to also build some static initializers that register all the invokers.

Another option, require users to register the signatures they want to use with `call` or `operator()` in `EMSCRIPTEN_BINDINGS`. Something like 
```cpp
EMSCRIPTEN_BINDINGS(invokers) {
  register_invoker<SomeReturnType, ArgType1, ArgType2>();
}
```

RReverser:
> Another option, require users to register the signatures they want to use with `call` or `operator()` in `EMSCRIPTEN_BINDINGS`. Something like

Yeah I really don't like that option, feels like exposing implementation details.

I wonder if we can use linker shenanigans here somehow? Like, have each template link to an external C++ method caller and then find them during embind_gen? The mangling scheme is pretty well-established so it should be possible to recover precise C++ types.

RReverser:
Or I guess you could even compute typeid-based signature in a static constexpr and store it in a custom Wasm section.

brendandahl:
> Or I guess you could even compute typeid-based signature in a static constexpr and store it in a custom Wasm section.

I tried to do something like this for another part of the bindings code, but `constexpr` isn't guaranteed to be computed at compile time. We'd probably need `consteval`.

RReverser:
[...]

PR #24619, opened 2025-06-24
[emval] Simplify dynamic invoker creation. NFC

RReverser:


RReverser:
I fixed CI, this should be ready for review (I'll just need to rebaseline again before merging).

src/lib/libemval.js
```diff
@@ -297,52 +297,41 @@ var LibraryEmVal = {
...
-    functionBody +=
-      `  var rv = ${invoker}(${argsList.join(', ')});\n`;
+    functionBody += `(${args})`;
     if (!retType.isVoid) {
-      params.push('toReturnWire', 'emval_returnValue');
-      args.push(toReturnWire, emval_returnValue);
-      functionBody +=
-        '  return emval_returnValue(toReturnWire, destructorsRef, rv);\n';
+      captures['toReturnWire'] = toReturnWire;
+      captures['emval_returnValue'] = emval_returnValue;
```
sbc100, line 325: Would it be slightly more compact to just list all the captures in the object literal above?   i.e. is there any harm in capturing stuff that isn't used?

RReverser: For now, just tiny memory overhead I guess, but when we implement AOT as shown in https://github.com/emscripten-core/emscripten/issues/24540, this will also result in larger method caller functions in the static JS.

I'd rather keep it.

brendandahl:
Looks like AOT mode is broken.

RReverser:
> Looks like AOT mode is broken.

Yeah which is weird, because I already fixed it - it was fine in https://github.com/emscripten-core/emscripten/commit/049e9ccc34698d49299e48423219fe9f796423c3. I think I force-pushed something wrong from the other laptop. Will fix.

PR #24628, opened 2025-06-27
[NFC][Testing] Test Poppler with optimizations

kripken:
Poppler's configure script chooses `-O2` for gcc, and nothing for all other
compilers, so we were never optimizing source files. Fix that by adding
optimizations in cflags.

This is important for test coverage, but also for branch hints, as LLVM
does not emit hints when not optimizing. Another fix for branch hints
is applied here: the likely macro was tailored (once again) to gcc. (This
is NFC for now, but if/when LLVM gets branch hint support, this will
become a useful real-world test of compiling and linking real-world code
with such hints.)

This also makes the Poppler benchmark 33% faster.

### DIFF
--- a/test/third_party/poppler/goo/GooLikely.h
+++ b/test/third_party/poppler/goo/GooLikely.h
@@ -11,7 +11,9 @@
 #ifndef GOOLIKELY_H
 #define GOOLIKELY_H
 
-#if defined(__GNUC__) && (__GNUC__ > 2) && defined(__OPTIMIZE__)
+// XXX EMSCRIPTEN: Remove gcc-specific detection of __builtin_expect.
+// #if defined(__GNUC__) && (__GNUC__ > 2) && defined(__OPTIMIZE__)
+#ifdef __EMSCRIPTEN__
 # define likely(x)      __builtin_expect((x), 1)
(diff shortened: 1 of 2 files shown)


dschuff (approved):

test/third_party/poppler/goo/GooLikely.h
```diff
@@ -11,7 +11,9 @@
 #ifndef GOOLIKELY_H
 #define GOOLIKELY_H
 
-#if defined(__GNUC__) && (__GNUC__ > 2) && defined(__OPTIMIZE__)
+// XXX EMSCRIPTEN: Remove gcc-specific detection of __builtin_expect.
+// #if defined(__GNUC__) && (__GNUC__ > 2) && defined(__OPTIMIZE__)
+#ifdef __EMSCRIPTEN__
```
sbc100, line 16: This should not be needed since emcc / clang defined `__GNUC__` to 4 and define `__OPTIMIZE__` in `-O2` and above.

kripken: Thanks, I think this was needed before the first part of this PR, which started to actually optimize... then I didn't notice this had become redundant. I'll remove it.

PR #24632, opened 2025-06-28
Fix test_env and test_environ to expect any locale.

juj:
Fixes test to pass on different locales on Windows.

### DIFF
--- a/test/core/test_environ.out
+++ b/test/core/test_environ.out
@@ -3,5 +3,5 @@ LOGNAME=web_user
 PATH=/
 PWD=/
 HOME=/home/web_user
-LANG=(C|en_US).UTF-8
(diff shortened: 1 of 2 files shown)


test/core/test_environ.out
```diff
@@ -3,5 +3,5 @@ LOGNAME=web_user
 PATH=/
 PWD=/
 HOME=/home/web_user
-LANG=(C|en_US).UTF-8
```
sbc100, line 6: The main point of these tests is to verify that setting en_US in outer environment is respected by emscripten.

If the `@with_env_modify({'LANG': 'en_US.UTF-8'})` line in the test code doesn't work on windows we should probably look into why its not works and/or disable the test under windows (with a bug open to investigate why it doesn't work).

PR #24634, opened 2025-06-29
Adds SDL3_image port

fredheidrich:


tools/ports/sdl3_image.py
```diff
@@ -0,0 +1,136 @@
...
+    libname += '-' + formats
+  if settings.PTHREADS:
+    libname += '-mt'
+  if settings.SUPPORT_LONGJMP == 'wasm':
+    libname += '-wasm-sjlj'
+  return libname + '.a'
+
+
+def get(ports, settings, shared):
+  ports.fetch_project('sdl3_image', f'https://github.com/libsdl-org/SDL_image/archive/{TAG}.zip', sha512hash=HASH)
```
kripken, line 56: Comparing to the sdl2_image port, I see this in the diff:
```diff
-  ports.fetch_project('sdl2_image', f'https://github.com/libsdl-org/SDL_image/archive/refs/tags/{TAG}.zip', sha512hash=HASH)
+  ports.fetch_project('sdl3_image', f'https://github.com/libsdl-org/SDL_image/archive/{TAG}.zip', sha512hash=HASH)
```
Is `/ref/tags/` not needed?

test/test_browser.py
```diff
@@ -2992,6 +2992,48 @@ def test_sdl2_image_formats(self):
       '--use-port=sdl2', '--use-port=sdl2_image:formats=jpg',
     ])
 
+  @requires_graphics_hardware
+  @also_with_wasm2js
```
kripken:
```suggestion
```
Perhaps we don't need wasm2js here, as there isn't something very specific to test in SDL3+wasm2js (vs SDL2+wasm2js).

test/test_browser.py
```diff
@@ -2992,6 +2992,48 @@ def test_sdl2_image_formats(self):
...
+    shutil.copy(test_file('screenshot.jpg'), 'screenshot.jpeg')
+    self.btest_exit('test_sdl3_image.c', 600, cflags=[
+      '--preload-file', 'screenshot.jpeg',
+      '-DSCREENSHOT_DIRNAME="/"', '-DSCREENSHOT_BASENAME="screenshot.jpeg"',
+      '-sUSE_SDL=3', '-sUSE_SDL_IMAGE=3', '--use-preload-plugins',
+    ])
+
+  @also_with_wasmfs
+  @requires_graphics_hardware
+  @with_all_sjlj
```
kripken:
```suggestion
  @requires_graphics_hardware
```
As above, removing some testing I think is less crucial.

kripken (commented):
Thanks! Looks generally good.

tools/ports/sdl3_image.py
fredheidrich: I looks like we can use either one. Both seems to alias the exact same file from GitHub.

kripken: Interesting about github aliases, sounds good.

kripken (approved):
lgtm, but is this still marked as draft because you have more work to do?

kripken:
Code size test failures should be fixed by merging in latest `main`.

The documentation test error should be fixable by the command mentioned in the error, `Please run tools/maint/update_settings_docs.py`

kripken:
Oh, please add a changelog mention here too, as in the other PR.

Thanks, it's great to get these added!

fredheidrich:
> lgtm, but is this still marked as draft because you have more work to do?

That's right, wanted to fix the failing doctest first -- will do that asap, thank you :)

test/browser/test_sdl3_image.c
```diff
@@ -0,0 +1,78 @@
...
+  chdir(SCREENSHOT_DIRNAME);
+  result = testImage(renderer, "./" SCREENSHOT_BASENAME); // relative path
+  assert(result != 0);
+
+  SDL_RenderPresent(renderer);
+
+  printf("you should see an image.\n");
+
+  return result;
+}
```
sbc100, line 78: How does this file differ from `test_sdl2_image.c`?  If not much, it might be worth using the same file?

fredheidrich: Well, the API has changed slightly from 2->3. It certainly possibly to pass the differences from the test suite but that would likely make the code a bit confusing to read (imho).

sbc100: Can you explain the differences? Or maybe just attach a diff of `test_sdl2_image.c` and `test_sdl3_image.c`?

PR #24638, opened 2025-06-30
Use local server for test_fetch_redirect. NFC

dipidoo:
fixes redirect test to use localhost


test/fetch/test_fetch_redirect.c
```diff
@@ -9,46 +9,93 @@
...
   attr.onsuccess = onsuccess;
   attr.onreadystatechange = onreadystatechange;
-  emscripten_fetch_t *fetch = emscripten_fetch(&attr, "https://httpbin.org/status/307");
+  emscripten_fetch_t *fetch = emscripten_fetch(&attr, url);
   assert(fetch);
+}
+
+int main() {
+  async_code_idx = 0;
+  async_method_idx = 0;
```
sbc100: These two lines are not needed since they are already initialized to zero above.

test/common.py
```diff
@@ -2148,6 +2148,25 @@ def do_POST(self):
...
+          self.send_header('Location', '/status/200')
+          self.end_headers()
+        elif code == 200:
+          self.send_response(200)
+          self.send_header('Content-type', 'text/plain')
+          self.end_headers()
+          self.wfile.write(b'OK')
+        else:
+          self.send_error(code)
+        return
```
sbc100: I don't think the `return` statement is needed here is it?

test/common.py
```diff
@@ -2160,6 +2179,25 @@ def do_GET(self):
         self.send_header('Content-type', 'text/html')
         self.end_headers()
         self.wfile.write(read_binary(test_file('browser_harness.html')))
+      elif info.path.startswith('/status/'):
+        code_str = info.path[len('/status/'):]
+        try:
+          code = int(code_str)
+        except ValueError:
+          self.send_error(400, 'Invalid status code')
+          return
```
sbc100: I don't think the try/catch is needed here is it?  If `code_str` is not a number maybe better to simple crash here rather than try to recover/continue.

dipidoo:
@sbc100 thank you for the review. is the branch recent enough or should I update the branch with latest main?

PR #24641, opened 2025-07-01
Add support for setting socket's blocking mode through ioctl with FIO…

inolen:
…NBIO

Today I was recompiling quakejs (12 years later, better late than never :)) against the latest emscripten. Everything was smooth sailing, but I hit one issue because ioctl() used to be under-implemented, and now it's more fleshed out but missing support for FIONBIO (https://github.com/torvalds/linux/blob/66701750d5565c574af42bef0b789ce0203e3071/fs/ioctl.c#L341).

I don't have the tests working locally right now, so I'm crossing my fingers that some CI job kicks in here to help verify. I modified `test_sockets_echo_server` which had support for both Winsock / POSIX sockets, and shimmed it to use ioctl() on POSIX to help test this.

Edit: Of course, this doesn't actually change or implement any support for O_NONBLOCK inside of libsockfs, it's just enabling the flag to be set just as it would be through fcntl.

### DIFF
--- a/src/lib/libsockfs.js
+++ b/src/lib/libsockfs.js
@@ -421,6 +421,14 @@ addToLibrary({
             }
             {{{ makeSetValue('arg', '0', 'bytes', 'i32') }}};
             return 0;
+          case {{{ cDefs.FIONBIO }}}:
+            var on = {{{ makeGetValue('arg', '0', 'i32') }}};
+            if (on) {
+              sock.stream.flags |= {{{ cDefs.O_NONBLOCK }}};
+            } else {
+              sock.stream.flags &= ~{{{ cDefs.O_NONBLOCK }}};
+            }
+            return 0;
           default:
             return {{{ cDefs.EINVAL }}};
         }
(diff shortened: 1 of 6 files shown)


inolen:
Any advice on these test failures? I spliced in the changes to add FIONBIO into struct_info.json, because gen_struct_info.py was throwing an error for me. Not sure if I missed something, or if they're unrelated.

kripken:
>  (12 years later, better late than never :))

Hi @inolen ! :smile: Been a while, indeed...

Those test errors may be unrelated. Updating to latest `main` might fix them, I'm pressing the "Update branch" button now. If they don't go away, I'll take a closer look.

kripken (approved):

kripken:
How does quakejs run now, btw?

inolen:
> How does quakejs run now, btw?

Seemingly works fine, do you have anything in particular you're curious about?

kripken:
I guess I was just wondering if performance was any different? (hopefully better...)

Issue #24648, opened 2025-07-02
OPFS leaves AccessHandles open when my app crashes


rickg-hcl:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.11-git (1e68a1b78e68bec44796a80617f18d29e7d52255)
clang version 21.0.0git (https:/github.com/llvm/llvm-project 2f05451198e2f222ec66cec4892ada0509519290)
Target: wasm32-unknown-emscripten
Thread model: posix

I have spent 2 days trying to create a simple test program the reproduces this, but I could not get it to reproduce.

In our very complicated emscripten application, when it crashes, there are access handles left dangling, such that when we restart, the opens fail with:
```
Failed to execute 'createSyncAccessHandle' on 'FileSystemFileHandle': Access Handles cannot be created if there is another open Access Handle or Writable stream associated with the same file.
```

reloading the web page does not resolve the issue, we have close the chrome browser to recover.

I will be creating a pull request that adds a function
```
void wasmfs_close_opfs_backend(backend_t backend);
```
That I have tested and solves our problem.

kripken:
If you build with `-sEXIT_RUNTIME`, WasmFS should shut down fully as the app exits, and then it might close all handles - but I'm not sure if the backend does that. Worth trying though.

cc @tlively

tlively:
Since this is a problem caused by exclusive locking, it would also be solved by #21869.

It looks like closing a file does cause the OPFS backend to close the file handle and release the lock, but I don't recall there being any shutdown routine that goes through and manually closes open files. I'm surprised the locks persist through a page reload, but that would be a question for the OPFS spec or implementation, not for Emscripten.

rickg-hcl:
@tlively That's the point of this issue.  We only have a problem when the app crashes, when files are not closed.  The PR I created just adds a function that I can call during Module.onAbort to clean up all the open files.  It should be harmless for anyone else that doesn't use it.
When we implemented this independently, we proactively closed any file handle that wasn't used for a few seconds to overcome this issue.

kripken:
If the app crashes, is `onAbort` guaranteed to be called? I am surprised that is enough. (edit: i mean that a crash can be of the entire tab at the browser level, etc.)

Perhaps there is some OPFS API that we can use at the opposite time, during startup, to get all open file handles and release them?

kripken:
[...]

PR #24649, opened 2025-07-02
add opfs_close

rickg-hcl:
https://github.com/emscripten-core/emscripten/issues/24648

add a function to close all access handles


kripken:
I am not opposed to landing this necessarily, but my feeling is that we should wait on adding this API. If we figure out the problem, or we get more user requests for it, we can reconsider then.

PR #24655, opened 2025-07-07
[compiler-rt] Avoid using musl's syscall layer. NFC

kleisauke:
Use libc calls instead, similar to the NetBSD and macOS implementations.


system/lib/compiler-rt/lib/sanitizer_common/sanitizer_emscripten.cpp
```diff
@@ -128,6 +136,92 @@ u64 MonotonicNanoTime() {
...
+void internal_usleep(u64 useconds) {
+  usleep(useconds);
+}
+
+void internal__exit(int exitcode) {
+  __wasi_proc_exit(exitcode);
+}
+
+tid_t GetTid() {
+  return gettid();
```
kleisauke, line 206: Note: this was previously using `pthread_self()`:
https://github.com/emscripten-core/emscripten/blob/b0c461c10482e12f88e192e04b3640acb1577f64/system/lib/compiler-rt/lib/sanitizer_common/sanitizer_linux.cpp#L683-L684

After commit 77859cd7e125efe23689505f4c181df81738e46d, we can safely use `gettid()` here.

system/lib/compiler-rt/lib/sanitizer_common/sanitizer_emscripten.cpp
```diff
@@ -128,6 +136,92 @@ u64 MonotonicNanoTime() {
...
+  }
+  return num;
+}
+
+uptr internal_write(fd_t fd, const void *buf, uptr count) {
+  __wasi_ciovec_t iov = { (const uint8_t*)buf, count };
+  size_t num;
+  if (__wasi_syscall_ret(__wasi_fd_write(fd, &iov, 1, &num))) {
+    return -1;
+  }
```
sbc100: If calling libc functions is OK in these `internal_xx` function why not just call libc `write` here?

system/lib/compiler-rt/lib/sanitizer_common/sanitizer_emscripten.cpp
```diff
@@ -128,6 +136,92 @@ u64 MonotonicNanoTime() {
 
 void GetMemoryProfile(fill_profile_f cb, uptr *stats) {}
 
+int internal_madvise(uptr addr, uptr length, int advice) {
+  return 0; // madvise is currently ignored
+}
+
+uptr internal_close(fd_t fd) {
+  return __wasi_fd_close(fd);
```
sbc100: Seems odd to use the lower lever `__wasi_fd_close` but the higher level `open` for `internal_open`.

system/lib/compiler-rt/lib/sanitizer_common/sanitizer_platform_interceptors.h
```diff
@@ -308,7 +308,7 @@ SANITIZER_WEAK_IMPORT void *aligned_alloc(__sanitizer::usize __alignment,
 #define SANITIZER_INTERCEPT_SENDMMSG SI_LINUX
 #define SANITIZER_INTERCEPT_SYSMSG SI_LINUX_NOT_ANDROID
 #define SANITIZER_INTERCEPT_GETPEERNAME SI_POSIX
-#define SANITIZER_INTERCEPT_IOCTL SI_POSIX && !SI_EMSCRIPTEN
+#define SANITIZER_INTERCEPT_IOCTL SI_POSIX_NOT_EMSCRIPTEN
```
sbc100, line 311: Why do we have a special define for this but not, for example, for  `SI_POSIX && !SI_SOLARIS` above?
`

kleisauke: Good question, I'm not  sure. I attempted to unify these checks somewhat, since we use `SI_POSIX_NOT_EMSCRIPTEN` elsewhere in this file, but I'm happy to split it out if you'd prefer.

sbc100: Oh. I see, its pre-existing.  LGTM then, although I'm not sure why we added this since its only used in one place right now.

system/lib/compiler-rt/lib/sanitizer_common/sanitizer_emscripten.cpp
kleisauke: Good question. I left it this way to ensure this change is still NFC, just like the previous implementation:
https://github.com/emscripten-core/emscripten/blob/b0c461c10482e12f88e192e04b3640acb1577f64/system/lib/compiler-rt/lib/sanitizer_common/sanitizer_linux.cpp#L344-L351

Looking at:
https://github.com/emscripten-core/emscripten/blob/b0c461c10482e12f88e192e04b3640acb1577f64/system/lib/libc/musl/src/unistd/write.c#L6-L16

It looks like doing `write()` here has the same effect.

system/lib/compiler-rt/lib/sanitizer_common/sanitizer_emscripten.cpp
kleisauke: Indeed. I left it this way to ensure this change is still NFC, just like the previous implementation:
https://github.com/emscripten-core/emscripten/blob/b0c461c10482e12f88e192e04b3640acb1577f64/system/lib/compiler-rt/lib/sanitizer_common/sanitizer_linux.cpp#L303-L305

Looking at:
https://github.com/emscripten-core/emscripten/blob/b0c461c10482e12f88e192e04b3640acb1577f64/system/lib/libc/musl/src/unistd/close.c#L16-L20

It sounds like doing `close()` here is more appropriate (as it wraps the `errno` with `__wasi_syscall_ret()`).

sbc100: Keeping the pre-existing behaviour (for now) seems reasonable. We can do cleanups / improvements separately and a followups.

kleisauke:
Looking at the test failures, I think we need to land PR #24652 first.

system/lib/compiler-rt/lib/sanitizer_common/sanitizer_emscripten.cpp
kleisauke: The above comment also applies to `internal_read()`, see:
https://github.com/emscripten-core/emscripten/blob/b0c461c10482e12f88e192e04b3640acb1577f64/system/lib/libc/musl/src/unistd/read.c#L6-L16

And `internal__exit()`:
https://github.com/emscripten-core/emscripten/blob/b0c461c10482e12f88e192e04b3640acb1577f64/system/lib/libc/musl/src/exit/_Exit.c#L6-L8

Commit ba629f83af758602a870e716f22d531174d36370 should fix this.

system/lib/compiler-rt/lib/sanitizer_common/sanitizer_emscripten.cpp
kleisauke: It looks like the callers of `internal_close()` aren't checking its return code anyway, so using `close()` here should be safe. Commit ba629f83af758602a870e716f22d531174d36370 addresses this.

PR #24666, opened 2025-07-08
Add support for PTHREAD_MUTEX_ROBUST

sbc100:
This was mostly just a case of copying the robust list handling code from the musl version of pthread_create.


system/lib/pthread/pthread_create.c
```diff
@@ -312,10 +312,29 @@ void _emscripten_thread_exit(void* result) {
   // Call into the musl function that runs destructors of all thread-specific data.
   __pthread_tsd_run_dtors();
 
-  if (!--libc.threads_minus_1) libc.need_locks = 0;
```
kripken, line 315: Was this not intentionally before `__tl_lock` so that it could potentially skip locking there?

kripken: But I suppose it doesn't matter much in `pthread_create`...

sbc100: This happens inside of `__tl_lock` in the musl version of this function.   This is just an attempt to keep the order of things consistent between the two.

PR #24672, opened 2025-07-09
Use JSON snapshots for all code size tests. NFC

RReverser:
Right now, code size tests in the codebase use two different formats - either everything in a single JSON file, or multiple single-line files.

I found the discrepancy a bit annoying and wanted to unify them.

Between the two, the one with individual files tends to be harder to review as Github (as well as local git tools) add quite a lot of boilerplate UI around each file, you need to manually scan filenames to understand which size diffs are related to each other, and overall feels more difficult to scan quickly as a human.

In this PR I'm unifying both towards a single JSON file format, and other supplementary metadata (`imports`, `sent` etc) goes into the same JSON as well so you can see all relevant context together during reviews.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -9248,7 +9248,7 @@ int main() {
 
     self.check_output_sizes('hello_world.js', 'hello_world.wasm', 'no_asserts.js', 'no_asserts.wasm', 'strict.js', 'strict.wasm')
 
-  def run_codesize_test(self, filename, cflags, check_funcs=True, check_full_js=False):
+  def run_codesize_test(self, filename, cflags, check_funcs=True, check_full_js=False, skip_gz=False):
     # in -Os, -Oz, we remove imports wasm doesn't need
     print('Running codesize test: %s:' % filename, cflags, check_funcs, check_full_js)
     filename = test_file('other/codesize', filename)
@@ -9319,7 +9319,7 @@ int main() {
         funcs.sort()
         info['funcs'] = [strip_numeric_suffixes(f) for f in funcs]
 
-    self.check_output_sizes(*outputs, **info)
+    self.check_output_sizes(*outputs, metadata=info, skip_gz=skip_gz)
 
   @parameterized({
     'O0': ([], True),
@@ -9385,14 +9385,16 @@ int main() {
     # we don't metadce with linkable code! other modules may want stuff
     # TODO(sbc): Investivate why the number of exports is order of magnitude
     # larger for wasm backend.
-    'dylink_all': (['-O3', '-sMAIN_MODULE'],),
+    # This test seems to produce different results under gzip on macOS and Windows machines
+    # so skip the gzip size reporting here.
+    'dylink_all': (['-O3', '-sMAIN_MODULE'], {'skip_gz': True}),
     'dylink': (['-O3', '-sMAIN_MODULE=2'],),
     # WasmFS should not be fully linked into a hello world program.
     'wasmfs': (['-O3', '-sWASMFS'],),
-    'single_file': (['-O3', '-sSINGLE_FILE'],), # noqa
+    'single_file': (['-O3', '-sSINGLE_FILE'],),
   })
-  def test_codesize_hello(self, args):
-    self.run_codesize_test('hello_world.c', args)
+  def test_codesize_hello(self, args, kwargs={}): # noqa
+    self.run_codesize_test('hello_world.c', args, **kwargs)
 
   @parameterized({
     'O3':                 ('mem.c', ['-O3']),
@@ -11904,7 +11906,7 @@ int main () {
 
     self.check_output_sizes(*outputs)
 
-  def check_output_sizes(self, *outputs: str, **metadata):
+  def check_output_sizes(self, *outputs: str, metadata=None, skip_gz=False):
     test_name = self.id().split('.')[-1]
     results_file = test_file('code_size', test_name + '.json')
 
@@ -11942,18 +11944,21 @@ int main () {
       total_output_size += size
       total_expected_size += expected_size
 
-      f_gz = f + '.gz'
-      size_gz = len(gzip.compress(contents))
-      expected_size_gz = expected_results.get(f_gz, inf)
-      update_and_print_diff(f_gz, size_gz, expected_size_gz)
-      total_output_size_gz += size_gz
-      total_expected_size_gz += expected_size_gz
+      if not skip_gz:
+        f_gz = f + '.gz'
+        size_gz = len(gzip.compress(contents))
+        expected_size_gz = expected_results.get(f_gz, inf)
+        update_and_print_diff(f_gz, size_gz, expected_size_gz)
+        total_output_size_gz += size_gz
+        total_expected_size_gz += expected_size_gz
 
     if len(outputs) > 1:
       update_and_print_diff('total', total_output_size, total_expected_size)
-      update_and_print_diff('total_gz', total_output_size_gz, total_expected_size_gz)
+      if not skip_gz:
+        update_and_print_diff('total_gz', total_output_size_gz, total_expected_size_gz)
 
-    obtained_results.update(metadata)
+    if metadata:
+      obtained_results.update(metadata)
 
     obtained_results_json = json.dumps(obtained_results, indent=2)
     expected_results_json = json.dumps(expected_results, indent=2)
@@ -11967,7 +11972,7 @@ int main () {
       if total_output_size < total_expected_size:
         print(f'Hey amazing, overall generated code size was improved by {total_expected_size - total_output_size} bytes!')
         print('If this is expected, rerun the test with --rebaseline to update the expected sizes')
-      self.assertTextDataIdentical(obtained_results_json, expected_results_json)
+      self.assertTextDataIdentical(expected_results_json, obtained_results_json)
 
   # Tests the library_c_preprocessor.js functionality.
   @crossplatform
--- a/test/code_size/test_codesize_hello_dylink_all.json
+++ b/test/code_size/test_codesize_hello_dylink_all.json
@@ -1,10 +1,7 @@
 {
   "a.out.js": 246743,
-  "a.out.js.gz": 81085,
   "a.out.nodebug.wasm": 597771,
-  "a.out.nodebug.wasm.gz": 330177,
   "total": 844514,
-  "total_gz": 411262,
   "sent": [
     "IMG_Init",
     "IMG_Load",


test/test_other.py
```diff
@@ -16300,3 +16204,10 @@ def test_install(self):
     # Some files, such as as maintenance tools should not be part of the
     # install.
     self.assertNotExists('newdir/tools/maint/')
+
+  def test_small_js_flags(self):
+    self.btest('browser_test_hello_world.c', '0', cflags=['-O3', '--closure=1', '-sINCOMING_MODULE_JS_API=[]', '-sENVIRONMENT=web', '--output-eol=linux'])
```
sbc100: I don't think you can move this test to test_other.py like this.   The `best` method is designed to run a browser test.. its not supposed to work outside of the browser tests.   I'm not even sure how this runs at all here.. the test is build with `-sENVIRONMENT=web` which means it should really run on node at all.

test/test_other.py
```diff
@@ -12435,10 +12341,8 @@ def test(args):
     print('sizes', normal, changed)
     # Changing this option to [] should decrease code size.
     self.assertLess(changed, normal)
-    # Check an absolute code size as well, with some slack.
-    self.check_expected_size_in_file('js',
-                                     test_file('other/test_INCOMING_MODULE_JS_API.js.size'),
-                                     changed)
+    # Check an absolute code size as well.
+    self.checkOutputSizes(['a.out.js'])
```
sbc100: We use `snake_case` for all our internal helper methods (even though the `assertFoo` family of functions uses camel.

test/test_other.py
```diff
@@ -12426,7 +12332,7 @@ def test_main_reads_params(self):
 
   def test_INCOMING_MODULE_JS_API(self):
     def test(args):
-      self.run_process([EMCC, test_file('hello_world.c'), '-O3', '--closure=1', '-sENVIRONMENT=node,shell', '--output-eol=linux'] + args)
+      self.run_process([EMCC, test_file('hello_world.c'), '-O3', '--closure=1', '-sENVIRONMENT=web', '--output-eol=linux'] + args)
```
sbc100: Maybe make this change separately since I imagine it will effect the code size numbers and it would be nice if this PR did not change them.

sbc100 (commented):
Thanks for doing this!  I've been thinking about doing this myself for a while now.

I really like the idea, especially for the code sizes themselves.  I'm less sure about the symbol lists since the extra syntax of quotes and commas makes them harder to read than flat files.

 Would you consider using json only for the file sizes and continue to use flat files for symbol lists?  Or maybe its not worth it?

test/test_other.py
```diff
@@ -11949,50 +11898,7 @@ def print_percent(actual, expected):
...
-      total_output_size += size
-      total_expected_size += expected_size
-
-      total_output_size_gz += size_gz
-      total_expected_size_gz += expected_size_gz
-
-    obtained_results['total'] = total_output_size
-    obtained_results['total_gz'] = total_output_size_gz
-    print('Total output size=' + str(total_output_size) + ' bytes, expected total size=' + str(total_expected_size) + ', delta=' + str(total_output_size - total_expected_size) + print_percent(total_output_size, total_expected_size))
-    print('Total output size gzipped=' + str(total_output_size_gz) + ' bytes, expected total size gzipped=' + str(total_expected_size_gz) + ', delta=' + str(total_output_size_gz - total_expected_size_gz) + print_percent(total_output_size_gz, total_expected_size_gz))
```
sbc100, line 11993: This output delta information expresses and a percent might still be useful, did we loose it?

test/test_other.py
RReverser: Uhh something went seriously wrong here. I was actually just removing it as per description in https://github.com/emscripten-core/emscripten/pull/24672/commits/66a74a3cda57e48e797bfbd16c6481e8b042a9ec since it essentially duplicates an existing test.

Let me fix that up...

test/test_other.py
RReverser: Nope, it's still there, just in the common.py.

RReverser:
> Would you consider using json only for the file sizes and continue to use flat files for symbol lists? Or maybe its not worth it?

I could, but I find it kind of nice to have in the same context for the same reason it's nice to have different sizes in the same context - eg there were times when code size increased because some symbol was accidentally exported that shouldn't have been, and then it's helpful for debugging the increase.

I guess we could output YAML / TOML / custom text output if the quotes and commas are the concern?

sbc100:
> > Would you consider using json only for the file sizes and continue to use flat files for symbol lists? Or maybe its not worth it?
> 
> I could, but I find it kind of nice to have in the same context for the same reason it's nice to have different sizes in the same context - eg there were times when code size increased because some symbol was accidentally exported that shouldn't have been, and then it's helpful for debugging the increase.
> 
> I guess we could output YAML / TOML / custom text output if the quotes and commas are the concern?

No this seems fine to me if its useful to have it all together.

test/test_other.py
RReverser: > even though the `assertFoo` family of functions uses camel

😭 why can't we be consistent. Okay...

test/test_other.py
RReverser: So this change was ported over from the removed `test_small_js_flags` in `test_browser.py`.

These tests were identical, except for `ENVIRONMENT`, so it made sense to merge them into one, and usually we care about code sizes for `ENVIRONMENT=web` and not shells anyway so I chose to keep this option when merging.

sbc100: I see... Are you sure the browser test is no longer useful?  One difference is that the browser test actually runs the resulting program, but maybe that is not important here?

test/common.py
```diff
@@ -1694,14 +1697,76 @@ def assertBinaryEqual(self, file1, file2):
     self.assertEqual(read_binary(file1),
                      read_binary(file2))
 
-  def check_expected_size_in_file(self, desc, filename, size):
+  def check_output_sizes(self, outputs: List[str], metadata=None):
+    test_name = self.id().split('.')[-1]
```
sbc100: I guess this makes a lot of sense to set the filename automatically like this, but it would break if we ever had two different test suites with the same test name... i guess that is very unlikely and we would notice right away.

test/test_other.py
sbc100: The problem is that the python unit test library goes against the python PEP8 coding standards (which we are using in general).

sbc100: From AI:

```
Adhering to PEP 8 in unittest:

While unittest itself uses camelCase for its built-in methods (like assertEqual and setUp), which predates the strong emphasis on snake_case in PEP 8 for functions and methods, your custom test methods and any helper methods within your test classes should still adhere to PEP 8's snake_case convention for readability and consistency within your own
```

test/test_other.py
RReverser: > but maybe that is not important here?

Yeah, exactly - this test is literally hello world and we run it (and much more complex tests) plenty of times throughout the test suite. The only thing it was used for is code size snapshot, but then we don't need to run it in the browser to get that.

test/common.py
RReverser: FWIW that's how existing logic already worked in `run_codesize_test`, I just copied it here.

RReverser:
(I'll do final rebaseline to fix CI before merging)

test/test_other.py
RReverser: I reverted this test merge to keep it simple. The browser test is still moved to test_other.py, but stays separate.

RReverser:
@sbc100 Anything left on this one? Could be nice to land before eg #24679 as it would make diff much more compact :)

RReverser:
~~Oh damn I just accidentally added unnecessary files, fixing up...~~ Done.

sbc100 (approved):

RReverser:
@sbc100 I think the CI check expectations script doesn't correctly handle removed/added tests, as it tries to find the same test both before and after this PR and fails.

What is the solution in such cases? Manual merge?

sbc100:
> @sbc100 I think the CI check expectations script doesn't correctly handle removed/added tests, as it tries to find the same test both before and after this PR and fails.
> 
> What is the solution in such cases? Manual merge?

Fair enough we can bypass that check for this PR.

RReverser:
> Fair enough we can bypass that check for this PR.

Can you please do it? I don't have the bypass rights :(

RReverser:
I rebaselined & fixed conflicts again, but it might need yet another rebaseline right before you merge it.

RReverser:
Ping.

sbc100:
lgtm!

Do we need another rebase or can we just land this now?

sbc100:
Oh, we do need another rebase (I see conflicts).

RReverser:
> Oh, we do need another rebase (I see conflicts).

Yeah see above:

> Can you please do it? I don't have the bypass rights :(

> I rebaselined & fixed conflicts again, but it might need yet another rebaseline right before you merge it.

I basically have to keep rebaselining, but I don't have rights to merge it, so unless you merge right after my rebaseline, it just goes in a loop 😅

I can do yet another rebaseline and hope you're free to merge soon, or can I leave it to you to run it locally right before merging - which is easier?

sbc100:
Can you do one more rebase now.  If it doesn't auto-merge, I'll take a look again before EOD

RReverser:
I think it still won't auto-merge because of

> I think the CI check expectations script doesn't correctly handle removed/added tests, as it tries to find the same test both before and after this PR and fails.

but yeah, I'll do one more.

RReverser:
Hm I accidentally re-added deleted files during rebase. Trying again.

RReverser:
Done.

RReverser:
I rebased but I see [test-mac-arm64 CI](https://app.circleci.com/pipelines/github/emscripten-core/emscripten/43867/workflows/e8a294b4-9e23-43ae-a557-723109f35139/jobs/974710) reporting a lot of

```
node:internal/child_process:420
    throw new ErrnoException(err, 'spawn');
    ^

Error: spawn Unknown system error -86
    at ChildProcess.spawn (node:internal/child_process:420:11)
    at spawn (node:child_process:753:9)
    at Compiler.run (/Users/distiller/project/node_modules/google-closure-compiler/lib/node/closure-compiler.js:76:26)
    at Object.<anonymous> (/Users/distiller/project/node_modules/google-closure-compiler/cli.js:83:10)
    at Module._compile (node:internal/modules/cjs/loader:1730:14)
    at Object..js (node:internal/modules/cjs/loader:1895:10)
    at Module.load (node:internal/modules/cjs/loader:1465:32)
    at Function._load (node:internal/modules/cjs/loader:1282:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24) {
  errno: -86,
  code: 'Unknown system error -86',
  syscall: 'spawn'
}

Node.js v22.16.0
building:WARNING: falling back to java version of closure compiler
node:internal/child_process:420
    throw new ErrnoException(err, 'spawn');
    ^
```

That's not related to my changes, is it?

sbc100:
> I rebased but I see [test-mac-arm64 CI](https://app.circleci.com/pipelines/github/emscripten-core/emscripten/43867/workflows/e8a294b4-9e23-43ae-a557-723109f35139/jobs/974710) reporting a lot of
> 
> ```
> node:internal/child_process:420
>     throw new ErrnoException(err, 'spawn');
>     ^
> 
> Error: spawn Unknown system error -86
>     at ChildProcess.spawn (node:internal/child_process:420:11)
>     at spawn (node:child_process:753:9)
>     at Compiler.run (/Users/distiller/project/node_modules/google-closure-compiler/lib/node/closure-compiler.js:76:26)
>     at Object.<anonymous> (/Users/distiller/project/node_modules/google-closure-compiler/cli.js:83:10)
>     at Module._compile (node:internal/modules/cjs/loader:1730:14)
>     at Object..js (node:internal/modules/cjs/loader:1895:10)
>     at Module.load (node:internal/modules/cjs/loader:1465:32)
>     at Function._load (node:internal/modules/cjs/loader:1282:12)
>     at TracingChannel.traceSync (node:diagnostics_channel:322:14)
>     at wrapModuleLoad (node:internal/modules/cjs/loader:235:24) {
>   errno: -86,
>   code: 'Unknown system error -86',
>   syscall: 'spawn'
> }
> 
> Node.js v22.16.0
> building:WARNING: falling back to java version of closure compiler
> node:internal/child_process:420
>     throw new ErrnoException(err, 'spawn');
>     ^
> ```
> 
> That's not related to my changes, is it?

Are those actually causing errors?  I think that is due to the arm bot not being able to execute the x86 binary for closure-compiler (i.e. no rossetta installed?) its supposed to fall back to java I think.

sbc100:
The real error looks like its coming from test_codesize_hello_dylink_all.. I recently added that but was thinking maybe I should revert it since it will be very sensitive.

RReverser:
> The real error looks like its coming from test_codesize_hello_dylink_all.

Yeah I wasn't sure if that's related to different Closure binary or not.

I did `emsdk install tot` + `emcc --clear-cache` before rebaselining locally, not sure why else CI would have different sizes for these couple of tests.

sbc100:
test_codesize_hello_dylink_all is highly sensitive since it includes basically all symbols ..

RReverser:
So should I do anything else here?

sbc100:
The reason the test failure is hard to read is because is exceeding the limits of `self.assertEqual`.

Can you do this:

```
obtained_results = json.dumps(obtained_results, indent=2)
expected_results = json.dumps(expected_results, indent=2)
self.assertTextDataIdentical(obtained_results, expected_results)
```

Our `assertTextDataIdentical` is better at diffing large things and showing just what changed.

RReverser:
> The reason the test failure is hard to read is because is exceeding the limits of `self.assertEqual`.

FWIW it also has proper logs above where the test is actually running, but neither test runner nor CircleCI capture stdout so it's not shown in the same place as test failure:

```
a.out.js: size=246743, expected=246743
a.out.js.gz: size=81085, expected=81085
a.out.nodebug.wasm: size=597771, expected=597771
a.out.nodebug.wasm.gz: size=330359, expected=330181, delta=178 (+0.05%)
total: size=844514, expected=844514
total_gz: size=411444, expected=411266, delta=178 (+0.04%)
test_codesize_hello_dylink_all (test_other.other.test_codesize_hello_dylink_all) ... FAIL
```

https://app.circleci.com/pipelines/github/emscripten-core/emscripten/43867/workflows/e8a294b4-9e23-43ae-a557-723109f35139/jobs/974710/parallel-runs/0/steps/0-111?invite=true#step-111-97665_38

RReverser:
I'll go to sleep for now, will take a look tomorrow evening if this is not merged by then.

RReverser:
I can't keep rebaselining... Can you please help with landing this?

sbc100:
Sure, I'm trying.   We will get there eventually :)

One downside I just notices to having the symbol list all in the single JSON is grep-ability:

```
$ git grep __wasm_init_tls
src/lib/libdylink.js:      '__wasm_init_tls',
system/lib/pthread/emscripten_tls_init.c:extern void __wasm_init_tls(void *memory);
system/lib/pthread/emscripten_tls_init.c:  __wasm_init_tls(tls_block);
system/lib/wasm_worker/library_wasm_worker.c:void __wasm_init_tls(void *memory);
system/lib/wasm_worker/library_wasm_worker.c:  __wasm_init_tls((void*)*sbrk_ptr);
system/lib/wasm_worker/wasm_worker_initialize.S:  // __wasm_init_tls(stackLowestAddress);
system/lib/wasm_worker/wasm_worker_initialize.S:  .functype __wasm_init_tls (PTR) -> ()
system/lib/wasm_worker/wasm_worker_initialize.S:  call __wasm_init_tls
system/lib/wasm_worker/wasm_worker_initialize.S:  // N.b. The function __wasm_init_tls above does not need
system/lib/wasm_worker/wasm_worker_initialize.S:  // So we must initialize __stack_pointer only *after* completing __wasm_init_tls:
test/other/codesize/test_codesize_minimal_pthreads.funcs:$__wasm_init_tls
test/other/codesize/test_codesize_minimal_pthreads_memgrowth.funcs:$__wasm_init_tls
```

Here I can see that `__wasm_init_tls` is a declared function in those two code size tests.

After this change the symbol will appear in the json output but it won't be clear which list it is part of..   Not a huge deal I suppose.  

It also means I can't do a simple `wc -l` to count the symbols... again not a big deal.  I still think we should land this.

RReverser:
If it does inhibit your workflow, I don't mind reverting these lists, just let me know. I suppose you could use `jq` instead with current workflow, but might be an overkill.

RReverser:
> I did `emsdk install tot` + `emcc --clear-cache` before rebaselining locally, not sure why else CI would have different sizes for these couple of tests.

Ah, so it was another of cross-platform shenanigans...

sbc100:
> > I did `emsdk install tot` + `emcc --clear-cache` before rebaselining locally, not sure why else CI would have different sizes for these couple of tests.
> 
> Ah, so it was another of cross-platform shenanigans...

Yes, this seems to be the first occurrence of it showup up in our CI machines, not just reports of user machines with this gzip discrepancy.

Note that the test in question was not previously measuring the gzip size of the wasm file which was why this didn't show up until this change.

RReverser:
Interesting. I guess it makes sense that it would show up in the most sensitive test like you said.

RReverser:
@sbc100 Thank you!

PR #24677, opened 2025-07-09
Fix HTML5 battery API

sbc100:
The old `navigator.battery` API is not longer present in modern browsers which all use `navigator.getBattery()`.

Tested locally on my laptop.

See https://developer.mozilla.org/en-US/docs/Web/API/BatteryManager


sbc100:
Alternatively I guess we should simply remove/deprecate this API since firefox removed it?

src/lib/libhtml5.js
```diff
@@ -2134,24 +2135,37 @@ var LibraryHTML5 = {
...
   emscripten_get_battery_status: (batteryState) => {
-    if (!battery()) return {{{ cDefs.EMSCRIPTEN_RESULT_NOT_SUPPORTED }}};
-    fillBatteryEventData(batteryState, battery());
+    if (!hasBatteryAPI()) return {{{ cDefs.EMSCRIPTEN_RESULT_NOT_SUPPORTED }}};
+    if (!batteryManager) {
+      navigator.getBattery().then((b) => {
+        batteryManager = b;
+      });
+      return {{{ cDefs.EMSCRIPTEN_RESULT_NO_DATA }}};
+    }
```
kripken, line 2167: It looks like this API fails until the async part happens - that makes me wonder how useful this API is? Do you know of users?

sbc100: This is also true for the `emscripten_get_mouse_status` API, so there is precedent, and that is what the `EMSCRIPTEN_RESULT_NO_DATA` result code is for I believe.

kripken: I see, makes sense.

PR #24678, opened 2025-07-09
[embind] Don't truncate `unsigned long` under wasm64

RReverser:
This makes it consistent with other bindings like EM_ASM.

As part of this, I had to change vector and map sizes and indices to explicitly take `unsigned int` instead, so that users don't have to upgrade all their code to deal with unexpected bigints in e.g. `vec.size()`.

Technically this means you can't pass around a C++ vector >4GB to JS and back, but I'm sure we can live with it.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -38,8 +38,10 @@ See docs/process.md for more on how version tagging works.
   wrapped with `WebAssembly.Suspending` functions. To automatically wrap library
   functions for use with JSPI they must now explicitly set
   `myLibraryFunction__async: true`.
-- On wasm64 Embind will now pass `size_t` as a `bigint` instead of `number` to
-  match behaviour of other bindings like `EM_ASM`.
+- Removed special casing for `size_t` in Embind, since it was also inadvertently
+  affecting `unsigned long` on wasm64. Both will now match the behaviour of
+  other 64-bit integers on wasm64 and will be passed as `bigint` instead of
+  `number` to the JavaScript code. (#24678)
 
 4.0.10 - 06/07/25
 -----------------
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -38,6 +38,8 @@ See docs/process.md for more on how version tagging works.
   wrapped with `WebAssembly.Suspending` functions. To automatically wrap library
   functions for use with JSPI they must now explicitly set
   `myLibraryFunction__async: true`.
+- On wasm64 Embind will now pass `size_t` as a `bigint` instead of `number` to
+  match behaviour of other bindings like `EM_ASM`.
 
 4.0.10 - 06/07/25
 -----------------
--- a/system/include/emscripten/bind.h
+++ b/system/include/emscripten/bind.h
@@ -573,13 +573,6 @@ struct SignatureCode : __em_asm_sig<int> {};
 template<typename T>
 struct SignatureCode<T, decltype(__em_asm_sig<T>::value)> : __em_asm_sig<T> {};
 
-// TODO: should we add this override to em_asm?
-// Most places, including Embind, use `p` for `size_t` (aka `unsigned long`) but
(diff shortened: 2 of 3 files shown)


sbc100 (commented):
IIUC, this is the wrong direction.  `size_t` and `void *` should be the same type on the JS side whenever possible.

The fact that `EM_ASM` cannot distinguish size_t from `unsigned long` is a bug/limitation in EM_ASM.

No point dragging embind down to the level of EM_ASM.

RReverser:
> The fact that `EM_ASM` cannot distinguish size_t from `unsigned long` is a bug/limitation in EM_ASM.

But neither can Embind, as it's a limitation of C/C++ not specific bindings.

In #24611 you said:

> (Without looking at this specific instance yet) I think the only times we want to use i53 numbers for 64-bit values is:
> 
> 1. When we know that thing is a pointer type (e.g. when `p` is used in signature of a JS function)
> 2. When there is an explicit opt-in (i.e. `__i53abi` tag on JS functions).
> 
> Otherwise 64-bit values should be preserved either via bigint (or pair-of-numbers)

In this case `size_t` aka `unsigned long` is a 64-bit value, and 1) we don't know that it's a pointer type and 2) there is no explicit opt-in, so per conditions above we should preserve it as a bigint on wasm64?

RReverser:
To clarify, `SignatureCode<size_t>` is a misnomer there.

It can look like it's only overriding signature for an explicit `size_t`, but because it's simply an type alias, what that line actually does is defines `SignatureCode<unsigned long>`.

So right now all `unsigned long` values - which are 64-bit integers on wasm64 - get truncated to `i53` in Embind. Based on previous discussions, I assume this is not what we want, and `unsigned long` should be a `bigint` on wasm64 to preserve the full range instead.

sbc100:
> To clarify, `SignatureCode<size_t>` is a misnomer there.
> 
> It can look like it's only overriding signature for an explicit `size_t`, but because it's simply an type alias, what that line actually does is defines `SignatureCode<unsigned long>`.
> 
> So right now all `unsigned long` values - which are 64-bit integers on wasm64 - get truncated to `i53` in Embind. Based on previous discussions, I assume this is not what we want, and `unsigned long` should be a `bigint` on wasm64 to preserve the full range instead.

Oh I see, so that real change here that we want is for `unsigned long` to *not* be truncated.

Perhaps the title should "Don't truncate 64-bit unsigned long values under wasm64" and the unfortunate side effect is that size_t cannot be truncated (even though we would like it to be, ideally).

RReverser:
Sgtm, changed the title.

ChangeLog.md
```diff
@@ -38,6 +38,8 @@ See docs/process.md for more on how version tagging works.
   wrapped with `WebAssembly.Suspending` functions. To automatically wrap library
   functions for use with JSPI they must now explicitly set
   `myLibraryFunction__async: true`.
+- On wasm64 Embind will now pass `size_t` as a `bigint` instead of `number` to
+  match behaviour of other bindings like `EM_ASM`.
```
sbc100: Can you update the comment too.   Something like:  On wasm64 Embind will now pass 64-bit integers as bigint instead of number.   One downside to this is that size_t will now surface as bigint, unlike pointers themselves which are are converted to i15 numbers.

RReverser: > On wasm64 Embind will now pass 64-bit integers as bigint instead of number.

That's a bit overly generic statement. It was already passing `int64_t` aka `long long` and `uint64_t` aka `unsigned long long` - which are 64-bit integers - as bigint on all targets. 

On `wasm64` it was even passing `long` - which is 64-bit on that target - as bigint too.

It's only `size_t` aka `unsigned long` that had this special inconsistent override that produced number even on wasm64.

sbc100: Maybe something more like "Removed special case for size_t since it was also effecting all `unsigned long` .."

RReverser: Updated

PR #24679, opened 2025-07-09
Update closure-compiler. NFC

sbc100:


kripken (commented):
The code diffs look fine, but most of the size changes are small but worse. Maybe worth investigating one to see what changed?

sbc100:
[files.zip](https://github.com/user-attachments/files/21243105/files.zip)

sbc100:
I attached the JS generated for the test_minimal_runtime_code_size_hello_embind test.  The size differs by just one byte, but if I clang-format the files they don't differ at all.

Any idea how to find which is the differing byte here?

sbc100:
Ah, I found an interesting hunk in the code generated by test_codesize_mem_O3_standalone_narg_flto:

```
@@ -139,7 +139,7 @@
   if (b.instantiateWasm)
     return new Promise(d => {b.instantiateWasm(c, (f, e) => {d(a(f, e))})});
   z ??= b.locateFile ? b.locateFile("a.out.wasm", n) : n + "a.out.wasm";
-  return a((await C(c)).instance)
+  return function(d) { return a(d.instance) }(await C(c))
```

sbc100:
Looks like maybe some subtle async/await difference ?

kripken:
```diff
-  return a((await C(c)).instance)
+  return function(d) { return a(d.instance) }(await C(c))
```
Hmm these are identical in every way I think, except that a stack trace would show that nameless thunk function..? Weird closure doesn't remove it, since it does inlining etc. so it is ok with changes to stack traces in principle.

sbc100:
Closing favor of #26869

PR #24680, opened 2025-07-10
Fix unresolved symbol in standalone debug build.

vchigrin:
Ensure WASM binaries, produced in standalone debug mode, does not require `__throw_exception_with_stack_trace` function. In non-standalone mode it implemented in JS.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -15182,6 +15182,15 @@ w:0,t:0x[0-9a-fA-F]+: formatted: 42
     err = self.expect_fail(base_cmd + ['-sMEMORY_GROWTH_LINEAR_STEP=1mb'])
     self.assertContained('error: MEMORY_GROWTH_LINEAR_STEP is not compatible with STANDALONE_WASM', err)
 
+  def test_standalone_imports(self):
+    # Ensure standalone binary will not have __throw_exception_with_stack_trace
+    # debug helper dependency, caused by exception-related code.
+    src_path = test_file('core/test_exceptions.cpp')
+    self.run_process([EMXX, '-O0', '-fwasm-exceptions', '-sSTANDALONE_WASM', src_path])
+    imports = self.parse_wasm('a.out.wasm')[0]
+    for name in imports:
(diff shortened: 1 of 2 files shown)


vchigrin:
Guess new behavior is desirable based on  this reply
https://github.com/emscripten-core/emscripten/discussions/19150#discussioncomment-5574760
particluary, sentence
"That shouldn't be emitted in standalone mode, so we should add an if to remove that. "

sbc100:
Perhaps we could add a test for this?   

Perhaps you could update the existing `test_f_exception` in `test_other.py`  it looks like it already has a standalone version of that test

vchigrin:
Added a test - I found that separate test for this scenario looks better then integrating it into `test_f_exception`.
Checked that new test fails without first commit and passes with it.

test/test_other.py
```diff
@@ -15167,6 +15167,15 @@ def test_standalone_settings(self):
...
     self.assertContained('error: MEMORY_GROWTH_LINEAR_STEP is not compatible with STANDALONE_WASM', err)
 
+  def test_standalone_imports(self):
+    # Ensure standalone binary will not have __throw_exception_with_stack_trace
+    # debug helper dependency, caused by exception-related code.
+    src_path = test_file('core/test_exceptions.cpp')
+    self.run_process([EMXX, '-O0', '-fwasm-exceptions', '-sSTANDALONE_WASM', src_path])
+    imports = self.parse_wasm('a.out.wasm')[0]
+    for name in imports:
+        self.assertTrue(name.startswith('wasi_'), "Unexpected import %s" % name)
```
sbc100: Use two-space indentation, and single quotes for strings.

vchigrin: My bad. Fixed, thank you for noticing.

vchigrin:
Sorry, last CI run failed with some errors that, I thought, dos not relate to my changes. So I rebased PR on the current `main` branch - did not realize that it will required manual approve for CI run...

sbc100 (approved):

PR #24684, opened 2025-07-10
Enable growable array buffers

sbc100:
Currently this is still behind a flag so we never turn it on by default.
    
Fixes: #24287

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -320,7 +320,7 @@ commands:
             CHROME_FLAGS_BASE: "--no-first-run -start-maximized --no-sandbox --use-gl=swiftshader --user-data-dir=/tmp/chrome-emscripten-profile --enable-experimental-web-platform-features --enable-features=JavaScriptSourcePhaseImports"
             # Increase the window size to avoid flaky sdl tests see #24236.
             CHROME_FLAGS_HEADLESS: "--headless=new --window-size=1024,768 --remote-debugging-port=1234"
-            CHROME_FLAGS_WASM: "--enable-experimental-webassembly-features --js-flags=\"--experimental-wasm-stack-switching --experimental-wasm-type-reflection\""
+            CHROME_FLAGS_WASM: "--enable-experimental-webassembly-features --js-flags=\"--experimental-wasm-stack-switching --experimental-wasm-type-reflection --experimental-wasm-rab-integration\""
             CHROME_FLAGS_NOCACHE: "--disk-cache-dir=/dev/null --disk-cache-size=1 --media-cache-size=1 --disable-application-cache --incognito"
             # The runners lack sound hardware so fallback to a dummy device (and
             # bypass the user gesture so audio tests work without interaction)
@@ -820,6 +820,7 @@ jobs:
             other.test_gen_struct_info
             other.test_native_call_before_init
             other.test_node_unhandled_rejection
+            other.test_*growable_arraybuffers
             core2.test_hello_world
             core2.test_esm_integration*
             core0.test_esm_integration*
@@ -1023,6 +1024,7 @@ jobs:
             skip:browser.test_webgl_offscreen_canvas_in_pthread
             skip:browser.test_webgl_offscreen_canvas_in_mainthread_after_pthread
             skip:browser.test_glut_glutget
+            skip:browser.test_*_growable_arraybuffers
             "
   test-browser-firefox-wasm64:
     executor: focal
--- a/src/lib/libcore.js
+++ b/src/lib/libcore.js
@@ -182,7 +182,9 @@ addToLibrary({
     try {
       // round size grow request up to wasm page size (fixed 64KB per spec)
       wasmMemory.grow({{{ toIndexType('pages') }}}); // .grow() takes a delta compared to the previous size
+#if !GROWABLE_ARRAYBUFFERS
       updateMemoryViews();
+#endif
 #if MEMORYPROFILER
       if (typeof emscriptenMemoryProfiler != 'undefined') {
         emscriptenMemoryProfiler.onMemoryResize(oldHeapSize, wasmMemory.buffer.byteLength);
@@ -320,6 +322,7 @@ addToLibrary({
 #endif // ALLOW_MEMORY_GROWTH
   },
 
+#if !GROWABLE_ARRAYBUFFERS
   // Called after wasm grows memory. At that time we need to update the views.
   // Without this notification, we'd need to check the buffer in JS every time
   // we return from any wasm, which adds overhead. See
@@ -330,6 +333,7 @@ addToLibrary({
 #endif
     updateMemoryViews();
   },
+#endif
 
   _emscripten_system: (command) => {
 #if ENVIRONMENT_MAY_BE_NODE
@@ -2415,7 +2419,7 @@ function wrapSyscallFunction(x, library, isWasi) {
 
   library[x + '__deps'] ??= [];
 
-#if PURE_WASI
+#if PURE_WASI && !GROWABLE_ARRAYBUFFERS
   // In PURE_WASI mode we can't assume the wasm binary was built by emscripten
   // and politely notify us on memory growth.  Instead we have to check for
   // possible memory growth on each syscall.
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -4739,17 +4739,21 @@ Module["preRun"] = () => {
   # Tests memory growth in pthreads mode, but still on the main thread.
   @parameterized({
     '': ([], 1),
+    'growable_arraybuffers': (['-sGROWABLE_ARRAYBUFFERS', '-Wno-experimental'], 1),
     'proxy': (['-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME'], 2),
   })
   @no_2gb('uses INITIAL_MEMORY')
   @no_4gb('uses INITIAL_MEMORY')
   def test_pthread_growth_mainthread(self, cflags, pthread_pool_size):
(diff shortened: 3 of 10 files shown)


.circleci/config.yml
```diff
@@ -820,6 +820,7 @@ jobs:
             other.test_gen_struct_info
             other.test_native_call_before_init
             other.test_node_unhandled_rejection
+            other.test_pthread_growth_growable_sab
```
kleisauke:
```suggestion
            other.test_pthread_growth_growable_arraybuffers
```

kleisauke:
IIUC, we could also conditionally exclude these two behind `GROWABLE_ARRAYBUFFERS`:
https://github.com/emscripten-core/emscripten/blob/8f12acf2befa5cfc31e2169ee1919147b9ba998f/system/lib/standalone/standalone.c#L133-L136
https://github.com/emscripten-core/emscripten/blob/8f12acf2befa5cfc31e2169ee1919147b9ba998f/src/lib/libcore.js#L2418-L2424

sbc100:
> IIUC, we could also conditionally exclude these two behind `GROWABLE_ARRAYBUFFERS`:
> 
> https://github.com/emscripten-core/emscripten/blob/8f12acf2befa5cfc31e2169ee1919147b9ba998f/system/lib/standalone/standalone.c#L133-L136
> 
> 
> https://github.com/emscripten-core/emscripten/blob/8f12acf2befa5cfc31e2169ee1919147b9ba998f/src/lib/libcore.js#L2418-L2424

Yeah.. I thought of doing that, but didn't want to create yet another library variant.   But I think it might be worth it.

sbc100:
Done!

system/lib/standalone/standalone.c
```diff
@@ -124,7 +124,7 @@ size_t emscripten_get_heap_max() {
 }
 
 int emscripten_resize_heap(size_t size) {
-#if defined(EMSCRIPTEN_MEMORY_GROWTH)
+#if defined(EMSCRIPTEN_REPORT_MEMORY_GROWTH)
```
kleisauke: This makes `emscripten_resize_heap()` no-op when building  with `-sSTANDALONE_WASM -sALLOW_MEMORY_GROWTH -sGROWABLE_ARRAYBUFFERS`, I think it still needs to call `__builtin_wasm_memory_grow()` in that case.

Fortunately, it looks like the `EMSCRIPTEN_PURE_WASI` macro is only used in the inner `#if` directive below, so I think you can adjust the `-pure` library variant instead (i.e. `is_pure=settings.PURE_WASI or settings.GROWABLE_ARRAYBUFFERS`).

.circleci/config.yml
```diff
@@ -320,7 +320,7 @@ commands:
             CHROME_FLAGS_BASE: "--no-first-run -start-maximized --no-sandbox --use-gl=swiftshader --user-data-dir=/tmp/chrome-emscripten-profile --enable-experimental-web-platform-features --enable-features=JavaScriptSourcePhaseImports"
             # Increase the window size to avoid flaky sdl tests see #24236.
             CHROME_FLAGS_HEADLESS: "--headless=new --window-size=1024,768 --remote-debugging-port=1234"
-            CHROME_FLAGS_WASM: "--enable-experimental-webassembly-features --js-flags=\"--experimental-wasm-stack-switching --experimental-wasm-type-reflection\""
+            CHROME_FLAGS_WASM: "--enable-experimental-webassembly-features --js-flags=\"--experimental-wasm-stack-switching --experimental-wasm-type-reflection --experimental-wasm-rab-integration\""
```
kripken, line 323: Not specific to this PR, but I wonder if we should add these flags at the test level, that is, let tests add chrome flags?

test/test_browser.py
```diff
@@ -4728,6 +4728,7 @@ def test_pthread_hello_thread(self, opts, modularize):
   # Tests that a pthreads build of -sMINIMAL_RUNTIME works well in different build modes
   @parameterized({
     '': ([],),
+    'growable_arraybuffers': (['-sGROWABLE_ARRAYBUFFERS'],),
```
kripken: This doesn't enable memory growth - is this still testing what we want?

test/test_browser.py
```diff
@@ -4739,6 +4740,7 @@ def test_minimal_runtime_hello_thread(self, opts):
   # Tests memory growth in pthreads mode, but still on the main thread.
   @parameterized({
     '': ([], 1),
+    'growable_arraybuffers': (['-sGROWABLE_ARRAYBUFFERS'], 1),
```
kripken: Do we need to skip this in Firefox for now?

test/test_other.py
```diff
@@ -14285,22 +14285,30 @@ def test_pthread_kill(self):
   @node_pthreads
   @parameterized({
     '': ([], 1),
+    'growable_arraybuffers': (['-sGROWABLE_ARRAYBUFFERS'], 1),
     'proxy': (['-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME'], 2),
   })
   def test_pthread_growth_mainthread(self, cflags, pthread_pool_size):
+    if '-sGROWABLE_ARRAYBUFFERS' in cflags:
+      self.node_args.append('--experimental-wasm-rab-integration')
+      self.require_node_canary()
```
kripken, line 14295: Maybe worth adding a code size test for this, which would verify we are shipping a lot less code? Or some other kind of check that we aren't instrumenting the code in the slow way.

.circleci/config.yml
sbc100: Currently we use a single browser instance for all tests so that would require some refactoring.

test/test_browser.py
sbc100: Oops, wrong test.  Fixed.

system/lib/standalone/standalone.c
sbc100: Hmm.. I guess we don't have enough testing for this code since I don't see failures from this obvious mistake :-/

test/test_other.py
sbc100: Hmm .. good idea.

sbc100:
Added a new test

system/lib/standalone/standalone.c
kleisauke: Indeed, there's some test coverage missing here. FWIW, this line:
https://github.com/emscripten-core/emscripten/blob/d3432e7a2a6585331c94b041973fb16c9f332ce4/src/lib/libcore.js#L2418

Should probably also be adjusted to `#if PURE_WASI && !GROWABLE_ARRAYBUFFERS`.

sbc100: But then we need a new variant of the library, which i would prefer not to add if possible.

kleisauke: You don't need a new library variant if you adjust only JS code. That change prevents `emscripten_notify_memory_growth()` from being called under `-sGROWABLE_ARRAYBUFFERS`, since that symbol no longer exists in that mode.

kleisauke: ... let me put this into a separate suggestion to avoid confusion.

src/lib/libcore.js
```diff
@@ -320,6 +322,7 @@ addToLibrary({
 #endif // ALLOW_MEMORY_GROWTH
   },
 
+#if !GROWABLE_ARRAYBUFFERS
```
kleisauke, line 325: This function is still being used here without a guard for `!GROWABLE_ARRAYBUFFERS`:
https://github.com/emscripten-core/emscripten/blob/a3cf37123035461427d06dfc4372cded984baa1b/src/lib/libcore.js#L2422-L2427

(i.e. that `#if` should be adjusted to `if PURE_WASI && !GROWABLE_ARRAYBUFFERS`)

test/test_other.py
```diff
@@ -14285,23 +14285,49 @@ def test_pthread_kill(self):
   @node_pthreads
   @parameterized({
     '': ([], 1),
+    'growable_arraybuffers': (['-sGROWABLE_ARRAYBUFFERS', '-Wno-experimental'], 1),
     'proxy': (['-sPROXY_TO_PTHREAD', '-sEXIT_RUNTIME'], 2),
   })
   def test_pthread_growth_mainthread(self, cflags, pthread_pool_size):
+    if '-sGROWABLE_ARRAYBUFFERS' in cflags:
+      self.node_args.append('--experimental-wasm-rab-integration')
```
RReverser, line 14293: Worth appending to v8_args too, like we do for other experimental features?

sbc100: Done.

sbc100:
This change should be good to go now.  PTAL

test/test_other.py
```diff
@@ -14285,23 +14285,49 @@ def test_pthread_kill(self):
...
+      self.require_node_canary()
+    else:
+      self.cflags.append('-Wno-pthreads-mem-growth')
```
RReverser, line 14297: Nit: I'd keep this inline in the original flags, after all if that warning shows up in "growable array buffer" mode, we also want the test to error out as it would be unexpected.

RReverser (approved):
Left a nitpick, but otherwise very exciting!

test/test_other.py
sbc100: I'm not sure I understand.  The whole point of moving this flag so that it is not specified in "growable_arraybuffers" mode is so that if that warning ever does show up in that mode the test would error.

If we unconditionally specify `-Wno-pthreads-mem-growth` we would never see that error or warning since that flag disables it.

RReverser: Yup, my bad - I mixed up `-Wno-` with `-Werror-` 🤦‍♂️

PR #24693, opened 2025-07-13
Fix pointer arguments in addFunction + wasm64

RReverser:
Without this fix, any usages of addFunction  - e.g. dylink - would add broken functions to the webassembly table, which result in errors like "Cannot mix BigInt and other types, use explicit conversions" when invoked.

To fix this, we need to do essentially the opposite of what we do in `dynCall` and convert all pointer arguments to numbers before invoking the actual JS function, and the pointer result to bigint before returning to wasm.


kripken (approved):

test/test_other.py
```diff
@@ -15491,6 +15490,36 @@ def test_add_js_function_bigint(self, memory64, wasm_function):
...
+
+        void* p1 = (void*)26;
+        assert(f(p1) == p1 + 1);
+
+        void* p2 = (void*)493921253191;
+        assert(f(p2) == p2 + 1);
+      }
+    ''')
+
+    self.do_runf('main.c', '')
```
sbc100, line 15521: I'm a little surprised we don't have any existing tests of addFunction being used with JS functions with 'p' in their signatures. 

I'd like to take a moment to try and figure out why no existing test covered this.

RReverser: Okay, as long as it doesn't block the fix too long.

RReverser: I suspect it's possible that even if there were any tests, they 1) weren't doing pointer arithmetic and 2) weren't returning a pointer _from JS to Wasm_, only the other way around.

Without these conditions, e.g. if you have a simple tests that only logs the passed argument, the issues wouldn't trigger.

sbc100: I'm just a little suprised we don't run into this issue with our exist dynamic linking tests.  The main user or `addFunction` for JS function with a signature passed are in dynamic linking, firstly in `dlsym` and secondly (which I imagine occurs way more often) in `reportUndefinedSymbols` which will fill in missing GOT entries using JS functions.   I'm surprised we don't run into this there.

sbc100: Oh wait.. I think its because JS libraries functions are modified in place when they are constructed, right?   So this change is not necessary for JS library functions at all, only for JS functions that come from elsewhere?

Is that right?

RReverser: > Oh wait.. I think its because JS libraries functions are modified in place when they are constructed, right? So this change is not necessary for JS library functions at all, only for JS functions that come from elsewhere?
> 
> Is that right?

Hm, I don't know tbh, can't really answer that. You're definitely more familiar with both JS library and signature setup.

RReverser: > The main user or `addFunction` for JS function with a signature passed are in dynamic linking, firstly in `dlsym` and secondly (which I imagine occurs way more often) in `reportUndefinedSymbols` which will fill in missing GOT entries using JS functions. I'm surprised we don't run into this there.

See above though:

> Without these conditions, e.g. if you have a simple tests that only logs the passed argument, the issues wouldn't trigger.

So it depends on _what_ those functions are doing with pointers, not merely "are functions with pointers passed to `addFunction`".

sbc100: If that is the case I'm less convinced we should do this since this would end up creating extra wrappers / work for the JS library code. 

There are others places today where folks need to handle the "pointers are bigints" problem themselves, for example I think EM_JS/EM_ASM has no way to deal with this today and you need to modify your code to handle wasm64.    I wonder if this is just another case where we don't try to paper over the difference.  How many folks are calling `addFunction` directly?  Let along with a the sig argument set.   Did you run into this issue because you are doing this?

sbc100: All library JS functions are modified here:  https://github.com/emscripten-core/emscripten/blob/b490c4440bb467fa6171a9ceec47e1c67052ffbd/src/jsifier.mjs#L267-L277

So (IIUC) this change is not necessary for JS library functions, in fact its somewhat detrimental in that case.

RReverser: > for example I think EM_JS/EM_ASM has no way to deal with this today and you need to modify your code to handle wasm64

Right, but the way it was discussed in the past gave me an impression it's considered a bug / a limitation of the current system rather than desired outcome? UPD: I think it's also only EM_JS, EM_ASM has a way to determine pointers.

> How many folks are calling `addFunction` directly? Let along with a the sig argument set. Did you run into this issue because you are doing this?

Yeah. I can work around it manually, but contributing a fix to the `addFunction` itself seemed like the better choice since it will likely affect other users too when they add wasm64 support.

Doing things like `ptr + length` is certainly not uncommon, and any such code would break if it ends up in `addFunction` (eg via dylink) + compiled for wasm64.

RReverser: Note that the function is wrapped only if it has `p` somewhere in its signature.

So if you say JS library functions get `p` converted to `i`/`j` statically, then they should be skipped from the wrapping anyway.

sbc100: So which functions are effected then?   i.e. how did you come across this bug?  Are you trying trying to add non-js-library functions manually using `addFunction` with manual signature agument specified?

sbc100: > So if you say JS library functions get `p` converted to `i`/`j` statically, then they should be skipped from the wrapping anyway.

I don't think the current wrapping process modifies that `.sig` property of the function to remove the `p` elements.  Perhaps if it did that would alleviate that concern about double wrapping.   (But we would need to be sure that we don't need to process those `p` values elsewhere at runtime too).

RReverser: > Are you trying trying to add non-js-library functions manually using `addFunction` with manual signature agument specified?

Yeah, or, well, not quite manual - I'm computing signature in Embind (https://github.com/emscripten-core/emscripten/pull/24611 was related to this). I'm trying to use `addFunction` for another Embind-related PR and without this fix it would fail on wasm64 due to BigInt vs number incompatibility.

I can definitely fix it one level up, just for my own function, but as I said, I thought fixing it for other `addFunction` users would be the right choice.

If you strongly feel it's not, I can leave this PR open and make that other PR with just a localized hotfix.

RReverser: Ping - so what's the preferred course of action here?

sbc100: I'm not too keen on landing this, since its not necessary for JS libraries function (which are modified at code-generation time).

If you do want to continue to move forward with this we would need to make sure this code didn't create extra wrapping for JS library functions (perhaps by removing the 'p' elements from the `.sig` which we generate at compile time).  I'd also like to see this split out into a separate named function (e.g. `wrapJSFunctionAndConvertPointerArgsAndParams` .. I'll leave it to you to come with this nicer name).

I'm tempted to say we shouldn't bother landing this since (I hope) mode usage of `addFunction` is internal and not user-visible.  Calling `addFunction` from user code doesn't work in multi-threaded builds, for example (since the tables on different threads would then get out of sync).

RReverser: Ok, fair enough. Given that it's possible to use `addFunction` from user code, and that it lives in its own `libaddfunction.js`, I assumed it's part of public API, but it sounds like it's just one of many functions that isn't marked as `__internal: true` for backward compat / legacy reasons rather than intended to be used by external developers?

If so, I'm happy to also close this and fix just my individual usage of `addFunction` in the Embind PR where I needed it.

sbc100: Yeah, I think its does have some users, but I'd like to discourage its usage if possible.   I'd rather see dynamic dispatch mechanisms be built on the JS side instead of by injecting into (modifying the wasm table at runtime).    (I think only the dynamic linker should be doing that really, but I've never really articulated that in any official docs anywhere).

RReverser: Hmm I'm finding even more usecases for `addFunction` where it would save quite a bit of JS code as well as make dispatch faster, pretty much win-win.

> we would need to make sure this code didn't create extra wrapping for JS library functions (perhaps by removing the 'p' elements from the `.sig` which we generate at compile time).

So I think I'll spend some more time sorting this out after all.

> I'd rather see dynamic dispatch mechanisms be built on the JS side

Pretty much the opposite preference here, the more I play with it - if we can avoid extra JS work and perform "direct" typed calls, why not do that?

> I'm tempted to say we shouldn't bother landing this since (I hope) mode usage of `addFunction` is internal and not user-visible.

I also remembered that it's actually been a documented API for invoking JS callbacks from Wasm: https://emscripten.org/docs/porting/connecting_cpp_and_javascript/Interacting-with-code.html#calling-javascript-functions-as-function-pointers-from-c

sbc100: > Pretty much the opposite preference here, the more I play with it - if we can avoid extra JS work and perform "direct" typed calls, why not do that?

But is there any real performance advantage over having an additional lookup table on the JS side?  

Also, remember that `addFunction` does not currently work with threading, is that a limitation you are OK with?

RReverser: > But is there any real performance advantage over having an additional lookup table on the JS side?

There is for variadic functions, anywhere where you otherwise have to encode type info and number of arguments out of band. 

> Also, remember that `addFunction` does not currently work with threading, is that a limitation you are OK with?

Yeah that's fine, I'm only lazily storing it in thread-locals.

PR #24694, opened 2025-07-13
Use the more optimal overload for copying `val`

RReverser:
I found by accident (while stepping via the debugger) that for `val(v)` where `v` is `val&` C++ chooses the `T&&` overload as more precise one instead of `const val&`.

The runtime behaviour of the two is equivalent, so I'm not sure it can be easily tested, but forwarding to the const overload would certainly be a bit more optimal.

### DIFF
--- a/system/include/emscripten/val.h
+++ b/system/include/emscripten/val.h
@@ -368,6 +368,8 @@ public:
     }
   }
 
+  val(val& v) : val(static_cast<const val&>(v)) {}
+
   ~val() {
     if (uses_ref_count()) {
       internal::_emval_decref(as_handle());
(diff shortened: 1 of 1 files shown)


system/include/emscripten/val.h
```diff
@@ -368,6 +368,8 @@ class EMBIND_VISIBILITY_DEFAULT val {
     }
   }
 
+  val(val& v) : val(static_cast<const val&>(v)) {}
```
brendandahl, line 374: Comment for why we have this would be helpful.

brendandahl (approved):

system/include/emscripten/val.h
RReverser: Good idea, adding...

PR #24697, opened 2025-07-14
Move test_TextDecoder from test_browser.py to test_other.py. NFC

sbc100:
There is nothing browser-specific about this test.

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -4257,33 +4257,6 @@ Module["preRun"] = () => {
   def test_utf16_textdecoder(self):
     self.btest_exit('benchmark/benchmark_utf16.cpp', 0, cflags=['--embed-file', test_file('utf16_corpus.txt') + '@/utf16_corpus.txt', '-sEXPORTED_RUNTIME_METHODS=UTF16ToString,stringToUTF16,lengthBytesUTF16'])
 
-  @also_with_threads
-  @parameterized({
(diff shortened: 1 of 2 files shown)


sbc100:
> I think TextDecoder is a Web API, not a JS language feature. The spec is here, and not part of TC39:
> 
> https://encoding.spec.whatwg.org/#interface-textdecoder
> 
> So I believe it is not implemented in V8. It is in the embedder, so in Chrome and separately in node.js. So I think it makes sense to test both perhaps?

This test is really not testing the API though.. its really just testing the size of the generate code.

As you can see its not some tests of the API, but just using hello world.   In terms of actually exercising the APIs, that happens in other tests

I will add `@requires_node` since it won't run on d8 or spidermonkey, you are correct.

kripken (approved):
Good point, if this is mainly testing size then that can be in `other`.

Separately, I wonder if we need this setting any more? It is widely supported aside from shells, so perhaps we can automatically enable it based on ENVIRONMENT?

sbc100:
> Good point, if this is mainly testing size then that can be in `other`.
> 
> Separately, I wonder if we need this setting any more? It is widely supported aside from shells, so perhaps we can automatically enable it based on ENVIRONMENT?

This change is actually making way for a followup that tried to simplify things a little.   Currently it looks like we can at least get rid of the `=0` option.

Issue #7133, opened 2018-09-15
No glutReshapeFunc() callback on browser window resizing


erik-larsen:
In the test [hello_world_gles.c](https://github.com/kripken/emscripten/blob/incoming/tests/hello_world_gles.c),
the callback registered to glutReshapeFunc() is not being called when the browser window is resized, even when using html which stretches the canvas to cover the browser window ('soft fullscreen').

Here's a live example of the problem (try resizing the browser window):
https://erik-larsen.github.io/emscripten-glut-resizable/hello_world_gles.html

And here's a fix:
https://erik-larsen.github.io/emscripten-glut-resizable/hello_world_gles_resizable.html

The fix detects canvas size changes on redraws and explicitly calls glutReshapeWindow().  This doesn't seem ideal though, as the canvas size check has to be done on every frame, plus this seems like something the GLUT implementation should handle under the hood.  As well, the SDL2 implementation does send resize window callbacks under [similar circumstances](https://erik-larsen.github.io/emscripten-sdl2-ogles2/hello_triangle.html).

I put together a repository for this issue and the workaround fix here:
https://github.com/erik-larsen/emscripten-glut-resizable

-Erik

erik-larsen:
The problem appears to be **updateResizeListeners** is never called on browser window resize.  For comparison, when resizing the window with an SDL sample, updateResizeListeners is called via this stack: 

updateResizeListeners (hello_triangle.js:5795)
setCanvasSize (hello_triangle.js:5802)
_emscripten_set_canvas_size (hello_triangle.js:9718)
_Emscripten_HandleResize (hello_triangle.js:134253)
dynCall_iiii (hello_triangle.js:168663)
handlerFunc (hello_triangle.js:7171)
jsEventHandler (hello_triangle.js:6973)

There's no **_Emscripten_HandleResize** nor **_emscripten_set_canvas_size** in the GLUT sample, so browser resize events never make it to updateResizeListeners.

erik-larsen:
Would a reasonable fix here involve adding

`document.addEventListener('resize', GLUT.onResize, true);`
to `_glutInit()?`   

And then do the appropriate stuff in GLUT.onResize (which does not exist yet).

erik-larsen:
[...]

PR #9835, opened 2019-11-14
Fix glutReshapeFunc not being called (closes #7133)

haberbyte:
This fixes the issue in #7133 for me and the reshape function callback is called on resize.

I also added { passive: false } to the mouse event listeners as Chrome is complaining about it.
I also noticed some quite old "requestFullScreen" deprecations, should i remove those methods as well? I would like to cleanup the glut library code a bit.

PR #24699, opened 2025-07-14
Fix GLUT window resizing when CSS scaling

erik-larsen:
When using GLUT and the window is resized with the canvas dependent upon it due to CSS scaling, the result is a stretched canvas with blocky pixel scaling.   Here's a CSS scaling example:

    <style>
        canvas {
            position: fixed;
            width: 75%;
            height: 75%;
        }
    </style>
    <canvas id="canvas"></canvas>

While position fixed isn't strictly necessary, it more readily shows the problem as it makes the canvas size directly dependent upon the browser window.  For comparison, SDL behaves properly in this same scenario.

## Fix 

Three issues were found:
1. On window resize, glutReshapeFunc is never called.

2. Even with glutReshapeFunc working, the dimensions passed to it do not include CSS scaling.  Specifically, the canvas width and height are never updated with the canvas clientWidth and clientHeight, which does include scaling.

3. On GLUT program startup, glutMainLoop calls glutReshapeWindow, which is slightly problematic for the case of loading the page while already in fullscreen.  This is a problem because, while an initial resize is needed on startup, glutReshapeWindow also forces an exit from fullscreen mode.

Here are the proposed fixes:

1. Register a new resize callback `GLUT.reshapeHandler` using `window.addEventListener`, replacing `Browser.resizeListeners.push`. Previous work in this area (see below) utilized `resizeListeners`, however this fix takes a different route that is self-contained and I think simpler:
    - Using `window.addEventListener` keeps the fix entirely within `libglut.js`, avoiding any `libbrowser.js` changes as in previous attempts.  As well, `updateResizeListeners` doesn't pass CSS-scaled canvas dimensions, so changing `updateResizeListeners` implementation might be necessary and this could impact other non-GLUT clients, going beyond this GLUT-only fix.
    - Since `glutInit` already utilizes `window.addEventListener` for all other event handling, doing the same for the resize event seems consistent and simpler, as it avoids mixing event handling methods for GLUT.

2. Create a new resize callback function, `GLUT.reshapeHandler`, which does the following:
    - Updates `canvas` dimensions (via `Browser.setCanvasSize`) to `canvas.clientWidth` and `clientHeight`, so that CSS scaling is accounted for. If no CSS scaling is present, `clientWidth` and `clientHeight` match `canvas.width` and `height`, so these values are safe to use in all cases, scaling or not.
    - After updating the canvas size, pass `clientWidth` and `clientHeight` to `glutReshapeFunc`.  This is needed so that GLUT reshape callbacks can properly update their viewport transform by calling `glViewport` with the actual canvas dimensions.

3. At GLUT startup in `glutMainLoop`, call `GLUT.reshapeHandler` instead of `glutReshapeWindow`.  
    - As mentioned above, `glutReshapeWindow` has an unwanted side effect of always forcing an exit from fullscreen (and this is by design, according to the [GLUT API](https://www.opengl.org/resources/libraries/glut/spec3/node23.html)).

## Testing

Manual testing:
1. Window resizing with no CSS, CSS scaling, CSS pixel dimensions, and a mix of these for canvas width and height.
2. Entering and exiting fullscreen, and loading a page while already in fullscreen.
3. No DPI testing done (window.devicePixelRatio != 1), as GLUT is not currently DPI aware and this fix does not address it.  I did confirm on Retina Mac that this fix doesn't make this issue any better or worse.

Automated tests:
1. Added test/browser/test_glut_resize.c, with tests to assert canvas size matches target size under various scenarios (no CSS, CSS scaling, CSS pixel dimensions, and a mix of these), as well as canvas size always matching canvas client size (clientWidth and clientHeight).  
2. Since programmatic browser window resizing is not allowed for security reasons, these tests dispatch a resize event after each CSS style change as a workaround.
3. Also added tests to assert canvas size consistency after glutMainLoop and glutReshapeWindow API calls.

## Related work

All the previous work in this area worked toward enabling GLUT resize callbacks via Emscripten’s built-in event handling (specifically `Browser.resizeListeners` and `updateResizeListeners`).  As mentioned above, this fix takes a different approach that is entirely self-contained within `libglut.js`.  

This 2013 [commit](https://github.com/Emscripten-core/Emscripten/commit/6d6490e61ef9a63cbf314faa19e152796a21f3d3) added `GLUT.reshapeFunc` to `Browser.resizeListeners`, presumably to handle window resizing.  However there is no test code with that commit, and as of current Emscripten, `updateResizeListeners()` is never called on window resizing with GLUT programs, so this code is currently a no-op.

[Issue 7133](https://github.com/Emscripten-core/Emscripten/issues/7133) (that I logged in 2018, hi again!) got part of the way on a fix, but used `glutReshapeWindow` which has the previously mentioned side effect of exiting fullscreen.  This was closed unresolved.

[PR 9835](https://github.com/Emscripten-core/Emscripten/pull/9835) proposed a fix for 7133.  Also closed unresolved, this fix involved modifying `libbrowser.js` in order to get resize callbacks to GLUT via `resizeListeners`.  While this got resize callbacks working, in my testing it didn’t pass CSS-scaled canvas size in the callback (the all-important clientWidth and clientHeight).

I also looked at how SDL handles resizing, which uses `resizeEventListeners`, but decided the more straightforward fix was to use `addEventListener`.  Last, I looked at [GLFW CSS scaling test](https://github.com/emscripten-core/emscripten/blob/main/test/browser/test_glfw3_css_scaling.c) which was helpful in writing the automated tests and also to confirm that no DPI ratio work is addressed by this fix.

Fixes: #7133


sbc100:
@ypujante will certainly have some thoughts in this area.

sbc100:
In general I think moving away from `Browser.resizeListeners` and into a GLUT-specific logic seems like a regression since there will be less shared code, but I get the it might be needed in this case.

erik-larsen:
I couldn't find a path using `Browser.resizeListeners` that doesn't widen the fix beyond GLUT.  Given GLUT is on life support (?)  a self-contained fix seems like the better way to go.  As well, [glutInit](https://github.com/emscripten-core/emscripten/blob/main/src/lib/libglut.js#L309) is already using `window.addEventListener` for all other event callbacks so this seems consistent with that.

sbc100:
Fair enough.  It looks like the use of `Browser.resizeListeners` was added in 6d6490e61ef9a63cbf314faa19e152796a21f3d3, but I don't see any specific rationale for it.

src/lib/libglut.js
```diff
@@ -298,6 +298,15 @@ var LibraryGLUT = {
...
       _glutPostRedisplay();
+    },
+
+    reshapeHandler: () => {
+      // Use clientWidth and clientHeight, which include CSS scaling of the canvas
+      var canvas = Browser.getCanvas();
+      Browser.setCanvasSize(canvas.clientWidth, canvas.clientHeight, true);
+      if (GLUT.reshapeFunc) {
+        {{{ makeDynCall('vii', 'GLUT.reshapeFunc') }}}(canvas.clientWidth, canvas.clientHeight);
+      }
```
sbc100: Could we keep the `Browser.resizeListeners.push` below and delete 3 lines?

i.e. could we execute these 3 lines via the `resizeListeners` system?  That way the other calls to `Browser.updateResizeListeners` in `libbrowser.js` will continue to work for GLUT.

erik-larsen: I'm not entirely sure what you mean by 'these 3 lines' - if you mean could we register `reshapeHandler` with `resizeListeners.push`, yes that should be possible.  However, it requires more than just registering the callback, because currently `updateResizeListeners` is never called for GLUT.  I think possibly due to linker dependency analysis not finding a dependency and optimizing out the call to `updateResizeListeners`?   There'd need to be some additional code to get `updateResizeListeners` to be called as well.

erik-larsen: Also worth mentioning `GLUT.reshapeHandler` is necessary as a replacement for the `glutReshapeWindow` call in [glutMainLoop](https://github.com/emscripten-core/emscripten/blob/main/src/lib/libglut.js#L637), so that we don't force exiting fullscreen if we enter the GLUT page in fullscreen.

sbc100: What I mean here:

1. Keep the new `addEventListener` line and have it call this function (maybe call it `resizeHandler`?)
2. Also keep the `Browser.resizeListeners.push` line and have that just make the call to `GLUT.reshapeFunc`.
3. `setCanvasSize` will then call the `Browser.resizeListeners` list from the resizeHandler.

erik-larsen: GLUT nomenclature for resize is (for whatever reason) _reshape_, hence the `reshapeFunc` naming.  My preference would be to stay with that convention, as grepping for 'reshape' will turn up all parts of GLUT resizing in `libglut.js`: `glutReshapeWindow, GLUT.reshapeFunc, GLUT.reshapeHandler`, etc., instead of having to grep for reshape and resize.

Registering `GLUT.reshapeFunc` with  `Browser.resizeListeners.push` is certainly possible, let me try out some code and get back with results. Previously `updateResizeListeners` wasn't ever being called on window resizes, but as you point out, maybe `setCanvasSize` will do this.

sbc100: Right, I'm wonder if we have just add some code to get the existing `resizeListeners` firing in GLUT mode.

I get that `reshape` comes from glut, but the event being registered for is what the web calls "resize" so I suppose both are logical enough.  But we are going to need two different names here.  One for the `resize` event handler that we receive from the browser and forward onto the `Browser` object and then another for the `updateResizeListeners` callback we receive from the `Browser` object.

erik-larsen: So `setCanvasSize` does indeed call `updateResizeListeners`, but that won't work here, because `setCanvasSize` is never called on window resizing, at least with GLUT.   `GLUT.reshapeHandler` does call `setCanvasSize`, but that's _after_ being called back on a window resize event.

```
Current proposal:
1. addEventListener('resize', GLUT.reshapeHandler)
2. window resize event -> GLUT.reshapeHandler -> setCanvasSize

Suggested change:
1. Browser.resizeListeners.push
2. window resize event -> <nothing happens: no setCanvasSize call, no updateResizeListeners call)
3. some additional code is necessary to get updateResizeListeners called
```

I'm not seeing an advantage to pursuing this suggested change.  What resizing cases does `resizeListeners` cover that `addEventListener` does not?

sbc100: Sorry I'm not suggesting you remove the window resize event handler here.

I'm suggesting that you simplify it by removing the last 3 lines.

erik-larsen: Ah, I think I understand now, please confirm.  

The callback sequence would be:
` window resize -> GLUT.resizeCanvas -> setCanvasSize -> updateResizeListeners -> GLUT.reshapeFunc`

And the setup would be:

Add this:
```
    resizeCanvas: () => {
      // Use clientWidth and clientHeight, which include CSS scaling of the canvas
      var canvas = Browser.getCanvas();
      Browser.setCanvasSize(canvas.clientWidth, canvas.clientHeight, true);
    }
```
Keep this:
```
    Browser.resizeListeners.push((width, height) => {
      if (GLUT.reshapeFunc) {
        {{{ makeDynCall('vii', 'GLUT.reshapeFunc') }}}(width, height);
      }
    });
```
Add this:
```
window.addEventListener('resize', GLUT.resizeCanvas, true);
```

sbc100: Yes, that looks like what I am suggesting.   Sorry for the confusion.

The main reason I think this might be a good idea is that GLUT would then continue to receive resize events from other sources (It looks like resizeListeners are called from libbrowser.js for several different reasons).

erik-larsen: Confirmed these code changes work, though need to call `setCanvasSize` with `noUpdates = false` so updateResizeListeners is called.  And I think the best name would be `GLUT.onResize()`, to go with all the other `GLUT.on*()` handlers.

It is more complex though, having two callbacks registered in order to resize, and having the resizing work done in two stages, rather than one callback registered and one resize handler.    On the other hand, with this approach GLUT is set up to handle anyone that calls `Browser.setCanvasSize, setFullscreenCanvasSize, setWindowedCanvasSize`.

If this seems like the right direction to go, please let me know and I'll do further testing and commit an update.

ypujante:
I don't have any knowledge of GLUT so I am not sure what I can contribute.

I will say that there was a [PR](https://github.com/emscripten-core/emscripten/pull/22900) that I authored to restore CSS scaling that I had accidentally removed when working on Hi DPI support for `libglfw.js`. I am not sure if that can be of any help but I thought I would point to it (there is a test that you might want to look at, maybe it can help with this effort)

erik-larsen:
Thanks - I did look at [GLFW CSS scaling test](https://github.com/emscripten-core/emscripten/blob/main/test/browser/test_glfw3_css_scaling.c) which was helpful in writing the automated tests and also to confirm that no DPI ratio work is addressed by this fix.  GLUT doesn't handle Hi DPI but this fix doesn't make it any better or worse.

erik-larsen:
Looking into CI failures now.

erik-larsen:
Added potential fix for CI tests.

erik-larsen:
All of the CI errors related to this PR appear to be fixed.  The remaining CI errors look unrelated to the fix, `test_glut_resize.c`, and GLUT in general:

[test-windows](https://app.circleci.com/pipelines/github/emscripten-core/emscripten/43811/workflows/eb39acaa-40fe-4c8d-8a8f-f514e4c7ab8a/jobs/973181)
`ninja not installed. An error occurred during installation:
 Error downloading 'ninja.1.13.1' from 'https://community.chocolatey.org/api/v2/package/ninja/1.13.1'.`

[test-other](https://app.circleci.com/pipelines/github/emscripten-core/emscripten/43811/workflows/eb39acaa-40fe-4c8d-8a8f-f514e4c7ab8a/jobs/973191)
`FAIL [0.002s]: test_codesize_hello_dylink_all (test_other.other)`

[test-mac-arm64](https://app.circleci.com/pipelines/github/emscripten-core/emscripten/43811/workflows/eb39acaa-40fe-4c8d-8a8f-f514e4c7ab8a/jobs/973205)
`FAIL: test_codesize_hello_dylink_all (test_other.other.test_codesize_hello_dylink_all)`

[test-browser-chrome-wasm64-4gb](https://app.circleci.com/pipelines/github/emscripten-core/emscripten/43811/workflows/eb39acaa-40fe-4c8d-8a8f-f514e4c7ab8a/jobs/973194)
`[49122:49122:0717/193332.423191:ERROR:CONSOLE:130] "got top level error: exit(0)", source: blob:http://localhost:8888/7de7e610-10a0-46d6-a81e-4ec04042ecec (130)`

[test-browser-chrome]()
```
FAIL [1.920s]: test_gl_stride (test_browser.browser)
TypeError: Cannot read properties of undefined (reading 'clearColor')
    at _glClearColor (http://localhost:8888/test.js:4040:49)

FAIL [1.495s]: test_webgl_multi_draw_arrays (test_browser.browser)
TypeError: Cannot read properties of undefined (reading 'GLctx')
    at _emscripten_webgl_enable_WEBGL_multi_draw (http://localhost:8888/test.js:1833:107)

FAIL [0.975s]: test_webgl_multi_draw_arrays_instanced (test_browser.browser)
TypeError: Cannot read properties of undefined (reading 'GLctx')
    at _emscripten_webgl_enable_WEBGL_multi_draw (http://localhost:8888/test.js:1833:107)

FAIL [1.074s]: test_webgl_multi_draw_elements (test_browser.browser)
TypeError: Cannot read properties of undefined (reading 'GLctx')
    at _emscripten_webgl_enable_WEBGL_multi_draw (http://localhost:8888/test.js:1833:107)

FAIL [0.986s]: test_webgl_multi_draw_elements_instanced (test_browser.browser)
TypeError: Cannot read properties of undefined (reading 'GLctx')
    at _emscripten_webgl_enable_WEBGL_multi_draw (http://localhost:8888/test.js:1833:107)
```

erik-larsen:
I forgot to mention my motivation for seeing this fix in GLUT:  I'm interested in preserving old OpenGL graphics demos.  Emscripten is the perfect vehicle for this, as I can compile C and C++ to the web, and then the demos can be run forever more, on any device with a web browser.   Without Emscripten, a lot of this code is lost to time.

erik-larsen:
@kripken We discussed this issue back in [2018](https://github.com/Emscripten-core/Emscripten/issues/7133 ) and [2019](https://github.com/Emscripten-core/Emscripten/pull/9835).  This new fix avoids any changes to libbrowser.js and non-GLUT code paths. 

@sbc100 Your recommended code changes have been integrated and testing has passed (the latest CI test failures appear unrelated to these changes).

test/browser/test_glut_resize.c
```diff
@@ -0,0 +1,208 @@
...
+  glutCreateWindow("test_glut_resize.c");
+
+  /* Set up glut callback functions */
+  glutIdleFunc(start_tests);
+  glutReshapeFunc(reshape);
+  glutDisplayFunc(NULL);
+
+  glutMainLoop();
+  return 0;
+}
```
sbc100: Missing newline

sbc100 (approved):

src/lib/libglut.js
```diff
@@ -633,10 +646,9 @@ var LibraryGLUT = {
   },
 
   glutMainLoop__proxy: 'sync',
-  glutMainLoop__deps: ['$GLUT', 'glutReshapeWindow', 'glutPostRedisplay'],
+  glutMainLoop__deps: ['$GLUT', 'glutPostRedisplay'],
   glutMainLoop: () => {
-    var canvas = Browser.getCanvas();
-    _glutReshapeWindow(canvas.width, canvas.height);
+    GLUT.onResize();
```
kripken, line 652: Maybe add a comment here? It's not obvious to me why we need to resize when starting the main loop.

sbc100: Maybe yes, although this is pre-existing behaviour (i.e. the existing code is doing a reshape without any comment).

kripken: Oh, right, thanks @sbc100 

Still, the new comment can't hurt, thanks @erik-larsen

sbc100:
Hmm.. I recently added test_codesize_hello_dylink_all and I think it might be too sensitive.  I think I will revert that.

sbc100:
Can you rebase (or merge) one more time.

erik-larsen:
@sbc100 I see @kripken merged, are we good or is another merge needed?

sbc100:
Looks like you need to do a `emsdk install tot` followed by `./tools/main/rebaseline_tests.py`.   if that doesn't work when you do I can take a look too.

erik-larsen:
Okay, updated to emsdk tot and latest emscripten, ran `./tools/maint/rebaseline_tests.py`.  Results:

Ran 60 tests in 170.629s

OK (skipped=1)
Total core time: 2301.424s. Wallclock time: 170.629s. Parallelization: 13.49x.
processing test/code_size/test_codesize_cxx_ctors1.json
processing test/code_size/test_codesize_cxx_ctors2.json
processing test/code_size/test_codesize_cxx_except.json
processing test/code_size/test_codesize_cxx_mangle.json
processing test/code_size/test_codesize_cxx_noexcept.json
processing test/code_size/test_codesize_cxx_wasmfs.json
processing test/code_size/test_codesize_hello_O0.json
processing test/code_size/test_codesize_hello_dylink_all.json
processing test/code_size/test_unoptimized_code_size.json

Automatic rebaseline of codesize expectations. NFC

This is an automatic change generated by tools/maint/rebaseline_tests.py.

The following (9) test expectation files were updated by

```
code_size/test_codesize_cxx_ctors1.json: 149256 => 149232 [-24 bytes / -0.02%]
code_size/test_codesize_cxx_ctors2.json: 148662 => 148638 [-24 bytes / -0.02%]
code_size/test_codesize_cxx_except.json: 194695 => 194671 [-24 bytes / -0.01%]
code_size/test_codesize_cxx_mangle.json: 258786 => 258762 [-24 bytes / -0.01%]
code_size/test_codesize_cxx_noexcept.json: 151674 => 151650 [-24 bytes / -0.02%]
code_size/test_codesize_cxx_wasmfs.json: 176935 => 176911 [-24 bytes / -0.01%]
code_size/test_codesize_hello_O0.json: 37467 => 37453 [-14 bytes / -0.04%]
code_size/test_codesize_hello_dylink_all.json: 844590 => 844684 [+94 bytes / +0.01%]
code_size/test_unoptimized_code_size.json: 175109 => 175067 [-42 bytes / -0.02%]

Average change: -0.01% (-0.04% - +0.01%)
```
Anything else I can do?

sbc100:
That doesn't looks quite right... can you make sure you are up-to-date with the emscripten main branch.  And also try `emcc --clear-cache` before you run the rebaseline script?

erik-larsen:
Ok.  Ran `emcc --clear-cache` and reran rebaseline:

Ran 60 tests in 187.014s

OK (skipped=1)
Total core time: 2500.130s. Wallclock time: 187.014s. Parallelization: 13.37x.
processing test/code_size/test_codesize_cxx_ctors1.json
processing test/code_size/test_codesize_cxx_ctors2.json
processing test/code_size/test_codesize_cxx_except.json
processing test/code_size/test_codesize_cxx_mangle.json
processing test/code_size/test_codesize_cxx_noexcept.json
processing test/code_size/test_codesize_cxx_wasmfs.json
processing test/code_size/test_codesize_hello_O0.json
processing test/code_size/test_codesize_hello_dylink_all.json
processing test/code_size/test_unoptimized_code_size.json

Automatic rebaseline of codesize expectations. NFC

This is an automatic change generated by tools/maint/rebaseline_tests.py.

The following (9) test expectation files were updated by

```
code_size/test_codesize_cxx_ctors1.json: 149232 => 149256 [+24 bytes / +0.02%]
code_size/test_codesize_cxx_ctors2.json: 148638 => 148662 [+24 bytes / +0.02%]
code_size/test_codesize_cxx_except.json: 194671 => 194695 [+24 bytes / +0.01%]
code_size/test_codesize_cxx_mangle.json: 258762 => 258786 [+24 bytes / +0.01%]
code_size/test_codesize_cxx_noexcept.json: 151650 => 151674 [+24 bytes / +0.02%]
code_size/test_codesize_cxx_wasmfs.json: 176911 => 176935 [+24 bytes / +0.01%]
code_size/test_codesize_hello_O0.json: 37453 => 37467 [+14 bytes / +0.04%]
code_size/test_codesize_hello_dylink_all.json: 844684 => 844702 [+18 bytes / +0.00%]
code_size/test_unoptimized_code_size.json: 175067 => 175109 [+42 bytes / +0.02%]

Average change: +0.02% (+0.00% - +0.04%)
```

sbc100:
I don't think this change should include changes to most of those test files.   Something odd must be going on there.

erik-larsen:
Yeah I don't see how a change to libglut.js would cause changes elsewhere like this.  Maybe something with my environment?  Tried on an intel iMac on Sonoma 14.7.6 and a Macbook with latest Sequoia - same diffs.

sgi-demos:
Tried Macbook air M4 Sequoia 15.5, same diffs.

erik-larsen:
Curious that the diffs of the second run exactly cancels out the first run?  I'm not familiar with rebaseline_tests.py, would one run affect the next?

sbc100:
If you are having trouble rebaselining I can update the PR for you with the correct expectations.

erik-larsen:
Yes, please.  This is my first PR, and I'm happy to learn how to rebaseline if you can point me to documentation, otherwise please have at it.

PR #19559, opened 2023-06-08
Use dedicated pointer types in syscall layer. NFC

kleisauke:
Split out from #13007.

PR #24706, opened 2025-07-15
[WasmFS] Use `const char*` in JS API impl. NFC

kleisauke:
In line with #23825.

### DIFF
--- a/system/lib/wasmfs/js_api.cpp
+++ b/system/lib/wasmfs/js_api.cpp
@@ -64,7 +64,7 @@ int _wasmfs_read_file(const char* path, uint8_t** out_buf, off_t* out_size) {
 
 // Writes to a file, possibly creating it, and returns the number of bytes
 // written successfully. If the file already exists, appends to it.
-int _wasmfs_write_file(const char* pathname, char* data, size_t data_size) {
+int _wasmfs_write_file(const char* pathname, const char* data, size_t data_size) {
   auto parsedParent = path::parseParent(pathname);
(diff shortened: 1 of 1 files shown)


kleisauke:
Test failures are due to https://discourse.llvm.org/t/llvm-21-x-release-branch-created-main-is-now-22/87374.

sbc100:
https://github.com/emscripten-core/emscripten/pull/24707

system/lib/wasmfs/js_api.cpp
```diff
@@ -64,7 +64,7 @@ int _wasmfs_read_file(const char* path, uint8_t** out_buf, off_t* out_size) {
 
 // Writes to a file, possibly creating it, and returns the number of bytes
 // written successfully. If the file already exists, appends to it.
-int _wasmfs_write_file(const char* pathname, char* data, size_t data_size) {
+int _wasmfs_write_file(const char* pathname, const char* data, size_t data_size) {
```
sbc100: I wonder why we don't use `uint8_t` here to like that `.write` method below does?

sbc100 (approved):

system/lib/wasmfs/js_api.cpp
kleisauke: Commit 251c091e23dc0d31f6372ba93e39db47df4ca9b0 uses `uint8_t` here, which would also avoid the cast below.

Note that there are still many `intptr_t` casts in this file, these will be addressed in PR #19559.

PR #24708, opened 2025-07-15
Centralize fullscreenElement usage. NFC

sbc100:
I will likely followup with another PR to reduce the size of this function using `MIN_XXX_VERSION` macros, but I wanted to keep this change simple and NFC.


kripken (approved):

kripken:
Can we not remove these prefixes entirely like the others?

sbc100:
> Can we not remove these prefixes entirely like the others?

Yup, planning a folllowup.

PR #24712, opened 2025-07-16
Use getFullscreenElement over legacy fullscreen attribute. NFC

sbc100:
The getFullscreenElement utility has fallbacks for old browsers going back as far as we support.

See https://caniuse.com/mdn-api_document_fullscreenelement

### DIFF
--- a/src/lib/libglfw.js
+++ b/src/lib/libglfw.js
@@ -632,7 +632,7 @@ var LibraryGLFW = {
       var resizeNeeded = false;
 
       // If the client is requesting fullscreen mode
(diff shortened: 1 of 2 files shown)


bootstrap.bat
```diff
@@ -23,7 +23,7 @@
 :: %~dp0 expansions will not work.
 :: So first try if %~dp0 might work, and if not, manually look up this script from PATH.
 @if exist "%~f0" (
-  set MYDIR=%~dp0
+  set MYDIR="%~dp0"
```
kripken: unrelated?

kripken (commented):
lgtm without unrelated changes

Issue #16706, opened 2022-04-12
--closure 1 breaks threads when WASM is set to 0 and PROXY_TO_PTHREAD is set to 1


Username404-59:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.8 (3ff7eff373d31d8c0179895165462019221f192e)
clang version 15.0.0 (https://github.com/llvm/llvm-project 80ec0ebfdc5692a58e0832125f2c6a991df9d63f)
Target: wasm32-unknown-emscripten
Thread model: posix
```

**Output with `-v` appended:**
```
[...]

PR #24720, opened 2025-07-16
Run wasm2js2.test_pthread_proxying as part of CI. NFC

sbc100:
This confirms that wasm2js + pthreads + closure compiler are compatible.

See #16706.  I believe this issues was likely fixed when we removed the separate worker.js file.

Replaces: #24719
Fixes: #16706

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -437,6 +437,9 @@ class TestCoreBase(RunnerCore):
   def maybe_closure(self):
     if '--closure=1' not in self.cflags and self.should_use_closure():
       self.cflags += ['--closure=1']
+      if self.is_wasm2js():
+        # wasm2js output currently contains closure warnings
+        self.cflags += ['-Wno-closure']
       logger.debug('using closure compiler..')
       return True
     return False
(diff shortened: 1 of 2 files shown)


test/test_core.py
```diff
@@ -437,6 +437,9 @@ def setup_esm_integration(self):
   def maybe_closure(self):
     if '--closure=1' not in self.cflags and self.should_use_closure():
       self.cflags += ['--closure=1']
+      if self.is_wasm2js():
+        # wasm2js output currently contains closure warnings
+        self.cflags += ['-Wno-closure']
```
kripken, line 442: doesn't this mean we'd lose coverage for wasm2js+closure?

kripken: if there is a bug in the new test enabled here, perhaps disable closure on wasm2js specifically there?

sbc100: Basically closure + wasm2js genrates warnings, and by default we run with `-Werror` in our test code.

This means that any tests that do wasm2js + closure also need to disabled closure warnings using `-Wno-closure`.  This is already done elsewhere in the tests (search for `-Wno-closure`) in test_other.py, for example.

Did you perhaps think that this line was disabling closure itself (and not just closure warnings)?

kripken: Oh, yes, I missed the "W" and read that as `--closure 0` somehow... sorry!

PR #24727, opened 2025-07-16
Simplify MODULARIZE code to always use async factory function

sbc100:
The docs here were wrong.  We have the `addRunDependency` system that can delay module's readiness for an arbitrary amount of time, even if async compilation is disabled.

This `addRunDependency` system does not exist in MINUMAL_RUTTIME, so in some cases we can avoid the ready promise in MINIMAL_RUNTIME mode, but that is a pretty subtle distinction that only applies to MINIMAL_RUNTIME. Also even in this mode we were still marking the factory function as async (note the lack of USE_READY_PROMISE in the existing condition).


site/source/docs/tools_reference/settings_reference.rst
```diff
@@ -1863,14 +1863,9 @@ MODULARIZE
...
 page, it will use the global scope. With ``MODULARIZE`` set, we instead emit
-the code wrapped in a function that returns a promise. The promise is
-resolved with the module instance when it is safe to run the compiled code,
-similar to the ``onRuntimeInitialized`` callback. You do not need to use the
-``onRuntimeInitialized`` callback when using ``MODULARIZE``.
-
-(If WASM_ASYNC_COMPILATION is off, that is, if compilation is
-*synchronous*, then it would not make sense to return a Promise, and instead
-the Module object itself is returned, which is ready to be used.)
+the code wrapped in an async function. This functions returns a promise that
```
kripken:
```suggestion
the code wrapped in an async function. This function returns a promise that
```

site/source/docs/tools_reference/settings_reference.rst
```diff
@@ -1863,14 +1863,9 @@ MODULARIZE
...
+resolves to a module instance when once it is safe to run the compiled code
```
kripken:
```suggestion
resolves to a module instance when it is safe to run the compiled code
```

kripken: (could also be "once", but not "when once" together, I think)

kripken (approved):

sbc100:
I updated the test case and the ChangeLog.  @kripken can you can another look and confirm that you agree that this simplification also improves correctness and is worth the potentially-breaking change.

ChangeLog.md
```diff
@@ -20,6 +20,11 @@ See docs/process.md for more on how version tagging works.
 
 4.0.12 (in development)
 -----------------------
+- In `-sMODULARIZE` mode the factory function will now always return a promise,
+  even when `WASM_ASYNC_COMPILATION` is disabled.  This is because emscripten
+  has other features that might also return async module creation (e.g. loading
+  files over the next, or any other user of the `addRunDependency` API).  For
```
```suggestion
  files over the web, or any other user of the `addRunDependency` API).  For
```

svenpaulsen:
Not sure why you guys constantly introduce breaking changes in patch version releases... :-/

sbc100:
> Not sure why you guys constantly introduce breaking changes in patch version releases... :-/

Sorry if this broke your use case.  Can you explain how so we can try to mitigate?

BTW, Emscripten does not do semantic versioning (at least not right now) so any no two releases are guaranteed to be 100% compatible and there is no such thing as a patch release, per say.   Of course we still do our best to avoid and limit breakages.  For more discussion on possible ways to improve that see the https://github.com/emscripten-core/emscripten/issues/19053

svenpaulsen:
The fix was quite easy, but in the past I had explicitly disabled async compilation - something that no longer had any effect after upgrading emsdk. It would've been super helpful to have a deprecation notice/warning when using options like `BINARYEN_ASYNC_COMPILATION` or `WASM_ASYNC_COMPILATION` during compilation. It's just frustrating when your code suddenly behaves differently without any changes on your end - so I hope you'll excuse the tone of my previous message. ;-)

sbc100:
Thanks for the feedback.  We did include a changelog entry to this change in behaviour:

https://github.com/emscripten-core/emscripten/blob/6e482b581b245aea4d16a328074745a3cfb7f022/ChangeLog.md?plain=1#L36-L40

However, I agree it would have been good to provide some kind of compile time warning too.  Its a little tricky to do that kind of thing since we don't want to continue showing the warning once a user has seen it an understood it.  But I am open to ideas about how we might do this.  For example, any user who uses `-sMODULARIZE` in combination with `-sWASM_ASYNC_COMPILATION=0` could see a warning like "The behaviour of XXX and YYY settings combined changes in version X.Y.X of emscripten".. which could then be suppressed by `-Wno-behaviour-change` but I fear it would over time grow to effect too many folks and everyone would end up building with `-Wno-behaviour-change`.    Ultimately I think maybe the ChangeLog is the currently the best place we have for this.

BTW, `-sWASM_ASYNC_COMPILATION=0` does still change the internal instanation to be syncronous, its just the outer function that creates the "Module" is now always async.   I suppose is makes `-sWASM_ASYNC_COMPILATION=0` kind of pointless in that case (at least in the `-sMODULARIZE` case) since the effect is not really observable.

Is there a particular reason you were using `-sMODULARIZE` with `-sWASM_ASYNC_COMPILATION=0`?

PR #24732, opened 2025-07-17
[AUDIO_WORKLET] Add support for MEMORY64 with 2GB and 4GB heap

cwoffenden:
This pulls the changes out of #23508 to add just the `MEMORY64` support.

The already merged interactive tests (from #23659) can be run with:

```
test/runner.py interactive_2gb.test_audio_worklet_params_mixing
test/runner.py interactive64_4gb.test_audio_worklet_params_mixing
```

The browser tests with wasm64 have been re-enabled.

Note: the intention is to add the performance improvements (from last October) once this is merged (which are still relevant and produce the same improvements).

~Will need a once-over when #24734 lands.~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -557,6 +557,7 @@ jobs:
             asan.test_minimal_runtime_global_initializer
             asan.test_fs_js_api_wasmfs
             asan.test_modularize_instance_pthreads
+            asan.test_minimal_runtime_hello_world
             lsan.test_dylink_dso_needed
             lsan.test_stdio_locking
             lsan.test_dlfcn_basic
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -8788,13 +8788,14 @@ NODEFS is no longer included by default; build with -lnodefs.js
   @no_wasmfs('https://github.com/emscripten-core/emscripten/issues/16816')
   @no_modularize_instance('MODULARIZE=instance is not compatible with MINIMAL_RUNTIME')
   @parameterized({
-    'default': ([],),
+    '': ([],),
     'streaming_inst': (['-sMINIMAL_RUNTIME_STREAMING_WASM_INSTANTIATION'],),
     'no_export': (['-sDECLARE_ASM_MODULE_EXPORTS=0'],),
   })
   @requires_node  # TODO: Support for non-Node.js shells under MINIMAL_RUNTIME
   def test_minimal_runtime_hello_world(self, args):
     self.maybe_closure()
+    self.cflags += ['--pre-js', test_file('minimal_runtime_exit_handling.js')]
     self.do_runf('small_hello_world.c', 'hello!', cflags=['-sMINIMAL_RUNTIME'] + args)
 
   # Test that printf() works in MINIMAL_RUNTIME=1


src/lib/libwebaudio.js
```diff
@@ -222,32 +222,33 @@ let LibraryWebAudio = {
     assert(EmAudio[contextHandle] instanceof (window.AudioContext || window.webkitAudioContext), `Called emscripten_create_wasm_audio_worklet_processor_async() on a context handle ${contextHandle} that is not an AudioContext, but of type ${typeof EmAudio[contextHandle]}`);
 #endif
 
-    options >>= 2;
     let audioParams = [],
-      numAudioParams = HEAPU32[options+1],
-      audioParamDescriptors = HEAPU32[options+2] >> 2,
+      processorName = UTF8ToString({{{ makeGetValue('options', C_STRUCTS.WebAudioWorkletProcessorCreateOptions.name, '*') }}}),
```
cwoffenden: Hmm, why would `closure` reduce this to:
```
ja=(a,b,c,d)=>{let e=[],g=(g=K[b>>2])?V(g):"" ...
```
It should be `b` and not the undeclared `g`.

cwoffenden: It's the `let` keyword that breaks this when used with `closure`.

cwoffenden:
These failing size related tests are unrelated to the changes:
```
test_minimal_runtime_code_size_hello_webgl_wasm2js
test_minimal_runtime_code_size_hello_webgl_wasm
test_minimal_runtime_code_size_hello_webgl2_wasm2js
test_minimal_runtime_code_size_hello_webgl2_wasm
```
They don't fail for me with Emscripten 4.0.12 built from `sdk-main-64bit`.

The failing `test_codesize_hello_dylink_all` is fixed in #24742.

sbc100:
> These failing size related tests are unrelated to the changes:
> 
> ```
> test_minimal_runtime_code_size_hello_webgl_wasm2js
> test_minimal_runtime_code_size_hello_webgl_wasm
> test_minimal_runtime_code_size_hello_webgl2_wasm2js
> test_minimal_runtime_code_size_hello_webgl2_wasm
> ```
> 
> They don't fail for me with Emscripten 4.0.12 built from `sdk-main-64bit`.
> 
> The failing `test_codesize_hello_dylink_all` is fixed in #24742.

To successfully rebaseline codesize tests on main you need to first run `emsdk install tot` to ensure you have the very latest llvm + binaryen running.

For pre-existing codesize changes we have a bot that automatically updates them.. I'll trigger it now so that main is up-to-date.

src/lib/libwebaudio.js
```diff
@@ -222,32 +222,33 @@ let LibraryWebAudio = {
...
 
-    options >>= 2;
-    let audioParams = [],
-      numAudioParams = HEAPU32[options+1],
-      audioParamDescriptors = HEAPU32[options+2] >> 2,
+    var audioParams = [],
+      processorName = UTF8ToString({{{ makeGetValue('options', C_STRUCTS.WebAudioWorkletProcessorCreateOptions.name, '*') }}}),
+      numAudioParams = {{{ makeGetValue('options', C_STRUCTS.WebAudioWorkletProcessorCreateOptions.numAudioParams, 'i32') }}},
+      audioParamDescriptors = {{{ makeGetValue('options', C_STRUCTS.WebAudioWorkletProcessorCreateOptions.audioParamDescriptors, '*') }}},
       i = 0;
```
sbc100: Can we convert to one-line-per-declaration here?  I think it would help make this more readable, and should not effect minified code size.

cwoffenden:
> To successfully rebaseline codesize tests on main you need to first run `emsdk install tot` to ensure you have the very latest llvm + binaryen running.

What's the difference between `tot` and `sdk-main-64bit`? I've always used `sdk-main-64bit` because I can quickly switch between Emscripten's origin (`main` in the repo) and my own remote's branch.

sbc100:
> > To successfully rebaseline codesize tests on main you need to first run `emsdk install tot` to ensure you have the very latest llvm + binaryen running.
> 
> What's the difference between `tot` and `sdk-main-64bit`? I've always used `sdk-main-64bit` because I can quickly switch between Emscripten's origin (`main` in the repo) and my own remote's branch.

`tot` is like `4.0.10` in that installs pre-built binaries from our build servers (https://ci.chromium.org/p/emscripten-releases/g/main/console)

`sdk-main-64bit` is not used my many folks, it builds everything from source (which most ppl don't want to do).   I think it updates the main branch each time its run but I don't remember for sure.

The `tot` binaries are what the emscripten CI tests run against.

The easiest way to use any emscripten checkout without emsdk binaries is to do set `EM_CONFIG=/path/to/emsdk/.emscripten`.  Then you can do a git checkout of emscripten and run `./emcc ...` and the emscripten checkout will use llvm and binaryen binaries from emsdk.   See https://emscripten.org/docs/tools_reference/emsdk.html#how-do-i-use-my-own-emscripten-fork-with-the-sdk

cwoffenden:
> `sdk-main-64bit` is not used my many folks, it builds everything from source (which most ppl don't want to do).

I think it’s a throwback from LLVM fixes we were doing, and it’s stuck. I’ll switch to `tot`.

cwoffenden:
Unrelated fails seem Chrome related, same as #24750 for example.

(Re-run and passed)

src/lib/libwebaudio.js
```diff
@@ -262,18 +263,20 @@ let LibraryWebAudio = {
     assert(EmAudio[contextHandle], `Called emscripten_create_wasm_audio_worklet_node() with a nonexisting/already freed Web Audio Context handle ${contextHandle}!`);
     assert(EmAudio[contextHandle] instanceof (window.AudioContext || window.webkitAudioContext), `Called emscripten_create_wasm_audio_worklet_node() on a context handle ${contextHandle} that is not an AudioContext, but of type ${typeof EmAudio[contextHandle]}`);
 #endif
-    options >>= 2;
 
     function readChannelCountArray(heapIndex, numOutputs) {
+      if (!heapIndex) return void 0;
```
sbc100: What does `return void 0` mean here?     Maybe just `0`?

cwoffenden:
@sbc100 Any chance of a review of this? It's a shorter replacement for the earlier one, discussed here way back when:

https://github.com/emscripten-core/emscripten/pull/23508#discussion_r1970467511

src/audio_worklet.js
```diff
@@ -25,8 +25,8 @@ function createWasmAudioWorkletProcessor(audioParams) {
       assert(opts.callback)
       assert(opts.samplesPerChannel)
 #endif
-      this.callback = getWasmTableEntry(opts.callback);
-      this.userData = opts.userData;
+      this.callback = getWasmTableEntry({{{ toIndexType("opts.callback") }}});
+      this.userData = {{{ toIndexType("opts.userData") }}};
```
sbc100: This second line doesn't look quite right to me.

userData I assume is a pointer, which means it should be Number and not a bigint (it should be converted to a Number when it first arrived on the JS side).   

If it needs to be converted back to a pointer (i64) when `this.callback` then you should be using `to64` at the callback call side, and not here.

`toIndexType` should only be used for places where you are direclty accessing the table, or the memory bounds.   

Can you test with `-sMEMORY64=2` as well perhaps to catch these cases where there is difference between `to64` and `toIndexType`?

src/lib/libwebaudio.js
```diff
@@ -262,18 +263,20 @@ let LibraryWebAudio = {
...
+      heapIndex = {{{ getHeapOffset('heapIndex', 'i32') }}};
       let channelCounts = [];
       while (numOutputs--) channelCounts.push(HEAPU32[heapIndex++]);
       return channelCounts;
     }
 
+    let optionsOutputs = options ? {{{ makeGetValue('options', C_STRUCTS.EmscriptenAudioWorkletNodeCreateOptions.numberOfOutputs, 'i32') }}} : 0;
```
sbc100: If we are using `let` then I guess we can/should use `const` here too?

Actually I'd be tempted to convert to using `var` instead since it minifies better.

sbc100:
(Probably not worth adding a `-sMEMORY64=2` mode actually.. although maybe test locally with it?

src/lib/libwebaudio.js
cwoffenden: Happily - I just followed the original style.

src/lib/libwebaudio.js
cwoffenden: It's replicating what the original code did:

https://github.com/emscripten-core/emscripten/blob/bf6eadf48cbc10be215de78fd5b67683be1a4a70/src/lib/libwebaudio.js#L276

This code is littered with the same expression (a shorter version of `undefined`, though I expect Closure replaces these?).

src/audio_worklet.js
cwoffenden: Yes, it's to convert it back to a pointer. It's constant, so I removed the need to keep doing the conversion.

I can test with `-sMEMORY64=2` and put the conversions near the call.

src/lib/libwebaudio.js
cwoffenden: The original was all `let` so I left it as-is (apart from the one that needed changing because of an issue with Closure). I can change them to `var`.

src/audio_worklet.js
cwoffenden: Tested manually with `-sMEMORY64=2` (and `-sMEMORY64=1`) with most of the conversions now removed.

cwoffenden:
> (Probably not worth adding a `-sMEMORY64=2` mode actually.. although maybe test locally with it?

Tested locally, which helped get this down to a minimum.

src/audio_worklet.js
```diff
@@ -65,56 +65,66 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
-      outputDataPtr = (dataPtr += numOutputs * {{{ C_STRUCTS.AudioSampleFrame.__size__ }}}) >> 2;
+      k = {{{ getHeapOffset('outputsPtr', 'u32') }}};
+      outputDataPtr = (dataPtr += numOutputs * {{{ C_STRUCTS.AudioSampleFrame.__size__ }}});
       for (i of outputList) {
         // Write the AudioSampleFrame struct instance
         HEAPU32[k + {{{ C_STRUCTS.AudioSampleFrame.numberOfChannels / 4 }}}] = i.length;
         HEAPU32[k + {{{ C_STRUCTS.AudioSampleFrame.samplesPerChannel / 4 }}}] = this.samplesPerChannel;
         HEAPU32[k + {{{ C_STRUCTS.AudioSampleFrame.data / 4 }}}] = dataPtr;
+#if MEMORY64
+        HEAPU32[k + {{{ C_STRUCTS.AudioSampleFrame.data / 4 + 1 }}}] = dataPtr / 0x100000000;
```
sbc100: I don't love this special case with the division.

How about this for the memory64 case: `makeSetValue(k * 4, C_STRUCTS.AudioSampleFrame.data, dataPtr. '*')`

cwoffenden: I can change everything to `makeSetValue` then we get assertions. I couldn’t previously because this was a separate file without the ability to run the JS tools.

sbc100: I think one reason to avoid `makeSetValue` is that it doesn't work well with pre-computed shifted values like this.. so I think there are cases where it can be a regression?  (Unless I'm misunderstanding).

If that is that case then maybe we should leave a wholesale migration to `makeGetValue` as a followup?

cwoffenden: The values end up getting shuffled to and from precomputed values in places (or offsets applied). What about like this:

https://github.com/emscripten-core/emscripten/blob/bced635a5fcd0c0e82146dbb39c114e8df2f2393/src/audio_worklet.js#L72

I only converted one section to do this as a test. It works as before but is written compactly. I can either revert in the morning, continue and test, or do this as a separate PR (I'd like to see the generated code difference, but it's past the end of the day here).

sbc100: Yeah, that is exactly what I mean.  In the new code the `>> 2` happens on each line instead of just once.  

Maybe its not and issue since the engine can optimize it.. but I'm not 100% sure.

cwoffenden: It's 5 bytes difference with this change to the JS (3 bytes of that was switching to explicit u32 types). And there are little savings to be had in other places where there are muls followed by shifts.

We're swapping plusses for shifts, which is all the same in the end, and we get the benefit of `ASSERTIONS`.

sbc100: OK, if you are convinced its not a perf regression then lets go with makeSetValue/makeGetValue everywhere.  I think you might be the person most concerned about the perf of this code anyway.

cwoffenden: Moving to `makeSetValue` after Closure changes this:
```.js
for (l of e) {
    L[g] = l.length;
    L[g + 1] = this.s;
    L[g + 2] = n;
}
```
Into this:
```.js
for (l of e) {
    L[g >> 2] = l.length;
    L[g + 4 >> 2] = this.s;
    L[g + 8 >> 2] = n;
}
```
Personally I can live with that since this code here is run probably once per audio callback , so 375x per second, compared with this, which I want to change next, getting called 96'000x per sec:

https://github.com/emscripten-core/emscripten/blob/052cd1a081d2194c1db4634e5f582e90e80f8750/src/audio_worklet.js#L121

src/audio_worklet.js
```diff
@@ -168,7 +169,7 @@ class BootstrapMessages extends AudioWorkletProcessor {
         //
         // '_wsc' is short for 'wasm call', using an identifier that will never
         // conflict with user messages
-        messagePort.postMessage({'_wsc': d.callback, args: [d.contextHandle, 1/*EM_TRUE*/, d.userData] });
+        messagePort.postMessage({'_wsc': d.callback, args: [d.contextHandle, 1/*EM_TRUE*/, {{{ to64('d.userData') }}}] });
```
sbc100, line 176: Why is this needed?  I would hope that userData can be sent over as JS pointer (AKA a number) and never a bigint.

src/audio_worklet.js
```diff
@@ -65,56 +65,57 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
         // Marshal the audio parameters array
-        HEAPF32.set(paramArray, dataPtr>>2);
-        dataPtr += paramArray.length*4;
+        HEAPF32.set(paramArray, {{{ getHeapOffset('dataPtr', 'float') }}});
+        dataPtr += paramArray.length * {{{ getNativeTypeSize('float') }}};
       }
 
       // Call out to Wasm callback to perform audio processing
-      if (didProduceAudio = this.callback(numInputs, inputsPtr, numOutputs, outputsPtr, numParams, paramsPtr, this.userData)) {
+      if (didProduceAudio = this.callback(numInputs, {{{ to64('inputsPtr') }}}, numOutputs, {{{ to64('outputsPtr') }}}, numParams, {{{ to64('paramsPtr') }}}, {{{ to64('this.userData') }}})) {
```
sbc100: Can we use the the `makeDynCall` mechanism to avoid this manual argument munging? 

Maybe that way we can completely avoid the need for the `to64` stuff here?

src/audio_worklet.js
cwoffenden: Without that conversion it will error:
```
Uncaught TypeError: Cannot convert 0 to a BigInt
    at MessagePort._EmAudioDispatchProcessorCallback (index.js:1:40067)
_EmAudioDispatchProcessorCallback	@	index.js:1
```
It either needs the conversion as above or where the function is then called:

https://github.com/emscripten-core/emscripten/blob/052cd1a081d2194c1db4634e5f582e90e80f8750/src/lib/libwebaudio.js#L215

src/audio_worklet.js
cwoffenden: I'll take a look (I have recollection that I tried it already, since it makes the most sense).

cwoffenden: It was to avoid doing this each audio frame:
```
getWasmTableEntry(this.callback).call(null, a1, BigInt(a2), a3, BigInt(a4), a5, BigInt(a6), BigInt(a7)))(numInputs, inputsPtr, numOutputs, outputsPtr, numParams, paramsPtr, this.userData)
```

cwoffenden: Which I can, by doing:
```
this.callback = {{{ makeDynCall('bipipipp', 'opts.callback') }}};
```
The callback is immutable.

src/audio_worklet.js
sbc100: In general it should be converted then the function is called, but maybe its not possible here since you don't know the function signature at the point of the call?

src/audio_worklet.js
sbc100: sgtm

src/audio_worklet.js
cwoffenden: The `wasmCall` above is used by both the message and by these collection of functions:

https://github.com/emscripten-core/emscripten/blob/052cd1a081d2194c1db4634e5f582e90e80f8750/src/lib/libwebaudio.js#L326

So exactly that, we don't know the signature, and this is used to broker all the different types of calls:

https://github.com/emscripten-core/emscripten/blob/052cd1a081d2194c1db4634e5f582e90e80f8750/src/audio_worklet.js#L173

cwoffenden: Left as-is and made a note.

src/audio_worklet.js
cwoffenden: Done.

src/audio_worklet.js
sbc100: Ok, can we add a comment then?  Something like "Convert pointer argments here since we don't have enough information available at the actual callsite"

cwoffenden: Done.

src/audio_worklet.js
```diff
@@ -25,7 +25,7 @@ function createWasmAudioWorkletProcessor(audioParams) {
       assert(opts.callback)
       assert(opts.samplesPerChannel)
 #endif
-      this.callback = getWasmTableEntry(opts.callback);
+      this.callback = {{{ makeDynCall('bipipipp', 'opts.callback') }}};
```
sbc100: Hmm.. i don't think I've seen `b` before?  Is that a thing?

sbc100 (approved):
Thanks!

src/audio_worklet.js
cwoffenden: Is it not `bool`? I took a guess and it didn’t seem to fail… I can change it to `i`.

sbc100: Nope, we only have letters for wasm types, not all types.

cwoffenden:
I’ll follow up with a bit of a tidy once this lands, just to keep the changes from sprawling. There’s a case also to write a test for when constructing the AW with options.

PR #17849, opened 2022-09-14
Use Node's path helpers under NODERAWFS

kleisauke:
Ensures that absolute paths are handled correctly on Windows as well. Using these helpers should be safe, since binaries linked with `-sNODERAWFS` can only be used on Node.js, see:
https://github.com/emscripten-core/emscripten/blob/e98554fd68b0bd16af151d9401c7eda004a13a2b/src/library_noderawfs.js#L28

Context: https://github.com/emscripten-core/emscripten/pull/16296#issuecomment-1079979886.

Resolves: #17360.
Resolves: #17819.
Supersedes: #17821.

(If preferred, I can split the last ~3~2 non-functional changes)

PR #24733, opened 2025-07-17
[WasmFS] Fix absolute path access under NODERAWFS

### LINKED ISSUE #24830: WasmFS NodeRawFS is not consistent with NodeRawFS (non-WasmFS version)

```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.11 (d3432e7a2a6585331c94b041973fb16c9f332ce4)
clang version 21.0.0git (https:/github.com/llvm/llvm-project 0f0079c29da4b4d5bbd43dced1db9ad6c6d11008)
Thread model: posix
```

Consider this example:
```cpp
/* test.cc */
#include <stdio.h>
#include <limits.h>
#include <unistd.h>

int main(int argc, char** argv) {
  char buf[PATH_MAX + 1];
  ssize_t len = readlink("/proc/self/exe", buf, sizeof(buf));
  if (len > PATH_MAX) {
    return 2;
  }
  printf("self exe: %s", buf);
  fflush(stdout);
  return 0;
}
```

```
% emcc -sEXIT_RUNTIME -sNODERAWFS test.cc && node a.out.js
self exe: /home/trybka/.nvm/versions/node/v22.13.1/bin/node
```
vs. 

```
% emcc -sEXIT_RUNTIME -sNODERAWFS -sWASMFS test.cc && node a.out.js
self exe: 
```

This is because the "root" of the VFS for NodeRawFS should be `/`, but when using WasmFS it is `.` (I assume, via https://github.com/emscripten-core/emscripten/blob/e5c41dbbc93b3a4ab3e0c30935223865facfdbcb/system/lib/wasmfs/backends/noderawfs_root.cpp#L11)

If I change the `.` to `/` then it works the same in both cases.

kleisauke:
Resolves: #24830.
Supersedes: #24833.


kleisauke:
Marking as ready for review with a few caveats:
- ~`wasmfs.test_readdir_rawfs` appears to be failing.~ **Edit:** see comment below.
- ~It also fails in the wasm-vips Windows CI when writing to `C:\Users\RUNNER~1\AppData\Local\Temp`, see:
  https://github.com/kleisauke/wasm-vips/actions/runs/23491851397/job/68365978350
  (this was workarounded in commit https://github.com/kleisauke/wasm-vips/commit/52f2d3cbea1574086a2dd05e8f66fc38e39314fd)~ **Edit:** see comment below.
- Code size expectations need to be rebaselined.
- ~Some changes could be split into separate PRs.~ **Edit:** done.

kleisauke:
The Windows CI issue was caused due to the temporary directory being located on `C:\` while the current working directory was on `D:\`, which broke WasmFS's single-root (`/`) assumption.

This has been fixed in commit d364aeb69f37ab63f93343922eae0d508f280177, which also appears to resolve the test failure in `wasmfs.test_readdir_rawfs`.

system/lib/wasmfs/backends/node_backend.cpp
```diff
@@ -180,6 +180,9 @@ class NodeDirectory : public Directory {
 
 private:
   std::string getChildPath(const std::string& name) {
+    if (state.path.empty()) {
+      return name;
+    }
```
tlively, line 190: Can you add a comment up on the definition of `path` saying what it means for a path to be empty?

system/lib/wasmfs/paths.cpp
```diff
@@ -66,12 +66,21 @@ ParsedParent doParseParent(std::string_view path,
...
+    return -ENOENT;
+  }
+
+  auto root = wasmFS.getRootDirectory();
+
+  // If the root backend is not virtualized, we assume there is only a single
+  // backend, making parent traversal unnecessary. Simply return the current
+  // directory along with the full path.
+  if (!root->getBackend()->isVirtualized()) {
+    return {std::make_pair(std::move(curr), path)};
```
tlively: The whole point of virtualization is that a virtual backend looks the same as a "real" backend. Can we get away without this part of the change?

system/lib/wasmfs/wasmfs.cpp
```diff
@@ -106,8 +106,15 @@ std::shared_ptr<Directory> WasmFS::initRootDirectory() {
   // The root directory is its own parent.
   lockedRoot.setParent(rootDirectory);
 
+  // For non-virtualized backends (e.g. NODERAWFS), the root filesystem is
+  // externally managed, so we skip creating standard directories like /dev/
+  // and /tmp/.
+  if (!rootBackend->isVirtualized()) {
+    return rootDirectory;
+  }
```
tlively: Let's use a more precise overridable hook here, e.g. `shouldMountSpecialFiles`, rather than checking whether the backend is virtual.

src/lib/libwasmfs_node.js
tlively, line 1: Are all the changes here just cleanups that are not directly related to the absolute path access problem?

src/lib/libwasmfs_node.js
```diff
@@ -5,7 +5,27 @@
  */
 
 addToLibrary({
-  $wasmfsNodeIsWindows: !!process.platform.match(/^win/),
+#if !ENVIRONMENT_MAY_BE_NODE
```
tlively: What's the goal of this change?

system/lib/wasmfs/paths.cpp
kleisauke: IIRC, this change was related to `-sNODERAWFS` on Windows, which doesn't have a single-root directory

I initially attempted a different fix (see e.g. the changes before commit d364aeb69f37ab63f93343922eae0d508f280177), but it failed when the current working directory and the absolute path to be accessed were on different drives, see comment https://github.com/emscripten-core/emscripten/pull/24733#issuecomment-4127788414.

src/lib/libwasmfs_node.js
kleisauke: This change was to allow a single Wasm binary to run both on Node.js and on the web (if you build with `-sWASMFS -sNODERAWFS` the Wasm binary is restricted to Node.js only).

I reverted this (and its test case) with commit ff4bba22a66cd470c1d9f200b315de849bc66d20. A PR will be opened for this later.

src/lib/libwasmfs_node.js
kleisauke: Yup, I've just separated these changes into individual PRs.

system/lib/wasmfs/wasmfs.cpp
kleisauke: I've just separated this change to PR #26607.

system/lib/wasmfs/backends/node_backend.cpp
kleisauke: Added a couple of comments with f04892c645351fe01e17ca6d66f0f1145a472dc0.

system/lib/wasmfs/paths.cpp
kleisauke: I renamed this `isVirtualized()` check with a `requiresPathTraversal()` check via commit cb9bb3ece93950ee59d46ace904a3eca263016aa.

src/lib/libwasmfs_node.js
kleisauke: ... I just opened PR #26608 for this change.

system/lib/wasmfs/backends/node_backend.cpp
```diff
@@ -296,6 +304,10 @@ class NodeBackend : public Backend {
   std::shared_ptr<Symlink> createSymlink(std::string target) override {
     WASMFS_UNREACHABLE("TODO: implement NodeBackend::createSymlink");
   }
+
+  virtual bool requiresPathTraversal() override {
```
tlively: How about calling this `usesRawPaths`? (And flipping the result, of course.) I think that's slightly easier to understand than `requiresPathTraversal`, though it's not perfect.

system/lib/wasmfs/paths.cpp
```diff
@@ -66,12 +66,21 @@ ParsedParent doParseParent(std::string_view path,
...
+    return -ENOENT;
+  }
+
+  auto root = wasmFS.getRootDirectory();
+
+  // If the root backend does not require path traversal, it can operate on the
+  // full path directly. In that case, skip WasmFS path parsing and pass the
+  // path through unchanged.
+  if (!root->getBackend()->requiresPathTraversal()) {
+    return {std::make_pair(std::move(curr), path)};
```
tlively, line 76: Is this going to cause WasmFS to model the underlying file system as completely flat, so that there is a single mount directory directly containing every file ever opened, no matter how deep that file is in the underlying directory structure?

If so, is that going to cause problems with reading directory entries?

tlively (commented):
Thanks for splitting this up. It's much more focused and easier to review now. Mostly LGTM, with some bikeshedding and a correctness concern.

system/lib/wasmfs/backends/node_backend.cpp
kleisauke: I think `usesRawPaths` it's too implementation-specific. How about `requiresPathResolution`? See commit e21bc63820ff7f89947102450879e78301381fd3.

system/lib/wasmfs/paths.cpp
kleisauke: Indeed, this is going to break in such situations as it implicitly assumes that the entire file system behaves like the root backend.

Commit e21bc63820ff7f89947102450879e78301381fd3 changes this to `curr->getBackend()` instead, which allows:
```console
$ tree / -L 2 --dirsfirst
/ (MEMFS)
├── dev
│   ├── null -> SpecialFiles::getNull()
│   ├── random -> SpecialFiles::getRandom()
│   ├── stderr -> SpecialFiles::getStderr()
│   ├── stdin -> SpecialFiles::getStdin()
│   ├── stdout -> SpecialFiles::getStdout()
│   └── urandom -> SpecialFiles::getURandom()
├── tmp
└── host (NODERAWFS)
```
(i.e. the Node filesystem is mounted under `/host`, while the `wasmfs_create_root_dir()` hook is not overriden)

kleisauke:
Marking as draft due to dependency on PR #26607 (i.e. the skip for `other.test_fs_dev_random` needs to be removed in this PR).

system/lib/wasmfs/paths.cpp
kleisauke: ... actually, that's not a great example, since it already worked without issues (AFAIK). I suspect it would have failed previously with the following layout:
```console
$ tree / -L 3 --dirsfirst
/ (NODERAWFS)
└── memory (MEMFS)
    ├── dev
    │   ├── null -> SpecialFiles::getNull()
    │   ├── random -> SpecialFiles::getRandom()
    │   ├── stderr -> SpecialFiles::getStderr()
    │   ├── stdin -> SpecialFiles::getStdin()
    │   ├── stdout -> SpecialFiles::getStdout()
    │   └── urandom -> SpecialFiles::getURandom()
    └── tmp
```

So, the `wasmfs_create_root_dir()` hook is overridden via `-sNODERAWFS`, while `MEMFS` is mounted at `/memory`. Though, `/memory/dev` and `/memory/tmp` aren't present, but that's an issue for PR #26607.

I suppose that we're missing some test coverage for this.

system/lib/wasmfs/backends/node_backend.cpp
tlively: That sounds good to me.

kleisauke:
Looking at the `chdir()` and `wasmfs_root_override.cpp‎` workaround in PR https://github.com/dotnet/runtime/pull/130141, I think .NET could also benefit from this.

/cc @pavelsavara FYI.

pavelsavara:
- when using NODERAWFS, could also default virtual cwd match the host cwd ?
- we also have hack for running on windows. Especially `D:\runtime` instead of `C:\runtime` was not working well. This is with legacy VFS too.

https://github.com/dotnet/runtime/blob/aa7a6fdc5db9aecfeba30881a9dae2fc0cce0f59/src/coreclr/hosts/corerun/wasm/libCorerun.js#L262-L270

Sorry if that's off topic here.

kleisauke:
@pavelsavara No worries, I'm happy to answer these questions here.

> * when using NODERAWFS, could also default virtual cwd match the host cwd ?

The current working directory will be whatever the OS working directory was when you started your Node.js program. See for example:

<details>
  <summary>Details</summary>

`test.c`:
```c
#include <stdio.h>
#include <dirent.h>

int main(void) {
    DIR *d = opendir(".");

    struct dirent *ent;
    while ((ent = readdir(d)) != NULL)
        puts(ent->d_name);

    closedir(d);
    return 0;
}
```

```console
$ emcc test.c -sWASMFS -sNODERAWFS
$ mkdir work
$ cd work
$ touch NEW_FILE
$ node ../a.out.js
.
..
NEW_FILE
```
</details>

Though, I just noticed that `getcwd()` is not aware of this under `WASMFS` mode, so that still needs to be fixed. See:

<details>
  <summary>Details</summary>

`test.c`:
```c
#include <stdio.h>
#include <unistd.h>
#include <limits.h>

int main(void) {
    char cwd[PATH_MAX];
    puts(getcwd(cwd, sizeof(cwd));
    return 0;
}
```

```console
$ echo $PWD
/home/kleisauke/emscripten
$ emcc test.c -sWASMFS -sNODERAWFS
$ node a.out.js
/
$ emcc test.c -sNODERAWFS
$ node a.out.js
/home/kleisauke/emscripten
```
</details>

> * we also have hack for running on windows. Especially `D:\runtime` instead of `C:\runtime` was not working well. This is with legacy VFS too.

I think this hack can be removed after PR #17849 landed.

kleisauke:
Abandoning due to inactivity. I'm also not particularly happy with this approach (although I can't think of a better solution at the moment).

As noted in https://github.com/emscripten-core/emscripten/pull/27319#issuecomment-4948119851, we should probally also disable dcache under NODERAWFS.
https://github.com/emscripten-core/emscripten/blob/d10aa3af1bbc06af6c9ead1cd418237d09ca5353/system/lib/wasmfs/file.h#L248-L264

(That said, I'm not going to take the lead on this, as my spare time is limited at the moment)

PR #24734, opened 2025-07-17
[AUDIO_WORKLET] Wait for Wasm instance to start using audio worklet.

brendandahl:
This fixes intermittent test failures where the AudioWorklet was used
before the Wasm module was loaded.


src/audio_worklet.js
```diff
@@ -170,6 +172,8 @@ class BootstrapMessages extends AudioWorkletProcessor {
         // conflict with user messages
         messagePort.postMessage({'_wsc': d.callback, args: [d.contextHandle, 1/*EM_TRUE*/, d.userData] });
       } else if (d['_wsc']) {
+        // Wait for the Wasm module to be instantiated.
+        await readyPromise;
```
sbc100: Didn't we already do this above?

src/lib/libcore.js
```diff
@@ -117,7 +117,7 @@ addToLibrary({
     // if exit() was called explicitly, warn the user if the runtime isn't actually being shut down
     if (keepRuntimeAlive() && !implicit) {
       var msg = `program exited (with status: ${status}), but keepRuntimeAlive() is set (counter=${runtimeKeepaliveCounter}) due to an async operation, so halting execution but not exiting the runtime or preventing further async execution (you can use emscripten_force_exit, if you want to force a true shutdown)`;
-#if MODULARIZE
+#if USE_READY_PROMISE
       readyPromiseReject?.(msg);
 #endif // MODULARIZE
```
sbc100: Comment needs updating here (just delete it I guess)

src/postamble_modularize.js
```diff
@@ -11,10 +11,7 @@ if (runtimeInitialized)  {
   moduleRtn = Module;
 } else {
   // Set up the promise that indicates the Module is initialized
-  moduleRtn = new Promise((resolve, reject) => {
-    readyPromiseResolve = resolve;
-    readyPromiseReject = reject;
-  });
+  moduleRtn = readyPromise = createReadyPromise();
```
sbc100, line 14: Can we just do `moduleRtn = readyPromise` here and unconditionally call `createReadyPromise` in `runtime_common.js
`?

src/postamble.js
```diff
@@ -139,7 +139,7 @@ function run() {
 
 #if PTHREADS || WASM_WORKERS
   if ({{{ ENVIRONMENT_IS_WORKER_THREAD() }}}) {
-#if MODULARIZE
+#if USE_READY_PROMISE
     readyPromiseResolve?.(Module);
```
sbc100, line 143: I think you can perhaps remove the question marks here too. i.e. we can make USE_READY_PROMISE imply that readyPromiseResolve exists?

src/postamble_modularize.js
brendandahl: I think if I do that, I'll have to undo the work you did in https://github.com/emscripten-core/emscripten/pull/24418/files and add back a try/catch.

sbc100:
I'm a little wary about extending the use of the module promise outside of the existing usage in `-sMODULARIZE`.     I wonder if there is some other existing signal we could to use here?    Perhaps the `dependenciesFulfilled` callback?   I'm not sure if like that any better though :-/

brendandahl:
That won't work with minimal_runtime. I suppose I could use one of the other hooks like ATINITS or ATPOSTCTORS that works in regular and minimal runtime.

brendandahl:
Those hooks don't work for wasm workers, but I think i've found another way. With non minimal runtime and wasm workers a message handler is installed that queues all the messages. The cleanup code for this was being installed, but not the setup code.

brendandahl:
Closing in favor of #24778

Issue #24504, opened 2025-06-05
Using standalone file_packager


lkwinta:
I am working on porting Geant 4, physics simulator into Web Assembly. I will be downloading many huge datafiles so I figured out that manually creating emscripten datafiles with file packager might be great way to handle this problem.

 Unfortunately, I have got into problem that I can't fix. Unfortunately, my project is quite modern and the only way I can load WASM is with modularize option - I can't pass global var as pre module. 

Here is my data structure. I am using react with typescriprt. All data files and wasm binaries are inside public directory, hence I have to overwrite `locateFile` function. WASM processig will be done inside web worker.

![Image](https://github.com/user-attachments/assets/5f7fd8c4-533b-46fb-a735-431167d02879)

For the main Wasm import I am using such syntax:
```js
import {default as MainModule} from "./geant4_wasm/geant4_wasm.js";

var preModule = {
 ...
}

var mod = MainModule(preModule);
```

I was trying to load `file_packager's` js. output using importScripts function inside web worker with global variable syntax but for some reason, while using webpack it doesn't work - global variable is not transfered to the imported file - i tried logging inside preload files, it loads but without initialization preModule.
```js
var preModule = {
 ....
}

var Module = preModule;
importScripts("/preload_G4EMLOW8.6.1.js"); // file loads but with empty Module
// I have to access public directory in this case since unimported files are not packed inside react package
```

I kindly ask for adding similar modularize option for file_packager as for emcc or for suggestion how could i solve my problem.

lkwinta:
PR: https://github.com/emscripten-core/emscripten/pull/24737

PR #24737, opened 2025-07-18
[file_packager.py] Add --export-es6 to file_packager.py

lkwinta:
I have added modularize option flag to `file_packager.py` script, like in main js generator 
I think it will be helpful since in modern world of react it is inconvinent to add generated files as `<script/>` tags.
This pull request introduces a new `--export-es6` option to the `file_packager.py` tool, enabling the generation of modularized JavaScript output. It enables to import generated JS loading stub into ES6 environment. The main module which will take effect of script execution is now passed as script argument. 

Use scenario:
```
emcc -s MODULARIZE -s EXPORT_NAME=MainModule ...
file_packager.py --export-es6 ...
```

```js
import MainModule from 'main_module_wasm'

// two datafiles that will be loaded at runtime with module
import { default as loadDataFile_specialName } from 'data_file_preload' 
import loadDataFile from `data_file_preload2`

var preMod = {
   // i. e. locateFile
}

var mod = MainModule(preMod)
mod.then(async (module) =>  {
    await loadDataFile(module);
    await loadDataFile_specialName(module;

    // module has now loaded data file in VFS
});
```

Fixes: #24504

### DIFF
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -758,6 +758,13 @@ def generate_js(data_target, data_files, metadata):
   if options.has_embedded and not options.obj_output:
     diagnostics.warn('--obj-output is recommended when using --embed.  This outputs an object file for linking directly into your application is more efficient than JS encoding')
 
+  catch_handler = ''
+  if options.export_es6:
+    catch_handler += '''
+        .catch((error) => {
+          loadDataReject(error);
+        })'''
+
   for counter, file_ in enumerate(data_files):
     filename = file_.dstpath
     dirname = os.path.dirname(filename)
@@ -1056,13 +1063,6 @@ def generate_js(data_target, data_files, metadata):
 
     code += '''
       Module['preloadResults'] ??= {};\n'''
-
-  catch_handler = ''
-  if options.export_es6:
-    catch_handler += '''
-        .catch((error) => {
-          loadDataReject(error);
-        })'''
 
     if options.use_preload_cache:
       code += '''


grzanka:
@sbc100 could you please take a look ?

tools/file_packager.py
```diff
@@ -1148,9 +1165,26 @@ def generate_js(data_target, data_files, metadata):
...
   })();\n'''
 
+  if options.modularize and not options.from_emcc:
+    ret +=  '''
+  (() => {
+  var real_createModule = createModule;
+
+  createModule = function (moduleArg) {
+    if (new.target) throw new Error("createModule() should not be called with `new createModule()`");
+    return real_createModule(moduleArg);
```
sbc100: We got rid of this thing in #23960.. is there some reason to bring it back here?

tools/file_packager.py
```diff
@@ -1148,9 +1165,26 @@ def generate_js(data_target, data_files, metadata):
...
+    ret += '''
+    });'''
+
   ret += '''
   })();\n'''
 
+  if options.modularize and not options.from_emcc:
+    ret +=  '''
+  (() => {
+  var real_createModule = createModule;
```
sbc100: Does this assume that createModule is the string passed to `-sEXPORT_NAME` ?

sbc100:
Can you explain a little more in the PR description exactly what `--modularize` is doing here?  It looks like its generating an ES6 module from the data files, is that right?  How is the generated module supposed to interact with the program into which the files are to be loaded?

tools/file_packager.py
lkwinta: We would like to use file packager standalone to generate multiple data files loaded dynamically instead of packing everything into large one. I couldn't get preloading module working in react environment using global variables - it seems to be working only when using <script> tags.

Same was with the main module, only after enabling -sMODULARIZE I was able to preload module with desired functionality.

tools/file_packager.py
lkwinta: I was thinking about integrating that functionality like in emcc but I decided that I would like to keep it simple for now. It shouldn't be a problem I think... If it is ES6 import it could be renamed anyway.

lkwinta:
> Can you explain a little more in the PR description exactly what `--modularize` is doing here? It looks like its generating an ES6 module from the data files, is that right? How is the generated module supposed to interact with the program into which the files are to be loaded?

I have added some use case description.

As described in the related issue we are just trying to make the dynamic loading of large necessary dependencies inside web workers run from react application. Together with XHR lazy loading this seems to be only valid solution to our problem that might solve our problem.

tools/file_packager.py
```diff
@@ -616,13 +620,26 @@ def generate_js(data_target, data_files, metadata):
   if options.from_emcc:
     ret = ''
   else:
-    ret = '''
+    if options.modularize:
+      ret = '''
+  var createModule = (() => {
```
sbc100: Perhaps this should be called `loadDataFile` line in your example?

tools/file_packager.py
```diff
@@ -616,13 +620,26 @@ def generate_js(data_target, data_files, metadata):
...
+
+  return (async function(moduleArg = {}) {
+    var Module = moduleArg;
```
sbc100: Do we need this extra inner function here or can we just do:
```
export default loadDataFile(Module) {
  ...
}
```

Also, I think the function doesn't need to be async because it hooks into the Module object using `addRunDependency` / `removeRunDependency`.  Although if the Module is already loaded and running that might not work, so we might need to change that too.

lkwinta: Function is async because I didn't touch inner code, i. e. web requests which return promises but I agree that it might need changing. 

The inner function prevents from using `new` keywoard, just like in `emcc` ES 6 export code

tools/file_packager.py
lkwinta: Maybe I should add a comment or add another import as a showcase. Indeed this is called, but since that is a default export the syntax to renaming is 
`import { default as loadDataFile } from 'data_file_preload'`
instead of
`import { createModule as loadDataFile } from 'data_file_preload'`

but maybe indeed we should integrate it with -s EXPORT_NAME

tools/file_packager.py
sbc100: I'm not sure we need the generated code here to look like the emscripten-generate program.   For example, the file loading code just take a single `Module` argument, I think, it doesn't need the whole moduleArg dictionary thing.

Also, the emcc output no longer contains the `new` keyword check.  See #23960.

lkwinta: Okay, my bad, I was basing on emsdk release which doesn't have those changs.

What do you suggest to pass inside that function? Only Module's FS and used stuff like locate file?

sbc100: I believe the generated code expected something called `Module` which is the whole module object?  But I could be wrong.

sbc100: You could also call it with your module args object and it looks like it will inject a `Module.preRun` in that case.  i.e. it works if its passed either module args or that actual module.

lkwinta: > I believe the generated code expected something called `Module` which is the whole module object? But I could be wrong.

Exactly, generated module want's to use modules function like `locateFile` and `FS_CreatePath`. Probably it messes with modules internal state while creating files in VFS.

You mean that I could just use it like this:

```js
var preMod = {...}

var mod = MainModule(preMod);
 
createModule(preMod); //function from fille packager generated code
```

Isn't it the same as calling it with `mod` arg since `MainModule(preMod)` takes `preMod` by reference and extends it with fields? Od that's not the case I think we can't use it like this because we need to fill in Module's (`mod`) virtual FS paths.

Original use case was:
```html
<script>
var preMod = { .... }
</script>
<script src="load_wasm"/> 
<script src="load_datafile"/>
<script>
// use Module to access wasm code
</script>
```

since it oprrates on the same global context IT passes whole Module dictionary.

In @Kripken's demo linked in emscripten docs, DOM manipulation is used to insert <script> tags at runtime to load datafile, but we can't use it because of react and WebWorkers.

tools/file_packager.py
lkwinta: I have changed that, now the function is named by `export_name` option

tools/file_packager.py
lkwinta: My bad, I didn't understood until now, you meant the warning about `new.target`. I have removed it from my code.

tools/file_packager.py
sbc100: I don't think you need to use EXPORT_NAME here.. can't you just always call the function `loadDataFile`?   Since it exported by default I don't even think it really matter what it is called.

Why not just?  

```
export default loadDataFile(Module) {
  ...
}
```

Also, as I think I said already I don't think need the inner function here do you?  And I don't think anything need to be async because we never return a promise.

lkwinta: Okay, now I see your point, I think that I have simplified it now.

I will take a look into those tests that are failing for some reason.

lkwinta:
@sbc100 what do you think about it now?

tools/file_packager.py
```diff
@@ -1153,7 +1171,11 @@ def generate_js(data_target, data_files, metadata):
     }
     loadPackage(%s);\n''' % json.dumps(metadata)
 
-  ret += '''
+  if options.modularize and not options.from_emcc:
```
sbc100: How about just doing `if options.modularize` and make `--from-emcc` and `--modularize` mutually exclusive (i.e. error if both are specified).

tools/file_packager.py
sbc100: How about just `export default function loadDataFile(Module) {`

(Argument should not be optional, and can just be called Module, I think).

tools/file_packager.py
lkwinta: Good point

tools/file_packager.py
```diff
@@ -621,13 +632,24 @@ def generate_js(data_target, data_files, metadata):
   if options.from_emcc:
     ret = ''
   else:
-    ret = '''
+    if options.modularize:
+      ret = '''
+  export default function loadDataFile(Module) {
+                            '''
```
sbc100: Maybe make this just a single line `ret = 'export default function loadDataFile(Module) {'`

tools/file_packager.py
```diff
@@ -1153,7 +1175,11 @@ def generate_js(data_target, data_files, metadata):
     }
     loadPackage(%s);\n''' % json.dumps(metadata)
 
-  ret += '''
+  if options.modularize:
+    ret += '''
```
sbc100, line 1176: How about just `ret += '\n}\n'`

sbc100: Can you add a comment here.  `// End the `loadDataFile` function`

sbc100 (commented):
Looks like we still need a test for this.

And might be worth adding to the changelog.

Otherwise lgtm.

tools/file_packager.py
lkwinta: Okay I will apply that, I am not really familiar with this multi line string formatting rules and I tried not to break already existing formatting

lkwinta:
> Looks like we still need a test for this.
> 
> And might be worth adding to the changelog.
> 
> Otherwise lgtm.

I have added two simple tests - from what I have seen `file_packager.py` tests aren't really extensive.

test/test_other.py
```diff
@@ -3929,6 +3947,21 @@ def test_file_packager_embed(self):
...
+    err = self.run_process([FILE_PACKAGER, 'test.data', '--modularize', '--preload', 'data.txt', '--js-output=data.js'], stderr=PIPE).stderr
+    self.assertEqual(MESSAGE, err)
+
+    generated_content = read_file('data.js')
+
+    expected_opening = "export default function loadDataFile(Module) {"
+    expected_closing = "\n};\n// END the loadDataFile function"
+
+    self.assertContained(expected_opening, generated_content)
+    self.assertContained(expected_closing, generated_content)
```
sbc100: Can you make this more like `test_file_packager_embed`... I think we should test that the resulting package actually works.

So I would so something like this:

1. Build an simple emscripten program with `-sMODULARIZE -sEXPORT_ES6`, just like the one in `test_file_packager_embed`.
2. Create a simple launch script that loads the module and creates a new instance, then passes that result to `loadDataFile`.

Thinking about this, that might not work because the main function will already have been run before we can call `loadDataFile`.

To work around this I would use preRun.  Something like this in your launcher script:

```
import loadDataFile from "./file_data.mjs";
import moduleFactory from "./emcc_program.mjs";

await moduleFactory({preRun: loadFileData});
console.log("done");
```

lkwinta: I think that still has issues because we have to pass module as arg to loadDataFile but preRun methods doesn't get any arg... Maybe this would work? I have decided to change it a little bit and used lembind to export function to be run in correct order.

sbc100: IIUC `preRun` callbacks (indeed all the callback) get the module passed as the first argument.

See https://github.com/emscripten-core/emscripten/blob/a2b1fddae62a74eff771f6ad599bd36a89f9b177/src/lib/libcore.js#L1815-L1816

lkwinta: shall I change that to use preRun or leave it like it is now

sbc100: I think preRun would be better.  Certainly involving embind seems like overkill here.

test/test_other.py
```diff
@@ -3788,21 +3788,34 @@ def check(text):
...
+    proc2 = self.run_process([FILE_PACKAGER, 'test.data', '--quiet', '--preload', 'data2.txt'], stderr=PIPE, stdout=PIPE)
+    check(proc2.stdout)
+    self.assertEqual(proc2.stderr, '')
+
+    self.assertTextDataIdentical(clean(proc.stdout), clean(proc2.stdout))
+
+    proc2 = self.run_process([FILE_PACKAGER, 'test.data', '--quiet', '--preload', 'data2.txt', '--modularize'], stderr=PIPE, stdout=PIPE)
+    check(proc2.stdout)
+    self.assertEqual(proc2.stderr, '')
+
```
sbc100: Maybe skip the changes to this test.. since it already quite a long test and its not clear what the duplication of these various sections is doing.

test/test_other.py
```diff
@@ -3929,6 +3947,47 @@ def test_file_packager_embed(self):
...
+    create_file('data.txt', 'hello data')
+    err = self.run_process([FILE_PACKAGER, 'test.data', '--modularize', '--preload', 'data.txt', '--js-output=dataFileLoader.mjs'], stderr=PIPE).stderr
+    self.assertEqual(MESSAGE, err)
+
+    create_file('test.cpp', '''
+    #include <stdio.h>
+    #include <emscripten/bind.h>
+
+    int test_fun() {
```
sbc100: You can just use `EMSCRIPTEN_KEEPALIVE` here to make this function callback from the outside rather then using embind.  Then you can do `mod._test_func()` from the JS side (node the leading underscore).

Either that or use the `preRun` approach.

lkwinta: okay, i will use KEEPALIVE

test/test_other.py
lkwinta: Should I duplicate the test - since we probably should check if with `--modularize` we don't break formatting

lkwinta:
@sbc100 I keep getting errors that `require` is not found even if I have added `@requires_node`  annotation, do you have any hint why that might be?

EDIT: Maybe it is because node doesn't allow require inside ES6 imported files? But I think that emcc's code also uses require even with ES6 mode...

ChangeLog.md
```diff
@@ -20,6 +20,8 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- `--modularize` flag was added to `file_packager.py` available when run
```
sbc100: I wonder if `--modularize` is the right name for this flag?   How about `--export-es6` or `--module` or `--esm`?   I'm not really sure what is best but in emscripten `-sMODULARIZE` does not imply ES6 modules .. so maybe confusing?

ChangeLog.md
```diff
@@ -20,6 +20,8 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- `--modularize` flag was added to `file_packager.py` available when run 
+  standalone, to enable ES6 imports of generated JavaScript code
```
sbc100: Can you add `(#24737)` to the end of this line.

sbc100 (commented):
LGTM % final comments.

ChangeLog.md
lkwinta: sure

ChangeLog.md
lkwinta: got it

test/test_other.py
```diff
@@ -3929,6 +3934,45 @@ def test_file_packager_embed(self):
...
+    }
+    ''')
+    self.run_process([EMCC, 'test.c', '-sFORCE_FILESYSTEM', '-sMODULARIZE', '-sEXPORT_ES6', '-o', 'moduleFile.js'])
+
+    create_file('run.js', '''
+    import loadDataFile from './dataFileLoader.js'
+    import {default as loadModule} from './moduleFile.js'
+
+    const sleep = (delay) => new Promise((resolve) => setTimeout(resolve, delay))
+    var module = loadModule();
```
sbc100: Can just do `var module = await loadModule()` avoiding the extra function below?

Maybe top level await is not available?   If so at least maybe just do `loadModule().then(...` to avoid having two things called `module` so close together.

test/test_other.py
```diff
@@ -3929,6 +3934,45 @@ def test_file_packager_embed(self):
...
+    module.then(async (module) => {
+      loadDataFile(module);
+      // sleep so we don't have to use monitorRunDependencies logic
+      await sleep(2000);
```
sbc100: This seems unfortunate.  Could `loadDataFile` not also return a promise perhaps?

lkwinta: It could for sure but I need permision to fix one thing:

```
 ret += '''
      function fetchRemotePackage(packageName, packageSize, callback, errback) {
        %(node_support_code)s
        fetch(packageName)
          .catch((cause) => Promise.reject(new Error(`Network Error: ${packageName}`, {cause}))) // If fetch fails, rewrite the error to include the failing URL & the cause.
          .then((response) => {
            if (!response.ok) {
              return Promise.reject(new Error(`${response.status}: ${response.url}`));
            }

            if (!response.body && response.arrayBuffer) { // If we're using the polyfill, readers won't be available...
              return response.arrayBuffer().then(callback);
            }

            const reader = response.body.getReader();
            const iterate = () => reader.read().then(handleChunk).catch((cause) => {
              return Promise.reject(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}));
            });
```

looking at the rest of the code, I think those `Promise.reject` statements are wrong and they should be replaced by `errback()` as for example in `fetchCachedPackage`

sbc100: I see yes, that does looks like it needs fixing.  Would you mind sending a separate PR for that?

sbc100: I wonder if we should instead re-write `fetchRemotePackage` to be and `async` function?  (Maybe we can do that after first fixing the minor bug you found).

lkwinta: I also found that some node related functions generate Promise reject... is that intentional?

lkwinta: I feel like those should call `handleError()` instead

```js
 function runMetaWithFS() {
    Module['addRunDependency']('dataFileLoader.js.metadata');
    var metadataUrl = Module['locateFile'] ? Module['locateFile']('dataFileLoader.js.metadata', '') : 'dataFileLoader.js.metadata';
    if (isNode) {
          require('fs').readFile(metadataUrl, 'utf8', (err, contents) => {
            if (err) {
              return Promise.reject(err);
            } else {
              loadPackage(JSON.parse(contents));
            }
          });
          return;
        }
    fetch(metadataUrl)
      .then((response) => {
        if (response.ok) {
          return response.json();
        }
        return Promise.reject(new Error(`${response.status}: ${response.url}`));
      })
      .then(loadPackage);
  }
```

edit: handleError is out of scope because it is local to loadPackage

lkwinta: > I see yes, that does looks like it needs fixing. Would you mind sending a separate PR for that?

https://github.com/emscripten-core/emscripten/pull/24871

lkwinta: > I wonder if we should instead re-write `fetchRemotePackage` to be and `async` function? (Maybe we can do that after first fixing the minor bug you found).

Maybe it would be correct way to do it but in fact it is async right now since we are not waiting for fetch, when fetch promise is called, `callback` is called which triggers rest of the code

ChangeLog.md
sbc100: We just added one new flag right?  Not two?  I think just add `--export-es6` should be good.

test/test_other.py
```diff
@@ -3945,6 +3950,46 @@ def test_file_packager_embed(self):
     output = self.run_js('a.out.js')
     self.assertContained('hello data', output)
 
+  def test_file_packager_standalone_modularize(self):
```
sbc100: Rename this test something like `test_file_packager_standalone_export_es6`?

test/test_other.py
```diff
@@ -3919,6 +3919,11 @@ def test_file_packager_returns_error_if_target_equal_to_jsoutput(self):
     err = self.expect_fail([FILE_PACKAGER, 'test.data', '--js-output=test.data'])
     self.assertContained(MESSAGE, err)
 
+  def test_file_packager_returns_error_if_emcc_and_modularize(self):
+    MESSAGE = 'error: Can\'t use modularize option together with --from-emcc since the code should be embedded within emcc\'s code'
+    err = self.expect_fail([FILE_PACKAGER, 'test.data', '--modularize', '--from-emcc'])
```
sbc100: `--modularize` is no longer a thing is it?

tools/file_packager.py
```diff
@@ -1153,7 +1183,14 @@ def generate_js(data_target, data_files, metadata):
     }
     loadPackage(%s);\n''' % json.dumps(metadata)
 
-  ret += '''
+  if options.export_es6:
+    ret += '''
+  return readyPromise;
+};
```
sbc100: No semicolon after `function` def.

tools/file_packager.py
```diff
@@ -621,13 +632,27 @@ def generate_js(data_target, data_files, metadata):
   if options.from_emcc:
     ret = ''
   else:
-    ret = '''
+    if options.export_es6:
+      ret = '''export default function loadDataFile(Module) {
+  var readyPromiseResolve, readyPromiseReject;
+  var readyPromise = new Promise((resolve, reject) => {
+    readyPromiseResolve = resolve;
+    readyPromiseReject = reject;
```
sbc100: Can you give these three variables different name so they don't match the ones use use in the emscripten itself.

Also, I think you don't need the `var readyPromise`.. you can just do:"

```
export default function loadDataFile(Module) {
    return new Promise((loadDataResolve, loadDataReject) => {
      ... (other generated code goes here)
    }
 }
 ```

Comment (lkwinta) [reply tools/file_packager.py]: I am also now getting this error... Which is really strange because the file is there...
obraz

I have tried on my local machine with npx serve and it just works

**Comment by lkwinta** [comment]
I manage to solve the issue, apperantly we do not have any tests with --no-node support? Is it possible to run test from web context at all to test if fetch works insteads of nodes method?

Now happy path works but promise is not rejected where it should be most likely - there is were I need that PR: #24871

**Comment by lkwinta** [comment]
> I manage to solve the issue, apperantly we do not have any tests with --no-node support? Is it possible to run test from web context at all to test if fetch works insteads of nodes method?
> 
> Now happy path works but promise is not rejected where it should be most likely - there is were I need that PR: #24871

maybe, I should use the same syntax as in emcc ? and just do the following

```
if (isNode) {
      const { createRequire } = await import('module');
      var require = createRequire(import.meta.url);
}
```

without es6 import and with mind that var inside if will be accessible outside ? @sbc100 what do you think?

File: test/test_other.py [hunk 19]
```diff
@@ -3945,6 +3950,43 @@ def test_file_packager_embed(self):
...
+    create_file('run.js', '''
+    import loadDataFile from './dataFileLoader.js'
+    import {default as loadModule} from './moduleFile.js'
+
+    loadModule().then((module) => {
+      loadDataFile(module)
+        .catch((cause) => Promise.reject(cause))
+        .then(() => {
+          module._test_fun();
+        }
```
sbc100: I don't think you need the `.catch` here.   Doing `return loadDataFile(..` should be enough to reject the promise if loadDataFile fails/rejects.

sbc100: Also, I think just using top-level-await might be more clear here:

```
var module = await loadModule();
await loadDataFile(module):
module._test_fun();
```

test/test_other.py
```diff
@@ -3945,6 +3950,43 @@ def test_file_packager_embed(self):
...
+      );
+    });
+    ''')
+
+    output = self.run_js('run.js')
+    self.assertContained('hello data', output)
```
sbc100: Maybe combine there two lines.

tools/file_packager.py
```diff
@@ -485,6 +491,11 @@ def main():  # noqa: C901, PLR0912, PLR0915
     diagnostics.error('TARGET should not be the same value of --js-output')
     return 1
 
+  if options.from_emcc and options.export_es6:
+    diagnostics.error('Can\'t use --export-es6 option together with --from-emcc since the code should be embedded '
+        'within emcc\'s code')
```
sbc100, line 496: I would just make this into a single line.  We allow longer-than-normal python lines IIRC

tools/file_packager.py
```diff
@@ -621,20 +632,36 @@ def generate_js(data_target, data_files, metadata):
...
 
   ret += '''
   Module['expectedDataFileDownloads'] ??= 0;
-  Module['expectedDataFileDownloads']++;
-  (() => {
+  Module['expectedDataFileDownloads']++;'''
+
+  if not options.export_es6:
+    ret += '''
+  (async () => {'''
```
sbc100: Why is `async` being added here for the non-export-es6 case?

tools/file_packager.py
```diff
@@ -1062,14 +1089,13 @@ def generate_js(data_target, data_files, metadata):
...
         // the rest of this function immediately.
-        fetchRemotePackage(REMOTE_PACKAGE_NAME, REMOTE_PACKAGE_SIZE).then((data) => {
-          if (fetchedCallback) {
-            fetchedCallback(data);
-            fetchedCallback = null;
-          } else {
-            fetched = data;
-          }
-        })
+        let packageData = await fetchRemotePackage(REMOTE_PACKAGE_NAME, REMOTE_PACKAGE_SIZE);
```
sbc100: See the comment above, we need to avoid `async` here since we don't want to suspend the function at this point

test/test_other.py
lkwinta: I will apply that later, I have changed some logic to be more universal and to allow catching thrown error, take a look please

tools/file_packager.py
```diff
@@ -1126,9 +1152,15 @@ def generate_js(data_target, data_files, metadata):
   else:
     ret += '''
     }
-    loadPackage(%s);\n''' % json.dumps(metadata)
+    await loadPackage(%s);\n''' % json.dumps(metadata)
```
sbc100: If this is the very end of the function then you can just do `return loadPackage` I think?

(doing `return` instead of `await` might also mean you don't need to make the outer function `async`).

test/test_other.py
lkwinta: I wonder, if I shouldn't add some throw error method to clear run dependencies - or maybe generate separate stub that allows to do that if we want ourselfs in external catch block

lkwinta: > I wonder, if I shouldn't add some throw error method to clear run dependencies - or maybe generate separate stub that allows to do that if we want ourselfs in external catch block

It could be other PR tho, I have just observed anoying behaviour that when it fails, main module constantly spams run dependencies message, being able to clean those to graceful escape error would be nice

sbc100: I'm not sure that removing the run dependency is the way to fix that.. since that would result in the module running.  If a dependency fails we want to block the running forever I think.. not just ignore it.   But I think the existing code might not do that :(

tools/file_packager.py
lkwinta: Because I wanted to wait for the loadPackage which was made async in one of yours PR chain

tools/file_packager.py
lkwinta: Oh, I didnt notice, that might complicate things a bit i think

tools/file_packager.py
```diff
@@ -1062,15 +1097,16 @@ def generate_js(data_target, data_files, metadata):
...
-      }\n'''
+        fetchRemotePackage(REMOTE_PACKAGE_NAME, REMOTE_PACKAGE_SIZE)
+          .then((data) => {
+            if (fetchedCallback) {
+              fetchedCallback(data);
+              fetchedCallback = null;
+            } else {
+              fetched = data;
+            }
+          })%s;
```
sbc100, line 1116: It looks like there is no current error handling here, for this fetch.  It looks like the code is trying to do a pre-fetch and maybe we can just ignore failures (like we currently do) for the pre-fetch?

tools/file_packager.py
```diff
@@ -1020,6 +1051,10 @@ def generate_js(data_target, data_files, metadata):
 
     code += '''
       Module['preloadResults'] ??= {};\n'''
+    catch_case = '''
+          .catch((error) => {
+            loadDataReject(error);
+          })'''
```
sbc100: How about doing:

```
catch_handler = ''
if export_es6:
  catch_handler = ''' ... '''
```

Then you don't need repeated `(catch_case if options.export_es6 else '')` below.

tools/file_packager.py
```diff
@@ -710,14 +741,14 @@ def generate_js(data_target, data_files, metadata):
...
-          this.requests[this.name] = null;
+          this.requests[this.name] = null;%s
         }
       };
 
       var files = metadata['files'];
       for (var i = 0; i < files.length; ++i) {
         new DataRequest(files[i]['start'], files[i]['end'], files[i]['audio'] || 0).open('GET', files[i]['filename']);
-      }\n''' % (create_preloaded if options.use_preload_plugins else create_data)
+      }\n''' % (create_preloaded if options.use_preload_plugins else create_data, ready_promise if options.export_es6 else '')
```
sbc100: How about:

```
finish_handler = create_preloaded if options.use_preload_plugins else create_data
if export_es6:
   finish_handler += '\nloadDataResolve();'
```

Then use `finish_handler` here in the string template.

tools/file_packager.py
sbc100: In that case this block could be reverted

sbc100 (commented):
lgtm % final comments

tools/file_packager.py
lkwinta: Huh, without this `catch` here in es6 mode, i can't handle error on my own, the whole js crashes because of 'unhandled error on promise'

Fine:
```
async function F1() {
    throw new Error();
}

F1().catch((err)=>console.log(err))
```

`Unhandled error in promise`:
```
async function F1() {
    async function F2(){
    throw new Error();
    }
    F2();
}

F1().catch((err)=>console.log(err))
```

sbc100: But isn't that just the existing behaviour?   There is not catch mechanism for this particular fetch and so errors here will always go the to the global error handlers.    I think that is the current/expected behaviour.   We could change that perhaps?

Given that there is not really any way to recover from such errors why not just let the global error handler handle these errors?  Or the global unhandledRejection handler.

lkwinta: I meant that when the throw is two or more level deep I cant catch the error even with top level try catch block which is an issue - most likely es6 mode will be used together with some frontend and we don't want to crash whole JS but rather give some information and maybe possibility to retry

lkwinta: And inside webworker we can't access window context to get access to `unhandledrejection` handler

sbc100: You can handle unhandled rejections using the worker global scope:

```
    // worker.js
    self.addEventListener('unhandledrejection', (event) => {
        console.error('Unhandled Promise Rejection in Worker:', event.reason);
    });
```

But I understand that it would be nice to correctly reject the promise here, so this part lgtm.

We can refactor later perhaps.

lkwinta: Sure

lkwinta:
So @sbc100 do we merge that?

sbc100:
Yup, LGTM.

I just landed #24917 so maybe worth rebasing/merging one more time.

lkwinta:
> Yup, LGTM.
> 
> I just landed #24917 so maybe worth rebasing/merging one more time.

done

PR #24744, opened 2025-07-20
[wasmfs] Expose wasmfs headers file for out-of-tree filesystems

martenrichter:
This moves wasmfs header files to the public include file directory, so that it is possible to develop wasmfs file systems, out of tree.
Open points:

- There are some prototypical classes of wasmfs, which may be beneficial for certain types of file systems. I did not include them, as I restricted the files to the bare minimum of abstract base classes.

- It would be good if out-of-tree filesystems could detect, if the wasmfs headers are in the public include directory and may have some kind of semantic versioning scheme, if the interface changes (e.g., to use the preprocessor). Do you think a global Emscripten macro would help?


sbc100:
For versioning we already have `emscripten/version.h` which looks like this:

```
/* Automatically generated by tools/system_libs.py */                                
#define __EMSCRIPTEN_major__ 4                                                       
#define __EMSCRIPTEN_minor__ 0                                                       
#define __EMSCRIPTEN_tiny__ 12    
```

system/include/emscripten/wasmfs_internal/backend.h
```diff
@@ -4,10 +4,11 @@
 // found in the LICENSE file.
 
 // This file defines the modular backend abstract class.
+// The declarations from this files should only be used by file system implementers
 
 #pragma once
 
-#include "file.h"
+#include <emscripten/wasmfs_internal/file.h>
```
sbc100: Can you leave these as-is, if the files remain in the same directory?

system/lib/wasmfs/virtual.h
```diff
@@ -15,7 +15,7 @@
 
 #pragma once
 
-#include "file.h"
+#include <emscripten/wasmfs_internal/file.h>
```
sbc100, line 18: I would hope we could keep `wasmfs_internal` headers in the `system/lib/wasmfs` directory and then have them installed via the `install_system_headers` help in `tools/system_libs.py`.  No need to relocate them in git to `system/include` I would hope.

system/include/emscripten/wasmfs_internal/backend.h
martenrichter: I will revert for headers..., but it may take until next weekend...

system/lib/wasmfs/virtual.h
martenrichter: Ok, I will check, if this PR is necessary at all.

system/include/emscripten/wasmfs_internal/backend.h
martenrichter: Ok, fixed.

system/lib/wasmfs/virtual.h
martenrichter: Ok, change it as suggested,

PR #24748, opened 2025-07-21
Use threaded HTTP server for test harness.

### LINKED ISSUE #10232: browser.test_offset_converter test is flakey
A few times now I've seen it fail, and strangly enough it fails often on both chrome and firefox in the same run.

e.g.
https://app.circleci.com/jobs/github/emscripten-core/emscripten/231312

```
test_offset_converter (test_browser.browser) ... emcc:WARNING: Wasm source map won't be usable in a browser without --source-map-base
[test error (see below), automatically retrying]
Expected to have '/report_result?1' == '/report_result?0', diff:

--- expected
+++ actual
@@ -1 +1 @@
-/report_result?1
+/report_result?0
```

brendandahl:
The test `test_offset_converter_pthread` hangs during the `fetch` call to report results to the server. This seems to be caused by Chrome opening an earlier connection to the server and leaving it open which in turn causes pythons`SimpleHTTPRequestHandler` to stop responding to new requests. Using the `ThreadingHTTPServer` allows python to keep processing requests and fixes the hang.

More info on this issue can be found in:  https://bugs.python.org/issue31639 and the chrome issue https://issues.chromium.org/issues/40978518.

Fixes #10232

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -9,7 +9,7 @@ from pathlib import Path
 from subprocess import PIPE, STDOUT
 from typing import Dict, Tuple
 from urllib.parse import unquote, unquote_plus, urlparse, parse_qs
-from http.server import HTTPServer, SimpleHTTPRequestHandler
+from http.server import ThreadingHTTPServer, SimpleHTTPRequestHandler
 import contextlib
 import difflib
 import hashlib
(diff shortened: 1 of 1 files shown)


sbc100:
Can you explain in the PR description a little about what the differences are here and why it might help?

brendandahl:
Yeah, sorry this was meant to be a draft. Still exploring if this is the issue on circleci.

brendandahl:
Some more info I found on the python tracker https://bugs.python.org/issue31639 and a chromium bug linked https://issues.chromium.org/issues/40978518.

sbc100 (approved):

PR #24749, opened 2025-07-21
Add @requires_git_checkout decorator. NFC

sbc100:
A recent change caused the emscripten-releases roller to break: #24723.

This was reverted in #24741 but we should be able to re-land after this change.


emcc.py
```diff
@@ -52,7 +52,7 @@
 # steps.  In packaged versions (e.g. emsdk) this file does not exist (because
 # it is excluded in tools/install.py) and these steps are assumed to have been
 # run already.
-if os.path.exists(utils.path_from_root('.git')) and os.path.exists(utils.path_from_root('bootstrap.py')):
+if os.path.exists(utils.path_from_root('.git')):
```
kripken: The comment needs to be updated too, with an explanation for why this file can always be assumed to exist.

sbc100: I just reverted this line to keep this change small.

PR #24753, opened 2025-07-21
Add git pre-push hook. NFC

sbc100:
Currently this script just runs `ruff check` to avoid pushing anything with python lint errors (since ruff is super fast this seems like a reasonable thing to do).


sbc100:
> Does this mean that all developers must have ruff installed to do pushes?

Absolutely not.  It just means that those that install it will generated fewer lint failures on the CI side.

kripken:
Oh, because the hook must be installed manually, we don't do it by default? sgtm

kripken (approved):

sbc100:
Yes, you currently need to install them manually.  At some point I will maybe advertise this more widely.

PR #24757, opened 2025-07-21
Update LLVM libraries to 20.1.8

aheejin:


sbc100 (approved):
lgtm if the tests pass

system/lib/compiler-rt/readme.txt
```diff
@@ -20,4 +20,4 @@ Modifications
 
 For a list of changes from upstream see the compiler-rt files that are part of:
 
-https://github.com/llvm/llvm-project/compare/llvmorg-20.1.4...emscripten-core:emscripten-libs-20
+https://github.com/llvm/llvm-project/compare/llvmorg-20.1.8...emscripten-core:emscripten-libs-20
```
kleisauke, line 23: Just wondering, does this changeset only include modifications to existing files? I noticed that `compiler-rt/lib/sanitizer_common/sanitizer_emscripten.cpp` is missing there.

aheejin: Yes, we don't maintain emscripten-only files in https://github.com/emscripten-core/llvm-project repo.

PR #24765, opened 2025-07-22
[llvm-libc] Import setjmp from llvm-libc

google-yfyang:
Defined a wasm-specific wasm/sigsetjmp.cpp that just fowards the call to setjmp.

Defines in the musl header don't quite work with llvm-libc's setjmp.


system/lib/llvm-libc/src/setjmp/wasm/sigsetjmp.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+#if !defined(LIBC_TARGET_ARCH_IS_WASM)
+#error "Invalid file include"
+#endif
+
+namespace LIBC_NAMESPACE_DECL {
+[[gnu::returns_twice]] int sigsetjmp(jmp_buf sigjmp_buf, int savesigs) {
+    return setjmp(sigjmp_buf);
+}
+}
```
sbc100: Missing new line here at EOF

system/lib/libc/musl/include/setjmp.h
```diff
@@ -26,7 +26,7 @@ typedef struct __jmp_buf_tag {
  || defined(_BSD_SOURCE)
 typedef jmp_buf sigjmp_buf;
 /* XXX EMSCRIPTEN: No signals support, alias sigsetjmp and siglongjmp to their non-signals counterparts. */
-#if __EMSCRIPTEN__
+#if __EMSCRIPTEN__ && !defined(LLVM_LIBC)
```
sbc100, line 29: I wonder if we could remove this musl patch and instead do this redirection in source files, then our musl patch would be smaller and we could avoid the need for this new LLVM_LIBC macro.

I'm fine landing this patch either way, we can always clean this up as a followup.

google-yfyang: I am fine either way as well. 
On the hand, this should no longer be of any issue once we switch over to use llvm-libc public headers over the musl ones.

sbc100: True, but I honestly think that process will take O(years).   We can't delete musl until llvm-libc is has complete parity, or at least close enough that we accept some minor regressions.

google-yfyang: Certainly. As I expect to see more of these as we import more llvm-libc, I feel like we should probably try to address this properly.

google-yfyang: PTAL. The forwarding is now done in source files.

system/lib/libc/musl/src/setjmp/sigsetjmp.c
```diff
@@ -0,0 +1,5 @@
+#include <setjmp.h>
+
+int sigsetjmp(jmp_buf sigjmp_buf, int savesigs) {
+    return setjmp(sigjmp_buf);
```
sbc100: Can you make this into separate change that lands first?

tools/system_libs.py
```diff
@@ -1363,6 +1366,7 @@ def get_files(self):
         filenames=['thread_profiler.c'])
 
     libc_files += glob_in_path('system/lib/libc/compat', '*.c')
+    libc_files += glob_in_path('system/lib/libc/musl/setjmp', '*.c')
```
sbc100: Can you put these two wrappers into the existing `emscripten_libc_stubs.c` instead?

sbc100: Actually it looks like `system/lib/libc/musl/src/signal/siglongjmp.c` already exists and does the right thing.

Can you update the existing `system/lib/libc/musl/src/signal/sigsetjmp.c` to include this stub and then add those two files to list on line 1330 in `system_libs.py`.

system/lib/libc/musl/src/setjmp/sigsetjmp.c
google-yfyang: See https://github.com/emscripten-core/emscripten/pull/24779

PR #24766, opened 2025-07-22
[test] Handling buffering of subprocess output. NFC

sbc100:
The python unittest framework has a stdout/stderr buffering mechanism that we don't take advantage of but that I would like to move towards using more.

This change allows our tests to run in buffered mode without generating output on the underlying stdout and stderr file descriptors.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -13,6 +13,7 @@ from http.server import HTTPServer, SimpleHTTPRequestHandler
 import contextlib
 import difflib
 import hashlib
+import io
 import itertools
 import logging
 import multiprocessing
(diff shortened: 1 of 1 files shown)


test/common.py
```diff
@@ -1763,19 +1764,36 @@ def clear(self):
...
     except subprocess.CalledProcessError as e:
       if check and e.returncode != 0:
         print(e.stdout)
         print(e.stderr)
         self.fail(f'subprocess exited with non-zero return code({e.returncode}): `{shlex.join(cmd)}`')
 
+    if stdout_buffering:
+      sys.stdout.write(rtn.stdout)
+    if stderr_buffering:
+      sys.stderr.write(rtn.stderr)
```
kripken, line 1794: I don't understand the change here... if we don't do (say) stdout buffering then we let `run_process` write to stdout directly. otoh if we do stdout buffering then we write that buffer to `sys.stdout` ourselves. Isn't the result the same, maybe only with some slight delay?

sbc100: The issue is that the unittest.py module overrides `sys.stdout` and `sys.stderr` in the expectation that will capture all output: https://github.com/python/cpython/blob/aafb1435d803c68ea3bbd2748f1e141c9deaadb0/Lib/unittest/result.py#L65-L71

However, overriding these doesn't effect subprocess, and the default of `stdout=None, stderr=None` means that the actually process's stdout and stderr are inherited not the user-space overrides of `sys.stdout` and `sys.stderr`

kripken: I see, thanks, so you are manually applying the override to subprocess as well... and I guess subprocess writes directly to the streams rather than go through Python's `sys.stdout/err`.

sbc100: Yes, exactly, hopefully the PR description capture that somwhat.

PR #19569, opened 2023-06-08
Remove legacy fields from stat struct. NFC

sbc100:
Musl needs these in order to support the linux kernel but we don't need to use the same ABI in emscripten.

Fixes: #19567

PR #13007, opened 2020-12-10
[WIP] Merge musl-1.2.1 into #13006

kleisauke:
Just a test PR to see how this behaves in CI and to compare my changes with #13006. See https://github.com/emscripten-core/emscripten/issues/7279 for context.

Lines with `TODO(kleisauke):` needs some further investigation.

PR #24769, opened 2025-07-22
Use 64-bit fields for fsblkcnt_t and fsfilcnt_t

sbc100:
We were running into issues where some of our bots were reporting numbers larger than 2^32 when calling statfs on the underlying filesystem.

### DIFF
--- a/src/lib/libsyscall.js
+++ b/src/lib/libsyscall.js
@@ -66,11 +66,11 @@ var SyscallsLibrary = {
     writeStatFs(buf, stats) {
       {{{ makeSetValue('buf', C_STRUCTS.statfs.f_bsize, 'stats.bsize', 'i32') }}};
       {{{ makeSetValue('buf', C_STRUCTS.statfs.f_frsize, 'stats.bsize', 'i32') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_blocks, 'stats.blocks', 'i32') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_bfree, 'stats.bfree', 'i32') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_bavail, 'stats.bavail', 'i32') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_files, 'stats.files', 'i32') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_ffree, 'stats.ffree', 'i32') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_blocks, 'stats.blocks', 'i64') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_bfree, 'stats.bfree', 'i64') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_bavail, 'stats.bavail', 'i64') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_files, 'stats.files', 'i64') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_ffree, 'stats.ffree', 'i64') }}};
       {{{ makeSetValue('buf', C_STRUCTS.statfs.f_fsid, 'stats.fsid', 'i32') }}};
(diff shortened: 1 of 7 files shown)


src/lib/libsyscall.js
```diff
@@ -66,11 +66,11 @@ var SyscallsLibrary = {
...
-      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_blocks, 'stats.blocks', 'i32') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_bfree, 'stats.bfree', 'i32') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_bavail, 'stats.bavail', 'i32') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_files, 'stats.files', 'i32') }}};
-      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_ffree, 'stats.ffree', 'i32') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_blocks, 'stats.blocks', 'i64') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_bfree, 'stats.bfree', 'i64') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_bavail, 'stats.bavail', 'i64') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_files, 'stats.files', 'i64') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.statfs.f_ffree, 'stats.ffree', 'i64') }}};
```
kleisauke, line 73: These types should be `u64`. Though, the previous signedness was also incorrect, I just opened PR #24772 for that.

kleisauke:
Note that this is an ABI break. Rust had a lot of issues with the previous stat-related ABI break, see e.g. https://github.com/emscripten-core/emscripten/pull/19569#issuecomment-1605440414 and https://github.com/rust-lang/rust/issues/131467. Though, I haven't yet verified whether the same applies to the `statfs` and `statvfs` structs.

kleisauke:
PR #24773 (which is split from #13007) makes updating `alltypes.h` a bit easier. That is, you only have to remove these two:
https://github.com/emscripten-core/emscripten/blob/4bb13baaf45ae6663b3df26dd0ff0e5a89fabfdd/system/lib/libc/musl/arch/emscripten/bits/alltypes.h.in#L26
https://github.com/emscripten-core/emscripten/blob/4bb13baaf45ae6663b3df26dd0ff0e5a89fabfdd/system/lib/libc/musl/arch/emscripten/bits/alltypes.h.in#L28

and regenerate `alltypes.h` using `update_alltypes.sh`.

sbc100:
> Note that this is an ABI break. Rust had a lot of issues with the previous stat-related ABI break, see e.g. [#19569 (comment)](https://github.com/emscripten-core/emscripten/pull/19569#issuecomment-1605440414) and [rust-lang/rust#131467](https://github.com/rust-lang/rust/issues/131467). Though, I haven't yet verified whether the same applies to the `statfs` and `statvfs` structs.

True, that is why I mention in the ChangeLog.   Note that we don't (currently) make any guarantees that object files and libraries built with one version of emscripten can be used by another version.

sbc100:
> PR #24773 (which is split from #13007) makes updating `alltypes.h` a bit easier. That is, you only have to remove these two:
> 
> https://github.com/emscripten-core/emscripten/blob/4bb13baaf45ae6663b3df26dd0ff0e5a89fabfdd/system/lib/libc/musl/arch/emscripten/bits/alltypes.h.in#L26
> 
> 
> https://github.com/emscripten-core/emscripten/blob/4bb13baaf45ae6663b3df26dd0ff0e5a89fabfdd/system/lib/libc/musl/arch/emscripten/bits/alltypes.h.in#L28
> 
> and regenerate `alltypes.h` using `update_alltypes.sh`.

Maybe lets land this first, then the new `.in` files will be two lines shorter.

src/lib/libsyscall.js
kleisauke: Actually, `-sWASM_BIGINT=0` doesn't support `u64`, so that won't work. I'll mark this as resolved.

kleisauke:
> Though, I haven't yet verified whether the same applies to the `statfs` and `statvfs` structs

Grepping for these structs in Rust returned no results, so it's likely/hopefully safe.

PR #24772, opened 2025-07-23
Fix signedness/type mismatches when marshaling stat-related structs

kleisauke:


sbc100:
lgtm, but I'm fine with #24769 landing first too.

src/lib/libsyscall.js
```diff
@@ -60,20 +60,20 @@ var SyscallsLibrary = {
       {{{ makeSetValue('buf', C_STRUCTS.stat.st_mtim.tv_nsec, '(mtime % 1000) * 1000 * 1000', SIZE_TYPE) }}};
       {{{ makeSetValue('buf', C_STRUCTS.stat.st_ctim.tv_sec, 'Math.floor(ctime / 1000)', 'i64') }}};
       {{{ makeSetValue('buf', C_STRUCTS.stat.st_ctim.tv_nsec, '(ctime % 1000) * 1000 * 1000', SIZE_TYPE) }}};
-      {{{ makeSetValue('buf', C_STRUCTS.stat.st_ino, 'stat.ino', 'i64') }}};
+      {{{ makeSetValue('buf', C_STRUCTS.stat.st_ino, 'stat.ino', 'u64') }}};
```
sbc100: IIRC the signed-ness of the type only really matter for `makeGetValue`.. is that right?    i.e. its when reading the bits that its important to interpret them.

Supporting evidence for this is that the `makeSetValue` code doesn't even have code paths for handling `u64` and `u53` here: https://github.com/emscripten-core/emscripten/blob/90250aa8dfc74e38876896ef95c342780dfad132/src/parseTools.mjs#L453-L466

I think maybe just always use the signed array views works for setting values?

I'm not sure what happens if we write a negative number to an unsigned typed array, but I guess its undefined behaviour from our POV and we should assert in debug build.

kleisauke: According to the failing tests, there's seems to be indeed support missing for `-sWASM_BIGINT=0` and `makeSetValue(x, y, z, 'u64')`. Commit 5511235e9d09962a83a9db202b8ed2c32dce3114 reverts that part.

I'm not sure if this signedness mismatch only affects `makeGetValue()`, since `getHeapForType()` is also used in `makeSetValue()` (e.g. `HEAP32` versus `HEAPU32`).

sbc100: > I'm not sure if this signedness mismatch only affects `makeGetValue()`, since `getHeapForType()` is also used in `makeSetValue()` (e.g. `HEAP32` versus `HEAPU32`).

I'm not sure either, but I would say that seems like signedness is critical when reading values (since the bytes need to be interpreted), but maybe not critical when writing them.

kleisauke:
The codesize tests appear to be quite sensitive, especially `other.test_codesize_hello_dylink_all`. IIRC, we allowed some slop/slack factor in the past, but it seems that's no longer configurable.

sbc100:
Yes, I apologize for that.    I'm not sure if we should remove that test again.. maybe?   Or we should have it auto-update as part of the review process.

kleisauke:
I'm not certain what the ideal solution is, but one possibility is to create a bot that automatically updates test expectations whenever a specific label (perhaps reuse https://github.com/emscripten-core/emscripten/labels/code%20size or introduce "needs-rebaseline"?) is applied to a PR. Once the bot completes the update, it would remove the label.

Using labels helps prevent potential abuse, since applying them requires triage access or higher (this idea is also somewhat inspired by LLVM's convenient `/cherry-pick` command, which also requires proper permissions, such as the ability to edit the milestone).

kleisauke:
... ah, I forgot that this only works when the "Allow edits by maintainers" option is enabled, though I believe it's enabled by default.

sbc100:
> I'm not certain what the ideal solution is, but one possibility is to create a bot that automatically updates test expectations whenever a specific label (perhaps reuse [ code size](https://github.com/emscripten-core/emscripten/labels/code%20size) or introduce "needs-rebaseline"?) is applied to a PR. Once the bot completes the update, it would remove the label.
> 
> Using labels helps prevent potential abuse, since applying them requires triage access or higher (this idea is also somewhat inspired by LLVM's convenient `/cherry-pick` command, which also requires proper permissions, such as the ability to edit the milestone).

Yes, we already have an action that will create rebasline CLs for main: https://github.com/emscripten-core/emscripten/actions/workflows/rebaseline-tests.yml

Figuring out how to make this work PRs in flight is what I'd like to do.  

I've not get written a bot that responds to labels of commands but it seems possible.

PR #24776, opened 2025-07-23
Limit array buffer tracing in TRACE_WEBGL_CALLS

sbc100:
It seems like a recent change to chrome means that error logging is now sent to stderr, which means certain browser tests that use TRACE_WEBGL_CALLS are hitting the 50mb log limit in circleCI.  Better to avoid logging huge array buffers.

### DIFF
--- a/src/lib/libwebgl.js
+++ b/src/lib/libwebgl.js
@@ -663,12 +663,24 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
       var contextHandle = glCtx.canvas.GLctxObject.handle;
       glCtx[f] = function(...args) {
         var ret = orig.apply(this, args);
-        // Some GL functions take a view of the entire linear memory.  Replace
-        // such arguments with the string 'HEAP' to avoid serializing all of
-        // memory.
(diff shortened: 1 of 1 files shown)


src/lib/libwebgl.js
```diff
@@ -663,12 +663,24 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
...
-        // such arguments with the string 'HEAP' to avoid serializing all of
-        // memory.
         for (var i in args) {
-          if (ArrayBuffer.isView(args[i]) && args[i].byteLength === HEAPU8.byteLength) {
-            args[i] = 'HEAP';
+          if (ArrayBuffer.isView(args[i])) {
+            // Some GL functions take a view of the entire linear memory.  Replace
+            // such arguments with the string 'HEAP' to avoid serializing all of
+            // memory.
+            if (args[i].byteLength === HEAPU8.byteLength) {
```
kripken, line 671: This could have compared `.buffer` between them to be more precise, but this is probably good enough in practice.

sbc100: Just FYI, this is pre-existing code

PR #24778, opened 2025-07-23
[AUDIO_WORKLET] Wait for Wasm instance to start using audio worklet.

brendandahl:
Fixes intermittent test failures where the AudioWorklet was used before the Wasm module was loaded in MINIMAL_RUNTIME.

### DIFF
--- a/src/audio_worklet.js
+++ b/src/audio_worklet.js
@@ -12,6 +12,7 @@
 // the node constructor's "processorOptions" field, we can share the necessary
 // bootstrap information from the main thread to the AudioWorkletGlobalScope.
 
+var instantiatePromise;
 if (ENVIRONMENT_IS_AUDIO_WORKLET) {
 
 function createWasmAudioWorkletProcessor(audioParams) {
@@ -161,6 +162,10 @@ class BootstrapMessages extends AudioWorkletProcessor {
     messagePort = this.port;
     /** @suppress {checkTypes} */
     messagePort.onmessage = async (msg) => {
+#if MINIMAL_RUNTIME
+      // Wait for the module instantiation before processing messages.
+      await instantiatePromise;
+#endif
       let d = msg.data;
       if (d['_wpn']) {
         // '_wpn' is short for 'Worklet Processor Node', using an identifier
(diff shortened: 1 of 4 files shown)


sbc100 (commented):
How is this solved in the regular runtime?

brendandahl:
Regular runtime instantiates on the worker synchronously. We could potentially do the same for minimal, but it will remove some of the shared instantiation code.

src/postamble_minimal.js
```diff
@@ -92,6 +92,8 @@ var wasmExports;
 var wasmModule;
 #endif
 
+var instantiatePromise;
```
sbc100: How about declaring this in `src/audio_worklet.js`

Also, the `instantiatePromise = ` usage below should be wrapped in `if AUDIO_WORKET` I guess?   Although no code size tests were effected maybe because closure can delete this?  Still best not to generate it in the first place though.. especially with MINIMAL_RUNTIME.

brendandahl: I'd prefer to leave it as is since adding more ifdefs is going to make this even more unreadable. Closure/rollup/webpack are all able to get rid of unused variables.  Are people running minimal runtime without some minification step?

sbc100: IIRC the goal of minimal runtime is to have the default output be minimal, not depend on DCE steps

@juj, is that correct?

juj: Yes absolutely, it is definitely something that MINIMAL_RUNTIME should achieve: smallest output possible.

I can understand that the #ifdef jungle can look a bit messy, but it is worth it in MINIMAL_RUNTIME philosophy. Also the intent is that we should not rely on CLOSURE_COMPILER to optimize code in the "easy cases" that we can deal with ourselves. This is because not every user is able to utilize closure: it requires adhering to defining a heavier set of internal/external public code boundary with the externs model. So not all users can utilize it.

Note that Closure also has challenges of optimizing away variables at global scope.

src/audio_worklet.js
```diff
@@ -12,6 +12,7 @@
 // the node constructor's "processorOptions" field, we can share the necessary
 // bootstrap information from the main thread to the AudioWorkletGlobalScope.
 
+var instantiatePromise;
```
sbc100, line 16: `#if MINIMAL_RUNTIME` here too for consistency.

I'm hoping we can find a nice way to do this one day..

PR #24779, opened 2025-07-23
[llvm-libc] Move the Emscripten patch for sigsetjmp and siglongjmp in setjmp.h to corresponding sources

google-yfyang:
This change helps llvm-libc imports to look cleaner.


tools/system_libs.py
```diff
@@ -1022,6 +1022,7 @@ def get_files(self):
 
 
 class libc(MuslInternalLibrary,
+           libcompiler_rt,
```
sbc100: I don't think this line is needed is it?

system/lib/libc/musl/src/signal/sigsetjmp.c
```diff
@@ -0,0 +1,6 @@
+#include <setjmp.h>
```
sbc100, line 2: Can you wrap this whole file in `if __EMSCRIPTEN__`?

sbc100 (approved):
Thanks!

tools/system_libs.py
google-yfyang: Yeah you are right.

system/lib/libc/musl/src/signal/sigsetjmp.c
google-yfyang: Done.

sbc100 (approved):

sbc100:
Ah, I see the problem with now.

Currently we don't build libc with different sjlj variants, but these new files would vary between the different modes.  So we would need to inherit from `SjLjLibrary`.. but sadly that would introduce yet another axis for libc variation adding N new versions of libc to build.

I'm not quite sure what the best option is.   It could be simply to compile these new files into libcompiler-rt instead (which already has this variant).  Its not very pretty, since these functions logically belong in libc.. but it would at least work.

google-yfyang:
> Ah, I see the problem with now.
> 
> Currently we don't build libc with different sjlj variants, but these new files would vary between the different modes. So we would need to inherit from `SjLjLibrary`.. but sadly that would introduce yet another axis for libc variation adding N new versions of libc to build.
> 
> I'm not quite sure what the best option is. It could be simply to compile these new files into libcompiler-rt instead (which already has this variant). Its not very pretty, since these functions logically belong in libc.. but it would at least work.

The inheritance from `SjLjLibrary` was necessary when I tried to get sources udner `system/lib/libc/musl/src/setjmp/` to work. For reasons that I don't understand, moving the files to `signal/` no longer requires the inheritance.  Build and test can execute without any issue.

sbc100:
core0.test_siglongjmp_wasm_legacy is failing, I think because libc doesn't have different versions for the different types of setjmp/longjmp support

sbc100:
My solution would be to build build `siglongjmp.c` and `sigsetjmp.c` into compiler-rt rather than libc.

Its kind of a hack and probably requires a comment such as:

```
# HACK: These two files normally belong in libc, but because they depend on the
# SjLj ABI we build them into compiler-rt instead.  This is because compiler-rt
# already has variants for the different SjLj ABIs.  Including these in libc would
# require libc to also be built with SjLjLibrary variants which would double the
# total number of libc variants.
```

google-yfyang:
> My solution would be to build build `siglongjmp.c` and `sigsetjmp.c` into compiler-rt rather than libc.
> 
> Its kind of a hack and probably requires a comment such as:
> 
> ```
> # HACK: These two files normally belong in libc, but because they depend on the
> # SjLj ABI we build them into compiler-rt instead.  This is because compiler-rt
> # already has variants for the different SjLj ABIs.  Including these in libc would
> # require libc to also be built with SjLjLibrary variants which would double the
> # total number of libc variants.
> ```

Yes I did see that just now. I am also running into this error once the variant issue is addressed.
```
-- begin program output --
node:internal/process/promises:392
      new UnhandledPromiseRejection(reason);
      ^

UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<Exception>".
    at throwUnhandledRejectionsMode (node:internal/process/promises:392:7)
    at processPromiseRejections (node:internal/process/promises:475:17)
    at process.processTicksAndRejections (node:internal/process/task_queues:106:32) {
}
Thrown at:
    at throwUnhandledRejectionsMode (node:internal/process/promises:392:7)
    at processPromiseRejections (node:internal/process/promises:475:17)
    at processTicksAndRejections (node:internal/process/task_queues:106:32)


Node.js v24.4.1
-- end program output --
FAIL

FAIL: test_siglongjmp_wasm (test_core.core0.test_siglongjmp_wasm)
```

Adding `nothrow` attribute to siglongjmp doesn't appear to address this. 

I am starting to wonder if this is worthwhile given the alternative is to just have a LLVM-LIBC define.

sbc100:
> I am starting to wonder if this is worthwhile given the alternative is to just have a LLVM-LIBC define.

Yup, thats fair.. lets abandon this for now maybe?

google-yfyang:
> > I am starting to wonder if this is worthwhile given the alternative is to just have a LLVM-LIBC define.
> 
> Yup, thats fair.. lets abandon this for now maybe?

Sounds good.

PR #24782, opened 2025-07-24
Fix MINIMAL_RUNTIME_STREAMING_WASM_INSTANTIATION under node. NFC

sbc100:
We cannot actually use this API under node since the fetch API doesn't support local files.

However, this failure was going undetected for two reason:

1. The test was only looking for the string 'hello' in the output which also appears in backtraces.
2. The instantiation error was being swallowed by a catch block that simply printed the error the console, so the node process was exiting with success.

### DIFF
--- a/src/postamble_minimal.js
+++ b/src/postamble_minimal.js
@@ -108,13 +108,17 @@ var imports = {
 
 #if MINIMAL_RUNTIME_STREAMING_WASM_INSTANTIATION
 // https://caniuse.com/#feat=wasm and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiateStreaming
-#if MIN_FIREFOX_VERSION < 58 || MIN_CHROME_VERSION < 61 || MIN_NODE_VERSION < 180100 || MIN_SAFARI_VERSION < 150000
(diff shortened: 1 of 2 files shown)


src/postamble_minimal.js
```diff
@@ -108,13 +108,17 @@ var imports = {
...
-#if MIN_FIREFOX_VERSION < 58 || MIN_CHROME_VERSION < 61 || MIN_NODE_VERSION < 180100 || MIN_SAFARI_VERSION < 150000
+#if MIN_FIREFOX_VERSION < 58 || MIN_CHROME_VERSION < 61 || MIN_SAFARI_VERSION < 150000 || ENVIRONMENT_MAY_BE_NODE
 #if ASSERTIONS && !WASM2JS
 // Module['wasm'] should contain a typed array of the Wasm object data, or a
 // precompiled WebAssembly Module.
 assert(WebAssembly.instantiateStreaming || Module['wasm'], 'Must load WebAssembly Module in to variable Module.wasm before adding compiled output .js script to the DOM');
 #endif
 (WebAssembly.instantiateStreaming
+#if ENVIRONMENT_MAY_BE_NODE
+  // Node's fetch API cannot by used for local files, so we cannot use instantiateStreaming
```
kripken:
```suggestion
  // Node's fetch API cannot be used for local files, so we cannot use instantiateStreaming
```

PR #24777, opened 2025-07-23
Rebaseline tests after recent binaryen change.

sbc100:
Also, temporarily disable test_codesize_cxx_lto until https://github.com/WebAssembly/binaryen/pull/7748 rolls in.

PR #24783, opened 2025-07-24
Restore test_other.other.test_codesize_cxx_lto after Binaryen fix. NFC

kripken:
Followup to

https://github.com/emscripten-core/emscripten/pull/24777/files#diff-59f6b346090f49748193fef565b0c7e223f93ecef5bb116fa370d0b864af9493

after

https://github.com/WebAssembly/binaryen/pull/7748

landed.

Code size savings are from

https://github.com/WebAssembly/binaryen/pull/7728


sbc100 (commented):
Nice!

Looks like `_tzset_js` could be eliminated from the JS side which was a decent size.

BTW the convention for "NFC" that we(I) have been using is to put it at the end of the PR title without square brackets.

PR #24784, opened 2025-07-24
[pthreads] Use blob to create worker in single file mode.

brendandahl:
This keeps the source readable and avoids console error messages including the full base64 data url in the console output.


src/lib/libpthread.js
```diff
@@ -425,7 +425,12 @@ var LibraryPThread = {
       // See https://github.com/webpack/webpack/issues/12638
       worker = new Worker(new URL('{{{ pthreadWorkerScript }}}', import.meta.url), {{{ pthreadWorkerOptions }}});
 #else // EXPORT_ES6
+#if SINGLE_FILE
```
sbc100: One issue here is that `SINGLE_FILE` can also be used on pure JS output.  i.e. we don't know that we are running inside a page with the generated HTML.   e.g. `./emcc -sSINGLE_FILE -o out.js foo.c`

sbc100 (approved):
Nice!

PR #24787, opened 2025-07-24
[embind] Handle -fno-exceptions and try/catch redefinition macros.

brendandahl:
When some code bases disable exceptions they also redefine try/catch with preprocessor macros. This causes the val coroutine function `unhandled_exception` to fail to compile with
`error: use of undeclared identifier 'error'`.

### DIFF
--- a/system/include/emscripten/val.h
+++ b/system/include/emscripten/val.h
@@ -728,6 +728,9 @@ public:
   auto initial_suspend() noexcept { return std::suspend_never{}; }
   auto final_suspend() noexcept { return std::suspend_never{}; }
 
+// When exceptions are disabled we don't define unhandled_exception and rely
+// on the default terminate behavior.
+#ifdef __cpp_exceptions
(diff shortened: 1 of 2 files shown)


system/include/emscripten/val.h
```diff
@@ -728,6 +728,9 @@ class val::promise_type {
   auto initial_suspend() noexcept { return std::suspend_never{}; }
   auto final_suspend() noexcept { return std::suspend_never{}; }
 
+// When exceptions are disabled we don't define unhandled_exception and rely
+// on the default terminate behavior.
+#ifdef __cpp_exceptions
   // On an unhandled exception, reject the stored promise instead of throwing
   // it asynchronously where it can't be handled.
   void unhandled_exception() {
```
sbc100, line 736: Where are the references to this function?

brendandahl: https://en.cppreference.com/w/cpp/coroutine/coroutine_traits.html

kripken (approved):

PR #24791, opened 2025-07-28
[embind] Fix and test for incorrect dynCalls signature

### LINKED ISSUE #24762: RuntimeError: null function or function signature mismatch
I am compiling C++ code using emsdk 4.0.11, and when calling it on the web, I encounter the following error:
```cpp
RuntimeError: null function or function signature mismatch
    at hello2d.wasm.dynCall_viid (hello2d.wasm:0xcea710)
    at wrapper (hello2d.js:5554:16)
    at hello2d.js:734:12
    at dynCallLegacy (hello2d.js:1253:10)
    at dynCall (hello2d.js:1258:13)
    at hello2d.js:5296:66
    at BaseView.updateSize (eval at createJsInvoker (hello2d.js:5470:10), <anonymous>:8:1)
    at updateSize (common.ts:221:28)
    at window.onload (index.ts:52:13)

```
The error seems to be caused by a function signature mismatch, but the same code compiles and works fine when using emsdk 4.0.10 or earlier versions.

The implementation of BaseView.updateSize is as follows:
```cpp
void BaseView::updateSize(float devicePixelRatio) {
  if (!canvasID.empty()) {
    int width = 0;
    int height = 0;
    emscripten_get_canvas_element_size(canvasID.c_str(), &width, &height);
    auto sizeChanged = appHost->updateScreen(width, height, devicePixelRatio);
    if (sizeChanged && window) {
      window->invalidSize();
    }
  }
}

void BaseView::setImagePath(const std::string& imagePath) {
  auto image = tgfx::Image::MakeFromFile(imagePath.c_str());
  if (image) {
    appHost->addImage("bridge", std::move(image));
  }
}

```
The binding of BaseView.updateSize is as follows:
```
#include <emscripten/bind.h>
#include "BaseView.h"
using namespace emscripten;

EMSCRIPTEN_BINDINGS(TGFXDemo) {
  class_<BaseView>("BaseView")
      .function("setImagePath", &BaseView::setImagePath)
      .function("updateSize", &BaseView::updateSize)
      .function("draw", &BaseView::draw);
}

```
In JavaScript, it is used as follows:
```
const scaleFactor = window.devicePixelRatio;
BaseView.updateSize(scaleFactor);

```

What's odd is that setImagePath works fine, but there is an issue specifically with the updateSize function's signature. Can you please help analyze what changes in emsdk 4.0.11 might have caused this problem?

Any insights or suggestions would be greatly appreciated. Thank you!

cwoffenden:
Fixes #24762. Embind with C++ float types _and_ `-sDYNCALLS=1` (and by extension `-sASYNCIFY=1`) fails with:
```
Assertion failed: bad function pointer type - sig is not in dynCalls: 'iid'
```
This adds a one-char fix (from @RReverser) and adds a couple of regression tests:
```
test/runner other.test_embind_dyncalls
test/runner other.test_embind_async
```
Without the fix the only the original `other.test_embind` tests will pass.

~Noted during writing this: the comprehensive `embind_test.cpp` does not look to be run.~


RReverser:
Thanks for working on this!

> Noted during writing this: the comprehensive `embind_test.cpp` does not look to be run.

Which tests did you run? It's executed by https://github.com/emscripten-core/emscripten/blob/ec93a04a8a09dc21ecf7bcc20356a04b3bb492b1/test/test_other.py#L3341, that is, `other.test_embind*` variants.

I think it might be better to add a one-liner DYNCALLS variant over there in the `parameterized` array instead of creating a new C++ file, check that adding these flags is enough to reproduce the issue, and that the fix fixes them.

cwoffenden:
> I think it might be better to add a one-liner DYNCALLS variant

That was my original plan, but on searching Emcripten's repo for `embind_test.cpp` (which I'd seen and manually built but running looked to require some magic I couldn't find) I didn't see any Python files referring to it:

screenshot

Annoyingly, running a find locally I see the two references in `test_other.py`, which would've saved me some time. I'll take a look at changing this tomorrow.

sbc100:
> > I think it might be better to add a one-liner DYNCALLS variant
> 
> That was my original plan, but on searching Emcripten's repo for `embind_test.cpp` (which I'd seen and manually built but running looked to require some magic I couldn't find) I didn't see any Python files referring to it:
> 
> screenshot
> Annoyingly, running a find locally I see the two references in `test_other.py`, which would've saved me some time. I'll take a look at changing this tomorrow.

I've never had much luck with github's search abilities sadly.   I would recommend sticking with `git grep` for that kind of thing.

test/test_other.py
```diff
@@ -3356,6 +3356,9 @@ def test_embind_no_raw_pointers(self, filename):
     'o2_mem_growth': ['-O2', '-sALLOW_MEMORY_GROWTH', test_file('embind/isMemoryGrowthEnabled=true.cpp')],
     'o2_closure': ['-O2', '--closure=1', '--closure-args', '--externs ' + shlex.quote(test_file('embind/underscore-externs.js')), '-sASSERTIONS=1'],
     'strict_js': ['-sSTRICT_JS'],
+    # DYNCALLS tests the legacy native function API, ASYNCIFY currently implies DYNCALLS
+    'dyncalls': ['-sDYNCALLS=1'],
+    'async': ['-sASYNCIFY=1'],
```
sbc100: This test is fairly slow so if you can just add one that might be preferable.

Also, is this actually a bug in em_asm?  Can you repro this use EM_ASM?   Or does EM_ASM just not care about `d` vs `f`?

cwoffenden:
Now using parameters with `other.test_embind`.

test/test_other.py
cwoffenden: I added two to catch any future case where `ASYNCIFY` runs without `DYNCALLS`. I can remove the`ASYNCIFY` option.

I can take a look at `EM_ASM` tomorrow (it breaks when `dynCall_sig()` is used instead of `wasmTable` lookup).

RReverser: > Or does EM_ASM just not care about `d` vs `f`?

The latter. `em_asm` always had a latent typo / bug where C and C++ parts defined signature for `float` differently ('f' in C: https://github.com/emscripten-core/emscripten/blob/ec93a04a8a09dc21ecf7bcc20356a04b3bb492b1/system/include/emscripten/em_asm.h#L84 and 'd' in C++: https://github.com/emscripten-core/emscripten/blob/ec93a04a8a09dc21ecf7bcc20356a04b3bb492b1/system/include/emscripten/em_asm.h#L152), but because em_asm itself encodes both as `double`s (which is a small missed optimisation, but that's a separate topic), it didn't cause any issues.

Now that I'm reusing signatures for Embind too, and Embind does encode `float` vs `double` differently, it did reveal itself for dyncalls apparently.

I still want to use same source of signatures for both, exactly because it helps to reveal this kind of consistency issues (or like that one for `size_t` I recently fixed on Embind side) and helps to reuse more code between different binding systems in the future.

sbc100: Thats, I get this consistency is the goal and its good to fix this bug.

I just wanted to check my assumption as to why this long-standing bug went unnoticed.

RReverser (approved):

PR #24793, opened 2025-07-28
Cleanup handleWebGLProxying. NFC

sbc100:
The check here was changed from `SHARED_MEMORY` to `PTHREADS` since wasm workers don't support proxying.

### DIFF
--- a/src/lib/libhtml5_webgl.js
+++ b/src/lib/libhtml5_webgl.js
@@ -545,45 +545,45 @@ var LibraryHtml5WebGL = {
   emscripten_webgl_get_parameter_i64v: (param, dst) => writeI53ToI64(dst, GLctx.getParameter(param)),
 };
 
-function handleWebGLProxying(funcs) {
-#if SHARED_MEMORY
-// Process 'sync_on_webgl_context_handle_thread' and 'sync_on_current_webgl_context_thread' pseudo-proxying modes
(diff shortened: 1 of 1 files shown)


src/lib/libhtml5_webgl.js
```diff
@@ -545,45 +545,45 @@ var LibraryHtml5WebGL = {
...
+  // proxying mechanism, either proxying on-demand, unconditionally, or never,
+  // depending on build modes.
+  // 'sync_on_webgl_context_handle_thread' is used for function signatures that
+  // take a HTML5 WebGL context handle object as the first argument.
+  // 'sync_on_current_webgl_context_thread' is used for functions that operate on
+  // the implicit "current WebGL context" as activated via
+  // emscripten_webgl_make_current() function.
   function listOfNFunctionArgs(func) {
-    var args = [];
+    const args = [];
```
kripken, line 561: Do we prefer `const`? It is larger, and I don't think provides practical benefit for our glue code.

It also seems not much used:
```
$ grep "var " src/lib/lib* | wc -l
4026
$ grep "const " src/lib/lib* | wc -l
175
$ grep "let " src/lib/lib* | wc -l
167
```

sbc100: We prefer it in the compiler itself (which this is part of even though its inside a library file) but not the in generated code.

PR #24794, opened 2025-07-28
Simplify `new Function` call. NFC

sbc100:
This code was originally added I suppose before rest/spread operators were allowed.

IIUC this code is equivalent.

### DIFF
--- a/src/lib/libhtml5_webgl.js
+++ b/src/lib/libhtml5_webgl.js
@@ -603,8 +603,7 @@ function handleWebGLProxying(funcs) {
         funcs[i + '__deps'].push('$' + i + '_before_on_calling_thread');
         funcBody = `${i}_before_on_calling_thread(${funcArgsString}); ` + funcBody;
       }
-      funcArgs.push(funcBody);
-      funcs[i] = new (Function.prototype.bind.call(Function, Function, ...funcArgs));
+      funcs[i] = new Function(...funcArgs, funcBody);
     } else if (targetingOffscreenFramebuffer) {
       // When targeting only OFFSCREEN_FRAMEBUFFER, unconditionally proxy all GL calls to
       // main thread.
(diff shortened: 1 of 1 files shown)


src/lib/libhtml5_webgl.js
```diff
@@ -603,8 +603,7 @@ function handleWebGLProxying(funcs) {
         funcs[i + '__deps'].push('$' + i + '_before_on_calling_thread');
         funcBody = `${i}_before_on_calling_thread(${funcArgsString}); ` + funcBody;
       }
-      funcArgs.push(funcBody);
-      funcs[i] = new (Function.prototype.bind.call(Function, Function, ...funcArgs));
+      funcs[i] = new Function(...funcArgs, funcBody);
```
RReverser: I'll do you one better 

```suggestion
      funcs[i] = new Function(funcArgs, funcBody);
```

sbc100: I don't seem to see that one form listed in the docs.  Can you pass an array as the first arg?  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/Function

RReverser: It's not on MDN but each argument you pass can be a comma-separated string by itself, because `Function` constructor doesn't evaluate or even validate them at all, simply inlines them as-is (that's why MDN mentions support for comments, whitespaces and stuff).

Arrays are automatically stringified as comma-separated strings, so... yeah. I've been doing that in a few places in Emscripten before, and works great with less code.

RReverser: This section hints at it:

> All arguments passed to the function, except the last, are treated as the names of the identifiers of the parameters in the function to be created, in the order in which they are passed. The function will be dynamically compiled as a function expression, with the source assembled in the following fashion:
>
> ```js
>  function anonymous(${args.join(",")}) {
>  ${functionBody}
>  }
>  ```

PR #24795, opened 2025-07-28
fix dynCall deps

walkingeyerobot:
If `DYNCALLS` and `WASM_BIGINT` are both false, `dynCall` relies on both `dynCallLegacy` and `getWasmTableEntry`. This ensures both are properly depended on.


sbc100 (commented):
I would hope we can come up with a test for this.

Since WASM_BIGINT and DYNCALLS both default to false I'm somewhat suprised we didn't catch this with any existing tests.

sbc100:
I spend a little time debugging and figured out why we didn't have coverage of this.   See #24796

sbc100:
Oops, I meant #24797

PR #24798, opened 2025-07-28
Fix skipping reftests and add helper to skip ref tests.

brendandahl:
Reftests that were being skipped were still trying to run the reftest snapshot code and report a result. Since this was async, the result would sometimes come in while the next test was started and cause "excessive responses from previous test".

This should fix flakes with several test_webgl tests.

### DIFF
--- a/test/browser_reporting.js
+++ b/test/browser_reporting.js
@@ -5,7 +5,7 @@ var hasModule = typeof Module === 'object' && Module;
 
 var reportingURL = 'http://localhost:8888';
 
-async function reportResultToServer(result) {
+async function reportResultToServer(result, extra = '') {
   if (reportResultToServer.reported) {
     // Only report one result per test, even if the test misbehaves and tries to report more.
     reportStderrToServer(`excessive reported results, sending ${result}, test will fail`);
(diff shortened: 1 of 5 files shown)


test/browser_reporting.js
```diff
@@ -52,6 +52,11 @@ function reportStdoutToServer(message) {
   }
 }
 
+async function skipTest(message) {
+  skipTest.skipped = true;
+  await reportResultToServer(message, 'skipped:');
```
sbc100: Why not do `'skipped:' + message` here and avoid the extra argument to reportResultToServer?

sbc100 (commented):
Nice work!

sbc100:
I wonder if we can (also) make the server robust againt "excessive responses from previous test"?   What happen right now in that case?

test/browser_reporting.js
brendandahl: The colon gets escaped and then doesn't match the check in python.

sbc100: Can you explain more?  Maybe we do a better fix here.. but I'm not sure exactly what it looks like.

sbc100: Can you explain more?  Maybe we do a better fix here.. but I'm not sure exactly what it looks like.

brendandahl: `encodeURIComponent(':')` becomes `'%3A'` so then the python string matching for `skipped:` doesn't match.

sbc100: How about `reportResultToServer('skipped:' + reason, noEncode=true)`?

sbc100: Actually, a better fix might be to move the `output = unquote(output)` line up to line 2405 in test/common.py

test/reftest.js
```diff
@@ -23,6 +23,7 @@ function reftestUnblock() {
 function doReftest() {
   if (reftestBlocked) return;
   if (doReftest.done) return;
+  if (skipTest.skipped) return;
```
sbc100: How about just using the existing `reportResultToServer.reported` here? Then it will also handle errors / exceptions too.

sbc100: (and you don't need skipTest.skipped at all)

PR #24799, opened 2025-07-29
Fix IDBFS autoPersist on mkdir

handlerug:
FS.mkdir does not use mkdir under the hood but rather mknod. The patched version of mknod installed by IDBFS only invokes IDBFS.queuePersist() when the node's stream is closed, which never happens for mkdir as the handle is immediately discarded and no stream is ever created.

Furthermore, the bare fact of file/directory creation should be persisted even if no writes are made to the node.

The test doesn't catch this because reconcilation happens in a JavaScript task, and no tasks or microtasks can execute while code is executing. IDBFS.syncfs only has a chance to run after the entirety of the test function finishes executing, at which point all of file/directory operations have been executed and persisted in memory. As long as one of those operations calls IDBFS.queuePersist, it ensures that the whole memory filesystem will get persisted to IDB and therefore shadows missing calls to IDBFS.queuePersist in other operations.

Remove the mkdir node op patching (which has never been a real node op) and make mknod queue a persist operation as soon as the node is created without waiting for the caller to interact with a stream.

Add a separate test that runs every operation that modifies the filesystem in isolation and ensures the said operation calls `IDBFS.queuePersist()`.

### DIFF
--- a/src/lib/libidbfs.js
+++ b/src/lib/libidbfs.js
@@ -54,7 +54,7 @@ addToLibrary({
       // If the automatic IDBFS persistence option has been selected, then automatically persist
       // all modifications to the filesystem as they occur.
       if (mount?.opts?.autoPersist) {
-        mnt.idbPersistState = 0; // IndexedDB sync starts in idle state
+        mount.idbPersistState = 0; // IndexedDB sync starts in idle state
         var memfs_node_ops = mnt.node_ops;
         mnt.node_ops = {...mnt.node_ops}; // Clone node_ops to inject write tracking
         mnt.node_ops.mknod = (parent, name, mode, dev) => {
--- a/src/lib/libidbfs.js
+++ b/src/lib/libidbfs.js
@@ -85,10 +85,12 @@ addToLibrary({
             if (n.memfs_stream_ops.close) return n.memfs_stream_ops.close(stream);
           };
 
+          // Persist the node we just created to IndexedDB
+          IDBFS.queuePersist(mnt.mount);
+
           return node;
         };
         // Also kick off persisting the filesystem on other operations that modify the filesystem.
-        mnt.node_ops.mkdir   = (...args) => (IDBFS.queuePersist(mnt.mount), memfs_node_ops.mkdir(...args));
         mnt.node_ops.rmdir   = (...args) => (IDBFS.queuePersist(mnt.mount), memfs_node_ops.rmdir(...args));
         mnt.node_ops.symlink = (...args) => (IDBFS.queuePersist(mnt.mount), memfs_node_ops.symlink(...args));
         mnt.node_ops.unlink  = (...args) => (IDBFS.queuePersist(mnt.mount), memfs_node_ops.unlink(...args));
--- a/test/fs/test_idbfs_autopersist.c
+++ b/test/fs/test_idbfs_autopersist.c
@@ -1,0 +1,245 @@
+/*
+ * Copyright 2025 The Emscripten Authors.  All rights reserved.
+ * Emscripten is available under two separate licenses, the MIT license and the
+ * University of Illinois/NCSA Open Source License.  Both these licenses can be
+ * found in the LICENSE file.
+ */
+
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <emscripten.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/stat.h>
+#include <errno.h>
+
+EM_JS_DEPS(deps, "$callUserCallback");
+
+enum {
+  TEST_CASE_OPEN,
+  TEST_CASE_CLOSE,
+  TEST_CASE_SYMLINK,
+  TEST_CASE_UNLINK,
+  TEST_CASE_RENAME,
+  TEST_CASE_MKDIR,
+};
+
+static void test_case_open(void) {
+  switch (TEST_PHASE) {
+    case 1: {
+      int fd = open("/working1/file", O_RDWR | O_CREAT | O_EXCL, 0777);
+      assert(fd != -1);
+      break;
+    }
+    case 2: {
+      struct stat st;
+      int res = lstat("/working1/file", &st);
+      assert(res == 0);
+      assert(st.st_size == 0);
+      break;
+    }
+    default:
+      assert(false);
+  }
+}
+
+static void test_case_close(void) {
+  switch (TEST_PHASE) {
+    case 1: {
+      int fd = open("/working1/file", O_RDWR | O_CREAT | O_EXCL, 0777);
+      assert(fd != -1);
+      break;
+    }
+    case 2: {
+      int fd = open("/working1/file", O_RDWR | O_CREAT, 0777);
+      assert(fd != -1);
+      ssize_t bytes_written = write(fd, "foo", 3);
+      assert(bytes_written == 3);
+      int res = close(fd);
+      assert(res == 0);
+      break;
+    }
+    case 3: {
+      struct stat st;
+      int res = lstat("/working1/file", &st);
+      assert(res == 0);
+      assert(st.st_size == 3);
+      break;
+    }
+    default:
+      assert(false);
+  }
+}
+
+static void test_case_symlink(void) {
+  switch (TEST_PHASE) {
+    case 1: {
+      int fd = open("/working1/file", O_RDWR | O_CREAT | O_EXCL, 0777);
+      assert(fd != -1);
+      break;
+    }
+    case 2: {
+      int res = symlink("/working1/file", "/working1/symlink");
+      assert(res == 0);
+      break;
+    }
+    case 3: {
+      struct stat st;
+      int res = lstat("/working1/symlink", &st);
+      assert(res == 0);
+      break;
+    }
+    default:
+      assert(false);
+  }
+}
+
+static void test_case_unlink(void) {
+  switch (TEST_PHASE) {
+    case 1: {
+      int fd = open("/working1/file", O_RDWR | O_CREAT | O_EXCL, 0777);
+      assert(fd != -1);
+      break;
+    }
+    case 2: {
+      int res = unlink("/working1/file");
+      assert(res == 0);
+      break;
+    }
+    case 3: {
+      struct stat st;
+      int res = lstat("/working1/file", &st);
+      assert(res == -1);
+      assert(errno == ENOENT);
+      break;
+    }
+    default:
+      assert(false);
+  }
+}
+
+static void test_case_rename(void) {
+  switch (TEST_PHASE) {
+    case 1: {
+      int fd = open("/working1/file", O_RDWR | O_CREAT | O_EXCL, 0777);
+      assert(fd != -1);
+      break;
+    }
+    case 2: {
+      int res = rename("/working1/file", "/working1/file_renamed");
+      assert(res == 0);
+      break;
+    }
+    case 3: {
+      struct stat st;
+      int res = lstat("/working1/file_renamed", &st);
+      assert(res == 0);
+      res = lstat("/working1/file", &st);
+      assert(res == -1);
+      assert(errno == ENOENT);
+      break;
+    }
+    default:
+      assert(false);
+  }
+}
+
+static void test_case_mkdir(void) {
+  switch (TEST_PHASE) {
+    case 1: {
+      int res = mkdir("/working1/dir", 0777);
+      assert(res == 0);
+      break;
+    }
+    case 2: {
+      struct stat st;
+      int res = lstat("/working1/dir", &st);
+      assert(res == 0);
+      break;
+    }
+    default:
+      assert(false);
+  }
+}
+
+EMSCRIPTEN_KEEPALIVE
+void finish(void) {
+  emscripten_force_exit(0);
+}
+
+EMSCRIPTEN_KEEPALIVE
+void test(void) {
+  switch (TEST_CASE) {
+    case TEST_CASE_OPEN: test_case_open(); break;
+    case TEST_CASE_CLOSE: test_case_close(); break;
+    case TEST_CASE_SYMLINK: test_case_symlink(); break;
+    case TEST_CASE_UNLINK: test_case_unlink(); break;
+    case TEST_CASE_RENAME: test_case_rename(); break;
+    case TEST_CASE_MKDIR: test_case_mkdir(); break;
+    default: assert(false);
+  }
+
+  EM_ASM({
+    // Wait until IDBFS has persisted before exiting
+    runOnceIDBFSIdle(() => {
+      callUserCallback(_finish);
+    });
+  });
+}
+
+int main(void) {
+  EM_ASM({
+    globalThis.runOnceIDBFSIdle = (callback) => {
+      const { mount } = FS.lookupPath('/working1').node;
+      assert('idbPersistState' in mount, 'mount object must have idbPersistState');
+      if (mount.idbPersistState !== 0) {
+        // IDBFS hasn't finished persisting. Check again after all pending tasks have executed
+        setTimeout(() => runOnceIDBFSIdle(callback), 0);
+        return;
+      }
+      callback();
+    };
+
+    FS.mkdir('/working1');
+    FS.mount(IDBFS, {
+      autoPersist: true
+    }, '/working1');
+  });
+
+  if (TEST_PHASE == 1) {
+    EM_ASM({
+      // The first phase of a test case must start from an empty filesystem.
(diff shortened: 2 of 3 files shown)


src/lib/libidbfs.js
```diff
@@ -85,10 +85,14 @@ addToLibrary({
             if (n.memfs_stream_ops.close) return n.memfs_stream_ops.close(stream);
           };
 
+          // Persist IndexedDB right away if the node we just created is a directory
```
sbc100: Can you explain why this is needed?  (Either here or in the PR description)

Also, is there some kind of test we could write for this?

handlerug: As the PR description and the commit message mention, `FS.mkdir` uses `mknod` under the hood:
https://github.com/emscripten-core/emscripten/blob/1d05ff842aa2f48b4e488a9598d1912b3a92c8e2/src/lib/libfs.js#L739-L748

The `node` that we return would otherwise only call `IDBFS.queuePersist` when an open handle/stream would get closed, which does not happen for creating a directory as evidenced by the implementation of `FS.mkdir`. It is clear that the intent here was to persist IDBFS on directory creation, so I added code that calls `IDBFS.queuePersist` if `mknod` was called with a mode of directory.

I will try to write a test for this behavior a bit later.

sbc100: I see that makes sense.

How about updating the comment/description to say something like.  "Force a persist now since there is not corresponding `close` call for newly creating directories".

sbc100: Does the same problem exist for folks that directly call `mknod` to make files?  (i.e. they never get closed, so the persist is never triggered?)

sbc100: Actually it looks like the persistence should probably only be enabled if `autoPersist` is set, so maybe we just need to add `mknod` to list of `audoPersist` things by the `// Also kick off persisting the filesystem on other operations that modify the filesystem` comment.

test/fs/test_idbfs_sync.c
```diff
@@ -67,12 +67,32 @@ void test() {
...
+      ccall('test2', 'v');
+    }, 0);
+  );
+}
+
+void test2() {
+  struct stat st;
+
+  // a directory using which we will check if auto-persistence is working
+  if ((stat("/working1/dir2", &st) != -1) || (errno != ENOENT))
```
sbc100: With this `stat` fail if `queuePersist` does not happen?

I would have thought it would work in any cases since the in-memory FS contains this directory, no?

src/lib/libidbfs.js
handlerug: > Does the same problem exist for folks that directly call `mknod` to make files? (i.e. they never get closed, so the persist is never triggered?)

Actually yes, that is a good point. I removed the check for file mode and made it so that an `mknod` invocation will queue a persist operation immediately. I believe that without this, files created via `open()` would not get persisted without a `close()` call either.

> Actually it looks like the persistence should probably only be enabled if `autoPersist` is set, so maybe we just need to add `mknod` to list of `audoPersist` things by the `// Also kick off persisting the filesystem on other operations that modify the filesystem` comment.

`mknod` patching is already inside an `if` statement that checks whether `autoPersist` is set. I can remove the `IDBFS.queuePersist` call from the patched and add a line that patches `mknod` again if you like.

> Also, is there some kind of test we could write for this?

I fixed the fact that the test wasn't checking whether all write operations (in the sense that they write to the filesystem, not literally `write()`) call `IDBFS.queuePersist` (I tested that the test now fails with my fix to `mknod` reverted). However, I only fixed it for `mkdir`, and the fix is somewhat hacky. Ideally we want to be sure that every write operation is tested, but splitting the `test` function at every write operation would make an unmaintainable mess. JSPI would help a lot (since it uses stack switching under the hood without requiring any changes to the code), but I'm not sure how to integrate it and it's not very well supported by browsers at the moment.

test/fs/test_idbfs_sync.c
handlerug: This is the first part of the test, so it checks for non-existence of the directory before creating it. It's still the "filesystem seeding" part of code as before, now it's just split before mkdir so we can make sure mkdir calls `IDBFS.queuePersist`. See the part after `#else` for the code that checks whether `/working1/dir2` actually got persisted.

handlerug: FYI I also updated the commit message and the PR description with a more thorough explanation of why this is needed.

sbc100: I find it a little hard to understand with the `#ifdef FIRST` straddling the function boundary here.   I wonder if we could structure this to avoid that.

Also I opened #24820 to make this test more readable in general.  The second one to land will need to deal with some conflicts I guess.  I'm happy to wait if you prefer.

test/fs/test_idbfs_sync.c
```diff
@@ -67,12 +67,32 @@ void test() {
...
     result = -12000 - errno;
   else if (mkdir("/working1/dir", 0777) != 0)
     result = -13000 - errno;
 
+  EM_ASM(
+    setTimeout(() => {
+      // Run the remainder of the test after all already scheduled JavaScript
+      // tasks have executed (such as the one scheduled by IDBFS.queuePersist).
+      // This is to ensure that mkdir also triggers a call to queuePersist and
+      // to isolate mkdir from queuePersist invocations made by other syscalls.
```
sbc100: I'm afraid I still don't understand why this is needed.   Why can't we run the remainder of this test directly?    How can we tell that the mkdir triggers a call to `queuePersist`?

handlerug: When a write operation is invoked, it calls `IDBFS.queuePersist()`, which schedules a sync. The important thing is that the sync operation doesn't happen synchronously, partly because of the intention to buffer up the writes and sync them all in one go via the usage of `setTimeout` to schedule the sync operation, partly because the IndexedDB APIs are asynchronous themselves so even if `setTimeout` wasn't used this would still be the case.

`setTimeout`:

https://github.com/emscripten-core/emscripten/blob/9886169e009c1bb1d8461733edf47c73308b88eb/src/lib/libidbfs.js#L36-L41

Asynchronous code (which isn't limited to this bit of code, it's just how the IndexedDB API works):

https://github.com/emscripten-core/emscripten/blob/9886169e009c1bb1d8461733edf47c73308b88eb/src/lib/libidbfs.js#L100-L113

That means--and this might be an incorrect assumption on my part--that while the synchronous C code is running (or rather while the JS that invokes the Wasm stuff is running), the JavaScript event loop, the "main thread", is blocked. No `setTimeout` or promise callbacks can run, the page is frozen, everything is waiting for the current task to stop executing and yield so that execution can switch elsewhere (e.g. back into userland code like `setTimeout`, promise, IndexedDB, or other async API callbacks, or yield into the browser).

The implication of that is, if we don't yield between write operations, they will be executed within a single task. If any of those write operations call `IDBFS.queuePersist()`, it will schedule a task to run after ours, which will sync the whole memory FS (in the state after we yielded from C) to IDB. That means that if just one of the write operations call `IDBFS.queuePersist()`, all of the write calls that happened synchronously during the same task will get persisted.

This is bad for testing because we want to ensure that every write operation calls `IDBFS.queuePersist()`. I'm not entirely sure what is the best way to check for that. The way I did it here is specific to the `mkdir` bug--I made sure that no `IDBFS.queuePersist` was scheduled at the time of the `mkdir` call by letting the JS engine run pending tasks (one of which is the `queuePersist` one). Ideally we want to test every write operation. The end-to-end approach could be to make a write call, then inside `setTimeout` call `syncfs(true)`, then call `stat`. Right now I assume the Emscripten runtime is teared down and restarted after the first half of the test is done and before the second half starts.

Does this make sense?

sbc100: Yes I think that makes sense.

What you are trying to do isolute the mkdir so that it doesn't run alongside any other operations.

How about we add another phase to the test.   So you have FIRST and SECOND.   In the SECOND phase you could call *only* mkdir.   After the second phase, without your fix, the mkdir would be lost I assume?

handlerug: What do you think about making a separate test with two phases each for each write operation that should auto-persist? I think it makes sense since we are trying to verify each write operation's correctness, and it might be nicer than arbitrarily splitting an existing test.

handlerug: > After the second phase, without your fix, the mkdir would be lost I assume?

Yup.

sbc100: > What do you think about making a separate test with two phases each for each write operation that should auto-persist? I think it makes sense since we are trying to verify each write operation's correctness, and it might be nicer than arbitrarily splitting an existing test.

I'm not sure it worth creating that many new tests.. unless we can do it in way that its expressed clearly and concisely.  I'd rather not have N different tests here if we can avoid it.

handlerug: Alright, I've pushed a commit that adds a separate test. I tried to keep it clean. It seems worth it to me to have a test that actually checks if every operation persists or not. What do you think?

test/fs/test_idbfs_autopersist.c
```diff
@@ -0,0 +1,208 @@
...
+  unlink("/working1/file_renamed");
+  unlink("/working1/symlink");
+  rmdir("/working1/dir");
+}
+
+void test() {
+#ifdef TEST_PREPARE
+  prepare();
+  goto end;
+#endif
```
sbc100: Instead of `goto end` here can you just use `#else` and have `#endif` instead of the end label?

test/test_browser.py
```diff
@@ -1353,6 +1353,32 @@ def test_fs_idbfs_sync(self, args):
     print('done first half')
     self.btest_exit('fs/test_idbfs_sync.c', cflags=['-lidbfs.js', f'-DSECRET="{secret}"', '-sEXPORTED_FUNCTIONS=_main,_test,_finish', '-lidbfs.js'] + args)
 
+  def test_fs_idbfs_autopersist(self):
+    common_cflags = ['-lidbfs.js', '-sEXPORTED_FUNCTIONS=_main,_test,_finish']
```
sbc100: You can just do `self.cflags +=` here instead of having `common_cflags`.

test/test_browser.py
```diff
@@ -1353,6 +1353,32 @@ def test_fs_idbfs_sync(self, args):
...
+    ]
+
+    def get_defines(current_case_name, phase):
+      return [
+        f'-DTEST_CASE={test_cases.index(current_case_name) if current_case_name in test_cases else -1}',
+        f'-DTEST_PHASE={phase}',
+      ]
+
+    print('preparing')
+    self.btest_exit('fs/test_idbfs_autopersist.c', cflags=common_cflags + get_defines(None, 0) + ['-DTEST_PREPARE'])
```
sbc100: If you use `if/else/endif` as mentioned about you can just remove `get_defines(None, 0)` completely here.

Then you can do `self.cflags += f'-DTEST_CASE={test_cases.index(current_case_name)}` here on this line and remove the `get_defines` function completely.

Then you can just do `cflags=[f'-DTEST_PHASE=1]'` and  `cflags=[f'-DTEST_PHASE=2']` below.

test/test_browser.py
```diff
@@ -1353,6 +1353,32 @@ def test_fs_idbfs_sync(self, args):
...
+  def test_fs_idbfs_autopersist(self):
+    common_cflags = ['-lidbfs.js', '-sEXPORTED_FUNCTIONS=_main,_test,_finish']
+    test_cases = [
+      'open',
+      'close',
+      'symlink',
+      'unlink',
+      'rename',
+      'mkdir',
+    ]
```
sbc100: Instead of this list here you can use `parameterize` to have the `case` case passed in as an argument.

Then each case gets its own test that can be run and debugged on its own.

src/lib/libidbfs.js
```diff
@@ -54,7 +54,7 @@ addToLibrary({
       // If the automatic IDBFS persistence option has been selected, then automatically persist
       // all modifications to the filesystem as they occur.
       if (mount?.opts?.autoPersist) {
-        mnt.idbPersistState = 0; // IndexedDB sync starts in idle state
+        mount.idbPersistState = 0; // IndexedDB sync starts in idle state
```
sbc100, line 57: Is this a separate bug?   What was the effect of this bug?

handlerug: Yes. The effect wasn't observable to users because the code doesn't strictly check for equality with `0` and `undefined` is considered falsy: https://github.com/emscripten-core/emscripten/blob/052cd1a081d2194c1db4634e5f582e90e80f8750/src/lib/libidbfs.js#L35

This fix is needed for the `runOnceIDBFSIdle` test helper.

test/test_browser.py
```diff
@@ -1353,6 +1353,19 @@ def test_fs_idbfs_sync(self, args):
...
+    'close': ('close', 3),
+    'symlink': ('symlink', 3),
+    'unlink': ('unlink', 3),
+    'rename': ('rename', 3),
+    'mkdir': ('mkdir', 2),
+  })
+  def test_fs_idbfs_autopersist(self, test_case, phase_count):
+    self.cflags += ['-lidbfs.js', '-sEXPORTED_FUNCTIONS=_main,_test,_finish', f'-DTEST_CASE=TEST_CASE_{test_case}']
+    for phase in range(phase_count):
+      self.btest_exit('fs/test_idbfs_autopersist.c', cflags=[f'-DTEST_PHASE={phase+1}'])
```
sbc100: Much better!

handlerug:
Pushed new code with your suggestions, except now that the test is parameterized there isn't a clean way to run some code (`-DTEST_PREPARE`) before any of the subtests, so I got rid of the separate "preparation" phase and added a new phase for each test case where preparation is necessary. Now each test case starts from a clean filesystem.

test/fs/test_idbfs_autopersist.c
```diff
@@ -0,0 +1,237 @@
...
+    default:
+      assert(false);
+  }
+}
+
+void finish() {
+  emscripten_force_exit(0);
+}
+
+void test() {
```
sbc100: You can mark `finish` and `test` as `EMSCRIPTEN_KEEPALIVE` and avoid passing `-sEXPORTED_FUNCTIONS` on the command line.

test/fs/test_idbfs_autopersist.c
```diff
@@ -0,0 +1,237 @@
...
+
+EM_JS_DEPS(deps, "$callUserCallback");
+
+enum {
+  TEST_CASE_open,
+  TEST_CASE_close,
+  TEST_CASE_symlink,
+  TEST_CASE_unlink,
+  TEST_CASE_rename,
+  TEST_CASE_mkdir,
```
sbc100: Make these all caps.

test/test_browser.py
```diff
@@ -1353,6 +1353,19 @@ def test_fs_idbfs_sync(self, args):
     print('done first half')
     self.btest_exit('fs/test_idbfs_sync.c', cflags=['-lidbfs.js', f'-DSECRET="{secret}"', '-sEXPORTED_FUNCTIONS=_main,_test,_finish', '-lidbfs.js'] + args)
 
+  @parameterized({
+    'open': ('open', 2),
+    'close': ('close', 3),
+    'symlink': ('symlink', 3),
+    'unlink': ('unlink', 3),
+    'rename': ('rename', 3),
+    'mkdir': ('mkdir', 2),
```
sbc100: Use uppercase here too.. maybe use the full name so it obvious what is going on?  e.g. `open': ('TEST_CASE_OPEN', 2),`

handlerug: Sounds sensible. Initially I didn't want to do it since it would be repetitive, but I think it's fine.

test/fs/test_idbfs_autopersist.c
handlerug: Yeah I actually wanted to use that at first, but `test_idbfs_sync.c` doesn't use `EMSCRIPTEN_KEEPALIVE` and I figured I better stick to the existing style.

sbc100: Feel free to update that test too.. otherwise we can do it as a followup.

handlerug: I think I'd rather do it in a follow-up PR

handlerug: https://github.com/emscripten-core/emscripten/pull/24864

handlerug: (Not sure if you realized but I have already applied this suggestion to this PR, I should've worded my comment more clearly. Unless you just haven't gotten around to reviewing in which case no worries)

sbc100 (approved):

Issue #24691, opened 2025-07-11
Allow custom scripts to be injected into default shell file


arsnyder16:
I have some situations where i am using the file packager script to generate some data files. 

In order to inject these data files in the shell html i need make my own copy, which i would rather not do since i am now maintaining a snapshot of code that i would prefer to always be aligned with emscripten. 

It would be nice to allow the default shell file to be used by adding an additional arguments during linking 

`--pre-script=otherdata.js --pre-script=yetanotherdata.js`
```html
    <script async type="text/javascript" src="otherdata.js"></script>
    <script async type="text/javascript" src="yetanotherdata.js"></script>
```

These would then be injected in the shell file as part of the generating the replacement for 
```
{{{ SCRIPT }}}
``` 
or a new replacement value like:
```
    {{{ PRE_SCRIPT }}}
    {{{ SCRIPT }}}
```

sbc100:
Firstly, is there some reason why `--pre-js` `--post-js` doesn't work for you?

Secondly, is there some reason why you are having emscripten generate the html for you?   We find that most folks only use the generated html during the prototyping phase and then end up creating the html themselves and having emscripten only generate .js and .wasm file.   

If you do need emscripten to generate html for you (for example if you have lots of small programs that each need their own html file), presumably you already want to customize it don't you?   I imagine you don't the emscripten logo and other things, right?

arsnyder16:
As far as `--pre-js` `--post-js`, Do you have an example of loading the `file_packager.js` dynamically in those callbacks? How do i avoid a race condition with the doRun and those scripts loading.  I find that i need to make sure these `file_packager.py` generated javascript files need to be inserted before the main javascript file. That ensures the scripts are executed in order they are defined in the html file 

I am just wrapping some simple gtest executables, so the html isn't that important but the default functionality is nice since it shows the console output.

sbc100:
You can use emscripten `--embed-file` and/or `--preload-file` to automatically run file_packager.js internally and embed the results in the pre-js.  i.e. pre-js is the default place where file_packager.js output is designed to go.

The code generated by file_packaged uses the addRunDependency mechanism to delay the running of the program until the files are ready.

arsnyder16:
Yes this is my current approach. I have many artifacts for my gtests that are either input or archived snapshots of results. Unfortunately my data file is becoming larger than 2Gi so I am hitting a limit.



[...]

PR #24802, opened 2025-07-29
[file_packager] Split data files when file size exceeds ArrayBuffer limit

arsnyder16:
Currently ArrayBuffer cannot be larger than 2046 on chrome. So bundling a large amount of files using file_packager.py the `.data` will not be allowed to load. Breaking up files into multiple data files by passes this issue. 
```
error while handling : http://localhost:3040/Gtest.data Error: Unexpected error while handling : http://localhost:3040/Gtest.data 
RangeError: Array buffer allocation failed
```

Note this was a solution proposed for issues described in this thread #24691


arsnyder16:
@sbc100 Is this aligned with what you were thinking?

tools/file_packager.py
```diff
@@ -459,23 +460,25 @@ def main():  # noqa: C901, PLR0912, PLR0915
...
+    elif leading == 'preload-size-limit':
+      options.preload_size_limit = int(arg.split('=', 1)[1] if '=' in arg else arg)
     else:
       err('Unknown parameter:', arg)
       return 1
 
-  options.has_preloaded = any(f.mode == 'preload' for f in data_files)
-  options.has_embedded = any(f.mode == 'embed' for f in data_files)
+  has_preloaded = any(f.mode == 'preload' for f in data_files)
+  has_embedded = any(f.mode == 'embed' for f in data_files)
```
sbc100: This change looks unrelated?  Is so perhaps make this a separate PR?

tools/file_packager.py
```diff
@@ -650,20 +673,22 @@ def generate_js(data_target, data_files, metadata):
           code += ('''Module['FS_createPath'](%s, %s, true, true);\n'''
                    % (json.dumps('/' + '/'.join(parts[:i])), json.dumps(parts[i])))
           partial_dirs.append(partial)
-
-  if options.has_preloaded:
+  has_preloaded = any(f.mode == 'preload' for f in data_files)
+  has_embedded = any(f.mode == 'embed' for f in data_files)
```
sbc100: Can these values change from above to here?   Surely the the mode of the data file is fixed and doesn't change?

tools/file_packager.py
```diff
@@ -650,20 +673,22 @@ def generate_js(data_target, data_files, metadata):
...
     with open(data_target, 'wb') as data:
-      for file_ in data_files:
-        file_.data_start = start
-        curr = utils.read_binary(file_.srcpath)
-        file_.data_end = start + len(curr)
-        if AV_WORKAROUND:
-            curr += '\x00'
-        start += len(curr)
-        data.write(curr)
+      if file_.mode == 'preload':
```
sbc100: I guess the old code assume that either all files were preloads or embedded, and not a combination of both?

Do you want to be able to do both at the same time?  I think maybe we should just make that invalid?

sbc100 (commented):
Can we add some tests for this?     What are the specific limits you are running into?   Are those limits not fixed?  i.e. does it need to be configurable?

tools/file_packager.py
arsnyder16: It is related. Basically if you are mixing preload and embed, we can't assume that all data_files that you start with are part of the each bundle. So during `generate_js` we need to look at the data files being used and reevaluate.  I can remove but i would think i am leaving gaps in in the logic

tools/file_packager.py
arsnyder16: looks like we allow a mix of preloaded and embeded, each chunk could have any mix of these modes so we need to evaluate independently

sbc100: How about we disallow combining the two modes?

tools/file_packager.py
arsnyder16: maybe that warrants a separate PR since that is a breaking change. I will leave that to the maintainer to introduce the breaking change

sbc100: Absolutely, yes.  I'm looking into it now.

arsnyder16:
> Can we add some tests for this? What are the specific limits you are running into? Are those limits not fixed? i.e. does it need to be configurable?

I wanted to make sure i was on the right track before adding tests. 
The limits are described in conversation in the issue i mentioned. We can hard code the limit to 2Gi as a likely guess at a true limit. If that is what you prefer its simpler from testing perspective . Seemed more flexible to make it configurable to anyone calling the utility, but i get the simplicity side of it

sbc100:
This does seem like a reasonable approach.

I'm still a little fuzzy on exactly why and when this might be needed in the real world, I think I need to go re-read our original discussion, but I also think including more information in the PR (i.e. in the description, or in comments) would be good.

arsnyder16:
> This does seem like a reasonable approach.
> 
> I'm still a little fuzzy on exactly why and when this might be needed in the real world, I think I need to go re-read our original discussion, but I also think including more information in the PR (i.e. in the description, or in comments) would be good.

Updated the description, based on that if you want me to hard code the limit into the package i can take that approach

tools/file_packager.py
arsnyder16: @sbc100 Looks like the plan is to deprecate support for combing of these two modes.  I adjusted this PR so the splitting only applies when you are exclusively preloaded

arsnyder16:
@sbc100 Going to look at adding tests for this. Do you have an recommendations? Would you like me to generate temporary files (s) so that i can reach the 2Gi limit?

arsnyder16:
@sbc100 This should be ready for review, not sure about the test failure if they are flaky or if its related to my change. I don't see them fail locally.

juj:
Peculiarly I find that in Chrome, the max ArrayBuffer limit is not 2GB (2048MB), but 2046MB:

image

Firefox supports up to 16GB ArrayBuffers:

image

In node.js I get up to 4GB ArrayBuffers:

image

What is your target environment?

arsnyder16:
@juj Thanks! My target is chrome, i incorrectly associated in the ArrayBuffer documentation `[Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/ArrayBuffer#exceptions) to `MAX_INT`.  I didn't specifically test the exact limit, rather was just trying to follow the documentation but clearly misinterpreted it.   

Looks like Chrome has the lowest limit, i will change to use that limit. 

Safari appears to be 4Gi
image

arsnyder16:
@juj @sbc100 Anything else on this?

test/test_other.py
```diff
@@ -7452,14 +7458,61 @@ def test_underscore_exit(self):
...
-    create_file('huge.dat', 'a' * (1024 * 1024 * 257))
+    self.create_huge_file('huge.dat', 1024 * 1024 * 257)
     create_file('tiny.dat', 'a')
     err = self.run_process([FILE_PACKAGER, 'test.data', '--preload', 'tiny.dat'], stdout=PIPE, stderr=PIPE).stderr
     self.assertNotContained(MESSAGE, err)
     err = self.run_process([FILE_PACKAGER, 'test.data', '--preload', 'huge.dat'], stdout=PIPE, stderr=PIPE).stderr
     self.assertContained(MESSAGE, err)
     self.clear()
 
+  def test_file_packager_huge_no_split(self):
```
juj: Could you add short comments before each function, that describe the test intent (why/what is the purpose for the test to exist?)

E.g.

```
# Verify that when packaging up to 2046MB of data, file packager should not split up the generated data file.
```
and
```
# Verify that when size exceeds 2046MB, file packager should split up the generated data file into two.
```
and so on?

test/test_other.py
```diff
@@ -7452,14 +7458,61 @@ def test_underscore_exit(self):
...
+    for i in range(7):
+      self.create_huge_file(f'huge{i}.dat', 1024 * 1024 * 256)
+    self.create_huge_file('huge8.dat', 1024 * 1024 * 254)
+    err = self.run_process([FILE_PACKAGER, 'test.data', '--preload', 'huge8.dat'] + [f'huge{i}.dat' for i in range(7)], stdout=PIPE, stderr=PIPE).stderr
+    self.assertContained('warning: file packager is creating an asset bundle of 2046 MB. this is very large, and browsers might have trouble loading it', err)
+    self.assertExists('test.data')
```
juj, line 6485: here also maybe `self.assertNotExists('test_1.data')`?

tools/file_packager.py
```diff
@@ -544,15 +547,60 @@ def was_seen(name):
...
+  if options.has_preloaded and not options.has_embedded:
+    file_chunks = [[]]
+    current_size = 0
+    for file_ in data_files:
+      fsize = os.path.getsize(file_.srcpath)
+      if current_size + fsize <= PRELOAD_DATA_FILE_LIMIT:
+        file_chunks[-1].append(file_)
+        current_size += fsize
+      elif fsize > PRELOAD_DATA_FILE_LIMIT:
+        diagnostics.error('error: cannot package file greater than %d MB does not exist' % (PRELOAD_DATA_FILE_LIMIT / (1024 * 1024)))
```
juj: Can you rephrase this error? I'm not sure how to parse it.

Maybe the intent was `error: cannot package file <filename>, which is larger than maximum individual file size limit %d MB.`?

tools/file_packager.py
```diff
@@ -544,15 +547,60 @@ def was_seen(name):
...
+    ret = generate_js(data_file, data_files, metadata)
+    if options.force or len(data_files):
+      if options.jsoutput is None:
+        print(ret)
+      else:
+        # Overwrite the old jsoutput file (if exists) only when its content
+        # differs from the current generated one, otherwise leave the file
+        # untouched preserving its old timestamp
+        base, ext = options.jsoutput.rsplit('.', 1)
+        js_file = f"{base}{f'_{counter}' if counter else ''}.{ext}"
```
juj: Maybe this 

```py
        base, ext = options.jsoutput.rsplit('.', 1)
        js_file = f"{base}{f'_{counter}' if counter else ''}.{ext}"
```
could go in a helper? Given it is used twice.

tools/file_packager.py
```diff
@@ -544,15 +547,60 @@ def was_seen(name):
...
+        targets.append(js_file)
+        if os.path.isfile(js_file):
+          old = utils.read_file(js_file)
+          if old != ret:
+            utils.write_file(js_file, ret)
+        else:
+          utils.write_file(js_file, ret)
```
juj: Also here this could be nice to be located into a helper function `write_file_if_changed(...)`, or similar to make the main body of the logic here lighter to follow?

test/test_other.py
```diff
@@ -7452,14 +7458,61 @@ def test_underscore_exit(self):
...
+    self.assertEqual(os.path.getsize('test.data'), (1024 * 1024 * 256) * 7)
+    self.assertEqual(os.path.getsize('test_1.data'), (1024 * 1024 * 254) + 1)
+    self.clear()
+
+  def test_file_packager_huge_split_too_large(self):
+    self.create_huge_file('huge.dat', (1024 * 1024 * 1024) + ((1022 * 1024 * 1024) + 1))
+    proc = self.run_process([FILE_PACKAGER, 'test.data', '--preload', 'huge.dat'], check=False, stdout=PIPE, stderr=PIPE)
+    self.assertEqual(proc.returncode, 1)
+    self.assertContained('error: cannot package file greater than 2046 MB does not exist', proc.stderr)
+    self.clear()
```
juj, line 6556: It looks like the above tests verify that file packager does split up files, though there is no functional test to verify that e.g. the bytes were split appropriately, and that the final end result loads up properly? Would that be important to test?

arsnyder16: I am not sure how critical it is, looks like all the file_packager logic is part of these unit tests, seems like out of scope for this PR to add tests that verify the loading of the file_packager results

sbc100: There are many tests that verify load of file packager results.   Basically any test that uses `emcc` with `--preload-file` is verifying this.

There are also tests that call `FILE_PACKGER` directly and then execute the result.  See `test_file_packager_separate_metadata` for example.

arsnyder16: @sbc100 Thanks for pointing out these decks, `test_file_packager_separate_metadata` was a good example to follow to test this logic

arsnyder16: @sbc100 @juj I updated to verify the package(s) load but on CircleCI they seem to be crashing immediately. I suspect OOMKilled. They run fine for me locally

arsnyder16: @juj @sbc100 Any guidance here?

arsnyder16: @juj @sbc100 Could use some guidance here to get this through the finish line, now that i addressed the feedback to actually load the results in these tests they cause the circleci processes to run out of memory.

arsnyder16:
@juj @sbc100 Circling back here, is it possible to work this in? We were close before i just needed guidance on how to handle the CI machine hitting OOMKilled

test/test_other.py
```diff
@@ -6381,14 +6387,164 @@ def test_underscore_exit(self):
...
+      assert(stat("huge3.dat", &buf) == 0);
+      assert(buf.st_size == 1024 * 1024 * 256);
+      assert(stat("huge4.dat", &buf) == 0);
+      assert(buf.st_size == 1024 * 1024 * 256);
+      assert(stat("huge5.dat", &buf) == 0);
+      assert(buf.st_size == 1024 * 1024 * 256);
+      assert(stat("huge6.dat", &buf) == 0);
+      assert(buf.st_size == 1024 * 1024 * 256);
+      assert(stat("huge7.dat", &buf) == 0);
+      assert(buf.st_size == 1024 * 1024 * 254);
```
kripken: There is a lot of repetition in these tests, that we can maybe improve?

1. This could be a loop.
2. The very similar tests could all use the same core, with `parameterized` to differentiate them, avoiding duplication of code between them?

arsnyder16: Feedback addressed. Still looking for guidance how to make these run on the CI machines since the js process seems to be getting OOMKilled.  These run fine for me locally

arsnyder16: One way sort of around this is instead of trying to lock to the chrome limit
```
# chrome limit is 2MB under 2Gi
PRELOAD_DATA_FILE_LIMIT = 2046 * 1024 * 1024
```

Instead just created an arbitrary limit for splitting say 512MB/1Gi going with an approach like that would allow me to test the behavior without going to the more extreme limit.  It does seem a little odd to try and align with chrome exactly

arsnyder16: I suppose adding a new argument to the packager to set this limit will allow the tests to be written at a lower limit while leaving the default at the chrome limit

kripken: Does it OOM when running the file packager, or when running the output in a VM?

tools/file_packager.py
```diff
@@ -545,13 +548,53 @@ def was_seen(name):
...
+    if len(file_chunks) > 1:
+      diagnostics.warn('warning: file packager is splitting bundle into %d chunks' % len(file_chunks))
+
+    for counter, data_files in enumerate(file_chunks):
+      metadata = {'files': []}
+
+      def construct_data_file_name(base, ext):
+        return f"{base}{f'_{counter}' if counter else ''}.{ext}"
+      data_file = construct_data_file_name(*data_target.rsplit('.', 1))
+      js_file = None if options.jsoutput is None else construct_data_file_name(*options.jsoutput.rsplit('.', 1))
```
sbc100: Can you write this as a multi-line if/else?

sbc100: Should this be `os.path.splitext`?

tools/file_packager.py
```diff
@@ -88,6 +88,9 @@
 
 DEBUG = os.environ.get('EMCC_DEBUG')
 
+# chrome limit is 2MB under 2Gi
+PRELOAD_DATA_FILE_LIMIT = 2046 * 1024 * 1024
```
sbc100: Yes, perhaps for testing you could set this via and environment variable?

e.g. `EM_FILE_PACKAGER_MAX_CHUNK_SIZE_MB=10`

test/test_other.py
arsnyder16: When running the output in a vm

tools/file_packager.py
arsnyder16: Yea that's a good call, I will make that change

test/test_other.py
kripken: If the OOM is only when running, then an option would be to skip that part in the testing. That is, we could test that the packager emits the right stuff, but not run it.

But, if you and @sbc100 like the env var idea, sgtm

arsnyder16: Yea it was suggested early on to make sure we test that the js is emitted correctly and will load each chunk. I think there is a lot of value in making sure that is covered. The environment variable is low risk and it does inherently build in some customization/flexibility for the future if different environments have different constraints

tools/file_packager.py
arsnyder16: cleaned up

arsnyder16: done

arsnyder16:
@sbc100 @kripken All feedback addressed.

kripken (approved):

Issue #24803, opened 2025-07-29
Does the file packager need to simultanious embedding and preloading


sbc100:
Supporting both at the same time complicates the code and I would hope this configuration is not necessary or useful.

We do have at least one test for this configuration: https://github.com/emscripten-core/emscripten/blob/ee5848479f9521bbb5b33c6d8c8f9158f1551ce8/test/test_other.py#L1951

This came up while reviewing #24802

sbc100:
Looks like the test was added in #17457 because an interactive test (test_sdl_audio) does this...

PR #24806, opened 2025-07-29
[file-packager] Warn about mixing --preload and --embed in the same command. NFC

sbc100:
See #24803

### DIFF
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -466,6 +466,11 @@ def main():  # noqa: C901, PLR0912, PLR0915
   options.has_preloaded = any(f.mode == 'preload' for f in data_files)
   options.has_embedded = any(f.mode == 'embed' for f in data_files)
 
+  if options.has_preloaded and options.has_embedded:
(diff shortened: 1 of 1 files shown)


tools/file_packager.py
```diff
@@ -466,6 +466,9 @@ def main():  # noqa: C901, PLR0912, PLR0915
   options.has_preloaded = any(f.mode == 'preload' for f in data_files)
   options.has_embedded = any(f.mode == 'embed' for f in data_files)
 
+  if options.has_preloaded and options.has_embedded:
+    diagnostics.warn('file packager support for --preload and --embed is scheduled for deprecation.  Is you need this feature please comment at https://github.com/emscripten-core/emscripten/issues/24803')
```
```suggestion
    diagnostics.warn('file packager support for using both --preload and --embed at the same time is scheduled for deprecation.  Is you need this feature please comment at https://github.com/emscripten-core/emscripten/issues/24803')
```

kripken: Just to make sure this isn't read as "we are deprecating `--preload` as well as `--embed`"

PR #24807, opened 2025-07-29
Bump minimum supported node version 10.19.0 -> 12.22.9

sbc100:
This is the minimum version that we support for running the generated code.  The version is old enough to cover both Ubuntu LTS (22.04) (Jammy) and Debian/stable (bookworm).


tools/feature_matrix.py
```diff
@@ -21,9 +21,9 @@
 OLDEST_SUPPORTED_CHROME = 55  # December 1, 2016
 OLDEST_SUPPORTED_FIREFOX = 50  # November 15, 2016
 OLDEST_SUPPORTED_SAFARI = 101000  # September 20, 2016
-# 10.19.0 is the oldest version of node that we do any testing with.
+# 12.22.9 is the oldest version of node that we do any testing with.
 # Keep this in sync with the test-node-compat in .circleci/config.yml.
-OLDEST_SUPPORTED_NODE = 101900
+OLDEST_SUPPORTED_NODE = 122209 # Janurary 11, 2022
```
kripken, line 26: Is there a reason to keep node so much more recent (2022) than the browsers (2016)?

Or, is there a particular benefit to bumping node from 10 to 12?

sbc100: I'm trying to remove the complications around the `--experimental-wasm-bigint` flag.  Right now we have to worry about 3 different cases:

1. Really old versions of node that don't support that flag
2. Somewhat old version that support the feature but require the flag
3. New versions don't support that flag because its now the default.

But bumping to 12 we no longer need to worry bout the first case which simplifies things somewhat.

Also, I don't think we have any reason to support older versions.

sbc100: > Is there a reason to keep node so much more recent (2022) than the browsers (2016)?

I think the reason here is that folks who run servers don't tend to let them get quite that out of date.   I don't think we have any server-side users who are running such ancient distros.. but I could be wrong.

test/common.py
```diff
@@ -1228,11 +1227,6 @@ def setUp(self):
         self.cflags += building.get_emcc_node_flags(node_version)
         self.cflags.append('-Wno-transpile')
 
-      # This allows much of the test suite to be run on older versions of node that don't
-      # support wasm bigint integration
-      if node_version[0] < feature_matrix.min_browser_versions[feature_matrix.Feature.JS_BIGINT_INTEGRATION]['node'] / 10000:
-        self.cflags.append('-sWASM_BIGINT=0')
```
sbc100, line 1234: I'm particularly interested in delete this gross couple of lines.. here we are basically saying that if the node version we are testing against is old enough we change the default settings for `-sWASM_BIGINT`... which seems very hacky.

tools/feature_matrix.py
kripken: Yeah, my concern is that I feel I know more about web users than node users... I'm really not sure what kind of node versions people use. And 2022 feels fairly recent to me, though that's subjective.

kripken: Maybe worth asking on the mailing list before making this change?

sbc100:
This change ended up landing as #25375

PR #24808, opened 2025-07-29
Avoid injecting `--experimental-wasm-bigint` into `#!` line when creating node-runnable output

sbc100:
This was the last remaining usage of `node_bigint_flags` in the compiler so this function
was moved into test code where it is still used.

IIUC its better not to do this at all for a few reasons:

1. The version of node used in the emscripten compiler is not the same as the version that will be used the run the output program, so basing these flags on `settings.NODE_JS` is just wrong.
2. If folks want to target older versions of node we already have ways to do that `-sMIN_NODE_VERSION`.  They can also add this flag themselves when running the output program.  I don't think is out business to inject node flags here.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -22,8 +22,8 @@ See docs/process.md for more on how version tagging works.
 -----------------------
 - The `#!` line that emscripten, under some circumstances, will add to the
   generated JS code no longer injects the `--experimental-wasm-bigint` node
-  flag.  This flag is not needed on recent versions of node, so its not possible
-  to know if its safe to include. (#24808)
+  flag.  This flag is not needed on recent versions of node, and in fact
+  errors there, so it's not possible to know if it's safe to include. (#24808)
 - In `-sMODULARIZE` mode the factory function will now always return a promise,
   even when `WASM_ASYNC_COMPILATION` is disabled.  This is because emscripten
   has other features that might also return async module creation (e.g. loading
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,10 @@ See docs/process.md for more on how version tagging works.
 
 4.0.12 (in development)
 -----------------------
+- The `#!` line that emscripten, under some circumstances, will add to the
+  generated JS code no longer injects the `--experimental-wasm-bigint` node
+  flag.  This flag is not needed on recent versions of node, so its not possible
+  to know if its safe to include. (#24808)
 - In `-sMODULARIZE` mode the factory function will now always return a promise,
   even when `WASM_ASYNC_COMPILATION` is disabled.  This is because emscripten
   has other features that might also return async module creation (e.g. loading
--- a/tools/link.py
+++ b/tools/link.py
@@ -467,8 +467,6 @@ def get_binaryen_passes():
 def make_js_executable(script):
   src = read_file(script)
   cmd = config.NODE_JS
-  if settings.WASM_BIGINT:
-    cmd += shared.node_bigint_flags(config.NODE_JS)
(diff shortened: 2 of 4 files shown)


sbc100:
The one test failure here is fixed in #24810

kripken:
> I don't think is out business to inject node flags here.

I'm not sure I agree - doesn't it make sense to add flags when it makes things just work, and when we know based on flags what will make things work?

Is this brittle for some reason?

sbc100:
> > I don't think is out business to inject node flags here.
> 
> I'm not sure I agree - doesn't it make sense to add flags when it makes things just work, and when we know based on flags what will make things work?
> 
> Is this brittle for some reason?

The problem is that we have no idea what version of node the final output will run on.

Should we inject all possible flags here that we might need?  What then happens in your run the script on a version of node that is either too old (i.e. before the flag existed) or too new (i.e. once the flag is removed).  Including flags like there here would cause the script to fail to start.

Also, if somebody wants a script to run on an older version of node they should really be setting MIN_NODE_VERSION to a version before bigint support was added, then the script would not depend on bigint at all.

sbc100:
> > > I don't think is out business to inject node flags here.
> > 
> > 
> > I'm not sure I agree - doesn't it make sense to add flags when it makes things just work, and when we know based on flags what will make things work?
> > Is this brittle for some reason?
> 
> The problem is that we have no idea what version of node the final output will run on.
> 
> Should we inject all possible flags here that we might need? What then happens in your run the script on a version of node that is either too old (i.e. before the flag existed) or too new (i.e. once the flag is removed). Including flags like there here would cause the script to fail to start.
> 
> Also, if somebody wants a script to run on an older version of node they should really be setting MIN_NODE_VERSION to a version before bigint support was added, then the script would not depend on bigint at all.

Remember the only thing we know about the version of node on which the script will be run is `MIN_NODE_VERSION`.   Is that version doesn't support bigint then we already disable this feature.  If that version does support bigint then we don't need the flag.  Trying to support the version where it was available but behind a flag is not, I believe, practical given the current inputs we have.

To achieve this we could write a more complex launcher script that is able to detect the runtime version of node and pass these optional/experimental flags, but I don't think its worth it to do that.

kripken:
> To achieve this we could write a more complex launcher script that is able to detect the runtime version of node and pass these optional/experimental flags, but I don't think its worth it to do that.

Right, good point. If we wanted to add flags, we should really do that, so things fully work. But I agree that is excessive.

tools/link.py
```diff
@@ -467,8 +467,6 @@ def check_human_readable_list(items):
 def make_js_executable(script):
   src = read_file(script)
   cmd = config.NODE_JS
-  if settings.WASM_BIGINT:
-    cmd += shared.node_bigint_flags(config.NODE_JS)
```
kripken, line 471: Let's document this in the changelog? lgtm otherwise.

sbc100:
I split out the purely NFC part: #24815

ChangeLog.md
```diff
@@ -20,6 +20,10 @@ See docs/process.md for more on how version tagging works.
 
 4.0.12 (in development)
 -----------------------
+- The `#!` line that emscripten, under some circumstances, will add to the
+  generated JS code no longer injects the `--experimental-wasm-bigint` node
+  flag.  This flag is not needed on recent versions of node, so its not possible
+  to know if its safe to include. (#24808)
```
```suggestion
  flag.  This flag is not needed on recent versions of node, and in fact
  errors there, so it's not possible to know if it's safe to include. (#24808)
```

kripken (approved):
lgtm % comment

PR #24811, opened 2025-07-29
[embind] Add pointer policies for setting val object properties.

brendandahl:
After #24175, `val` required a pointer policy when creating a `val` with a pointer. In `val.set(key, value)`, a temporary `val` object is created, but there was no way to set the policy from user code. This patch allows the user to pass in `allow_raw_pointers()` to enable this.


brendandahl:
an aside: @RReverser do you have any thoughts on how to add policies to to val's `call` and `operator()`. Unfortunately, they both take variadic arguments already, so I don't see a great way to add on a `Policy...`. I was thinking either
1) Make a wrapper struct/tuple thing for Policy arguments
2) Try to separate the arguments and filter the Policy arguments into a list, but this looks rather painful to do in c++11.

RReverser:
@brendandahl Can't think of a good way either.

My only other ideas were

- to ask user to provide a single `WithPolicies<...>` object itself, which might be easier to split out via SFINAE than a variadic list
- create an `AllowRawPointers<T>` type that is a zero-cost wrapper around `T*` (a struct with a single field `T*` which is ABI-compatible with `T*` itself in Wasm), so user can write things like `call<AllowRawPointers<Foo>>(...)` or `call<void>(AllowRawPointers(ptr), 42, ...)`

PR #24814, opened 2025-07-30
Use `var` and `undefined` in libwebaudio.js. NFC

sbc100:
The convention for JS library code is to use var for better minification.

The use of `void 0` instead of `undefined` I believe is a minor code size optimization, but its somewhat obscure and (as you can see from the fact that the code expectations didn't change) it an optimization that is already performed on the generated code anyway.

Split out from #24732


sbc100:
@cwoffenden

PR #24815, opened 2025-07-30
[test] Remove unnecessary bigint flags. NFC

sbc100:
Split out from #24808


kripken (commented):
Wait, this removes it, doesn't there need to be a corresponding place that adds it?

sbc100:
> Wait, this removes it, doesn't there need to be a corresponding place that adds it?

`WASM_BIGINT` is the default these days.    And `node_bigint_flags` are added as part of `setUp` as of #24796

PR #24821, opened 2025-07-30
Simplify em_js func types check. NFC

RReverser:
As far as I can tell, the only reason we had to re-parse the module here is because update_metadata didn't update em_js_func_types which might've been changed by legalization.

I've consolidates imports reading into read_module_imports so it now consistently reads any relevant info both during the initial module parsing and in the update_metadata.


tools/emscripten.py
```diff
@@ -317,26 +317,15 @@ def emscript(in_wasm, out_wasm, outfile_js, js_syms, finalize=True, base_metadat
...
         c_sig = c_sig.split(',')
-      if em_js_func in import_map:
-        imp = import_map[em_js_func]
-        assert imp.kind == webassembly.ExternType.FUNC
-        signature = types[imp.type]
-        if len(signature.params) != len(c_sig):
-          diagnostics.warning('em-js-i64', 'using 64-bit arguments in EM_JS function without WASM_BIGINT is not yet fully supported: `%s` (%s, %s)', em_js_func, c_sig, signature.params)
+      signature = metadata.em_js_func_types[em_js_func]
+      if len(signature.params) != len(c_sig):
+        diagnostics.warning('em-js-i64', 'using 64-bit arguments in EM_JS function without WASM_BIGINT is not yet fully supported: `%s` (%s, %s)', em_js_func, c_sig, signature.params)
```
sbc100, line 328: Is this change independent of the refactor in extract_metadata.py?   Or are they linked somehow?

RReverser: It's linked, as per PR description without the fixed up update_metadata we couldn't use em_js_func_types here as they would not be legalised yet.

sbc100: I think it was likely just an accident of history that it ended up this.  Nice simplification.

sbc100 (approved):

PR #24827, opened 2025-07-31
Add support for clang's color diagnostics flags

sbc100:
emcc will not honor the following clang flags:

```
-fcolor-diagnostics
-fdiagnostics-color
-fdiagnostics-color=always
-fno-color-diagnostics
-fdiagnostics-color=never
-fansi-escape-codes
```

### DIFF
--- a/tools/cmdline.py
+++ b/tools/cmdline.py
@@ -513,6 +513,12 @@ def parse_args(newargs):  # noqa: C901, PLR0912, PLR0915
       options.cpu_profiler = True
     elif check_flag('--threadprofiler'):
       settings_changes.append('PTHREADS_PROFILING=1')
+    elif arg in ('-fcolor-diagnostics', '-fdiagnostics-color', '-fdiagnostics-color=always'):
+      diagnostics.color_enabled = True
+    elif arg in ('-fno-color-diagnostics', '-fdiagnostics-color=never'):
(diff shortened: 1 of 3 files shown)


tools/cmdline.py
```diff
@@ -513,6 +513,12 @@ def consume_arg_file():
       options.cpu_profiler = True
     elif check_flag('--threadprofiler'):
       settings_changes.append('PTHREADS_PROFILING=1')
+    elif arg in ('-fcolor-diagnostics', '-fdiagnostics-color', '-fdiagnostics-color=always'):
+      diagnostics.color_enabled = True
+    elif arg in ('-fno-color-diagnostics', '-fdiagnostics-color=never'):
+      diagnostics.color_enabled = False
+    elif arg == '-fansi-escape-codes':
+      diagnostics.color_enabled = False
```
kripken: Should this not be `force_ansi = True`?

If not, where is that set to true?

kripken (commented):
"not" in the description should be "now"

tools/cmdline.py
kripken: lgtm with this suggestion applied, if it makes sense.

sbc100: Good catch!

kripken (approved):

Issue #24824, opened 2025-07-31
tot regression: can't invoke any compilation on Windows


RReverser:
Invoking e.g. `emcc hello_world.c` with tot version on Windows results in:

```
shared:INFO: (Emscripten: Running sanity checks)
emcc: Traceback (most recent call last):
  File "D:\emsdk\upstream\emscripten\emcc.py", line 748, in <module>
    sys.exit(main(sys.argv))
  File "D:\emsdk\python\3.9.2-nuget_64bit\lib\contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "D:\emsdk\upstream\emscripten\emcc.py", line 741, in main
    ret = run(args)
  File "D:\emsdk\upstream\emscripten\emcc.py", line 474, in run
    return link.run(options, linker_args)
  File "D:\emsdk\upstream\emscripten\tools\link.py", line 3066, in run
    target, wasm_target = phase_linker_setup(options, linker_args)
  File "D:\emsdk\python\3.9.2-nuget_64bit\lib\contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "D:\emsdk\upstream\emscripten\tools\link.py", line 1772, in phase_linker_setup
    diagnostics.warning('experimental', '-sASYNCIFY=2 (JSPI) is still experimental')
  File "D:\emsdk\upstream\emscripten\tools\diagnostics.py", line 262, in warning
    manager.warning(warning_type, message, *args)
  File "D:\emsdk\upstream\emscripten\tools\diagnostics.py", line 238, in warning
    warn(msg)
  File "D:\emsdk\upstream\emscripten\tools\diagnostics.py", line 168, in warn
    diag(WARN, msg, *args)
  File "D:\emsdk\upstream\emscripten\tools\diagnostics.py", line 140, in diag
    output = output_color(level_colors[level]) + bold()
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

As far as I can tell from history, this is likely a regression caused by #24804 cc @sbc100.

I'm guessing it's not caught by CI because CI doesn't test in TTY mode or something?

sbc100:
Oh wow.. i guess that is a major floor in our CI, it must not be testing windows with stdout set to real terminal.    Investigating.

sbc100:
It looks like this code only runs if `sys.stderr.isatty()` which I guess is not true on any of our CI systems?

PR #24828, opened 2025-07-31
Fix windows color terminal output

sbc100:
We cannot actually test this is working, but we can at least test
that it doesn't crash.

Fixes: #24824

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -12407,7 +12407,15 @@ int main(void) {
     output = self.expect_fail([EMCC, '-fansi-escape-codes', flag, 'src.c'])
     self.assertIn("\x1b[1msrc.c:1:13: \x1b[0m\x1b[0;1;31merror: \x1b[0m\x1b[1mexpected '}'\x1b[0m", output)
     # Verify that emcc errors show up as red and bold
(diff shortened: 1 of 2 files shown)


sbc100:
The fix is not actually part of this PR yet, just waiting to verify that the CI fails.

test/test_other.py
```diff
@@ -12407,7 +12407,15 @@ def test_color_diagnostics_force(self, flag):
     output = self.expect_fail([EMCC, '-fansi-escape-codes', flag, 'src.c'])
     self.assertIn("\x1b[1msrc.c:1:13: \x1b[0m\x1b[0;1;31merror: \x1b[0m\x1b[1mexpected '}'\x1b[0m", output)
     # Verify that emcc errors show up as red and bold
-    self.assertIn("emcc: \x1b[31m\x1b[1m", output)
+    self.assertIn('emcc: \x1b[31m\x1b[1m', output)
+
+    if WINDOWS:
+      # Also test without -fansi-escape-codes on windows.
+      # In those mode the code will use kennel calls such as SetConsoleTextAttribute to
```
```suggestion
      # In those mode the code will use kernel calls such as SetConsoleTextAttribute to
```

PR #24832, opened 2025-07-31
[embind] Add pointer policies for val's call, operator(), and new_.

brendandahl:
After #24831 24175, pointer policies were required in more places, but there
was no way to pass the policy into val's `call`, `operator()`, and `new_`
methods and use pointers.

All of these methods take variadic template arguments already
which made passing separate variadic policy arguments challenging. I used
some C++14/17 tricks to separate the regular arguments and policy
arguments.


brendandahl:
reviewer: The first commit is just moving policy code into wire.h, so it's probably easier to look at the second commit on its own.

brendandahl:
@sbc or @RReverser mind taking a look? The changes ignorning the moved code are relatively small [01eafbe](https://github.com/emscripten-core/emscripten/pull/24832/commits/01eafbed954c5045ee220cb16b9dd41a335412f8)

sbc100:
So does this mean some code now will not work of behave differently under C++11 vs C++17?  

I wonder if we should document this or mention in the changelog?

system/include/emscripten/val.h
```diff
@@ -573,8 +573,27 @@ class EMBIND_VISIBILITY_DEFAULT val {
...
 
-  template<internal::EM_INVOKER_KIND Kind, typename Policy, typename Ret, typename... Args>
+  template<internal::EM_INVOKER_KIND Kind, typename Ret, typename... Args>
   static Ret internalCall(EM_VAL handle, const char *methodName, Args&&... args) {
+    using namespace internal;
+#if __cplusplus >= 201703L
+    using Policy = WithPolicies<FilterTypes<isPolicy, Args...>>;
+    auto filteredArgs = Filter<isNotPolicy>(args...);
+    return std::apply(
+        [&](auto&&... filteredArgs) -> decltype(auto) {
```
sbc100: Should we name `filteredArgs` differently here so it doesn't shadow the outer `filteredArgs` 2 lines up?

brendandahl:
@sbc100 review ping

RReverser:
I'm sorry I'm kinda swamped at work at the moment, not much time for Emscripten reviews :(

brendandahl:
> What happens if you use `allow_raw_pointers` with older C++ version? Will it simply not compile? If so, then I think that is better than compiling but with different behaviour. Perhaps detail that in the changelog?

It will compile, but it will break at runtime. I can add an assert to disallow them for <c++17. A user could put the allow raw pointers in an `#ifdef` if they really need both to work. FWIW, this behavior has changed recently i.e. we use to allow raw pointers by default in val, but then we disallowed them (and broke people).

I'm hoping to switch embind to c++17 and newer in the near future, so this mess will go away.

sbc100:
> > What happens if you use `allow_raw_pointers` with older C++ version? Will it simply not compile? If so, then I think that is better than compiling but with different behaviour. Perhaps detail that in the changelog?
> 
> It will compile, but it will break at runtime. I can add an assert to disallow them for <c++17. A user could put the allow raw pointers in an `#ifdef` if they really need both to work. FWIW, this behavior has changed recently i.e. we use to allow raw pointers by default in val, but then we disallowed them (and broke people).
> 
> I'm hoping to switch embind to c++17 and newer in the near future, so this mess will go away.

I think an assert would be good, to avoid differing behaviors with different C++ versions.  

Also, the PR says C++14/17 but it sounds like its 17 that we need here?

sbc100:
Link to the C++17 update bug: https://github.com/emscripten-core/emscripten/issues/24850.  Hopefully that can land soon?  I didn't see any objections yet.

Issue #20189, opened 2023-09-05
 XMLHttpRequest.status is read only


MemomiAlexCherkashin:
https://github.com/emscripten-core/emscripten/blob/0566a76b500bd2bbd535e108f657fce1db7f6f75/src/Fetch.js#L408

xhr.status = 200 - cannot be overridden, as property is read only, based on mozila specification:
[https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest](url)

sbc100:
Is this causing a runtime failure for you, or maybe some kind of linter failure, or was this found just by looking at the code?

MemomiAlexCherkashin:
we see it when access some file from some our internal domains when use emscripten Fetcher API. Most of the cases - it avoids this section as server returns 200 itself.

philb90:
This is resulting in (very occasional) runtime exceptions in browser for me.

sbc100:
It seems like this could only happen if the server status was actually zero... can this happen in practice?   Under what circumstances?

sbc100:
Fix is in #24841, but I was unable to reproduce the issue myself.  If anyone has repro steps for this I would love to know how to trigger it.

philb90:
Thanks for the quick action on this.

> It seems like this could only happen if the server status was actually zero... can this happen in practice? Under what circumstances?

For us it was happening on a handful of requests per million; I didn't get as far as establishing the exact circumstances, my assumption is it's a failure case of some sort.

sbc100:
> Thanks for the quick action on this.
> 
> > It seems like this could only happen if the server status was actually zero... can this happen in practice? Under what circumstances?
> 
> For us it was happening on a handful of requests per million; I didn't get as far as establishing the exact circumstances, my assumption is it's a failure case of some sort.

Its very odd failure though since it seems like the response has data, but not status code .. If you are ever able to capture one of these errors in wild it would be great to get a full dump of the xhr object to see if we can figure out what is going on.

Do you happen to know if this effects just one browser?

PR #24841, opened 2025-08-01
[Fetch] Avoid assigned to XMLHttpRequest.status field

sbc100:
I have no idea how or when this assigned is triggered but apparently there are cases where it can be.  I tried doing fetches from `file://` but there were all blocked.

Fixes: #20189

### DIFF
--- a/src/Fetch.js
+++ b/src/Fetch.js
@@ -415,9 +415,10 @@ function fetchXHR(fetch, onsuccess, onerror, onprogress, onreadystatechange) {
     writeI53ToI64(fetch + {{{ C_STRUCTS.emscripten_fetch_t.dataOffset }}}, e.loaded - ptrLen);
     writeI53ToI64(fetch + {{{ C_STRUCTS.emscripten_fetch_t.totalBytes }}}, e.total);
     {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.readyState, 'xhr.readyState', 'i16') }}}
+    var status = xhr.status;
     // If loading files from a source that does not give HTTP status code, assume success if we get data bytes
-    if (xhr.readyState >= 3 && xhr.status === 0 && e.loaded > 0) xhr.status = 200;
-    {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.status, 'xhr.status', 'i16') }}}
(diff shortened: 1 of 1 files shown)


kripken:
>  I tried doing fetches from file:// but there were all blocked.

Browsers block file:// fetches by default, but I seem to recall a flag for it, and some searching suggested
```
chrome --allow-file-access-from-files
```
(But you should only use that for careful local testing!)

src/Fetch.js
```diff
@@ -415,9 +415,10 @@ function fetchXHR(fetch, onsuccess, onerror, onprogress, onreadystatechange) {
...
     writeI53ToI64(fetch + {{{ C_STRUCTS.emscripten_fetch_t.totalBytes }}}, e.total);
     {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.readyState, 'xhr.readyState', 'i16') }}}
+    var status = xhr.status;
     // If loading files from a source that does not give HTTP status code, assume success if we get data bytes
-    if (xhr.readyState >= 3 && xhr.status === 0 && e.loaded > 0) xhr.status = 200;
-    {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.status, 'xhr.status', 'i16') }}}
+    if (xhr.readyState >= 3 && xhr.status === 0 && e.loaded > 0) status = 200;
+    {{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.status, 'status', 'i16') }}}
     if (xhr.statusText) stringToUTF8(xhr.statusText, fetch + {{{ C_STRUCTS.emscripten_fetch_t.statusText }}}, 64);
     onprogress?.(fetch, xhr, e);
```
kripken, line 423: The onprogress handler will receive an xhr with a different status than the struct, which seems potentially dangerous, or at least confusing..?

Another option would be to copy the xhr object and modify that, but I'm not sure what is better?

sbc100: The onprogress callback is always set to `reportProgress` which doesn't ever use this argument.

kripken: Oh good, lgtm then.

sbc100:
I'll wait for #24845 land first so its more obvious this is safe.

PR #24849, opened 2025-08-01
Mark MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION as requiring html output

sbc100:
The code for handling this setting and performing the streaming exists only in the generated html.

Also error out if MINIMAL_RUNTIME_STREAMING_WASM_* are used without MINIMAL_RUNTIME.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- emcc will now error if `MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION` is used
+  when not generating html output.  This was always incompatible but previously
+  ignored. (#24849)
 
 4.0.12 - 08/01/25
 -----------------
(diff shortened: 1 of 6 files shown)


ChangeLog.md
```diff
@@ -20,6 +20,8 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- The `MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION` setting is only compatible
```
kripken: Do we need an entry here? This was always the case, and this PR only adds a clear error. This entry makes it seem like the PR removes functionality.

lgtm without this (unless I'm missing something)

sbc100: I updated to make it more clear

ChangeLog.md
```diff
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- emcc will now error if `MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION` is used
+  when not generating html output.  This was always incomabible but previously
```
```suggestion
  when not generating html output.  This was always incompatible but previously
```

kripken (commented):
Perhaps make this a warning and not an error? I don't feel strongly though.

sbc100:
> Perhaps make this a warning and not an error? I don't feel strongly though.

The problem is that if we do that then we have to consider that combination of options downstream in the rest of the code.

I guess we could invent a mechanism by which you could resolve incompatible options but disabled one of them with warning.. but we don't have that today.    I think for everyone's sanity its best to ban outright these incompatible options.

PR #24857, opened 2025-08-04
Only create compiler entry points for the current platform (by default)

sbc100:


tools/maint/create_entry_points.py
```diff
@@ -81,20 +85,22 @@ def generate_entry_points(cmd, path):
...
+        with open(os.path.join(__rootdir__, entry_point + '.ps1'), 'w') as f:
+          f.write(ps1_data)
 
   generate_entry_points(entry_points, os.path.join(__scriptdir__, 'run_python'))
   generate_entry_points(compiler_entry_points, os.path.join(__scriptdir__, 'run_python_compiler'))
 
 
 if __name__ == '__main__':
-  sys.exit(main())
+  sys.exit(main('--all' in sys.argv))
```
kripken, line 106: Where is this going to be used from?

kripken (approved):

tools/maint/create_entry_points.py
sbc100: Maybe nowhere?   I just figured some folks might want to generate all possible versions of the scripts.  As a developer its quite useful so you can see what would happen on a windows machine for example.

PR #24861, opened 2025-08-05
Cleanup some test. NFC

sbc100:
- Avoid assignment (prefer mutation) to self.cflags (see #24860 for why this is better).
- Parameterize test_wasm_intrinsics_simd and remove unneeded arguments.
- Use inline cflags where it improves test readability.
- Remove use of EMSCRIPTEN_KEEPALIVE with `main` function.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -2315,8 +2315,6 @@ Success!''')
 
   @no_ubsan('local count too large for VMs')
   def test_indirectbr(self):
-    self.cflags = [x for x in self.cflags if x != '-g']
-
     self.do_core_test('test_indirectbr.c')
 
   @no_asan('local count too large for VMs')
@@ -6549,9 +6547,8 @@ void* operator new(size_t size) {
   @also_with_asyncify_and_jspi
   def test_cubescript(self):
     # uses register keyword
-    self.cflags += ['-std=c++03', '-Wno-dynamic-class-memaccess']
+    self.cflags += ['-std=c++03', '-Wno-dynamic-class-memaccess', '-I', test_file('third_party/cubescript')]
     self.maybe_closure()
-    self.cflags += ['-I', test_file('third_party/cubescript')]
     # Test code contains memory leaks
     if '-fsanitize=address' in self.cflags:
       self.cflags += ['--pre-js', test_file('asan-no-leak.js')]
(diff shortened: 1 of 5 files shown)


sbc100:
@tlively for test_wasm_intrinsics_simd changes

test/test_core.py
```diff
@@ -6549,9 +6547,8 @@ def test_mmap_anon_pthreads(self):
   @also_with_asyncify_and_jspi
   def test_cubescript(self):
     # uses register keyword
-    self.cflags += ['-std=c++03', '-Wno-dynamic-class-memaccess']
+    self.cflags += ['-std=c++03', '-Wno-dynamic-class-memaccess', '-I', test_file('third_party/cubescript')]
```
kripken, line 6550: Why not remove `self.cflags +=` here as well, like below? Is there a difference in this test?

sbc100: You mean inline it into the `do_core_test` beflow?

I guess I don't have hard and fast rule for when to do that.  My rule of thumb tends to that it should inline if:

1. Its unconditional
2. Its short enough to fit in ~100 columns.

test/test_core.py
```diff
@@ -6564,23 +6561,18 @@ def test_relocatable_void_function(self):
...
   @wasm_simd
-  def test_wasm_intrinsics_simd(self):
-    def run():
-      self.do_runf('test_wasm_intrinsics_simd.c', 'Success!')
+  @parameterized({
+    '': ([],),
+    'unsigned_char': (['-funsigned-char'],),
+  })
+  def test_wasm_intrinsics_simd(self, args):
     # Improves test readability
```
tlively: This comment goes with the `-Wno-c++11-narrowing`, but it looks like that flag was mistakenly dropped.

test/test_core.py
```diff
@@ -6905,9 +6895,6 @@ def line_splitter(data):
 
       return out
 
-    # remove -g, so we have one test without it by default
-    self.cflags = [x for x in self.cflags if x != '-g']
```
tlively, line 6909: Are we worried about losing this coverage?

test/test_core.py
kripken: Sounds good. I was just curious if there was a functional reason I was missing.

kripken (approved):
lgtm % open comments

test/test_core.py
sbc100: But this test is not doesn't contain any such warnings.. and its not even a C++ test.

I.e. if I can remove the `-Wno-xxx` flag and no warnings show up then its not doing anything right?

test/test_core.py
sbc100: We have versions of the core test suite that include `-g` and tests that don't.

Notably I don't think any of the ones that we run by default include `-g` .. so this just seems pointless.

test/test_core.py
tlively: Oh interesting, I guess so. We should just remove the comment, too, then.

sbc100:
> LGTM with comment removed.

oops, i thought I had done that already.

Issue #24743, opened 2025-07-19
issue with `MINIMAL_RUNTIME`: wasmExports is undefined


kapodamy:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.10 (b7dc6e5747465580df5984e723b9d1f10d8e804b)
clang version 21.0.0git (https:/github.com/llvm/llvm-project 8f7e57485ee73205e108d74abb5565d5c63beaca)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /c/emsdk/upstream/bin

compile options used:
```
-sSTRICT
-sASYNCIFY
-sALLOW_MEMORY_GROWTH
-sWASM_BIGINT
-sFILESYSTEM=0
-sDYNAMIC_EXECUTION=0
-sNO_POLYFILL
-sENVIRONMENT=web
-sMINIMAL_RUNTIME=1
-sMINIMAL_RUNTIME_STREAMING_WASM_INSTANTIATION
-sEXPORT_KEEPALIVE=1

-sMODULARIZE=1 

-sEXPORT_NAME=Module
-sEXPORTED_FUNCTIONS=@./src/exports.txt
--pre-js ./src/pre.js
-sINCOMING_MODULE_JS_API=print,printErr
-sALLOW_TABLE_GROWTH 
-sEXPORTED_RUNTIME_METHODS=addFunction,removeFunction
````

the issue is on the glue js file:
```javascript
    WebAssembly.instantiateStreaming(fetch("lua.wasm"), imports).then(output => {
        var wasmExports = output.exports;
        assignWasmExports(wasmExports);
        wasmTable = wasmExports["__indirect_function_table"];
        wasmMemory = wasmExports["memory"];
        updateMemoryViews();
        initRuntime(wasmExports);
        ready()
    });
```
The property **exports** is not defined in **output** param. If i change to `var wasmExports = output.instance.exports;` it works

kapodamy:
looking here, line 161 should be `wasmExports = output.instance.exports; `
https://github.com/emscripten-core/emscripten/blob/b7dc6e5747465580df5984e723b9d1f10d8e804b/src/postamble_minimal.js#L147-L166

sbc100:
Wow, the logic there is pretty convoluted.  Clearly its wrong in your case.  I think we would want to make sure we don't break some other config.

[...]

PR #24863, opened 2025-08-05
Fix for MINIMAL_RUNTIME_STREAMING_WASM_INSTANTIATION + ENVIRONMENT=web

sbc100:
With `MINIMAL_RUNTIME_STREAMING_WASM_INSTANTIATION` we know that we always get a only an instance back.

The default case is unchanged: assume we get a pair of instance/module back.

With pthreads we have to support both cases.

Fixes: #24743

### DIFF
--- a/src/postamble_minimal.js
+++ b/src/postamble_minimal.js
@@ -136,7 +136,9 @@ assert(Module['wasm'], 'Must load WebAssembly Module in to variable Module.wasm 
 
 {{{ exportJSSymbols() }}}
 
-WebAssembly.instantiate(Module['wasm'], imports).then((output) => {
+// Add missingProperties supression here because closure compiler doesn't know that
+// WebAssembly.instantiate is polymorphic in its return value.
+WebAssembly.instantiate(Module['wasm'], imports).then(/** @suppress {missingProperties} */ (output) => {
 #endif
 
 #if !LibraryManager.has('libexports.js')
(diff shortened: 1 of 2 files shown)


kripken (approved):

magic-akari:
I've implemented the approach below to ensure the WebAssembly module loads correctly in both environments with and without `compileStreaming` support. However, I noticed that with this pull request change, `MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION` no longer enters the `wasmExports = (output.instance || output).exports;` branch, and the condition has been narrowed to only allow `PTHREADS`.

Since my use case doesn't require `PTHREADS`, I'm wondering if there might be an alternative way to restore the previous behavior?

```JavaScript
async function load(module) {
    if (typeof Response === "function" && module instanceof Response) {
        if ("compileStreaming" in WebAssembly) {
            try {
                return await WebAssembly.compileStreaming(module);
            } catch (e) {
                if (module.headers.get("Content-Type") != "application/wasm") {
                    console.warn(
                        "`WebAssembly.compileStreaming` failed because your server does not serve wasm with `application/wasm` MIME type. Falling back to `WebAssembly.instantiate` which is slower. Original error:\n",
                        e,
                    );
                } else {
                    throw e;
                }
            }
        }

        return module.arrayBuffer();
    }

    return module;
}

let wasm;
export default async function initAsync(input) {
    if (wasm !== undefined) {
        return wasm;
    }

    if (typeof input === "undefined") {
        input = new URL("lib.wasm", import.meta.url);
    }

    if (
        typeof input === "string" ||
        (typeof Request === "function" && input instanceof Request) ||
        (typeof URL === "function" && input instanceof URL)
    ) {
        input = fetch(input);
    }

    wasm = await load(await input).then((wasm) => Module({ wasm }));
}
```

PR #24865, opened 2025-08-05
Fix WASM_WORKERS + MINIMAL_RUNTIME to transfer compiled module

sbc100:
The current code was transferring `Module['wasm']` which is normally a byte array.  The new code uses the same pattern that pthreads uses to stash the compiled module and postMessage that instead, which should be more performant.


src/postamble_minimal.js
```diff
@@ -150,11 +148,11 @@ WebAssembly.instantiate(Module['wasm'], imports).then((output) => {
...
   // Depending on the build mode, Module['wasm'] can mean a different thing.
-#if MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION || MINIMAL_RUNTIME_STREAMING_WASM_INSTANTIATION || PTHREADS
+#if MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION || MINIMAL_RUNTIME_STREAMING_WASM_INSTANTIATION || PTHREADS || WASM_WORKERS
   // https://caniuse.com/#feat=wasm and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiateStreaming
-#if MIN_FIREFOX_VERSION < 58 || MIN_CHROME_VERSION < 61 || MIN_SAFARI_VERSION < 150000 || ENVIRONMENT_MAY_BE_NODE || PTHREADS
-  // In pthreads, Module['wasm'] is an already compiled WebAssembly.Module. In
-  // that case, 'output' is a WebAssembly.Instance.
+#if MIN_FIREFOX_VERSION < 58 || MIN_CHROME_VERSION < 61 || MIN_SAFARI_VERSION < 150000 || ENVIRONMENT_MAY_BE_NODE || PTHREADS || WASM_WORKERS
+  // In pthreads and wasm workers, Module['wasm'] is an already compiled
+  // WebAssembly.Module. In that case, 'output' is a WebAssembly.Instance.
```
kripken, line 155: The PR description says `Module['wasm']` is a byte array. Was this comment wrong then, in the old code?

sbc100: When the worker or pthread starts up it assigns to recieved wasm module to `Module['wasm']` to the comment is correct.

However on the main thread `Module['wasm']` is a byte array, so we don't want to send that over.  Instead we send over `wasmModule` which is the compiled version.

This change basically makes wasm works behave like pthreads already do in this regard.

sbc100: I updated the PR description to try to give some more context.

test/code_size/test_minimal_runtime_code_size_hello_wasm_worker.json
```diff
@@ -1,10 +1,10 @@
 {
   "a.html": 519,
   "a.html.gz": 357,
-  "a.js": 830,
-  "a.js.gz": 525,
+  "a.js": 855,
```
kripken, line 4: Do you know what causes this 3% regression?

sbc100: This just a few bytes due to:

1. A new variable to track the compiler wasm module
2. The instantation handler being able to handle both types of return value `a.instance.exports` vs `(a.instance || a).exports;`

I think its worth it to get compiled module sharing between workers, and to be consistent with the pthread behaviour.

kripken: Could we remove `Module['wasm']` after this as a later simplification? That might get back this size, and more.

sbc100: `Module['wasm']` in `MINIMAL_RUNTIME` is how the wasm bytes/module is supplied.  Its up the surrounding HTML to set this ahead of time so we can't get rid of it.

kripken: I see.

Then I wonder if it isn't worth a little more ifdefing, in return for 3% better code size? That is the goal of minimal runtime mode, after all.

I'm not opposed to such tradeoffs in general, but the benefit here seems small compared to the cost?

sbc100: Its already ifdef'd such this change only effects `-sWASM_WORKERS`.

The reason its regression is that the old code did not share the Wasm module between workers. 

The regression is necessary because of the way `WebAssembly.instantiate` returns different result for re-compiled modules.  There is no way to have the same instantiation code on workers and on the main thread without this regession (not as far as I know anyway).

Another way of putting it:  The reason this worked before is that wasm workers were all sharing the same instantiation path as the main thread.  i.e. they were not sharing the compiled module.  I believe this was a bug.  Once I fixed that bug, the regression in the instantiation code becomes necessary to handle to two different result types (but only for `-sWASM_WORKERS`)

kripken: Oh, I see, yes, that does sound like a bug, and definitely worth fixing.

PR #24866, opened 2025-08-06
[AUDIO_WORKLET] fix emscripten_create_audio_context for wasm64

cwoffenden:
I spotted there was no coverage for `emscripten_create_audio_context()` and then that it fails with wasm64 (it accesses the heap directly). This is a minor fix with an addition to the test.

Note: the code is slightly smaller than the original but still performs the same (omitting `latencyHint` or `sampleRate` results in `undefined`).

### DIFF
--- a/src/lib/libwebaudio.js
+++ b/src/lib/libwebaudio.js
@@ -58,11 +58,10 @@ var LibraryWebAudio = {
 #if ASSERTIONS
     if (!ctx) console.error('emscripten_create_audio_context failed! Web Audio is not supported.');
 #endif
-    options >>= 2;
 
     var opts = options ? {
-      latencyHint: HEAPU32[options] ? UTF8ToString(HEAPU32[options]) : undefined,
-      sampleRate: HEAP32[options+1] || undefined
+      latencyHint: UTF8ToString({{{ makeGetValue('options', C_STRUCTS.EmscriptenWebAudioCreateAttributes.latencyHint, '*') }}}) || undefined,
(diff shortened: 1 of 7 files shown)


src/lib/libwebaudio.js
```diff
@@ -58,11 +58,10 @@ var LibraryWebAudio = {
 #if ASSERTIONS
     if (!ctx) console.error('emscripten_create_audio_context failed! Web Audio is not supported.');
 #endif
-    options >>= 2;
 
     var opts = options ? {
-      latencyHint: HEAPU32[options] ? UTF8ToString(HEAPU32[options]) : undefined,
-      sampleRate: HEAP32[options+1] || undefined
+      latencyHint: UTF8ToString({{{ makeGetValue('options', C_STRUCTS.EmscriptenWebAudioCreateAttributes.latencyHint, '*') }}}) || undefined,
```
sbc100, line 63: Is the logic here the same?   Won't UTF8ToString return an empty string if the ptr is null?

sbc100 (approved):
lgtm % comments

src/lib/libwebaudio.js
cwoffenden: `UTF8ToString()` returns the empty string and JS sees `''` as false (so takes the `|| undefined`). Try this in the console:
```
> '' || undefined
undefined
> 'blah' || undefined
'blah'
```
(It's unnatural and should be banned, for sure!)

sbc100: OK great, lgtm then!

PR #24871, opened 2025-08-06
[file_packager.py] Fix `errback` call in `fetchRemotePackage`

lkwinta:
`errback` argument was ignored in `fetchRemotePackage`, `Promise.reject` was called instead

### DIFF
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -972,10 +972,10 @@ def generate_js(data_target, data_files, metadata):
         %(node_support_code)s
         Module['dataFileDownloads'] ??= {};
         fetch(packageName)
-          .catch((cause) => Promise.reject(new Error(`Network Error: ${packageName}`, {cause}))) // If fetch fails, rewrite the error to include the failing URL & the cause.
+          .catch((cause) => errback(new Error(`Network Error: ${packageName}`, {cause}))) // If fetch fails, rewrite the error to include the failing URL & the cause.
           .then((response) => {
             if (!response.ok) {
-              return Promise.reject(new Error(`${response.status}: ${response.url}`));
+              return errback(new Error(`${response.status}: ${response.url}`));
             }
 
             if (!response.body && response.arrayBuffer) { // If we're using the polyfill, readers won't be available...
@@ -984,7 +984,7 @@ def generate_js(data_target, data_files, metadata):
 
             const reader = response.body.getReader();
             const iterate = () => reader.read().then(handleChunk).catch((cause) => {
-              return Promise.reject(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}));
+              return errback(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}));
             });
 
             const chunks = [];
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -636,10 +636,8 @@ def generate_js(data_target, data_files, metadata):
   if options.support_node:
     ret += "    var isNode = typeof process === 'object' && typeof process.versions === 'object' && typeof process.versions.node === 'string';\n"
   ret += '''
-    function handleError(error) {
-      console.error('package error:', error);
-
-      return Promise.reject(error);
+    function throwPackageError(error) {
+      throw error;
     };
 
     function loadPackage(metadata) {\n'''
@@ -1050,7 +1048,7 @@ def generate_js(data_target, data_files, metadata):
         function preloadFallback(error) {
           console.error(error);
           console.error('falling back to default preload behavior');
-          fetchRemotePackage(REMOTE_PACKAGE_NAME, REMOTE_PACKAGE_SIZE, processPackageData, handleError);
+          fetchRemotePackage(REMOTE_PACKAGE_NAME, REMOTE_PACKAGE_SIZE, processPackageData, throwPackageError);
         };
 
         openDatabase(
@@ -1090,7 +1088,7 @@ def generate_js(data_target, data_files, metadata):
         } else {
           fetched = data;
         }
-      }, handleError);\n'''
+      }, throwPackageError);\n'''
 
       code += '''
       Module['preloadResults'][PACKAGE_NAME] = {fromCache: false};
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -635,12 +635,12 @@ def generate_js(data_target, data_files, metadata):
 
   if options.support_node:
     ret += "    var isNode = typeof process === 'object' && typeof process.versions === 'object' && typeof process.versions.node === 'string';\n"
-  ret += ''' 
+  ret += '''
     function handleError(error) {
       console.error('package error:', error);
       return error;
     };
-      
+
     function loadPackage(metadata) {\n'''
 
   code = '''
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -641,7 +641,7 @@ def generate_js(data_target, data_files, metadata):
       return error;
     };
       
-    function loadPackage(metadata) {\n''''
+    function loadPackage(metadata) {\n'''
 
   code = '''
       function assert(check, msg) {
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -638,7 +638,6 @@ def generate_js(data_target, data_files, metadata):
   ret += '''
     function handleError(error) {
       console.error('package error:', error);
-      return error;
     };
 
     function loadPackage(metadata) {\n'''
@@ -1119,7 +1118,7 @@ def generate_js(data_target, data_files, metadata):
         if (isNode) {
           require('fs').readFile(metadataUrl, 'utf8', (err, contents) => {
             if (err) {
-              return handleError(err);
+              handleError(err);
             } else {
               loadPackage(JSON.parse(contents));
             }
@@ -1140,8 +1139,8 @@ def generate_js(data_target, data_files, metadata):
         if (response.ok) {
           return response.json();
         }
-        return handleError(new Error(`${response.status}: ${response.url}`));
       })
+      .catch((cause) => handleError(new Error(`Network Error: ${packageName}`, {cause}))) // If fetch fails, rewrite the error to include the failing URL & the cause.
       .then(loadPackage);
   }
 
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -988,9 +988,9 @@ def generate_js(data_target, data_files, metadata):
             }
 
             const reader = response.body.getReader();
-            const iterate = () => reader.read().then(handleChunk).catch((cause) => {
-              return errback(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}));
-            });
+            const iterate = () => reader.read().then(handleChunk).catch((cause) =>
+              errback(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}))
+            );
 
             const chunks = [];
             const headers = response.headers;
@@ -1135,14 +1135,15 @@ def generate_js(data_target, data_files, metadata):
     var metadataUrl = Module['locateFile'] ? Module['locateFile']('%(metadata_file)s', '') : '%(metadata_file)s';
     %(node_support_code)s
     fetch(metadataUrl)
+      .catch((cause) =>
+        throwPackageError(new Error(`Network Error: ${packageName}`, {cause}))
+      ) // If fetch fails, rewrite the error to include the failing URL & the cause.
       .then((response) => {
         if (response.ok) {
           return response.json();
         }
+        throwPackageError(new Error(`Response error: ${response}`));
       })
-      .catch((cause) =>
-        throwPackageError(new Error(`Network Error: ${packageName}`, {cause}))
-      ) // If fetch fails, rewrite the error to include the failing URL & the cause.
       .then(loadPackage);
   }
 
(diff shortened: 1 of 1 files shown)


lkwinta:
@sbc100

sbc100 (approved):

sbc100:
By the way it looks like this code was added in #22016, but the code in replaced also did `throw new Error` rather than calling `errback`.

Also, it looks like errback is just doing  `console.error('package error:', error);` .. not sure thats any better. 

This change does seem perfectly good, but I'm wondoring how it solves your issue if the result is just to call `console.error`?

lkwinta:
> By the way it looks like this code was added in #22016, but the code in replaced also did `throw new Error` rather than calling `errback`.
> 
> Also, it looks like errback is just doing `console.error('package error:', error);` .. not sure thats any better.
> 
> This change does seem perfectly good, but I'm wondoring how it solves your issue if the result is just to call `console.error`?

I will now have one place where I can reject the promise returned by function.

Other thing that comes to my mind is changing console.errors in whole script to Module.printErr to enable filtering or guard the logs with some flag

lkwinta:
@sbc100 I think that I have more proposal as part of this mr

tools/file_packager.py
```diff
@@ -635,7 +635,13 @@ def generate_js(data_target, data_files, metadata):
 
   if options.support_node:
     ret += "    var isNode = typeof process === 'object' && typeof process.versions === 'object' && typeof process.versions.node === 'string';\n"
-  ret += '    function loadPackage(metadata) {\n'
+  ret += ''' 
+    function handleError(error) {
+      console.error('package error:', error);
+      return error;
```
sbc100: Why add `return error` here?

tools/file_packager.py
```diff
@@ -1138,7 +1140,7 @@ def generate_js(data_target, data_files, metadata):
         if (response.ok) {
           return response.json();
         }
-        return Promise.reject(new Error(`${response.status}: ${response.url}`));
+        return handleError(new Error(`${response.status}: ${response.url}`));
```
sbc100: I think the `return Promise.reject` is needed here to avoid the `.then` firing below (i.e. to break the promise chain).

Maybe single `.catch(handleError)` on line 1146 is what is needed here to make sure we call `handleError`

tools/file_packager.py
```diff
@@ -1117,7 +1119,7 @@ def generate_js(data_target, data_files, metadata):
         if (isNode) {
           require('fs').readFile(metadataUrl, 'utf8', (err, contents) => {
             if (err) {
-              return Promise.reject(err);
+              return handleError(err);
```
sbc100: I think you can delete the `return` keyword here.

tools/file_packager.py
sbc100: If you google "how to reject a promise from a then handler" you will see that we either need to throw something here or return a rejected promise object.

lkwinta: Right

tools/file_packager.py
lkwinta: I think think about it now again it is really unnecessary

lkwinta:
Should I merge when the pipeline passes now?

tools/file_packager.py
```diff
@@ -1138,8 +1139,8 @@ def generate_js(data_target, data_files, metadata):
         if (response.ok) {
           return response.json();
         }
-        return Promise.reject(new Error(`${response.status}: ${response.url}`));
```
sbc100, line 1141: I think this line is still needed to reject the promise and then the `.catch` on line 1143 should just be `catch(handleError)`.

Without this line the promise chain continue with `undefined` as the value.

lkwinta: Now I get it, based on how it was implemented in `fetchRemotePackage` I thought that catch cancels the promise

lkwinta: Maybe the HandleError should return Promise.reject, because we return it often, even in catch ?

lkwinta: Or just throw

sbc100: Yes, I think maybe it should just throwing is best.  In which case you can rename it to `throwPackageError`... and in that case maybe we don't need the console.error there.

lkwinta: Done

sbc100 (approved):

lkwinta:
@sbc100 I have fixed the pipeline, some returns where neccessary since errback had different meaning, I wonder if we should modify catch statement in fetchRemotePackage now

tools/file_packager.py
```diff
@@ -984,7 +989,7 @@ def generate_js(data_target, data_files, metadata):
 
             const reader = response.body.getReader();
             const iterate = () => reader.read().then(handleChunk).catch((cause) => {
-              return Promise.reject(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}));
+              return errback(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}));
```
sbc100: I don't think this `return` is needed here.

tools/file_packager.py
```diff
@@ -1138,8 +1139,10 @@ def generate_js(data_target, data_files, metadata):
         if (response.ok) {
           return response.json();
         }
```
sbc100, line 1149: I think you still need to do something here when the response is not `ok`... you either need the `throw` or return a rejected promise object I think.

sbc100: Actaully I think just doing `throwPackageError` here on this line is probably what you want.

lkwinta: I mean we call fetchRemotePackage with two different error handlers, once we just wants to throw na error, but in one scenario, with cache enabled we want to call fallback handler and not throw error

tools/file_packager.py
lkwinta: I agree,

tools/file_packager.py
lkwinta: Done

lkwinta: I wonder now, if we take this code: 

```js
function preloadFallback(error) {
          console.error(error);
          console.error('falling back to default preload behavior');
          fetchRemotePackage(REMOTE_PACKAGE_NAME, REMOTE_PACKAGE_SIZE, processPackageData, throwPackageError);
        };

        openDatabase(
          (db) => checkCachedPackage(db, PACKAGE_PATH + PACKAGE_NAME,
              (useCached, metadata) => {
                Module['preloadResults'][PACKAGE_NAME] = {fromCache: useCached};
                if (useCached) {
                  fetchCachedPackage(db, PACKAGE_PATH + PACKAGE_NAME, metadata, processPackageData, preloadFallback);
                } else {
                  fetchRemotePackage(REMOTE_PACKAGE_NAME, REMOTE_PACKAGE_SIZE,
                    (packageData) => {
                      cacheRemotePackage(db, PACKAGE_PATH + PACKAGE_NAME, packageData, {uuid:PACKAGE_UUID}, processPackageData,
                        (error) => {
                          console.error(error);
                          processPackageData(packageData);
                        });
                    }
                  , preloadFallback);
                }
              }, preloadFallback)
        , preloadFallback);
```

then this catch in fetchRemotePackage will not fail 

```js
        fetch(packageName)
          .catch((cause) => errback(new Error(`Network Error: ${packageName}`, {cause}))) // If fetch fails, rewrite the error to include the failing URL & the cause.
          .then((response) => {
```

I think that just adding promise Rejsect here for security iss what we want

tools/file_packager.py
```diff
@@ -972,20 +977,25 @@ def generate_js(data_target, data_files, metadata):
         %(node_support_code)s
         Module['dataFileDownloads'] ??= {};
         fetch(packageName)
-          .catch((cause) => Promise.reject(new Error(`Network Error: ${packageName}`, {cause}))) // If fetch fails, rewrite the error to include the failing URL & the cause.
+          .catch((cause) => {
+            errback(new Error(`Network Error: ${packageName}`, {cause}));
+            // We want to be sure to cancel promise chain because we sometimes call this function with method
+            // which doesn't throw error hance promise chain is not broken
+            return Promise.reject(new Error(`Network Error: ${packageName}`, {cause}));
```
sbc100, line 984: This is starting to look pretty ugly because we mixing callbacks and promises .. I think converting to using `async` `await` is ultimately what we we probably want to do here.

lkwinta: I have prototyped something, but maybe it deservse separate PR...

```js
function fetchRemotePackage(packageName, packageSize) {
        return new Promise((fetchPackageResolve, fetchPackageReject) => {
          if (isNode) {
          require('fs').readFile(packageName, (err, contents) => {
            if (err) {
              fetchPackageReject(err);
            } else {
              fetchPackageResolve(contents.buffer);
            }
          });
          return;
        }
          fetch(packageName)
            .catch((cause) => {
              err = new Error(`Network Error: ${packageName}`, {cause});
              fetchPackageReject(err);
              return Promise.reject(err);
            }) // If fetch fails, rewrite the error to include the failing URL & the cause.
            .then((response) => {
              if (!response.ok) {
                fetchPackageReject(new Error(`${response.status}: ${response.url}`));
                return;
              }

              if (!response.body && response.arrayBuffer) { // If we're using the polyfill, readers won't be available...
                return response.arrayBuffer().then(fetchPackageResolve);
              }

              const reader = response.body.getReader();
              const iterate = () => reader.read().then(handleChunk).catch((cause) => {
                fetchPackageReject(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}));
                return;
              });

              const chunks = [];
              const headers = response.headers;
              const total = Number(headers.get('Content-Length') ?? packageSize);
              let loaded = 0;

              const handleChunk = ({done, value}) => {
                if (!done) {
                  chunks.push(value);
                  loaded += value.length;
                  Module['dataFileDownloads'][packageName] = {loaded, total};

                  let totalLoaded = 0;
                  let totalSize = 0;

                  for (const download of Object.values(Module['dataFileDownloads'])) {
                    totalLoaded += download.loaded;
                    totalSize += download.total;
                  }

                  Module['setStatus']?.(`Downloading data... (${totalLoaded}/${totalSize})`);
                  return iterate();
                } else {
                  const packageData = new Uint8Array(chunks.map((c) => c.length).reduce((a, b) => a + b, 0));
                  let offset = 0;
                  for (const chunk of chunks) {
                    packageData.set(chunk, offset);
                    offset += chunk.length;
                  }
                  fetchPackageResolve(packageData.buffer);
                }
              };

              Module['setStatus']?.('Downloading data...');
              return iterate();
            });
        });
      };
      ```

sbc100:
I took as stab as converting part of the file_packager output to async/await: https://github.com/emscripten-core/emscripten/pull/24871.    Its not the relevant part .. but its a start.

lkwinta:
> I took as stab as converting part of the file_packager output to async/await: #24871. Its not the relevant part .. but its a start.

So should I fix that now ?

sbc100:
> > I took as stab as converting part of the file_packager output to async/await: #24871. Its not the relevant part .. but its a start.
> 
> So should I fix that now ?

I've made some progress on it, so can do this part if you like?

sbc100:
> > > I took as stab as converting part of the file_packager output to async/await: #24871. Its not the relevant part .. but its a start.
> > 
> > 
> > So should I fix that now ?
> 
> I've made some progress on it, so can do this part if you like?

Ooops that should have been: #24882

lkwinta:
> > > I took as stab as converting part of the file_packager output to async/await: #24871. Its not the relevant part .. but its a start.
> > 
> > 
> > So should I fix that now ?
> 
> I've made some progress on it, so can do this part if you like?

Not right now for sure, but feel free to take and use my prototype, I have changed it now after some tests, at least happy path should pass with current version pasted in my comment.

sbc100:
I think I'm going to go for full conversion to async/await which should make your change much simpler (I hope)

sbc100:
Part 2: https://github.com/emscripten-core/emscripten/pull/24883

lkwinta:
> I think I'm going to go for full conversion to async/await which should make your change much simpler (I hope)

So we wait with this PR and the export_es6 one for the conversion to async?

sbc100:
Yes I think the async conversion work will supersede this PR

lkwinta:
> Yes I think the async conversion work will supersede this PR

@sbc100 to be honest I don't see the point of this PR anymore after your changes. I think it will be possible to Reject main functions promise from just one or two places without the need of handleError

sbc100:
> > Yes I think the async conversion work will supersede this PR
> 
> @sbc100 to be honest I don't see the point of this PR anymore after your changes. I think it will be possible to Reject main functions promise from just one or two places without the need of handleError

That sounds reasonable to me.   

The non-es6 case I think the point of handleError is to show something on the console to the user so they know the package loading has failed.

For the es6 case i don't think that is necessary, so I think you could just do `handleError = reject` maybe?

lkwinta:
> > > Yes I think the async conversion work will supersede this PR
> > 
> > 
> > @sbc100 to be honest I don't see the point of this PR anymore after your changes. I think it will be possible to Reject main functions promise from just one or two places without the need of handleError
> 
> That sounds reasonable to me.
> 
> The non-es6 case I think the point of handleError is to show something on the console to the user so they know the package loading has failed.
> 
> For the es6 case i don't think that is necessary, so I think you could just do `handleError = reject` maybe?

So I think I can close that

sbc100:
Yes, I think this PR can be closed.  Is that what you are asking?  I think the conversion async removed the need for this fix.

PR #24874, opened 2025-08-06
Remove internal USE_READY_PROMISE

sbc100:
This was only used in one very specific case which required all the following to be enabled:

- MINIMAL_RUNTIME
- MODULARIZE
- SINGLE_FILE
- !PTHREADS
- html output

This combination rare because most folks don't have use the emscripten generated html output.  So that rules out most of our users, then they would also have to choose `MODULARIZE` but not actually want to create multiple instances or pass in any arguments.  This is because in order to support this very special use case the generated HTML was setup to stamp out a singleton instance of the module with no arguments passed in.

While it *is* sometimes useful to stamp out a single instance of the module we don't do this in any other circumstances.  If a user wants to do this its easy enough to do with a `--post-js` file.  In fact we do this in out test code when we want to test the `-sMODULARIZE` settings.  See `test/modularize_post_js.js`.

So think this change is really an improvment and I think its very unlikely that many folks will be using this specific combination of settings.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,14 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- The `-sMODULARIZE` setting generates a factory function that must be called
+  before the program is instantiated that run.  However, emscripten previously
+  had very special case where this instantiation would happen automatically
+  (with no parameterization) under certain specific circumstances: When
+  `-sMINIMAL_RUNTIME`, `-sSINGLE_FILE` and `-sMODULARIZE` were used in
+  combination with default html output.  This special case was removed.  If you
+  want to instantiate the module on startup you can still do so by adding a call
+  the factory function in `--extern-post-js`. (#24874)
 - emcc will now error if `MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION` is used
   when not generating html output.  This was always incompatible but previously
   ignored. (#24849)
--- a/src/postamble_modularize.js
+++ b/src/postamble_modularize.js
@@ -4,7 +4,6 @@
 // We assign to the `moduleRtn` global here and configure closure to see
 // this as and extern so it won't get minified.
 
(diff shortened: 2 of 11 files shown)


ChangeLog.md
```diff
@@ -20,6 +20,14 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- The `-sMODULARIZE` setting generates a factory function that must be called
+  before the program is instantiated that run.  However, emscripten previously
+  had very special case where this instantiation would happen automatically
+  (with no parameterization) under certain specific circumstances: When
+  `-sMINIMAL_RUNTIME`, `-sSINGLE_FILE` and `-sMODULARIZE` were used in
+  combination with defeault html otuput.  This special case was removed.  If you
```
```suggestion
  combination with default html otuput.  This special case was removed.  If you
```

kripken (approved):
I agree this simplification makes sense.

I also think the optimization to create a singleton instance in an efficient way is nice, but if we want to do it, we should have it work consistently, e.g. as MODULARIZE_INSTANCE used to.

sbc100:
> I agree this simplification makes sense.
> 
> I also think the optimization to create a singleton instance in an efficient way is nice, but if we want to do it, we should have it work consistently, e.g. as MODULARIZE_INSTANCE used to.

Indeed and we do have the new `-sMODULARIZE=instance` (which is currently limited to `-sEXPORT_ES6` but we could extend it).

kripken:
Oh, I forgot we have `-sMODULARIZE=instance`. It might be nice to use that for code size, then, to offset the size regression here.

sbc100:
> Oh, I forgot we have `-sMODULARIZE=instance`. It might be nice to use that for code size, then, to offset the size regression here.

But we still want to record codesize number of the actual `-sMODULARIZE` setting (even if we also do it for `-sMODULARIZE=instance`.

Also, `-sMODULARIZE=instance` is currently limited to `-sEXPORT_ES6` and marked as experimental.

PR #24875, opened 2025-08-06
Remove duplicate `also_with_wasmfs` test decorator. NFC

sbc100:


test/test_browser.py
```diff
@@ -77,26 +77,6 @@ def do_GET(s):
...
-      self.set_setting('WASMFS')
-      self.cflags = self.cflags.copy() + ['-DWASMFS']
-      f(self, *args, **kwargs)
-    else:
-      f(self, *args, **kwargs)
-
-  parameterize(metafunc, {'': (False,),
-                          'wasmfs': (True,)})
-
-  return metafunc
```
kripken, line 97: Comparing to the version the code will now use, the new version has this:
```python
    if wasmfs:
      ..
    else:
      self.cflags += ['-DMEMFS']
```
I suppose that define is harmless, as I see no uses of it in C/CPP files. Perhaps we can even remove it.

kripken (approved):

PR #24877, opened 2025-08-07
Fix function assert in test_legalize_js_ffi.

brendandahl:
After https://github.com/WebAssembly/binaryen/pull/7802 the output of the test changed from

`(func $add_f (param $0 f32) (param $1 f32) (result f32)` to
`(func $add_f (type $5) (param $0 f32) (param $1 f32) (result f32)`


brendandahl:
Whoops didn't see the fix was already open.

PR #24878, opened 2025-08-07
[AUDIO_WORKLET] Tidy and document the worklet processing code. NFC

cwoffenden:
This renames some of the single-letter variables and documents their use.

Since I've spent a lot of time in this code it's extra work to keep remembering that `k` is the struct address (previously the heap index), `j` is the audio channel, etc. I also removed some of the code that had side effects, e.g. `c = (a += b)`; some but not all).

The end result (IMO) is readable and even saves a few bytes after Closure.

(This is the only planned change like this ahead of the optimisation PR)

### DIFF
--- a/src/audio_worklet.js
+++ b/src/audio_worklet.js
@@ -49,20 +49,14 @@ function createWasmAudioWorkletProcessor(audioParams) {
 
       var numInputs = inputList.length;
       var numOutputs = outputList.length;
-      var numParams = 0;
 
       var entry; // reused list entry or index
       var subentry; // reused channel or other array in each list entry or index
-      var structPtr; // working start to each struct record
-      var dataPtr; // start of the data section, usually after all structs
-
-      var bytesPerChannel = this.samplesPerChannel * {{{ getNativeTypeSize('float') }}};
-      var stackMemoryNeeded = (numInputs + numOutputs) * {{{ C_STRUCTS.AudioSampleFrame.__size__ }}};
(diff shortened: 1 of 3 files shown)


src/audio_worklet.js
```diff
@@ -47,79 +47,89 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+        ++numParams;
+      }
 
       // Allocate the necessary stack space.
       inputsPtr = stackAlloc(stackMemoryNeeded);
 
       // Copy input audio descriptor structs and data to Wasm
-      k = inputsPtr;
+      structPtr = inputsPtr;
       dataPtr = inputsPtr + numInputs * {{{ C_STRUCTS.AudioSampleFrame.__size__ }}};
```
sbc100: Maybe just declare there down here.. no need to declare them above, right?

sbc100: I believe in the old code the `var`s were all group together to avoid repeating the `var` keyword.   Now that we can run closure on this code we should just declare each `var` at the place where its first needed.

cwoffenden: I did originally declare them where used, but each added to the file size (a few bytes), so I grouped them manually. Closure didn't appear to be grouping them as expected. Let me take another pass (probably tomorrow).

cwoffenden: Quick check, if I move them around to be near their use site it's a delta of `6`. I think it's worth it, your call.

sbc100: That makes no sense to me since I though closure took care of that.. oh well.  lgtm.

PR #24883, opened 2025-08-07
[file_packager] Convert openDatabase to async/await. NFC

sbc100:
Followup to #24882


sbc100:
Best viewed with "hide whitespace".

sbc100:
I hope you don't mind me splitting this up into small peices like this.  Its easier for me to get my head around each step on its own and I hope its easier to review.

tools/file_packager.py
```diff
@@ -805,40 +805,32 @@ def generate_js(data_target, data_files, metadata):
         var DB_VERSION = 1;
         var METADATA_STORE_NAME = 'METADATA';
         var PACKAGE_STORE_NAME = 'PACKAGES';
-        function openDatabase(callback, errback) {'''
-      if options.support_node:
-        code += '''
-            if (isNode) {
-              return errback();
-            }'''
```
kripken: Looks like node always errored. Is this fixing that path?

kripken: (do we test it?)

sbc100: I think the `typeof indexedDB == 'undefined'` just below should be enough to handle the node case.

sbc100: Hmm.. maybe I should split that out, or at least see why it was added.

tools/file_packager.py
```diff
@@ -805,34 +805,32 @@ def generate_js(data_target, data_files, metadata):
         var DB_VERSION = 1;
         var METADATA_STORE_NAME = 'METADATA';
         var PACKAGE_STORE_NAME = 'PACKAGES';
-        function openDatabase(callback, errback) {
+        async function openDatabase() {
           if (typeof indexedDB == 'undefined') {
             return errback('using IndexedDB to cache data can only be done on a web page or in a web worker');
```
lkwinta, line 810: @sbc100 @kripken I think that this introduces a bug, it should be be throw or promise reject since errback doesn't exist anymore

sbc100: Good point!  Will fix

sbc100: Looks like this went unnoticed because the `errback is not defined` error was caught an the fallback was run anyway:

```
$ node ./src.js
ReferenceError: errback is not defined
    at openDatabase (/usr/local/google/home/sbc/dev/wasm/emscripten/out/test/src.js:173:13)
    at runWithFS (/usr/local/google/home/sbc/dev/wasm/emscripten/out/test/src.js:333:9)
    at callRuntimeCallbacks (/usr/local/google/home/sbc/dev/wasm/emscripten/out/test/src.js:1164:26)
    at preRun (/usr/local/google/home/sbc/dev/wasm/emscripten/out/test/src.js:841:3)
    at run (/usr/local/google/home/sbc/dev/wasm/emscripten/out/test/src.js:4658:3)
    at Object.removeRunDependency (/usr/local/google/home/sbc/dev/wasm/emscripten/out/test/src.js:950:7)
    at loadPackage (/usr/local/google/home/sbc/dev/wasm/emscripten/out/test/src.js:362:34)
    at runMetaWithFS (/usr/local/google/home/sbc/dev/wasm/emscripten/out/test/src.js:371:18)
falling back to default preload behavior
done
```

Still worth fixing though.

sbc100: Fixed in https://github.com/emscripten-core/emscripten/pull/24886

PR #24885, opened 2025-08-08
[file_packager] Convert checkCachedPackage to async/await. NFC

sbc100:
Followup to #24882 and #24883


sbc100:
Part 3.  Also best viewed with ignore whitespace.

test/test_browser.py
```diff
@@ -640,20 +640,22 @@ def make_main(path):
...
           var cached = 0;
-          var packages = Object.keys(Module['preloadResults']);
-          packages.forEach(function(package) {
-            var fromCache = Module['preloadResults'][package]['fromCache'];
-            if (fromCache)
-              ++ cached;
-          });
+          for (var result of Object.values(Module['preloadResults'])) {
+            var fromCache = result['fromCache'];
+            if (fromCache) {
```
kripken:
```suggestion
            if (result['fromCache']) {
```

tools/file_packager.py
```diff
@@ -1044,24 +1049,25 @@ def generate_js(data_target, data_files, metadata):
...
+                    cacheRemotePackage(db, PACKAGE_PATH + PACKAGE_NAME, packageData, {uuid:PACKAGE_UUID}, processPackageData,
+                      (error) => {
+                        console.error(error);
+                        processPackageData(packageData);
+                      });
+                  }
+                , preloadFallback);
+              }
+            })
+          }).catch(preloadFallback);
```
kripken, line 1070: PR title says "async/await", but I don't see await in the code? Maybe this is just a step to there, and the PR title should be adjusted?

tools/file_packager.py
```diff
@@ -882,20 +882,25 @@ def generate_js(data_target, data_files, metadata):
           }
         }
 
-        /* Check if there's a cached package, and if so whether it's the latest available */
-        function checkCachedPackage(db, packageName, callback, errback) {
+        /*
+         * Check if there's a cached package, and if so whether it's the latest available.
+         * Resolves to `null` if the package is missing our out-of-date.
```
```suggestion
         * Resolves to `null` if the package is missing or out-of-date.
```

kripken (approved):

tools/file_packager.py
sbc100: Sorry, what I means was the the functions themselves are now `async` , which means they now return promises instead of using callback.

The call sites (here) are still no using async/await.  That will be a followup.

I updated the PR description.

kripken: sg

I see new changes in this PR: now `cacheRemotePackage` is also modified. Was that meant to be in this PR?

kripken: Oh, I see the updated PR text now, so the scope grew. lgtm.

kripken (approved):

PR #24888, opened 2025-08-08
[test] Improve test_preload_caching_indexeddb_name. NFC

sbc100:
Split out from #24885


sbc100:
Landing these test updates separately so its clear they are not related specifically to the fix in #24885

Issue #24265, opened 2025-05-06
Deprecate and remove -sUSE_WEBGPU


kainino0x:
Emscripten's implementation of webgpu.h is unmaintained. Dawn now maintains "emdawnwebgpu", a fork of Emscripten's implementation that is up to date, resolving #23432.

We want people to move to emdawnwebgpu. However, the API has changed substantially, so it will take people time to switch over. Plus, there might be new bugs in that implementation. So we need a decently long deprecation period before removing the old implementation.

Deprecation PR: #24220
Removal TBD.

donthomasitos:
I think this should be done rather sooner than later. The API changes are manageable, and if ppl don't switch we'll never find the bugs in the `emdawnwebgpu` implementation anyway. Right now it's super confusing.

kainino0x:
Planning to do this around August 18 once there has been some time for the bugfixed Emdawnwebgpu to bake.

kainino0x:
@sbc100 how do you think this should be done? Was thinking either:
1. remove support for the option, but leave a special error message saying to use `--use-port=emdawnwebgpu`
2. remove old code, but keep `-sUSE_WEBGPU` and translate it to `--use-port=emdawnwebgpu`

donthomasitos:
I vote for option 1 🫡

kainino0x:
Thanks for the bump, I'm going to get this done now.

PR #24889, opened 2025-08-08
Roll emdawnwebgpu port to v20250807

kainino0x:
From https://github.com/google/dawn/releases/tag/v20250807.221415

This version includes the last breaking change to the core API and fixes the known showstopping bugs. It also improves the documentation included in the port file so it's easier to understand what to do with it.

After this has landed for a little while it will be a good time to finally remove -sUSE_WEBGPU from Emscripten. (#24265)


sbc100 (approved):

kainino0x:
oops, didn't realize the last roll still was one breaking change behind. Updated the test and the commit message.

PR #24890, opened 2025-08-08
Require `id` in addRunDependency/removeRunDependency

sbc100:
Since the dawn of time we've been emitting warnings when these are called without an ID.  It seem pretty reasonable to require one at this point.

One of my motivations here it to update/replace this interface with a more modern one based on promises.  Making this API consistently use IDs simplifies things.  Also, even though this could be breaking change I think its highly unlikely there are any users who are not specifying IDs, since we have had warnings in debug builds since 2012: f67ad608988b.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,10 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- The `addRunDependency`/`removeRunDependency` now assert in debug builds if
+  they are not passed an `id` parameter.  We have been issuing warnings in
+  this case since 2012 (f67ad60), so it seems highly unlikely anyone is not
+  passing IDs here. (#24890).
 - The `-sMODULARIZE` setting generates a factory function that must be called
   before the program is instantiated that run.  However, emscripten previously
   had very special case where this instantiation would happen automatically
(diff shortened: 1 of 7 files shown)


src/preamble.js
```diff
@@ -289,12 +286,9 @@ function removeRunDependency(id) {
 #endif
 
 #if ASSERTIONS
-  if (id) {
-    assert(runDependencyTracking[id]);
-    delete runDependencyTracking[id];
-  } else {
-    err('warning: run dependency removed without ID');
-  }
+  assert(id, 'removeRunDependency required an ID');
```
kripken:
```suggestion
  assert(id, 'removeRunDependency requires an ID');
```

sbc100:
Done.  I also added some more context to the PR description

kripken:
Maybe the changelog entry wasn't pushed? I don't see it here.

sbc100:
Done

ChangeLog.md
```diff
@@ -20,6 +20,10 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- The `addRunDependency`/`removeRunDependency` now assert in debug builds if
```
kripken, line 23:
```suggestion
- `addRunDependency` and `removeRunDependency` now assert in debug builds if
```

PR #24891, opened 2025-08-08
[AUDIO_WORKLET] Optimised output buffer copy

cwoffenden:
A reworking of #22753, which "improves the copy back from the audio worklet's heap to JS by 7-12x depending on the browser." From the previous description:

Since we pass in the stack for the worklet from the caller's heap, its address doesn't change. And since the render quantum size doesn't change after the audio worklet creation, the stack positions for the audio buffers do not change either. This optimisation adds one-time subarray views and replaces the float-by-float copy with a simple `set()` per channel (per output).

The existing interactive tests (written for the original PR) can be run for comparison:
```
test/runner interactive.test_audio_worklet_stereo_io
test/runner interactive.test_audio_worklet_2x_stereo_io
test/runner interactive.test_audio_worklet_mono_io
test/runner interactive.test_audio_worklet_2x_hard_pan_io
test/runner interactive.test_audio_worklet_memory_growth
test/runner interactive.test_audio_worklet_hard_pans
```
These test various input/output arrangements as well as parameters (parameters are interesting because, depending on the browser, the sizes change as the params move from static to varying).

The original benchmark of the extracted copy is still valid:

https://wip.numfum.com/cw/2024-10-29/index.html

This is tested with 32- and 64-bit wasm (which required a reordering of how structs and data were stored to avoid alignment issues).

Some explanations:
- Fixed-position output  buffer views are created once in the`WasmAudioWorkletProcessor` constructor
- Stack allocations for the `process()` call are split into aligned struct data (see the comments) and audio/param data
- The struct writes are simplified by this splitting of data
- `ASSERTIONS` are used to ensure everything fits and correctly aligns
- The tests account for size changes in the params, which can vary from a single float to 128 floats (a single float nicely showing up any 8-byte alignment issues for wasm64)

~~Future improvements: the output views are sequential, so instead of of being individual views covering each channel the views could cover one to however-many-views needed, with a single `set()` being enough for all outputs.~~


src/audio_worklet.js
```diff
@@ -28,51 +28,104 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+      var oldStackPtr = stackSave();
+      var viewDataIdx = {{{ getHeapOffset('stackAlloc(this.maxBuffers * this.bytesPerChannel)', 'float') }}};
+#if WEBAUDIO_DEBUG
+      console.log(`AudioWorklet creating ${this.maxBuffers} buffer one-time views (for a stack size of ${wwParams.stackSize} at address 0x${(viewDataIdx * 4).toString(16)})`);
+#endif
+      this.outputViews = [];
+      for (var n = this.maxBuffers; n > 0; n--) {
+        // Added in reverse so the lowest indices are closest to the stack top
+        this.outputViews.unshift(
+          HEAPF32.subarray(viewDataIdx, viewDataIdx += this.samplesPerChannel)
```
juj: This line of pre-creating the views does not seem to be safe.

The problem is that if the WebAssembly.Memory object is grown after creating this views, then these `outputViews` objects become null typed array views.

Consider:

image

So in order to be able to precreate the views, they would have to be updated when the buffer is grown.

-----------------

Note that the above behavior is super subtle. The "neutering" occurs only if it is the AudioWorker context that grows the WebAssembly Memory. If another Worker context grows the memory, then `this.outputViews` will actually remain valid to view the old heap size.

sbc100: There has been recent work to fix this issue: #24684

We cannot turn it on by default yet, but in the mean time you should be able to do something like this:  https://github.com/emscripten-core/emscripten/blob/678bcbf3d4c87e18b4e9e6edd4bd768333cb43ca/src/growableHeap.js#L7-L14

i.e. you can recreate the views (only) when needed.

cwoffenden: Thanks both for the feedback. I'll look at:

- Splitting out the view creation ✅
- Looking for changes and recreating views ✅
- At the same time I might as well copy all channels at once with a single `set()` ❌ (not possible)
- Add an audio test that grows the heap ✅

cwoffenden: Added #24931 to test the heap growing (next will fix the breakage).

cwoffenden: With #24931 this PR as-is plays back fine on Mac Chrome, Firefox and Safari (even though `HEAPF32.buffer != this.outputViews[0].buffer`). Next I'll get the AW to grow the memory to repro juj's example.

Regardless, I'll catch the change and recreate the views.

src/audio_worklet.js
```diff
@@ -28,51 +28,115 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+     * STACK_OVERFLOW_CHECK failing (since the stack will be full, and 16 bytes
+     * being the minimum allocation size due to alignments) and leaves room for
+     * a single AudioSampleFrame as a minumum.
+     */
+    createOutputViews() {
+      // These are still alloc'd to take advantage of the overflow checks, etc.
+      var oldStackPtr = stackSave();
+      var viewDataIdx = {{{ getHeapOffset('stackAlloc(this.maxBuffers * this.bytesPerChannel)', 'float') }}};
+#if WEBAUDIO_DEBUG
+      console.log(`AudioWorklet creating ${this.maxBuffers} buffer one-time views (for a stack size of ${wwParams.stackSize} at address 0x${(viewDataIdx * 4).toString(16)})`);
```
sbc100: You can use the `ptrToString` helper here (it adds the 0x prefix and the hex encoding)

cwoffenden:
It's not possible to merge the output data copy into a single `set()` (nor the input data), since the receiving float arrays are independent.

cwoffenden:
This is done, ~~though I can rebase with all the tests.~~ and merged so the tests are available.

src/audio_worklet.js
```diff
@@ -32,51 +32,119 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+      var stackMemoryStruct = (numInputs + numOutputs) * {{{ C_STRUCTS.AudioSampleFrame.__size__ }}};
+      var stackMemoryData = 0;
+      for (entry of inputList) {
+        stackMemoryData += entry.length * this.bytesPerChannel;
+      }
+      // Collect the total number of output channels (mapped to array views)
+      var outputViewsNeeded = 0;
+      for (entry of outputList) {
+        outputViewsNeeded += entry.length;
+      }
```
juj, line 118: The above loop multiplies by `* this.bytesPerChannel;` outside the loop, though the loop before that multiplies inside the loop. A small micro-opt would be to merge to multiply once outside. (though looks good to me either way)

src/audio_worklet.js
```diff
@@ -32,51 +32,119 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
-      // shouldn't be required (to be verified)
+      // context that created this processor. Even though this 'render quantum
+      // size' is fixed at 128 samples in the 1.0 spec, it will be variable in
+      // the 1.1 spec. It's passed in now, just to prove it's settable, but will
+      // eventually be a property of the  AudioWorkletGlobalScope (globalThis).
       this.samplesPerChannel = opts.samplesPerChannel;
+      this.bytesPerChannel = this.samplesPerChannel * {{{ getNativeTypeSize('float') }}};
+
+      // Creates the output views (see createOutputViews() docs)
+      this.maxBuffers = Math.min(((wwParams.stackSize - /*minimum alloc*/ 16) / this.bytesPerChannel) | 0, /*sensible limit*/ 64);
```
juj: It looks like we don't ever use `this.maxBuffers` beyond the call to `this.createOutputViews()` below?

One way to recoup some of the code size increase would be to avoid storing `this.xxx` members that aren't needed outsided this scope.

src/audio_worklet.js
```diff
@@ -32,51 +32,119 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+     * a single AudioSampleFrame as a minumum.
+     */
+    createOutputViews() {
+      // These are still alloc'd to take advantage of the overflow checks, etc.
+      var oldStackPtr = stackSave();
+      var viewDataIdx = {{{ getHeapOffset('stackAlloc(this.maxBuffers * this.bytesPerChannel)', 'float') }}};
+#if WEBAUDIO_DEBUG
+      console.log(`AudioWorklet creating ${this.maxBuffers} buffer one-time views (for a stack size of ${wwParams.stackSize} at address ${ptrToString(viewDataIdx * 4)})`);
+#endif
+      this.outputViews.length = 0;
```
juj: Maybe just have `this.outputViews.length = 0;` instead be `this.outputViews = [];`

It is an extremely rare occurrence that `createOutputViews()` needs to be called again. It will happen in an app lifetime maybe 10-20 times at most. (being generous). So there's no perf implication to always reassigning to an empty `= [];` from scratch and letting the old GC.

src/audio_worklet.js
```diff
@@ -32,51 +32,119 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+      // context that created this processor. Even though this 'render quantum
+      // size' is fixed at 128 samples in the 1.0 spec, it will be variable in
+      // the 1.1 spec. It's passed in now, just to prove it's settable, but will
+      // eventually be a property of the  AudioWorkletGlobalScope (globalThis).
       this.samplesPerChannel = opts.samplesPerChannel;
+      this.bytesPerChannel = this.samplesPerChannel * {{{ getNativeTypeSize('float') }}};
+
+      // Creates the output views (see createOutputViews() docs)
+      this.maxBuffers = Math.min(((wwParams.stackSize - /*minimum alloc*/ 16) / this.bytesPerChannel) | 0, /*sensible limit*/ 64);
+      this.outputViews = [];
```
juj: This line `this.outputViews = [];` could maybe be avoided altogether (see comment below)

src/audio_worklet.js
```diff
@@ -118,20 +171,53 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+        // The preallocated 'outputViews' already have the correct offsets and
+        // sizes into the stack (recall from createOutputViews() that they run
+        // backwards).
         for (entry of outputList) {
           for (subentry of entry) {
-            // repurposing structPtr for now
-            for (structPtr = 0; structPtr < this.samplesPerChannel; ++structPtr) {
-              subentry[structPtr] = HEAPF32[outputDataPtr++];
-            }
+            subentry.set(this.outputViews[--outputViewsNeeded]);
```
juj, line 222: This looks so powerful! 👍

src/audio_worklet.js
juj: Oh I see, we need to know the number of buffers when we call again to `this.createOutputViews();`.

Maybe then this could do

```js
this.outputViews = new Array(Math.min(((wwParams.stackSize - /*minimum alloc*/ 16) / this.bytesPerChannel) | 0, /*sensible limit*/ 64));
```

and instead of

```
this.outputViews.unshift(
```
inside `createOutputViews()`, the function would always assign to `this.outputViews[i] = `, reusing the array with fixed pre-initialized length?

juj (approved):
LGTM, the logic looks solid now.

I had a couple of minor comments, but those are cosmetic. LGTM either way, this looks good to land to me.

src/audio_worklet.js
cwoffenden: I think this has merit and will result in a saving. Assertions can be against the array length.

I'll try to take a look today.

cwoffenden:
> LGTM, the logic looks solid now.

Thanks!

src/audio_worklet.js
cwoffenden: I'll take a look over this. The most compact would probably be to take the incoming sizes (but I work on the assumption that the browser will do something wrong at some point in the future and throw the maths off).

juj:
One thing that comes to mind that may be worth probing: The TLS section is allocated at the end of the stack, (it should be exclusive of the stack size though).

So you may try to place a global variable `__thread int foo = 42;` variable in the test [test/webaudio/audioworklet_params_mixing.c](https://github.com/emscripten-core/emscripten/pull/24891/files#diff-c8d7263765ee962b990c7807762fed326cb7af0679703ebe63f2520347c1b6f7), and add an `assert(foo == 42);` to verify that its value never changes during the test. This assert would be inside the audio worklet callback process call.

This will verify that the allocation from stack top won't stomp on TLS variables.

cwoffenden:
> One thing that comes to mind that may be worth probing

[snip]

I can add this as a separate PR, internals like this were one of my worries when starting this last year (most of those worries went away once it was working, but I'm still missing insights like this).

juj:
Yeah that sounds good.

src/audio_worklet.js
cwoffenden: I factored out tracking the length. Update incoming.

src/audio_worklet.js
cwoffenden: I took a look and reused the `stackMemoryData` var (committed). I'm not sure it improves the readability, whereas for `outputViewsNeeded` it made sense. I'm not sure there are enough channels to warrant it.

cwoffenden:
Changes in as per @juj's suggestions. ~~CI is breaking but it's not related.~~

cwoffenden:
> Yeah that sounds good.

#25024 tests this.

cwoffenden:
@sbc100 if we can land this in the next days, I'll have time in the next weeks to prep for the WebAudio 1.1 API and how I'd like to extend the current API to enable support (I don't want to merge this later).

juj:
I'm happy with the PR, and code size increase is ok for the feature. I'll defer landing though to Sam, since he's still got a yellow dot on the review.

src/audio_worklet.js
```diff
@@ -32,51 +32,117 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+      this.ctorOldStackPtr = stackSave();
+#endif
+    }
+
+    /**
+     * Create up-front as many typed views for marshalling the output data as
+     * may be required (with an arbitrary maximum of 64, for the case where a
+     * multi-MB stack is passed), allocated at the *top* of the worklet's stack
+     * (and whose addresses are fixed). The 'minimum alloc' firstly stops
+     * STACK_OVERFLOW_CHECK failing (since the stack will be full, and 16 bytes
```
sbc100: What is the 'minimum alloc' here refering to?

Also there is the `rbitrary maximum of 64` in this function?   Is the comment referring to stuff outside of the createOutputViews perhaps?

src/audio_worklet.js
```diff
@@ -32,51 +32,117 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+     * audio output data back.
+     *
      * @param {Object} parameters
      */
     process(inputList, outputList, parameters) {
-      // Marshal all inputs and parameters to the Wasm memory on the thread stack,
-      // then perform the wasm audio worklet call,
-      // and finally marshal audio output data back.
+#if ALLOW_MEMORY_GROWTH
+      // recreate the output views if the heap has changed
```
sbc100: Can you add a `TODO` here to support `GROWABLE_ARRAYBUFFERS`

src/audio_worklet.js
```diff
@@ -32,51 +32,117 @@ function createWasmAudioWorkletProcessor(audioParams) {
...
+      this.outputViews = new Array(Math.min(((wwParams.stackSize - /*minimum alloc*/ 16) / this.bytesPerChannel) | 0, /*sensible limit*/ 64));
+#if ASSERTIONS
+      console.assert(this.outputViews.length > 0, `AudioWorklet needs more stack allocating (at least ${this.bytesPerChannel})`);
+#endif
+      this.createOutputViews();
+
+#if ASSERTIONS
+      // Explicitly verify this later in process(). Note to self, stackSave is a
+      // bit of a misnomer as it simply gets the stack address.
+      this.ctorOldStackPtr = stackSave();
```
sbc100, line 57: Is this used anywhere?  I would expect to see it checked somewhere.

cwoffenden: It got deleted in the multiple refactors. The original code is here:

https://github.com/cwoffenden/emscripten/blob/6b3f01eb7327c2ee7f8056a4849bdddf536fb5c1/src/audio_worklet.js#L100

I'll re-add it tomorrow.

src/audio_worklet.js
cwoffenden: If the stack size passed to the worklet is exactly a multiple of the buffers, the stack overflow check will fail because extra bytes are required. The allocator needs additional bytes to work (I didn't investigate, I'm assuming sentinel or other reason, or simply return slot) so this needs taking into account, and 16 bytes leaving free (Emscripten's minimum allocation size).

The 64 is here:

https://github.com/emscripten-core/emscripten/blob/d9e17a5a3dcc8020fa320157f736bd1b47686f83/src/audio_worklet.js#L43

Refactoring shunted this around just before the weekend. I'll tidy this tomorrow.

src/audio_worklet.js
cwoffenden: The test was re-added.

src/audio_worklet.js
cwoffenden: Done.

src/audio_worklet.js
cwoffenden: Comments moved to the correct places (and clarified).

cwoffenden:
All done from me.

cwoffenden:
Could one of you land this? Two ticks, so I’m guessing we’re all good?

juj:
Well it will be my pleasure. Thank you so much for the great effort in this - it is a superb performance optimization.

cwoffenden:
> it is a superb performance optimization

Thanks!

PR #24893, opened 2025-08-08
[file_packager] Convert fetchRemotePackage to async. NFC

sbc100:
Part 4 in the sequence following #24882, #24883 and #24885.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -3984,10 +3984,7 @@ More info: https://emscripten.org
 
     create_file('post.js', 'MyModule(Module).then(() => console.log("done"));')
 
-    self.run_process([EMCC, 'main.c', '--extern-pre-js=embed.js', '--extern-post-js=post.js', '-sMODULARIZE', '-sEXPORT_NAME=MyModule', '-sFORCE_FILESYSTEM'])
-
-    result = self.run_js('a.out.js')
-    self.assertContained('|hello world|', result)
+    self.do_runf('main.c', '|hello world|', cflags=['--extern-pre-js=embed.js', '--extern-post-js=post.js', '-sMODULARIZE', '-sEXPORT_NAME=MyModule', '-sFORCE_FILESYSTEM'])
 
   def test_preprocess(self):
     # Pass -Werror to prevent regressions such as https://github.com/emscripten-core/emscripten/pull/9661
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -635,11 +635,11 @@ def generate_js(data_target, data_files, metadata):
 
   if options.support_node:
     ret += "    var isNode = typeof process === 'object' && typeof process.versions === 'object' && typeof process.versions.node === 'string';\n"
-  ret += '    function loadPackage(metadata) {\n'
+  ret += '    async function loadPackage(metadata) {\n'
 
   code = '''
       function assert(check, msg) {
-        if (!check) throw msg + new Error().stack;
+        if (!check) throw new Error(msg);
       }\n'''
 
   # Set up folders
@@ -808,7 +808,7 @@ def generate_js(data_target, data_files, metadata):
 
         async function openDatabase() {
           if (typeof indexedDB == 'undefined') {
-            throw 'using IndexedDB to cache data can only be done on a web page or in a web worker';
+            throw new Error('using IndexedDB to cache data can only be done on a web page or in a web worker');
           }
           return new Promise((resolve, reject) => {
             var openRequest = indexedDB.open(DB_NAME, DB_VERSION);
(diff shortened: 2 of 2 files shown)


tools/file_packager.py
```diff
@@ -953,76 +953,62 @@ def generate_js(data_target, data_files, metadata):
...
+        var response = fetch(packageName);
+        if (!response.ok) {
+          return Promise.reject(`${response.status}: ${response.url}`);
+        }
 
-            if (!response.body && response.arrayBuffer) { // If we're using the polyfill, readers won't be available...
-              return response.arrayBuffer().then(callback);
-            }
+        if (!response.body && response.arrayBuffer) { // If we're using the polyfill, readers won't be available...
+          returnd response.arrayBuffer();
```
lkwinta: typo

test/test_other.py
```diff
@@ -3987,10 +3987,7 @@ def test_file_packager_modularize(self):
 
     create_file('post.js', 'MyModule(Module).then(() => console.log("done"));')
 
-    self.run_process([EMCC, 'main.c', '--extern-pre-js=embed.js', '--extern-post-js=post.js', '-sMODULARIZE', '-sEXPORT_NAME=MyModule', '-sFORCE_FILESYSTEM'])
-
-    result = self.run_js('a.out.js')
-    self.assertContained('|hello world|', result)
+    self.do_runf('main.c', '|hello world|', cflags=['-sRUNTIME_DEBUG', '--extern-pre-js=embed.js', '--extern-post-js=post.js', '-sMODULARIZE', '-sEXPORT_NAME=MyModule', '-sFORCE_FILESYSTEM'])
```
kripken: Is RUNTIME_DEBUG intentional?

tools/file_packager.py
```diff
@@ -953,76 +953,62 @@ def generate_js(data_target, data_files, metadata):
...
         Module['dataFileDownloads'] ??= {};
-        fetch(packageName)
-          .catch((cause) => Promise.reject(new Error(`Network Error: ${packageName}`, {cause}))) // If fetch fails, rewrite the error to include the failing URL & the cause.
-          .then((response) => {
-            if (!response.ok) {
-              return Promise.reject(new Error(`${response.status}: ${response.url}`));
-            }
+        var response = await fetch(packageName);
+        if (!response.ok) {
+          return Promise.reject(`${response.status}: ${response.url}`);
```
kripken: Wait, as an async function, can't this just throw an Error, and those awaiting it will get a rejection?

sbc100: Yes, that is true, IIRC those are the same thing.  Happy to throw here instead.

test/test_other.py
sbc100: Reverted

tools/file_packager.py
```diff
@@ -1079,14 +1067,20 @@ def generate_js(data_target, data_files, metadata):
...
-      if (!fetched) fetchRemotePackage(REMOTE_PACKAGE_NAME, REMOTE_PACKAGE_SIZE, (data) => {
-        if (fetchedCallback) {
-          fetchedCallback(data);
-          fetchedCallback = null;
-        } else {
-          fetched = data;
-        }
-      }, handleError);\n'''
+      if (!fetched) {
+        // Note that we don't use await here because we to execute the
```
```suggestion
        // Note that we don't use await here because we want to execute the
```

tools/file_packager.py
```diff
@@ -953,76 +953,62 @@ def generate_js(data_target, data_files, metadata):
...
         Module['dataFileDownloads'] ??= {};
-        fetch(packageName)
-          .catch((cause) => Promise.reject(new Error(`Network Error: ${packageName}`, {cause}))) // If fetch fails, rewrite the error to include the failing URL & the cause.
-          .then((response) => {
-            if (!response.ok) {
-              return Promise.reject(new Error(`${response.status}: ${response.url}`));
-            }
+        var response = await fetch(packageName);
+        if (!response.ok) {
+          throw `${response.status}: ${response.url}`;
```
RReverser: Please wrap it back into `new Errror`.

Unfortunately, Emscripten codebase already suffers from many legacy codepaths that just throw strings, but it's something I'm hoping we'll fix over time, not introduce more.

It's a subtle difference for those who are not familiar with it, but while JS supports throwing arbitrary primitives from its early days, anything but descendants of `Error` results in an error with no stacktrace and no way to find out where the error occurred or why, which makes the error messages almost useless.

tools/file_packager.py
```diff
@@ -953,76 +953,62 @@ def generate_js(data_target, data_files, metadata):
...
-            }
+        if (!response.body && response.arrayBuffer) { // If we're using the polyfill, readers won't be available...
+          return response.arrayBuffer();
+        }
 
-            const reader = response.body.getReader();
-            const iterate = () => reader.read().then(handleChunk).catch((cause) => {
-              return Promise.reject(new Error(`Unexpected error while handling : ${response.url} ${cause}`, {cause}));
-            });
+        const chunks = [];
```
RReverser, line 974: Anything here and on can be removed nowadays - all browsers we care about support `response.arrayBuffer()` so it's enough to return just that.

RReverser: Ah, or is the goal to show actual progress? Then perhaps the opposite, remove the polyfill.

tools/file_packager.py
```diff
@@ -953,76 +953,62 @@ def generate_js(data_target, data_files, metadata):
...
+        const reader = response.body.getReader();
 
-            const handleChunk = ({done, value}) => {
-              if (!done) {
-                chunks.push(value);
-                loaded += value.length;
-                Module['dataFileDownloads'][packageName] = {loaded, total};
+        while (1) {
+          var {done, value} = await reader.read();
+          if (done) break;
```
RReverser, line 984: We can also do

```suggestion
        for await (var value of reader) {
```

tools/file_packager.py
```diff
@@ -1128,7 +1122,7 @@ def generate_js(data_target, data_files, metadata):
     %(node_support_code)s
     var response = await fetch(metadataUrl);
     if (!response.ok) {
-      throw new Error(`${response.status}: ${response.url}`);
+      throw `${response.status}: ${response.url}`;
```
RReverser: Same here re:Error wrapping.

tools/file_packager.py
sbc100: Sure.

RReverser: Another option is probably to just use `abort(...)` here which does the right thing already.

tools/file_packager.py
sbc100: That is a very new JS feature though right?   I'll maybe leave that as a followup since its not a feature we currently depend onn

tools/file_packager.py
sbc100: yes, the goal is to show progress.

I don't know which polyfill is being referred to above, but its not something emscripten includes. 

Maybe it can be remove now?  It was added in #22016 but I don't see anything about if/when the polyfill might be used. Probably best to leave that to a separate cleanup PR.

tools/file_packager.py
RReverser: Hm relatively new, yeah.

lkwinta (approved):

tools/file_packager.py
kripken: `abort()` might be best, yeah.

Otherwise, I think in a release build it can make sense to throw a raw string? We don't care about the stack trace/debuggability there, and it is smaller.

sbc100: I don't think we can use abort here because this code doesn't always run inside of emscripten-generated code.. it can also run before the emscripten code loads.

PR #24898, opened 2025-08-09
Fix test_prepost_run_dep after #24890

sbc100:
- Cleanup the test
- Add and `id` to the `addRunDependency` call.
- Add `.unref` to prevent the dependency watcher from keeping the node process alive.  Prior to #24890 the lack of `id` in used in the test caused it to skip the dependency watcher.
- Add `--closure=1` to make sure closure doesn't minify the `.unref` call.
- Add some RUNTIME_DEBUG logging to add/removeRunDependency.

### DIFF
--- a/src/preamble.js
+++ b/src/preamble.js
@@ -251,6 +251,9 @@ function addRunDependency(id) {
 #endif
 
 #if ASSERTIONS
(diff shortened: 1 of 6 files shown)


sbc100:
Landing TBR to green the tree

PR #11785, opened 2020-08-01
Add support for `--preload-file` and under node

sbc100:
I did this to all me to test the pre-load plugins which don't run
with `--embed-file`.  Is it worth landing this for testing reasons
alone.  Shoudl the preload plugins work under node?

PR #24904, opened 2025-08-11
[file_packager] Remove unused `errback` fetchRemovePackage. NFC

sbc100:
This argument has been used since 3839dd4de.

Even when this function was first added in 74f3242118d, `errback` was only used for the IDB failures, not the fetch failures. Fetch failure were left to propagate to the global error handlers.

The exception here was that when node support was added in #11785 it, I believe incorrectly used errback to handle file read errors (doesn't match the web version that does *not* handle fetch error using errback).

I'm not claiming this is correct/best behaviour, this is simply an NFC to cleanup an already-unused argument.

This is small cleanup in advance of #24893.

### DIFF
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -953,18 +953,12 @@ def generate_js(data_target, data_files, metadata):
     if options.support_node:
       node_support_code = '''
         if (isNode) {
-          require('fs').readFile(packageName, (err, contents) => {
-            if (err) {
-              errback(err);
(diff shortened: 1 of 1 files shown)


kripken:
Isn't errback used on line 958 (for node)?

sbc100:
Oops, you are correct, however I believe that usage was wrong so I updated it.

sbc100:
I also updated the PR description.

PR #18443, opened 2022-12-30
Don't default to SDLv1 for un-versioned SDL includes

sbc100:
With SDL you can choose to either include the header files via the          
explictly version prefix (e.g. `<SDL/SDL_audio.h>` or                    
`<SDL2/SDL_audio.h>`.  Or some folks configure the include path such        
that the unversioned includes can be used (`#include <SDL_audio.h>`).  

Emscripten allows the unversioned form by putting the `SDL` or `SDL2`       
directory on the include path based on the USE_SDL settings, but the        
fact that were defaulting to enabling version 1 was causing some issues  
(See #18072 and #184402 for more details).                               
                                                                         
After this change we default to not adding either of these include          
paths.  This means that folks who want to use un-versioned includes will 
now need to opt into one version of the other via `-sUSE_SDL`.           
                                                                         
Note that folks using the explict `<SDL/SDL_audio.h>` includes are not   
effected by this (that will always result in an SDL v1 header being         
included).                                                               
                                                                         
In order to make it more obvious that use of `<SDL_audio.h>` now         
requires an opt in, I created dummy headers that should give effected       
users and clear error message.                                           
                                                                         
We've already been doing this in STRICT module for a while now.          
                                                                         
Fixes: #18072, #18440

PR #24907, opened 2025-08-11
Make sdl-config/sdl2-config into simple shell scripts

sbc100:
This removes the python dependency and the need to .bat and shell launchers.

These script have always be shell scripts upstream and there should be no expectation that they can be run on windows.

See https://github.com/libsdl-org/SDL/issues/3516

In addition fix `sdl-config.py` so that it explicitly includes `-sUSE_SDL`. This has been required since #18443.


ChangeLog.md
```diff
@@ -20,6 +20,8 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- The `sdl-config` and `sdl2-config` were simplified to avoid using python
+  and the `.bat` file versions were removed. (#24907)
```
```suggestion
  and the `.bat` file versions were removed, following upstream. (#24907)
```

ChangeLog.md
```diff
@@ -20,6 +20,8 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- The `sdl-config` and `sdl2-config` were simplified to avoid using python
```
```suggestion
- `sdl-config` and `sdl2-config` were simplified to avoid using python
```

kripken: or:
```suggestion
- The `sdl-config` and `sdl2-config` scripts were simplified to avoid using python
```

kripken (approved):

PR #24910, opened 2025-08-11
Mark `--preload-file` as incompatible with `MINIMAL_RUNTIME`. NFC

sbc100:
The file preloading mechanism relies on either `Module.preRun` or `Module.calledRun`, neither of which are available with `MINIMAL_RUNTIME`.


tools/link.py
```diff
@@ -893,6 +893,9 @@ def limit_incoming_module_api():
     if options.use_preload_plugins or len(options.preload_files):
       exit_with_error('MODULARIZE=instance is not compatible with --embed-file/--preload-file')
 
+  if settings.MINIMAL_RUNTIME and len(options.preload_files):
+    exit_with_error('MINIMAL_RUNTIME is not compatible --preload-file')
```
kripken:
```suggestion
    exit_with_error('MINIMAL_RUNTIME is not compatible with --preload-file')
```

kripken (approved):

PR #24913, opened 2025-08-12
Remove support for windows 7 from .bat file launchers

sbc100:
This removed that handling of `EM_WORKAROUND_WIN7_BAD_ERRORLEVEL_BUG` which is workaround for a windows 7 issue.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,8 @@ See docs/process.md for more on how version tagging works.
 
 4.0.13 (in development)
 -----------------------
+- Support for Windows 7 and the `EM_WORKAROUND_WIN7_BAD_ERRORLEVEL_BUG`
+  environment variable was removed.
 - `sdl-config` and `sdl2-config` scripts were simplified to avoid using python
   and the `.bat` file versions were removed, matching upstream SDL. (#24907)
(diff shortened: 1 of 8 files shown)


sbc100:
@juj, can you confirm that you no longer need Windows 7 support?

juj:
We no longer need Windows 7 support (we only support Windows 10 + 11), but I do not trust Windows+Python enough to drop `EM_WORKAROUND_WIN7_BAD_ERRORLEVEL_BUG` like this.

While we triaged the issue to be a Windows 7 specific issue back then, I really want to see in action that we don't get this issue in latest code if we drop this workaround.

So instead what I propose we'll do is that I'll drop this workaround env. var. from being used in production in next Emscripten version that we update Unity to, and if that goes good and we get no user bugs, then we can remove these from Emscripten altogether. So.. ask again in a year or two? :|

sbc100:
OK, sounds good.

So IIUC, you are currently setting this environment variable in the SDK that you ship?

juj:
> So IIUC, you are currently setting this environment variable in the SDK that you ship?

Yeah that's correct. Whenever we invoke Emscripten, we set the env. var.

sbc100:
OK, lets revisit once you have shipped a version with it removed for N months.

PR #24914, opened 2025-08-12
Convert `FS_handledByPreloadPlugin` and plugin `handle` callback to async.

sbc100:
The `FS_handledByPreloadPlugin` is completely internal so changing this
should be completely non-functional.

If external projects are implementing custom preload plugins they would
need to update their `handle` methods.  I've added an assertion in debug
builds to detect plugins that are not async.

This change will allow a followup which is to convert that whole file preloading
mechanism to async.

### DIFF
--- a/src/lib/libbrowser.js
+++ b/src/lib/libbrowser.js
@@ -47,32 +47,34 @@ var LibraryBrowser = {
       imagePlugin['canHandle'] = function imagePlugin_canHandle(name) {
         return !Module['noImageDecoding'] && /\.(jpg|jpeg|png|bmp|webp)$/i.test(name);
       };
-      imagePlugin['handle'] = function imagePlugin_handle(byteArray, name, onload, onerror) {
+      imagePlugin['handle'] = async function imagePlugin_handle(byteArray, name) {
         var b = new Blob([byteArray], { type: Browser.getMimetype(name) });
         if (b.size !== byteArray.length) { // Safari bug #118630
           // Safari's Blob can only take an ArrayBuffer
           b = new Blob([(new Uint8Array(byteArray)).buffer], { type: Browser.getMimetype(name) });
         }
         var url = URL.createObjectURL(b);
-        var img = new Image();
-        img.onload = () => {
+        return new Promise((resolve, reject) => {
+          var img = new Image();
+          img.onload = () => {
 #if ASSERTIONS
-          assert(img.complete, `Image ${name} could not be decoded`);
+            assert(img.complete, `Image ${name} could not be decoded`);
 #endif
-          var canvas = /** @type {!HTMLCanvasElement} */ (document.createElement('canvas'));
-          canvas.width = img.width;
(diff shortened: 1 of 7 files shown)


sbc100:
If we think folks have external plugins that we want to continue to support we could use a different name such as `handleAsync`?

src/lib/libbrowser.js
```diff
@@ -47,86 +47,99 @@ var LibraryBrowser = {
...
+            if (done) return;
+            done = true;
+            Browser.preloadedAudios[name] = new Audio(); // empty shim
+            reject();
+          }
+          var b = new Blob([byteArray], { type: Browser.getMimetype(name) });
+          var url = URL.createObjectURL(b); // XXX we never revoke this!
+  #if ASSERTIONS
+          assert(typeof url == 'string', 'createObjectURL must return a url as a string');
+  #endif
```
kripken: I thought we were removing these?

src/lib/libbrowser.js
```diff
@@ -47,86 +47,99 @@ var LibraryBrowser = {
...
+            if (done) return;
+            done = true;
+            Browser.preloadedAudios[name] = audio;
+            resolve(byteArray);
+          }
+          function fail() {
+            if (done) return;
+            done = true;
+            Browser.preloadedAudios[name] = new Audio(); // empty shim
+            reject();
```
kripken: Is this from older code?

sbc100: Sorry, bad rebase

src/lib/libdylink.js
```diff
@@ -17,27 +17,26 @@ var LibraryDylink = {
     // Use string keys here to avoid minification since the plugin consumer
     // also uses string keys.
     var wasmPlugin = {
-      'promiseChainEnd': Promise.resolve(),
+      promiseChainEnd: Promise.resolve(),
```
kripken: The comment above suggests quoting is necessary, and the lines below use quoting. Is this line special somehow in not needing it?

sbc100: Not for this attribute, which is internal to this one plugin, not part of the public interface.

sbc100: I can split out this change if you prefer.

kripken: No need to split it out, but maybe move it to a more separate place? Like if the comment about quoting were below it, things would be clearer I think.

sbc100: Already split it out

sbc100:
What do you think about changing the behaviour of the `handle` callback?  Should we try to keep the old behaviour too?  At the cost of code size?  Or maybe we could somehow detect the old style in debug builds?

sbc100:
Added and assertion to catch external plugins that might need updating.

kripken:
I don't think we need to support the old style. While people might be making new plugins, I imagine not many, and this is a good modernization to the API.

kripken (approved):

PR #24917, opened 2025-08-13
Replace `FS_createPreloadedFile` with async version

sbc100:
The new `FS_preloadFile` API replaces the `FS_createPreloadedFile` API.

Left the old version around for compatibility.

Followup to #24914

### DIFF
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -677,11 +677,14 @@ def generate_js(data_target, data_files, metadata):
           % (start / (1024 * 1024)))
 
     create_preloaded = '''
-          Module['FS_createPreloadedFile'](this.name, null, byteArray, true, true,
-            () => Module['removeRunDependency'](`fp ${that.name}`),
-            () => err(`Preloading file ${that.name} failed`),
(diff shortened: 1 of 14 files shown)


tools/file_packager.py
```diff
@@ -677,10 +677,12 @@ def generate_js(data_target, data_files, metadata):
           % (start / (1024 * 1024)))
 
     create_preloaded = '''
-          Module['FS_createPreloadedFile'](this.name, null, byteArray, true, true,
-            () => Module['removeRunDependency'](`fp ${that.name}`),
-            () => err(`Preloading file ${that.name} failed`),
-            false, true); // canOwn this data in the filesystem, it is a slide into the heap that will never change\n'''
+          try {
+            await Module['FS_preloadFile'](this.name, null, byteArray, true, true, false, true); // canOwn this data in the filesystem, it is a slide into the heap that will never change
```
```suggestion
            await Module['FS_preloadFile'](this.name, null, byteArray, true, true, false, true); // canOwn this data in the filesystem, it is a slice into the heap that will never change
```
Pre-existing, but may as well fix it

kripken (approved):

PR #24919, opened 2025-08-13
[draft] Support running parallel browser tests.

brendandahl:
I want to test this out on circle ci to see how much it helps. Locally, the full browser suite runs in 160s vs 2,057s, but I've got 128 cores.

Still need to cleanup and comment on this some more.


brendandahl:
The medium instance only has 2 cores (4 w/ hyperthread?), but `test-browser-chrome-2gb` does see a nice speedup going from ~44m->~22m. May be worth trying bigger instances.

arsnyder16:
> but I've got 128 cores.
👀

brendandahl:
New pr #25069

PR #24925, opened 2025-08-14
Avoid `wasm-instantiate` run dependency when not using WASM_ASYNC_COMPILATION. NFC

sbc100:


src/preamble.js
```diff
@@ -895,11 +895,14 @@ function getWasmImports() {
 #if DECLARE_ASM_MODULE_EXPORTS
     assignWasmExports(wasmExports);
 #endif
+#if WASM_ASYNC_COMPILATION
     removeRunDependency('wasm-instantiate');
+#endif
     return wasmExports;
   }
-  // wait for the pthread pool (if any)
```
sbc100, line 901: This comment was simply wrong.  The run dependency for the pthread pool is in libpthread.js.

kripken (approved):

PR #24931, opened 2025-08-14
[AUDIO_WORKLET] Add interactive heap growing test NFC

cwoffenden:
This adds an interactive test to force growing the heap during playback:
```
test/runner interactive.test_audio_worklet_memory_growth
```
Tested with `interactive64` and `interactive_2gb` (for `interactive64_4gb` the heap is already at the browser's max in testing so can't be grown).

It works by alloc'ing and leaking 2/3 of the current size until it can no longer do so. Emscripten regrows its wasm memory in the process, invalidating any data views (see #24891).

**Edit: test can now grow from both the main and audio thread.**


test/webaudio/audioworklet_grow_heap.c
```diff
@@ -0,0 +1,126 @@
+#include <assert.h>
```
sbc100: How about calling this `audio_worklet_memory_growth` or `audio_worklet_mem_growth`

cwoffenden: Done!

cwoffenden:
I'd be good to land this and its companion hard-pan test. Both are interactive so don't affect the CI.

PR #24933, opened 2025-08-14
Update feature matrix to highlight unconditionally required DedicatedWorkerGlobalScope.name parameter.

juj:
Way back I recall I did not want to target Worker's `name` parameter given it had browser version requirements, and listed to be a [debugging feature](https://html.spec.whatwg.org/multipage/workers.html#dom-worker-dev).

Though now that we have crossed that line, update minimum browser version feature matrix to require `name` support (e.g. after #24190), up[date the minimum required browser versions to highlight this fact.

We could maintain separate min required versions for singlethreaded and multithreaded builds, although given that the minimum browser version bump here is relatively small, probably no bother.

Although the version bump does mean that Emscripten does no longer support iPhone 5 devices. Fortunately they are very old by now, so practically absent.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -15226,8 +15226,8 @@ out.js
     self.assertTextDataIdentical(expected, response)
 
   def test_min_browser_version(self):
-    err = self.expect_fail([EMCC, test_file('hello_world.c'), '-Wno-transpile', '-Werror', '-sWASM_BIGINT', '-sMIN_SAFARI_VERSION=120000'])
-    self.assertContained('emcc: error: MIN_SAFARI_VERSION=120000 is not compatible with WASM_BIGINT (150000 or above required)', err)
+    err = self.expect_fail([EMCC, test_file('hello_world.c'), '-Wno-transpile', '-Werror', '-sWASM_BIGINT', '-sMIN_SAFARI_VERSION=130000'])
+    self.assertContained('emcc: error: MIN_SAFARI_VERSION=130000 is not compatible with WASM_BIGINT (150000 or above required)', err)
 
     err = self.expect_fail([EMCC, test_file('hello_world.c'), '-Wno-transpile', '-Werror', '-pthread', '-sMIN_CHROME_VERSION=73'])
     self.assertContained('emcc: error: MIN_CHROME_VERSION=73 is not compatible with pthreads (74 or above required)', err)
@@ -15241,7 +15241,7 @@ out.js
     self.assertNotContained('--signext-lowering', err)
 
     # Specifying an older browser version should trigger the lowering pass
-    err = self.run_process(cmd + ['-sMIN_SAFARI_VERSION=120000'], stderr=subprocess.PIPE).stderr
+    err = self.run_process(cmd + ['-sMIN_SAFARI_VERSION=120200'], stderr=subprocess.PIPE).stderr
     self.assertContained('--signext-lowering', err)
     err = self.run_process(cmd + ['-sMIN_FIREFOX_VERSION=61'], stderr=subprocess.PIPE).stderr
     self.assertContained('--signext-lowering', err)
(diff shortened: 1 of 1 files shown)


tools/feature_matrix.py
```diff
@@ -15,12 +15,14 @@
...
+# Emscripten unconditionally requires support for:
+# - DedicatedWorkerGlobalScope.name parameter for multithreading support, which
+#   landed first in Chrome 70, Firefox 55 and Safari 12.2.
 # TODO(sbc): Design a of policy for managing these values.
-OLDEST_SUPPORTED_CHROME = 55  # December 1, 2016
-OLDEST_SUPPORTED_FIREFOX = 50  # November 15, 2016
-OLDEST_SUPPORTED_SAFARI = 101000  # September 20, 2016
+OLDEST_SUPPORTED_CHROME = 70  # October 16, 2018
+OLDEST_SUPPORTED_FIREFOX = 55  # August 08, 2017
+OLDEST_SUPPORTED_SAFARI = 120200  # March 25, 2019
```
sbc100: These need to be kept in sync with the docs in for  `MIN_CHROME_VERSION` etc in `src/settings.js` I think.

(Can you update comment above to mention this too)

juj: The src/settings.js `min_*_version` fields are already newer than these.

sbc100: I meant in the comments.  e.g.:

```
// Minimum supported value is 55, which was released on 2016-12-01 (see               
// feature_matrix.py).     
```

(these comments regarding the min versions would need updating).

juj: Oh I see, ok.

juj:
> If we can bump these version that would be great!
> 
> @kripken @dschuff any objections?

To note that this PR is not a version bump, but the bump has already occurred in several previous PRs. We are in it already.

sbc100:
> > If we can bump these version that would be great!
> > @kripken @dschuff any objections?
> 
> To note that this PR is not a version bump, but the bump has already occurred in several previous PRs. We are in it already.

Thats only true for users using pthreads or wasm workers I think though?   Those users are already tend to be targetting more recent browers anyway.  For users targetting legacy browsers its unlikely they will be expecting pthreads to work.

src/settings.js
```diff
@@ -1885,7 +1885,7 @@ var MIN_FIREFOX_VERSION = 79;
 // older, i.e. iPhone 4s, iPad 2, iPad 3, iPad Mini 1, Pod Touch 5 and older,
 // see https://github.com/emscripten-core/emscripten/pull/7191.
 // MAX_INT (0x7FFFFFFF, or -1) specifies that target is not supported.
-// Minimum supported value is 101000 which was released in 2016-09 (see
+// Minimum supported value is 120200 which was released on 2019-03-25 (see
```
kripken, line 1890: The rest of the comment above might need updates too. "14.1" should be 15, unrelated to this PR though. And iPhone 4s should be 5 based on the PR description?

sbc100:
Codesize failures will be fixed by rebase/merge I believe.

sbc100:
I'm excited to make some cleanups once this lands.. but maybe will wait a release or two to see if there is feedback.

src/settings.js
juj: Added a second note to specifically list multithreaded code, since the above refers to singlethreaded code.

juj:
> I'm excited to make some cleanups once this lands.. but maybe will wait a release or two to see if there is feedback.

Cool. I was thinking maybe multithreading in general might be one feature to use the Feature mechanism for, if we wanted to distinguish singlethreaded vs multithreaded support rigorously.

src/settings.js
kripken: Thanks!

PR #24940, opened 2025-08-14
Move `createWasm` call from python to JS template. NFC

sbc100:
We should always prefer JS template code to python-generated JS (IMHO).

### DIFF
--- a/src/parseTools.mjs
+++ b/src/parseTools.mjs
@@ -80,6 +80,11 @@ export function preprocess(filename) {
       .replace(/\bimport\.meta\b/g, 'EMSCRIPTEN$IMPORT$META')
       .replace(/\bawait import\b/g, 'EMSCRIPTEN$AWAIT$IMPORT');
   }
+  if (MODULARIZE) {
+    // Same for out use of "top-level-await" which is not actually top level
+    // in the case of MODULARIZE.
+    text = text.replace(/\bawait createWasm\(\)/g, 'EMSCRIPTEN$AWAIT(createWasm())');
+  }
   // Remove windows line endings, if any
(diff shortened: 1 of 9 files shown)


src/postamble.js
```diff
@@ -6,6 +6,23 @@
 
 // === Auto-generated postamble setup entry stuff ===
 
+#if MODULARIZE == 'instance'
+var wasmExports;
+#elif WASM_ASYNC_COMPILATION
+
+#if MODULARIZE
+// In modularize mode the generated code is within a factory function so we
+// can use await here (since its not top-level-await).
```
```suggestion
// can use await here (since it's not top-level-await).
```

src/parseTools.mjs
```diff
@@ -80,6 +80,11 @@ export function preprocess(filename) {
       .replace(/\bimport\.meta\b/g, 'EMSCRIPTEN$IMPORT$META')
       .replace(/\bawait import\b/g, 'EMSCRIPTEN$AWAIT$IMPORT');
   }
+  if (MODULARIZE) {
+    // Same for out use of "top-level-await" which is not actually top level
```
kripken, line 84:
```suggestion
    // Same for our use of "top-level-await" which is not actually top level
```

src/postamble.js
```diff
@@ -6,6 +6,23 @@
...
+var wasmExports = await createWasm();
+#else
+var wasmExports;
+createWasm();
+#endif
+
+#else
+var wasmExports = createWasm();
+#endif
```
kripken, line 27: Maybe add `//` comments here? The multiple nested ifs make this hard to read.

kripken (approved):

PR #24943, opened 2025-08-15
emrun.py fixes

juj:
Fix emrun.py to work with undefined EXITSTATUS, and fix emrun to not print spammy errors in different error conditions, like if browser is closed at just the right moment, or socket is torn down by peer before emrun is shutting down.

### DIFF
--- a/emrun.py
+++ b/emrun.py
@@ -570,7 +570,7 @@ class HTTPWebServer(socketserver.ThreadingMixIn, HTTPServer):
 
 # Processes HTTP request back to the browser.
 class HTTPHandler(SimpleHTTPRequestHandler):
-  def send_head(self):
+  def send_headers(self):
     self.protocol_version = 'HTTP/1.1'
     global page_last_served_time
     path = self.translate_path(self.path)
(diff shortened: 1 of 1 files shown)


emrun.py
```diff
@@ -707,10 +716,21 @@ def do_POST(self):
         have_received_messages = True
       elif data.startswith('^exit^'):
         if not emrun_options.serve_after_exit:
-          page_exit_code = int(data[6:])
+          # Emrun callers
+          try:
+            # Treat sending 'undefined' (i.e. EXITSTATUS = undefined) as exit code zero...
+            page_exit_code = (0 if data == '^exit^undefined' else int(data[6:]))
```
sbc100, line 722: Its seems odd to me that this should be possible. 

When can it happen that `^exit^` is triggered without an exit status at all?   Should we treat this instead as an error?   Or as `Unexpected page exit result` perhaps?

juj: I don't quite remember anymore what the flow was where that happened, this was a few months ago that I triaged this. I'd be fine landing it with this, and iterating to adjust the flow when/if this comes up again.

juj:
I'll close this until I have a chance to diagnose the flow closer.

PR #24944, opened 2025-08-15
Fix emrun with EXIT_RUNTIME=0

juj:
I run emrun-based test harness on https://github.com/juj/wasm_webgpu and it uses `EXIT_RUNTIME=0` in the harness without problems. Not sure why this was deemed to be not working, but all the tests there are based on this mode.

After I updated Emscripten, the wasm_webgpu tests stopped working when this check had been added.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -734,9 +734,6 @@ def phase_linker_setup(options, linker_args):  # noqa: C901, PLR0912, PLR0915
     options.post_js.append(utils.path_from_root('src/emrun_postjs.js'))
     if settings.MINIMAL_RUNTIME:
       exit_with_error('--emrun is not compatible with MINIMAL_RUNTIME')
-    # emrun mode waits on program exit
-    if user_settings.get('EXIT_RUNTIME') == '0':
-      exit_with_error('--emrun is not compatible with EXIT_RUNTIME=0')
     settings.EXIT_RUNTIME = 1
     # emrun_postjs.js needs this library function.
     settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$addOnExit']


juj:
Not necessarily.

You can use emrun to launch a page, and then manually close the page. Emrun will detect that browser was closed, and shut itself down as well.

Or you can use emrun to launch a page, and then programmatically call window.close() to close the window and the browser, emrun will then again detect that browser was closed.

Or you can call emrun_quit() to close the page (which calls window.close() internally), for the same effect.

You can call emrun_quit() or window.close() asynchronously if your program is asynchronous. E.g. wasm_webgpu tests are asynchronous because acquiring a WebGPU adapter and device are asynchronous tasks.

sbc100:
As an actual user of emrun I'll defer to you on this.

QQ: If I run hello_world (unmodified) using emrun (but with out -sEXIT_RUNTIME) then the program will appear to run forever, right?  I guess this is working as intended since programs without `-sEXIT_RUNTIME` don't exit per say.

sbc100 (approved):

tools/link.py
```diff
@@ -734,9 +734,6 @@ def phase_linker_setup(options, linker_args):  # noqa: C901, PLR0912, PLR0915
     options.post_js.append(utils.path_from_root('src/emrun_postjs.js'))
     if settings.MINIMAL_RUNTIME:
       exit_with_error('--emrun is not compatible with MINIMAL_RUNTIME')
-    # emrun mode waits on program exit
-    if user_settings.get('EXIT_RUNTIME') == '0':
-      exit_with_error('--emrun is not compatible with EXIT_RUNTIME=0')
     settings.EXIT_RUNTIME = 1
```
sbc100, line 737: Wait.. shouldn't we also delete this line then?

The reason for the above error is that we are about to set `-sEXIT_RUNTIME=1` .. so if the user explicitly asks for `-sEXIT_RUNTIME=0` then that is a contradiction.

How about we change the above error into a warnings: `warning: ignoring -sEXIT_RUNTIME=0 because --emrun used`.

What I'm trying to avoid is having emcc silently ignore the user's command line settings (or at least tell the user when two settings contradict each other, as in this case).

juj: Oh I see, the line `settings.EXIT_RUNTIME = 1` will override user's command line if user passes `emcc --emrun -sEXIT_RUNTIME=0`. Hmm that's not good.

Actually I'm now confused how wasm_webgpu test suite was working with that line 740 with `settings.EXIT_RUNTIME = 1` in place. Or maybe the user command line setting is applied only afterwards, so line 740 won't override it?

juj:
I'll close this until I get a chance to verify how the EXIT_RUNTIME override behavior works.

PR #24946, opened 2025-08-15
Skip test_pthread_create_sync_instantiation in ESM_INTEGRATION mode. NFC

sbc100:
This test was added in #24938 but can't run in ESM_INTEGRATION mode


test/test_core.py
```diff
@@ -9169,6 +9169,8 @@ def test_fpic_static(self):
   })
   @node_pthreads
   def test_pthread_create(self, args):
+    if self.get_setting('WASM_ESM_INTEGRATION') and args:
```
kripken, line 9172: The `'and args'` pattern seems brittle, as we may add more to the list on 9167-9168, and forget to update this. How about checking for async compilation in the args?

sbc100:
Landing TBR to green the tree.

PR #24949, opened 2025-08-15
Refactor pthread and wasm worker init sequence

sbc100:
Rather than running `createWasm` and then returning a promise based on the `wasmModuleReceived` we now do something closer to the way it works under `MINIMUL_RUNTIME`: We delay both `createWasm` and `run` calls until the module is received via postMessage.

### DIFF
--- a/src/postamble.js
+++ b/src/postamble.js
@@ -6,25 +6,37 @@
 
 // === Auto-generated postamble setup entry stuff ===
 
-#if MODULARIZE == 'instance'
 var wasmExports;
-#elif WASM_ASYNC_COMPILATION
+
+#if MODULARIZE != 'instance'
+
+#if WASM_WORKERS || PTHREADS
+if ({{{ ENVIRONMENT_IS_MAIN_THREAD() }}}) {
+// Call createWasm on startup if we are the main thread.
+// Worker threads call this once they receive the module via postMessage
+#endif
+
+#if WASM_ASYNC_COMPILATION
 
 #if MODULARIZE
 // In modularize mode the generated code is within a factory function so we
 // can use await here (since it's not top-level-await).
-var wasmExports = await createWasm();
+wasmExports = await createWasm();
 #else
-// With async instantation wasmExports is assigned asyncronously when the
-// the instance is received.
-var wasmExports;
+// With async instantation wasmExports is assigned asynchronously when the
+// instance is received.
 createWasm();
 #endif
 
 #else
-// With sync compilation wasmExports is directly returned from createWasm()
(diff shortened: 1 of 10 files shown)


sbc100:
This is needed for #24925 and the other run-dependency refactoring I'm working on.

src/postamble.js
```diff
@@ -6,26 +6,38 @@
...
-#if MODULARIZE == 'instance'
 var wasmExports;
-#elif WASM_ASYNC_COMPILATION
+
+#if MODULARIZE != 'instance'
+
+#if WASM_WORKERS || PTHREADS
+if ({{{ ENVIRONMENT_IS_MAIN_THREAD() }}}) {
+// Call createWasm on startup if we are the main thread.
+// Worker threads call this once they recieve the module via postMessage
```
```suggestion
// Worker threads call this once they receive the module via postMessage
```

src/postamble.js
```diff
@@ -6,26 +6,38 @@
...
+#endif
+
+#if WASM_ASYNC_COMPILATION
 
 #if MODULARIZE
 // In modularize mode the generated code is within a factory function so we
-// can use await here (since it's not top-level-await).
-var wasmExports = await createWasm();
+// can use await here (since its not top-level-await).
```
```suggestion
// can use await here (since it's not top-level-await).
```
No, this was right before. "since it is not [..]" - this is a contraction of "it is" to "it's", with an apostrophe.

src/runtime_pthread.js
```diff
@@ -122,7 +122,13 @@ if (ENVIRONMENT_IS_PTHREAD) {
         Module['wasm'] = msgData.wasmModule;
         loadModule();
 #else
-        wasmModuleReceived(msgData.wasmModule);
+        wasmModule = msgData.wasmModule;
+#if MODULARIZE == 'instance'
+        init()
```
kripken:
```suggestion
        init();
```

src/wasm_worker.js
```diff
@@ -18,7 +18,13 @@ function startWasmWorker(props) {
   Module['wasm'] = props.wasm;
   loadModule()
 #else
-  wasmModuleReceived(props.wasm);
+  wasmModule = props.wasm;
+#if MODULARIZE == 'instance'
+  init()
```
kripken:
```suggestion
  init();
```

src/wasm_worker.js
```diff
@@ -18,7 +18,13 @@ function startWasmWorker(props) {
...
+#else
+  createWasm()
```
kripken:
```suggestion
  createWasm();
```

PR #24952, opened 2025-08-16
Move add/removeRunDependency to library code. NFC

sbc100:
This change just moves the code in the JS library.  For now we still included these symbol unconditionally.  I follow up by trying to limit the inclusion to when its actually needed.


tools/gen_struct_info.py
```diff
@@ -219,6 +219,7 @@ def generate_cmd(js_file_path, src_file_path, cflags):
                                '-Werror',
                                '-Wno-format',
                                '-sBOOTSTRAPPING_STRUCT_INFO',
+                               '-sWASM_ASYNC_COMPILATION=0',
```
kripken, line 222: Unrelated?

src/modules.mjs
```diff
@@ -430,8 +430,6 @@ function exportRuntimeSymbols() {
   // All possible runtime elements that can be exported
   let runtimeElements = [
     'run',
-    'addRunDependency',
-    'removeRunDependency',
```
kripken, line 434: Is this a breaking change we need to mention in the changelog? I think exporting these as runtime elements might no longer work?

sbc100: No, this is actually NFC, or at least intended to me.

These symbols just moved from runtime symbols to library symbols, but library symbols can be exported both using `-sEXPORTED_RUNTIME_METHODS=foo` and `-sDEFAULT_LIBRARY_FUNCS_TO_INCLUDE=$foo`.. so this should be be user visible.

tools/gen_struct_info.py
sbc100: This is needed since libcore.js is not included when boostrapping so we can't use the addRunDependencies mechanism while boostrapping after this change.

src/modules.mjs
kripken: Yes, I suppose some test would have failed if this were breaking.

PR #24954, opened 2025-08-17
Set CMAKE_CXX26_COMPILE_FEATURES to cxx_std_26 in toolchain file

TheGoldMonkey:
Set CMAKE_CXX26_COMPILE_FEATURES to cxx_std_26 in toolchain file. Without this, when setting target_compile_features(target PUBLIC cxx_std_26), the `-std=gnu++26` doesn't get added to the compiler flags.


cmake/Modules/Platform/Emscripten.cmake
```diff
@@ -209,6 +209,7 @@ if (EMSCRIPTEN_FORCE_COMPILERS)
         set(CMAKE_CXX23_COMPILE_FEATURES "cxx_std_23")
         set(CMAKE_CXX_COMPILE_FEATURES "${CMAKE_CXX_COMPILE_FEATURES};cxx_std_23")
         if ("${CMAKE_VERSION}" VERSION_GREATER_EQUAL "3.25") # 3.25+
+          set(CMAKE_CXX26_COMPILE_FEATURES "cxx_std_26")
```
sbc100, line 212: I guess that `test_cmake_compile_features` currently fails with 3.25 then?  And this line would fix it?

TheGoldMonkey: No I didn't look at that test at all. I was using this toolchain file, and set target_compile_definitions(target cxx_std_26) on a target, and saw that it wasn't correctly adding the flag as it should before this change. https://cmake.org/cmake/help/latest/command/target_compile_features.html

When doing cxx_std_23, it correctly added the -std=gnu++23. After this change, it correctly added the -std=gnu++26.

PR #24956, opened 2025-08-18
Fix test other.test_browser_language_detection

juj:
Fix test other.test_browser_language_detection to allow any language (e.g. en_US, en_FI, fi_FI, and so on). This allows developers who are not en_US to also run the test.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -6912,7 +6912,7 @@ int main(int argc, char **argv) {
 
     # We support both "C" and system LANG here since older versions of node do
     # not expose navigator.languages.
-    self.assertContained(f'LANG=({expected_lang}|C.UTF-8)', self.run_js('a.out.js'), regex=True)
+    self.assertContained(f'LANG=({expected_lang}|en_US.UTF-8|C.UTF-8)', self.run_js('a.out.js'), regex=True)
 
     # Accept-Language: fr,fr-FR;q=0.8,en-US;q=0.5,en;q=0.3
     create_file('pre.js', 'var navigator = { language: "fr" };')


test/test_other.py
```diff
@@ -6901,7 +6901,8 @@ def test_browser_language_detection(self):
     # We support both "C" and "en_US" here since older versions of node do
     # not expose navigator.languages.
     lang = os.environ.get('LANG', 'C.UTF-8')
-    self.assertContained(f'LANG=({lang}|en_US.UTF-8|C.UTF-8)', self.run_js('a.out.js'), regex=True)
+    # User may have their own language set, so match regex ".._.." to catch any language, e.g. en_US, en_FI, etc.
```
sbc100: That line above is supposed to capture that.  

i.e. `lang = os.environ.get('LANG', 'C.UTF-8')` is supposed to detect the users current lang. 

Is it not working windows?  Perhaps there is another way to detect the current langauge there, or we can just add `@no_windows` since we are depending on `$LANG` which I guess is not a windows thing?

juj: > we are depending on `$LANG` which I guess is not a windows thing?

To my understanding the test intent here is that `getenv("LANG")` works in the Emscripten compiled code, and it should pull from `navigator.languages`. That is behavior on the compiled output side, not the compiler host side.

The earlier line seems to be some papering to old version of node. In my node I get `fi_EN` as the language.

juj: Oh now I understand the intent with `LANG`, got confused by the comment.

> we can just add `@no_windows` since we are depending on `$LANG` which I guess is not a windows thing?

That would be weakening test coverage on Windows, so not sure that would be good.

sbc100: In that case we should find a way to detect the current language on windows (i.e. a replacement for `getenv('LANG')`).

Perhaps we can use a python API to find the current language python is using and assume the node subprocess will do that same (that is what the `getenv('LANG')` is trying to do here).

sbc100: An alternative would be find a way to force node run in a particular locale and then verify that that is reported correctly when we run emscripten.

test/test_other.py
```diff
@@ -6901,7 +6901,8 @@ def test_browser_language_detection(self):
     # We support both "C" and "en_US" here since older versions of node do
     # not expose navigator.languages.
     lang = os.environ.get('LANG', 'C.UTF-8')
```
sbc100: Maybe rename `lang` to `expected_lang` or `host_lang` to make this more clear for future readers.

test/test_other.py
juj: > In that case we should find a way to detect the current language on windows (i.e. a replacement for `getenv('LANG')`).

Why?

The actual locale is not relevant to the test here. Just read whatever the locale is, why make things complicated?

sbc100: The point of this test is to verify that the actual machine locale is reported here, not some fake/dummy value.

sbc100: I guess we could run `node -e 'console.log(navigator.languages[0])'` to capture the expected value?   Or better still use the current python locale, looks like you can do `locale.getlocale()`

juj: ok, how about now?

sbc100 (approved):

test/test_other.py
sbc100: Out of interest did `locale.getlocale()` not work?

juj: Unfortunately not 
image

test/test_other.py
```diff
@@ -6898,10 +6898,20 @@ def test_browser_language_detection(self):
     # Test HTTP Accept-Language parsing by simulating navigator.languages #8751
     self.run_process([EMCC,
                       test_file('test_browser_language_detection.c')])
-    # We support both "C" and "en_US" here since older versions of node do
+
+    expected_lang = os.environ.get('LANG', None)
```
sbc100: You can drop the `, None` here, since that is the default.

sbc100:
The lsan failures an `main` are a known issue due to and llvm-side change, which has now been reverted.

Feel free to ignore those failures when landing this.

PR #24958, opened 2025-08-18
Skip all core tests that require MODULARIZE from being run in STRICT_JS mode.

juj:
Skip all core tests that require MODULARIZE from being run in STRICT_JS mode, since the two are not compatible.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -6332,7 +6332,7 @@ PORT: 3979
   @with_env_modify({'LC_ALL': 'latin-1', 'PYTHONUTF8': '0', 'PYTHONCOERCECLOCALE': '0'})
   @crossplatform
   @no_modularize_instance('uses MODULARIZE')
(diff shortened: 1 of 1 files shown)


test/test_core.py
```diff
@@ -8025,6 +8027,7 @@ def get_wat_addr(call_index):
       self.assertLessEqual(dwarf_addr, end_wat_addr)
 
   @no_modularize_instance('uses -sMODULARIZE')
+  @no_strict_js('This test verifies behavior in MODULARIZE mode, and STRICT_JS is not compatible with MODULARIZE')
```
sbc100: This string can just be `MODULARIZE is not compatible with STRICT_JS`, which is the format we use elsewhere for this types of skips.

juj: check

juj:
Yes, we are looking to "upgrade" to shipping with `use strict;` in the future.

PR #24960, opened 2025-08-18
Fix tests that depend on UTF8ToString().

juj:
E.g. `wasmfs` test mode pulls in less JS code, so don't get to implicitly/automatically get in `UTF8ToString()`.


test/test_core.py
```diff
@@ -1963,15 +1963,18 @@ def test_inlinejs4(self):
 
   def test_em_asm(self):
     self.maybe_closure()
+    self.set_setting('DEFAULT_LIBRARY_FUNCS_TO_INCLUDE', ['$UTF8ToString'])
```
sbc100: These days we can do this the C++ source code it self using `EM_JS_DEPS(deps, "$UTF8ToString")`

juj: ok

sbc100 (approved):

sbc100:
Sorry, there was another recent codesize change #24970, so you might need to rebase again.

PR #24962, opened 2025-08-18
Fix regression to enable specifying EMTEST_BROWSER env. var in native path format on Windows.

juj:
Fix regression to enable specifying EMTEST_BROWSER env. var in native path format on Windows.

This is something that used to work a long time ago, and at some refactor got lost. On Windows environment variables that point to directories or files canonically use native path slashes `\` to do so.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -2343,8 +2343,10 @@ class BrowserCore(RunnerCore):
     if WINDOWS:
       # On Windows env. vars canonically use backslashes as directory delimiters, e.g.
       # set EMTEST_BROWSER=C:\Program Files\Mozilla Firefox\firefox.exe
-      # and spaces are not escaped, so convert backslashes to forward slashes to feed to shlex split.
-      EMTEST_BROWSER = EMTEST_BROWSER.replace('\\', '/')
+      # and spaces are not escaped. But make sure to also support args, e.g.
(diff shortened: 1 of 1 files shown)


test/common.py
```diff
@@ -2340,6 +2340,11 @@ def browser_open(cls, url):
     if not EMTEST_BROWSER:
       logger.info('No EMTEST_BROWSER set. Defaulting to `google-chrome`')
       EMTEST_BROWSER = 'google-chrome'
+    if WINDOWS:
+      # On Windows env. vars canonically use backslashes as directory delimiters, e.g.
+      # set EMTEST_BROWSER=C:\Program Files\Mozilla Firefox\firefox.exe
+      # and spaces are not escaped, so convert backslashes to forward slashes to feed to shlex split.
```
sbc100: This doesn't handle the space problem does it?    I think would would need to use quotes or escape the spaces in this case.

sbc100 (approved):
lgtm % comment

test/common.py
juj: Good point, there was actually something in subprocess.Popen() that papered over bad quotes. I.e. running `subprocess.Popen(['C:\Program', 'Files\Mozilla', 'Firefox\firefox.exe'])` surprisingly worked.

Adjusted the split to account for spaces.

PR #8711, opened 2019-05-30
Port LeakSanitizer to emscripten

quantum5:
This seems sufficiently stable for an initial version.

The bullet, poppler, sqlite3 tests all run fine under LSan, without false positives. Leaks were introduced into sqlite3 and it was caught by LSan.

Limitations:

* Reduced execution speed, but it's much better after #8778 
* No pthread support for now

PR #24963, opened 2025-08-18
Remove wasm offset converter

sbc100:
The v8 bug that made this necessary was fixed in 2019: https://crbug.com/v8/9172

From running some tests I confirmed the node v12 and newer work fine without the offset converter.  Older node versions such as v10 will show a warning about the legacy backtrace format and will not see symbol names in their backtraces (when using ASAN etc).

Fixes: #24587


RReverser:
Don't have time to do a thorough review right now, but looks like a nice cleanup!

src/lib/libstack_trace.js
```diff
@@ -104,19 +104,14 @@ var LibraryStackTrace = {
...
       // so we try using the offset converter. If that doesn't work,
       // we pack index and offset into a "return address"
-#if !USE_OFFSET_CONVERTER
-      abort('Legacy backtrace format detected but -sUSE_OFFSET_CONVERTER not present.')
-#else
-#if ASSERTIONS
-      assert(wasmOffsetConverter, 'wasmOffsetConverter global not found');
-#endif
-      return wasmOffsetConverter.convert(+match[1], +match[2]);
+      warnOnce('legacy backtrace format detected, this version of v8 is not longer supported by emscripten backtrace mechanism')
```
kripken: The comment above is no longer relevant. And I think we can make this an `abort()`?

src/lib/libstack_trace.js
```diff
@@ -242,32 +237,25 @@ var LibraryStackTrace = {
...
-      } else if (match = /^(.+?)@/.exec(frame)) {
-        name = match[1];
-      } else {
-        return 0;
-      }
+    var name;
+    var match;
+    if (match = /^\s+at .*\.wasm\.(.*) \(.*\)$/.exec(frame)) {
+      name = match[1];
+    } else if (match = /^\s+at (.*) \(.*\)$/.exec(frame)) {
```
kripken, line 263: Why is this changing?

test/core/test_return_address.c
```diff
@@ -11,8 +11,10 @@
 
 void *emscripten_return_address(int level);
 const char *emscripten_pc_get_function(void* pc);
+void emscripten_stack_snapshot();
 
 __attribute__((noinline)) void func() {
+  emscripten_stack_snapshot();
```
kripken, line 20: Why did this change?

sbc100: The `emscripten_pc_get_function` depends on `UNWIND_CACHE` being populated to included the address in question.   As it happens, for native frames at least, the old `WASM_OFFSET_CONVERTER` code obviated this need, but for JS frames the `UNWIND_CACHE` always needed to be populated, and the asan etc (the primary / only user of this API, since its not a public API) the cache is normally populated automatically.

I'll add a comment.

src/lib/libstack_trace.js
sbc100: I think we still want the code to run on ancient v8 versions, even if that address -> function name mapping isn't working there.

sbc100: Updated the comment.

ChangeLog.md
```diff
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 4.0.14 (in development)
 -----------------------
+- The wasm offset converter and `WASM_OFFSET_CONVERTER` setting were removed.
+  This features only existed to work around v8 and bug that was long since
```
```suggestion
  This feature only existed to work around a v8 bug that is long since
```

src/lib/libstack_trace.js
```diff
@@ -104,19 +104,14 @@ var LibraryStackTrace = {
...
-#else
 #if ASSERTIONS
-      assert(wasmOffsetConverter, 'wasmOffsetConverter global not found');
-#endif
-      return wasmOffsetConverter.convert(+match[1], +match[2]);
+    } else if (match = /\bwasm-function\[(\d+)\]:(\d+)/.exec(frame)) {
+      // Older versions of v8 (e.g node v10) give function index and offset in
+      // the function.  That format is not supported since it does not provide
+      // the information we need to map the frame to a global program counter.
+      warnOnce('legacy backtrace format detected, this version of v8 is not longer supported by emscripten backtrace mechanism')
```
kripken:
```suggestion
      warnOnce('legacy backtrace format detected, this version of v8 is no longer supported by the emscripten backtrace mechanism')
```

test/test_other.py
```diff
@@ -14388,7 +14367,7 @@ def test_pthread_growth_mainthread(self, cflags, pthread_pool_size):
     else:
       self.cflags.append('-Wno-pthreads-mem-growth')
     self.set_setting('PTHREAD_POOL_SIZE', pthread_pool_size)
-    self.do_runf('pthread/test_pthread_memory_growth_mainthread.c', cflags=['-pthread', '-sALLOW_MEMORY_GROWTH', '-sINITIAL_MEMORY=32MB', '-sMAXIMUM_MEMORY=256MB'] + cflags)
+    self.do_runf('pthread/test_pthread_memory_growth_mainthread.c', cflags=['-Oz', '-pthread', '-sALLOW_MEMORY_GROWTH', '-sINITIAL_MEMORY=32MB', '-sMAXIMUM_MEMORY=256MB'] + cflags)
```
kripken: Why did this change?

src/lib/libstack_trace.js
sbc100: Previously we used two completely different mechanism for getting the functions name:

1. For JS functions to match against the line in the backtrace.
2. For native functions we rely on wasmOffsetConverter.

However as @RReverser pointed out in #24587, there is no need to do that now that the v8 bug has been fixed, and we can use the same approach for both native and JS functions.

test/test_other.py
sbc100: Reverted

src/source_map_support.js
```diff
@@ -61,11 +61,6 @@ class WasmSourceMap {
 
   lookup(offset) {
     var normalized = this.normalizeOffset(offset);
-#if USE_OFFSET_CONVERTER
-    if (!wasmOffsetConverter.isSameFunc(offset, normalized)) {
-      return null;
-    }
-#endif
```
sbc100, line 68: I'm not sure what this bit of code was supposed to be doing.   It looks like its an early exit if the "normalized" is not in the same function as the original offset.

It was added in #8711, but the code also has to work in the absence of `USE_OFFSET_CONVERTER`.. so I'm tempted to say that if the tests pass we should just land this.

src/lib/libstack_trace.js
kripken: Hmm, ok.

I spent a little time trying to figure out what these two lines (249 and 251) accept and reject, and I think I get it now. Perhaps add some comments though? I mean something like saying on the first, "this is the pattern emitted by [..], which looks like [example]" etc.

PR #24967, opened 2025-08-19
Cleanup source map loading

sbc100:
- Move LOAD_SOURCE_MAP to settings_internal.  Its set whenever source maps are used with ASAN and really is only used internally.  We have no tests that set it explicitly.
- Move all code to `source_map_support.js`
- Load `source_map_support.js` during postamble.  This does run under `MINIMAL_RUNTIME`, but `LOAD_SOURCE_MAP` has always been no-op on `MINIMAL_RUNTIME` anyway.


kripken (approved):
lgtm but I would verify this manually in the browser as well - I don't think we have test coverage of browsers successfully loading and using source maps (when they fail, it is silently except for a console log at best)

sbc100:
We do have at least one browser test that would fail without the source map loaded correctly: browser.test_pthread_asan_leak.  I ran it locally to verify.

sbc100:
Ignoring unrelated lsan failures on main

PR #24972, opened 2025-08-19
Improve test_emmalloc_memory_statistics to be more robust to different test modes.

juj:
This way it'll run in lto, thinlto, wasm64 and optimized modes without being so brittle.


test/test_core.py
```diff
@@ -931,22 +931,21 @@ def test_emmalloc_usable_size(self):
...
-      out_suffix = '64'
-    else:
-      out_suffix = ''
-
-    self.set_setting('INITIAL_MEMORY', '128mb')
     self.set_setting('MALLOC', 'emmalloc')
-    self.cflags += ['-g']
-    self.do_core_test('test_emmalloc_memory_statistics.c', out_suffix=out_suffix)
+    self.set_setting('INITIAL_MEMORY', '128MB')
+    output = self.do_run(open(test_file('test_emmalloc_memory_statistics.c')).read())
```
sbc100: This can just be  `output = self.do_runf('test_emmalloc_memory_statistics.c')`

sbc100 (commented):
lgmt % comment

sbc100 (commented):
I guess you know this test better than most, so you can confirm that this more limited checking is not loosing any important information or test coverage?

juj:
> I guess you know this test better than most, so you can confirm that this more limited checking is not loosing any important information or test coverage?

Not really - depending on the exact numbers is too brittle. Verifying that the fields don't come out zero is well enough.

PR #24973, opened 2025-08-19
Fix script.test_fcoverage_mapping

juj:
This is not really the intent of what -sSTRICT was to be about, but this fixes the test at least.

### DIFF
--- a/test/code_size/test_codesize_hello_dylink_all.json
+++ b/test/code_size/test_codesize_hello_dylink_all.json
@@ -1,7 +1,7 @@
 {
   "a.out.js": 246665,
-  "a.out.nodebug.wasm": 597809,
-  "total": 844474,
+  "a.out.nodebug.wasm": 597720,
+  "total": 844385,
   "sent": [
     "IMG_Init",
     "IMG_Load",


test/test_core.py
```diff
@@ -9824,7 +9824,7 @@ def test_fcoverage_mapping(self):
 '''
     self.set_setting('NODERAWFS')
     self.set_setting('EXIT_RUNTIME')
-    self.do_core_test('test_hello_world.c', cflags=['-fprofile-instr-generate', '-fcoverage-mapping', '-g'])
+    self.do_core_test('test_hello_world.c', cflags=['-fprofile-instr-generate', '-fcoverage-mapping', '-g', '-lstubs'])
```
sbc100: It looks like this is due to `uname` reference which is coming from `system/lib/compiler-rt/lib/profile/InstrProfilingUtil.c`.

Can you instead remove `COMPILER_RT_HAS_UNAME` from `tools/system_libs.py` and add `|| defined(__EMSCRIPTEN__)` to line 57 of InstrProfilingPort.h?

juj: Nice, that's a better fix.

sbc100 (approved):

PR #24974, opened 2025-08-19
Make addRunDependency/removeRunDependency optional

sbc100:
This saves on code size is configurations where its not needed. i.e. in `MODULARIZE` mode where we use `await` to wait for instantiation, or if sync instantiation is used.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -1127,10 +1127,12 @@ def phase_linker_setup(options, linker_args):  # noqa: C901, PLR0912, PLR0915
         settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$warnOnce']
 
       settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$getValue', '$setValue']
-      # TODO(sbc): Remove these forced dependencies.
-      settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$runDependencies', '$addRunDependency', '$removeRunDependency']
 
     settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$ExitStatus']
+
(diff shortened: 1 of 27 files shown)


tools/link.py
```diff
@@ -1127,11 +1127,13 @@ def limit_incoming_module_api():
...
 
       settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$getValue', '$setValue']
-      # TODO(sbc): Remove these forced dependencies.
-      settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$runDependencies', '$addRunDependency', '$removeRunDependency']
 
     settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$ExitStatus']
 
+  # Certain configurations require the removeRunDependency/addRunDependency system.
+  if settings.LOAD_SOURCE_MAP or (settings.WASM_ASYNC_COMPILATION and not settings.MODULARIZE):
+    settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE += ['$runDependencies', '$addRunDependency', '$removeRunDependency']
```
kripken: What about `tools/file_packager.py`? Some of the add/removes there are covered by the `src/lib/libfs_shared.js` change in this PR, but that just handles file preloading, and there is a bunch of other stuff there?

sbc100: Good catch.  Added for that case too.

kripken: In file_packager.py I also see LZ4FS uses it, am I reading that wrong?

sbc100: Right, but LZ4FS (and other library files) are able to use the `__deps` system to pull this in.

kripken: Oh, ok, sgtm.

kripken (approved):

PR #24976, opened 2025-08-19
[testing] Use customizable port for the browser test server.

brendandahl:
Add a helper to copy `browser_reporting.js` file with the test server URL inlined. We could do `globalThis.location` for some tests, but that doesn't work for audio worklets and when using blob urls.

Where possible just remove the reference to `http:://localhost:8888` and use `/`.


test/common.py
```diff
@@ -2306,7 +2306,7 @@ class BrowserCore(RunnerCore):
   # suite early, as otherwise we will wait for the timeout on every
   # single test (hundreds of minutes)
   MAX_UNRESPONSIVE_TESTS = 10
-  PORT = 8888
+  PORT = 8889
```
brendandahl: I've changed this to confirm everything is still working. I'll revert before merging.

test/common.py
```diff
@@ -2463,6 +2471,7 @@ def compile_btest(self, filename, cflags, reporting=Reporting.FULL):
     if not os.path.exists(filename):
       filename = test_file(filename)
     self.run_process([compiler_for(filename), filename] + self.get_cflags() + cflags)
+    utils.delete_file('browser_reporting.js')
```
sbc100, line 2475: Maybe just keep this around?

test/common.py
```diff
@@ -2349,6 +2349,8 @@ def setUpClass(cls):
     cls.harness_in_queue = multiprocessing.Queue()
     cls.harness_out_queue = multiprocessing.Queue()
     cls.harness_server = multiprocessing.Process(target=harness_server_func, args=(cls.harness_in_queue, cls.harness_out_queue, cls.PORT))
+    cls.SERVER_URL = f'http://localhost:{cls.PORT}'
+    cls.HARNESS_URL = f'{cls.SERVER_URL}/run_harness'
```
sbc100: Are these two lines needed?   Can we just use the class level declarations above?

sbc100 (approved):
Nice!

test/common.py
brendandahl: There are some tests that have assertions on how many files are in the output dir.

sbc100: Really?  Which ones?  Maybe we should fix those instead.

sbc100: lgtm either way.. but maybe add a comment if you leave this in.

test/common.py
brendandahl: Yeah, can remove this. Will be needed for my later patch where the port is set differently.

PR #24978, opened 2025-08-19
Convert loadDylibs to async. NFC

sbc100:


src/lib/libdylink.js
```diff
@@ -1187,21 +1187,19 @@ var LibraryDylink = {
...
-      ), Promise.resolve())
-      .then(() => {
-        // we got them all, wonderful
-        reportUndefinedSymbols();
-        removeRunDependency('loadDylibs');
-  #if DYLINK_DEBUG
-        dbg('loadDylibs done!');
-  #endif
-      });
+    for (var lib of dynamicLibraries) {
```
RReverser, line 1193: I wonder if this would benefit from Promise.all? Is it safe to load them in parallel?

sbc100: No, we have to load them in the given order I believe.

I thought there was a commend about this, but I don't see it now.

PR #24985, opened 2025-08-20
[file_packager] Convert DataRequest to JS class

sbc100:
- Convert prototype properties to methods and class properties.
- Remove unused `send` method
- Use for .. of to simplify code

I think this whole class can just be removed, but cleaning up first seemed like a good first step.


sbc100:
Closing in favor of #24988

PR #24988, opened 2025-08-20
[file_packager] Remove unnecessary DataRequest class

sbc100:
This class dates back to 2416d19b0 where it looks like this was a standin for an XMLHttpRequest, but that doesn't seem to be needed at all anymore.

### DIFF
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -682,51 +682,25 @@ def generate_js(data_target, data_files, metadata):
     create_preloaded = '''
           try {
             // canOwn this data in the filesystem, it is a slice into the heap that will never change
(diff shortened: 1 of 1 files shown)


tools/file_packager.py
```diff
@@ -1019,7 +995,7 @@ def generate_js(data_target, data_files, metadata):
       }\n''' % {'node_support_code': node_support_code}
 
     code += '''
-      function processPackageData(arrayBuffer) {
+      async function processPackageData(arrayBuffer) {
```
kripken, line 998: Why is this part needed?

sbc100: Prior to this the await code was hidden inside the `async finish` member function.  Now the code runs in the calling scope so that needs to be async.

The `async` was recently added in #24917.

Issue #24905, opened 2025-08-11
test_stat_chmod_rawfs is flaky


sbc100:
Likely some kind of timing issue.

juj:
The exact `assert(s.st_mtime == lastmtime);` assert failure flakiness also appears in e.g. variants

```
thinlto1.test_stat_chmod
lto0.test_stat_chmod_nodefs
wasm2js3.test_stat_chmod_wasmfs
core0.test_stat_chmod_nodefs
core0g.test_stat_chmod_nodefs
```

So not specific to a single mode or a FS backend.

PR #17401, opened 2022-07-08
Use 64 bits for time_t

sbc100:
This brings us back in line with upstream musl.  The change to 32-bit
was only recently made in #16966. The reason we made this change was
made was because we had certain C library calls that were implemented in
JS that returned `time_t`.  Since returning 64-bit values from JS
functions is not always easy (we don't always have WASM_BIGINT
available) that simplest solution was to define `time_t` to 32-bit which
doesn't have issues at the JS boundary.

However, in the intervening time many of the `time_t`-returning function
have been moved into native code (See #16606 and #16439) with only two
remaining: _mktime_js and _timegm_js.  So this change redefines just
those two functions to return `int` while keeping `time_t` itself as
64-bit.

Fixes: #17393

PR #24989, opened 2025-08-20
Possibly fix flaky test seen to error in wasm2js3.test_stat_chmod_wasmfs.

juj:
`time_t` is `long long`, test has a narrowing conversion to `int`.

Fails nondeterministically in http://clbri.com:8010/api/v2/logs/20165/raw_inline

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -5686,7 +5686,7 @@ got: 10
       self.skipTest('mode bits work differently on windows')
     if nodefs and self.get_setting('WASMFS'):
       self.skipTest('test requires symlink creation which currently missing from wasmfs+noderawfs')
-    self.do_runf('stat/test_chmod.c', 'success')
+    self.do_runf('stat/test_chmod.c', 'success', cflags=['-Werror=conversion'])
 
   @also_with_wasmfs
   def test_stat_mknod(self):
--- a/test/stat/test_chmod.c
+++ b/test/stat/test_chmod.c
@@ -22,7 +22,7 @@ void create_file(const char *path, const char *buffer, int mode) {
   int fd = open(path, O_WRONLY | O_CREAT | O_EXCL, mode);
   assert(fd >= 0);
 
-  int err = write(fd, buffer, sizeof(char) * strlen(buffer));
+  int err = (int)write(fd, buffer, sizeof(char) * strlen(buffer));
   assert(err ==  (sizeof(char) * strlen(buffer)));
 
   close(fd);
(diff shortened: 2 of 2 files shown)


sbc100 (commented):
It seems likely this is a timing issue since its rely on sleep(1) and assumes the file created/touched before after the sleep will have different timestamps.

I had thought we had a bug open about this test already but I don't see one.

Perhaps you could open a bug and mark this test with `@flaky(...bug url...)`.. that way it should get re-tried automatically in CI (see EMTEST_RETRY_FLAKY, you probably want to set that to a non-zero value in your CI too).

sbc100:
I found the existing issue : https://github.com/emscripten-core/emscripten/issues/24905

juj:
Could be, hence the "possibly fix."

Nevertheless, long long shouldn't be narrowed to int, so this change is correct. If there's a filesystem that actually has 64-bit timestamps, then the code would be incorrect there. After PR #17401 this test wasn't adapted accordingly.

juj:
> Perhaps you could open a bug and mark this test with `@flaky(...bug url...)`.. that way it should get re-tried automatically in CI (see EMTEST_RETRY_FLAKY, you probably want to set that to a non-zero value in your CI too).

Separate PR.

test/test_core.py
```diff
@@ -5686,7 +5686,7 @@ def test_stat_chmod(self):
       self.skipTest('mode bits work differently on windows')
     if nodefs and self.get_setting('WASMFS'):
       self.skipTest('test requires symlink creation which currently missing from wasmfs+noderawfs')
-    self.do_runf('stat/test_chmod.c', 'success')
+    self.do_runf('stat/test_chmod.c', 'success', cflags=['-Werror=conversion'])
```
sbc100, line 5689: Makes me wonder if we should make an effort to enable this in whole test suite.

We currently enable `-Werror` by default: https://github.com/emscripten-core/emscripten/blob/4cab0399ccb663ccacc3d5b15fa765a8ec651b8e/test/common.py#L1202

But we don't enable `-Wall` (which looks like it doesn't actually include `-Wconversion` anyway )

juj: I was expecting you might suggest that, and did give it a quick test before this PR, and was met with a million `double` -> `float` narrowing warnings with most tests failing miserably. It would be a nice idea, something to work on later.

I also searched the repo for other uses of `.st_mtime` and this was the only occurrence that assigned that to an `int`, other places have been converted already before.

sbc100: Yeah, no worries, I think we are long way from `-Wall` or `-Wextra` by default in the test suite !

PR #24990, opened 2025-08-20
Skip @also_with_modularize runs in strict_js suite (fixes e.g. strict_js.test_pthread_proxying_modularize)

juj:


test/common.py
```diff
@@ -683,6 +683,8 @@ def also_with_modularize(f):
   @wraps(f)
   def metafunc(self, modularize, *args, **kwargs):
     if modularize:
+      if self.get_setting('STRICT_JS'):
+        self.skipTest('MODULARIZE is not compatible with STRICT_JS')
```
sbc100, line 687: Maybe match the wording below?  Either option seems fine but they should be consistent.

juj: I followed your previous review in https://github.com/emscripten-core/emscripten/pull/24958#discussion_r2283464888 for the message, so it's at least consistent with something.

sbc100: Can you update line 689 below to match?

juj: sure

PR #25000, opened 2025-08-21
Fix wasm2js0 memory growth tests

juj:
Fix wasm2js.test_memorygrowth, wasm2js.test_memorygrowth_2, wasm2js0.test_minimal_runtime_memorygrowth


test/test_core.py
```diff
@@ -2095,7 +2095,11 @@ def test_minimal_runtime_memorygrowth(self):
     self.do_runf(src, 'OOM', assert_returncode=NON_ZERO)
     # Win with it
     self.set_setting('ALLOW_MEMORY_GROWTH')
-    self.do_runf(src, '*pre: hello,4.955*\n*hello,4.955*\n*hello,4.955*')
+    if self.is_wasm2js() and '-O0' in self.cflags:
+      expect = '*pre: hello,4.955*\nWarning: Enlarging memory arrays, this is not fast! 16908288,20316160\n*hello,4.955*\n*hello,4.955*'
+    else:
```
kripken: Perhaps it would be simpler to add a helper function `self.remove_growth_warning(output)` which would remove the warning, and call it from these places?

juj: Hmm, the output is handled internally inside `self.do_runf(src, expect)`, the test code here doesn't get access to it, but only feeds the expectation. So the function `self.do_runf(src, expect)` would have to be broken apart for that.

Or did you mean that the `remove_growth_warning()` function would be called on these expectation strings?

kripken: Ah, I thought `do_runf` had a way to filter strings. We used to, but I guess it was removed ("nicerizer function"). Though there is a `regex` param, but using it on these strings could be annoying...

My concern is that the hardcoded `16908288,20316160` figures will get out of date and need annoying updates, so I think it is worth finding a way to avoid that.

One way might be to avoid using `do_runf` here, and call `self.build` then run and compare manually.

sbc100: If you want to do the assertion checking outside of `do_runf` i think you can just grab the output from the return value:

```
output = self.do_runf(...)
output = .. filter output ..
```

It not has nice a doing it in single `do_runf` call, but it should work.

Alternatively we could find a way to split the `Warning` output (stderr) from the actual test output (stdout).  Perhaps a stdout_only, param or something like that?

juj: > i think you can just

Thanks, super useful. Didn't realize that the test would skip asserting and return output like this. Refactored to use this format, how does this look?

test/test_core.py
```diff
@@ -2095,7 +2100,10 @@ def test_minimal_runtime_memorygrowth(self):
     self.do_runf(src, 'OOM', assert_returncode=NON_ZERO)
     # Win with it
     self.set_setting('ALLOW_MEMORY_GROWTH')
-    self.do_runf(src, '*pre: hello,4.955*\n*hello,4.955*\n*hello,4.955*')
+    expect = '*pre: hello,4.955*\n*hello,4.955*\n*hello,4.955*'
+    output = self.do_runf(src)
+    output = self.remove_growth_warning(output)
+    self.assertContained(expect, output)
```
sbc100: nit: Maybe just inline the `expect` here and below?  

lgtm either way.

juj: ok

kripken (approved):

PR #25002, opened 2025-08-21
Skip simd2.test_autodebug_wasm

juj:
CC @kripken is SIMD+EMCC_AUTODEBUG combination expected/intended to work? Or is this skip good?


juj:
Kk, that makes sense.

PR #25004, opened 2025-08-21
Fix FS forceLoadFile() function

juj:
Fixes `strict_js.test_files` and other `strict_js` suite runs.

### DIFF
--- a/test/code_size/test_codesize_file_preload.json
+++ b/test/code_size/test_codesize_file_preload.json
@@ -1,10 +1,10 @@
 {
-  "a.out.js": 22699,
-  "a.out.js.gz": 9379,
+  "a.out.js": 22684,
+  "a.out.js.gz": 9375,
   "a.out.nodebug.wasm": 1733,
   "a.out.nodebug.wasm.gz": 980,
-  "total": 24432,
-  "total_gz": 10359,
+  "total": 24417,
+  "total_gz": 10355,
   "sent": [
     "a (fd_write)"
   ],
--- a/test/other/codesize/test_codesize_file_preload.expected.js
+++ b/test/other/codesize/test_codesize_file_preload.expected.js
@@ -2804,7 +2804,6 @@ var FS = {
       // Command-line.
       try {
         obj.contents = readBinary(obj.url);
-        obj.usedBytes = obj.contents.length;
       } catch (e) {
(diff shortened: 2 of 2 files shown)


sbc100 (commented):
What is the `strict_js` error that this fixes?

Is the current code actually broken?  How to the tests pass without this change?

juj:
Yes the code was broken before, it assigns to the field without a setter. In lazy JS assigning to a field without a setter is silently ignored. In strict_js, it throws a no-assignment-allowed error.

src/lib/libfs.js
```diff
@@ -1843,7 +1846,15 @@ FS.staticInit();`;
...
         usedBytes: {
-          get: function() { return this.contents.length; }
+          get: function() { return this.contents.length; },
+          set: function(newLength) {
+            // Resize the buffer by allocating new.
+            if (newLength != this.contents.length) {
+              var newBuffer = new Uint8Array(newLength);
+              newBuffer.set(this.contents);
+              this.contents = newBuffer;
+            }
```
sbc100: This seems like it does more than just fix the "cannot assign to `usedBytes` in strict mode" though.    it also changes what the assignment does, even in non-strict mode. 

In non-strict mode what does the assignment do?  Is it a no-op?

juj: Yes, it is a no-op.

```js
//'use strict';

var node = {
  contents: new Uint8Array(8),
};
Object.defineProperties(node, {
  usedBytes: {
    get: function() { return this.contents.length; }
  }
});

console.log(node.usedBytes);
node.usedBytes = 16;
console.log(node.usedBytes);
```
prints
```
8
8
```
in non-strict mode, but if you comment out use strict, then `node.usedBytes = 16;` throws a JS exception.

juj: This is the line that throws:

https://github.com/emscripten-core/emscripten/blob/5788dcc32540f2f3d13e6e33f35ac01188b7917d/src/lib/libfs.js#L1713

I am not sure how else to interpret the intent here.

juj: Or actually, the more I read into this, it does seem like the assignment to `.usedBytes` should be redundant altogether.

sbc100 (approved):
Nice!

PR #25007, opened 2025-08-21
Fix llvmlibc.test_asan*

juj:


tools/system_libs.py
```diff
@@ -1006,7 +1006,7 @@ class llvmlibc(DebugLibrary, AsanInstrumentedLibrary, MTLibrary):
   def get_files(self):
     files = glob_in_path('system/lib/llvm-libc/src/assert', '*.cpp')
     files += glob_in_path('system/lib/llvm-libc/src/complex', '**/*.cpp')
-    files += glob_in_path('system/lib/llvm-libc/src/string', '**/*.cpp')
+    files += glob_in_path('system/lib/llvm-libc/src/string', '**/*.cpp', excludes=['memset.cpp', 'memcpy.cpp'] if self.is_asan else [])
```
sbc100, line 1009: The llvm libc test config is experimental, and many things are known to not work yet, including asan.

@google-yfyang is working it on it.

I'm not sure if this is the way we want to do asan integration here.. maybe?  maybe not.

juj: The asan tests failed on duplicate symbols, and after this change they passed. We could land this and I could run this on my CI.

Or we can say no to free coverage and not land. Whichever works.

sbc100: I think we are too early in the llvm-libc experiment (it really is very early days) for it to be useful to have you testing it (or reporting errors like this).

@google-yfyang may disagree though?

google-yfyang: I don't think I mind either. As Sam mentioned it doesn't really make sense to run these tests yet given that the experimental overlay mode is incompatible with asan tests. When we eventually exit out of this experimental overlay mode, we should get these tests failures resolved for free. 

On the other hand, if we plan to have real users use overlay mode in some capacity, it might make sense to have asan tests running?

juj:
Well, I ran the whole llvmlibc suite, and there were only these few errors. Feels odd if you would frown on getting free work for test fixes.

sbc100:
> Well, I ran the whole llvmlibc suite, and there were only these few errors. Feels odd if you would frown on getting free work for test fixes.

Its just that llvmlibc configuration is very experimental, and known to not work in all cases yet.  For example, its known to not work with asan.  While fixing it is great in this case (assuming this is the fix we want for asan?)  please be aware that this suite is not expected to fully pass at all times, so I would advise against trying to run it an expecting it to fully pass.

lgtm to this change if @google-yfyang is ok with it.  I just wanted you to be aware of the current expeirmental nature of this test config.

juj:
I got started on this CI work since trying to land anything to Emscripten has been such a pita for a long time, because everything is always red and flaky, and trying to find a big picture is hard.

I miss the days when we could land incremental improvements to the right direction with ease. Even if this PR isn't the ultimate fix, it still makes the test suite look more green than what it was before, so I think it would be a nudge in the right direction, and help see a big picture.

sbc100:
I'm not saying we can't land this, I just want to make sure you understand that not all the test configuration in the core test suite are expected to pass fully at any given time.  

Perhaps we should have document that better so that you don't need to spend your time debugging stuff like this (even if the result, this PR, is very useful, and welcome, its likely not a good use of your time I fear).

juj:
I understand this. The reason I authored this is that the fix looked trivial.

If it didn't, I wouldn't have bothered with it.

I can definitely skip running `llvmlibc` suite on our CI and ignore its results as well. I just thought to contribute here and on #25008 with visibility since I know the full `llvmlibc` suite is not running regularly on CircleCI otherwise. Though if that is superfluous noise and not interesting, I can definitely skip doing such "drive-by" contributions. Just thought to be a "good citizen."

sbc100:
> I understand this. The reason I authored this is that the fix looked trivial.
> 
> If it didn't, I wouldn't have bothered with it.
> 
> I can definitely skip running `llvmlibc` suite on our CI and ignore its results as well. I just thought to contribute here and on #25008 with visibility since I know the full `llvmlibc` suite is not running regularly on CircleCI otherwise. Though if that is superfluous noise and not interesting, I can definitely skip doing such "drive-by" contributions. Just thought to be a "good citizen."

Ok, sorry I misunderstood.  I didn't want to you wasting your time on configs that are known to be failing.

This change lgtm!

PR #25013, opened 2025-08-21
[testing] Improve running browser tests locally.

brendandahl:
This changes how the browser test runner works by default. It will now automatically detect the browser type and:
 - Use the same options used in CircleCI (for chrome and firefox)
 - Create and delete a temporary profile directory
 - Terminate the browser
 
Adds two new options:
 -  `EMTEST_BROWSER_AUTO_CONFIG` or `--browser-auto-config`: disables the automatic configuration,
 - `EMTEST_HEADLESS` or `--headless`: run the test in headless mode

Last, two flags were removed from chrome:
- `--disk-cache-dir` this was causing errors (on CI and locally)
- `--incognito` not allowed in corporate environments and not needed since the profile directory is automatically deleted.

These changes make it much easier to reproduce the circle ci environment locally and make running a test multiple times more reliable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -325,7 +325,7 @@ commands:
             # bypass the user gesture so audio tests work without interaction)
             CHROME_FLAGS_AUDIO: " --use-fake-device-for-media-stream --autoplay-policy=no-user-gesture-required"
           command: |
-            export EMTEST_BROWSER="/usr/bin/google-chrome $CHROME_FLAGS_BASE $CHROME_FLAGS_HEADLESS $CHROME_FLAGS_WASM $CHROME_FLAGS_NOCACHE $CHROME_FLAGS_AUDIO"
+            export EMTEST_BROWSER="/usr/bin/google-chrome"
             # There are tests in the browser test suite that using libraries
             # that are not included by "./embuilder build ALL".  For example the
             # PIC version of libSDL which is used by test_sdl2_misc_main_module
@@ -400,7 +400,7 @@ commands:
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
             DISPLAY: ":0"
           command: |
-            export EMTEST_BROWSER="$HOME/firefox/firefox -headless -profile $HOME/tmp-firefox-profile/"
+            export EMTEST_BROWSER="$HOME/firefox/firefox"
             # There are tests in the browser test suite that using libraries
             # that are not included by "./embuilder build ALL".  For example the
             # PIC version of libSDL which is used by test_sdl2_misc_main_module
@@ -423,14 +423,8 @@ commands:
           environment:
             EMTEST_LACKS_SOUND_HARDWARE: "1"
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
-            # --no-sandbox becasue we are running as root and chrome requires
-            # this flag for now: https://crbug.com/638180
-            CHROME_FLAGS_BASE: "--no-first-run -start-maximized --no-sandbox --use-gl=swiftshader --user-data-dir=/tmp/chrome-emscripten-profile"
-            CHROME_FLAGS_HEADLESS: "--headless=new --remote-debugging-port=1234"
-            CHROME_FLAGS_WASM: "--enable-experimental-webassembly-features"
-            CHROME_FLAGS_NOCACHE: "--disk-cache-dir=/dev/null --disk-cache-size=1 --media-cache-size=1 --disable-application-cache --incognito"
           command: |
-            export EMTEST_BROWSER="/usr/bin/google-chrome $CHROME_FLAGS_BASE $CHROME_FLAGS_HEADLESS $CHROME_FLAGS_WASM $CHROME_FLAGS_NOCACHE"
+            export EMTEST_BROWSER="/usr/bin/google-chrome"
             test/runner sockets
       - upload-test-results
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -378,9 +378,9 @@ commands:
             EMTEST_LACKS_OFFSCREEN_CANVAS: "1"
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
             EMTEST_HEADLESS: "1"
-            EMTEST_BROWSER: "$HOME/firefox/firefox"
             DISPLAY: ":0"
           command: |
+            export EMTEST_BROWSER="$HOME/firefox/firefox"
             # There are tests in the browser test suite that using libraries
             # that are not included by "./embuilder build ALL".  For example the
             # PIC version of libSDL which is used by test_sdl2_misc_main_module
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -315,8 +315,8 @@ commands:
           environment:
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
             EMTEST_HEADLESS: "1"
+            EMTEST_BROWSER: "/usr/bin/google-chrome"
           command: |
-            export EMTEST_BROWSER="/usr/bin/google-chrome"
             # There are tests in the browser test suite that using libraries
             # that are not included by "./embuilder build ALL".  For example the
             # PIC version of libSDL which is used by test_sdl2_misc_main_module
@@ -378,9 +378,9 @@ commands:
             EMTEST_LACKS_OFFSCREEN_CANVAS: "1"
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
             EMTEST_HEADLESS: "1"
+            EMTEST_BROWSER: "$HOME/firefox/firefox"
             DISPLAY: ":0"
           command: |
-            export EMTEST_BROWSER="$HOME/firefox/firefox"
             # There are tests in the browser test suite that using libraries
             # that are not included by "./embuilder build ALL".  For example the
             # PIC version of libSDL which is used by test_sdl2_misc_main_module
@@ -404,8 +404,8 @@ commands:
             EMTEST_LACKS_SOUND_HARDWARE: "1"
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
             EMTEST_HEADLESS: "1"
+            EMTEST_BROWSER: "/usr/bin/google-chrome"
           command: |
-            export EMTEST_BROWSER="/usr/bin/google-chrome"
             test/runner sockets
       - upload-test-results
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -314,16 +314,7 @@ commands:
           name: run tests (<< parameters.title >>)
           environment:
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
-            # --no-sandbox because we are running as root and chrome requires
-            # this flag for now: https://crbug.com/638180
-            CHROME_FLAGS_BASE: "--no-first-run -start-maximized --no-sandbox --enable-unsafe-swiftshader --use-gl=swiftshader  --user-data-dir=/tmp/chrome-emscripten-profile --enable-experimental-web-platform-features --enable-features=JavaScriptSourcePhaseImports"
-            # Increase the window size to avoid flaky sdl tests see #24236.
-            CHROME_FLAGS_HEADLESS: "--headless=new --window-size=1024,768 --remote-debugging-port=1234"
-            CHROME_FLAGS_WASM: "--enable-experimental-webassembly-features --js-flags=\"--experimental-wasm-stack-switching --experimental-wasm-type-reflection --experimental-wasm-rab-integration\""
-            CHROME_FLAGS_NOCACHE: "--disk-cache-dir=/dev/null --disk-cache-size=1 --media-cache-size=1 --disable-application-cache --incognito"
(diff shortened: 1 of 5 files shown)


test/common.py
```diff
@@ -45,15 +45,20 @@
 
 # User can specify an environment variable EMTEST_BROWSER to force the browser
 # test suite to run using another browser command line than the default system
-# browser.
-# There are two special value that can be used here if running in an actual
+# browser. If only the path to the browser executable is given, the tests
+# will run in headless mode with a temporary profile with the same options
+# used in CircleCI. To use a custom start command specify the executable and
```
kripken:
```suggestion
# used in CI. To use a custom start command specify the executable and
```
(future-proofing)

test/runner.py
```diff
@@ -392,6 +392,8 @@ def parse_args():
                       help='Automatically update test expectations for tests that support it.')
   parser.add_argument('--browser',
                       help='Command to launch web browser in which to run browser tests.')
+  parser.add_argument('--headed', action='store_true',
+                      help='Run browser tests in regular (non-headless) mode.')
```
kripken: Does this have a default of None, i.e. not headed, i.e. headless? (the reverse of the current behavior)

test/common.py
```diff
@@ -76,6 +81,38 @@
...
+    'default': {},
+    'headless': '-headless',
+  },
+  'other': {
+    'data_dir_flag': '',
+    'default': {},
+    'headless': '',
+  },
+}
+DEFAULT_BROWSER_DATA_DIR = '/tmp/emscripten-profile'
```
sbc100: I think we should using `out/browser-profile` (or at least something in the out directory)

test/common.py
```diff
@@ -2335,6 +2370,15 @@ def browser_restart(cls):
       logger.info('Browser did not respond to `terminate`.  Using `kill`')
       cls.browser_proc.kill()
       cls.browser_proc.wait()
+    if cls.browser_data_dir:
+      utils.delete_dir(cls.browser_data_dir)
+      cls.browser_data_dir = None
```
sbc100, line 2387: It might be useful for debugging to leave this around after terminate, and instead clean it up on startup.

test/runner.py
sbc100: Should we keep the default as headed for now to avoid breakage and have CI use `--headless`?

brendandahl: New default is headless. I did default for two reasons: 1) in preparation for parallel mode, where having many chrome windows open at once is undesirable. 2) most browser test frameworks run headless by default these days.

Another option, we could have different defaults for parallel test vs single test. Or only allow headless for parallel.

test/common.py
brendandahl: This is the current default on ci. Doesn't the out directory get cleared out automatically?

juj: `/tmp/emscripten-profile` reads like a linux-ism?

Maybe Python's `tempfile.mkdtemp(prefix='emscripten-profile-dir')`?

test/common.py
```diff
@@ -2343,6 +2387,26 @@ def browser_open(cls, url):
     if not EMTEST_BROWSER:
       logger.info('No EMTEST_BROWSER set. Defaulting to `google-chrome`')
       EMTEST_BROWSER = 'google-chrome'
+
+    # If only the the browser is specified, use the default arguments used in circleci.
+    browser_args = shlex.split(EMTEST_BROWSER)
```
juj: I think this reproduces the issue with `EMTEST_BROWSER` from before, where on Windows developers would natively do

```
set EMTEST_BROWSER=c:\path\to\browser.exe
```
to set an environment variable to point to an executable, that would not parse through `shlex.split()`.

See https://github.com/emscripten-core/emscripten/pull/24962/files on how to make the parsing work out natively on both Windows and Linux/macOS.

test/common.py
sbc100: Its `out/test` that is specifically clears for each test (doesn't apply to parallel mode).

The `out` directory itself can be used for any other stuff like this.   Think of it as the scratch space for emscripten development.  In that sense I think putting the profiles here makes perfect sense (and its what I do on my machine).

test/runner.py
juj: `--headed` is an odd command line parameter. I wonder if `--windowed` or `--visible` might work better?

Though I'd be slightly more leaning in favor of `--headless` being the option, and visible window mode being the default.

sbc100: How about `--headless=0` AKA `--no-headless`.

Then we can use `default=true` if we want it to be the default...

test/common.py
```diff
@@ -2343,6 +2387,26 @@ def browser_open(cls, url):
...
+    if len(browser_args) == 1:
+      logger.info('No EMTEST_BROWSER flags set. Defaulting to CI configuration.')
```
juj: What should a user pass here to not default to CI configuration, if they don't have any flags to pass?

test/common.py
juj: > The `out` directory itself

Btw, just to note that writing any files under Emscripten (like the `out/` directory) is strictly forbidden except for testing harness purposes. This is because on users' systems, Emscripten is installed under `C:\Program Files`, which is read-only application directory.

sbc100: Indeed, its for emscripten development and testing only.

test/common.py
brendandahl: added a new option `--browser-auto-config` to control this explicity

brendandahl:
I still need to fix the profile dir and I also totally missed that for firefox we create a profile file with some settings. I need to add that in here...

sbc100:
> I still need to fix the profile dir and I also totally missed that for firefox we create a profile file with some settings. I need to add that in here...

Maybe we can find equivalent command line flags?

brendandahl:
> > I still need to fix the profile dir and I also totally missed that for firefox we create a profile file with some settings. I need to add that in here...
> 
> Maybe we can find equivalent command line flags?

There weren't equivalents a few years ago, and from a quick search I don't see anything has changed.

test/runner.py
brendandahl: It seems people are leaning towards "headed" by default, so I changed it back. The env var `EMTEST_HEADLESS` or `--headless` make it easy enough to turn it on.

brendandahl:
To summarize changes:
- default to windowed mode (add --headless to do the opposite)
- setup firefox profile js
- added flag to skip auto profile setup
- put profile dir in out/ dir

test/common.py
```diff
@@ -2343,6 +2414,25 @@ def browser_open(cls, url):
     if not EMTEST_BROWSER:
       logger.info('No EMTEST_BROWSER set. Defaulting to `google-chrome`')
       EMTEST_BROWSER = 'google-chrome'
+
+    if EMTEST_BROWSER_AUTO_CONFIG:
+      logger.info('Using default to CI configuration.')
```
kripken:
```suggestion
      logger.info('Using default CI configuration.')
```

kripken (approved):

.circleci/config.yml
```diff
@@ -314,18 +314,9 @@ commands:
...
-            CHROME_FLAGS_HEADLESS: "--headless=new --window-size=1024,768 --remote-debugging-port=1234"
-            CHROME_FLAGS_WASM: "--enable-experimental-webassembly-features --js-flags=\"--experimental-wasm-stack-switching --experimental-wasm-type-reflection --experimental-wasm-rab-integration\""
-            CHROME_FLAGS_NOCACHE: "--disk-cache-dir=/dev/null --disk-cache-size=1 --media-cache-size=1 --disable-application-cache --incognito"
-            # The runners lack sound hardware so fallback to a dummy device (and
-            # bypass the user gesture so audio tests work without interaction)
-            CHROME_FLAGS_AUDIO: " --use-fake-device-for-media-stream --autoplay-policy=no-user-gesture-required"
+            EMTEST_HEADLESS: "1"
           command: |
-            export EMTEST_BROWSER="/usr/bin/google-chrome $CHROME_FLAGS_BASE $CHROME_FLAGS_HEADLESS $CHROME_FLAGS_WASM $CHROME_FLAGS_NOCACHE $CHROME_FLAGS_AUDIO"
+            export EMTEST_BROWSER="/usr/bin/google-chrome"
```
sbc100: Can you move this up to the `environment` section now?

.circleci/config.yml
```diff
@@ -398,9 +377,10 @@ commands:
             # OffscreenCanvas support is not yet done in Firefox.
             EMTEST_LACKS_OFFSCREEN_CANVAS: "1"
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
+            EMTEST_HEADLESS: "1"
             DISPLAY: ":0"
           command: |
-            export EMTEST_BROWSER="$HOME/firefox/firefox -headless -profile $HOME/tmp-firefox-profile/"
+            export EMTEST_BROWSER="$HOME/firefox/firefox"
```
sbc100, line 383: Ditto

.circleci/config.yml
```diff
@@ -423,14 +403,9 @@ commands:
...
-            # --no-sandbox becasue we are running as root and chrome requires
-            # this flag for now: https://crbug.com/638180
-            CHROME_FLAGS_BASE: "--no-first-run -start-maximized --no-sandbox --use-gl=swiftshader --user-data-dir=/tmp/chrome-emscripten-profile"
-            CHROME_FLAGS_HEADLESS: "--headless=new --remote-debugging-port=1234"
-            CHROME_FLAGS_WASM: "--enable-experimental-webassembly-features"
-            CHROME_FLAGS_NOCACHE: "--disk-cache-dir=/dev/null --disk-cache-size=1 --media-cache-size=1 --disable-application-cache --incognito"
+            EMTEST_HEADLESS: "1"
           command: |
-            export EMTEST_BROWSER="/usr/bin/google-chrome $CHROME_FLAGS_BASE $CHROME_FLAGS_HEADLESS $CHROME_FLAGS_WASM $CHROME_FLAGS_NOCACHE"
+            export EMTEST_BROWSER="/usr/bin/google-chrome"
```
sbc100: Ditto

test/common.py
```diff
@@ -76,6 +82,53 @@
 if 'EM_BUILD_VERBOSE' in os.environ:
   exit_with_error('EM_BUILD_VERBOSE has been renamed to EMTEST_BUILD_VERBOSE')
 
+
+# Default flags used to run browsers in CI testing:
+class BrowserConfig:
+  def __init__(self, data_dir_flag, default_flags, headless_flags):
+      self.data_dir_flag = data_dir_flag
```
sbc100: Indentation looks wrong here.

test/common.py
```diff
@@ -76,6 +82,53 @@
...
+    )
+
+
+class FirefoxConfig(BrowserConfig):
+  def __init__(self):
+    super().__init__(
+      data_dir_flag = '-profile ',
+      default_flags = {},
+      headless_flags = '-headless',
+    )
```
sbc100: Can these just be static data members:

```
class FirefoxConfig(BrowserConfig):
  data_dir_flag = '-profile '
  default_flags = {}
  headless_flags = '-headless'
```

test/common.py
```diff
@@ -2343,6 +2401,22 @@ def browser_open(cls, url):
...
+        utils.delete_dir(cls.browser_data_dir)
+      os.mkdir(cls.browser_data_dir)
+      if is_chrome():
+        config = ChromeConfig()
+      elif is_firefox():
+        config = FirefoxConfig()
+      EMTEST_BROWSER += f" {config.data_dir_flag}{cls.browser_data_dir} {' '.join(config.default_flags)}"
+      if EMTEST_HEADLESS == 1:
+        EMTEST_BROWSER += f" {config.headless_flags}"
+      config.configure(cls.browser_data_dir)
```
sbc100, line 2418: Maybe "else error" here, "EMTEST_BROWSER_AUTO_CONFIG only currently works with firefox or chrome"

test/common.py
```diff
@@ -76,6 +82,40 @@
...
+  data_dir_flag = '-profile '
+  default_flags = ()
+  headless_flags = '-headless'
+
+  @staticmethod
+  def configure(data_dir):
+    shutil.copy(test_file('firefox_user.js'), os.path.join(data_dir, 'user.js'))
+
+
+DEFAULT_BROWSER_DATA_DIR = path_from_root('out/browser-profile')
```
sbc100: Move this down to be alongside `LAST_TEST`?

kainino0x:
How am I meant to turn off auto-config? I don't think there's any way to set `EMTEST_BROWSER_AUTO_CONFIG` that will disable it (e.g. `EMTEST_BROWSER_AUTO_CONFIG=0` is truthy), the only thing I could find is `--browser-auto-config=` or `--browser-auto-config ''` which is very strange.

And then even doing that, `if EMTEST_BROWSER_AUTO_CONFIG:` doesn't work because `EMTEST_BROWSER_AUTO_CONFIG` is `'0'` which is still truthy.

(What I actually need to do is run WebGPU tests in Chrome, which auto config disables by using swiftshader. Fortunately I can deal with this for now by adding `--enable-unsafe-webgpu` which enables the unsafe webgpu-on-swiftshader.)

brendandahl:
That went through a few iterations and I must have broke something. Fix in #25384

Issue #25015, opened 2025-08-21
(thin)lto0.test_emscripten_get_compiler_setting and other failures


juj:
The following tests are failing

```
lto0.test_emscripten_get_compiler_setting
lto0.test_pthread_run_script
lto0.test_return_address

thinlto0.test_emscripten_get_compiler_setting
thinlto0.test_pthread_run_script
thinlto0.test_return_address
```

with

```
wasm-ld: error: libc-debug.a(stderr.o): attempt to add bitcode file after LTO (stderr)
wasm-ld: error: libc-debug.a(fprintf.o): attempt to add bitcode file after LTO (fprintf)
```

See more at https://github.com/emscripten-core/emscripten/pull/24999

PR #25016, opened 2025-08-21
Skip failing lto0 and thinlto0 tests.

juj:
Skip failing lto0 and thinlto0 tests. https://github.com/emscripten-core/emscripten/issues/25015

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -1907,7 +1907,7 @@ int main() {
     self.do_runf('test_emscripten_get_now.c', 'Timer resolution is good')
 
   def test_emscripten_get_compiler_setting(self):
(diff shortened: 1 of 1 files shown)


test/test_core.py
```diff
@@ -2721,6 +2724,9 @@ def test_pthread_tls(self):
 
   @no_modularize_instance('uses global Module objecgt')
   def test_pthread_run_script(self):
+    if '-O0' in self.cflags and ('-flto' in self.cflags or '-flto=thin' in self.cflags):
```
sbc100: How about `if not self.is_optimizing()` (to catch the case where no `-O` flag is passed)

sbc100 (approved):

test/test_core.py
juj: Great idea

PR #25017, opened 2025-08-21
Fix strict.test_unistd_misc*

juj:


test/test_core.py
```diff
@@ -6181,6 +6181,8 @@ def test_unistd_io(self):
   @no_windows('https://github.com/emscripten-core/emscripten/issues/8882')
   @also_with_nodefs
   def test_unistd_misc(self):
+    if self.get_setting('STRICT'):
+      self.cflags += ['-lstubs']
```
sbc100: Actually maybe better to add `-sALLOW_UNIMPLEMENTED_SYSCALLS` explicitly (which is the setting which STRICT is adding that makes this test break (I think).

Alternatively you could just skip in strict mode with "test is not compatible with strict mode (uses unimplemented syscalls)"?

PR #25019, opened 2025-08-22
Split up test other.test_standalone_system_headers to parallelize test runtime.

juj:
Split up test other.test_standalone_system_headers to parallelize test runtime. It is still a slow test (> minute runtime), but much faster than before.

Before: `Ran 1 test in 272.024s`

After:
`Total core time: 386.746s. Wallclock time: 64.692s. Parallelization: 5.98x.`

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -10069,7 +10069,23 @@ int main() {
   # Tests that Emscripten-provided header files can be cleanly included standalone.
   # Also check they can be included in C code (where possible).
   @is_slow_test
-  def test_standalone_system_headers(self):
+  @parameterized({
+    'ab': ('ab',),
+    'cd': ('cd',),
+    'ef': ('ef',),
+    'gh': ('gh',),
+    'ij': ('ij',),
+    'kl': ('kl',),
+    'mn': ('mn',),
+    'op': ('op',),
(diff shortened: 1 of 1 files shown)


test/test_other.py
```diff
@@ -10088,10 +10104,15 @@ def test_standalone_system_headers(self):
...
 
     for directory, headers in directories.items():
-      print('dir: ' + directory)
       for header in headers:
         if not header.endswith('.h'):
           continue
+        # Process files depending on first char of the file in different test cases for parallelization, though
+        # special case SDL_ prefix to parallelize better.
+        first_char = (header[4] if header.startswith('SDL_') else header[0]).lower()
+        if first_char not in prefix or (prefix == '*' and ord(first_char) >= ord('a') and ord(first_char) <= ord('z')):
```
sbc100: `(prefix == '*' and firstchar.isalpha())`

juj: That wouldn't be strictly accurate, since the set of characters that return true on `.isalpha()` is much more than a-z. Though updated to smooth review along.

juj:
> Should we now remove `@is_slow_test`?

No, the test is still unfortunately slow, just not quite as slow as before.

PR #25024, opened 2025-08-22
[AUDIO_WORKLET] Add TLS variable test NFC

cwoffenden:
As per @juj's suggestion, this extends one of the audio playback tests to also verify the JS set-up code doesn't stomp on the TLS section. This mixer test is idea since the param data can grow and shrink during playback.


juj (approved):
Thanks - looks good 👍

PR #25025, opened 2025-08-22
Add debugging prints to test_stat_chmod

juj:
I locally ran `test/runner --repeat 100 --failfast *.test_stat_chmod*` on my Linux workstation, and could not coax out a single failure.

Nevertheless, on that very same workstation, the test maybe about 10%-20% of the time when run as part of the CI.

So add debug prints to help suggest if the issue is about close-but-not-quite timestamps, or something else.

Related to #24905.


juj:
Debugged this on my local CI, not needed anymore.

PR #25029, opened 2025-08-23
Fix a race condition in test_chmod

### LINKED ISSUE #24905: test_stat_chmod_rawfs is flaky

juj:
Fix logic error in test_chmod that resulted in a race condition. The test would compare file metadata modification times of `file` vs `otherfile` to each other, when the intent was to test modification time of `otherfile` before vs after its `fchmodat()` operation.

Fixes #24905.

### DIFF
--- a/test/stat/test_chmod.c
+++ b/test/stat/test_chmod.c
@@ -49,7 +49,8 @@ void test() {
   struct stat s;
 
   //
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
Nice find!

PR #25039, opened 2025-08-24
Fix wasmfs.test_fs_nodefs_readdir on Windows.

juj:
Windows does not support fifos, so the test expectation needs to look different.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -5868,7 +5868,7 @@ got: 10
     self.cflags += ['-lnodefs.js']
     suffix = ''
     if self.get_setting('WASMFS'):
-      suffix = '.wasmfs'
+      suffix = '.wasmfs_win' if WINDOWS else '.wasmfs'
     elif self.is_wasm2js():
       suffix = ".wasm2js"
     self.do_run_in_out_file_test('fs/test_nodefs_readdir.c', out_suffix=suffix)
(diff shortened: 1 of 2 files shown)


test/test_core.py
```diff
@@ -5868,7 +5868,7 @@ def test_fs_nodefs_readdir(self):
     self.cflags += ['-lnodefs.js']
     suffix = ''
     if self.get_setting('WASMFS'):
-      suffix = '.wasmfs'
+      suffix = '.wasmfs_win' if WINDOWS else '.wasmfs'
```
sbc100, line 5872: IIUC the issue here is the order in which directories are listed? 

I wonder if we can instead have either:

1. Have wasmfs list directories in a deterministic / alphabetical order
2. Update the test to avoid depending on specific ordering.

Adding another separate copy of the expectations seems less desirable.

juj: This is not an ordering problem, but a problem that Windows does not support FIFO objects on the filesystem, so the test is unable to add one. And hence fails to `readdir()` it in the filesystem on Windows.

juj:
Ping, any further thoughts here?

sbc100:
In general, I don't love the solution of having get more different output expectation file, since it makes it hard to see how/why the different configs have different output. 

But I don't see any other great options here.

test/test_core.py
sbc100: Perhaps add a comment as to why these three different output are needed?

sbc100 (approved):

test/test_core.py
juj: check

PR #25040, opened 2025-08-25
Weak link confstr

hoodmane:
I think it would be useful if this did not say that it Emscripten is "glibc 2.14". If it's weak linked, I can easily replace the function and make it say something else.


sbc100:
Should we also try to just remove the emscripten patch here?

Also, why does it need to be weak linked?  Can't you just provide your own implementation and the normal rules of `.a` file linking will mean that your definition is chosen?

Issue #15014, opened 2021-09-09
test_pthread_abort is failing on macOS & flaky on linux


sbc100:
Its not clear what changed but this test recently started failing more consistently:

https://chromium-review.googlesource.com/c/emscripten-releases/+/3151159
https://ci.chromium.org/ui/p/emscripten-releases/builders/try/mac/b8836587739777488289/overview

```
======================================================================
FAIL: test_pthread_abort (test_core.wasm0)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/s/w/ir/cipd_bin_packages/cpython3/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/opt/s/w/ir/cipd_bin_packages/cpython3/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/opt/s/w/ir/cipd_bin_packages/cpython3/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/opt/s/w/ir/k/install/emscripten/tests/common.py", line 188, in decorated
    f(self, *args, **kwargs)
  File "/opt/s/w/ir/k/install/emscripten/tests/test_core.py", line 2361, in test_pthread_abort
    self.do_run_in_out_file_test('pthread/test_pthread_abort.c', assert_returncode=NON_ZERO)
  File "/opt/s/w/ir/k/install/emscripten/tests/common.py", line 1018, in do_run_in_out_file_test
    self._build_and_run(srcfile, expected, **kwargs)
  File "/opt/s/w/ir/k/install/emscripten/tests/common.py", line 1074, in _build_and_run
    self.assertContained(expected_output, js_output)
  File "/opt/s/w/ir/k/install/emscripten/tests/common.py", line 744, in assertContained
    self.fail("Expected to find '%s' in '%s', diff:\n\n%s\n%s" % (
  File "/opt/s/w/ir/cipd_bin_packages/cpython3/lib/python3.8/unittest/case.py", line 753, in fail
    raise self.failureException(msg)
AssertionError: Expected to find 'onAbort called
' in 'undefined
pthread sent an error! undefined:undefined: abort(undefined) at Error
    at jsStackTrace (eval at importScripts (/opt/s/w/ir/tmp/t/tmpw4m97k_f/emscripten_test_wasm0_hsj_lnsx/test_pthread_abort.worker.js:38:16), <anonymous>:2232:19)
[...]

PR #25041, opened 2025-08-25
Sequentialize uncaught exceptions in Node.js Workers

juj:
Sequentialize uncaught exceptions in Node.js Workers into the same postMessage() stream that other messages go through, to ensure they are received by the main thread in the order they were generated.

Fixes https://github.com/emscripten-core/emscripten/issues/15014.

See https://github.com/nodejs/node/issues/59617.

### DIFF
--- a/src/lib/libpthread.js
+++ b/src/lib/libpthread.js
@@ -241,6 +241,9 @@ var LibraryPThread = {
       worker.onmessage = (e) => {
         var d = e['data'];
         var cmd = d.cmd;
+#if PTHREADS_DEBUG
+        dbg(`main thread: received message '${cmd}' from worker. ${JSON.stringify(d)}`);
+#endif
 
         // If this message is intended to a recipient that is not the main
         // thread, forward it to the target thread.
@@ -283,6 +286,13 @@ var LibraryPThread = {
           // Worker wants to postMessage() to itself to implement setImmediate()
           // emulation.
           worker.postMessage(d);
+#if ENVIRONMENT_MAY_BE_NODE
+        } else if (cmd === 'uncaughtException') {
+          // Message handler for Node.js specific out-of-order behavior:
+          // https://github.com/nodejs/node/issues/59617
+          // A pthread sent an uncaught exception event. Re-raise it on the main thread.
+          worker.onerror(d.error);
+#endif
         } else if (cmd === 'callHandler') {
           Module[d.handler](...d.args);
         } else if (cmd) {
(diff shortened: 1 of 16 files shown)


src/runtime_common.js
```diff
@@ -40,6 +40,21 @@ if (ENVIRONMENT_IS_NODE && {{{ ENVIRONMENT_IS_WORKER_THREAD() }}}) {
...
+  // Node.js Workers do not pass postMessage()s and uncaught exception events to the parent
+  // thread necessarily in the same order where they were generated in sequential program order.
+  // See https://github.com/nodejs/node/issues/59617
+  // To remedy this, capture all uncaughtExceptions in the Worker, and sequentialize those over
+  // to the same postMessage pipe that other messages use.
+  process.on("uncaughtException", (err) => {
+#if PTHREADS_DEBUG
+    dbg(`uncaughtException on worker thread: ${err.message}`);
+#endif
+    parentPort['postMessage']({ cmd: 'uncaughtException', error: err });
```
sbc100: I think you can just do `postMessage(...)` here can't you?  (because it was assigned to globalThis above).

juj: I think it's nicer to use the native node.js API here because this code block is under the native "we're in Node.js" world. 

The `postMessage()` API is rather targeted for emulating the Web Worker like API out to the pthreads code that was developed assuming browser Web Workers.

juj:
> So if I'm understanding correctly, issue is with the `onAbort` handler, which is being called via the `callHandler` message is racing with the `onerror` handler for uncaught exceptions?
> 
> Nice work find this!

Thanks - that looks like the case. For some reason Node.js internally uses a different mechanism to propagate postMessage()s vs process.on('uncaughtError')s from Worker to parent.

src/runtime_common.js
sbc100: Perhaps, but its slightly worse for code size right?   lgtm either way.

juj: Hmm no argument there, updated.

src/lib/libpthread.js
```diff
@@ -241,6 +241,9 @@ var LibraryPThread = {
       worker.onmessage = (e) => {
         var d = e['data'];
         var cmd = d.cmd;
+#if PTHREADS_DEBUG
+        dbg(`main thread: received message '${cmd}' from worker. ${JSON.stringify(d)}`);
```
sbc100: How is `${JSON.stringify(d)}` different from just `${d}` here? 

Is this line going to produce a log (too much?) of debug output?

juj: I was getting `[Object]` prints in some of the debug logs before, so JSON.stringify helped. Though this should be a regular `{}` so maybe that won't matter.

> Is this line going to produce a log (too much?) of debug output?

You mean in general? I understand `PTHREADS_DEBUG` flag users won't matter about print output, but getting useful logs is important. I left the debug prints that helped diagnose the issue, thinking that those would probably be the useful ones debugging other types of problems as well.

PR #25042, opened 2025-08-25
fix: Big endian support does not work since 4.0.9

slavek-kucera:
This PR fixes issues with big endian support introduced in the following PRs:
- #24295 - The change did not consider the `littleEndianHeap` transform implications,
- #24283 - The new LE_HEAP* functions were not added to link.py, causing runtime failure on BE system.

The `littleEndianHeap` transform code is simplified and updated to detect the code introduced by `growableHeap` pass.
A new test was added to verify the combined transformation generates working code.


juj:
What hardware do you run on that is Big Endian? Are there maybe single board computers, or specific Big Endian OSes for ARM computers that would fit this target?

I'm pondering how this could be kept working in the future. I presume e.g. CircleCI doesn't have BE targets in its catalog.

slavek-kucera:
I am running it on s390x both z/linux and z/os.

The testing could be handled as described [here](https://github.com/emscripten-core/emscripten/pull/23700#issuecomment-2682684468). I am just not familiar enough with the project to implement it without a lot of research or trial and error.

juj:
Oh that's cool, I wasn't familiar with qemu, interesting to see it can be used to run a single program invocation like that.

Gave that a spin at https://github.com/emscripten-core/emscripten/compare/main...juj:emscripten:bigendian_test_suite , and indeed it does work out, and some Emscripten core tests pass at least.

sbc100:
I think we should probably add an "experimental" warning when using the BIGENDIAN setting, since its maintained at only best effort level.   There are probably only one or two users our there so I think its fine for it to be incumbent upon those users to keep it alive and kicking.

slavek-kucera:
Would it be ok to pre-install the dependencies into the `emscripten-ci` image including the BE node binaries?
If so, we could place the node into e.g. /opt/node-be and introduce a setup-bigendian command into the ci (as documented in the juj's branch) and run a couple of basic tests.

juj:
I think it would be good to be able to run a smoke test or two at least. Though I don't know what the capabilities of CircleCI would be.

I ran the core suite in bigendian mode locally on the contents of this PR. Many tests do pass there out of the box:

though some are still red.

Skipping individual tests in big endian mode is easy, if you would like to work through the failures in the core suite, and skip any tests that might be still failing: https://github.com/emscripten-core/emscripten/commit/46a2cdf75f78839cec47a3010e24cf02894826bd

That way you could mark down TODOs/BUGs to get towards a passing big endian suite.

juj:
To reproduce a red test in that graph, you can run e.g.

```
test/runner bigendian.test_segfault
```

```
$ test/runner bigendian.test_segfault
Running test_core: (1 tests)
test_segfault (test_core.bigendian.test_segfault) ... get_null()
new D2()
-- begin program output --
Aborted(segmentation fault, exceeded the top of the available dynamic heap when loading 1 bytes at address 1325465856. DYNAMICTOP=77824)
/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:505
  /** @suppress {checkTypes} */ var e = new WebAssembly.RuntimeError(what);
                                        ^

RuntimeError: Aborted(segmentation fault, exceeded the top of the available dynamic heap when loading 1 bytes at address 1325465856. DYNAMICTOP=77824)
    at abort (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:505:41)
    at SAFE_HEAP_INDEX (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:389:29)
    at SAFE_HEAP_LOAD (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:398:14)
    at _fd_write (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:1110:21)
    at wasm://wasm/0001fcf2:wasm-function[21]:0x7d8
    at wasm://wasm/0001fcf2:wasm-function[38]:0xe63
    at wasm://wasm/0001fcf2:wasm-function[43]:0x1c9d
    at wasm://wasm/0001fcf2:wasm-function[42]:0x134c
    at wasm://wasm/0001fcf2:wasm-function[41]:0x11c3
    at wasm://wasm/0001fcf2:wasm-function[50]:0x2169
    at wasm://wasm/0001fcf2:wasm-function[20]:0x754
    at wasm://wasm/0001fcf2:wasm-function[17]:0x6c4
    at wasm://wasm/0001fcf2:wasm-function[8]:0x4fd
    at /mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:553:12
    at callMain (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:1368:15)
    at doRun (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:1410:24)
    at run (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:1420:5)
    at removeRunDependency (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:830:7)
    at receiveInstance (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:639:5)
    at receiveInstantiationResult (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:655:12)
    at createWasm (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:678:17)
Thrown at:
    at abort (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:505:41)
    at SAFE_HEAP_INDEX (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:389:29)
    at SAFE_HEAP_LOAD (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:398:14)
    at _fd_write (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:1110:21)
    at $func21 (wasm://wasm/0001fcf2:1:2009)
    at $func38 (wasm://wasm/0001fcf2:1:3684)
    at $func43 (wasm://wasm/0001fcf2:1:7326)
    at $func42 (wasm://wasm/0001fcf2:1:4941)
    at $func41 (wasm://wasm/0001fcf2:1:4548)
    at $func50 (wasm://wasm/0001fcf2:1:8554)


Node.js v22.16.0
-- end program output --
FAIL

FAIL: test_segfault (test_core.bigendian.test_segfault)
  File "/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/test/test_core.py", line 304, in decorated
    f(self, *args, **kwargs)
  File "/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/test/test_core.py", line 1784, in test_segfault
    self.do_run(src, 'marfoosh')
  File "/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/test/common.py", line 1969, in do_run
    return self._build_and_run(filename, expected_output, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/test/common.py", line 2020, in _build_and_run
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/test/common.py", line 1600, in run_js
AssertionError: JS subprocess failed (qemu-s390x -L /usr/s390x-linux-gnu/ /mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/nodebe/node-v22.16.0-linux-s390x/bin/node --stack-trace-limit=50 --trace-uncaught /mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js): 1 (expected=0).  Output:
Aborted(segmentation fault, exceeded the top of the available dynamic heap when loading 1 bytes at address 1325465856. DYNAMICTOP=77824)
/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:505
  /** @suppress {checkTypes} */ var e = new WebAssembly.RuntimeError(what);
                                        ^

RuntimeError: Aborted(segmentation fault, exceeded the top of the available dynamic heap when loading 1 bytes at address 1325465856. DYNAMICTOP=77824)
    at abort (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:505:41)
    at SAFE_HEAP_INDEX (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:389:29)
    at SAFE_HEAP_LOAD (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:398:14)
    at _fd_write (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:1110:21)
    at wasm://wasm/0001fcf2:wasm-function[21]:0x7d8
    at wasm://wasm/0001fcf2:wasm-function[38]:0xe63
    at wasm://wasm/0001fcf2:wasm-function[43]:0x1c9d
    at wasm://wasm/0001fcf2:wasm-function[42]:0x134c
    at wasm://wasm/0001fcf2:wasm-function[41]:0x11c3
    at wasm://wasm/0001fcf2:wasm-function[50]:0x2169
    at wasm://wasm/0001fcf2:wasm-function[20]:0x754
    at wasm://wasm/0001fcf2:wasm-function[17]:0x6c4
    at wasm://wasm/0001fcf2:wasm-function[8]:0x4fd
    at /mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:553:12
    at callMain (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:1368:15)
    at doRun (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:1410:24)
    at run (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:1420:5)
    at removeRunDependency (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:830:7)
    at receiveInstance (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:639:5)
    at receiveInstantiationResult (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:655:12)
    at createWasm (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:678:17)
Thrown at:
    at abort (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:505:41)
    at SAFE_HEAP_INDEX (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:389:29)
    at SAFE_HEAP_LOAD (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:398:14)
    at _fd_write (/mnt/ssd1000/emscripten_buildbot/worker_root/x10dax-linux-mint22/emscripten_x10dax/build/emscripten/main/out/test/src.js:1110:21)
    at $func21 (wasm://wasm/0001fcf2:1:2009)
    at $func38 (wasm://wasm/0001fcf2:1:3684)
    at $func43 (wasm://wasm/0001fcf2:1:7326)
    at $func42 (wasm://wasm/0001fcf2:1:4941)
    at $func41 (wasm://wasm/0001fcf2:1:4548)
    at $func50 (wasm://wasm/0001fcf2:1:8554)


Node.js v22.16.0


Ran 1 test in 30.909s

FAILED (failures=1)
```

slavek-kucera:
It looks like most of the failing tests are SAFE_HEAP and asan related and then x86 specific.

As far as I can tell, the SAFE_HEAP is currently incompatible with both big endian and growable heap support. I'd say the same is true for the asan support (for the same reason).

slavek-kucera:
I think there might be a way to make all these options compatible.
The only transforms that actually have to replace the `HEAPxx[idx]` code seems to be the `littleEndianHeap` and `growableHeap`.

The `safeHeap` and `asanify` only perform checks on the index so they could be implemented as `HEAPxx[idx]` => `HEAPxx[func(HEAPxx, idx)]` transforms, where `func` returns `idx` and they could be easily composed (`f1(HEAPxx, f2(HEAPxx, idx))`).
The `growableHeap` and `littleEndianHeap` are compatible with this PR and so they could run last in that order.

I could look into that after this PR is merged.

juj:
I should be able to plop in a run of the core suite with the big endian Node.js on my test lab CI at http://clbri.com:8010/#/builders

Posted a PR https://github.com/emscripten-core/emsdk/pull/1590 that will help automate the setup of the big endian Node on the CI, it should be a trivial one-liner setup then.

sbc100:
> I should be able to plop in a run of the core suite with the big endian Node.js on my test lab CI at http://clbri.com:8010/#/builders
> 
> Posted a PR [emscripten-core/emsdk#1590](https://github.com/emscripten-core/emsdk/pull/1590) that will help automate the setup of the big endian Node on the CI, it should be a trivial one-liner setup then.

Regarding the `NODE_JS` config setting, you can use `NODE_JS_TEST` (used to run tests) which is separate setting to `NODE_JS` (used during compilation).   The former defaults to the latter.

slavek-kucera:
I put together a patch that would make all the options compatible https://github.com/slavek-kucera/emscripten/compare/fix_big_endian_after_4_0_9...fix_big_endian_other_options_preview

I was able to successfully run with `-pthread -sALLOW_MEMORY_GROWTH=1 -sSAFE_HEAP=1 -sSUPPORT_BIG_ENDIAN=1 -sPROXY_TO_PTHREAD=1 -sEXIT_RUNTIME=1` on both little and big endian platform.

juj:
> Regarding the `NODE_JS` config setting, you can use `NODE_JS_TEST` (used to run tests) which is separate setting to `NODE_JS` (used during compilation). The former defaults to the latter.

Gave that a try, though there seems to be some interaction with `js_engines` variable that either needs some manual configuring that I missed, or maybe something that `js_engines` didn't anticipate with `NODE_JS_TEST` variable? Posted at https://github.com/emscripten-core/emscripten/pull/25063#issuecomment-3228388422 about that.

Also, posted PR #25068 to land the bigendian0 suite.

slavek-kucera:
> I'm currently on vacation so can't look in depth but I think the right fix here should be just reordering transforms. I theory they shouldn't have to know about each other.

The transforms are replacing the very identifiers other transforms are looking for, so I don't think this is correct.

RReverser:
> The transforms are replacing the very identifiers other transforms are looking for, so I don't think this is correct.

Yeah sorry, I realised that shortly after posting my comment, so deleted it but I guess not quickly enough.

Your approach from the last comment, combining the transforms, seems like a better one.

juj:
> I put together a patch that would make all the options compatible [slavek-kucera/emscripten@fix_big_endian_after_4_0_9...fix_big_endian_other_options_preview](https://github.com/slavek-kucera/emscripten/compare/fix_big_endian_after_4_0_9...fix_big_endian_other_options_preview)

I ran the `bigendian` suite with these patches on top. Definitely more tests pass now, here is the result I got:

https://clbri.com/dump/emscripten/toolchain_profiler.results_20250827_1943.html

juj:
After https://github.com/emscripten-core/emscripten/pull/25074 lands, the big endian test suite will be live on `emscripten linux x64` runner at http://clbri.com:8010/#/builders

That CI is my test lab CI, so not sure if it will be able to live indefinitely, though I'll try to keep it alive into the future.

It will be a post-landing runner unfortunately, so will only run the bigendian0 suite on Emscripten's main branch after code has landed to it. But it should show the latest status then.

sbc100:
> After #25074 lands, the big endian test suite will be live on `emscripten linux x64` runner at http://clbri.com:8010/#/builders
> 
> That CI is my test lab CI, so not sure if it will be able to live indefinitely, though I'll try to keep it alive into the future.
> 
> It will be a post-landing runner unfortunately, so will only run the bigendian0 suite on Emscripten's main branch after code has landed to it. But it should show the latest status then.

We also run a bunch of post-landing test  suites on the "Test Suites" running here: https://ci.chromium.org/p/emscripten-releases/g/main/console.  We consider that an "FYI bot" which is what chromium calls bots that don't close the tree, and could , therefore, be red to long-ish periods of time.   

We should probably document here in `test_core.py` which of the core test configurations are maintained in FYI/best-effort/post-landing way.

For configurations such as bigendian I don't really think its reasonable for the core team them green at all times so FYI/best-effort/post-landing is where I think it belongs.

juj:
Ok, all the necessary pieces for the bigendian0 suite infra have landed in emsdk and emscripten repository. This was the first live run of that suite against Emscripten `main` branch: http://clbri.com:8010/#/builders/11/builds/303

slavek-kucera:
Looking through the last test report, there is still a few SSE/AVX tests running.
As for the webidl tests, I am not using this feature so I am not familiar with how it works, but I suspect the problem might be [here](https://github.com/emscripten-core/emscripten/blob/29f8535fd6960c32cec165e008fe9e3a5c5e89f7/tools/webidl_binder.py#L241) where the code won't use the little endian transforms.

juj:
You can go ahead and mark any still failing tests as `@no_big_endian('TODO: Fails with error xyz')` to get the suite looking green for the tests that do pass. That will then give a foothold over keeping track of the passing tests to stay green in the Emscripten main branch.

slavek-kucera:
It looks like with node24 even the EH tests are passing.

As far as I can tell. The only problem is with test_modularize_instance_embind. It is as if the littleEndian transform is not applied at all, I am not sure why at the moment.

juj:
I ran the test suites on top of this PR, at http://clbri.com:8010/#/builders/11/builds/321

Looks like this is in a very good shape now, only few reds remaining.

src/lib/liblittle_endian_heap.js
```diff
@@ -115,8 +115,8 @@ function LE_HEAP_UPDATE() {
   },
   $LE_ATOMICS_WAITASYNC: (heap, offset, value, timeout) => {
     const order = LE_ATOMICS_NATIVE_BYTE_ORDER[heap.BYTES_PER_ELEMENT - 1];
-    /** @suppress {checkTypes} */
-    return Atomics.waitAsync(heap, offset, order(value), timeout);
+    // this is suppressing incorrect closure JSC_INEXISTENT_PROPERTY warning that cannot be suppress via annotation
+    return Atomics.waitAsync && Atomics.waitAsync(heap, offset, order(value), timeout);
```
juj: Try adding `Atomics.waitAsync` to this list: https://github.com/emscripten-core/emscripten/blob/452e7dbc926aed36e13d90391aa576447cbaea3a/src/closure-externs/closure-externs.js#L63-L69

slavek-kucera: It looks like any way you define it, there is a test case that breaks.

juj: Interesting.. can you list some test cases that croak with the different approaches?

slavek-kucera: I think each of the following failed differently due to the closure compiler warning
- test_modularize_closure_pre
- test_closure_full_js_library_pthread
- test_audio_worklet_minimal_runtime_pthreads_and_closure

juj: Oh I see. Try with the following:

```diff
diff --git a/src/closure-externs/closure-externs.js b/src/closure-externs/closure-externs.js
index 1681cd2b4..b5bfc5857 100644
--- a/src/closure-externs/closure-externs.js
+++ b/src/closure-externs/closure-externs.js
@@ -64,6 +64,11 @@ var Atomics = {};
 Atomics.compareExchange = function() {};
 Atomics.exchange = function() {};
 Atomics.wait = function() {};
+/**
+ * @param {number=} maxWaitMilliseconds
+ * @suppress {duplicate, checkTypes}
+ */
+Atomics.waitAsync = function(i32a, index, value, maxWaitMilliseconds) {};
 Atomics.notify = function() {};
 Atomics.load = function() {};
 Atomics.store = function() {};
```
I think that should help Closure avoid tripping itself up in different build modes.

slavek-kucera: This version seems to work, thanks.

slavek-kucera:
@juj Could you please re-test the latest version? The only tests failing at the moment should be the code_size ones (which might change 10 times before the review is done).

juj:
Certainly. Here's two runs of the latest on this branch:
http://clbri.com:8010/#/builders/11/builds/360
http://clbri.com:8010/#/builders/11/builds/359

(ignore the orange mark on wasm2js3, that is my CI flagging any all-caps string `WARNING` as an orange run output)

juj:
`other.test_gen_sig_info` could be something relevant.

slavek-kucera:
I agree except both `other.test_gen_sig_info` and the web_idl tests are passing for me locally.
Is it possible that the cache is not cleared?

juj:
Hmmm, maybe... Disturbing the version string will definitely clear the cache. Did that here: https://github.com/juj/emscripten/commit/58618f791f5a5888767142c3ee60b09270ded274

and running a new one now: http://clbri.com:8010/#/builders/11/builds/361

Back in the day we had a versioning convention to bump the revision version whenever any system file is changed, to ensure that caches for users fill be fresh for the changes.

slavek-kucera:
Any idea why those 3 tests are timing out? They run in about 6-7 second locally.

juj:
> Any idea why those 3 tests are timing out? They run in about 6-7 second locally.

Sigh, this is really something peculiar... I can reproduce the issue, but only in the parallel test runner, by running

```
test/runner bigendian0
```
then the three tests hang. Running e.g.
```
test/runner bigendian0.test_webidl
```
does not hang.

When the hang occurs, all the threads on the system (not just the ones running the test) are stuck (and interrupted by pressing Ctrl-C) on

```
KeyboardInterrupt
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/multiprocessing/process.py", line 313, in _bootstrap
    self.run()
    ~~~~~~~~^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/multiprocessing/pool.py", line 114, in worker
    task = get()
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/multiprocessing/queues.py", line 384, in get
    with self._rlock:
         ^^^^^^^^^^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/multiprocessing/synchronize.py", line 95, in __enter__
    return self._semlock.__enter__()
           ~~~~~~~~~~~~~~~~~~~~~~~^^
```

The hang does not occur on all runs of `test/runner bigendian0`, but sill on a good >50% times the suite is run.

This seems like yet another of the countless cases of buggy Python multiprocessing functionality.. I'll poke into this further, feel free to ignore these failures if the tests pass for you locally.

juj:
Actually I take that back.. these python threads are doing what they are supposed to be doing.

It is indeed the `qemu-s390x` process hanging. It is consuming 0% of CPU, and running the test case.

If I remove the stdout and stderr redirection to file from the Python jsrun code, so that it prints to terminal stdout.. then I can see that all the expected prints from the executed test case .js come out correctly. It is just that either node.js is refusing to shut down, or somehow qemu-s390x is refusing to shut down after the quit. I don't know which.

And I don't quite understand why this is happening only in parallel runner, and never in single-threaded runner.

Manually running the test .js file in qemu-s390x node never hangs. Also tried to create a python script that hammers that test .js script across all threads using Python threading, though that never hangs either. I am quite stumped.

If this hang only occurs on my CI, then it is fine to go ahead, and I can mark those `test_webidl` tests as skips for the time being. It would be good to get this PR reviewed and landed, since there is a lot of development in it already. @kripken would you have a chance to look at the codegen passes maybe?

juj:
The `test_webidl*` tests time out after five minutes.

If I add a JS code block

```diff
diff --git a/src/shell.js b/src/shell.js
index 5e9e11b6e..c37841974 100644
--- a/src/shell.js
+++ b/src/shell.js
@@ -7,6 +7,19 @@
 "use strict";
 
 #endif
+var xxtimeout = setTimeout(() => {
+  console.log('EXPLICIT EXIT');
+  process.exit(1);
+}, 50000);
+xxtimeout.unref();
```
with the expectation that the hanging node.js invocations should then quit after 50 seconds and not five minutes.. this never happens.

So this suggests that the qemu-s390x processes are properly hung, and not e.g. idle in the Node.js event loop.

juj:
And if I add

```diff
diff --git a/test/test_core.py b/test/test_core.py
index 26a510bb5..fae9def54 100644
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -7822,6 +7822,7 @@ void* operator new(size_t size) {
   @no_modularize_instance('uses Module global')
   @no_strict('TODO: Fails in -sSTRICT mode due to an unknown reason.')
   def test_webidl(self, mode, allow_memory_growth):
+    time.sleep(30)
     self.set_setting('WASM_ASYNC_COMPILATION', 0)
     if self.maybe_closure():
       # avoid closure minified names competing with our test code in the global name space
```
in the beginning of test_webidl tests, then the tests begin to pass and never hang. Tried 10 runs of the suite - zero hangs, whereas without the sleep, the suite hangs almost always in those test_webidl tests.

Something is properly wrong either in Python, or in qemu-s390x..

RReverser:
I wonder if it would be easier to wrap all the global typed arrays + dataview with JS Proxy endian swap for big-endian simulation for testing.

In theory that should work quite well - there are no other APIs with observable endianness in JS - and would be much easier than having a dedicated VM.

We already do something similar for bigint64 array polyfill here: https://github.com/emscripten-core/emscripten/blob/7ed425edb82161c3d60ef5d00f939446edddce42/src/polyfill/bigint64array.js#L39

slavek-kucera:
> I wonder if it would be easier to wrap all the global typed arrays + dataview with JS Proxy endian swap for big-endian simulation for testing.

This should perhaps be done not only for the bigendian problem. As far as I can tell all the other memory access enhancing features are subtlety broken when the typedarray is passed as a parameter - growable heap does not check for reallocation, safe heap and asan don't verify accessed offset and of course big endian does not read/write the memory correctly (this one is not so subtle).

slavek-kucera:
@juj It looks like it is indeed qemu problem. I ran 1000 tests with 64x parallel instances and eventually the qemu got stuck.
I ran the same test on ubuntu 24.04 and did not observe the problem (I additionally ran also 10k with 128x - also no issues).
The versions are of course different
emsdk image offers `qemu-s390x version 6.2.0 (Debian 1:6.2+dfsg-2ubuntu6.26)`
ubuntu 24.04 `qemu-s390x version 8.2.2 (Debian 1:8.2.2+ds-0ubuntu1.9)`

juj:
Ok I got it now. This  looks like either a qemu bug or a node-s390x bug.

Attached a repro that repeatedly hammers launching the webidl test case JS from a python script. When run wide with # of parallel cores on the system, then the qemu-s390x launched nodes start to hang at exit after a few minutes.

[qemu_node_hang_testcase.tar.gz](https://github.com/user-attachments/files/22084195/qemu_node_hang_testcase.tar.gz)

The issue reproes at least on this 256 thread workstation (2x Epyc 7763). If I reduce the parallel number of qemu instances to <= 32 in the test, then I am able to run the testcase for more than one hour without a single hang.

I adjusted my CI to limit parallelization when running the `bigendian0` suite, and that "fixed" the issue from happening.

@slavek-kucera I'm curious if you might be able to also repro the problem with the above test case?

juj:
Ok, that's interesting. My qemu is

```
$ qemu-s390x --version
qemu-s390x version 8.2.2 (Debian 1:8.2.2+ds-0ubuntu1.9)
Copyright (c) 2003-2023 Fabrice Bellard and the QEMU Project developers
```
(this is running on a Linux Mint 22)

slavek-kucera:
Your test case starts failing almost immediately within the emsdk image. But seems to be running fine in the ubuntu:24.04.

juj:
Does the s390x Node.js version differ between the two? I added Node.js 22.16.0 to Emsdk manifest to match version with the other OSes. (that is what my CI runs on) I do also have an open PR at https://github.com/emscripten-core/emsdk/pull/1597 to update Node to 24.7.0 that also includes the s390x version. I wonder if that might help any.

juj:
What is the emsdk image you are referring to? Someone has made e.g. a docker image with emsdk in it?

slavek-kucera:
emscripten/emsdk

juj:
Gotcha, yeah - https://hub.docker.com/r/emscripten/emsdk - should then install that same s390x Node 22.16.0.

slavek-kucera:
It looks like I mixed the node versions between the images. Node 22 fails with both versions of qemu, Node 24 seems ok

juj:
Ok that's fantastic. Once https://github.com/emscripten-core/emsdk/pull/1597 lands I can then change the CI to test on Node 24.7.0 and that should avoid the issue.

test/test_core.py
```diff
@@ -8600,6 +8638,7 @@ def test_wasm2js(self):
   @no_asan('no wasm2js support yet in asan')
   @requires_wasm2js
   @also_with_minimal_runtime
+  @no_big_endian('wasm2js is currently not compatible with big endian')
```
juj, line 8550: Can you add a line after this line https://github.com/emscripten-core/emscripten/blob/7ed425edb82161c3d60ef5d00f939446edddce42/tools/link.py#L847 inside `if settings.SUPPORT_BIG_ENDIAN:`

```py
if settings.WASM2JS:
  exit_with_error('WASMJ2S is currently not compatible with SUPPORT_BIG_ENDIAN')
```

juj:
Regarding the code size changes, there is one regression that CircleCI reports that looks odd:

```
FAIL [0.006s]: test_minimal_runtime_code_size_hello_webgl2_wasm (test_other.other.test_minimal_runtime_code_size_hello_webgl2_wasm)
  File "C:\Users\circleci\emsdk\python\3.13.3_64bit\Lib\unittest\case.py", line 58, in testPartExecutor
    yield
  File "C:\Users\circleci\emsdk\python\3.13.3_64bit\Lib\unittest\case.py", line 651, in run
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "C:\Users\circleci\emsdk\python\3.13.3_64bit\Lib\unittest\case.py", line 606, in _callTestMethod
    if method() is not None:
       ~~~~~~^^
  File "C:\Users\circleci\path with spaces ()\test\common.py", line 986, in resulting_test
    return func(self, *args)
  File "C:\Users\circleci\path with spaces ()\test\test_other.py", line 12026, in test_minimal_runtime_code_size
    self.check_output_sizes(*outputs)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "C:\Users\circleci\path with spaces ()\test\test_other.py", line 12094, in check_output_sizes
    self.assertTextDataIdentical(expected_results_json, obtained_results_json)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\circleci\path with spaces ()\test\common.py", line 1703, in assertTextDataIdentical
    return self.assertIdentical(text1, text2, msg, fromfile, tofile)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\circleci\path with spaces ()\test\common.py", line 1723, in assertIdentical
    self.fail(fail_message)
    ~~~~~~~~~^^^^^^^^^^^^^^
  File "C:\Users\circleci\emsdk\python\3.13.3_64bit\Lib\unittest\case.py", line 732, in fail
AssertionError: Unexpected difference:
--- expected
+++ actual
@@ -3,8 +3,8 @@
   "a.html.gz": 321,
   "a.js": 4437,
   "a.js.gz": 2281,
-  "a.wasm": 8299,
-  "a.wasm.gz": 5645,
-  "total": 13190,
-  "total_gz": 8247
+  "a.wasm": 8317,
+  "a.wasm.gz": 5660,
+  "total": 13208,
+  "total_gz": 8262
 }

For full output run with --verbose.
```
The `test_minimal_runtime_code_size_hello_webgl2_wasm` code size test is one of the most important code size tests, and I can't quite figure out what in this PR could regress that.

When I locally run the code size test `other.test_minimal_runtime_code_size_hello_webgl2_wasm` on this branch, I get the same numbers as CircleCI does:

```
C:\emsdk\emscripten\main>test\runner other.test_minimal_runtime_code_size_hello_webgl2_wasm
Running test_other: (1 tests)
test_minimal_runtime_code_size_hello_webgl2_wasm (test_other.other.test_minimal_runtime_code_size_hello_webgl2_wasm) ... 'C:\emsdk\emscripten\main\emcc.bat' -o a.html -sMINIMAL_RUNTIME=2 -sENVIRONMENT=web -sTEXTDECODER=2 -sDYNAMIC_EXECUTION=0 -sABORTING_MALLOC=0 -sALLOW_MEMORY_GROWTH=0 -sDECLARE_ASM_MODULE_EXPORTS -sMALLOC=emmalloc -sGL_EMULATE_GLES_VERSION_STRING_FORMAT=0 -sGL_EXTENSIONS_IN_PREFIXED_FORMAT=0 -sGL_SUPPORT_AUTOMATIC_ENABLE_EXTENSIONS=0 -sGL_SUPPORT_SIMPLE_ENABLE_EXTENSIONS=0 -sGL_TRACK_ERRORS=0 -sGL_POOL_TEMP_BUFFERS=0 -sGL_WORKAROUND_SAFARI_GETCONTEXT_BUG=0 -sGL_ENABLE_GET_PROC_ADDRESS=0 -sNO_FILESYSTEM -sSTRICT --output-eol linux -Oz --closure=1 -DNDEBUG -ffast-math 'C:\emsdk\emscripten\main\test\minimal_webgl\main.c' 'C:\emsdk\emscripten\main\test\minimal_webgl\webgl.c' --js-library 'C:\emsdk\emscripten\main\test\minimal_webgl\library_js.js' -lGL -sMODULARIZE -sMAX_WEBGL_VERSION=2
a.html: size=454, expected=454
a.html.gz: size=321, expected=321
a.js: size=4437, expected=4437
a.js.gz: size=2281, expected=2281
a.wasm: size=8299, expected=8299
a.wasm.gz: size=5645, expected=5645
total: size=13190, expected=13190
total_gz: size=8247, expected=8247
ok

Ran 1 test in 4.659s

OK
```
i.e. the regression does not reproduce. It does not look like your PR branch would be that out of date with the upstream `main` branch either. A bit puzzling..

juj:
I triggered a rebaseline run on `main`, https://github.com/emscripten-core/emscripten/pull/25133 . Once that lands, you can try merging/rebasing with `origin/main` to see if that helps with the code size test changes.

slavek-kucera:
I think this PR is ready now.

juj (approved):
Fantastic. Gave this a run at http://clbri.com:8010/#/builders/11/builds/452 and everything looks green now (ignore the code size tests red there, I merged that with main to test)

@RReverser do you have a moment to check?

juj:
Hey this would be great to land soon so it doesn't leave simmering, since I did all the effort to add testing for as well.

If @RReverser is not available for review, then maybe @kripken might be able to take a peek at the acorn JS optimizer pass logic? To me it all looks good to land.

src/lib/libemval.js
```diff
@@ -109,6 +110,85 @@ var LibraryEmVal = {
...
+          if (view instanceof Uint32Array)
+            return i => dv.getUint32(i * 4, true)
+          if (view instanceof Float32Array)
+            return i => dv.getFloat32(i * 4, true)
+          break;
+        case 8:
+          if (view instanceof BigInt32Array)
+            return i => dv.getBigInt32(i * 8, true)
+          if (view instanceof BigUint32Array)
+            return i => dv.getBigUint32(i * 8, true)
```
RReverser: I assume these were supposed to be BigInt**64**. The tests should've caught these typos, worth extending to cover these cases too.

src/lib/libemval.js
```diff
@@ -109,6 +110,85 @@ var LibraryEmVal = {
...
+    dst.set(src);
+  },
+#else
+  _emval_new_array_from_memory_view__deps: ['$Emval'],
+  _emval_new_array_from_memory_view: (view) => {
+    view = Emval.toValue(view);
+    var reader = (() => {
+      if (view.BYTES_PER_ELEMENT===1)
+        return i => view[i];
+      const dv = new DataView(view.buffer, view.byteOffset);
```
RReverser: Nit: you could get away with a lot less code by simply building a map from typed array names (since the list is exhaustive anyway) and avoiding custom reader wrappers. Something like:

```js
const dv = new DataView(view.buffer, view.byteOffset);
const reader = {
  Int8Array: dv.getInt8,
  Uint8Array: dv.getUint8,
  Int16Array: dv.getInt16,
  ...
}[view.constructor.name];
var a = new Array(view.length);
for (var i = 0; i < view.length; i++) a[i] = reader.call(dv, i * view.BYTES_PER_ELEMENT, true);
return Emval.toHandle(a);
```

Same for the writer below.

tools/acorn-optimizer.mjs
```diff
@@ -1249,21 +1210,40 @@ function isHEAPAccess(node) {
     node.type === 'MemberExpression' &&
     node.object.type === 'Identifier' &&
     node.computed && // notice a[X] but not a.X
-    isEmscriptenHEAP(node.object.name)
+    isEmscriptenHEAP(node.object.name) &&
+    node.object.name
+  );
+}
+
+function isGrowHEAPAccess(node) {
```
RReverser, line 1218: Hm it's not ideal that we produce a sequence and then have to pattern-match it back.

Can't we just merge the pass handlers together instead, so it only has to look for `HEAP8[...]` and such accesses, and does any enabled transforms in one go?

tools/webidl_binder.py
```diff
@@ -238,20 +238,16 @@ def build_constructor(name):
...
 
 /** @suppress {duplicate} (TODO: avoid emitting this multiple times, it is redundant) */
 function ensureString(value) {
   if (typeof value === 'string') {
     var intArray = intArrayFromString(value);
     var offset = ensureCache.alloc(intArray, HEAP8);
-    ensureCache.copy(intArray, HEAP8, offset);
+    for (var i = 0; i < intArray.length; i++) {
+      HEAP8[offset + i] = intArray[i];
+    }
```
RReverser, line 250: It would be nicer if we could use & pattern-match `HEAP8.set(intArray, offset);` and such in the heap transforms, so that little-endian can use the performant path (`.set` is essentially binding for fast `memcpy`) instead of copying bytes one by one.

src/lib/libemval.js
slavek-kucera: I thought about something like that but I was not sure if this is expected to work classes that extend the typedarrays.

tools/acorn-optimizer.mjs
slavek-kucera: I agree, but this sounds like a larger rewrite of how the optimizer is invoked.
At the moment only these two transforms seem to be interfering with each other. Moreover, if the idea with using proxies is pursued all of these passes could be eliminated completely...

tools/webidl_binder.py
slavek-kucera: Actually, I don't think there is any little-endian/big-endian issue here. This is dealing with bytes and it was just a direct expansion of the copy function.

I think it might be even possible to just call the stringToUTF8Array directly (which is what intArrayFromString does) and avoid the extra copy. The patch would look something like
```javascript
var ensureCache = {
  ...,
  alloc(array, view) {
    return this.allocBytes(array.length * view.BYTES_PER_ELEMENT);
  },
  allocBytes(len) {
    assert(ensureCache.buffer);
    len = alignMemory(len, 8); // keep things aligned to 8 byte boundaries
    var ret;
    if (ensureCache.pos + len >= ensureCache.size) {
      // we failed to allocate in the buffer, ensureCache time around :(
      assert(len > 0); // null terminator, at least
      ensureCache.needed += len;
      ret = Module['_webidl_malloc'](len);
      ensureCache.temps.push(ret);
    } else {
      // we can allocate in the buffer
      ret = ensureCache.buffer + ensureCache.pos;
      ensureCache.pos += len;
    }
    return ret;
  }
};

/** @suppress {duplicate} (TODO: avoid emitting this multiple times, it is redundant) */
function ensureString(value) {
  if (typeof value === 'string') {
    var len = lengthBytesUTF8(value) + 1;
    var offset = ensureCache.allocBytes(len);
    stringToUTF8Array(value, HEAP8, offset, len);
    return offset;
  }
  return value;
}

```

src/lib/libemval.js
slavek-kucera: As for tests, I think that `test/embind/test_i64_val.cpp` tests exactly this, but it also directly manipulates the HEAPxx arrays via an alias, so the code is not adjusted by the acorn transform and therefore the test is disabled at the moment on big endian.

juj: > worth extending to cover these cases too.

If there is something that doesn't yet/still work on bigendian with this PR, it is fine to not fix it here, but leave it for a later PR. This is a large PR already, so to ease development and allow a comfortable iteration workflow, it's ok to land this in the state it is, and look at improving the suite to cover more cases in a followup.

I ran the tests on bigendian0 suite on this PR to see that it is green, so that's a good first step.

src/lib/libemval.js
RReverser: I don't think we use anything like that. Extending typed arrays is very uncommon, but if we end up making one, we can add a prototype walk later.

tools/acorn-optimizer.mjs
RReverser: Proxies have significant performance cost, if you are referring to my earlier comment, I only meant using them for testing suite.

I believe merging these two passes should make this PR simpler / smaller rather than more complex, as you don't need to worry about these pass interactions.

src/lib/libemval.js
RReverser: > If there is something that doesn't yet/still work on bigendian with this PR, it is fine to not fix it here, but leave it for a later PR.

As far as I can tell, this is something that worked before this PR and gets broken by the change due to typos I commented on, we just don't have tests to catch this regression.

That's different from, say, leaving new functionality for later.

juj: If the typos only regress bigendian0, and those regressions are not showing up in the current test suite, then that's fine, and we can still land. That is because the big endian feature is a niche one.

Of course really great to point out the failures, and help slavek-kucera work on adding new tests and fixing those as well. That can be a separate PR still.

tools/webidl_binder.py
slavek-kucera: I tried including the change, but something isn't working so I backed it out as the current version is no worse than the original implementation.

src/lib/libemval.js
RReverser: Oh right, you can use `view[Symbol.toStringTag]` instead. That one is preserved even in subclasses and is meant for differentiating between built-in objects.

tools/webidl_binder.py
RReverser: > as the current version is no worse than the original implementation

Yeah that's fair, feel free to leave this one as future improvement.

src/lib/libemval.js
slavek-kucera: The typos are fixed. The affected code is only active with the big endian support and I don't see how this could have worked before on bigendian machines.

src/lib/libemval.js
slavek-kucera: It looks like you can overload that as well, the current solution works universally.

tools/acorn-optimizer.mjs
slavek-kucera: I fail to see the point of using proxies only in tests - that way tests pass and production fails.

I am definitely not merging those two passes in this PR.

slavek-kucera: As for the performance impact - are there any measurements available that would indicate how bad it is?
Because we are not talking about replacing just the typed array access, but really replacing function calls - whether it is the growable heap refresh, or the index checks in case of asan and safe heap.

src/lib/libemval.js
RReverser: You can, but only if someone intentionally wants their object to _not_ be treated as a type array.

That symbol intentionally exists for correct checking of type brands, something that even `instanceof` fails on when e.g. object comes from another worker or iframe.

And, either way, our views here are all known in advance since this intrinsic is only ever invoked on built-in `HEAP8`, `HEAP16` etc so there's no need to worry about edge cases like we're discussing, whereas code savings are pretty noticeable.

tools/acorn-optimizer.mjs
RReverser: > I fail to see the point of using proxies only in tests

When I mentioned using proxies, it was in context of simulating big-endian environment without a dedicated VM, not for implementation details.

> As for the performance impact - are there any measurements available that would indicate how bad it is?

You could benchmark it in this particular case, but in general from past work on JS engines the problem is that `Proxy` prevents _all_ JS optimisations as the engine can no longer operate on a statically known type and e.g. translate typed array access to a literal memory load/store, can no longer assume that index is integer (since proxy can only get it as a string and operate on it as such), can no longer treat methods as statically known and so on. It gets pretty bad and best avoided at all costs.

RReverser: > I am definitely not merging those two passes in this PR.

Why not? It's just about having a single handler for both passes in the same file, nobody's suggesting a huge refactoring, just simplify the current implementation.

slavek-kucera: > > I am definitely not merging those two passes in this PR.
> 
> Why not? It's just about having a single handler for both passes in the same file, nobody's suggesting a huge refactoring, just simplify the current implementation.

Because as far as I can tell none of the passes are currently parametrized. They are either called or not based on the flags.
The new combined pass would need to alter its behavior based on *somehow* propagated parameters, or there would need to be 3 new passes that would represent the three available combinations.
Either way this is completely unnecessary change to an already large PR and it is something that could be done separately at any time once the big-endian tests are available.

RReverser: I just think this deep check is just way too complex to maintain. If it's possible to make the passes _more_ independent instead, that would also work.

For example, what if we instead produced pattern like `(growMemViews(), nameXX[idx])` instead of `(growMemViews(), nameXX)[idx]`? That would mean checking for assignment access in growable pass separately, but it would make all other passes simpler since all of them would only need to match `nameXX[idx]` regardless of whether it's wrapped or not.

slavek-kucera: You would need to alter the `growableHeap` transform to now handle assignments differently.
The current version works, it has been tested.
If you want to further improve it it can be done in a follow up PR.

Just keep in mind that the current approach is flawed anyway. All the transforms should actually be performing deep checks: `(((HEAP32)))[x]` or `(((HEAP32)))[x] = val` should be transformed as well by littleEndianHeap, safeHeap and asanify transforms.

RReverser: As a reviewer, I mostly want to make sure that we don't add too much complexity _in this PR_. I'm happy to leave the final review / decision to @kripken though.

> All the transforms should actually be performing deep checks: `(((HEAP32)))[x]` or `(((HEAP32)))[x] = val` should be transformed as well by littleEndianHeap, safeHeap and asanify transforms.

Do we ever produce those variables in parentheses? If not, it seems somewhat out of scope of these static analysis passes, just like other edge cases like `var x = HEAP8; x[i] = val` and such.

slavek-kucera: The transforms are applied also on user-provided code. That's the reason why some of the tests need to be turned off, because they are accessing the heap via aliases or custom variables etc. and the transforms are not able to recognize the situation.
That is also one of the reasons why I think the proxy approach is very interesting, because that could actually work universally.

src/lib/libemval.js
slavek-kucera: This has been implemented.

tools/acorn-optimizer.mjs
RReverser: I think we could use Proxy for ASAN & safe heap, as those are expected to have slowdown anyway.

The growable heap should work fine with user code after #24295 as it will transform even `var x = HEAP8;` like any other access, and ensuring that heap is grown at that point should be enough.

The only real problem is the big-endian transform since it cares about specific type of access to swap bytes correctly on the way in and out, and since it's going to be used for production code, adding Proxy to such a hot path (every heap integer read/write from JS) would quickly add up.

RReverser: Although I guess you could make a mix of both. You could have something like

```js
var dv = new DataView(wasmMemory.buffer);
var HEAP32 = proxied(new Int32Array(wasmMemory.buffer));
```

and then transform would find all native usages like `HEAP32[x]` and turn them into `dv.getInt32(x)`, and anything not caught by the static transform would take the slower proxy path. Should get you a nice balance of performance and compatibility.

slavek-kucera: There is also the possibility to implement higher order operations more efficiently. Instead of e.g. copying elements element by element a bulk operation could be implemented with a single heap growth check, or single safe heap or asan check on the whole range. If the operation is between arrays of the same width, the operation could by implemented using byte-level operations etc.

RReverser: > Instead of e.g. copying elements element by element a bulk operation could be implemented with a single heap growth check

Oh it already does, as long as you use the `.set` methods (which is what should be used for efficient copying anyway).

kripken: I don't like this complexity, but I also don't have a better idea than anything that has already been proposed. I'm ok with it.

juj:
Fantastic - let's land this to open room for future work.

The big endian support will of course be on the periphery of Emscripten features, so it is not granted that it will stay without regressions - or that authors landing new work will know how to guarantee keeping it functional. But at least there is now the `bigendian0` suite that can be run, and emsdk integration to help go with it, so there is some coverage.

My CI runs the bigendian0 suite on x64 Linux, so I should see if regressions arise there.

juj:
@slavek-kucera I got the following output on the `bigendian0` suite, that might be interesting: http://clbri.com:8010/api/v2/logs/59822/raw_inline

PR #25048, opened 2025-08-25
Proxy process exit from pthread to main thread using a postMessage()

juj:
Proxy process exit from pthread to main thread using a postMessage() and not the WebAssembly heap-based proxying queue, so that all postMessage()s will flush in-order before the process exit command does.

Fixes `other.test_pthread_print_override_modularize` flakiness.

There the problem was that a pthread performs a print, which is proxied over to the main thread using a `print` callHandler.

Then pthread quits, which would result in `exitOnMainThread` being called, but that would utilize the WebAssembly heap based proxying queue.

On the main thread then depending on the phase of the moon, the main thread will either flush the Wasm heap queue or the postMessage() queue first.

By `postMessage()`ing also the process exit call, then we can ensure that all previous postMessage()s, such as the print() call, will be flushed sequentially before the process exit call is processed.

### DIFF
--- a/src/lib/libpthread.js
+++ b/src/lib/libpthread.js
@@ -75,6 +75,9 @@ var LibraryPThread = {
                    '$removeRunDependency',
 #endif
                    '$spawnThread',
+#if EXIT_RUNTIME
+                   '$exitOnMainThread',
+#endif
                    '_emscripten_thread_free_data',
                    'exit',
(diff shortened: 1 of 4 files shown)


juj:
Turns out this doesn't quite work out, because posixtest has tests that expect receiving a synchronous process exit message. Will have to think of something else.

PR #16050, opened 2022-01-17
file_packager.py: Add option to embed file data in wasm binary

sbc100:
This change not only adds the new option but uses this option
whenever `file_packager` is called from within emcc.

Very heavily inspired by the larger change in #16028.

Hopefully we can find a way deprecate and remove the old JS
embedding code since that seems strictly worse in every(?) way:

- Larger code size (JS base64 encoding is larger than binary)
- No possiblity of zero copy, memory-backed files
- Less compatible with standalone wasm / WASI

PR #25049, opened 2025-08-25
[file-packager] Drop support for --embed with JS output

sbc100:
The object file output is more efficient way of embedding and has been recommended via a warning since #16050 (early 2022).

With this simplification the file packager now has essentially two output modes:

1. Preloading via JS
2. Embedding via and object file

### DIFF
--- a/tools/file_packager.py
+++ b/tools/file_packager.py
@@ -67,7 +67,6 @@ Notes:
     subdir\file, in JS it will be subdir/file. For simplicity we treat the web platform as a *NIX.
 """
 
-import base64
 import ctypes
 import fnmatch
 import hashlib
@@ -136,11 +135,6 @@ options = Options()
 
 def err(*args):
   print(*args, file=sys.stderr)
-
-
-def base64_encode(b):
-  b64 = base64.b64encode(b)
-  return b64.decode('ascii')
 
 
 def has_hidden_attribute(filepath):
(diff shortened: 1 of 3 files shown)


tools/file_packager.py
```diff
@@ -455,9 +455,10 @@ def main():  # noqa: C901, PLR0912, PLR0915
   options.has_embedded = any(f.mode == 'embed' for f in data_files)
 
   if options.has_preloaded and options.has_embedded:
-    diagnostics.warn('support for using --preload and --embed in the same command is scheduled '
-        'for deprecation.  If you need this feature please comment at '
-        'https://github.com/emscripten-core/emscripten/issues/24803')
+    diagnostics.error('--preload and --embed now now mutually exclusive (See https://github.com/emscripten-core/emscripten/issues/24803')
```
```suggestion
    diagnostics.error('--preload and --embed are mutually exclusive (See https://github.com/emscripten-core/emscripten/issues/24803')
```

tools/file_packager.py
```diff
@@ -455,9 +455,10 @@ def main():  # noqa: C901, PLR0912, PLR0915
...
+
+  if options.has_embedded and not options.obj_output:
+    diagnostics.error('--obj-output is now required when using --embed.  This outputs an object file for linking directly into your application is more efficient than the old JS encoding')
```
```suggestion
    diagnostics.error('--obj-output is required when using --embed.  This outputs an object file for linking directly into your application which is more efficient than the old JS encoding')
```

tools/file_packager.py
kripken: (even a single "now" seems unneeded to me - we can just state the current rules, not a change in them)

kripken (approved):
Please add a changelog mention.

PR #25052, opened 2025-08-26
Use functools.lru_cache to implement memoize (NFC)

aheejin:
In Python 3.9 this can be replaced by `functools.cache`, but at least this can be a one-liner now.

### DIFF
--- a/system/lib/libc/musl/src/errno/__strerror.h
+++ b/system/lib/libc/musl/src/errno/__strerror.h
@@ -2,8 +2,12 @@
  * This file is included multiple times to declare and define a structure
  * with these messages, and then to define a lookup table translating
  * error codes to offsets of corresponding fields in the structure. */
-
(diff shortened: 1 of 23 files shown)


tools/utils.py
```diff
@@ -112,19 +112,8 @@ def delete_contents(dirname, exclude=None):
...
-  def helper(*args, **kwargs):
-    assert not kwargs
-    key = (func.__name__, args)
-    if key not in results:
-      results[key] = func(*args)
-    return results[key]
-
-  return helper
+# TODO(sbc): Replace with functools.cache, once we update to python 3.9
+memoize = lru_cache(maxsize=None)
```
sbc100: Can you use the full `functools.lru_cache` here since its only in one place

aheejin: Done

PR #25053, opened 2025-08-26
Implement --emit-symbol-map locally without calling out wasm-opt

sbc100:
The binaryen version was outputting strange mangled C++ names.

Doing this in python should be faster since we don't need to parse the whole binary.

### DIFF
--- a/test/other/test_symbol_map.O2.symbols
+++ b/test/other/test_symbol_map.O2.symbols
@@ -1,7 +1,7 @@
 0:run_js
 1:emscripten_asm_const_int
 2:__wasm_call_ctors
-3:foo::cpp_func\28int\29
+3:foo::cpp_func(int)
 4:middle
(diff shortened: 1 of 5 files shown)


sbc100:
Still needs a test..

kripken (approved):
Oh nice, yeah, this makes a lot of sense.

lgtm with a test as you mention.

Please also update the changelog about the escaping change.

kripken:
test lgtm

kripken:
Oh, but we still need an escaping test.

sbc100:
> Oh, but we still need an escaping test.

Yes, I mentioned that already in the comment  above(and in the PR description)

sbc100:
Added testing #25060.  Now you can see the fix in action.

kripken (approved):

PR #25055, opened 2025-08-26
Merge two similar features-section-parsing functions (NFC)

aheejin:
`webassembly.Module` has two similar functions. This makes code use `get_target_features`, which returns a map so it can be queried by features.

This also adds `@memoize` decorator to `get_target_features` so its result will be cached.


sbc100 (approved):
Nice!

PR #25057, opened 2025-08-26
Fix minimap selection extend limits in toolchain profiler view.

juj:


tools/toolchain_profiler.results_template.html
```diff
@@ -588,11 +588,12 @@
     .attr('d', function(d) { return d.path; });
 
   // draw the selection area
+  var selectionExtent = [
+         d3.min(items, function(d) { return d.start; }),
```
sbc100, line 592: `(d) => d.start`

sbc100:
I wonder if we can just load the profile data into devtools and use that UI instead of rolling out own?

juj:
Maybe.. I don't know enough of Devtools UI capabilities to say.

PR #25061, opened 2025-08-26
Restore test_unistd_write_broken_link test on Windows.

juj:
This now passes.


juj:
Actually, it turns out that this does still fail on Windows.

It was just that the test `test_unistd_write_broken_link` doesn't fail on Windows, but the variant `test_unistd_write_broken_link_rawfs` does. So I'll close this.

PR #25065, opened 2025-08-26
[testing] Disable background networking in chrome.

brendandahl:
This turns off a bunch of background downloading tasks and makes chrome run much more reliably for me.


brendandahl:
I should probably note this is what puppeteer does too.

sbc100 (approved):

kripken (approved):

PR #25067, opened 2025-08-27
Added minor docs for 64-bit conversions NFC

cwoffenden:
Minor clarification of the difference between `toIndexType()` and `to64()`.

### DIFF
--- a/src/parseTools.mjs
+++ b/src/parseTools.mjs
@@ -1003,7 +1003,9 @@ function toIndexType(x) {
 }
 
 // Converts a value to BigInt if building for wasm64, regardless of whether the
-// memory is 32- or 64-bit. Used for passing 64-bit values to the JS API.
+// memory is 32- or 64-bit. Used for passing pointer-width values to native
+// code (since pointers are presented as Number in JS and BigInt in wasm we need
+// this conversion before passing them).
 function to64(x) {
   if (!MEMORY64) return x;
   return `BigInt(${x})`;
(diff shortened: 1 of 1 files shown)


src/parseTools.mjs
```diff
@@ -995,11 +995,15 @@ function from64Expr(x) {
...
 
+// Converts a value to BigInt if building for wasm64, with both 64-bit pointers
+// and 64-bit memory. Used for indices into the memory tables, for example.
 function toIndexType(x) {
   if (MEMORY64 == 1) return `BigInt(${x})`;
   return x;
 }
 
+// Converts a value to BigInt if building for wasm64, regardless of whether the
+// memory is 32- or 64-bit. Used for passing 64-bit values to the JS API.
```
sbc100: I wonder if we should convert all the comments in this file to jsdoc?  (doesn't have to be part of this PR).

sbc100: How about "Used for passing pointer-width values to native code.   Since pointers are presented as Number in JS and BigInt in wasm we need to convert before passing them".

cwoffenden: > I wonder if we should convert all the comments in this file to jsdoc?

It's probably worth doing but there are 1000+ lines to scroll through.

I updated to comment with your suggestion.

sbc100 (approved):

cwoffenden:
@sbc100 or @juj could I prod one of you to land this docs-only change?

PR #25069, opened 2025-08-27
[testing] Use parallel test runner for browser tests.

brendandahl:
When using the parallel test runner each worker in the pool will get its own browser and server to run independently. Workers in the pool do not allow daemon processes so the HTTP servers have been converted to use threads.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -17,6 +17,7 @@ import io
 import itertools
 import logging
 import multiprocessing
+import threading
 import os
 import re
 import shlex
@@ -99,7 +100,7 @@ class ChromeConfig:
     # Disable various background tasks downloads (e.g. updates).
     '--disable-background-networking',
   )
-  headless_flags = '--headless=new --window-size=1024,768 --remote-debugging-port=1234'
+  headless_flags = '--headless=new --window-size=1024,768'
 
   @staticmethod
   def configure(data_dir):
@@ -2160,8 +2161,16 @@ class RunnerCore(unittest.TestCase, metaclass=RunnerMeta):
 # Run a server and a web page. When a test runs, we tell the server about it,
 # which tells the web page, which then opens a window with the test. Doing
 # it this way then allows the page to close() itself when done.
-def harness_server_func(in_queue, out_queue, port):
+def harness_server_func(in_queue, out_queue, port, base_path):
   class TestServerHandler(SimpleHTTPRequestHandler):
+    # Convert the path relative to what the base_path is set to.
+    # TODO is this still needed anymore? Tests seem to run fine without it.
+    def translate_path(self, path):
+      path = SimpleHTTPRequestHandler.translate_path(self, path)
+      relpath = os.path.relpath(path, os.getcwd())
+      fullpath = os.path.join(base_path, relpath)
+      return fullpath
+
     # Request header handler for default do_GET() path in
     # SimpleHTTPRequestHandler.do_GET(self) below.
     def send_head(self):
@@ -2236,6 +2245,7 @@ def harness_server_func(in_queue, out_queue, port):
         print(f'do_POST: unexpected POST: {urlinfo}')
 
     def do_GET(self):
+      nonlocal base_path
       info = urlparse(self.path)
       if info.path == '/run_harness':
         if DEBUG:
@@ -2263,7 +2273,7 @@ def harness_server_func(in_queue, out_queue, port):
         # the test is reporting its result. first change dir away from the
         # test dir, as it will be deleted now that the test is finishing, and
         # if we got a ping at that time, we'd return an error
-        os.chdir(path_from_root())
+        base_path = path_from_root()
         # for debugging, tests may encode the result and their own url (window.location) as result|url
         if '|' in self.path:
           path, url = self.path.split('|', 1)
@@ -2306,7 +2316,7 @@ def harness_server_func(in_queue, out_queue, port):
           # tell the browser to load the test
           self.wfile.write(b'COMMAND:' + url.encode('utf-8'))
           # move us to the right place to serve the files for the new test
-          os.chdir(dir)
+          base_path = dir
         else:
           # the browser must keep polling
           self.wfile.write(b'(wait)')
@@ -2342,8 +2352,27 @@ def harness_server_func(in_queue, out_queue, port):
   # do not have the correct MIME type
   SimpleHTTPRequestHandler.extensions_map['.mjs'] = 'text/javascript'
 
-  httpd = ThreadingHTTPServer(('localhost', port), TestServerHandler)
-  httpd.serve_forever() # test runner will kill us
+  return ThreadingHTTPServer(('localhost', port), TestServerHandler)
+
+
+class ServerThread(threading.Thread):
+  """A generic thread class to create and run a server."""
+  def __init__(self, server_creator, *server_args):
+    super().__init__()
+    self.server = None
+    self.server_creator = server_creator
+    self.server_args = server_args
+
+  def stop(self):
+    """Shuts down the server if it is running."""
+    if self.server:
+      self.server.shutdown()
+
+  def run(self):
+    """Creates the server instance and serves forever until stop() is called."""
+    self.server = self.server_creator(*self.server_args)
+    # Start the server's main loop (this blocks until shutdown() is called)
+    self.server.serve_forever()
 
 
 class Reporting(Enum):
@@ -2358,15 +2387,22 @@ class Reporting(Enum):
   FULL = 2
 
 
+def get_worker_id():
+  current = multiprocessing.current_process()
+  if current.name != "MainProcess":
+    # This relies on the fact that workers in the pool haven incremental names
+    # in the format of Thread-X.
+    worker_id_str = current.name.split('-')[-1]
+    return int(worker_id_str)
+  return 0
+
+
 class BrowserCore(RunnerCore):
   # note how many tests hang / do not send an output. if many of these
   # happen, likely something is broken and it is best to abort the test
   # suite early, as otherwise we will wait for the timeout on every
   # single test (hundreds of minutes)
   MAX_UNRESPONSIVE_TESTS = 10
-  PORT = 8888
-  SERVER_URL = f'http://localhost:{PORT}'
-  HARNESS_URL = f'{SERVER_URL}/run_harness'
   BROWSER_TIMEOUT = 60
 
   unresponsive_tests = 0
@@ -2405,6 +2441,9 @@ class BrowserCore(RunnerCore):
     if EMTEST_BROWSER_AUTO_CONFIG:
       logger.info('Using default CI configuration.')
       cls.browser_data_dir = DEFAULT_BROWSER_DATA_DIR
+      if cls.WORKER_ID:
+        # Running in parallel mode, give each browser its own profile dir.
+        cls.browser_data_dir += '-' + str(cls.WORKER_ID)
       if os.path.exists(cls.browser_data_dir):
         utils.delete_dir(cls.browser_data_dir)
       os.mkdir(cls.browser_data_dir)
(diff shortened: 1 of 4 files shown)


test/common.py
```diff
@@ -2160,8 +2161,16 @@ def get_zlib_library(self, cmake):
 # Run a server and a web page. When a test runs, we tell the server about it,
 # which tells the web page, which then opens a window with the test. Doing
 # it this way then allows the page to close() itself when done.
-def harness_server_func(in_queue, out_queue, port):
+def harness_server_func(in_queue, out_queue, port, base_path):
   class TestServerHandler(SimpleHTTPRequestHandler):
+    # Convert the path relative to what the base_path is set to.
+    # TODO is this still needed anymore? Tests seem to run fine without it.
```
kripken: I'm confused - the TODO suggests this is old code, which might not be needed. But this is new code adding in this PR? At least I can't see it as originating from any old code moved around in this diff.

brendandahl: I should have said this old behavior of switching dirs. Currently, the server does `chdir` a bunch switching between the root folder and the test directory. With the server being on a thread we can't do `chdir` since it will affect the process and tests. It looks like this change came in https://github.com/emscripten-core/emscripten/commit/c2530ebfbfc1342d81ac635b35adef91b8fd9d99

kripken: Hmm, interesting. Not sure what the reason for that was...

If tests run fine without this, I guess we can leave it out?

test/common.py
```diff
@@ -2236,6 +2245,7 @@ def do_POST(self):
         print(f'do_POST: unexpected POST: {urlinfo}')
 
     def do_GET(self):
+      nonlocal base_path
```
kripken: Hmm, is this actually going up and modifying the `base_path` param from above?

test/common.py
```diff
@@ -2306,7 +2316,7 @@ def do_GET(self):
           # tell the browser to load the test
           self.wfile.write(b'COMMAND:' + url.encode('utf-8'))
           # move us to the right place to serve the files for the new test
-          os.chdir(dir)
+          base_path = dir
```
sbc100: Can we land the change to stop using `chdir` separately perhaps?   It might be nice to review that part in isolation.

test/common.py
```diff
@@ -2433,23 +2472,32 @@ def browser_open(cls, url):
   @classmethod
   def setUpClass(cls):
     super().setUpClass()
+    # In parallel mode this ID contains which pool worker this process is.
+    cls.WORKER_ID = get_worker_id()
+    cls.PORT = 8888 + cls.WORKER_ID
+    cls.SERVER_URL = f'http://localhost:{cls.PORT}'
+    cls.HARNESS_URL = f'{cls.SERVER_URL}/run_harness'
```
sbc100, line 2586: Can this stuff be done at the class level instead?   

Also, we probably should be using ALL_CAPS for these.. although maybe we don't need to change that in this PR

test/parallel_testsuite.py
```diff
@@ -62,11 +67,18 @@ def run(self, result):
     # issues.
     # multiprocessing.set_start_method('spawn')
     tests = list(self.reversed_tests())
+    contains_browser_test = any(test.is_browser_test() for test in tests)
     use_cores = cap_max_workers_in_pool(min(self.max_cores, len(tests), num_cores()))
+    if contains_browser_test:
+      # TODO experiment with this number. In browser tests we'll be creating
+      # a chrome instance per worker which is expensive.
+      use_cores = int(use_cores / 2)
```
sbc100: It seems like this limit should not be applied the result of `num_cores` since it would be nice if `EMCC_CORES` or `EMTEST_CORES` or `-j` was honored precisely.  

So maybe this logic should go inside the `cap_max_workers_in_pool` function it should take a "is_browser" 4th argument?

test/test_browser.py
```diff
@@ -70,11 +70,7 @@ def do_GET(s):
         s.sendheaders([], length)
         s.wfile.write(data[start:end + 1])
 
-  # CORS preflight makes OPTIONS requests which we need to account for.
-  expectedConns = 22
-  httpd = HTTPServer(('localhost', 11111), ChunkedServerHandler)
-  for _ in range(expectedConns + 1):
-    httpd.handle_request()
```
sbc100: Is this just no needed anymore?  Maybe can be a separate PR too?

sbc100:
Can you explain more what you mean by "Workers in the pool do not allow daemon processes so the HTTP servers have been converted to use threads."?

Some more content in the PR description about exactly what is changing here would be good too think,

I'm very keep to see this land though!

test/common.py
brendandahl: The `chdir` change requires the threaded server part. I could land the threaded server part and removing the chdir.

test/common.py
brendandahl: These need to be assigned later than the class level otherwise they all end up with the same value of the initial forked process.

brendandahl:
> Can you explain more what you mean by "Workers in the pool do not allow daemon processes so the HTTP servers have been converted to use threads."?

Workers in the pool are already daemon processes and do not allow you to create sub-daemon processes.

brendandahl:
More on the daemonic stuff: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.daemon

test/common.py
sbc100: Why is that?  Does `get_worker_id()` not work at load time?

test/common.py
```diff
@@ -2358,15 +2387,22 @@ class Reporting(Enum):
   FULL = 2
 
 
+def get_worker_id():
+  current = multiprocessing.current_process()
+  if current.name != "MainProcess":
+    # This relies on the fact that workers in the pool haven incremental names
+    # in the format of Thread-X.
+    worker_id_str = current.name.split('-')[-1]
+    return int(worker_id_str)
```
sbc100: This seems pretty hacky... but maybe there is no better way?

test/common.py
```diff
@@ -2479,15 +2479,22 @@ class Reporting(Enum):
   FULL = 2
 
 
+def get_worker_id():
+  current = multiprocessing.current_process()
+  if current.name != "MainProcess":
```
sbc100: Use single quotes

test/common.py
```diff
@@ -2584,7 +2600,7 @@ def is_browser_test(self):
 
   def add_browser_reporting(self):
     contents = read_file(test_file('browser_reporting.js'))
-    contents = contents.replace('{{{REPORTING_URL}}}', BrowserCore.SERVER_URL)
+    contents = contents.replace('{{{REPORTING_URL}}}', self.SERVER_URL)
```
sbc100, line 2618: How is this different?

test/parallel_testsuite.py
```diff
@@ -102,6 +107,7 @@ def run(self, result):
       results = [pool.apply_async(run_test, (t, failfast_event, lock, progress_counter, len(tests))) for t in tests]
       results = [r.get() for r in results]
       results = [r for r in results if r is not None]
+      [pool.apply(tear_down, ()) for i in range(use_cores)]
```
sbc100: Is this supposed to ensure that every single worker runs `tear_down`?     How do we know one process don't end up being used for more than 1 of the N calls?

test/common.py
brendandahl: We no longer have a class variable on `BrowserCore`, it's on the cls object.

test/common.py
brendandahl: Class level code is not re-executed in each worker. It's run once in the main thread.

test/common.py
sbc100: I'm a bit confused though because this method is on the BrowserCore class so isn't the class object and BrowserCore the same thing here?   It is setUpClass called with the subclass of BrowserCore?

brendandahl: I misread what `cls` is in `BrowserCore.setUpClass` it's the child class `<class 'test_browser.browser'>`, not a `BrowserCore`.

I can change `setUpClass` to put the properties on `BrowserCore`, but then that seems like a bit of a strange place to modify the parent class.

sbc100: If self works here that seems fine

test/common.py
brendandahl: Added a less hacky way...

brendandahl:
Chatted with Sam, he's fine with landing as is.

juj:
I wonder if this might have regressed my CI: http://clbri.com:8010/api/v2/logs/46527/raw_inline

```
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/runner.py", line 603, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/runner.py", line 592, in main
    num_failures = run_tests(options, suites)
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/runner.py", line 434, in run_tests
    res = testRunner.run(suite)
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/unittest/runner.py", line 240, in run
    test(result)
    ~~~~^^^^^^^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/unittest/suite.py", line 84, in __call__
    return self.run(*args, **kwds)
           ~~~~~~~~^^^^^^^^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/parallel_testsuite.py", line 141, in run
    num_tear_downs = sum([pool.apply(tear_down, ()) for i in range(use_cores)])
                          ~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/multiprocessing/pool.py", line 360, in apply
    return self.apply_async(func, args, kwds).get()
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/multiprocessing/pool.py", line 458, in apply_async
    self._check_running()
    ~~~~~~~~~~~~~~~~~~~^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/multiprocessing/pool.py", line 353, in _check_running
    raise ValueError("Pool not running")
ValueError: Pool not running
program finished with exit code 1
```

test/parallel_testsuite.py
```diff
@@ -114,6 +135,12 @@ def run(self, result):
     # Filter out the None results which can occur in failfast mode.
     if self.failfast:
       results = [r for r in results if r is not None]
+      # Send a task to each worker to tear down the browser and server. This
+      # relies on the implementation detail in the worker pool that all workers
+      # are cycled through once.
+      num_tear_downs = sum([pool.apply(tear_down, ()) for i in range(use_cores)])
```
juj, line 141: Looks like this line crashes on me.

test/parallel_testsuite.py
```diff
@@ -94,10 +107,18 @@ def run(self, result):
...
     tests = list(self if self.failing_and_slow_first else self.reversed_tests())
-    use_cores = cap_max_workers_in_pool(min(self.max_cores, len(tests), num_cores()))
+    contains_browser_test = any(test.is_browser_test() for test in tests)
+    use_cores = cap_max_workers_in_pool(min(self.max_cores, len(tests), num_cores()), contains_browser_test)
     print('Using %s parallel test processes' % use_cores, file=sys.stderr)
     with multiprocessing.Manager() as manager:
-      with multiprocessing.Pool(use_cores) as pool:
+      # Give each worker a unique ID.
+      worker_id_counter = manager.Value('i', 0) # 'i' for integer, starting at 0
+      worker_id_lock = manager.Lock()
```
juj, line 116: I now realize that this PR added an unconditional use of `manager.Value()` and `manager.Lock()`, which are unsafe to use unless `python_multiprocessing_structures_are_buggy()` returns `False`.

So basically the `if python_multiprocessing_structures_are_buggy()` gate below is redundant/useless, since this code already has created a Lock and a Value.

brendandahl: Should we just disallow parallel runner with windows and affected versions of python?

sbc100: `python_multiprocessing_structures_are_buggy` is not just on windows is it?

If it is windows-only we should update `python_multiprocessing_structures_are_buggy`.

PR #18852, opened 2023-02-25
[Proxying] Send messages via in-memory mailbox queues

tlively:
Threads were previously notified of new work via postMessage messages that
carried pointers to the task queues to execute. There was no way to synchronously
pump or inspect these pending messages however, and there is no central registry
of all task queues for a thread, so this mechanism afforded no way to discover
or cancel pending work when a thread dies.

In preparation for implementing work cancellation, move the pending messages
into userspace by giving each thread a "mailbox", which is an `em_task_queue` in
the pthread struct. Instead of using `postMessage`, proxying queues now use the
thread mailbox API to notify threads of new work.

Internally, thread mailboxes still use postMessage to schedule work to be
executed when a thread returns to its event loop. Since the only task queues
involved in postMessages are now at known locations relative to the pthread
struct, there is no longer any need to store pointers to them in the postMessage
messages themselves. Removing these pointers works around tricky notification
and lifetime management edge cases that would have caused problems such as
dropped work or use-after-free bugs in future PRs.

When a thread dies because it exits or is canceled, it "closes" its mailbox by
decrementing a refcount and waiting to observe a refcount of 0. At this point,
the thread mailbox API ensures that no new messages will be enqueued on the
mailbox. Because the postMessage messages no longer contain task queue pointers,
it is safe to destroy the mailbox immediately after it is closed.

A user-visible behavior change this introduces is that proxied work is more
frequently completed _before_ a thread's main function begins running, since it
no longer gets ordered behind the `run` message in the JS postMessage queue. A
few tests are updated accordingly.

PR #18754, opened 2023-02-15
Avoid calling `process.exit` under node

sbc100:
Instead we always prefer to let exceptions bubble out and get handled by whatever handlers are installed.

Doing this did expose a few places where we were calling back into the runtime after it had exited.  This was previously being masked by the fact that we were bringing down the entire node process.  This means that when running tests under node we will now be a little more sensitive to use-after-exit bugs in our runtime and test code, but I would argue this is good thing.

This also means we no longer need to define logExceptionOnExit under node, which is a code size saving.

PR #25073, opened 2025-08-27
Fix race condition in checkMailbox

sbc100:
The `checkMailbox` callback can occur after the thread has terminated. In this case calling into native code can trigger the `makeAbortWrapper` wrapper that is put around each native function which then results in a "program has already aborted!" error being thrown.

Once solution to this is to make sure that the function which are called do not have `makeAbortWrapper` applied to them.

This was the technique I used in #18754, but the list of functions became stale when emscripten_proxy_execute_task_queue was removed in #18852.

A better solution is to wrap to whole function in callUserCallback, which takes case of checking if the runtime is alive before calling into native code.

Fixes: #20067


src/lib/libcore.js
```diff
@@ -2113,8 +2113,19 @@ addToLibrary({
   },
 
 #else // MINIMAL_RUNTIME
-  // MINIMAL_RUNTIME doesn't support the runtimeKeepalive stuff
-  $callUserCallback: (func) => func(),
+  $callUserCallback: (func) => {
+    // MINIMAL_RUNTIME doesn't support the runtimeKeepalive stuff, but under
+    // some circumstances it supportes `runtimeExited`
+#if EXIT_RUNTIME
+    if (runtimeExited) {
```
juj, line 2125: What would happen if we are building in `-sEXIT_RUNTIME=0` mode, and some code calls `abort()`, and then this async timeout for checkMailbox() triggers? Wouldn't that result in this safeguard passing right through into calling `_pthread_self()` again, since in `EXIT_RUNTIME=0` mode this check was compiled out altogether?

`_pthread_self()` is a C compiled function, so if the Worker is not hosting a pthread and hence doesn't have an active program stack, then even entering that function will not be safe (even though it might `return 0;`), since it could e.g. do a stack bump into corrupted space.

juj:
Thanks for looking into this. I'm happy with this as long as https://github.com/emscripten-core/emscripten/pull/25066 passes. Although if the checks are narrower, it might need to be expanded to cover both `-sEXIT_RUNTIME=0/1` options to make sure both cases are ok.

src/lib/libcore.js
sbc100: Yes, this could be a problem for `MINIMAL_RUNTIME`, since it doesn't track `ABORT` like the normal runtime does.

Do we have any other signal to know of the runtime is alive other than `ABORT`, `runtimeExited` and/or `pthread_self() == null`?   i.e. could we do any better?

I've love to consolidate those 3 myself into single "is_runtime_ok_or_valid_right_now" things, so adding yet another piece of state seems like the wrong direction.

juj: > Yes, this could be a problem for `MINIMAL_RUNTIME`, since it doesn't track `ABORT` like the normal runtime does.

I mean just in regular runtime, if one builds with `-sEXIT_RUNTIME=0`, then the above `if (runtimeExited)` build won't be compiled in, and if a pthread calls abort, then that would set `ABORT = true;` in the Worker and shut down its runtime? But this code wouldn't catch it?

sbc100: See the other implementation of `callUserCallback` for the regular runtime.  It does include that check.

juj: Gotcha, right - missed that this was the MINIMAL_RUNTIME path. That makes sense. LGTM.

juj (approved):

tlively (approved):

PR #25063, opened 2025-08-26
Use NODE_JS_TEST over NODE_JS in test code

sbc100:
See #20573

PR #25074, opened 2025-08-27
Fix JS engines to use NODE_JS_TEST

juj:
Emsdk generates

```
import os
emsdk_path = os.path.dirname(os.getenv('EM_CONFIG')).replace('\\', '/')
LLVM_ROOT = emsdk_path + '/llvm/git/build_main_64/bin'
NODE_JS = emsdk_path + '/node/22.16.0_64bit/bin/node'
EMSCRIPTEN_ROOT = emsdk_path + '/emscripten/main'
BINARYEN_ROOT = emsdk_path + '/binaryen/main_64bit_binaryen'
NODE_JS_TEST = ['qemu-s390x', '-L', '/usr/s390x-linux-gnu/', emsdk_path + '/node-big-endian-crosscompile/22.16.0_64bit/bin/node']
```
when activating regular and big endian node. This results in a non-working Emscripten installation that complains about NODE_JS not found.

Comment at https://github.com/emscripten-core/emscripten/pull/25063#issuecomment-3229281102 suggests that the intent is that if NODE_JS_TEST is populated, it should go to JS_ENGINES, so make that happen.

### DIFF
--- a/tools/config.py
+++ b/tools/config.py
@@ -64,7 +64,7 @@ def normalize_config_settings():
 
   # EM_CONFIG stuff
   if not JS_ENGINES:
-    JS_ENGINES = [NODE_JS]
+    JS_ENGINES = [NODE_JS_TEST if NODE_JS_TEST else NODE_JS]
 
   SPIDERMONKEY_ENGINE = fix_js_engine(SPIDERMONKEY_ENGINE, listify(SPIDERMONKEY_ENGINE))
   NODE_JS = fix_js_engine(NODE_JS, listify(NODE_JS))


juj:
Test failures seem to be flakes.

Running this PR in http://localhost:8010/#/builders/11/builds/287 does seem to work as expected.

tools/config.py
```diff
@@ -64,7 +64,7 @@ def normalize_config_settings():
 
   # EM_CONFIG stuff
   if not JS_ENGINES:
-    JS_ENGINES = [NODE_JS]
+    JS_ENGINES = [NODE_JS_TEST if NODE_JS_TEST else NODE_JS]
```
sbc100, line 67: Can you move these 2 lines to `test/common.py` after the assignment to `NODE_JS_TEST` (then you don't need the embedded if/else here).

juj: check

juj: There's some kind of type analysis checker?

`mypy` step fails with
```
test/common.py:142: error: List item 0 has incompatible type "Optional[Any]"; expected "List[str]"  [list-item]
Found 1 error in 1 file (checked 130 source files)
```

juj: Not sure if I did the change something wrong, though after moving the two lines, it seems the effect is as if I didn't have those lines at all, i.e. the run fails like it did before this PR altogether. (http://localhost:8010/api/v2/logs/33118/raw_inline)

juj: Oh I see, I still need the ternary, I misunderstood the move was intended to avoid that.

juj: > (then you don't need the embedded if/else here).

Hmm, it looks like I do, because in common.py, the check is

```
if not config.NODE_JS_TEST:
  config.NODE_JS_TEST = config.NODE_JS
```
but in emsdk case, `config.NODE_JS_TEST` *is* set, it has both. So this check doesn't do anything.

So it needs 

```
if not config.JS_ENGINES:
  config.JS_ENGINES = [config.NODE_JS_TEST if config.NODE_JS_TEST else config.NODE_JS]
```

juj: I reverted back to the original form, that helped the mypy type check, that insisted thinking that `config.NODE_JS` is `Optional[Any]` and not `List[Str]`

Issue #25075, opened 2025-08-27
wasm64.test_safe_stack failure


juj:
Running test `wasm64.test_safe_stack` on Node.js Nightly, I get

```
======================================================================
FAIL: test_safe_stack (test_core.wasm64.test_safe_stack)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.12/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/usr/lib/python3.12/unittest/case.py", line 634, in run
    self._callTestMethod(testMethod)
  File "/usr/lib/python3.12/unittest/case.py", line 589, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/test_core.py", line 9162, in test_safe_stack
    self.do_runf('core/test_safe_stack.c',
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 2013, in do_runf
    return self._build_and_run(filename, expected_output, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 2073, in _build_and_run
    self.assertContained(o, js_output, regex=regex)
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 1719, in assertContained
    self.assertTrue(re.search(values, string, re.DOTALL), 'Expected regex "%s" to match on:\n%s' % (values, string))
  File "/usr/lib/python3.12/unittest/case.py", line 727, in assertTrue
    raise self.failureException(msg)
AssertionError: None is not true : Expected regex "Aborted\(stack overflow \(Attempt to set SP to 0x[0-9a-fA-F]+, with stack limits \[0x[0-9a-fA-F]+ - 0x[0-9a-fA-F]+\]\)" to match on:
Aborted(stack overflow (Attempt to set SP to 0x00000-30, with stack limits [0x00000000 - 0x00000400]). If you require more stack space build with -sSTACK_SIZE=<bytes>)
[...]

PR #25079, opened 2025-08-27
Fix wasm64.test_stack_safe. NFC

sbc100:
This change fixes ptrToString to handle 64-bit pointers, as well as negative pointers.

For wasm64 have to deal with potentially negative pointer values flowing out of WebAssembly.  The `>>> 0` trick doesn't work since it would truncate the value to 32-bits.

Fixes: #25075

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -659,27 +659,6 @@ jobs:
     steps:
       - run-tests-linux:
           test_targets: "wasm2js1"
-  test-wasm64:
-    # We don't use `bionic` here since its too old to run recent node versions:
-    # `/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found`
-    executor: linux-python
-    steps:
-      - prepare-for-tests
-      # The linux-python image uses /home/circleci rather than /root and jsvu
-      # hardcodes /root into its launcher scripts so we need to reinstall v8.
-      - run: rm -rf $HOME/.jsvu
-      - install-v8
-      - install-node-canary
-      # When running wasm64 tests we need to make sure we use the testing
-      # version of node (node canary) when running the compiler output (e.g.
-      # in configure tests.
-      - run:
-          name: configure compiler to use node-canary
-          command: echo "NODE_JS = NODE_JS_TEST" >> ~/emsdk/.emscripten
-      - run-tests:
-          title: "wasm64"
-          test_targets: "wasm64"
-      - upload-test-results
   test-wasm64-4gb:
     environment:
       LANG: "C.UTF-8"
@@ -707,6 +686,7 @@ jobs:
             core_2gb.test_*em_asm*
             core_2gb.test_*embind*
             core_2gb.test_fs_js_api_wasmfs
+            wasm64.test_safe_stack
             wasm64l.test_hello_world
             wasm64l.test_bigswitch
             wasm64l.test_module_wasm_memory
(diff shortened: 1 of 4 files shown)


juj:
When reading pointer values, I always felt it unnatural to read `-0x1` for example.. `-0xFFFFFFFFFFFFFFFF` would be my style.. (given that pointers are always unsigned) but maybe that's just me.

LGTM either way.

sbc100:
> When reading pointer values, I always felt it unnatural to read `-0x1` for example.. `-0xFFFFFFFFFFFFFFFF` would be my style.. (given that pointers are always unsigned) but maybe that's just me.
> 
> LGTM either way.

Can you think of a nice way to convert negative JS number to unsigned i64 values?     The `>>> 0` trick is great for i32 but I don't know of an i64 trick.

I guess one way would use typesarrays and write as signed by read as unsigned?  Seems like more code than we would want here though.

juj:
> Can you think of a nice way to convert negative JS number to unsigned i64 values?

```
var ptrize = (x) => x < 0 ? 2n**64n + x : x; 
```
comes to mind - maybe that could work?

sbc100:
> > Can you think of a nice way to convert negative JS number to unsigned i64 values?
> 
> ```
> var ptrize = (x) => x < 0 ? 2n**64n + x : x; 
> ```
> 
> comes to mind - maybe that could work?

Nice, I took that idea.

sbc100:
We now see:

```
RuntimeError: Aborted(stack overflow (Attempt to set SP to 0xffffffffffffffd0, with stack limits [0x0000000000000000 - 0x0000000000000400]). If you require more stack space build with -sSTACK_SIZE=<bytes>)
```

Which matches the 32-bit version:
```
RuntimeError: Aborted(stack overflow (Attempt to set SP to 0xffffffd0, with stack limits [0x00000000 - 0x00000400]). If you require more stack space build with -sSTACK_SIZE=<bytes>)
```

.circleci/config.yml
```diff
@@ -659,27 +659,6 @@ jobs:
...
-      # When running wasm64 tests we need to make sure we use the testing
-      # version of node (node canary) when running the compiler output (e.g.
-      # in configure tests.
-      - run:
-          name: configure compiler to use node-canary
-          command: echo "NODE_JS = NODE_JS_TEST" >> ~/emsdk/.emscripten
-      - run-tests:
-          title: "wasm64"
-          test_targets: "wasm64"
-      - upload-test-results
```
kripken, line 682: Is this intentional?

src/lib/libcore.js
```diff
@@ -45,11 +45,15 @@ addToLibrary({
 #if ASSERTIONS
     assert(typeof ptr === 'number');
 #endif
-#if !CAN_ADDRESS_2GB && !MEMORY64
-    // With CAN_ADDRESS_2GB or MEMORY64, pointers are already unsigned.
+#if MEMORY64
+    // Convert to 64-bit unsigned value.  We need to use BigInt here since
+    // Number cannot represent the full 64-bit range.
```
kripken, line 50: Isn't Number ok because this is a pointer, so it will only use part of the full 64 bits, and that range is in Number?

.circleci/config.yml
sbc100: Yes, this whole block was unused.   We can bring it back if ever need it again.

src/lib/libcore.js
sbc100: Right but in this case the number is small negative value (since the SP went below zero).

We can try to show this as a native value (like the first version of this patch) but a large unsigned 64-bit value seems better and more consistent with the 32-bit case.

We show `0xffffffd0` for 32-bit and `0xffffffffffffffd0` for 64-bit.  But `0xffffffffffffffd0` is not representable with a 32-bit Number.

PR #25080, opened 2025-08-27
Avoid throwing of plain JS strings. NFC

sbc100:
Replaced with `assert`, `abort` or `throw new Error` accordingly.  We almost never want to be throwing plain values because we don't end up with a stack trace in that case.


brendandahl:
What's the criteria for choosing `throw new Error` vs `abort`?

sbc100:
> What's the criteria for choosing `throw new Error` vs `abort`?

I tried to use `abort` whenever I know that the code is running in the context of emscripten-generated code, and after the definition of `abort` itself (which I think happens in shell.js).

sbc100:
> > What's the criteria for choosing `throw new Error` vs `abort`?
> 
> I tried to use `abort` whenever I know that the code is running in the context of emscripten-generated code, and after the definition of `abort` itself (which I think happens in shell.js).

But I now that I wasn't very consistent :)

sbc100:
Ok, should be more consistent now.  I've used `abort` whereever I'm sure its ok to us it.

src/lib/libfs.js
```diff
@@ -1780,7 +1780,7 @@ FS.staticInit();`;
             }
 
             xhr.send(null);
-            if (!(xhr.status >= 200 && xhr.status < 300 || xhr.status === 304)) throw new Error("Couldn't load " + url + ". Status: " + xhr.status);
+            if (!(xhr.status >= 200 && xhr.status < 300 || xhr.status === 304)) abort("Couldn't load " + url + ". Status: " + xhr.status);
```
RReverser, line 1783: You mentioned in the PR description you used `assert` too, but looks like quite a few of conditional aborts like this one could still be replaced with assert.

Just for clarity of review, did these get missed or is it something you want to leave out of scope of this PR?

sbc100: The problem with assert is it that its only available in debug builds (or at least that is the idea).

sbc100: If there are `abort`s list this that are already inside `#if ASSERTIONS` then yes the should be converted to `assert`.    Did you spot any?

RReverser: Ah ok.

It's something I was wondering about from time to time - currently each assert is always wrapped in manual `#if ASSERTIONS`...

If it's meant to be only available in debug builds, maybe it's worth making it conditional in acorn-optimizer to avoid repetitions, or, vice versa, if we want to explicitly wrap only certain assertions in `#if`, then maybe the function definition should always be available so we could use it in all these other places too.

sbc100: The `assert` function is, in theory, only defined when `ASSERTIONS` setting set.  However, in practice this is only true in `MINIMAL_RUNTIME` or `STRICT` modes today, because historically we have not been great a wrapping the calls in `#if ASSERTIONS` in all places.

Regarding using acorn passed instead of our preprocessor shit, that is something I've looked into in the past but haven't managed to land yet.  My latest attempt was here: https://github.com/emscripten-core/emscripten/pull/22166.

sbc100:
@brendandahl ptal

PR #25081, opened 2025-08-27
When printing test harness stuff, print to stderr to be more consistent.

juj:
Iirc the original rationale for the test harness was that test harness infra related prints would go to stderr, and tests itself would print to stdout? (at least that is how I gathered it to be from the start of time..)

Most of the harness stuff goes to stderr, though a few stray lines do not. So move those to print to stderr also to be more consistent.

When observing the prints on Buildbot, this distinction helps a bit visually, since stderr is printed red, and individual test output then in gray.


sbc100 (commented):
LGTM.

I'm working on another change that will buffer all stdout/stderr from individual tests and only show it when a test fails. (i.e. when no tests fail you don't see any noise).

If we can make this mode the default then tests and be a noisy as they want..

sbc100 (approved):

PR #25082, opened 2025-08-27
Add support for drawing parallelization swimlane graphs of the multithreaded suite run.

juj:
Add support for drawing parallelization swimlane graphs of the multithreaded suite run.

Running
```
EMTEST_VISUALIZE=1 test/runner core0
./emprofile
```

will generate a visual swimlane chart in html form to show how well the suite parallelized.

Example graphs:

https://clbri.com/dump/emscripten/toolchain_profiler.results_20250827_1943.html

This helps visualize at a glance how well parallelization took place, but also to eyeball how many tests passed/failed/flaked (green, red, orange).

The code is a bit WIP, I'll have to clean it up a bit, but putting this up for looks to show the idea already. I'm looking about making my CI generate these graphs from all steps, so that slow tests can be identified at a glance.

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -205,7 +205,7 @@ class BufferedParallelTestResult:
   def test_short_name(self):
     # Given a test name e.g. "test_atomic_cxx (test_core.core0.test_atomic_cxx)"
     # returns a short form "test_atomic_cxx" of the test.
-    return self.test_name[:self.test_name.find(' ')]
+    return self.test_name.split(' ', 1)[0]
 
   def addDuration(self, test, elapsed):
     self.test_duration = elapsed


test/parallel_testsuite.py
```diff
@@ -76,6 +76,12 @@ def run(self, result):
     # issues.
     # multiprocessing.set_start_method('spawn')
 
+    # Remove old list of flaky tests before starting the run
+    if os.getenv('EMTEST_VISUALIZE'):
+      try:
+        os.remove(os.path.join(tempfile.gettempdir(), 'emscripten_flaky_tests'))
+      except:
+        pass
```
sbc100: You can use the `delete_file` helper utility here instead of try/catch maybe?

sbc100 (commented):
Nice idea!

test/parallel_testsuite.py
```diff
@@ -212,7 +212,9 @@ def log_test_run_for_visualization(self):
       # Write profiling entries for emprofile.py tool to visualize. This needs a unique identifier for each
       # block, so generate one on the fly.
       dummy_test_task_counter = os.path.getsize(profiler_log_file) if os.path.isfile(profiler_log_file) else 0
-      prof.write(f',\n{{"pid":{dummy_test_task_counter},"op":"start","time":{self.start_time},"cmdLine":["{self.test_short_name()}"],"color":"{colors[self.test_result]}"}}')
+      # Remove the redundant 'test_' suffix from each test, since character space is at a premium in the visualized graph.
```
sbc100: suffix -> prefix?

test/common.py
```diff
@@ -83,6 +83,11 @@
 if 'EM_BUILD_VERBOSE' in os.environ:
   exit_with_error('EM_BUILD_VERBOSE has been renamed to EMTEST_BUILD_VERBOSE')
 
+# If we are drawing a parallel swimlane graph of test output, we need to use a temp
+# file to track which tests were flaky so they can be graphed in orange color to
+# visually stand out.
+EMTEST_FLAKY_TEST_LOG_FILE = os.path.join(tempfile.gettempdir(), 'emscripten_flaky_tests') if os.getenv('EMTEST_VISUALIZE') else None
```
sbc100: No need for `EMTEST_` prefix here.

Also, since you aren't actually opening the file here maybe just `flaky_test_log_file = os.path.join(tempfile.gettempdir(), 'emscripten_flaky_tests')`  (i'm not a fan on the single line if/else syntax and it seems like we can avoid it here.)

Also, maybe give the file the `.txt` externsion?

Also, why not just write to `out/flaky_texts.txt`?  (why use tmpdir at all?)

test/common.py
```diff
@@ -231,6 +236,10 @@ def modified(*args, **kwargs):
         except AssertionError as exc:
           preserved_exc = exc
           logging.info(f'Retrying flaky test "{f.__name__}" (attempt {i}/{EMTEST_RETRY_FLAKY} failed): {exc}')
+          # Mark down that this was a flaky test.
+          if EMTEST_FLAKY_TEST_LOG_FILE:
+            open(EMTEST_FLAKY_TEST_LOG_FILE, 'a').write(f'{f.__name__}\n')
```
sbc100: Maybe just do this unconditionally?  It seems very cheap and useful for debugging.

test/parallel_testsuite.py
```diff
@@ -155,6 +176,11 @@ def __init__(self):
   def test(self):
     return self.buffered_result.test
 
+  def test_short_name(self):
+    # Given a test name e.g. "test_atomic_cxx (test_core.core0.test_atomic_cxx)"
+    # returns a short form "test_atomic_cxx" of the test.
+    return self.test_name[:self.test_name.find(' ')]
```
sbc100: `self.test_name.split(' ', 1)[0]`

test/parallel_testsuite.py
```diff
@@ -135,8 +141,23 @@ def combine_results(self, result, buffered_results):
     # parallel causes mis-orderings, this makes the results more readable.
     results = sorted(buffered_results, key=lambda res: str(res.test))
     result.core_time = 0
+
+    flaky_tests = open(common.EMTEST_FLAKY_TEST_LOG_FILE).read().split() if os.path.isfile(common.EMTEST_FLAKY_TEST_LOG_FILE) else []
```
sbc100: If the file is only persisting until the end of the test run can we not just use an in-memory list to store the list of flaky tests?

test/common.py

test/common.py
juj: ok, changed to `out/flaky_tests.txt`

test/parallel_testsuite.py
juj: That is what I had at first, but it algorithmically overworks, and looks clumsy with a constant `1` in it.

test/parallel_testsuite.py
juj: You are free to figure out how to do that. I agree it would be nicer, and spent about two hours trying to make that happen through multiple module imports and parallel python multiprocessing, and the best I got was a wall of pickling errors from trying to chain the data structure throughout the test runner.

sbc100: I see so it comes down to "shared data structures are hard in the python multi-processing world"?   Maybe add comment saying that.

juj: In order to make that work, IIUC one needs to create a shared string/string list variable using the `monitor` object created in the parallel test runner. Similar to how an integer is created here:

https://github.com/emscripten-core/emscripten/pull/25099/files#diff-fd5a5983b044aa841287bf197f2b5e46781162bf0e6a0ca8d1ccaddf1cc1c969R88

and then route that variable (plus an associated lock) over to `test/common.py` code to access in the flaky retrying code. 

The variable needs to pass through the member functions that lead to running the individual tests, something similar to how this variable routing takes place:

https://github.com/emscripten-core/emscripten/pull/25099/files#diff-fd5a5983b044aa841287bf197f2b5e46781162bf0e6a0ca8d1ccaddf1cc1c969R155-R157

And the code structure ends up reading more messy than what routing via files is. Since flaky tests occur very rarely during a test run, the files are not in a hot path in general.

juj:
Ping, any more thoughts here?

test/parallel_testsuite.py
```diff
@@ -85,6 +87,9 @@ def run(self, result):
     # issues.
     # multiprocessing.set_start_method('spawn')
 
+    # Remove any old stale list of flaky tests before starting the run
+    utils.delete_file(common.flaky_tests_log_filename)
```
sbc100, line 91: Shouldn't this apply to regular, non-parallel test runner too?  (You can always use `-j1` to trigger the use of the non-parallel runner with any suite).

test/parallel_testsuite.py
```diff
@@ -150,6 +155,10 @@ def combine_results(self, result, buffered_results):
     results = sorted(buffered_results, key=lambda res: str(res.test))
     result.core_time = 0
 
+    # shared data structures are hard in the python multi-processing world, so
+    # use a file to share the flaky test information across test processes.
+    flaky_tests = open(common.flaky_tests_log_filename).read().split() if os.path.isfile(common.flaky_tests_log_filename) else []
```
sbc100, line 160: I think `.splitlines()` would make sense here.

test/parallel_testsuite.py
```diff
@@ -192,6 +218,30 @@ def updateResult(self, result):
...
     result.stopTest(self.test)
     result.core_time += self.test_duration
+    self.log_test_run_for_visualization()
+
+  def log_test_run_for_visualization(self):
+    if os.getenv('EMTEST_VISUALIZE') and (self.test_result != 'skipped' or self.test_duration > 0.2):
+      profiler_logs_path = os.path.join(tempfile.gettempdir(), 'emscripten_toolchain_profiler_logs')
+      os.makedirs(profiler_logs_path, exist_ok=True)
+      profiler_log_file = os.path.join(profiler_logs_path, 'toolchain_profiler.pid_0.json')
+      prof = open(profiler_log_file, 'a')
```
sbc100: `with prof as open(profiler_log_file, 'a'):`

sbc100: Looks like it can move down to line 243 too

test/parallel_testsuite.py
sbc100: Actually you can just do `.readlines()` in a single call.

test/parallel_testsuite.py
```diff
@@ -192,6 +218,30 @@ def updateResult(self, result):
     self.buffered_result.updateResult(result)
     result.stopTest(self.test)
     result.core_time += self.test_duration
+    self.log_test_run_for_visualization()
+
+  def log_test_run_for_visualization(self):
+    if os.getenv('EMTEST_VISUALIZE') and (self.test_result != 'skipped' or self.test_duration > 0.2):
+      profiler_logs_path = os.path.join(tempfile.gettempdir(), 'emscripten_toolchain_profiler_logs')
+      os.makedirs(profiler_logs_path, exist_ok=True)
+      profiler_log_file = os.path.join(profiler_logs_path, 'toolchain_profiler.pid_0.json')
```
sbc100, line 227: What is the hardcoded `pid_0` here?   Should this be one-file-per-process?

test/parallel_testsuite.py
juj: `.readlines()` left behind a `\r` character at the end of each line for whatever reason.. that is what I had there first.

juj: err not the \r character but \n character.

test/parallel_testsuite.py
juj: No, because the parallel test harness has already joined, and it is only the very last combine printing that does this. (this has then the benefit of not slowing down the actual test run times)

test/parallel_testsuite.py
juj: check

test/parallel_testsuite.py
juj: Drawing parallel swimlane graphs out of a single-threaded test run is not particularly exciting. Adding support to the st test runner could be something for another PR, but I don't see anyone actually using it.

juj:
Any further thoughts here?

test/parallel_testsuite.py
sbc100: `readlines()` is supposed to work with text files so I'm somewhat confused but this... but not a big deal I guess

juj: image

sbc100: I yes I forgot you need to call `strip()` on each line.

Maybe make this into multiple lines:

```
flaky_tests = []
if os.path.isfile(common.flaky_tests_log_filename): 
  flaky_tests = [line.strip() for line in open(common.flaky_tests_log_filename).readlines()]
```

I've never likes the single line if/else syntax anyway.

test/parallel_testsuite.py
sbc100: Still, I think `self.test_name.split(' ', 1)[0]` i more pythonic, at least to my eye.

However I noticed while reviewing https://github.com/emscripten-core/emscripten/pull/25069 that you can maybe do `test._testMethodName` instead?

Or I noticed there is the alternative `test.id()` which returns `"%s.%s" % (strclass(self.__class__), self._testMethodName)`

test/parallel_testsuite.py
juj: Well, that's just worse than `open().read().split()` :/

test/parallel_testsuite.py
```diff
@@ -159,7 +168,19 @@ def combine_results(self, result, buffered_results):
       os.set_blocking(sys.stderr.fileno(), True)
 
     for r in results:
+      # Merge information of flaky tests into the test result
+      if r.test_result == 'success' and r.test_short_name() in flaky_tests:
+        r.test_result = 'warnings'
```
sbc100, line 173: Is this "warnings" test result only surfaced via the visualization, or does it show up on the command line output too?

juj: It only shows up in visualization. Python unit test runner does not have a concept of test finishing with warnings to my knowledge.

test/parallel_testsuite.py
juj: ok, fine.. I don't think it's particularly improving, but to get the review forward, updated.

PR #25093, opened 2025-08-28
Move preInit handling to processModuleArgs. NFC

sbc100:
This simplifies the startup code and paves the way for the removal of the use of `addRunDependency`/`removeRunDependency` for `wasm-instantiation`.

### DIFF
--- a/src/postamble.js
+++ b/src/postamble.js
@@ -288,20 +288,6 @@ function checkUnflushedContent() {
 #endif // EXIT_RUNTIME
 #endif // ASSERTIONS
 
-function preInit() {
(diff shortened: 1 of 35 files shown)


src/postamble.js
```diff
@@ -378,13 +363,12 @@ createWasm();
 wasmExports = createWasm();
 #endif
 
+run();
```
kripken, line 366: Previously this ran only on the main thread (old line 386). How does the new code handle that?

kripken: Oh, I misread the nested ifdefs. (Wish we had more comments on them...)

lgtm

PR #25096, opened 2025-08-28
Remove the `wasm-instantate` run dependency. NFC

sbc100:
Instead we can just delay the `run` call until after the async createWasm function returns/resolves.

This results in a nice code size saving.

### DIFF
--- a/src/postamble.js
+++ b/src/postamble.js
@@ -353,17 +353,18 @@ if ({{{ ENVIRONMENT_IS_MAIN_THREAD() }}}) {
 // In modularize mode the generated code is within a factory function so we
 // can use await here (since it's not top-level-await).
 wasmExports = await createWasm();
+run();
 #else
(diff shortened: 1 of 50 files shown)


src/postamble.js
```diff
@@ -353,17 +353,18 @@ if ({{{ ENVIRONMENT_IS_MAIN_THREAD() }}}) {
 // In modularize mode the generated code is within a factory function so we
 // can use await here (since it's not top-level-await).
 wasmExports = await createWasm();
+run();
 #else
 // With async instantation wasmExports is assigned asynchronously when the
 // instance is received.
-createWasm();
+createWasm().then(() => run());
```
kripken, line 360: IIUC this will delay preRun (including preloading data files) until the wasm is ready, but I think before we could do them in parallel?

sbc100: Maybe, but we already that just above where we `await createWasm`.  I will check..

PR #25098, opened 2025-08-29
Add a --bell option to parallel test runner to play a beep at the end of the test suite run.

juj:
So that I can watch Columbo without a care in the world while the tests are running.

### DIFF
--- a/test/runner.py
+++ b/test/runner.py
@@ -450,6 +450,10 @@ def run_tests(options, suites):
     for msg in resultMessages:
       print('    ' + msg)
 
+  if options.bell:
+    sys.stdout.write('\a')
+    sys.stdout.flush()
+
   return num_failures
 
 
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
This seems like it applies to all test runners, not just the parallel one?

Does it work on windows too?

ltgm, although seems like something you could just with with normal shell command chaining.  e.g. `./do_somerthing; echo "DONE'; ./play_sound`

juj:
> ltgm, although seems like something you could just with with normal shell command chaining. e.g. `./do_something; echo "DONE'; ./play_sound`

Yes. I can create a script to do that, but it's troublesome to have to create a .bat script to be able to get a bell.

This is because I was never able to make shell chaining work. The only way was to use `&&`, but that requires that the suite actually passes in the end.

image

i.e. `;` confuses the runner itself

`--bell` does work on Windows.

sbc100:
`;` I guess is not a windows thing but it certainly works on bash right?

What is the equivalent on windows, maybe none?

sbc100:
Oh it looks like windows using single & for chaining:  `command1 & command2 & command3`

sbc100:
Just for future reference, I'm find with having this extra option if its useful to you.

juj:
> Oh it looks like windows using single & for chaining: `command1 & command2 & command3`

Good point, I always thought it was `;` too for Windows. TIL.

PR #25099, opened 2025-08-29
Implement a progress counter to the parallel test suite runner.

juj:
To give a sense of where things are at:

image

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -183,7 +183,7 @@ class BufferedParallelTestResult:
 
   def compute_progress(self):
     with self.lock:
-      val = f'\033[96m[{int(self.progress_counter.value * 100 / self.num_tests)}%]\033[0m'
+      val = f'[{int(self.progress_counter.value * 100 / self.num_tests)}%]'
       self.progress_counter.value += 1
     return val
 
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -183,7 +183,7 @@ class BufferedParallelTestResult:
 
   def compute_progress(self):
     with self.lock:
-      val = f'[{int(self.progress_counter.value * 100 / self.num_tests)}%]'
+      val = f'\033[96m[{int(self.progress_counter.value * 100 / self.num_tests)}%]\033[0m'
       self.progress_counter.value += 1
     return val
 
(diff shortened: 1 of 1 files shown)


test/parallel_testsuite.py
```diff
@@ -176,33 +181,39 @@ def stopTest(self, test):
     # these results get passed back to the TextTestRunner/TextTestResult.
     self.buffered_result.duration = self.test_duration
 
+  def compute_progress(self):
+    with self.lock:
+      val = f'\033[96m[{int(self.progress_counter.value * 100 / self.num_tests)}%]\033[0m'
```
sbc100: In the other places in the codebase where we do console coloring we have different paths for linux and windows.

See `tools/colored_logger.py` and `tools/diagnostics.py`.   These two places already duplicate this stuff so it might be nice not to add a third.

juj: I am not sure I see the duplication. The other places are about adjusting logging `.info`, `.warning`, `.error` colors, but this is just making sure that the percentage value gets colors.

Anyhow, I'll drop the colors to avoid this overhead.

sbc100: I `diagnotics.py` we have:

```
def output_color_windows(color):                                                                         
  assert not force_ansi                                                                                  
  # TODO(sbc): This code is duplicated in colored_logger.py.  Refactor.                                  
  # wincon.h                                                                                             
  FOREGROUND_BLACK     = 0x0000 # noqa                                                                   
  FOREGROUND_BLUE      = 0x0001 # noqa                                                                   
  FOREGROUND_GREEN     = 0x0002 # noqa                                                                   
  FOREGROUND_CYAN      = 0x0003 # noqa                                                                   
  FOREGROUND_RED       = 0x0004 # noqa                                                                   
  FOREGROUND_MAGENTA   = 0x0005 # noqa                                                                   
  FOREGROUND_YELLOW    = 0x0006 # noqa                                                                   
  FOREGROUND_GREY      = 0x0007 # noqa       
```

Same duplication for ANSI I think.  `add_coloring_to_emit_ansi` duplicated the ansi color codes from diagnositcs:

```
# available colors                                                                                       
RED = 1                                                                                                  
GREEN = 2                                                                                                
YELLOW = 3                                                                                               
BLUE = 4                                                                                                 
MAGENTA = 5                                                                                              
CYAN = 6                                                                                                 
WHITE = 7    
```

PR #25102, opened 2025-08-29
[test] Cleanup JS engine handling

sbc100:
- Add `get_v8()` to match existing `get_nodejs()`
- Add `get_current_js_engine` helper

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -1061,17 +1061,25 @@ class RunnerCore(unittest.TestCase, metaclass=RunnerMeta):
     if self.get_setting('MEMORY64') == 2:
       self.skipTest('dynamic linking not supported with MEMORY64=2')
 
-  def require_v8(self):
+  def get_v8(self):
+    """Return v8 engine, if one is configured, otherwise None"""
     if not config.V8_ENGINE or config.V8_ENGINE not in config.JS_ENGINES:
+      return None
+    return config.V8_ENGINE
+
+  def require_v8(self):
+    v8 = self.get_v8()
+    if not v8:
       if 'EMTEST_SKIP_V8' in os.environ:
         self.skipTest('test requires v8 and EMTEST_SKIP_V8 is set')
       else:
         self.fail('d8 required to run this test.  Use EMTEST_SKIP_V8 to skip')
-    self.require_engine(config.V8_ENGINE)
+    self.require_engine(v8)
     self.cflags.append('-sENVIRONMENT=shell')
 
   def get_nodejs(self):
-    if config.NODE_JS_TEST not in self.js_engines:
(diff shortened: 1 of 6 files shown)


test/common.py
```diff
@@ -1061,17 +1061,25 @@ def check_dylink(self):
...
       else:
         self.fail('d8 required to run this test.  Use EMTEST_SKIP_V8 to skip')
-    self.require_engine(config.V8_ENGINE)
+    self.require_engine(v8)
     self.cflags.append('-sENVIRONMENT=shell')
 
   def get_nodejs(self):
-    if config.NODE_JS_TEST not in self.js_engines:
+    """Return nodejs engine, if one is configured, otherwise None"""
+    if config.NODE_JS_TEST not in config.JS_ENGINES:
```
kripken, line 1082: Why did `self` change to `config` here?

test/common.py
```diff
@@ -1463,7 +1476,9 @@ def verify_es5(self, filename):
     # use --quiet once its available
     # See: https://github.com/dollarshaveclub/es-check/pull/126/
     es_check_env = os.environ.copy()
-    es_check_env['PATH'] = os.path.dirname(config.NODE_JS_TEST[0]) + os.pathsep + es_check_env['PATH']
+    # Use NODE_JS here (that version of node that the compiler users) rather then NODE_JS_TEST (the
```
kripken:
```suggestion
    # Use NODE_JS here (the version of node that the compiler uses) rather then NODE_JS_TEST (the
```

test/common.py
```diff
@@ -1463,7 +1476,9 @@ def verify_es5(self, filename):
...
+    # version of node being used to run the tests).
```
kripken: Why use the compiler node and not the test node? This is part of a test, so I'm confused.

sbc100: Because what we trying to do is run the es_check tool here.. we are not trying to test emscripten output.

We don't care about have es-check run on that test version of node.

test/common.py
sbc100: These function get v8 or node if its available in the config.  The `self.js_engines` might be subset, or some specific engine that was configured, but these function i always want to return the engine if its available, regardless of the current `self.js_engines` which could be a subset.

sbc100: In practice I don't think it should make much differences because `self.js_engines` always starts of as copy of `config.JS_ENGINES` and is then limited on per-test based by things like `@require_node`

sbc100: I think you would only notice this change if you had a test that did `require_node` followed by `require_v8` or something complicated like that.

test/common.py
sbc100: Added more to the comment.

kripken (approved):

Issue #25103, opened 2025-08-29
TypeError race condition in parallel test runner


juj:
Ugh, the PR https://github.com/emscripten-core/emscripten/pull/25099 surfaces what looks to me like a Python internal multiprocessing bug.

On this run http://clbri.com:8010/api/v2/logs/36311/raw_inline,

error

```
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/lib/python3.12/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/usr/lib/python3.12/unittest/case.py", line 634, in run
    self._callTestMethod(testMethod)
  File "/usr/lib/python3.12/unittest/case.py", line 589, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 986, in resulting_test
    return func(self, *args)
           ^^^^^^^^^^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 784, in metafunc
    self.require_wasm_eh()
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 1186, in require_wasm_eh
    self.skipTest('test requires node v24 or d8 (and EMTEST_SKIP_EH is set)')
  File "/usr/lib/python3.12/unittest/case.py", line 711, in skipTest
    raise SkipTest(reason)
unittest.case.SkipTest: test requires node v24 or d8 (and EMTEST_SKIP_EH is set)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.12/unittest/case.py", line 633, in run
    with outcome.testPartExecutor(self):
  File "/usr/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/usr/lib/python3.12/unittest/case.py", line 63, in testPartExecutor
    _addSkip(self.result, test_case, str(e))
  File "/usr/lib/python3.12/unittest/case.py", line 89, in _addSkip
    addSkip(test_case, reason)
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/parallel_testsuite.py", line 206, in addSkip
[...]

PR #25104, opened 2025-08-29
Revert "Implement a progress counter to the parallel test suite runne…r. (#25099)"

juj:
This reverts commit e1283677a2465fdd8215d7a086d9404c83103f3c.

See https://github.com/emscripten-core/emscripten/issues/25103

This is not ready for prime time.

On several runs, Python loses multiprocessing lock.
On one run, Python crashes in a mystery last print `object address  : 0x7bd9d7d28cc0` and then non-zero exit code: http://clbri.com:8010/api/v2/logs/36405/raw_inline

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -20,14 +20,14 @@ NUM_CORES = None
 seen_class = set()
 
 
-def run_test(test, failfast_event, lock, progress_counter, num_tests):
+def run_test(test, failfast_event):
   # If failfast mode is in effect and any of the tests have failed,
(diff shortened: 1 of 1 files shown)


juj:
> On one run, Python crashes in a mystery last print `object address : 0x7bd9d7d28cc0`

Looks like that print comes from https://github.com/python/cpython/blob/5f7906715adea645d8141d5337851872cfca4381/Objects/object.c#L727

juj:
Well we got lucky.. this bug has already been fixed in upstream Python. https://github.com/emscripten-core/emscripten/pull/25110

PR #25110, opened 2025-08-29
Workaround old Python broken multiprocessing shared data structures

juj:
If running on an old python version that has broken multiprocessing shared data structures, avoid using such data structures. Fixes https://github.com/emscripten-core/emscripten/issues/25103

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -20,6 +20,13 @@ NUM_CORES = None
 seen_class = set()
 
 
+# Older Python versions have a bug with multiprocessing shared data
+# structures. https://github.com/emscripten-core/emscripten/issues/25103
+def python_multiprocessing_structures_are_buggy():
+  v = sys.version_info
+  return (v.major, v.minor, v.micro) <= (3, 12, 7) or (v.major, v.minor, v.micro) == (3, 13, 0)
+
+
 def run_test(test, failfast_event, lock, progress_counter, num_tests):
   # If failfast mode is in effect and any of the tests have failed,
   # and then we should abort executing further tests immediately.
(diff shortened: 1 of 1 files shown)


test/parallel_testsuite.py
```diff
@@ -20,6 +20,14 @@
 seen_class = set()
 
 
+# Older Python versions have a bug with multiprocessing shared data
+# structures. https://github.com/emscripten-core/emscripten/issues/25103
+# and https://github.com/python/cpython/issues/71936
+def python_multiprocessing_structures_are_buggy():
+  v = sys.version_info
+  return (v.major, v.minor, v.micro) <= (3, 12, 7) or (v.major, v.minor, v.micro) == (3, 13, 0)
```
sbc100, line 28: Perhaps at some point we can just declare that running the test suite requires > 3.13 (even though emscripten itself doesn't).

test/parallel_testsuite.py
```diff
@@ -182,38 +193,40 @@ def stopTest(self, test):
...
+      return ''
     with self.lock:
-      val = f'[{int(self.progress_counter.value * 100 / self.num_tests)}%]'
+      val = f'[{int(self.progress_counter.value * 100 / self.num_tests)}%] '
       self.progress_counter.value += 1
     return val
 
   def addSuccess(self, test):
-    print(self.compute_progress(), test, '... ok (%.2fs)' % (self.calculateElapsed()), file=sys.stderr)
+    print(f'{self.compute_progress()}{test}', '... ok (%.2fs)' % (self.calculateElapsed()), file=sys.stderr)
```
sbc100: Maybe just use a single f-string here to avoid mixing the two styles on the same line:

```
f'{self.compute_progress()}{test} ... ok ({self.calculateElapsed():.2f})'
```

sbc100 (approved):

PR #25111, opened 2025-08-29
Add basic support for Fibers with JSPI

zb3:
This PR is my attempt at supporting fibers using JSPI. My testing shows that they are indeed faster than asyncify (and the wasm file is noticeably smaller), however, there's a major limitation of using JSPI for coroutine support - a suspended coroutine might only be resumed in the same thread (worker), since the actual wasm stack is not serialized into memory. Additionally it's not possible to implement this in pure wasm as this still requires the JS trampoline.

Now let me advertise some self-noticed problems with the PR itself:
- coroutines can only be continued in the same thread, but there are no such assertions in this code (when testing I implemented them by xoring the key with pthread_ptr)
- `continuations` ids are not reused (does the same happen with `Asyncify.callStackId`?)
- continuation ids reuse the `rewind_id` structure field
- further lack of assertions
- in practice the `makeDynCall` might call `promising` when it was already done (maybe `getWasmTableEntry` could have an argument whether an async return is expected?)

What do you think about this? JSPI is already supported in chrome and might soon be supported in firefox. However the "stack switching" proposal will probably not be supported anytime soon.. so for me it would make sense to support fibers using JSPI.

### DIFF
--- a/src/lib/libasync.js
+++ b/src/lib/libasync.js
@@ -569,6 +569,10 @@ addToLibrary({
 
   $Fibers__deps: ['$Asyncify', 'emscripten_stack_set_limits', '$stackRestore'],
   $Fibers: {
+#if ASYNCIFY == 2
+    lastContinuationId: 0,
+    continuations: [],
+#endif
     nextFiber: 0,
     trampolineRunning: false,
     trampoline() {
@@ -599,6 +603,30 @@ addToLibrary({
 
       stackRestore({{{ makeGetValue('newFiber', C_STRUCTS.emscripten_fiber_s.stack_ptr,   '*') }}});
 
+#if ASYNCIFY == 2
+      const continuationId = {{{ makeGetValue('newFiber', C_STRUCTS.emscripten_fiber_s.asyncify_data + C_STRUCTS.asyncify_data_s.rewind_id, 'i32') }}};
+
+      if (continuationId !== 0) {
+#if ASYNCIFY_DEBUG
+        dbg('ASYNCIFY/FIBER: resuming fiber', newFiber);
+#endif
+        Fibers.continuations[continuationId]();
+      } else {
+        var entryPoint = {{{ makeGetValue('newFiber', C_STRUCTS.emscripten_fiber_s.entry, '*') }}};
+
+#if ASYNCIFY_DEBUG
+        dbg('ASYNCIFY/FIBER: entering fiber', newFiber, 'for the first time');
+#endif
+
+#if STACK_OVERFLOW_CHECK
+        writeStackCookie();
+#endif
+
+        var userData = {{{ makeGetValue('newFiber', C_STRUCTS.emscripten_fiber_s.user_data, '*') }}};
+
+        {{{ makeDynCall('vp', 'entryPoint', true) }}}(userData);
+      }
+#else // ASYNCIFY == 1
       var entryPoint = {{{ makeGetValue('newFiber', C_STRUCTS.emscripten_fiber_s.entry, '*') }}};
 
       if (entryPoint !== 0) {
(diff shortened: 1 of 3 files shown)


src/lib/libasync.js
```diff
@@ -569,6 +569,10 @@ addToLibrary({
 
   $Fibers__deps: ['$Asyncify', 'emscripten_stack_set_limits', '$stackRestore'],
   $Fibers: {
+#if ASYNCIFY == 2
+    lastContinuationId: 0,
+    continuations: [],
```
brendandahl, line 530: It looks like we add to this but never remove anything.

test/test_core.py
```diff
@@ -8338,12 +8338,23 @@ def test_async_ccall_promise(self, exit_runtime):
...
+  def test_fibers(self):
     self.maybe_closure()
     self.do_runf('test_fibers.cpp', '*leaf-0-100-1-101-1-102-2-103-3-104-5-105-8-106-13-107-21-108-34-109-*')
 
+  @parameterized({
+    '': (0,),
+    'legacy': (1,),
+  })
+  def test_fibers_setjmp(self, legacy):
+    self.maybe_closure()
```
brendandahl, line 8419: Please add a brief comment explaining what the test is doing.

src/lib/libasync.js
zb3: Did you mean nullifying the old continuation references [commited] or was it about the fact that ids only grow but are never reclaimed? Ids work just like `rewindId` and I don't see `callStackIdToFunc` being cleared either.. (not that it's perfect)

zb3:
> Maybe we want to add an assertion that we're using a fiber on the same thread it was created from? I'm not sure the asyncify version even supports resuming on different threads, so that may not really be important.

That assertion would be helpful but I'd need some guidance on how to implement it without compromising performance..
While I currently reuse `rewind_id`, I'd probably need to either expand the struct or make an union to also add pthread_self there.. how to best access it from JS?

Also looking at the documentation of  `rewind_id` I see it already says that it's only valid for the same thread.. yet its only the functions itself that are accessed through it (their arguments are stored in memory), so for example qemu-wasm works because these functions happen to be the same (otherwise it'd fail silently)..

That saying, adding an assertion that would also affect non-JSPI fibers would break qemu-wasm (for https://zb3.me/qemu-wasm-test/jspi-noffi/ I made a patch, but it's not generic).

UPDATE: I have no idea why some unrelated tests are now failing

src/lib/libasync.js
brendandahl: I meant nullyfing. The overflow is a separate issue, but seeing at it goes 2^53 it's probably fine.

brendandahl:
We can add the assertion in a follow up PR. To not impact performance we can have it only in debug builds. We have a similar check in the embind code for accessing val objects.

> UPDATE: I have no idea why some unrelated tests are now failing

The audioworklet is a flaky test.

test/test_core.py
```diff
@@ -8404,12 +8404,25 @@ def test_async_ccall_promise(self, exit_runtime):
     self.cflags += ['--pre-js', 'pre.js', '-sINCOMING_MODULE_JS_API=onRuntimeInitialized']
     self.do_runf('main.c', 'stringf: first\nsecond\n6.4')
 
-  @no_esm_integration('WASM_ESM_INTEGRATION is not compatible with ASYNCIFY=1')
```
brendandahl, line 8407: We probably still want this?

zb3: I thought `with_asyncify_and_jspi` includes that check, take a look: https://github.com/emscripten-core/emscripten/blob/3bad024b722cfd30ede8f59eba8a614e471f0240/test/test_core.py#L212-L228

sbc100:
Can we rebase this and see if we can land?

PR #25112, opened 2025-08-29
[test] Use `all_engines` more consistently. NFC

sbc100:
This allows us remove some the nested loops from these tests.

Ideally `all_engines` would generate separate tests for each engine but that can be a followup change.


test/test_other.py
```diff
@@ -1696,22 +1696,15 @@ def test_export_all_and_exported_functions(self):
...
     '': ([],),
     'closure': (['-O2', '--closure=1'],),
   })
   def test_stdin(self, args):
-    self.set_setting('ENVIRONMENT', 'node,shell')
-    self.emcc(test_file('module/test_stdin.c'), args=args, output_filename='out.js')
-
-    for engine in config.JS_ENGINES:
-      if engine == config.V8_ENGINE:
-        print('skipping v8 due to https://github.com/emscripten-core/emscripten/issues/25010')
```
sbc100: This is has been fixed.

sbc100:
I have a great  followup that actually splits these into one-test-per-engine

sbc100:
Here is my followup which I'm pretty happy with: https://github.com/emscripten-core/emscripten/pull/25115

Issue #24676, opened 2025-07-09
OPFS Hangs when using firefox


rickg-hcl:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
emcc -v
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.9 (e93c648a75ee38bb6e021c4c16fd18cf8f9896e9)
clang version 21.0.0git (https:/github.com/llvm/llvm-project 2f05451198e2f222ec66cec4892ada0509519290)
Target: wasm32-unknown-emscripten
Thread model: posix

**Failing command line in full:**
emcc -s WASMFS -s PROXY_TO_PTHREAD=1 -s PTHREAD_POOL_SIZE=16  -pthread  -o test.html test/wasmfs/wasmfs_opfs.c
emrun --browser=firefox test.html

The test runs, but if you reload the page a few times, it hangs.

kripken:
As this is a Firefox bug, the best place to file it is on their tracker,

https://bugzilla.mozilla.org/enter_bug.cgi?product=Firefox&_gl=1*i27ttf*_ga*NzUxODQ5NzA4LjE3NDQxNDQ4MjY.*_ga_2VC139B3XV*czE3NTIwOTk4NzYkbzEkZzAkdDE3NTIwOTk4NzYkajYwJGwwJGgw

brendandahl:
We recently added some deadlock detection to debug builds. You could try using the latest emscripten `./emsdk install tot` `./emsdk activate tot` and see if gives a backtrace.

rickg-hcl:
I didn't get a backtrace automatically, but I was able to use the firefox performance tab to get one

<img width="956" height="277" alt="Image" src="https://github.com/user-attachments/assets/b6793e3d-9db3-4f3d-b771-5d041992e981" />

and in my product, the main browser thread is hung with this stack:

<img width="592" height="380" alt="Image" src="https://github.com/user-attachments/assets/a421e6d0-a41f-426c-8d14-6f0678c0c2f5" />

rickg-hcl:
After some investigation, I've determined that the issue only occurs when I attempt to use filesystem operations like FS.readFile from the main browser thread while also doing c++ file operations on worker threads.  The main browser thread gets stuck waiting.  I can open an issue with Firefox, but this really seems to be a synchronization issue in the WASMFS part of emscripten.

rickg-hcl:
Minimum reproducible testcase. here:

```
[...]

Issue #20650, opened 2023-11-08
Using WasmFS / OPFS from the main thread deadlocks on Safari


fguinan:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.42 (6ede0b8fc1c979bb206148804bfb48b472ccc3da)
clang version 17.0.0 (https://github.com/llvm/llvm-project f3b64887de61020c09404bfee97b2fadd30df10a)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: <snip>

**Full link command and output with `-v` appended:**
emcc main.cpp -o hello.html -sWASMFS -pthread -sPTHREAD_POOL_SIZE=4

```cpp
#include <emscripten.h>
#include <emscripten/wasmfs.h>
#include <unistd.h>

#include <filesystem>
#include <string>

#include <thread>

extern "C"
{
    std::thread* bgThread = nullptr;

    EMSCRIPTEN_KEEPALIVE int main() 
    {
        // Initialize wasmfs in bg thread
        bgThread = new std::thread([]() 
        {
            auto wasmfs_backend = wasmfs_create_opfs_backend();
            const char* rootMount = "/persistent";
            bool exists = access(rootMount, F_OK) == 0;
            if (!exists)
                wasmfs_create_directory(rootMount, 0777, wasmfs_backend);
        });

         // Keep the standard library runtime alive even after exiting main`
        emscripten_exit_with_live_runtime();
        return 0;
    }

    EMSCRIPTEN_KEEPALIVE void checkForDirectory()
    {
        bool exists = std::filesystem::exists("/persistent/data");
        printf("exists = %d\n", exists);
    }
}
```

Load the html with the necessary COOP settings (in order to get pthread to work). Then in the web debugger console, make a call to `window.Module._checkForDirectory()`. In Chrome, the output is printed out successfully. In Safari, the output is never printed out. The main thread is stuck in a tight loop, and the debugger is no longer useable. If there were other web workers running, their execution also seems to pause.

sbc100:
Is this a regression?  (i.e. has this ever worked for you?)

fguinan:
We ported some code to web, and we focused our initial testing on Chrome. We just started trying to validate on Safari, which is when we discovered this issue. So this is not a regression from our point of view, but I can't say whether it's a regression in general in emscripten or Safari.

tlively:
[...]

PR #25114, opened 2025-08-29
[wasmfs] Fix data race in thread initialization.

### LINKED ISSUE #24370: wasmfs_create_opfs_backend() hangs on low end devices
Calls to wasmfs_create_opfs_backend() seem to not return on certain low end devices, or only return once every x reloads.

Tested on a dell 3100 chromebook (2 core machine) 

```
Thread model: posix
InstalledDir: /usr/lib/emsdk/upstream/bin
```

Compiling with `emcc Emscripten.c -o index.html -s WASM=1 -s USE_PTHREADS=1 -sWASMFS -sPTHREAD_POOL_SIZE=2`

```c
#include <assert.h>
#include <emscripten/console.h>
#include <emscripten/proxying.h>
#include <emscripten/threading.h>
#include <emscripten/wasmfs.h>
#include <pthread.h>
#include <stdio.h>

int mount_opfs() {
  emscripten_console_log("mount_opfs: starting");
  backend_t opfs = wasmfs_create_opfs_backend();
  emscripten_console_log("mount_opfs: created opfs backend");
  int ret = wasmfs_create_directory("/libsdl", 0777, opfs);
  emscripten_console_log("mount_opfs: mounted opfs");
  return ret;
}

void *thread_main(void *arg) {
  emscripten_console_log("trying to mount\n");
  mount_opfs();
  emscripten_console_log("finished mount\n");
}

int main() {
  int arg = 42;

  pthread_t thread;
  if (pthread_create(&thread, NULL, thread_main, &arg) != 0) {
    perror("pthread_create failed");
    return 1;
  }

  emscripten_console_log("Main thread continues\n");

  /* pthread_join(thread, NULL); */

  emscripten_console_log("Main thread done\n");
  return 0;
}
```

On my development PC, the log reaches `finished mount`, but on a lower end device it only hits `mount_opfs: starting` . The main thread continues however.

From my understanding it's spawning the proxy worker that fails, but I'm not sure how I would fix or work around this

brendandahl:
The `ProxyWorker` was creating a thread in the constructor initializer list and using a captured reference to the `started` member variable. This is problematic because it is not guaranteed that the class has been fully constructed during the initializer list.

What happens:
1) `ProxyWorker` initializers run
2) Thread starts and sets `started = true`
3) `ProxyWorker` finishes construction and sets `started = false`
4) `ProxyWorker` waits for `started` to be `true`

Step 4 will never finish in this case since the thread already ran.

To fix this we simply need to move the thread creation into the constructor body where the class is guaranteed to be fully constructed.

Fixes #24370, #24676, #20650

### DIFF
--- a/system/lib/wasmfs/thread_utils.h
+++ b/system/lib/wasmfs/thread_utils.h
@@ -34,32 +34,36 @@ class ProxyWorker {
 public:
   // Spawn the worker thread.
   ProxyWorker()
-    : queue(), thread([&]() {
-        // Notify the caller that we have started.
-        {
-          std::unique_lock<std::mutex> lock(mutex);
-          started = true;
-        }
-        cond.notify_all();
+    : queue() {
+    // Initialize the thread in the constructor to ensure the object has been
+    // fully constructed before thread starts using the object to avoid a data
+    // race. See #24370.
+    thread = std::thread([&]() {
+      // Notify the caller that we have started.
+      {
+        std::unique_lock<std::mutex> lock(mutex);
+        started = true;
+      }
+      cond.notify_all();
 
-        // Sometimes the main thread is spinning, waiting on a WasmFS lock held
-        // by a thread trying to proxy work to this dedicated worker. In that
-        // case, the proxying message won't be relayed by the main thread and
-        // the system will deadlock. This heartbeat ensures that proxying work
-        // eventually gets done so the thread holding the lock can make forward
-        // progress even if the main thread is blocked.
-        //
-        // TODO: Remove this once we can postMessage directly between workers
-        // without involving the main thread or once all browsers ship
-        // Atomics.waitAsync.
-        //
-        // Note that this requires adding _emscripten_proxy_execute_queue to
-        // EXPORTED_FUNCTIONS.
-        _wasmfs_thread_utils_heartbeat(queue.queue);
+      // Sometimes the main thread is spinning, waiting on a WasmFS lock held
+      // by a thread trying to proxy work to this dedicated worker. In that
+      // case, the proxying message won't be relayed by the main thread and
+      // the system will deadlock. This heartbeat ensures that proxying work
+      // eventually gets done so the thread holding the lock can make forward
+      // progress even if the main thread is blocked.
+      //
+      // TODO: Remove this once we can postMessage directly between workers
+      // without involving the main thread or once all browsers ship
+      // Atomics.waitAsync.
+      //
+      // Note that this requires adding _emscripten_proxy_execute_queue to
+      // EXPORTED_FUNCTIONS.
+      _wasmfs_thread_utils_heartbeat(queue.queue);
 
-        // Sit in the event loop performing work as it comes in.
-        emscripten_exit_with_live_runtime();
-      }) {
+      // Sit in the event loop performing work as it comes in.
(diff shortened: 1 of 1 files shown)


system/lib/wasmfs/thread_utils.h
```diff
@@ -34,32 +34,36 @@ class ProxyWorker {
...
   // Spawn the worker thread.
   ProxyWorker()
-    : queue(), thread([&]() {
-        // Notify the caller that we have started.
-        {
-          std::unique_lock<std::mutex> lock(mutex);
-          started = true;
-        }
-        cond.notify_all();
+    : queue() {
```
sbc100: Does clang-format not put this onto a single line now?

sbc100:
Nice work!!!!

system/lib/wasmfs/thread_utils.h
```diff
@@ -34,32 +34,36 @@ class ProxyWorker {
...
+    // Initialize the thread in the constructor to ensure the object has been
+    // fully constructed before thread starts using the object to avoid a data
+    // race. See #24370.
+    thread = std::thread([&]() {
+      // Notify the caller that we have started.
+      {
+        std::unique_lock<std::mutex> lock(mutex);
+        started = true;
```
tlively, line 47: I suspect this is still technically incorrect because the initialization write of `started` is not protected by the mutex, so it will not synchronize with this mutex-protected read on the worker thread. (Unless the `std::thread` constructor introduces some synchronization, but the documentation doesn't say.) Let's get rid of the lock and make the bool atomic in addition to the changes here.

brendandahl:
As mentioned in chat, I tried fixing this by moving `started` to the top of the decls list, but that didn't work. After thinking about it, I realized the thread also relies on the mutex and and condition variable too. I tried moving `std::thread thread` declaration to the end and that also fixes the issue.

Anyone have a preference on solution? I lean towards the current PR since any new members and dependencies will work as expected.

sbc100:
I think using the correct decl order is better, but please also a comment about the decl order being important.

sbc100:
Actually maybe apply both fixes, since didn't you also read the starting threads in initializers is bad?   Either way please add comments in the decl order.

system/lib/wasmfs/thread_utils.h
brendandahl: From what I interpret from the class initialization rules, `started` is guaranteed to be initialized before the thread is created, so we shouldn't need a lock at that point.

brendandahl: https://isocpp.org/wiki/faq/ctors#using-this-in-ctors

tlively: Yes, initialized on the current thread, but that initialization write does not necessarily become visible on any other thread without some form of synchronization.

brendandahl: It took me a little while to dig this up, but thread creation is a [synchronization point](https://timsong-cpp.github.io/cppwp/thread.thread.constr#6).

Initializing (unguarded) data before thread creation is pretty common pattern, so I was going to be surprised if that was actually UB.

tlively: Nice find, that makes sense.

tlively (approved):

sbc100 (approved):

PR #25115, opened 2025-08-30
[test] Re-implement the @all_engines decorator based on parameteraization. NFC

sbc100:
This means that we can remove the loop over JS_ENGINES from few more
tests.
    
It also splits up these tests in to smaller units which is always good
and means that it much more obvious what is going on when one of them
fails.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -450,19 +450,42 @@ def crossplatform(f):
   return f
 
 
-# without EMTEST_ALL_ENGINES set we only run tests in a single VM by
-# default. in some tests we know that cross-VM differences may happen and
-# so are worth testing, and they should be marked with this decorator
+# Without EMTEST_ALL_ENGINES set we only run tests in a single VM by default.
+# However, for some tests we know that cross-VM differences may happen and
+# so are worth testing, and they should be marked with this decorator which creates
+# N different variants of the test, one for each engine.
 def all_engines(f):
   assert callable(f)
 
+  if len(config.JS_ENGINES) == 1 or EMTEST_ALL_ENGINES:
+    return f
+
   @wraps(f)
-  def decorated(self, *args, **kwargs):
-    self.use_all_engines = True
+  def metafunc(self, engine, *args, **kwargs):
     self.set_setting('ENVIRONMENT', 'web,node,shell')
+    self.js_engines = [engine]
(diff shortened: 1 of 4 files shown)


test/common.py
```diff
@@ -449,6 +449,44 @@ def crossplatform(f):
...
+    else:
+      basename = os.path.basename(engine[0])
+      name = basename
+      suffix = 1
+      while name in engine_mapping:
+        name = f'{basename}_{suffix}'
+        suffix += 1
+    engine_mapping[name] = (engine,)
+
+  parameterize(metafunc, engine_mapping)
```
kripken, line 486: Does this not run the entire test in each engine, while before we built once and just executed the JS once per engine?

sbc100: Yes, it does work that way.  This means that there will be slight slowdown for folks with multiple engines configured.

However:

1. Configuring multiple engines is optional, and rare.  None of CI setups do this.
2. There are very few tests using this decorator (currently just 1 test on main, updated to 6 tests with #25112).

Given upsides (separate names failing tests, avoiding inner loops in tests that make debugging hard), I think its worth it in this case.

kripken: Wait, none of our CI setups test multiple engines? Don't we set up V8 on circleCI?

sbc100: We have CI tests that configure different engines, but never more than one engine at the same time.

`@all_engines` is a no-op unless you have multiple engines configured at the same time.

sbc100: Actually this might not be totally true, I think for v8 we *do* configure it to be in addition to node (i.e. we append it to JS_ENGINES) but for jsc and spidermonkey testing we always just do `JS_ENGINES = [spidermonkey]` or `JS_ENGINES = [jsc]`.

sbc100: But still, given how few tests run with `all_engines` I think this should be fine.

Also, of the 7 total tests that are currently marked with `all_engines` we can/should probably remove some of those annotations.

kripken: Fair enough, I guess it's worth the debuggability.

kripken (approved):

sbc100:
Oops this auto-merged even though there were failing tests.

Reverting for now and so I can re-land once I debug.

PR #25117, opened 2025-08-30
Clean up Closure warnings when targeting WASM2JS.

juj:
This enables Closure spam free wasm2js test suite runs.

Even though the `checkTypes` annotation doesn't do anything in wasm2js.js file for WebAssembly, I'm still adding it there because otherwise in six months me or someone else would look at that Closure warning and have a first impression "oh, that's just missing a checkTypes suppress."

Muting these warnings globally in Closure for WASM2JS runs isn't that bad, since any developer who is targeting WASM2JS will also very likely target regular Wasm runs, and there they will see any checkTypes/uselessCode related diagnostics, giving them the needed coverage.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -428,9 +428,6 @@ class TestCoreBase(RunnerCore):
   def maybe_closure(self):
     if '--closure=1' not in self.cflags and self.should_use_closure():
       self.cflags += ['--closure=1']
-      if self.is_wasm2js():
-        # wasm2js output currently contains closure warnings
-        self.cflags += ['-Wno-closure']
       logger.debug('using closure compiler..')
       return True
     return False


juj:
> Perhaps add a test confirming no warnings remain, somewhere?

Closure prints its warnings in all caps `WARNING - ...`. In my CI setup I use that to highlight the run results in orange, which is how I noticed this. So in that perspective we do have coverage.

(I presume we'd have to run all tests to ensure they really don't warn? so that would be a test to run all tests..)

kripken:
Well, I just mean a single test, for basic coverage. E.g. like this existing test:
```python
  @with_env_modify({'EMCC_LOGGING': '0'})  # this test assumes no emcc output
  def test_no_warnings(self):
    # build once before to make sure system libs etc. exist
    self.run_process([EMXX, test_file('hello_libcxx.cpp')])
    # check that there is nothing in stderr for a regular compile
    err = self.run_process([EMXX, test_file('hello_libcxx.cpp')], stderr=PIPE).stderr
    self.assertEqual(err, '')
```
But maybe it's not worth it, lgtm either way.

juj:
> But maybe it's not worth it, lgtm either way.

Hmm, I think probably better to work towards running the test suite with -Werror=closure altogether, so that all compilation with Closure would check against warnings.

sbc100:
We already run with `-Werror` by default which applies to closure warnings.  So as long as a test using the default flags we already cover this case, so I don't think we should need any special testing.

tools/building.py
```diff
@@ -666,6 +666,22 @@ def move_to_safe_7bit_ascii_filename(filename):
...
+    # to the polyfill is "in the wrong end". So mute this warning globally to
+    # allow clean Closure output.
+    cmd += ['--jscomp_off=checkTypes']
+
+    # WASM2JS codegen routinely generates expressions that are unused, e.g.
+    # WARNING - [JSC_USELESS_CODE] Suspicious code. The result of the 'bitor' operator is not being used.
+    #        s(0) | 0;
+    #        ^^^^^^^^
+    # Turn off this check in Closure to allow clean Closure output.
+    cmd += ['--jscomp_off=uselessCode']
```
sbc100, line 683: Does closure compile run on the generated wasm2js output code?  I guess it must otherwise this would not be needed.

juj: Yeah it does, I think this is because the WASM2JS output is so huge that it needs Closure to help make it half-reasonable.

juj:
> I think there are a bunch of tests that have `-Wno-closure` to avoid these issues.
> 
> Those should be removed as part of this PR, right?

Indeed, looks like these aren't needed anymore. Removed.

test/test_core.py
```diff
@@ -428,9 +428,6 @@ def setup_esm_integration(self):
   def maybe_closure(self):
     if '--closure=1' not in self.cflags and self.should_use_closure():
       self.cflags += ['--closure=1']
-      if self.is_wasm2js():
-        # wasm2js output currently contains closure warnings
-        self.cflags += ['-Wno-closure']
```
sbc100, line 433: I think there are few other places that need updating like this:
 
For example test_no_filesystem_code_size.

tools/building.py
```diff
@@ -666,6 +666,22 @@ def move_to_safe_7bit_ascii_filename(filename):
   if not settings.MINIFY_WHITESPACE:
     cmd += ['--formatting', 'PRETTY_PRINT']
 
+  if settings.WASM2JS:
+    # In WASM2JS mode, the WebAssembly object is polyfilled, which triggers
+    # Closure's built-in type check:
+    # externs.zip//webassembly.js:29:18: WARNING - [JSC_TYPE_MISMATCH] initializing variable
+    # We cannot fix this warning externally, since adding /** @suppress{checkTypes} */
+    # to the polyfill is "in the wrong end". So mute this warning globally to
+    # allow clean Closure output.
```
sbc100, line 675: Can you link to https://github.com/google/closure-compiler/issues/4108 here?

PR #25119, opened 2025-08-31
Skip tests that are tested explicitly under minimal runtime

juj:
Skip tests that are tested explicitly under minimal runtime, from being run in minimal0 test mode.

This way tests that verify minimal runtime behavior as part of e.g. `core0`, `core1` and so on, won't croak when attempted to be run under `minimal0.test_foo` - they are already covered so that mode doesn't apply.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -615,7 +615,7 @@ def also_with_minimal_runtime(f):
     if DEBUG:
       print('parameterize:minimal_runtime=%s' % with_minimal_runtime)
     if self.get_setting('MINIMAL_RUNTIME'):
-      self.skipTest('This test is verified for minimal runtime in the regular build modes, so running it in minimal0 suite does not apply.')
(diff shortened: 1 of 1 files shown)


test/common.py
```diff
@@ -614,7 +614,8 @@ def also_with_minimal_runtime(f):
   def metafunc(self, with_minimal_runtime, *args, **kwargs):
     if DEBUG:
       print('parameterize:minimal_runtime=%s' % with_minimal_runtime)
-    assert self.get_setting('MINIMAL_RUNTIME') is None
+    if self.get_setting('MINIMAL_RUNTIME'):
+      self.skipTest('This test is verified for minimal runtime in the regular build modes, so running it in minimal0 suite does not apply.')
```
sbc100: How about `self.skipTest('MINIMAL_RUNTIME already enabled in test config')`

PR #25121, opened 2025-08-31
If user is targeting ENVIRONMENT=node, then return unconditional 'true' from the Node.js detection

juj:
If user is targeting only ENVIRONMENT=node, then return unconditional 'true' from the Node.js detection code to optimize code size.

### DIFF
--- a/src/parseTools.mjs
+++ b/src/parseTools.mjs
@@ -1102,6 +1102,12 @@ function ENVIRONMENT_IS_WORKER_THREAD() {
 }
 
 function nodeDetectionCode() {
+  if (ENVIRONMENT == 'node') {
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
Any code size changes from this?

juj:
Yeah, I did get Closure optimizing ENVIRONMENT_IS_NODE checks away after that.

sbc100:
It turns out there was a test failure causes by this when running `test_environment` under v8.   This went undetected due to the fact that `.require_engine` was being called inside a look in the `test_environment` test which meant only the first engine was actually being testing.

When refactoring this test in https://github.com/emscripten-core/emscripten/pull/25115/ it exposed the bug and I fixed as part of that PR.

PR #25122, opened 2025-08-31
fix a bug in FS.getattr

warrenfalk:
The implementation of BFS's getattr requires a 'this' and calling it without this always fails.

image

### DIFF
--- a/src/lib/libfs.js
+++ b/src/lib/libfs.js
@@ -983,7 +983,7 @@ FS.staticInit();`;
       var node = stream.node;
       var getattr = stream.stream_ops.getattr;
       var arg = getattr ? stream : node;
-      getattr ??= node.node_ops.getattr;
+      getattr ??= node.node_ops.getattr.bind(node.node_ops);
       FS.checkOpExists(getattr, {{{ cDefs.EPERM }}})
       return getattr(arg);
     },


src/lib/libfs.js
```diff
@@ -983,7 +983,7 @@ FS.staticInit();`;
       var node = stream.node;
       var getattr = stream.stream_ops.getattr;
       var arg = getattr ? stream : node;
-      getattr ??= node.node_ops.getattr;
+      getattr ??= node.node_ops.getattr.bind(node.node_ops);
       FS.checkOpExists(getattr, {{{ cDefs.EPERM }}})
       return getattr(arg);
```
hoodmane: How about `getattr.call(node.node_ops, arg)` here to avoid the temporary?

warrenfalk: The problem is that `getattr` might have come from either `stream.stream_ops` or `node.node_ops` and so the wrong `this` might break something else. I'm happy to rewrite the function, but I don't have enough experience with this project to test it thoroughly.

src/lib/libfs.js
```diff
@@ -983,9 +983,10 @@ FS.staticInit();`;
       var node = stream.node;
       var getattr = stream.stream_ops.getattr;
       var arg = getattr ? stream : node;
-      getattr ??= node.node_ops.getattr.bind(node.node_ops);
+      var obj = getattr ? stream.stream_ops : node.node_ops;
+      getattr ??= node.node_ops.getattr;
       FS.checkOpExists(getattr, {{{ cDefs.EPERM }}})
-      return getattr(arg);
+      return getattr.call(obj, arg);
```
warrenfalk, line 989: @hoodmane like this?

hoodmane: LGTM

hoodmane (approved):
Would be good to have a test.

src/lib/libfs.js
sbc100: I think `doSetAttr` needs updating too.

I guess we don't have any test coverage for this because our FS callback don't use `this`.. :(

PR #25124, opened 2025-08-31
Add omitexports0 test suite

juj:
Add omitexports0 test suite to test -sDECLARE_ASM_MODULE_EXPORTS=0 option, and fix the test suite to pass in that mode.


test/test_core.py
```diff
@@ -7334,7 +7348,6 @@ def test_large_exported_response(self):
 
     self.set_setting('EXPORTED_FUNCTIONS', '@large_exported_response.json')
     self.do_run(src, 'waka 4999!')
-    self.assertContained('_exported_func_from_response_file_1', read_file('src.js'))
```
juj, line 7318: This test is redundant, and its success is implied by line 7323, so simplest to just omit it.

src/lib/libcore.js
```diff
@@ -1724,8 +1724,15 @@ addToLibrary({
   $dynCallLegacy__deps: ['$dynCalls'],
   $dynCallLegacy: (sig, ptr, args) => {
     sig = sig.replace(/p/g, {{{ MEMORY64 ? "'j'" : "'i'" }}})
+
+#if !DECLARE_ASM_MODULE_EXPORTS
+    var dynCallFn = globalThis[`dynCall_${sig}`];
+#else
+    var dynCallFn = dynCalls[sig];
```
sbc100: Doesn't the dynCalls export get populated, even in `DECLARE_ASM_MODULE_EXPORTS=0` mode?   

I can't see why it shouldn't be.

It looks simple enough to modify `exportWasmSymbols` to handle there `dynCall_` prefix (would only need to be `#if DYNCALLS || !WASM_BIGINT`).

juj: Very nice.. This didn't even occur to me, I was in the mindset that "of course the dyncalls would be statically emitted in DECLARE_ASM_MODULE_EXPORTS=1 and assigned from there to dynCalls, so that can't work."

Changed the implementation from `exportWasmSymbols()` to route the dynCalls over. That works much better, and also then resolves the dynamic linking + wasm_bigint case, for a simpler overall implementation.

src/lib/libcore.js
```diff
@@ -1627,6 +1631,11 @@ addToLibrary({
 
     for (var __exportedFunc in wasmExports) {
       var jsname = asmjsMangle(__exportedFunc);
+#if DYNCALLS || !WASM_BIGINT
+      if (jsname.startsWith('dynCall_')) {
+        dynCalls[jsname.substr(8)] = wasmExports[__exportedFunc];
```
sbc100, line 1636: `continue` here, or `else`, since we don't need these symbols globally.

sbc100 (approved):

test/common.py
```diff
@@ -750,6 +750,8 @@ def also_with_modularize(f):
   @wraps(f)
   def metafunc(self, modularize, *args, **kwargs):
     if modularize:
+      if self.get_setting('DECLARE_ASM_MODULE_EXPORTS') == 0:
+        self.skipTest('DECLARE_ASM_MODULE_EXPORTS is not compatible with MODULARIZE')
```
sbc100: `DECLARE_ASM_MODULE_EXPORTS=0`

test/test_core.py
```diff
@@ -375,7 +375,13 @@ def decorator(f):
...
       @wraps(f)
       def decorated(self, *args, **kwargs):
-        if (name + '=1') in self.cflags or self.get_setting(name):
+        nonlocal name
+        if '=' in name:
+          key, val = name.split('=', 1)
+        else:
+          key = name
+          val = 1
+        if (f'{key}={val}') in self.cflags or self.get_setting(key) == int(val):
```
sbc100: We might want to handle both `-sFOO=1` and just `-sFOO`.

However, `-sFOO` should be far more common, if not universal.

test/test_core.py
sbc100: Line 7323 is running `_exported_func_from_response_file_4999` which is different from `_exported_func_from_response_file_1`, so I'm not sure how they are the same?

test/test_core.py
```diff
@@ -8648,16 +8656,17 @@ def test_memprof_requirements(self):
...
+            if (typeof _emscripten_stack_get_base() === 'number' &&
+                typeof _emscripten_stack_get_end() === 'number' &&
+                typeof _emscripten_stack_get_current() === 'number' &&
+                 Module['___heap_base'] > 0) {
+              out('able to run memprof');
+              return 0;
+            }
+          } catch(e) { console.error(e); }
+          out('missing the required variables to run memprof');
+          return 1;
```
sbc100: Why is this change needed?

test/test_core.py
juj: > We might want to handle both `-sFOO=1` and just `-sFOO`.

The old code only handled `name + '=1'`, so it didn't handle `-sFOO` either.

Now that I read it, it looks to me that because `name` is always of form `FOO`, the old code doesn't look like it matched anything, since 'FOO=1` would never appear in `self.cflags`. (it would be `-sFOO=1` or `-sFOO` there, not `FOO=1`)

Updated to handle both.

src/lib/libcore.js
juj: But of course we do? User JS code might call `dynCall_ii(func, i1, i2)` to call a JS pointer?

sbc100: Oh I see, the `dynCalls` mapping is in addition to the individual functions, not instead of them?

lgtm then

test/test_core.py
juj: They aren't, but to test presence of _1, I would add

```
int y = EM_ASM_INT({ return Module._exported_func_from_response_file_1() });
printf("waka %d!\n", y);
```
in there to test that form.. but that is getting silly, since if _4999 is there, isn't it pretty much guaranteed to have _1 in there as well. Or if not, why wouldn't we be checking for _2, _3, _4 and so on all the way?

test/test_core.py
juj: if one had exported `foo` by populating `global['foo'] = function() { }`, then `typeof foo` still throws an exception because `foo` was not a statically written out `var` in the scope.

But calling `foo()` will still work, which is what this feature relies on.

sbc100: Hmm..  Under node it seems to work:

```
$ node
Welcome to Node.js v20.19.2.
Type ".help" for more information.
> typeof foo
'undefined'
> globalThis['foo'] = 1
1
> typeof foo
'number'
> 
```

src/lib/libcore.js
juj: Yeh, for regular users they'll still go out to the global scope.

test/test_core.py
juj: Err hmm good point. I botched something in the test.. it is only the last member that comes out as a WebAssembly.Global that was causing the failure. Good catch, updated the test now.

test/test_core.py
```diff
@@ -375,7 +375,15 @@ def decorator(f):
...
-        if (name + '=1') in self.cflags or self.get_setting(name):
+        nonlocal name
+        if '=' in name:
+          key, val = name.split('=', 1)
+        else:
+          key = name
+          val = 1
+        if int(val) == 1 and f'-s{key}' in self.cflags:
+          self.skipTest(note)
+        if (f'-s{key}={val}') in self.cflags or self.get_setting(key) == int(val):
```
sbc100: No need for braces here anymore

test/test_core.py
```diff
@@ -375,7 +375,15 @@ def decorator(f):
 
       @wraps(f)
       def decorated(self, *args, **kwargs):
-        if (name + '=1') in self.cflags or self.get_setting(name):
+        nonlocal name
```
sbc100: Why is `nonlocal` needed now but not before?

juj: Removed the nonlocal now - not needed anymore. At some point I was assigning to `name` directly, but that didn't work since it polluted the outer scope. Refactored but forgot to remove the nonlocal.

test/test_core.py
```diff
@@ -8651,7 +8661,7 @@ def test_memprof_requirements(self):
           if (typeof _emscripten_stack_get_base === 'function' &&
               typeof _emscripten_stack_get_end === 'function' &&
               typeof _emscripten_stack_get_current === 'function' &&
-              typeof Module['___heap_base'] === 'number') {
+              Module['___heap_base'] > 0) {
```
sbc100, line 8643: Hmm.. so the type of `Module['___heap_base']` in DECLARE_ASM_MODULE_EXPORTS=0 is no longer a number and is now a wasm global?    Does > 0 work on wasm globals?   Shouldn't we export the `.value` of the global instead?

juj: Yeah, it works.. looks like it attempts to be a smart object like that. `typeof` comes out different, and there may be other discrepancies that might necessitate unwrapping the .value at some point. But that wasn't necessary now, so didn't go there.

sbc100: Doesn't it seem strange that we use the unwrapped value for `DECLARE_ASM_MODULE_EXPORTS` mode but the wrapped value for `DECLARE_ASM_MODULE_EXPORTS=0` mode?

juj: Sure, though I'm not overeager to change it until there is a compelling use case that would say it has to be unified. E.g. for the purposes of accessing `Module['___heap_base']` in memory profiler will work the same in both.

sbc100: In that case I'm includes to say we should skip this test then since it depends on `Module['___heap_base']` being a number.

And we should open a bug to make `DECLARE_ASM_MODULE_EXPORTS=0` unwrap wasm globals like `DECLARE_ASM_MODULE_EXPORTS=1` does.

sbc100: I guess since this test is specifically about the memory profiler it doesn't matter, but maybe open the bug anyway since the behaviour seems wrong/inconsistent.

I'm supprised we don't have other tests that confirm that exported wasm globals are actually numbers?  We probably should if we don't already.

PR #25125, opened 2025-08-31
Fix EMPROFILE=1 toolchain profiler to not break if user passes an incorrect @rsp

juj:
Fix EMPROFILE=1 toolchain profiler to not break if user passes an incorrect @rsp file, but profile that command line as-is.

This way the incorrect @rsp file error will be properly raised from within emcc, and not the toolchain profiler.

### DIFF
--- a/tools/toolchain_profiler.py
+++ b/tools/toolchain_profiler.py
@@ -188,7 +188,12 @@ if EMPROFILE == 1:
 
     @staticmethod
     def record_subprocess_spawn(process_pid, process_cmdline):
-      expanded_cmdline = response_file.substitute_response_files(process_cmdline)
+      try:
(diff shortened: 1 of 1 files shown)


sbc100 (commented):
Can you give an example of where the old behaviour is not desirable?

Is it something like `@missing_file.txt`?  Wouldn't we just want to error in that case anyway?

juj:
> Wouldn't we just want to error in that case anyway?

Yeah, though we want the error to come from emcc.py via the proper route of trying to load the rsp, and not the toolchain profiler "stealing" the error print. The profiler should behave as if it was a transparent thing.

This way the profiler will then correctly also profile failed emcc invocations as well.

juj:
> How can `substitute_response_files` fail? is it just "file not found"?

That's right.

PR #25126, opened 2025-08-31
Fix emmalloc_trim() to work again

juj:
Found problems:

1. The convention for sbrk() changed at some point to require 16b alignment. (`__alignof__(max_align_t)`) This can create gaps in the alignment of emmalloc heap end sentinel region and the previous free region. This must require changing the assumption in emmalloc that heap end sentinel region would always be fixed 16b -> allow it to be arbitrary sized (in practice 16-31b).
2. sbrk() implementation had round-up to nearest max_align_t, but didn't handle negative sbrk() values.
3. in 4GB mode, a 32-bit sbrk() is not enough to accommodate support for both sbrk() increases and decreases (unless we take a convention that sbrk() can only be increased by < 2GB amount.. which probably won't hold in use cases that want to claim the whole heap immediately). Add a function sbrk64() that can be used in 4GB builds to be able to express both memory increases and decresases.
4. test case expectations have changed at some point. test_emmalloc_trim was written with an assumption that emmalloc would acquire the whole heap with sbrk() at startup.. but no longer does so. Adjusted test to not assume this.
5. printing bare memory values in test doesn't work well in e.g. ubsan test mode, so remove the direct value prints in the test, and just assert() the expectations.

Fixes: #23343


system/lib/libc/sbrk.c
```diff
@@ -93,6 +98,25 @@ void *sbrk(intptr_t increment_) {
   }
 }
 
+void *sbrk(intptr_t increment_) {
+#if defined(__wasm64__) // TODO || !defined(wasm2gb)
+  // In the correct https://linux.die.net/man/2/sbrk spec, sbrk() parameter is
+  // intended to be treated as signed, meaning that it is not possible in a
+  // 32-bit program to sbrk alloc (or dealloc) more than 2GB of memory at once.
```
sbc100, line 105: If this is indeed the case (I have no reason to believe its not) wouldn't it be better to fix the below bug rather than do this dance?

Maybe at least open a bug so that we can remove this code since it seems like its only need to workaround the bug?

juj: > wouldn't it be better to fix the below bug rather than do this dance?

Yes, totally agree. A later PR. (possibly.. I don't know how hard it will be to do)

juj: https://github.com/emscripten-core/emscripten/issues/25138

sbc100: Can you include the full bug URL in the code here so we can clean this up in the future?

juj: check

juj:
Would this look good now?

system/lib/emmalloc.c
```diff
@@ -1165,7 +1172,7 @@ struct mallinfo emmalloc_mallinfo() {
   struct mallinfo info;
   // Non-mmapped space allocated (bytes): For emmalloc,
   // let's define this as the difference between heap size and dynamic top end.
-  info.arena = emscripten_get_heap_size() - (size_t)sbrk(0);
+  info.arena = emscripten_get_heap_size() - (size_t)sbrk64(0);
```
sbc100: Seems like this line can be reverted?

system/lib/emmalloc.c
```diff
@@ -1227,73 +1234,118 @@ struct mallinfo emmalloc_mallinfo() {
...
 
   if (!listOfAllRegions) {
+#ifdef EMMALLOC_VERBOSE
+    MAIN_THREAD_ASYNC_EM_ASM(out('emmalloc_trim(): cannot trim memory, emmalloc is currently not initialized to manage any dynamic memory at all.'));
+#endif
     return 0; // emmalloc is not controlling any dynamic memory at all - cannot release memory.
   }
   uint8_t *previousSbrkEndAddress = listOfAllRegions->endPtr;
-  assert(sbrk(0) == previousSbrkEndAddress);
+  void *sbrk_addr = sbrk64(0);
```
sbc100: The rest of this file seems to use `camelCase` for local variable names.

sbc100: Seems like this can just be `sbrk(0)`.   Hopefully in the longer run we can remove `sbrk64` so the fewer references to it the better.

system/lib/emmalloc.c
```diff
@@ -1227,73 +1234,118 @@ struct mallinfo emmalloc_mallinfo() {
...
-  endSentinelRegion = (Region*)((uint8_t*)lastActualRegion + newRegionSize);
-  create_used_region(endSentinelRegion, sizeof(Region));
+  // Call sbrk() to shrink the memory area.
+  void *oldSbrk = sbrk64(-(int64_t)shrinkAmount);
+  assert((intptr_t)oldSbrk != -1); // Shrinking with sbrk() should never fail.
 
-  // And update the size field of the whole region block.
-  listOfAllRegions->endPtr = (uint8_t*)endSentinelRegion + sizeof(Region);
+  // Ask where sbrk() got us at.
+  void *sbrkNow = sbrk64(0);
```
sbc100: Same here.. it also means the comment above is accurate.

system/lib/emmalloc.c
```diff
@@ -61,6 +61,8 @@
 #include <emscripten/heap.h>
 #include <emscripten/threading.h>
 
+void *sbrk64(int64_t numBytes);
```
sbc100: Lets call this `_sbrk64` to avoid polluting the global namespace.

sbc100 (commented):
lgtm % comments

system/lib/emmalloc.c
juj: The idea is for emmalloc to only refer to this new sbrk64(). This way if user code only uses emmalloc and nothing else (doesn't sbrk() manually), then emmalloc will not pull in the flawed middle-man sbrk() function at all, but just the sbrk64() variant.

So that is why I changed emmalloc to only call to the "fixed" sbrk64().

system/lib/emmalloc.c
juj: I think it will be important for end users to be able to call this function as well. I am already thinking that at Unity we'll patch bdwgc to call to this fixed sbrk64() function.

system/lib/emmalloc.c
juj: I think the opposite direction: `sbrk()` is flawed, so we must migrate to using `sbrk64()` even in 4GB builds.

(2GB builds don't care, since there the problem doesn't exist)

sbc100: But we agree that even if `sbrk()` is limited in how much it can allocate we need to fix our current `sbrk()` so that it can shrink memory, right?  That is what #25138 is for I guess?

So is the idea that rather than updating all our malloc implementations to chop up `sbrk()` calls into in 2gb chunks, we would instead move them all over to using `sbrk64`?

sbc100: I think we should land this change using `_sbrk64()` and then we can bikeshed the final solution we want in #25138.

Once we make the new function public we can declare it in a header and remove the underscore prefix.

juj: > But we agree that even if `sbrk()` is limited in how much it can allocate we need to fix our current `sbrk()` so that it can shrink memory, right? That is what #25138 is for I guess?

Yeah, that's right. 25138 means that interpretation of existing 32-bit sbrk() must be turned back to signed again.

> So is the idea that rather than updating all our malloc implementations to chop up `sbrk()` calls into in 2gb chunks, we would instead move them all over to using `sbrk64`?

That's true. At first I was thinking that callers can just chop their allocs to 2gb pieces, but the issue there is that then there won't be a guarantee that such sbrk() allocs are contiguous.

For example in a multithreaded program with bdwgc, different threads can race to sbrk() for native malloc(), and bdwgc races to sbrk() for itself. So if it was patched to alloc in 2gb slices, another thread could come in between, and the result for the chopped alloc would result in noncontinuous >2GB memory area being received, which would be catastrophic.

So best to just provide a int64_t utilizing sbrk64() api instead.

juj: > I think we should land this change using `_sbrk64()` and then we can bikeshed the final solution we want in #25138.
> 
> Once we make the new function public we can declare it in a header and remove the underscore prefix.

Ok, updated.

juj:
Hmm I'm having trouble landing this: I can't get the rebaseline_expectations.py script to create expectations that will match what's desired. (tried running the script on Windows and on Linux)

Would it make sense to land, and then run the GitHub Action?

sbc100:
> Hmm I'm having trouble landing this: I can't get the rebaseline_expectations.py script to create expectations that will match what's desired. (tried running the script on Windows and on Linux)
> 
> Would it make sense to land, and then run the GitHub Action?

Thats one option.

But using `./emsdk install tot` followed by `./emcc --clear-cache` followed by `tools/maint/rebaseline_tests.py` should always works (modulo real bugs that we should track down).

juj:
> But using `./emsdk install tot` followed by `./emcc --clear-cache` followed by `tools/maint/rebaseline_tests.py` should always works (modulo real bugs that we should track down).

Ok, rebaselined with `emsdk install tot`, let's see if that does the trick

PR #25128, opened 2025-09-01
[AUDIO_WORKLET] Added TODO for max_align_t/minimum alloc NFC

cwoffenden:
This is a minor docs only change, since I spotted `max_align_t` had already changed before and may affect this.

### DIFF
--- a/src/audio_worklet.js
+++ b/src/audio_worklet.js
@@ -42,9 +42,10 @@ function createWasmAudioWorkletProcessor(audioParams) {
       // Prepare the output views; see createOutputViews(). The 'minimum alloc'
       // firstly stops STACK_OVERFLOW_CHECK failing (since the stack will be
       // full if we allocate all the available space, with 16 bytes being the
-      // minimum allo size due to alignments) leaving room for a single
+      // minimum alloc size due to alignments) leaving room for a single
       // AudioSampleFrame as a minumum. There's an arbitrary maximum of 64, for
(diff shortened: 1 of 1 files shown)


juj:
Yeh - it's unlikely that it will ever change to anything else from 16, but good to note down.

src/audio_worklet.js
```diff
@@ -42,9 +42,10 @@ function createWasmAudioWorkletProcessor(audioParams) {
       // Prepare the output views; see createOutputViews(). The 'minimum alloc'
       // firstly stops STACK_OVERFLOW_CHECK failing (since the stack will be
       // full if we allocate all the available space, with 16 bytes being the
-      // minimum allo size due to alignments) leaving room for a single
+      // minimum alloc size due to alignments) leaving room for a single
       // AudioSampleFrame as a minumum. There's an arbitrary maximum of 64, for
       // the case where a multi-MB stack is passed.
+      // TODO: if max_align_t changes the minimum alloc will need readdressing
       this.outputViews = new Array(Math.min(((wwParams.stackSize - /*minimum alloc*/ 16) / this.bytesPerChannel) | 0, /*sensible limit*/ 64));
```
sbc100, line 49: Actually I think you an replace `/*minimum alloc*/ 16` here with `{{{ STACK_ALIGN }}}`

cwoffenden: Tested and I can so did: #25150

PR #25130, opened 2025-09-01
Fix rounding in SSE conversion functions

### LINKED ISSUE #25129: _mm_cvttpd_epi32 emulation incorrectly rounds values near integer boundaries
4.0.8

---

**Description and Reproduction**

Emscripten’s emulation of `_mm_cvttpd_epi32` is currently broken for some inputs.  
For example, the following should print `1022` but instead prints `1023`:

```c
// Compile with: emcc -msimd128 -msse3 repro.c
#include <emmintrin.h>
#include <stdio.h>

int main() {
  __m128d v = _mm_set1_pd(1022.99998194495);
  __m128i i = _mm_cvttpd_epi32(v);
  int out[4];
  _mm_storeu_si128((__m128i*)out, i);
  printf("%d %d\n", out[0], out[1]);
}
```

**Expected output:**
```
1022 1022
```

**Actual output:**
```
1023 1023
```

---

**Root Cause**

The current emulation converts each `double` lane to `float` before truncating, which loses precision for values near integer boundaries.  
In this case, `1022.99998194495` becomes `1023.0f`, so truncation yields `1023` instead of `1022`.

---

**Suggested Fix**

Avoid the lossy cast to `float` and use the f64 truncation builtin:

```c
static inline __m128i _mm_cvttpd_epi32(__m128d a) {
  int m[4] = {0, 0, 0, 0};
  for (int i = 0; i < 2; ++i) {
    double e = a[i];
    if (!isfinite(e) || e >= 2147483648.0 || e < -2147483648.0) {
      m[i] = 0x80000000; // match Intel semantics for NaN / out of range
    } else {
      m[i] = __builtin_wasm_trunc_s_i32_f64(e); // <-- f64 truncation, not f32
    }
  }
  return wasm_i32x4_make(m[0], m[1], 0, 0);
}
```

This ensures correct results without losing precision.

juj:
Value 3.6 was  not that interesting float, replaced it with the more interesting 1022.99998194495 from #25129.

Fixes: #25129

### DIFF
--- a/system/include/compat/emmintrin.h
+++ b/system/include/compat/emmintrin.h
@@ -1010,7 +1010,7 @@ _mm_cvtsd_si64(__m128d __a)
   double e = __a[0];
   if (isnan(e) || isinf(e)) return 0x8000000000000000LL;
   long long x = llrint(e);
-  if (x != 0xFFFFFFFF00000000ULL && (x != 0 || fabs(e) < 2.f) && e <= LLONG_MAX && e >= LLONG_MIN)
+  if ((x != 0 || fabs(e) < 2.f) && e <= LLONG_MAX && e >= LLONG_MIN)
     return x;
   else
     return 0x8000000000000000LL;
@@ -1023,7 +1023,7 @@ _mm_cvttsd_si64(__m128d __a)
   double e = __a[0];
   if (isnan(e) || isinf(e) || e > LLONG_MAX || e < LLONG_MIN) return 0x8000000000000000LL;
   long long x = llrint(e);
-  if (x != 0xFFFFFFFF00000000ULL && (x != 0 || fabs(e) < 2.f))
+  if (x != 0 || fabs(e) < 2.f)
     // Use the trapping instruction here since we have explicit bounds checks
     // above
     return __builtin_wasm_trunc_s_i64_f64(e);
--- a/system/include/compat/emmintrin.h
+++ b/system/include/compat/emmintrin.h
@@ -1068,8 +1068,7 @@ _mm_cvttps_epi32(__m128 __a)
   for(int i = 0; i < 4; ++i)
   {
     float e = __a[i];
-    int x = lrint(e);
-    if ((x != 0 || fabs(e) < 2.0) && !isnanf(e) && e <= INT_MAX && e >= INT_MIN)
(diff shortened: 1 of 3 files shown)


system/include/compat/emmintrin.h
```diff
@@ -433,11 +433,11 @@ _mm_cvttpd_epi32(__m128d __a)
   int m[2];
   for(int i = 0; i < 2; ++i)
   {
-    float elem = __a[i];
-    if ((lrint(elem) != 0 || fabs(elem) < 2.0) && !isnanf(elem) && elem <= INT_MAX && elem >= INT_MIN)
+    double elem = __a[i];
+    if ((lrint(elem) != 0 || fabs(elem) < 2.0) && !isnanf(elem) && elem < 2147483648.0 && elem >= -2147483648.0)
```
sbc100: These hardcoded constants seem worse than using predefined macros.    Are there no macros for these?

juj: One could do e.g. `((float)INT_MAX + 1.0f)`, but it has the very annoying property that debug code then gets generated with silly arithmetic, and anyone debugging through SIMD code won't see the code resembling what it is supposed to be.

I could create some #defines, but seeing these literals is quite self-documenting, since then user doesn't need to hop back and forth to search their value.

system/include/compat/xmmintrin.h
```diff
@@ -597,9 +597,8 @@ _mm_cvtsi32_ss(__m128 __a, int __b)
 static __inline__ int __attribute__((__always_inline__, __nodebug__, DIAGNOSE_SLOW)) _mm_cvtss_si32(__m128 __a)
 {
   float e = ((__f32x4)__a)[0];
-  int x = lrint(e);
-  if ((x != 0 || fabsf(e)) < 2.f && !isnan(e) && e <= INT_MAX && e >= INT_MIN)
-    return x;
+  if (e < 2147483648.0f && e >= -2147483648.0f && (lrint(e) != 0 || fabsf(e) < 2.f))
```
sbc100, line 600: Here you are moving the INT_MAX range change first, should we do that same in _mm_cvttsd_si32?

juj: updated to be consistent

sbc100:
lgtm, but maybe lets wait for @tlively to chime in.

PR #25141, opened 2025-09-02
[testing] Use threads for the http servers.

brendandahl:
This is in preparation for the parallel browser tests where sub processes aren't allowed.

The server no longer changes the directory during tests, since we can now rely on the test runner changing the directory (same process).


test/test_browser.py
```diff
@@ -56,6 +58,11 @@ def do_OPTIONS(s):
       s.sendheaders([("Access-Control-Allow-Headers", "Range")], 0)
 
     def do_GET(s):
+       # CORS preflight makes OPTIONS requests which we need to account for.
```
sbc100: Indentation looks wrong here.

test/common.py
```diff
@@ -2513,7 +2529,7 @@ def is_browser_test(self):
 
   def add_browser_reporting(self):
     contents = read_file(test_file('browser_reporting.js'))
-    contents = contents.replace('{{{REPORTING_URL}}}', BrowserCore.SERVER_URL)
+    contents = contents.replace('{{{REPORTING_URL}}}', self.SERVER_URL)
```
sbc100: Why this change?

test/common.py
```diff
@@ -2395,8 +2389,27 @@ def log_request(code=0, size=0):
   # do not have the correct MIME type
   SimpleHTTPRequestHandler.extensions_map['.mjs'] = 'text/javascript'
 
-  httpd = ThreadingHTTPServer(('localhost', port), TestServerHandler)
-  httpd.serve_forever() # test runner will kill us
+  return ThreadingHTTPServer(('localhost', port), TestServerHandler)
+
+
+class ServerThread(threading.Thread):
```
sbc100: HttpServerThread?

test/common.py
```diff
@@ -2395,8 +2389,27 @@ def log_request(code=0, size=0):
...
+  """A generic thread class to create and run a server."""
```
sbc100: `.. run an http server`

test/common.py
```diff
@@ -2395,8 +2389,27 @@ def log_request(code=0, size=0):
...
+  def __init__(self, server_creator, *server_args):
```
sbc100: `server_ctor` or `server_constructor`.. or maybe `server_class`?

test/common.py
```diff
@@ -2395,8 +2389,27 @@ def log_request(code=0, size=0):
...
+    super().__init__()
+    self.server = None
+    self.server_creator = server_creator
+    self.server_args = server_args
```
sbc100: Can you just do `self.server = server_creator(*server_args)` here, rather then in the run function?

If so I guess you could even accept the full constructed server as the argument instead of the ctor + args?

sbc100 (commented):
Can you remind me why "sub processes aren't allowed."?

brendandahl:
> Can you remind me why "sub processes aren't allowed."?

Workers in the pool are already daemon processes and do not allow you to create sub processes.

More on the daemonic stuff: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.daemon

test/common.py
brendandahl: The server needs to be created on a seperate thread which is where `run` is called.

test/common.py
brendandahl: It takes a function, so I don't think those names work.

test/common.py
brendandahl: not needed here, removing

test/common.py
```diff
@@ -2488,21 +2501,24 @@ def setUpClass(cls):
     super().setUpClass()
     if not has_browser() or EMTEST_BROWSER == 'node':
       return
-    cls.harness_in_queue = multiprocessing.Queue()
-    cls.harness_out_queue = multiprocessing.Queue()
-    cls.harness_server = multiprocessing.Process(target=harness_server_func, args=(cls.harness_in_queue, cls.harness_out_queue, cls.PORT))
+
+    cls.harness_in_queue = queue.Queue()
+    cls.harness_out_queue = queue.Queue()
+    cls.harness_server = HttpServerThread(harness_server_func, cls.harness_in_queue, cls.harness_out_queue, cls.PORT)
```
sbc100: Can we at least rename `harness_server_func` to something like `make_test_server` or `make_test_harness` since that function now returns that server rather than running it?

kripken (approved):

PR #25142, opened 2025-09-02
Fix tools/maint/rebaseline_tests.py on Windows.

juj:


juj:
It complains that `test/runner` is an invalid file name, `/` cannot be a character in a file name.

Then if adding `shell=True` without other changes, it runs it as `test /runner`, i.e. launches `test.exe` with `/runner` as an argument to it.

`os.path.join()` gives the native `test\runner` format to it, and `shell=True` will then autodetect from the multitude of suffixes the one to launch, as if user had run `test\runner` on the command line.

tools/maint/rebaseline_tests.py
```diff
@@ -86,7 +86,7 @@ def main():
       print('tree is not clean')
       return 1
 
-    subprocess.check_call(['test/runner', '--rebaseline', '--browser=0'] + TESTS, cwd=root_dir)
+    subprocess.check_call([os.path.join('test', 'runner'), '--rebaseline', '--browser=0'] + TESTS, cwd=root_dir, shell=True)
```
sbc100, line 89: Can you use `bat_suffix` instead of `shell=True`?

sbc100: Oh, I wonder if using `bat_suffix` obviates the need for `os.path.join` here?

PR #25148, opened 2025-09-03
NFC: Use `TextEncoder` for stringToUTF8Array and lengthBytesUTF8 Function

PtPrashantTripathi:
This PR refactors the `$lengthBytesUTF8` function to use the native `TextEncoder` API.  
This change simplifies the codebase, improves correctness and reliability, and aligns the implementation with modern web standards.

The following code replaces the old, manually-coded version:

```js
  /**
   * Returns the number of bytes the given JavaScript string takes if encoded as a
   * UTF8 byte array, EXCLUDING the null terminator byte.
   *
   * @param {string} str - The JavaScript string to operate on.
   * @return {number} The length, in bytes, of the UTF-8 encoded string.
   */
  $lengthBytesUTF8: (str) => {
    var UTF8Encoder = new TextEncoder();
    return UTF8Encoder.encode(str).length;
  }
```

---

## **Change Log**

* **Removed** manual UTF-16 surrogate pair and character code logic.
* **Added** reliance on the native `TextEncoder` web API.
* **Improved** code simplicity, readability, and maintainability.
* **Ensured** accurate and robust UTF-8 byte length calculation for all Unicode characters.

---

## **Explanation**

The original implementation manually calculates the UTF-8 byte length by iterating through the string's characters and checking their `charCodeAt`.
While functional, this approach was **complex and fragile**:

* It required specific logic to correctly handle **multi-byte characters**.
* It had to identify and process **UTF-16 surrogate pairs**, which represent characters outside the Basic Multilingual Plane (e.g., emojis and certain CJK ideographs).

### **New Implementation Benefits**

The new implementation leverages the built-in `TextEncoder` API, which is **modern, standardized, and highly optimized**.

#### Why this is an improvement:

1. **Simplicity**

   * The manual loop is replaced by a **single, declarative line of code**.
   * Easier to read, understand, and **less prone to human error**.

2. **Robustness**

   * `TextEncoder` correctly handles **all Unicode edge cases** without complex, manual checks.
   * Works seamlessly for strings containing characters like:

     ```
     こんにちは
     or
     नमस्ते
     ```

     which require careful surrogate pair handling in the old implementation.

3. **Performance**

   * Native browser APIs like `TextEncoder` are implemented in **highly optimized C++**.
   * **Faster** than manual JavaScript logic, especially for **long or complex strings**.

---

## **Example**

To demonstrate correctness, consider this test case:

```js
// A string with characters that are multi-byte in UTF-8
const testString = "नमस्ते विश्व!";

// Old manual function:
// lengthBytesUTF8(testString) === 35

// New TextEncoder implementation:
const byteLength = new TextEncoder().encode(testString).length;
// byteLength === 35
```

Both methods yield the **correct result**, but the new implementation achieves it with a **more reliable and maintainable approach**.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -603,3 +603,4 @@ a license to everyone to use it as detailed in LICENSE.)
 * Sean Morris <sean@seanmorr.is>
 * Mitchell Wills <mwills@google.com> (copyright owned by Google, Inc.)
 * Han Jiang <jhcarl0814@gmail.com>
+* Pt. Prashant Tripathi <ptprashanttripathi@outlook.com>
--- a/src/lib/libstrings.js
+++ b/src/lib/libstrings.js
@@ -8,6 +8,10 @@
 #error "TEXTDECODER must be either 1 or 2"
 #endif
 
+#if TEXTENCODER != 1 && TEXTENCODER != 2
+#error "TEXTENCODER must be either 1 or 2"
+#endif
+
 addToLibrary({
   // In -sAUDIO_WORKLET builds, TextDecoder will not exist in AudioWorkletGlobalScope,
   // so we cannot try to unconditionally initialize it in that build mode.
@@ -16,6 +20,13 @@ addToLibrary({
   $UTF8Decoder: "new TextDecoder()",
 #else
   $UTF8Decoder: "globalThis.TextDecoder && new TextDecoder()",
+#endif
+
+  // TextEncoder constructor defaults to UTF-8
+#if TEXTENCODER == 2
+  $UTF8Encoder: "new TextEncoder()",
+#else
+  $UTF8Encoder: "typeof TextEncoder != 'undefined' ? new TextEncoder() : undefined",
 #endif
 
   $findStringEnd: (heapOrArray, idx, maxBytesToRead, ignoreNul) => {
@@ -149,9 +160,12 @@ addToLibrary({
    *                                   terminator.
    * @return {number} The number of bytes written, EXCLUDING the null terminator.
    */
+  $stringToUTF8Array__deps: [
+    '$UTF8Encoder',
 #if ASSERTIONS
-  $stringToUTF8Array__deps: ['$warnOnce'],
+    '$warnOnce',
 #endif
+  ],
   $stringToUTF8Array: (str, heap, outIdx, maxBytesToWrite) => {
 #if CAN_ADDRESS_2GB
     outIdx >>>= 0;
@@ -164,6 +178,28 @@ addToLibrary({
     if (!(maxBytesToWrite > 0))
       return 0;
 
+#if TEXTENCODER == 2
+    // Always use TextEncoder when TEXTENCODER == 2
+    var encoded = UTF8Encoder.encode(str);
+    var bytesToWrite = Math.min(encoded.length, maxBytesToWrite - 1); // -1 for null terminator
+    for (var i = 0; i < bytesToWrite; ++i) {
+      heap[outIdx + i] = encoded[i];
+    }
+    heap[outIdx + bytesToWrite] = 0;
+    return bytesToWrite;
+#else
+    // When using conditional TextEncoder
+    if (UTF8Encoder) {
+      var encoded = UTF8Encoder.encode(str);
+      var bytesToWrite = Math.min(encoded.length, maxBytesToWrite - 1); // -1 for null terminator
+      for (var i = 0; i < bytesToWrite; ++i) {
+        heap[outIdx + i] = encoded[i];
+      }
+      heap[outIdx + bytesToWrite] = 0;
+      return bytesToWrite;
+    }
+
+    // Fallback: manual UTF-8 encoding
     var startIdx = outIdx;
     var endIdx = outIdx + maxBytesToWrite - 1; // -1 for string null terminator.
     for (var i = 0; i < str.length; ++i) {
@@ -200,6 +236,7 @@ addToLibrary({
     // Null-terminate the pointer to the buffer.
     heap[outIdx] = 0;
     return outIdx - startIdx;
+#endif // TEXTENCODER == 2
   },
 
   /**
@@ -226,7 +263,18 @@ addToLibrary({
    * @param {string} str - JavaScript string to operator on
    * @return {number} Length, in bytes, of the UTF8 encoded string.
    */
+  $lengthBytesUTF8__deps: ['$UTF8Encoder'],
   $lengthBytesUTF8: (str) => {
+#if TEXTENCODER == 2
+    // Always use TextEncoder when TEXTENCODER == 2
+    return UTF8Encoder.encode(str).length;
+#else
+    // When using conditional TextEncoder
+    if (UTF8Encoder) {
+      return UTF8Encoder.encode(str).length;
+    }
+
+    // Fallback: manual calculation
     var len = 0;
     for (var i = 0; i < str.length; ++i) {
       // Gotcha: charCodeAt returns a 16-bit word that is a UTF-16 encoded code
@@ -245,6 +293,7 @@ addToLibrary({
       }
     }
     return len;
+#endif // TEXTENCODER == 2
   },
 
   $intArrayFromString__docs: '/** @type {function(string, boolean=, number=)} */',
--- a/src/settings.js
+++ b/src/settings.js
@@ -1742,6 +1742,13 @@ var EVAL_CTORS = 0;
 // [link]
 var TEXTDECODER = 1;
 
+// The default value of 1 means the generated code will use TextEncoder if
+// available and fall back to custom encoding code when it is not available.
+// If set to 2, we assume TextEncoder is always present and usable, and no
+// fallback JS code will be emitted.
+// [link]
+var TEXTENCODER = 1;
+
 // Embind specific: If enabled, assume UTF-8 encoded data in std::string binding.
 // Disable this to support binary data transfer.
 // [link]
--- a/site/source/docs/tools_reference/settings_reference.rst
+++ b/site/source/docs/tools_reference/settings_reference.rst
@@ -2645,6 +2645,18 @@ of ENVIRONMENT since TextDecoder is not available in those environments).
 
 Default value: 1
 
+.. _textencoder:
+
+TEXTENCODER
+===========
+
+The default value of 1 means the generated code will use TextEncoder if
+available and fall back to custom encoding code when it is not available.
+If set to 2, we assume TextEncoder is always present and usable, and no
+fallback JS code will be emitted.
+
+Default value: 1
+
 .. _embind_std_string_is_utf8:
 
 EMBIND_STD_STRING_IS_UTF8
--- a/tools/link.py
+++ b/tools/link.py
@@ -1183,16 +1183,21 @@ def phase_linker_setup(options, linker_args):  # noqa: C901, PLR0912, PLR0915
   if 'noExitRuntime' in settings.INCOMING_MODULE_JS_API:
     settings.DEFAULT_LIBRARY_FUNCS_TO_INCLUDE.append('$noExitRuntime')
 
-  # Default to TEXTDECODER=2 (always use TextDecoder to decode UTF-8 strings)
-  # in -Oz builds, since custom decoder for UTF-8 takes up space.
-  # When supporting shell environments, do not do this as TextDecoder is not
-  # widely supported there.
-  # In Audio Worklets TextDecoder API is intentionally not exposed
-  # (https://github.com/WebAudio/web-audio-api/issues/2499) so we also need to
-  # keep the JavaScript-based fallback.
-  if settings.SHRINK_LEVEL >= 2 and not settings.ENVIRONMENT_MAY_BE_AUDIO_WORKLET and \
+  # Default to TEXTDECODER=2 (always use TextDecoder for decoding UTF-8 strings)
+  # and TEXTENCODER=2 (always use TextEncoder for encoding UTF-8 strings)
+  # in -Oz builds to minimize code size, as custom UTF-8 encoder/decoder logic
+  # adds extra code footprint.
+  #
+  # Exceptions:
+  # - Shell environments: TextDecoder and TextEncoder are not consistently
+  #   available, so we must retain the JavaScript-based fallback.
+  # - Audio Worklets: The TextDecoder API is intentionally not exposed
+  #   (see: https://github.com/WebAudio/web-audio-api/issues/2499), requiring
+  #   the fallback decoder to remain in place.
+  if settings.SHRINK_LEVEL >= 2 and not settings.AUDIO_WORKLET and \
      not settings.ENVIRONMENT_MAY_BE_SHELL:
     default_setting('TEXTDECODER', 2)
+    default_setting('TEXTENCODER', 2)
 
   # If set to 1, we will run the autodebugger (the automatic debugging tool, see
   # tools/autodebugger).  Note that this will disable inclusion of libraries. This


src/lib/libstrings.js
```diff
@@ -218,31 +218,15 @@ addToLibrary({
...
-      } else if (c <= 0x7FF) {
-        len += 2;
-      } else if (c >= 0xD800 && c <= 0xDFFF) {
-        len += 4; ++i;
-      } else {
-        len += 3;
-      }
-    }
-    return len;
+    var UTF8Encoder = new TextEncoder();
```
juj: The approach here would be to instantiate the encoder once at global scope, so that each string length query won't allocate a new encoder. See how TextDecoder does it: https://github.com/emscripten-core/emscripten/blob/55e95ed90fc2fae967feb12c4b39f2f2b2a8997c/src/lib/libstrings.js#L13-L17

src/lib/libstrings.js
```diff
@@ -218,31 +218,15 @@ addToLibrary({
...
+    return UTF8Encoder.encode(str).length;
```
juj, line 270: This is a great idea, though likely results in disappointing overall performance, because it will allocate and throw away a string, producing temporary garbage.

I think it's probably better to address the call sites of `lengthBytesUTF8()` directly. For example,

```js
  $stringToNewUTF8: (str) => {
    var size = lengthBytesUTF8(str) + 1;
    var ret = _malloc(size);
    if (ret) stringToUTF8(str, ret, size);
    return ret;
  },
```
could become
```js
  $utf8Encoder: 'new TextEncoder("utf-8")',
  $stringToNewUTF8: (str) => {
    var utf8 = utf8Encoder.encode(str);
    var ptr = _malloc(utf8.length + 1);
#if !ABORTING_MALLOC
    if (ptr) {
#endif
      HEAPU8.set(utf8, ptr);
      HEAPU8[ptr + utf8.length] = 0;
#if !ABORTING_MALLOC
    }
#endif
    return ptr;
  },
```
when TextEncoder is enabled. This way both the `lengthBytesUTF8()` and `stringToUTF8()` will be replaced with just one encode use of TextEncoder.

PtPrashantTripathi: good idea

src/lib/libstrings.js
PtPrashantTripathi: Done

src/lib/libstrings.js
PtPrashantTripathi: Hey @juj 

I've implemented the TEXTENCODER pattern we discussed, following the same approach as the existing TEXTDECODER implementation. Here's what I've added:

## Changes Made:

- **Added TEXTENCODER compile-time flag** with the same 1/2 options as TEXTDECODER
- **Enhanced `stringToUTF8Array`** to use TextEncoder when available/beneficial 
- **Enhanced `lengthBytesUTF8`** to leverage TextEncoder for accurate byte length calculation
- **Maintained full backward compatibility** with existing fallback logic

The implementation follows the exact same pattern as TEXTDECODER:
- `TEXTENCODER=2`: Always use TextEncoder (modern environments)
- `TEXTENCODER=1`: Conditional usage - TextEncoder for longer strings (>16 chars), manual encoding for shorter ones

## Testing:
I've tried to double-check the logic but honestly, the vscode syntax highlighter is going crazy with all the `#if` preprocessor directives mixed with JavaScript 😅. Getting red squiggly lines everywhere since VSCode doesn't know how to handle this C/JS hybrid syntax.

**Quick question for the team**: What VSCode settings/extensions do you guys use to make this Emscripten code more readable? The syntax highlighting is pretty much useless for me right now with all the preprocessor directives.

Hope I didn't introduce any bugs! The pattern should be solid since it mirrors the existing TEXTDECODER implementation. Let me know if you spot anything that needs fixing.

Thanks!

juj:
> Quick question for the team: What VSCode settings/extensions do you guys use to make this Emscripten code more readable? The syntax highlighting is pretty much useless for me right now with all the preprocessor directives.

Personally I use Sublime Text as the editor, not sure what others use. It does also sometimes get confused with the `#if` preprocessors, though I've gotten by getting used to it.

> Hope I didn't introduce any bugs

You can see CircleCI logs on this PR, that take you to URLs of live CI runs that check if there are any errors. Looks like there is something to fix there - check out the report pages there.

juj:
> ```js
>   $lengthBytesUTF8: (str) => {
>     var UTF8Encoder = new TextEncoder();
>     return UTF8Encoder.encode(str).length;
>   }
> ```

Btw if you still wanted to implement use of TextEncoder into `lengthBytesUTF8`, you can do so with something like

```js
$lengthBytesUTF8__deps: ['$UTF8Encoder'],
$lengthBytesUTF8: (str) => {
#if SHRINK_LEVEL == 2
  return UTF8Encoder.encode(str).length;
#else
  .. the original implementation here
#endif
}
```

`SHRINK_LEVEL=2` corresponds to using `-Oz` on command line, i.e. "optimize maximally for smallest code size, ignoring any performance implications".

AUTHORS
```diff
@@ -601,3 +601,5 @@ a license to everyone to use it as detailed in LICENSE.)
 * Artur Gatin <agatin@teladochealth.com> (copyright owned by Teladoc Health, Inc.)
 * Christian Lloyd <clloyd@teladochealth.com> (copyright owned by Teladoc Health, Inc.)
 * Sean Morris <sean@seanmorr.is>
+* Pt. Prashant Tripathi <ptprashanttripathi@outlook.com>
+
```
sbc100: Remove extra newline

src/lib/libstrings.js
```diff
@@ -16,15 +20,21 @@ addToLibrary({
...
   $findStringEnd: (heapOrArray, idx, maxBytesToRead, ignoreNul) => {
     var maxIdx = idx + maxBytesToRead;
     if (ignoreNul) return maxIdx;
     // TextDecoder needs to know the byte length in advance, it doesn't stop on
     // null terminator by itself.
     // As a tiny code save trick, compare idx against maxIdx using a negation,
     // so that maxBytesToRead=undefined/NaN means Infinity.
-    while (heapOrArray[idx] && !(idx >= maxIdx)) ++idx;
-    return idx;
+    heapOrArray.indexOf(0)
```
sbc100: Don't you need a `return` keyword here?

This change seems separate from the TextEncoder change.   Can you submit this as a separate PR?   (I wonder if its results in any code size wins?)

src/lib/libstrings.js
```diff
@@ -162,6 +175,32 @@ addToLibrary({
...
       return 0;
 
+#if TEXTENCODER == 2
+    // Always use TextEncoder when TEXTENCODER == 2
+    var encoded = UTF8Encoder.encode(str);
+    var bytesToWrite = Math.min(encoded.length, maxBytesToWrite - 1); // -1 for null terminator
+    
+    for (var i = 0; i < bytesToWrite; ++i) {
+      heap[outIdx + i] = encoded[i];
+    }
```
sbc100, line 187: Can this loop be replaced with `heap.set(encoded, outIdx);`  ?   

I guess you would also need to do something like: `encoded = encoded.subarray(1, maxBytesToWrite - 1)`?

src/lib/libstrings.js
```diff
@@ -224,7 +264,22 @@ addToLibrary({
...
+#if TEXTENCODER == 2
+    '$UTF8Encoder',
+#endif
+  ],
   $lengthBytesUTF8: (str) => {
+#if TEXTENCODER == 2
+    // Always use TextEncoder when TEXTENCODER == 2
+    return UTF8Encoder.encode(str).length;
+#else
+    // When using conditional TextEncoder, use it for longer strings if available
```
sbc100: I don't see the `use it for longer strings if available` part in the code, I guess either update the comment or the code?

sbc100: I wonder if its even worth doing this here?  i.e. when TEXT_DECODER != 2 maybe we just only do the fallback?  Unless there is a performance advantage?

sbc100 (commented):
Can you update the PR description and title since this is no longer just about `lengthBytesUTF8`.

For the title, can you use the `. NFC` suffix rather than `Refactor:` prefix (this is just our local convention).

src/lib/libstrings.js
```diff
@@ -179,23 +180,17 @@ addToLibrary({
...
-    }
-    // Null-terminate the string
+    encoded = encoded.subarray(0, bytesToWrite)
+    heap.set(encoded, outIdx)
     heap[outIdx + bytesToWrite] = 0;
     return bytesToWrite;
 #else
     // When using conditional TextEncoder, use it for longer strings if available
     if (str.length > 16 && UTF8Encoder) {
       var encoded = UTF8Encoder.encode(str);
```
sbc100, line 193: We could also consider using `encodeInto` using a `subarray` of the heap?

The downside I think is that `envodeInto` was added a little later so might be available in fewer places.

src/lib/libstrings.js
PtPrashantTripathi: changed it

src/lib/libstrings.js
PtPrashantTripathi: done

AUTHORS
PtPrashantTripathi: done

src/lib/libstrings.js
juj: > We could also consider using `encodeInto` using a `subarray` of the heap?

That is a fine idea. I think it's good to work on as a separate PR, if someone wants to. Along with the appropriate `#if MIN_SAFARI_VERSION >= 140500` gates etc.

src/lib/libstrings.js
PtPrashantTripathi: **Yes, there is a performance benefit. The manual method takes more time than the native method. Here are some statistics:**

```sh
❯ node test.js
Benchmarking 1000 runs per method per string...

Testing string:
Native: Total = 0.885 ms, Avg = 88,483 ns
Manual: Total = 2.102 ms, Avg = 210,162 ns
Length Check: Native = 48,860, Manual = 48,860 → ✅ Match
```

src/lib/libstrings.js
```diff
@@ -224,7 +259,22 @@ addToLibrary({
    * @param {string} str - JavaScript string to operator on
    * @return {number} Length, in bytes, of the UTF8 encoded string.
    */
+  $lengthBytesUTF8__deps: [
+#if TEXTENCODER == 2
+    '$UTF8Encoder',
+#endif
+  ],
```
sbc100: This can all be on single line and wrapped in `#if TEXTENCODER == 2`:

```
#if TEXTENCODER == 2
    $lengthBytesUTF8__deps: ['$UTF8Encoder'],
#endif
```

PtPrashantTripathi:
Hi team,

I'd like to propose a change to our `emscripten` string utility library.

Currently, our UTF-8, UTF-16, and UTF-32 conversion functions rely on a lot of custom logic that manually loops through bytes. This approach is not as performant as it could be.

I've outlined a high-level approach to address this:

-----

### High-Level Architectural Change

The core idea is to move away from individual functions with custom loops and introduce dedicated **Processor** objects. These processors will act as a centralized, robust layer for all encoding and decoding operations.

```
+----------------+       +-------------------+
|  Old Approach  |       |  Proposed Approach  |
+----------------+       +-------------------+
|                |       |                   |
|  stringToUTF8  |       |  stringToUTF8     |
|       |        |       |       |           |
|  Manual Loop   |       |   UTF8Processor   |
+----------------+       |                   |
|  UTF8ToString  |       |  UTF8ToString     |
|       |        |       |       |           |
|  Manual Loop   |       |   UTF8Processor   |
+----------------+       +-------------------+
                                   |
                                   |
+------------------------------------------------+
|          UTF8Processor Logic                   |
+------------------------------------------------+
|  Try Native `TextEncoder` / `TextDecoder` API  |
|          (Faster)                              |
|------------------------------------------------|
|  Fallback to JavaScript Loop                   |
|           (Compatability)                      |
+------------------------------------------------+

```

-----

### Conceptual Code Example

This code block illustrates the core concept for a single `UTF8Processor` object. The `decode` and `encode` methods encapsulate the logic, preferring the fast native APIs while maintaining a fallback for older browser compatibility.

```javascript
// A conceptual high-level "Processor" object
const UTF8Processor = (() => {
  // Try to get native browser APIs
  const nativeDecoder = typeof TextDecoder !== 'undefined' ? new TextDecoder('utf-8') : null;
  const nativeEncoder = typeof TextEncoder !== 'undefined' ? new TextEncoder() : null;

  return {
    decode: (byteArray) => {
      if (nativeDecoder) {
        // Use the fast, native API
        return nativeDecoder.decode(byteArray);
      } else {
        // Fallback to manual JS implementation for older browsers
        // ...manual loop logic here...
      }
    },
    encode: (jsString) => {
      if (nativeEncoder) {
        // Use the fast, native API
        return nativeEncoder.encode(jsString);
      } else {
        // Fallback to manual JS implementation
        // ...manual loop logic here...
      }
    }
  };
})();
```

I don't have the skills to implement this refactor myself, but I believe it's an important change for improving performance. Could an existing developer please review this and help get it started?

cc : @juj @sbc100

juj:
> The core idea is to move away from individual functions with custom loops and introduce dedicated Processor objects.

Thanks for the idea. For the system library architecture that Emscripten has, this is unfortunately not a good approach. The system libraries need to be kept flat, C-like to that they generate minimal code size and allow effective automatic dead code elimination.

So object-oriented class-like abstractions do not fit well within these requirements. That is why TextEncoder and TextDecoder are accessed as separate global variables.

Also another requirement is to be able to fall back to manual code when the JS objects are not available. The `UTF8Processor` idea would tie encoder and decoder together, even though they would have been adopted in JavaScript APIs at different times, so gating their use would needlessly be tied to each other.

PtPrashantTripathi:
> Thanks for the idea. For the system library architecture that Emscripten has, this is unfortunately not a good approach. The system libraries need to be kept flat, C-like to that they generate minimal code size and allow effective automatic dead code elimination.

So, I had another thought based on what you said. What if we don't use a single big object at all? We could just add the logic directly to each function. That way, each one can check for its own native API and fall back to the manual code if it needs to. It keeps everything flat and simple.

PtPrashantTripathi:
@juj can you check why those 2 ci check are not running

src/settings.js
```diff
@@ -1776,6 +1776,13 @@ var EVAL_CTORS = 0;
 // [link]
 var TEXTDECODER = 1;
 
+// The default value of 1 means the generated code will use TextEncoder if
+// available and fall back to custom encoding code when it is not available.  
+// If set to 2, we assume TextEncoder is always present and usable, and no
+// fallback JS code will be emitted.
+// [link]
+var TEXTENCODER = 1;
```
sbc100, line 1750: I'm wondering if we should actually just make this an internal setting?   (i.e. move it to `settings_internal.js`).

Since we don't support targeting browsers that don't support TestEncoder at all (See https://caniuse.com/textencoder and https://github.com/emscripten-core/emscripten/blob/cdec9c1d60a760abb4ea88bcbce10948a54baeff/tools/feature_matrix.py#L18-L30).

That way we can just default to `=2` when ENVIRONMENT is web or node and `=1` when we need shell support?

The cost of adding more public options is pretty high (IMHO) so if we can avoid it, even initially, that would be great!

src/lib/libstrings.js
```diff
@@ -243,16 +289,18 @@ addToLibrary({
...
   $intArrayFromString: (stringy, dontAddNull, length) => {
     var len = length > 0 ? length : lengthBytesUTF8(stringy)+1;
-    var u8array = new Array(len);
+    var u8array = new Uint8Array(len);
     var numBytesWritten = stringToUTF8Array(stringy, u8array, 0, u8array.length);
-    if (dontAddNull) u8array.length = numBytesWritten;
-    return u8array;
+    if (dontAddNull) 
+      u8array = u8array.subarray(0, numBytesWritten);
+    return Array.from(u8array);
```
sbc100: Is this change needed?  If so then I guess this change technically changes the interface to `stringToUTF8Array`?   

If this part of the change is unrelated then perhaps is can be removed, or split into its own PR?

src/settings.js
PtPrashantTripathi: Actually, **TEXTDECODER** was public, which is why I kept **TEXTENCODER** in the same place.

src/lib/libstrings.js
PtPrashantTripathi: Yes, actually `intArrayFromString` originally had the functionality to remove the null at the end of the array. To support this old feature, I have kept it here. Without it, some of the test cases are failing.

juj:
Here I think we definitely do want to support a `-sTEXTENCODER=0` option that uses the old synchronous code.

This is because the TextEncoder-enabled paths all allocate temporary garbage. We do want to have a build path that provides as many garbage-free paths as possible. GC stuttering is a very sensitive and difficult thing to optimize for, so being able to clean up as many sources of temp garbage as possible is desirable there.

(Building no-runaway-temp-garbage real-time renderers/games is in general quite feasible with Emscripten, e.g. Unity and Unreal Engine 5 code paths do not generate per-frame temp garbage at all)

juj:
> @juj can you check why those 2 ci check are not running

Hmm, not sure why they are not progressing. CircleCI is a bit finicky at times. I generally push some new change commit to the PR to make the CircleCI backend re-do the whole thing.

src/lib/libstrings.js
sbc100: Can you explain more?

If the behaviour of `stringToUTF8Array` is unchanged then it seems like that code in `intArrayFromString` would not need changing.

If the behaviour of `stringToUTF8Array` *has* changed in some way can you explain exactly how its has changed?

sbc100:
> Here I think we definitely do want to support a `-sTEXTENCODER=0` option that uses the old synchronous code.
> 
> This is because the TextEncoder-enabled paths all allocate temporary garbage. We do want to have a build path that provides as many garbage-free paths as possible. GC stuttering is a very sensitive and difficult thing to optimize for, so being able to clean up as many sources of temp garbage as possible is desirable there.
> 
> (Building no-runaway-temp-garbage real-time renderers/games is in general quite feasible with Emscripten, e.g. Unity and Unreal Engine 5 code paths do not generate per-frame temp garbage at all)

In that case I'm not sure its worth adding this complexity.   i.e. if the current code has advantages in terms of code size and advantages in terms of not generating garbage, then maybe the perf win in some cases is not worth the extra compexity of adding a new setting (and the extra dimension in the testing matix).    Presumably most programs are not bottlenecked on converting JS strings to linear memory?

sbc100:
> (Building no-runaway-temp-garbage real-time renderers/games is in general quite feasible with Emscripten, e.g. Unity and Unreal Engine 5 code paths do not generate per-frame temp garbage at all)

If we were able to use `encodeInto` would that avoid the garbage?  I guess not since it would involve calling `heap.subarray(..)` to ceate the target array?   Are the engines able to avoid creating garbage for transient subarrays?

src/lib/libstrings.js
PtPrashantTripathi: In the `stringToUTF8Array`, we updated the logic to use the `set` method on the `heap` instead of looping through and assigning values one by one. This works only when both the source and target are `TypedArray` objects like `Uint8Array`.

So for that Previously in here, we were using a regular `Array(len)` for the `u8array heap`, which was incorrect since `set` wouldn't work in that case. Now, we explicitly create it as a `Uint8Array` like this:

```js
var u8array = new Uint8Array(len);
```

Additionally, earlier we had:

```js
if (dontAddNull) u8array.length = numBytesWritten;
```

I changed it to:

```js
if (dontAddNull) 
  u8array = u8array.subarray(0, numBytesWritten);
```

This ensures we properly handle cases where the null terminator should be removed, which also fixes some failing test cases.

PtPrashantTripathi:
> > @juj can you check why those 2 ci check are not running
> 
> Hmm, not sure why they are not progressing. CircleCI is a bit finicky at times. I generally push some new change commit to the PR to make the CircleCI backend re-do the whole thing.

@juj It looks like the GitHub Actions workflow is currently blocked and waiting for maintainer approval.

Here’s the link to the pending run: [https://github.com/emscripten-core/emscripten/actions/runs/17455947522](https://github.com/emscripten-core/emscripten/actions/runs/17455947522?utm_source=chatgpt.com)

Could you please take a look and approve it so the checks can proceed? since all the other checks are successfully completed

src/lib/libstrings.js
sbc100: I see, in that case I'm not sure we can just switch to using the `.set()` method, since that would be a breaking change to `stringToUTF8Array`.  i.e. it would mean the function no longer accepts `Array` objects, only `TypedArray` objects.

Is that right?

Its not impossible to imagine us changing that interface, but if we do I think we should do that first, as a separate change.

sbc100:
I don't see any maintainers approval button so I'm not sure what is blocking those github actions.

src/lib/libstrings.js
PtPrashantTripathi: #### Yes you are correct, I'll just revert back the change to accept both array and typed-array 



Meanwhile quick question what is the best way to run all the test case in local I am having a mac - so is there any just a quick test run command or somthing 
I have tried that test/runner script but it's running for more then 1.5 hour in my machine and still running...

PtPrashantTripathi:
> I don't see any maintainers approval button so I'm not sure what is blocking those github actions.

It might look somthing like this
Screenshot_2025-09-05-02-54-21-309_com android chrome-edit

Source: https://docs.github.com/en/copilot/how-tos/use-copilot-agents/coding-agent/review-copilot-prs#managing-github-actions-workflow-runs

src/lib/libstrings.js
sbc100: I normally run `./test/runner other --skip-slow` and `./test/runner core2 --skip-slow` locally.. that gives pretty good coverage and both commands should return pretty quickly.

sbc100:
> It might look somthing like this

Yup, I know that it looks like since it shows up for every new contributor.  It was missing for some reason, but its there now and I just pressed it.

src/lib/libstrings.js
```diff
@@ -162,6 +176,28 @@ addToLibrary({
...
+    var encoded = UTF8Encoder.encode(str);
+    var bytesToWrite = Math.min(encoded.length, maxBytesToWrite - 1); // -1 for null terminator
+    for (var i = 0; i < bytesToWrite; ++i) {
+      heap[outIdx + i] = encoded[i];
+    }
+    heap[outIdx + bytesToWrite] = 0;
+    return bytesToWrite;
+#else
+    // When using conditional TextEncoder, use it for longer strings if available
+    if (UTF8Encoder) {
```
sbc100, line 192: The comment says its only being used to longer strings but the code doesn't do that.   Do we want to do it?

PtPrashantTripathi: done

PtPrashantTripathi:
Hi @sbc100 can you approve the pr now

sbc100:
As I mentioned in https://github.com/emscripten-core/emscripten/pull/25148#issuecomment-3254123942 I'm just not sure about this change anymore.

It seems like it regresses code size while adding an addition setting (i.e. complexity which will cost us in the long run).   The benefit is performance but I'm not sure many real world apps are bottle-necked on this code.

PtPrashantTripathi:
Got it


Thank and Regards,

Pt. Prashant Tripathi
Data Engineer | Cloud and Data Technology
📧 Email: ***@***.***
📱 Mobile: +91-7223047175, +91-8435782545
📍 Location: Pune, India

________________________________

sbc100:
@juj WDYT?

I'm leaning no in the current form.

If we could avoid the size regression I would certainly change my mind.

I have couple of ideas for that:

1. Avoid the =2 mode, and only support =0 (never use it) or =1 (always use it).
2. Remove support non-TypedArray inputs so we can use `heap.set()` again.   This is something we could do accoss the board for the string API, where we have extra codepaths to handle the non-TypedArray path.

PtPrashantTripathi:
cant work on this anymore closing this PR

PR #25149, opened 2025-09-03
Fix export of unsigned WebAssembly globals when they grow > 2GB in size.

juj:
Fix export of unsigned WebAssembly globals when they grow > 2GB in size. Fix DECLARE_ASM_MODULE_EXPORTS=0 mode globals export.

Fixes test `core_2gb.test_memprof_requirements`.

### DIFF
--- a/src/lib/libcore.js
+++ b/src/lib/libcore.js
@@ -1629,17 +1629,23 @@ addToLibrary({
     var global_object = this;
 #endif
 
-    for (var __exportedFunc in wasmExports) {
-      var jsname = asmjsMangle(__exportedFunc);
+    for (var [name, exportedSymbol] of Object.entries(wasmExports)) {
+      name = asmjsMangle(name);
 #if DYNCALLS || !WASM_BIGINT
-      if (jsname.startsWith('dynCall_')) {
-        dynCalls[jsname.substr(8)] = wasmExports[__exportedFunc];
+      if (name.startsWith('dynCall_')) {
+        dynCalls[name.substr(8)] = exportedSymbol;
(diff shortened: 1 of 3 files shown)


juj:
CircleCI looks like is out of order for more than an hour now. Force-landing to save time and get Google CI green: https://ci.chromium.org/ui/p/emscripten-releases/builders/ci/linux-test-suites/b8704775493462556353/overview and to verify this fully on my CI.

Taking reviews after.

src/lib/libcore.js
```diff
@@ -1629,18 +1629,23 @@ addToLibrary({
...
-      if (jsname.startsWith('dynCall_')) {
-        dynCalls[jsname.substr(8)] = wasmExports[__exportedFunc];
+      if (name.startsWith('dynCall_')) {
+        dynCalls[name.substr(8)] = exportedSymbol;
       }
 #endif
+      // Globals are currently statically enumerated into the output JS.
+      // TODO: If the number of Globals grows large, consider giving them a
+      // similar DECLARE_ASM_MODULE_EXPORTS = 0 treatment.
+      if (typeof exportedSymbol.value === 'undefined') {
```
sbc100, line 1642: `(!exportedSymbol instanceof WebAssembly.Global)` might be more explicit?

sbc100 (commented):
Nice!

src/lib/libcore.js
juj: That is what I had first, but it won't work in wasm2js mode, because the polyfilled objects failed to pass that test.

PR #25153, opened 2025-09-03
Fix regression caused by #25141

juj:
Fix regression caused by https://github.com/emscripten-core/emscripten/pull/25141/files#diff-2ff8ac5244693c408612e27b5740480b110cb2187ef498ee73ad34b19e906169L20

(bad merge?)

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -17,6 +17,7 @@ import io
 import itertools
 import json
 import logging
+import multiprocessing
 import threading
 import os
 import re


sbc100:
I think the issue was that #25141 and #25066 raced and were never tested together. 

#25141 removed the last reference to `multiprocessing` but then #25066 added a fresh one

test/common.py
```diff
@@ -17,6 +17,7 @@
 import itertools
 import json
 import logging
+import multiprocessing
 import threading
```
brendandahl: This line needs to be removed. Looks like I removed the wrong line. Confused how my pr passed though...

juj:
The regression is e.g. here: http://clbri.com:8010/#/builders/11/builds/412

test/common.py
juj: I think that is because CircleCI tests PR branches as-is without merging to main.

So a PR on top of main can pass, but then main gets other PRs landed in the meanwhile (like my PR that added use of multithreading).

Individually both PRs get a green stamp on CircleCI, but when merged together, they are not compatible.

PR #25158, opened 2025-09-03
Remove `shell=True` from `rebaseline_tests.py`. NFC

sbc100:
This was added in #25142 to fix windows but inadvertently broke linux.

Turns out the `shell=True` means that first argument should be a single string, and not and array of arguments.  For some reason using an array along with `shell=True` was causing `subprocess.check_call` to run `test/runner` only without any aguments.


juj (approved):
Great catch - I'll have to remember that gotcha.

kripken (approved):

Issue #25162, opened 2025-09-03
-sJS_MATH duplicate symbols against libc


juj:
Running with `-sJS_MATH` is giving trouble:

```
wasm-ld: error: duplicate symbol: acos
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libjsmath.a(jsmath.o)
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libc.a(acos.o)

wasm-ld: error: duplicate symbol: acosf
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libjsmath.a(jsmath.o)
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libc.a(acosf.o)

wasm-ld: error: duplicate symbol: asin
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libjsmath.a(jsmath.o)
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libc.a(asin.o)

wasm-ld: error: duplicate symbol: asinf
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libjsmath.a(jsmath.o)
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libc.a(asinf.o)

wasm-ld: error: duplicate symbol: atan
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libjsmath.a(jsmath.o)
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libc.a(atan.o)

wasm-ld: error: duplicate symbol: atan2
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libjsmath.a(jsmath.o)
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libc.a(atan2.o)

wasm-ld: error: duplicate symbol: atan2f
>>> defined in C:\emsdk\emscripten\main\cache\sysroot\lib\wasm32-emscripten\pic\libjsmath.a(jsmath.o)
[...]

PR #25163, opened 2025-09-03
Add jsmathz test suite, and skip tests that are failing in it.

juj:
Add jsmathz test suite, and skip tests that are failing in it. https://github.com/emscripten-core/emscripten/issues/25162

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -390,6 +390,7 @@ no_strict = make_no_decorator_for_setting('STRICT')
 no_strict_js = make_no_decorator_for_setting('STRICT_JS')
 no_big_endian = make_no_decorator_for_setting('SUPPORT_BIG_ENDIAN')
 no_omit_asm_module_exports = make_no_decorator_for_setting('DECLARE_ASM_MODULE_EXPORTS=0')
+no_js_math = make_no_decorator_for_setting('JS_MATH')
 
 
 def is_sanitizing(args):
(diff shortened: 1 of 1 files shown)


test/test_core.py
```diff
@@ -1002,6 +1004,7 @@ def test_longjmp2(self):
 
   @needs_dylink
   @with_all_sjlj
+  @no_js_math('TODO: https://github.com/emscripten-core/emscripten/issues/25162')
```
sbc100: `JS_MATH is not compatible with MODULE_MODULE=1`.

Can you also add an error to link.py if these are used together?

juj: Updated.

tools/link.py
```diff
@@ -1088,6 +1088,14 @@ def limit_incoming_module_api():
 
   # Use settings
 
+  if settings.JS_MATH:
+    if settings.MAIN_MODULE:
```
sbc100: Only `MAIN_MODULE == 1` .. not the other two conditions here.

test/test_core.py
```diff
@@ -9675,6 +9688,7 @@ def test_stress_proxy_to_pthread_hello_world(self):
     self.parallel_stress_test_js_file(js_file, assert_returncode=0, expected='hello, world!', not_expected='error')
 
   @needs_dylink
+  @no_js_math('JS_MATH is not compatible with MAIN_MODULE')
```
sbc100, line 9691: `=1` specifically.  `=2` has no issue I believe.

tools/link.py
juj: Currently I see it doesn't work with any of these combinations.

test/test_core.py
juj: Check

PR #25164, opened 2025-09-04
Add test framework logging function. NFC

sbc100:
Followup to #25081

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -155,6 +155,14 @@ if not config.JS_ENGINES:
 requires_network = unittest.skipIf(os.getenv('EMTEST_SKIP_NETWORK_TESTS'), 'This test requires network access')
 
 
+def errlog(*args):
+  """Shorthand for print with file=sys.stderr
+
+  Use this for all internal test framework logging..
+  """
+  print(*args, file=sys.stderr)
+
+
 def load_previous_test_run_results():
   try:
     return json.load(open(PREVIOUS_TEST_RUN_RESULTS_FILE))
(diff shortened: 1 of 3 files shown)


test/common.py
```diff
@@ -155,6 +155,10 @@ def configure(data_dir):
 requires_network = unittest.skipIf(os.getenv('EMTEST_SKIP_NETWORK_TESTS'), 'This test requires network access')
 
 
+def log(msg):
+  print(msg, file=sys.stderr)
```
juj: maybe name this `loge()` or similar, otherwise it could be confusing to read at call sites.

juj:
This is both a refactor (print -> log) and a functional change (logging is now all stderr logging)?

Err.. LGTM either way

sbc100:
> This is both a refactor (print -> log) and a functional change (logging is now all stderr logging)?

I thought you said when you added the `file=sys.stderr` stuff that you wanted the logging from the test framework itself to all go to stderr?

juj:
> I thought you said when you added the `file=sys.stderr` stuff that you wanted the logging from the test framework itself to all go to stderr?

Yeah, that's ok change to me.

test/common.py
```diff
@@ -155,6 +155,10 @@ def configure(data_dir):
 requires_network = unittest.skipIf(os.getenv('EMTEST_SKIP_NETWORK_TESTS'), 'This test requires network access')
 
 
+def errlog(*args):
```
kripken, line 158: Maybe add a comment mentioning when this should be used?

PR #25166, opened 2025-09-04
Revise -sJS_MATH errors

juj:
Removing these two to see how they work out on the CI.


juj:
Looks good - landing to run on CI. (this is only currently exercised on my bots)

sbc100:
Thanks!

BTW, when I land these kind of small changes I normally use the chrome convention of saying something like "Landing TBR" (which stands for To Be Reviewed)

PR #25169, opened 2025-09-04
Replace emval_get_global. NFC

sbc100:
- Use a constant `emGlobalThis` instead of a function call
- Calculate the constant once at startup time if its needed.

The plan is to then use this library symbol whenever we need access to globalThis.

### DIFF
--- a/src/lib/libemval.js
+++ b/src/lib/libemval.js
@@ -122,50 +122,48 @@ var LibraryEmVal = {
   _emval_new_u16string__deps: ['$Emval'],
   _emval_new_u16string: (v) => Emval.toHandle(UTF16ToString(v)),
 
+  $emGlobalThis__internal: true,
 #if SUPPORTS_GLOBALTHIS
-  $emval_get_global: () => globalThis,
-#elif !DYNAMIC_EXECUTION
-  $emval_get_global: () => {
-    if (typeof globalThis == 'object') {
+  $emGlobalThis: 'globalThis',
+#else
+  $getGlobalThis__internal: true,
+  $getGlobalThis: () => {
+    if (typeof globalThis != 'undefined') {
       return globalThis;
     }
+#if DYNAMIC_EXECUTION
+    return new Function('return this')();
+#else
     function testGlobal(obj) {
-      obj['$$$embind_global$$$'] = obj;
-      var success = typeof $$$embind_global$$$ == 'object' && obj['$$$embind_global$$$'] == obj;
-      if (!success) {
-        delete obj['$$$embind_global$$$'];
-      }
+      // Use __emGlobalThis as a test symbol to see if `obj` is indeed the
+      // global object.
+      obj['__emGlobalThis'] = obj;
+      var success = typeof __emGlobalThis == 'object' && obj['__emGlobalThis'] === obj;
+      delete obj['__emGlobalThis'];
       return success;
     }
-    if (typeof $$$embind_global$$$ == 'object') {
-      return $$$embind_global$$$;
+    if (typeof self != 'undefined' && testGlobal(self)) {
+      return self; // This works for both "window" and "self" (Web Workers) global objects
     }
-    if (typeof global == 'object' && testGlobal(global)) {
-      $$$embind_global$$$ = global;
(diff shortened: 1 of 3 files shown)


sbc100:
Ultimately I think we can remove this function completely since we can remove support for `SUPPORTS_GLOBALTHIS == false`.. but for now this is still an improvment.

sbc100:
Followup: https://github.com/emscripten-core/emscripten/pull/25172

test/test_core.py
```diff
@@ -7504,8 +7504,10 @@ def test2():
     do_test(test2, level=2, prefix='hello_libcxx')
 
   @parameterized({
-    '': (['-lembind', '-sDYNAMIC_EXECUTION=0'],),
+    '': (['-lembind'],),
```
RReverser, line 7507: Was this an accidental change? There's a separate test for non-dynamic mode below.

src/lib/libemval.js
```diff
@@ -122,50 +122,46 @@ var LibraryEmVal = {
...
+#if DYNAMIC_EXECUTION
+    return new Function('return this')();
+#else
     function testGlobal(obj) {
-      obj['$$$embind_global$$$'] = obj;
-      var success = typeof $$$embind_global$$$ == 'object' && obj['$$$embind_global$$$'] == obj;
-      if (!success) {
-        delete obj['$$$embind_global$$$'];
-      }
+      // Use __emGlobalThis as a test symbol to see if `obj` is indeed the
```
RReverser, line 138: I don't think we need this at all tbh.

src/lib/libemval.js
```diff
@@ -122,50 +122,46 @@ var LibraryEmVal = {
...
+      // global object.
+      obj['__emGlobalThis'] = obj;
+      var success = typeof __emGlobalThis == 'object' && obj['__emGlobalThis'] === obj;
+      delete obj['__emGlobalThis'];
       return success;
     }
-    if (typeof $$$embind_global$$$ == 'object') {
-      return $$$embind_global$$$;
+    if (typeof global != 'undefined' && testGlobal(global)) {
```
kripken: The change from `== 'object'` (a few lines below) to `!= 'undefined'` (here) may run into problems if `global = 42`? Writing a field to a Number traps.

I guess this trapped before on `null`, though...

test/test_core.py
sbc100: Yes, I split it out into and explicit "no_dynamic" variant.

src/lib/libemval.js
RReverser: If you want, you can also wrap this into conditional compilation since `global` only exists on Node.js, and `self` elsewhere.

test/test_core.py
RReverser: 🤦🏼‍♂️ It's been a long day.

I was actually thinking of https://github.com/sbc100/emscripten/blob/67c1f24c42d562add252630836f94924bb0eca33/test/test_other.py#L3357.

src/lib/libemval.js
sbc100: Yeah, I changed that because I noticed that ` == 'object'` was begin lowered to something more verbose with `LEGACY_VM_SUPPORT` enabled.  I don't think its a big deal

RReverser: Yeah if someone declares a variable `global` that is not global, they're asking for trouble IMO. A lot of Node scripts refer to it without additional checks.

src/lib/libemval.js
sbc100: Sure, maybe you are right.   I'm not sure it worth further iteration though since this code is likely going to be delete soon.

src/lib/libemval.js
sbc100: Done

test/test_core.py
sbc100: In this case we need to combine both `-sDYNAMIC_EXECUTION=0` and `-sLEGACY_VM_SUPPORT` to test the long code path.

Also, doing it in core means it gets testings with `-O1` .. `-ON` and also with `--closure=1` enabled.

kripken (approved):
Sounds good.

PR #25172, opened 2025-09-04
Replace globalThis polyfill in exportWasmSymbols. NFC 

sbc100:
This polyfill is code a duplicate of the `emGlobalThis` polyfill.

### DIFF
--- a/src/lib/libcore.js
+++ b/src/lib/libcore.js
@@ -1650,20 +1650,12 @@ addToLibrary({
 #if !DECLARE_ASM_MODULE_EXPORTS
   // When DECLARE_ASM_MODULE_EXPORTS is not set we export native symbols
   // at runtime rather than statically in JS code.
-  $exportWasmSymbols__deps: ['$asmjsMangle'
+  $exportWasmSymbols__deps: ['$asmjsMangle', '$emGlobalThis',
 #if DYNCALLS || !WASM_BIGINT
     , '$dynCalls'
 #endif
   ],
   $exportWasmSymbols: (wasmExports) => {
-#if ENVIRONMENT_MAY_BE_NODE && ENVIRONMENT_MAY_BE_WEB
-    var global_object = (typeof process != "undefined" ? global : this);
-#elif ENVIRONMENT_MAY_BE_NODE
-    var global_object = global;
-#else
-    var global_object = this;
-#endif
-
     for (var [name, exportedSymbol] of Object.entries(wasmExports)) {
       name = asmjsMangle(name);
 #if DYNCALLS || !WASM_BIGINT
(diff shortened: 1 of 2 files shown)


src/lib/libcore.js
```diff
@@ -1612,23 +1612,50 @@ addToLibrary({
   emscripten_asm_const_async_on_main_thread: (emAsmAddr, sigPtr, argbuf) => runMainThreadEmAsm(emAsmAddr, sigPtr, argbuf, 0),
 #endif
 
+  $emGlobalThis__internal: true,
+#if SUPPORTS_GLOBALTHIS
+  $emGlobalThis: '=globalThis',
```
juj: It would be great in `SUPPORTS_GLOBALTHIS` mode to not have a `emGlobalThis` variable at all. The existence of aliasing `var emGlobalThis = globalThis;` is icky.

src/lib/libcore.js
```diff
@@ -1612,23 +1612,50 @@ addToLibrary({
...
+      return success;
+    }
+    if (typeof self != 'undefined' && testGlobal(self)) {
+      return self; // This works for both "window" and "self" (Web Workers) global objects
+    }
+#if ENVIRONMENT_MAY_BE_NODE
+    if (typeof global != 'undefined' && testGlobal(global)) {
+      return global;
+    }
+#endif
```
juj: Which usecase does this fix that was broken before? I'm not completely sure I follow.

(Are we missing a test case that would uncover the missing scenario?)

sbc100: I noticed we had two different polyfills for `globalThis` so this change unifies them.

This code here is not new code its just moved from embind to here.

It also removed the unnecessary polyfill in the case of `SUPPORTS_GLOBALTHIS` because the polyfil in `exportWasmSymbols` was always active before, not its conditional which should be code size win when `SUPPORTS_GLOBALTHIS` is enabled.

sbc100: > Which usecase does this fix that was broken before? I'm not completely sure I follow.
> 
> (Are we missing a test case that would uncover the missing scenario?)

Yes we could maybe find a way around that.. but its not as bad as the current `var global_object = this;` which is just broken in the case of strict JS mode (of in the case of EXPORT_ES6).

sbc100: I think in order demonstrate the current broken code you would need test that run in the browser and does `-sDECLARE_ASM_MODULE_EXPORTS=0 + -sSTRICT_JS` or `-sDECLARE_ASM_MODULE_EXPORTS=0 + -sEXPORT_ES6`

juj: Oh so I suppose it would be the combination of STRICT_JS + DECLARE_ASM_MODULE_EXPORTS=0 then that is not covered?

I wonder if we should add one test to verify that combination specifically, and that would then warrant this uglier form here.

sbc100:
I split out the moving of the function so it can be separated from the usage with `exportWasmSymbols`.

src/lib/libcore.js
sbc100: It not just that combination, it also has to run in the browser and not in node, since the node case will always work.

I will add a new browser test.

sbc100:
OK, so I was wrong about the strict JS mode.  Because `exportWasmSymbols` is an arrow function it actual *does* get its lexically scoped `this` .. strange.

I added a new test anyway just to confirm.

I think its still probably work landing this, if only to remove the duplication here.

PR #25174, opened 2025-09-04
Adjust comment on test_pthread_dlopen_many

juj:
Presumably it passes on wasm64 occassionally...?

https://github.com/emscripten-core/emscripten/issues/18887


sbc100 (commented):
I think the reason it was there is that it was too flaky under wasm64.. but that was a while ago so maybe things are better now?

juj:
Looks like it is still too flaky in wasm64, put that skip back.

sbc100 (approved):

Issue #25178, opened 2025-09-04
Cache out of date


juj:
The previous fix to emmalloc trim resulted in code cache needing to be cleared.

Can I run the [Tag release and update Changelog](https://github.com/emscripten-core/emscripten/actions/workflows/tag-release.yml) action to do that? Just put current SHA of main branch as the Changeset there?

sbc100:
We just releases 4.0.14 two days ago, but if you want to go another release.

We don't normally trigger a release just because we change something in one of the system libraries (which always requires a cache clear).    For git users we instead normally just rely on an "if in doubt clear the cache and try again" approach.

sbc100:
The `Tag release and update Changelog` action is designed to be triggered at the end of the chain of events the make up a release.   Specifically it gets triggered automatically from the emsdk change that tags a release, which in turn is automatically triggered when the emscripten-releases builders have all finished building a release.

The release is initially triggered by a human running https://chromium.googlesource.com/emscripten-releases/+/refs/heads/main/src/create_release_candidate.py

juj:
I see.. back in the day we would bump the version number whenever a cache clear was needed. (that was the idea of tying cache clear to the versioning)

On my CI I have made the bots test each commit on an existing cache, so they would rely on that being up to date. (it saves a nice amount of electricity to not have to rebuild the system libs on every commit)

sbc100:
If you are testing each commit the safest thing to do to clear the cache each time.

If we bumped the version every time we touched any header or source that gets built into a system library that would result in a lot of version bumps a believe.    We don't currently do that.

juj:
Maybe we could take the last modified timestamp of the `system/` directory as part of the cache hash? Then if something is changed there as part of a git pull, then it'd trigger a cache clear as well?

sbc100:
I suppose you could design a system that forces the cache to be cleaned if any file in the `system/` directory was newer than the oldest file in the cache?  But you would need to remember that changes to the compile driver itself can change compile and link flags so any change to `tools/` would probably have to trigger it too.

juj:
Hmm I'd imagine `tools/system_libs.py` would cover that well.

Rerunning test suite on my Linux bot takes ~30 minutes on a warm cache, and close to 50 minutes on a clean cache. It is quite a big difference, and a lot (majority) of our PRs that land are not touching the system libraries.

sbc100:
[...]

PR #25179, opened 2025-09-04
Add script tools/heuristic_clear_cache.py

juj:
Add script tools/heuristic_clear_cache.py that can be used as a heuristic to detect if cache clear might be necessary.

Resolves https://github.com/emscripten-core/emscripten/issues/25178.

### DIFF
--- a/tools/maint/heuristic_clear_cache.py
+++ b/tools/maint/heuristic_clear_cache.py
@@ -58,5 +58,6 @@ def heuristic_clear_cache():
   else:
     print('Cache timestamp is up to date, no clear needed.')
 
+
 if __name__ == '__main__':
     sys.exit(heuristic_clear_cache())


tools/heuristic_clear_cache.py
```diff
@@ -0,0 +1,62 @@
...
+  return mtime
+
+
+def newest_mtime(paths):
+  return max((recursive_mtime(path) for path in paths), default=0)
+
+
+def heuristic_clear_cache():
+  try:
+    system_libs_mtime = open(utils.path_from_root('cache', 'system_libs_mtime.txt')).read()
```
sbc100: You probably and to use `cache.cachdir` here since the cache isn't always in this default location.

tools/heuristic_clear_cache.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
```
sbc100: Can you put this in `tools/maint/..` which is where we put stuff for maintainers and/or git users.   This will also prevent this tool being shipping in emsdk, which is good.

sbc100 (approved):

tools/heuristic_clear_cache.py
juj: Fantastic, thanks.

tools/heuristic_clear_cache.py
juj: check

PR #25180, opened 2025-09-04
Fix failfast regression

juj:
Fix failfast regression from https://github.com/emscripten-core/emscripten/pull/25069


juj:
Example of failing run: http://clbri.com:8010/#/builders/11/builds/472

brendandahl:
This code still needs run. It just needs to be moved up into the end of `with multiprocessing.Pool(`

juj:
> This code still needs run. It just needs to be moved up into the end of `with multiprocessing.Pool(`

Could you submit a proper fix?

juj:
Not the right one..

PR #25185, opened 2025-09-04
Fix heuristic_clear_cache.py

juj:
Moving one directory deeper requires one more `..` in import, and fix cache read.


juj:
ruff is happy, landing to test on CI.

PR #25186, opened 2025-09-04
Demote assert to print

juj:


juj:
To save time, I'll force-land this to unblock my runners, they will also then print out the values here for further debugging.

brendandahl:
This is happening with fail-fast?

juj:
> This is happening with fail-fast?

Yes, that's right. I run `--failfast` on my bots.

juj:
http://clbri.com:8010/api/v2/logs/46878/raw_inline is saying

```
Expected 256 teardowns, got 77
```

juj:
Even though that is `--failfast`, there are no failures in the suite, but all the tests came out green (as in the above log)

test/parallel_testsuite.py
```diff
@@ -136,7 +136,8 @@ def run(self, result):
         # are cycled through once.
         num_tear_downs = sum([pool.apply(tear_down, ()) for i in range(use_cores)])
         # Assert the assumed behavior above hasn't changed.
-        assert(num_tear_downs == use_cores)
+        if num_tear_downs != use_cores:
```
sbc100, line 139: I think this only really matters for browser tests, right brendan?  since teardown doesn't mean anything otherwise?

Could we restore the assert but put it behind `contains_browser_test`?

juj: Wouldn't that then fail when I run browser tests on my CI?

juj: It seems rather that the logic is not sound that scheduling necessarily works that every thread gets one task.

Would it be possible to have a multiprocessing shared data structure where all the tearable items are added, and then the main thread would tear them down once the pool is done?

sbc100: We looked into a lot of options.   The problem with shared data structures is that python bug.. we can't always use them right?

brendandahl:
Does your ci bot really have 256 threads?

juj:
Yes, it is running on 2x AMD Epyc 7763. (Supermicro H21DSi-N6 and 512GB of DDR4)

test/parallel_testsuite.py
juj: That is true :/

PR #25187, opened 2025-09-05
[emsymbolizer] Move emsymbolizer.py to tools/ (NFC)

aheejin:
This way the functions can be possibly reused from files from `tools/`. Because we add it to https://github.com/emscripten-core/emscripten/blob/main/tools/maint/create_entry_points.py, users can access it from anywhere.


sbc100 (commented):
I'm tempted to say just move the entire `emsymbolizer.py` into `tools/` and then add emsymbolizer to the `entry_remap` in `create_entry_points.py`.

One less file at the top level directory then, which is a good thing IMHO.

aheejin:
@sbc100 Thanks! Didn't know about that.

dschuff (approved):

PR #25194, opened 2025-09-05
[ci] Increase EMCC_CORES for test-wasm64-4gb test suite. NFC

sbc100:
Since the CI didn't report any OOMs for this suite I assume this should be safe to land.

Test time goes from ~35 or ~21 minutes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -662,8 +662,6 @@ jobs:
   test-wasm64-4gb:
     environment:
       LANG: "C.UTF-8"
-      # Only run 2 tests at a time to avoid OOM (since each test used >4gb)
-      EMCC_CORES: "2"
     # We don't use `bionic` here since its too old to run recent node versions:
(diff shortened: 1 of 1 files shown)


juj:
> Since the CI didn't report any OOMs

I don't think there is an actual >4 GB memory allocation in this suite per VM even with the 4gb bump, but the memory usage stays virtual. I run the wasm64-4gb suite parallel on 256 threads on my linux box, and iirc it consumed only about 30GB of RAM at peak during the suite.

sbc100:
> > Since the CI didn't report any OOMs
> 
> I don't think there is an actual >4 GB memory allocation in this suite per VM even with the 4gb bump, but the memory usage stays virtual. I run the wasm64-4gb suite parallel on 256 threads on my linux box, and iirc it consumed only about 30GB of RAM at peak during the suite.

It certainly used to OOM in the past, which is why this limit was there.. but it seems OK now... at least for this sample size of 1 :-/

Issue #25012, opened 2025-08-21
Single-file JS-only file has reference to non-existing .wasm file (since 4.0.6)


pasieronen:
I'm building a single-file JS-only file (-sWASM=0 -sSINGLE_FILE=1), which eventually gets bundled with Webpack. With 4.0.5 everything works fine, but with 4.0.6, the resulting .js file has a reference to a .wasm file (which obviously does not exist with these options), and Webpack chokes on that.

The probable culprit is commit 975084c.

**How to reproduce:**

adder.c
```c
int add(int a, int b) { return a + b; }
```

index.js
```js
import createAdder from './adder.mjs'
(async () => {
  const Module = await createAdder()
  const add = Module.cwrap("add", "number", ["number","number"]);
  alert("2+3 = " + add(2, 3))
})();
```

webpack.config.cjs

```js
module.exports = {
  mode: "development",
  entry: "./index.js",
  output: {filename: "bundle.js", path: __dirname}
}
````

Commands:

```
emcc -sENVIRONMENT=web -sWASM=0 -sMODULARIZE=1 -sEXPORT_ES6=1 -sSINGLE_FILE=1 -sEXPORTED_FUNCTIONS='["_add"]' -sEXPORTED_RUNTIME_METHODS='["cwrap", "ccall"]'  adder.c -o adder.mjs
# succeeds, nothing special in output

npx webpack-cli@6.0.1
# fails with "Module not found: Error: Can't resolve 'adder.wasm' ...'
```

**Version of emscripten/emsdk:**

emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.6 (1ddaae4d2d6dfbb678ecc193bc988820d1fc4633)
clang version 21.0.0git (https:/github.com/llvm/llvm-project 4775e6d9099467df9363e1a3cd5950cc3d2fde05)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: (omitted)

sbc100:
Can you confirm if it happens with the latest release too?  (`emsdk install tot` or `emsdk install 4.0.13`)

pasieronen:
Yes, this happens with 4.0.13, too.

sbc100:
Thanks, I was able to reproduce locally.   In practice the URL is not actually used, but the bundler still sees it.  I will see I can find a way to omit this.

sbc100:
[...]

PR #25195, opened 2025-09-05
Fix for webpack + SINGLE_FILE + WASM=0

sbc100:
Without this fix the output code contains a references to the non-existent wasm file URL.

Fixes: #25012

### DIFF
--- a/src/preamble.js
+++ b/src/preamble.js
@@ -413,8 +413,17 @@ function instrumentWasmTableWithAbort() {
 #if !SOURCE_PHASE_IMPORTS && !WASM_ESM_INTEGRATION
 var wasmBinaryFile;
 
+#if WASM2JS && WASM != 2
+
+// When building with wasm2js these 3 functions all no-ops.
+function findWasmBinary(file) {}
+function getBinarySync(file) {}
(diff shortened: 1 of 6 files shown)


src/preamble.js
```diff
@@ -413,8 +413,17 @@ function instrumentWasmTableWithAbort() {
 #if !SOURCE_PHASE_IMPORTS && !WASM_ESM_INTEGRATION
 var wasmBinaryFile;
 
+#if WASM2JS && WASM != 2
+
+// When building with wasm2js these 3 functions all all no-ops.
```
brendandahl:
```suggestion
// When building with wasm2js these 3 functions all are no-ops.
```

sbc100:
@brendandahl any idea what these firefox failures might be coming from?  They seem unrelated to this change.

brendandahl:
I think they started being flaky when we had two many test runners.

juj:
If you're referring to the # of browsers, I now landed the https://github.com/emscripten-core/emscripten/pull/25220 CircleCI fix.

sbc100:
OK, ignoring those failures since I don't see how they could be related.

PR #25201, opened 2025-09-05
Do not halve the number of used cores in testing

juj:
Do not halve the number of used cores if the user has explicitly specified the number of cores to use in testing.

It is a bit silly to have to say `export EMTEST_CORES=128` to get 64 cores in actuality.

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -33,7 +33,7 @@ def python_multiprocessing_structures_are_buggy():
 
 
 def cap_max_workers_in_pool(max_workers, is_browser):
-  if is_browser and os.getenv('EMTEST_CORES') is None:
+  if is_browser and 'EMTEST_CORES' not in os.environ and 'EMCC_CORES' not in os.environ:
     # TODO experiment with this number. In browser tests we'll be creating
     # a chrome instance per worker which is expensive.
     max_workers = max_workers // 2
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -33,10 +33,10 @@ def python_multiprocessing_structures_are_buggy():
 
 
 def cap_max_workers_in_pool(max_workers, is_browser):
-  if is_browser:
+  if is_browser and os.getenv('EMTEST_CORES') is None:
     # TODO experiment with this number. In browser tests we'll be creating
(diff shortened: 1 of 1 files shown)


test/parallel_testsuite.py
```diff
@@ -33,10 +33,10 @@ def python_multiprocessing_structures_are_buggy():
 
 
 def cap_max_workers_in_pool(max_workers, is_browser):
-  if is_browser:
+  if is_browser and os.getenv('EMTEST_CORES') is None:
```
sbc100: I wonder if we should maybe just no call `cap_max_workers_in_pool` in the case of `EMTEST_CORES` being set.

Also IIUC `EMCC_CORES` will also be used if `EMTEST_CORES` is not specified, so I think we would need to test both.

How about

```
# If a user explicitly asks for number of cores, don't artificially cap it.
if 'EMTEST_CORES' in os.environ or `EMCC_CORES` in os.environ:
  return max_workers
```

juj: > I wonder if we should maybe just no call `cap_max_workers_in_pool` in the case of `EMTEST_CORES` being set.

In case of the capping to max 61 cores, I think that is a good idea, because it shields the user from Python crap, which a developer might not be aware of.

I.e. say a user is using EMCC_CORES=64 and it is hanging on Windows due to the Python bug, then they have no real way of discovering that having specified > 61 cores is the cause, and our workaround will go lost on them. So it is better to force the 61 cores here, even if user sets a higher number.

juj: Testing both EMCC_CORES and EMTEST_CORES sounds fine, updated.

sbc100 (approved):

brendandahl:
I think we need to adjust circleci now. We specify EMCC_CORES=4 (though there are only 2 cores on a medium instance). Before this change we had two runners going, but now there are 4.

sbc100:
> I think we need to adjust circleci now. We specify EMCC_CORES=4 (though there are only 2 cores on a medium instance). Before this change we had two runners going, but now there are 4.

Sounds like the right fix yes.

juj:
Oh I see, the CircleCI runners explicitly used EMCC_CORES to oversubscribe over the hardware threads? Let me post a PR.

PR #25209, opened 2025-09-07
Fix race condition in assigning the pthread_t object in main thread

### LINKED ISSUE #25026: test_pthread_proxying is flaky
This failure comes through every now and then:

```
FAIL: test_pthread_proxying (test_core.core0.test_pthread_proxying)
  File "/usr/lib/python3.12/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/usr/lib/python3.12/unittest/case.py", line 634, in run
  File "/usr/lib/python3.12/unittest/case.py", line 589, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 928, in resulting_test
    return func(self, *args)
           ^^^^^^^^^^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 385, in decorated
    f(self, *args, **kwargs)
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 700, in metafunc
    f(self, *args, **kwargs)
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/test_core.py", line 2606, in test_pthread_proxying
    self.do_run_in_out_file_test('pthread/test_pthread_proxying.c', interleaved_output=False)
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 1980, in do_run_in_out_file_test
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 2029, in _build_and_run
    self.assertContained(o, js_output, regex=regex)
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 1687, in assertContained
  File "/usr/lib/python3.12/unittest/case.py", line 715, in fail
AssertionError: Expected to find 'Testing async proxying
running widget 1 on main
running widget 2 on looper
running widget 3 on returner
Testing sync proxying
running widget 4 on looper
running widget 5 on returner
Testing sync_with_ctx proxying
running widget 6 on looper
running widget 7 on returner
Testing callback proxying
running widget 8 on main
running widget 9 on looper
running widget 10 on returner
Testing callback_with_ctx proxying
running widget 11 on main
running widget 12 on looper
running widget 13 on returner
Testing promise proxying
running widget 14 on worker
running widget 15 on looper
running widget 16 on returner
Testing promise_with_ctx proxying
running widget 17 on worker
running widget 18 on looper
running widget 19 on returner
Testing tasks queue growth
Testing proxying queue growth
work
work
work
work
done
' in 'Testing async proxying
running widget 1 on main
running widget 2 on looper
running widget 3 on returner
Testing sync proxying
running widget 4 on looper
running widget 5 on returner
Testing sync_with_ctx proxying
running widget 6 on looper
running widget 7 on returner
Testing callback proxying
running widget 8 on main
running widget 9 on looper
running widget 10 on returner
Testing callback_with_ctx proxying
running widget 11 on main
running widget 12 on looper
running widget 13 on returner
Testing promise proxying
running widget 14 on worker
running widget 15 on looper
running widget 16 on returner
Testing promise_with_ctx proxying
running widget 17 on unknown
running widget 18 on looper
running widget 19 on returner
Testing tasks queue growth
Testing proxying queue growth
work
work
work
work
done
looper_main
returner_main
quitting looper
returner_main
returner_main
program exited (with status: 0), but keepRuntimeAlive() is set (counter=0) due to an async operation, so halting execution but not exiting the runtime or preventing further async execution (you can use emscripten_force_exit, if you want to force a true shutdown)
', diff:

--- expected
+++ actual
@@ -21,7 +21,7 @@
 running widget 15 on looper
 running widget 16 on returner
 Testing promise_with_ctx proxying
-running widget 17 on worker
+running widget 17 on unknown
 running widget 18 on looper
 running widget 19 on returner
 Testing tasks queue growth
@@ -31,4 +31,10 @@
 work
 work
 done
+looper_main
+returner_main
+quitting looper
+returner_main
+returner_main
+program exited (with status: 0), but keepRuntimeAlive() is set (counter=0) due to an async operation, so halting execution but not exiting the runtime or preventing further async execution (you can use emscripten_force_exit, if you want to force a true shutdown)
```
which suggests some kind of a race condition in proxying. (a lot of other pthread proxying tests are already marked flaky)

juj:
Fix race condition in assigning the pthread_t object in main thread, when creating a pthread. Fixes #25026.

Also reduce the stress test case for `test_stress_pthread_proxying` so that it runs faster and reproduces more often.

### DIFF
--- a/system/lib/pthread/pthread_create.c
+++ b/system/lib/pthread/pthread_create.c
@@ -171,6 +171,11 @@ int __pthread_create(pthread_t* restrict res,
   struct pthread *new = (struct pthread*)offset;
   offset += sizeof(struct pthread);
 
+  // Assign the pthread_t object over immediately, so that by the time pthread_create_js()
+  // is dispatched to a pthread and the pthread main runs, the value will be visible to
+  // the thread to examine.
+  __atomic_store_n(res, new, __ATOMIC_SEQ_CST);
+
   new->map_base = block;
   new->map_size = size;
 
@@ -250,6 +255,10 @@ int __pthread_create(pthread_t* restrict res,
 
   int rtn = __pthread_create_js(new, &attr, entry, arg);
   if (rtn != 0) {
+    // Reset the pthread_t return value to zero (we assigned to it above,
+    // so by clearing it here we won't litter bits to caller)
+    __atomic_store_n(res, 0, __ATOMIC_SEQ_CST);
+
     if (!--libc.threads_minus_1) libc.need_locks = 0;
 
     // undo previous addition to the thread list
@@ -269,7 +278,6 @@ int __pthread_create(pthread_t* restrict res,
       self->prev,
       new);
 
-  *res = new;
   return 0;
 }
 
--- a/system/lib/pthread/pthread_create.c
+++ b/system/lib/pthread/pthread_create.c
@@ -171,11 +171,6 @@ int __pthread_create(pthread_t* restrict res,
   struct pthread *new = (struct pthread*)offset;
   offset += sizeof(struct pthread);
 
-  // Assign the pthread_t object over immediately, so that by the time pthread_create_js()
-  // is dispatched to a pthread and the pthread main runs, the value will be visible to
-  // the thread to examine.
-  __atomic_store_n(res, new, __ATOMIC_SEQ_CST);
-
   new->map_base = block;
   new->map_size = size;
 
@@ -252,6 +247,12 @@ int __pthread_create(pthread_t* restrict res,
   // by the time it returns the thread could be running and we
   // want libc.need_locks to be set from the moment it starts.
   if (!libc.threads_minus_1++) libc.need_locks = 1;
+
+  // Assign the pthread_t object over immediately, so that by the time pthread_create_js()
+  // is dispatched to a pthread and the pthread main runs, the value will be visible to
+  // the thread to examine.
+  // Use __atomic_store_n() instead of a_store() to avoid splicing the pointer.
+  __atomic_store_n(res, new, __ATOMIC_SEQ_CST);
 
   int rtn = __pthread_create_js(new, &attr, entry, arg);
   if (rtn != 0) {
--- a/test/code_size/test_codesize_minimal_pthreads.json
+++ b/test/code_size/test_codesize_minimal_pthreads.json
@@ -1,10 +1,10 @@
 {
   "a.out.js": 7659,
   "a.out.js.gz": 3776,
-  "a.out.nodebug.wasm": 19592,
-  "a.out.nodebug.wasm.gz": 9060,
-  "total": 27251,
-  "total_gz": 12836,
+  "a.out.nodebug.wasm": 19599,
+  "a.out.nodebug.wasm.gz": 9063,
+  "total": 27258,
+  "total_gz": 12839,
   "sent": [
     "a (memory)",
     "b (emscripten_get_now)",
--- a/test/code_size/test_codesize_minimal_pthreads_memgrowth.json
+++ b/test/code_size/test_codesize_minimal_pthreads_memgrowth.json
@@ -1,10 +1,10 @@
 {
   "a.out.js": 8086,
   "a.out.js.gz": 3977,
-  "a.out.nodebug.wasm": 19593,
-  "a.out.nodebug.wasm.gz": 9061,
-  "total": 27679,
-  "total_gz": 13038,
+  "a.out.nodebug.wasm": 19600,
+  "a.out.nodebug.wasm.gz": 9064,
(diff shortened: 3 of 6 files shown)


system/lib/pthread/pthread_create.c
```diff
@@ -171,6 +171,11 @@ int __pthread_create(pthread_t* restrict res,
   struct pthread *new = (struct pthread*)offset;
   offset += sizeof(struct pthread);
 
+  // Assign the pthread_t object over immediately, so that by the time pthread_create_js()
+  // is dispatched to a pthread and the pthread main runs, the value will be visible to
+  // the thread to examine.
+  __atomic_store_n(res, new, __ATOMIC_SEQ_CST);
```
sbc100: In the musl codebase you can/should use the `a_store` macro instead I think.

juj: This is not musl codebase file? This is our own, no?

sbc100: This file uses the musl atomics convetions, and its at-least-in-part based on the musl version of the same code.  See all the `a_store`, `a_cas` and `__wait` calls that exist already.

juj: `a_store` only assigns an integer, so it would splice the pointer in a 64-bit build.

Would you like

```c
#ifdef __wasm64__
  a_store_l((long*)res, (long)new);
#else
  a_store((int*)res, (int)new);
#endif
```

instead?

sbc100: This makes me wonder how musl deals with this issue.

The core here is heavily based on `./system/lib/libc/musl/src/thread/pthread_create.c` including the final `*res = new;` assignment.

sbc100: LGTM then with an extra comment about why we are not using `a_store` here.

I'm still curious why the pointer needs to be assigned though, isn't it enough for the `struct pthread` to be populated correctly?   Where does the target thread depend on `*res*` being set?

system/lib/pthread/pthread_create.c
```diff
@@ -171,6 +171,11 @@ int __pthread_create(pthread_t* restrict res,
   struct pthread *new = (struct pthread*)offset;
   offset += sizeof(struct pthread);
 
+  // Assign the pthread_t object over immediately, so that by the time pthread_create_js()
+  // is dispatched to a pthread and the pthread main runs, the value will be visible to
+  // the thread to examine.
```
sbc100: Can we move this to line 250?

juj: ok

system/lib/pthread/pthread_create.c
```diff
@@ -248,8 +248,18 @@ int __pthread_create(pthread_t* restrict res,
   // want libc.need_locks to be set from the moment it starts.
   if (!libc.threads_minus_1++) libc.need_locks = 1;
 
+  // Assign the pthread_t object over immediately, so that by the time pthread_create_js()
+  // is dispatched to a pthread and the pthread main runs, the value will be visible to
+  // the thread to examine.
```
sbc100, line 253: lgtm, but I wonder if you understand why exactly writing `res` fixes the issue?

__pthread_create_js rakes the `new` pointer where was `res` is basically a pointer to a pointer which is owned by the calling thread.   Do you know why is it important the `res` be set during `__pthread_create_js`?

system/lib/pthread/pthread_create.c
sbc100: (Also, would the standard C11 `atomic_store` work here? https://en.cppreference.com/w/c/atomic/atomic_store)

system/lib/pthread/pthread_create.c
juj: See the minimal test code below. It reads from res in https://github.com/emscripten-core/emscripten/pull/25209/files#diff-a1d042ebd40397e77e1318873e436c39125856dafa8417cd5b27443702b731b0R13

sbc100: Thanks for the clarification.  Sorry I didn't see that.

PR #25213, opened 2025-09-08
Add test for sequential consistency in printing from a pthread.

juj:
https://github.com/emscripten-core/emscripten/issues/19683#issuecomment-3263265733

### DIFF
--- a/test/pthread/test_pthread_print_seq_consistent.c
+++ b/test/pthread/test_pthread_print_seq_consistent.c
@@ -15,5 +15,5 @@ int main() {
   pthread_t t;
   pthread_create(&t, 0, pthread_main, 0);
   pthread_join(t, 0);
-  emscripten_out("thread has quit.");
+  emscripten_out("thread joined");
 }
--- a/test/pthread/test_pthread_print_seq_consistent.out
+++ b/test/pthread/test_pthread_print_seq_consistent.out
@@ -1,2 +1,2 @@
 thread has started.
-thread has quit.
+thread joined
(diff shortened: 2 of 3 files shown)


test/pthread/test_pthread_print_seq_consistent.js
```diff
@@ -0,0 +1,3 @@
+if (!ENVIRONMENT_IS_PTHREAD) {
+  Module.print = (text) => { console.log(text); }
+}
```
sbc100, line 3: Why is this file needed?  Shouldn't we be testing the default behaviour of `emscripten_out`

sbc100: Oh I see your are testing both.

What happens if you override `Module.print` on the pthreads too?     i.e. is the `if` condition import here?  I would hope not, since calls `out` should be proxied back to the main threads's `print` handler regardless.

test/pthread/test_pthread_print_seq_consistent.c
```diff
@@ -0,0 +1,19 @@
...
+void *pthread_main(void *arg) {
+  emscripten_out("thread has started.");
+  return 0;
+}
+
+int main() {
+  pthread_t t;
+  pthread_create(&t, 0, pthread_main, 0);
+  pthread_join(t, 0);
+  emscripten_out("thread has quit.");
```
sbc100: "thread joined"

test/pthread/test_pthread_print_seq_consistent.js
juj: > is the `if` condition import here?

Yes, it is important.

> I would hope not, since calls `out` should be proxied back to the main threads's `print` handler regardless.

No, that is not correct. If pthread overrides a print handler, then it is not proxied to the main thread. https://github.com/emscripten-core/emscripten/blob/80ca182d92aa33f874acd2f25718143ce8848f37/src/runtime_pthread.js#L88

juj:
I'll close this since we were not able to remove the broken behavior.

PR #25216, opened 2025-09-08
Move browser source map test to interactive tests.

brendandahl:
This test tries to manually launch a browser for verification. This causes issues when trying to launch a browser from headless mode over ssh. The test also doesn't launch the correct file URL in windowed mode, so I'm guessing this hasn't been checked for a very long time.

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -11,7 +11,6 @@ import shutil
 import subprocess
 import time
 import unittest
-import webbrowser
 import zlib
 from functools import wraps
 from http.server import BaseHTTPRequestHandler, HTTPServer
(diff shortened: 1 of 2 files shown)


brendandahl:
@sbc100  I think this is the test that was clobbering your terminal output.

test/test_browser.py
```diff
@@ -308,15 +304,11 @@ def test_zzz_html_source_map(self):
...
-If manually bisecting:
-  Check that you see src.cpp among the page sources.
-  Even better, add a breakpoint, e.g. on the printf, then reload, then step
-  through and see the print (best to run with --save-dir for the reload).
-''')
+    # If manually bisecting:
+    #   Check that you see src.cpp among the page sources.
+    #   Even better, add a breakpoint, e.g. on the printf, then reload, then step
+    #   through and see the print (best to run with --save-dir for the reload).
+    # webbrowser.open_new('file://src.html')
```
sbc100: I guess we should change these instructions to say something like `run "./emrun out/test/src.html"` and just delete this line?

brendandahl: Yeah, or maybe we should just delete the test all together? In its current state it's just checking that the map file exists. We have a number of other tests that do that in `test_other`?

sbc100: Or move it to interactive tests?

Issue #25217, opened 2025-09-08
EXIT_RUNTIME=0 semantics


juj:
Currently `src/settings.js` states:

```
// If 0, the runtime is not quit when main() completes (allowing code to
// run afterwards, for example from the browser main event loop). atexit()s
// are also not executed, and we can avoid including code for runtime shutdown,
// like flushing the stdio streams.
// Set this to 1 if you do want atexit()s or stdio streams to be flushed
// on exit.
var EXIT_RUNTIME = false;
```

If user passes `-sEXIT_RUNTIME=0`, does that mean

a) the generated code will never shut down, and no disk size will be expended to implement/track shutdown, or
b) falling out of `main()` will not shut down the page, but yield with a live runtime. The user can shut down the runtime either via explicit call to exit/emscripten_force_exit/dropping runtime keepalive refcount to zero.

In `-sMINIMAL_RUNTIME`, `-sEXIT_RUNTIME=0` crystal clear means `a)`. In regular runtime, it originally started out as meaning `b)`, though it seems like the multitude of `#if EXIT_RUNTIME`s in the src/lib*.js code seem to have kind of taken it to a direction of `a)` as well.

I wonder if things would be simpler if there were separate flags for these two usages?

Maybe something like `-sEXIT_RUNTIME=disabled` to mean that the concept of runtime shutting down doesn't need to be emitted to the web page (and falling off of `main()` should always yield), and `-sEXIT_RUNTIME=yield` to mean that falling out of main() should yield to the event loop, rather than shut down - and the ability to shut down later would be generated to the page?

One thing that makes things a bit more complex here is that in `-pthread` builds, the pthreads will need to have the ability to shut down at least enough of its runtime, so that the Worker hosting a pthread can acquire a different pthread to host later on.

sbc100:
My understanding has always been `(a)`.  i.e: If you build with `-sNO_EXIT_RUNTIME` you won't *ever* get your global destructors called because we remove all the of the atexit mechanisms and such (i.e. its mostly a code saving tool).

If there are places in the where where it seems that `EXIT_RUNTIME` is doing something different it could be where it overlaps with some other feature (e.g. threads) making it hard to fully to `(a)` in as much as might want, but `(a)` is still the intended meaning (I believe).  Can you point to the places in the code where you think we are doing it wrong?

This is one of the most bug-generating settings that we have an it has caused a *lot* headaches over the years.     I would be loath to make it even more complex by adding yet another setting here.

[...]

PR #25218, opened 2025-09-08
Update documentation on -sEXIT_RUNTIME.

juj:
Closes https://github.com/emscripten-core/emscripten/issues/25217

Update documentation to reflect the current state of affairs for the meaning of `-sEXIT_RUNTIME`.

### DIFF
--- a/site/source/docs/tools_reference/settings_reference.rst
+++ b/site/source/docs/tools_reference/settings_reference.rst
@@ -95,12 +95,15 @@ exception, but will not shut down the runtime.
 Set this to 1 if you do want to retain the ability to shut down the program.
 If 1, then completing main() will by default call exit(), unless a refcount
 keeps the runtime alive. Call emscripten_exit_with_live_runtime() to finish
-main() while keeping the runtime alive. Calling exit() will shut down the
-runtime, invoking atexit()s, and flushing stdio streams.
+main() while keeping the runtime alive. Calling emscripten_force_exit() will
+shut down the runtime, invoking atexit()s, and flushing stdio streams.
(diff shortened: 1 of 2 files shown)


site/source/docs/tools_reference/settings_reference.rst
```diff
@@ -85,16 +85,22 @@ Default value: true
...
+will yield to the event loop of the JS environment, allowing event handlers
+to run afterwards. If 0, C++ global destructors will not be emitted into the
+build either, to save on code size. Calling exit() will throw an unwinding
+exception, but will not shut down the runtime.
+
+Set this to 1 if you do want to retain the ability to shut down the program.
+If 1, then completing main() will by default call exit(), unless a refcount
+keeps the runtime alive. Call emscripten_exit_with_live_runtime() to finish
+main() while keeping the runtime alive. Calling exit() will shut down the
+runtime, invoking atexit()s, and flushing stdio streams.
```
sbc100: Actually, I think calling `exit()` won't do this while the refcount is non-zero.

If you want to override the refcount you would need to call `emscripten_force_exit()`.

Also, maybe mention `emscripten_runtime_keepalive_push()` alongside `emscripten_exit_with_live_runtime()` which is basically the same thing but without the unwind.

sbc100: We do have docs for `emscripten_force_exit` and `emscripten_exit_with_live_runtime` so maybe just say something like `See docs for ...` rather than duplicating?

juj: ok, updated wording.

sbc100 (approved):

PR #25222, opened 2025-09-08
Fix shlex splitting when launching parallel browser test runner on Windows.

juj:
Fix shlex splitting when launching parallel browser test runner on Windows.

Before this PR, running

```
set EMTEST_BROWSER=C:\Program Files\Mozilla Firefox\firefox.exe
test\runner browser
```
would fail with the parallel test harness attempting to launch browser command line

`["C:\Program Files\Mozilla Firefox\firefox.exe -profile C:\emsdk\emscripten\main\out\profile-1"]`

when it should instead have attempted to launch command line

`["C:\Program Files\Mozilla Firefox\firefox.exe", "-profile", "C:\emsdk\emscripten\main\out\profile-1"]`

Fix by having the browser config first do the shlex stuff to interpret `EMTEST_BROWSER`, and only then append the list of strings of command line args, rather than first append the command line args as a string spaghetti, and using shlex split to parse the resulting command line string.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -2517,29 +2517,25 @@ class BrowserCore(RunnerCore):
         EMTEST_BROWSER = '"' + EMTEST_BROWSER.replace("\\", "/") + '"'
     browser_args = shlex.split(EMTEST_BROWSER)
 
-    try:
-      if EMTEST_BROWSER_AUTO_CONFIG:
-        logger.info('Using default CI configuration.')
-        cls.browser_data_dir = DEFAULT_BROWSER_DATA_DIR
-        if worker_id is not None:
-          # Running in parallel mode, give each browser its own profile dir.
-          cls.browser_data_dir += '-' + str(worker_id)
-        if os.path.exists(cls.browser_data_dir):
-          utils.delete_dir(cls.browser_data_dir)
-        os.mkdir(cls.browser_data_dir)
-        if is_chrome():
-          config = ChromeConfig()
-        elif is_firefox():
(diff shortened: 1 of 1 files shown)


test/common.py
```diff
@@ -105,17 +104,24 @@ class ChromeConfig:
...
 
 class FirefoxConfig:
-  data_dir_flag = '-profile '
   default_flags = ()
-  headless_flags = '-headless'
+  headless_flags = ['-headless']
+
+  @staticmethod
+  def data_dir_cmdline(data_dir):
+    return ['-profile', data_dir]
```
sbc100, line 127: Does chrome not support `['--user-data-dir', xxx]` or firefox not support `-profile=foo`?   i.e. can we use the same method for both?

juj: Firefox at least does not support `-profile=foo`, but silently ignores it.

I don't have Chrome set up to test.

sbc100: lgtm then, but lets lets @brendandahl approve since he wrote this code.

sbc100:
This change seems to have broken a bunch of the `test_cubegeom` tests under chrome

sbc100:
I can't see why though.. the command line before and after looks identical

sbc100:
Oh, its because `--use-gl=swiftshader` is not longer added to EMTEST_BROWSER, only to browser_cmd.

sbc100:
I think maybe we should revert this to make the CI green again.  It should be easy enough to re-land.  Unless somehow has a fix ready?

PR #25223, opened 2025-09-08
[testing] Consistently log test name when retrying.

brendandahl:
Now in parallel mode it's unclear which test fails.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -248,7 +248,7 @@ def flaky(note=''):
           return f(*args, **kwargs)
         except AssertionError as exc:
           preserved_exc = exc
-          logging.info(f'Retrying flaky test "{f.__name__}" (attempt {i}/{EMTEST_RETRY_FLAKY} failed): {exc}')
+          logging.info(f'Retrying flaky test "{f.id()}" (attempt {i}/{EMTEST_RETRY_FLAKY} failed): {exc}')
           # Mark down that this was a flaky test.
           open(flaky_tests_log_filename, 'a').write(f'{f.__name__}\n')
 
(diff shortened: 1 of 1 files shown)


test/common.py
```diff
@@ -2631,7 +2631,7 @@ def run_browser(self, html_file, expected=None, message=None, timeout=None, extr
             self.assertContained(expected, output)
           except self.failureException as e:
             if extra_tries > 0:
-              print('[test error (see below), automatically retrying]')
+              print(f'[test error in: {self._testMethodName} (see below), automatically retrying]')
```
sbc100: Didn't we already do this? 

Or maybe we already did it but in the `@flaky` decorator above?   Looks like we use `f.__name__` there.. we should at least be consistent, no?

We should also be consistent about the exceptions we catch I think.  CC @juj who just updated the exception conditions in `@flaky`.

juj: Does `self._testMethodName` print the test suite name as well? If so, I think changing `f.__name__` to use `self._testMethodName` instead would be better, since that would show the suite as well.

sbc100: It looks like `self.id()` is the public way to get the full name, include the suite.   Maybe we should use that everywhere?

brendandahl: _testMethodName just prints the test name, not suite. `self.id()` does `test_browser.browser.test_wasmfs_opfs_errors`. I don't really have a preference

sbc100: I don't really mind either, since we only run one test suite a time is seems reasonble to print the shorter name, but `self.id()` is both less ambiguous and a public method (rather than a private field).

brendandahl: > Or maybe we already did it but in the `@flaky` decorator above? Looks like we use `f.__name__` there.. we should at least be consistent, no?

That's only when marked as flaky. We retry by default 1 time on failures. I think we should change that, but out of scope for this PR.

Issue #25203, opened 2025-09-06
Fix for fullscreen button in default html/shell


Revilo0509:
Please include the following in your bug report:

**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.14 (96371ed7888fc78c040179f4d4faa82a6a07a116)
clang version 22.0.0git (https:/github.com/llvm/llvm-project 1cc84bcc08f723a6ba9d845c3fed1777547f45f9)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /root/emsdk/upstream/bin

**JS Console Error**
Aborted('requestFullscreen' was not exported. add it to EXPORTED_RUNTIME_METHODS (see the Emscripten FAQ))
overrideMethod @ hook.js:608
emscript.js:704 Uncaught RuntimeError: Aborted('requestFullscreen' was not exported. add it to EXPORTED_RUNTIME_METHODS (see the Emscripten FAQ))
    at abort (emscript.js:699:11)
    at Object.get (emscript.js:556:9)
    at HTMLInputElement.onclick (emscript.html:115:132)
hook.js:608 user callback triggered after runtime exited or application aborted.  Ignoring.

**Fix**
I have found out that changing:
```html
<input type="button" value="Fullscreen" onclick="Module.requestFullscreen(document.getElementById('pointerLock').checked, 
                                                                                document.getElementById('resize').checked)">
```
To
```html
<input type="button" value="Fullscreen" onclick="Module.canvas.requestFullscreen(document.getElementById('pointerLock').checked, 
                                                                                document.getElementById('resize').checked)">
```

fixes atleast my issue.
OBS: I am not to experienced, and it may just be a user error. Anyway feels like something that should be reported. :P

PR #25224, opened 2025-09-09
Export `requestFullscreen` by default when targeting default shell html

sbc100:
This was broken by #24269 but it seems we don't have automated tests for it.

Fixes: #25203


sbc100:
I'm not really sure how to go about creating a test for this... we would need to simulate the button click I suppose?

kripken:
Actually a manual event is certainly not enough, for security reasons (web pages cannot jump into fullscreen and hide the browser chrome).

sbc100:
> Actually a manual event is certainly not enough, for security reasons (web pages cannot jump into fullscreen and hide the browser chrome).

There are sometimes flags to enable that kind of thing for testing/development though

Issue #25227, opened 2025-09-09
Typo in zlib port?


ColmTalbot:
I'm attempting to build a project using emconfigure and I think there's a typo in how zlib is being invoked [here](https://github.com/emscripten-core/emscripten/blob/main/tools/ports/zlib.py#L26). If I change this line, my project correctly finds with zlib.

**Version of emscripten/emsdk:**

emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.14 (96371ed7888fc78c040179f4d4faa82a6a07a116)
clang version 22.0.0git (https:/github.com/llvm/llvm-project 1cc84bcc08f723a6ba9d845c3fed1777547f45f9)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /emsdk/upstream/bin

**Failing command line in full:**

I don't have an example using emconfigure, but if I put together minimal example using `-sUSE_ZIB` rather than `-sUSE_ZLIB` I can pretty easily reproduce the error.

```console
# emcc test_zlib.c -o test_zlib.html -sUSE_ZIB=1
emcc: error: Attempt to set a non-existent setting: 'USE_ZIB'
 - did you mean one of USE_ZLIB, USE_BZIP2, USE_ICU?
 - perhaps a typo in emcc's  -sX=Y  notation?
 - (see src/settings.js for valid values)
```

*test_zlib.c*

```c
#include <stdio.h>
#include <zlib.h>

int main() {
    printf("zlib version: %s\n", zlibVersion());
    return 0;
}
```

kripken:
Fixed by #25228

PR #25228, opened 2025-09-09
TYPO: Fix typo in pkg config option for zlib

ColmTalbot:
Addresses #25227 

I'm not sure the best way to add testing for this, the best I can think of is duplicating the [existing tests for `bullet`](https://github.com/sbc100/emscripten/blob/ede0b27e12db07978f5cd3e028874030a9e5847c/test/test_other.py#L1094-L1101). I'm happy to add these in if they are useful.


sbc100:
Thanks!

PR #25229, opened 2025-09-09
Revert "Fix shlex splitting when launching parallel browser test runner on Windows."

sbc100:
Reverts emscripten-core/emscripten#25222

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -91,27 +91,21 @@ flaky_tests_log_filename = os.path.join(path_from_root('out/flaky_tests.txt'))
 
 # Default flags used to run browsers in CI testing:
 class ChromeConfig:
+  data_dir_flag = '--user-data-dir='
   default_flags = (
     # --no-sandbox because we are running as root and chrome requires
     # this flag for now: https://crbug.com/638180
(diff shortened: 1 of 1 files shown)


sbc100:
The `--use-gl=swiftshader` check (and others)  assume that `EMTEST_BROWSER` is a string so simply re-assigning the global to be a list is not necessarily the simple fix here.

juj (approved):
LGTM to revert to unblock. I'd be happy if you have an idea to patch as well.

juj:
Btw, is `CHROMIUM_BASED_BROWSERS = ['chrom', 'edge', 'opera']` a typo in common.py?

sbc100:
> Btw, is `CHROMIUM_BASED_BROWSERS = ['chrom', 'edge', 'opera']` a typo in common.py?

No, I think its trying to match `chome` and `chromium` both.

juj:
Oh I see.. `['chrome', 'chromium', 'edge', 'opera']` would certainly be more explicit, or having a comment.

sbc100:
> Oh I see.. `['chrome', 'chromium', 'edge', 'opera']` would certainly be more explicit, or having a comment.

Agreed, pretty sure I didn't write that code, I'm only guessing at why it is that way.

Also remember, there is also `google-chrome` and `google-chrome-unstable` and `google-chrome-beta` etc but I guess

PR #25233, opened 2025-09-09
Fix parallel test suite handling of EMTEST_BROWSER, attempt 2

juj:
Fix parallel test suite handling of EMTEST_BROWSER, attempt 2. Keep the browser cmdline as a string salad, and don't convert to a list.


juj:
Btw I think there is an idempotency bug in `def browser_open()`. The function appends to `EMTEST_BROWSER` every time `browser_open()` is called, but it could be called many times, e.g. from `browser_restart()` to restart a flaky/unresponsive browser test. So this function will then keep accumulating the parallel test runner args to `browser_open()`. 

Not attempting to fix that flaw in this PR (if it is a flaw).

PR #25235, opened 2025-09-09
[testing] Do not automatically retry browser tests on failure.

brendandahl:
Running the tests more than once by default is hiding flaky tests. Instead, we should run the tests once and if they become flaky either fix them or explicitly add `extra_tries` to the individual test.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -2694,7 +2694,7 @@ class BrowserCore(RunnerCore):
             post_build=None,
             cflags=None,
             timeout=None,
-            extra_tries=1,
+            extra_tries=0,
             reporting=Reporting.FULL,
             output_basename='test'):
     assert expected, 'a btest must have an expected output'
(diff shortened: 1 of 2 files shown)


sbc100 (approved):
lgtm .. but how much more flake are we likely to see now :-/

test/common.py
```diff
@@ -2588,7 +2588,7 @@ def assert_out_queue_empty(self, who):
   #                     synchronously, so we have a timeout, which can be hit if the VM
   #                     we run on stalls temporarily), so we let each test try more than
   #                     once by default
-  def run_browser(self, html_file, expected=None, message=None, timeout=None, extra_tries=1):
+  def run_browser(self, html_file, expected=None, message=None, timeout=None, extra_tries=0):
```
sbc100, line 2591: the `def btest` function also has `extra_tries=1` so will need updating.

juj:
> lgtm .. but how much more flake are we likely to see now :-/

A lot from what I am observing from my browser runs. Though looks good to do this, manual flake handling will definitely converge to better outcome.

juj:
Would be nice to remove the `extra_tries` mechanism altogether in favor of the same `@flaky` mechanism, and EMTEST_RETRY_FLAKY?

Although that probably would need to be improved to interact somehow with the browser teardown and bringup..

PR #25240, opened 2025-09-10
Only cast wasm globals to unsigned under wasm32

sbc100:
With wasm64 this is not needed and risks loosing information.


juj:
I'm not sure this will work with that `if` in place. Try building in Wasm64 mode and bump e.g. heap base to 3gb.. then I'd bet that either `__heap_base` will be a BigInt or negative. (likely negative.. I don't see how it would become unsigned in Wasm64 mode otherwise?)

Would be good to have a test case to verify that scenario, which this fixes.

sbc100:
> I'm not sure this will work with that `if` in place. Try building in Wasm64 mode and bump e.g. heap base to 3gb.. then I'd bet that either `__heap_base` will be a BigInt or negative. (likely negative.. I don't see how it would become unsigned in Wasm64 mode otherwise?)
> 
> Would be good to have a test case to verify that scenario, which this fixes.

Seem to work:

```
$ ./emcc -sMEMORY64 ~/test//hello.c -sGLOBAL_BASE=4gb -sEXPORTED_FUNCTIONS=___heap_base,_main
$ grep ___heap_base a.out.js 
var ___heap_base = Module['___heap_base'] = 4295037264;var wasmImports = {
```

Without this change is broken:

```
$ grep ___heap_base a.out.js 
var ___heap_base = Module['___heap_base'] = 69968;var wasmImports = {
```

sbc100:
> I'm not sure this will work with that `if` in place. Try building in Wasm64 mode and bump e.g. heap base to 3gb.. then I'd bet that either `__heap_base` will be a BigInt or negative. (likely negative.. I don't see how it would become unsigned in Wasm64 mode otherwise?)

The code being modified here is python and it doesn't have any issues with decoding i64 globals, so in memory64 mode these globals are just normal python numbers.

sbc100:
Turns out we need to this all our global exports (since we assume they are all addresses).

Added some more testing.

test/other/test_export_global_address.c
```diff
@@ -9,7 +9,11 @@ int g_foo = 4;
 
 EM_JS(int*, get_foo_from_js, (void), {
   assert(_g_foo !== undefined, "g_foo not exported to JS");
+#if __wasm64__
+  return BigInt(_g_foo);
+#else
   return _g_foo;
+#endif
```
juj, line 16: This is what I meant. This is punting the responsibility to the user to be aware that they can get either BigInts or Numbers.

Is that the correct model to propose? When I proposed it before, I got ridiculed off that Wasm64 will never have 64 bits, but it will be Wasm53, and so users should always see bare Numbers.

juj (approved):
LGTM

test/other/test_export_global_address.c
sbc100: Its a current limitation of EM_JS that it can't do that automatic conversion to int53, like we do almost everywhere else (e.g. all the of the JS libraries).

This limitation is unrelated to the current change, and this `ifdef` here is (sadly) needed when running the test under wasm64 for reasons unrelated to this PR or exported globals.

The exported wasm globals are never bigint here, even on wasm64.   Trying to export a global with a value larger than int53 will not currently works for this reason.   Luckilym, we don't really support exporting anyting but a pointer, which should always be withing int53.

PR #25246, opened 2025-09-10
[testing] Support the @flaky decorator on browser tests.

brendandahl:
For browser tests use the `extra_tries` mechanism that already exists to re-run the test without recompiling.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -242,10 +242,15 @@ def flaky(note=''):
 
   def decorated(f):
     @wraps(f)
-    def modified(*args, **kwargs):
+    def modified(self, *args, **kwargs):
+      # Browser tests have there own method of retrying tests.
+      if self.is_browser_test():
+        self.flaky = True
+        return f(self, *args, **kwargs)
+
       for i in range(EMTEST_RETRY_FLAKY):
         try:
-          return f(*args, **kwargs)
+          return f(self, *args, **kwargs)
         except (AssertionError, subprocess.TimeoutExpired) as exc:
           preserved_exc = exc
           logging.info(f'Retrying flaky test "{f.id()}" (attempt {i}/{EMTEST_RETRY_FLAKY} failed): {exc}')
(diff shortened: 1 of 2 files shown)


juj (approved):
Very nice!

test/common.py
```diff
@@ -2586,12 +2592,17 @@ def assert_out_queue_empty(self, who):
...
+  def run_browser(self, html_file, expected=None, message=None, timeout=None, extra_tries=None):
     if not has_browser():
       return
     assert '?' not in html_file, 'URL params not supported'
+    if extra_tries is None:
+      # Check if the test was marked with the @flaky decorator.
+      if self.extra_tries is not None:
+        extra_tries = self.extra_tries
+      else:
+        extra_tries = 0
```
sbc100: Might be nice if maybe the `@flaky` decorator to take an optional `extra_tries` count, then we could remove this code.

Can be a followup.

brendandahl:
We'll still probably want to add these to the flaky log, but it's a start...

test/common.py
```diff
@@ -242,7 +242,12 @@ def flaky(note=''):
 
   def decorated(f):
     @wraps(f)
-    def modified(*args, **kwargs):
+    def modified(self, *args, **kwargs):
+      # Browser tests have there own method of retrying tests.
+      if self.is_browser_test():
+        self.extra_tries = EMTEST_RETRY_FLAKY
+        return f(self, *args, **kwargs)
```
sbc100, line 249: Since this is basically a seperate decorator perhaps we should just define it in `test_browser.py`.

We could give it a unique name, or keep the the same, and it could take an integer `extra_tries`?

test/common.py
```diff
@@ -242,7 +242,12 @@ def flaky(note=''):
...
+
       for i in range(EMTEST_RETRY_FLAKY):
         try:
           return f(*args, **kwargs)
```
sbc100: I guess this needs a `self,` too now?

test/common.py
juj: I wonder if a follow-up cleaning could remove the `extra_tries` function argument from `run_browser()` altogether? And always read the `self.extra_tries` parameter?

test/common.py
```diff
@@ -2694,7 +2705,7 @@ def btest(self, filename, expected=None,
             post_build=None,
             cflags=None,
             timeout=None,
-            extra_tries=0,
+            extra_tries=None,
```
juj: Or even simplest, maybe this could be
```py
extra_tries = EMTEST_RETRY_FLAKY if self.flaky else 0,
```
and the decorator would assign
```
self.flaky = True
```

PR #25247, opened 2025-09-10
Fix passing native backslashes to Windows Firefox launch.

juj:
Fix passing native backslashes to Windows Firefox launch. Otherwise Firefox would treat -profile C:/foo/ to be a URL http:///foo/ to visit.


sbc100:
> Firefox would treat -profile C:/foo/ to be a URL http:///foo/ to visit.

Wow, that seems odd.  Maybe a bug?

sbc100 (approved):

test/common.py
```diff
@@ -2525,7 +2525,7 @@ def browser_open(cls, url):
         exit_with_error("EMTEST_BROWSER_AUTO_CONFIG only currently works with firefox or chrome.")
       if WINDOWS:
         # Remove directory delimiter backslashes to avoid shlex.split getting confused.
```
sbc100: I guess this comment needs updating.

TBH, I'm not sure `shlex.split` is supposed to be used on windows at all.  But if work then lgtm.

PR #25248, opened 2025-09-10
Fix EMTEST_BROWSER command line to not accumulate parameters every time a browser is opened.

juj:
Fix EMTEST_BROWSER command line to not accumulate parameters every time a browser is opened.

This prevents the `-profile ...` directive and other command line args from stacking up on every browser terminate + restart.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -2491,7 +2491,7 @@ def configure_test_browser():
     elif is_firefox():
       config = FirefoxConfig()
     else:
-      exit_with_error(f"EMTEST_BROWSER_AUTO_CONFIG only currently works with firefox or chrome. Browser was \"{EMTEST_BROWSER}\"")
+      exit_with_error(f'EMTEST_BROWSER_AUTO_CONFIG only currently works with firefox or chrome. Browser was "{EMTEST_BROWSER}"')
     EMTEST_BROWSER += ' ' + ' '.join(config.default_flags)
     if EMTEST_HEADLESS == 1:
       EMTEST_BROWSER += f" {config.headless_flags}"
--- a/test/common.py
+++ b/test/common.py
@@ -2491,7 +2491,7 @@ def configure_test_browser():
     elif is_firefox():
       config = FirefoxConfig()
     else:
-      exit_with_error("EMTEST_BROWSER_AUTO_CONFIG only currently works with firefox or chrome.")
+      exit_with_error(f"EMTEST_BROWSER_AUTO_CONFIG only currently works with firefox or chrome. Browser was \"{EMTEST_BROWSER}\"")
     EMTEST_BROWSER += ' ' + ' '.join(config.default_flags)
     if EMTEST_HEADLESS == 1:
       EMTEST_BROWSER += f" {config.headless_flags}"
--- a/test/common.py
+++ b/test/common.py
@@ -2526,7 +2526,6 @@ class BrowserCore(RunnerCore):
 
   @classmethod
   def browser_open(cls, url):
-    global worker_id
     browser_args = EMTEST_BROWSER
 
     if EMTEST_BROWSER_AUTO_CONFIG:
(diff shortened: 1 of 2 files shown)


test/common.py
```diff
@@ -2496,20 +2526,10 @@ def browser_restart(cls):
...
-      # On Windows env. vars canonically use backslashes as directory delimiters, e.g.
-      # set EMTEST_BROWSER=C:\Program Files\Mozilla Firefox\firefox.exe
-      # and spaces are not escaped. But make sure to also support args, e.g.
-      # set EMTEST_BROWSER="C:\Users\clb\AppData\Local\Google\Chrome SxS\Application\chrome.exe" --enable-unsafe-webgpu
-      EMTEST_BROWSER = '"' + EMTEST_BROWSER.replace("\\", "/") + '"'
-
-    if not EMTEST_BROWSER:
-      logger.info('No EMTEST_BROWSER set. Defaulting to `google-chrome`')
-      EMTEST_BROWSER = 'google-chrome'
+    global worker_id
```
sbc100: I think this line can be deleted.

test/runner.py
```diff
@@ -512,6 +512,8 @@ def configure():
   assert 'PARALLEL_SUITE_EMCC_CORES' not in os.environ, 'use EMTEST_CORES rather than PARALLEL_SUITE_EMCC_CORES'
   parallel_testsuite.NUM_CORES = os.environ.get('EMTEST_CORES') or os.environ.get('EMCC_CORES')
 
+  common.configure_test_browser()
```
sbc100, line 515: Could this be done in `BrowserCore.setUpClass` instead of modifying this file?

sbc100: I guess that it would happen for each subclass that is configured which is also wrong..

juj: I don't know. Moved to test.

juj: Oh I see, yeah, there's

```
C:\emsdk\emscripten\main\test\test_interactive.py:
   18: class interactive(BrowserCore):

C:\emsdk\emscripten\main\test\test_posixtest_browser.py:
   13: class posixtest_browser(BrowserCore):

C:\emsdk\emscripten\main\test\test_sockets.py:
  160: class sockets(BrowserCore):
```

so that would indeed amplify.

sbc100: I guess its unlikely the one would run those at the same time but possible.

Maybe some kind of "once" global state?

sbc100: Maybe `EMTEST_BROWSER` should not be modified in-place at all but we could have some kind of `full_browser_cmd` global that is initialized just once based on `EMTEST_BROWSER`?

sbc100:
When I run non-browser tests such as `other` I now see the `common:INFO: No EMTEST_BROWSER set. Defaulting to `google-chrome` message.. for some reason I see it twice too.

```
$ test/runner other.*code_size* other.*codesize* --rebase 
common:INFO: No EMTEST_BROWSER set. Defaulting to `google-chrome`
common:INFO: No EMTEST_BROWSER set. Defaulting to `google-chrome`
Running test_other: (66 tests)
Using 66 parallel test processes
...
```

sbc100:
Looks like the auto-roller is failing on this too:

https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket/8704074500004063569/+/u/Emscripten_testsuite__core0_/stdout

```
subprocess.check_call(`C:\b\s\w\ir\x\w\install\emscripten\test\runner.bat core0`, cwd=`C:\b\s\w\ir\cache\builder\emscripten-releases\emtest-out`)
  File "C:\b\s\w\ir\x\w\install\emscripten\test\runner.py", line 602, in <module>
    configure()
  File "C:\b\s\w\ir\x\w\install\emscripten\test\runner.py", line 515, in configure
    common.configure_test_browser()
  File "C:\b\s\w\ir\x\w\install\emscripten\test\common.py", line 2479, in configure_test_browser
    if WINDOWS and '"' not in EMTEST_BROWSER and "'" not in EMTEST_BROWSER:
TypeError: argument of type 'NoneType' is not iterable
```

juj:
I posted https://github.com/emscripten-core/emscripten/pull/25258 to address that this morning.

PR #25249, opened 2025-09-10
Remove unnecessary hasOwnProperty call. NFC

sbc100:
This was causing issues with another patch I'm working on and I realized we can simply remove it for code size win.

### DIFF
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -1313,11 +1313,11 @@ var LibraryDylink = {
 #if ASSERTIONS
     assert(lib, `Tried to dlsym() from an unopened handle: ${handle}`);
 #endif
-    if (!lib.exports.hasOwnProperty(symbol) || lib.exports[symbol].stub) {
+    newSymIndex = Object.keys(lib.exports).indexOf(symbol);
+    if (newSymIndex == -1 || lib.exports[symbol]?.stub) {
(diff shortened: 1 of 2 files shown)


src/lib/libdylink.js
```diff
@@ -1313,11 +1313,11 @@ var LibraryDylink = {
 #if ASSERTIONS
     assert(lib, `Tried to dlsym() from an unopened handle: ${handle}`);
 #endif
-    if (!lib.exports.hasOwnProperty(symbol) || lib.exports[symbol].stub) {
+    newSymIndex = Object.keys(lib.exports).indexOf(symbol);
+    if (newSymIndex == -1 || lib.exports[symbol]?.stub) {
```
kripken, line 1317:
```suggestion
    if (newSymIndex == -1 || lib.exports[symbol].stub) {
```
The code before does not seem to have checked that?

kripken (approved):

src/lib/libdylink.js
kripken: This comment was resolved but it looks like it might have been missed?

kripken: Or is there a reason we need the `?` now?

sbc100: Oops sorry.  Will fix

PR #25253, opened 2025-09-10
[testing] Speed up test_sdl_audio_beep_sleep.

brendandahl:
This test takes over 60 seconds to run. There's really no point for it using such long beeps and delays when running in an automated test. After the changes it finishes in ~4 seconds.

### DIFF
--- a/test/browser/test_sdl_audio_beep_sleep.cpp
+++ b/test/browser/test_sdl_audio_beep_sleep.cpp
@@ -22,7 +22,15 @@
 #undef main
 #endif
 
-const int tone_duration = 1000;
+// Set this to true to if manually testing in browser to give longer delays
+// and beeps.
+// #define INTERACTIVE true
+
+#ifdef INTERACTIVE
+const int tone_duration = 10000;
+#else
+const int tone_duration = 10;
+#endif
 
 struct BeepObject {
   double toneFrequency;
@@ -196,7 +204,11 @@ void nextTest(void *unused = 0) {
     return;
   }
 
+#ifdef INTERACTIVE
   emscripten_sleep(100);
+#else
+  emscripten_sleep(1);
+#endif
 
   printf("Playing back a beep for %d msecs at %d Hz tone with audio format %s, %d channels, and %d samples/sec.\n",
       tone_duration, (int)Hz, SdlAudioFormatToString(sdlAudioFormats[s]), channels[c], freqs[f]);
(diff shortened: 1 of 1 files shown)


test/browser/test_sdl_audio_beep_sleep.cpp
```diff
@@ -22,7 +22,15 @@
 #undef main
 #endif
 
-const int tone_duration = 1000;
+// Set this to true to if manually testing in browser to give longer delays
+// and beeps.
+// #define INTERACTIVE true
+
+#ifdef INTERACTIVE
+const int tone_duration = 10000;
```
sbc100: Should this be 1000?

test/browser/test_sdl_audio_beep_sleep.cpp
```diff
@@ -211,7 +223,11 @@ void update() {
     delete beep;
     beep = 0;
 #ifdef __EMSCRIPTEN__
+#ifdef INTERACTIVE
     emscripten_async_call(nextTest, 0, 1500);
+#else
+    emscripten_async_call(nextTest, 0, 1);
+#endif
 #else
     SDL_Delay(1500);
```
sbc100: Should this 1500 value also change with `INTERACTIVE`?     Maybe define this value above?

juj:
Iirc this was a test that was created initially only for the interactive suite, in order to verify crackling free audio playback.

But then at some point, the test was added also to browser suite for automatic testing, since interactive test is run so rarely.

So from that perspective, in the browser suite it would be fine for this test to only try out one of the audio playback combinations, e.g. 44.1 kHz signed 16. I don't think there is any benefit in testing the other combinations. That could make the test run in < 0.5s in browser suite.

brendandahl:
> Iirc this was a test that was created initially only for the interactive suite, in order to verify crackling free audio playback.
> 
> But then at some point, the test was added also to browser suite for automatic testing, since interactive test is run so rarely.
> 
> So from that perspective, in the browser suite it would be fine for this test to only try out one of the audio playback combinations, e.g. 44.1 kHz signed 16. I don't think there is any benefit in testing the other combinations. That could make the test run in < 0.5s in browser suite.

Makes sense, I'll maybe do in a follow up. For now, I'm just trying to prune down the list of tests that take over 30s.

juj:
> Makes sense, I'll maybe do in a follow up. For now, I'm just trying to prune down the list of tests that take over 30s.

That's great. I was looking at the EMTEST_VISUALIZE graphs of browser suite yesterday, and there are a few of those. Optimizing these will help a *lot* with parallelization and runtimes on my CI.

brendandahl:
I'd also like to separate out compile time and browser test time at some point. It seems a few of the tests that are running for a long time in the browser are hitting the unresponsive timeout, causing them to look flaky.

Issue #24324, opened 2025-05-14
Embind: `enums` are not serializable. Couldn't the implementation be simplified ?


FelixNumworks:
When I'm storing an enum value binded via embind in a js var, I can't use `JSON.stringify` on it.

## Current implementation:

I'm not understanding very well why the enums are binded the way they are today, attaching values to their constructor.

### Register enum

https://github.com/emscripten-core/emscripten/blob/1cb178b2b8d30a0c8860921322c633ca8ee35bd0/src/lib/libembind.js#L2277-L2295 

### Register enum value

https://github.com/emscripten-core/emscripten/blob/1cb178b2b8d30a0c8860921322c633ca8ee35bd0/src/lib/libembind.js#L2298-L2310

## Suggested implementation

Isn't it possible to replace with something along those lines:

### Register enum

```js
  _embind_register_enum: (rawType, name, size, isSigned) => {
    name = readLatin1String(name);
    var values = {};
    registerType(rawType, {
      name,
      values,
      fromWireType: (c) => {
        return Object.values(values).find((v) => v.value === c);
      },
      toWireType: (destructors, c) => c.value,
      argPackAdvance: GenericWireTypeSize,
      'readValueFromPointer': enumReadValueFromPointer(name, size, isSigned),
      destructorFunction: null,
    });
    exposePublicSymbol(name, values);
  },
```

### Register enum value

```js
  _embind_register_enum_value: (rawEnumType, name, enumValue) => {
    var enumType = requireRegisteredType(rawEnumType, 'enum');
    name = readLatin1String(name);
    enumType.values[name] = { value: enumValue };
  },
```

I'm just creating a `values` object tied to the enum type. And enum values are just a simple object `{ value: ... }`

It seems to me that it doesn't change how enums are used, but it makes them way more flexible, as they are just simple js object.

### Typescript 

You could even add the enum name to the the enum values to strongly type them in typescript

```js
  _embind_register_enum_value: (rawEnumType, name, enumValue) => {
    var enumType = requireRegisteredType(rawEnumType, 'enum');
    name = readLatin1String(name);
    enumType.values[name] = { value: enumValue, enumType: enumType.name };
  },
```

In the `.d.ts` file:
```
export interface MyEnumValue<T extends number> {
  value: T;
  enumType: "MyEnum";
}

export type MyEnum = MyEnumValue<0> | MyEnumValue<1> | MyEnumValue<2>;

interface EmbindModule {
[...]

Issue #19387, opened 2023-05-17
embind enums aren't compatible with jasmine equality checks


mitchellwills:
**Version of emscripten/emsdk:** `3.1.35`

All enum values of a given enum type generated by embind are considered equivalent by jasmine. Jasmine does object equality by doing a deep equals on most objects; however, this relies on being able to iterate over the object keys (using `Object.keys`). embind enum values are all objects with a single *non-enumerable* property with the underlying value. Since the `value` property isn't enumerable the object looks the same. For single values this is somewhat ok because the `toBe` matcher can be used instead; however, it becomes problematic when comparing the quality of objects/arrays using `toEquals` because the same logic is applied recursively.

For example
```javascript
module.MyEnum.VALUE1 === module.MyEnum.VALUE2 // false
module.MyEnum.VALUE1 == module.MyEnum.VALUE2 // false
expect(module.MyEnum.VALUE1).toBe(module.MyEnum.VALUE2);  // assertion failes
expect(module.MyEnum.VALUE1).toEqual(module.MyEnum.VALUE2);  // assertion incorrectly passes
expect([module.MyEnum.VALUE1]).toEqual([module.MyEnum.VALUE2]);  // assertion incorrectly passes
```

The [setup code](https://github.com/emscripten-core/emscripten/blob/daecbfcac67d5299b4583e287f14e27fb59dc85e/src/embind/embind.js#L2565-L2603) simplifies to roughly something like the following:

```javascript
  it('enum equality', () => {
    const MyEnum = {};

    var Value1 = Object.create(MyEnum.constructor.prototype, {
      value: {value: 1},  // not enumerable
    });
    var Value2 = Object.create(MyEnum.constructor.prototype, {
      value: {value: 2},  // not enumerable
    });

    expect(Value1 !== Value2).toBe(true);
    expect(Value1 != Value2).toBe(true);
    expect(Value1).toEqual(Value2);  // always passes
    expect([Value1]).toEqual([Value2]);  // always passes
  });
```

One option to fix this could be to simply make the `value` property enumerable, but not sure if that would be a breaking change.

sbc100:
@brendandahl

brendandahl:
Do you know offhand if this is how TypeScript generates enums?

mitchellwills:
[...]

Issue #18585, opened 2023-01-24
[question] TypeScript binding of enums?


harsszegi:
Hi,

I'm trying to wrap my head around how the enums are exposed with embind. Is there a documentation which provides a basic ".d.ts"-kinda description how is the ""class" structured, what interfaces it has? I'm having hard time to map them properly to ".d.ts", e.g. the mapped values don't behave like "normal" TypeScript enums (e.g. values mapped to numbers of strings).
Any idea?
Thanks,

zeyonaut:
A binding of the form:
```c++
emscripten::enum_<EnumClass>("EnumClass")
	.value("A", EnumClass::A)
	.value("B", EnumClass::B)
```
can be mapped to the following value-type pair:
```ts
EnumClass: {
  A: number,
  B: number,
};
```
One way you could ensure type safety in your resulting interface is to use some newtype in place of `number`. Technically, if you use a top-level `await` of your WebAssembly module, you could also manually define an `enum` whose values correspond to the values of the properties of the `EnumClass` object.

harichards:
You can do something like this:

**`my_enum.cc`**
```cpp
enum OldStyle {
    OLD_STYLE_ONE,
    OLD_STYLE_TWO
};

enum class NewStyle {
    ONE,
    TWO
};

EMSCRIPTEN_BINDINGS(my_enum_example) {
    enum_<OldStyle>("OldStyle")
        .value("ONE", OLD_STYLE_ONE)
        .value("TWO", OLD_STYLE_TWO)
        ;

    enum_<NewStyle>("NewStyle")
        .value("ONE", NewStyle::ONE)
        .value("TWO", NewStyle::TWO)
        ;
}
```

**`my_enum.d.ts`**
```ts
export declare enum OldStyle {
    ONE = 0,
    TWO = 1,
}

export declare enum NewStyle {
    ONE = 0,
    TWO = 1,
}

export interface MyModule extends EmscriptenModule {
    OldStyle: typeof OldStyle
    NewStyle: typeof NewStyle
}
```

I learned this by examining the output of [`tsembind`](https://github.com/ted537/tsembind). `EmscriptenModule` comes from [`@types/emscripten`](https://www.npmjs.com/package/@types/emscripten). For free functions, classes, class methods and static class methods, see [this comment](https://github.com/emscripten-core/emscripten/issues/10271#issuecomment-1179010012).

kirkwaiblinger:
Is that correct, though?

[...]

PR #25257, opened 2025-09-11
[embind] Add a way to register enum values as number or string

FelixNumworks:
This PR adds a param to `enum_` to be able to use enums values as plain string or plain numbers in javascript

#### Int enums:

```cpp
enum_<Animal>("Animal", enum_value_type::number)
 .value("Dog", Animal::Dog)
 .value("Cat", Animal::Cat)
```

emits:

```ts
// module.d.ts
export type Animal = 1 | 2;

interface EmbindModule {
    Animal: { Dog: 1, Cat: 2 },
}
```

#### String enums:

```cpp
enum_<Animal>("Animal", enum_value_type::string)
 .value("Dog", Animal::Dog)
 .value("Cat", Animal::Cat)
```

emits:

```ts
// module.d.ts
export type Animal = "Dog" | "Cat";

interface EmbindModule {
    Animal: { Dog: "Dog", Cat: "Cat" },
}
```

#### Object enums (default):

```cpp
enum_<Animal>("Animal", enum_value_type::object) 
 .value("Dog", Animal::Dog)
 .value("Cat", Animal::Cat)

 // OR
 Fix https://github.com/emscripten-core/emscripten/issues/24324
Fix https://github.com/emscripten-core/emscripten/issues/19387
Fix https://github.com/emscripten-core/emscripten/issues/18585

enum_<Animal>("Animal") 
 .value("Dog", Animal::Dog)
 .value("Cat", Animal::Cat)
```

emits:

```ts
// module.d.ts
export type AnimalValue<T extends number> {
  value: T
}

export type Animal = AnimalValue<1> | AnimalValue<2>;

interface EmbindModule {
  Animal: { Dog: Animal<1>, Cat: Animal<2> };
}
```

This doesn't conflict with current implementation of enums, as the parameter default value is the one keeping the same behaviour

Fixes: #24324, #19387, #18585


src/lib/libembind.js
```diff
@@ -2199,38 +2199,68 @@ var LibraryEmbind = {
...
-      constructor: ctor,
-      fromWireType: function(c) {
-        return this.constructor.values[c];
-      },
-      toWireType: (destructors, c) => c.value,
-      readValueFromPointer: enumReadValueFromPointer(name, size, isSigned),
-      destructorFunction: null,
-    });
-    exposePublicSymbol(name, ctor);
+    if (asString) {
```
brendandahl: Since this doesn't have any shared code let's split it out into it's own function `_embind_register_enum_string` or something. I'm guessing users will only use one way other other, so then if it's a totally seperate function it will be easier to do dead code elimination.

src/lib/libembind.js
```diff
@@ -2199,38 +2199,68 @@ var LibraryEmbind = {
...
 
-    var Enum = enumType.constructor;
-
-    var Value = Object.create(enumType.constructor.prototype, {
-      value: {value: enumValue},
-      constructor: {value: createNamedFunction(`${enumType.name}_${name}`, function() {})},
-    });
-    Enum.values[enumValue] = Value;
-    Enum[name] = Value;
+    if (enumType.asString) {
```
brendandahl: Same here, split into its own function. I think you'll have to store on the enum class in c++ whether it's a string or not and call the correct register function from there.

brendandahl (changes requested):

FelixNumworks:
I splitted most of the code between the two, but I kept a common class in the TS types generation, as they are very close to one another

brendandahl:
Sorry for the delay here. I got some feedback from a few different projects that also want enum to behave differently and not have a `.value`. Unfortunately, all the other projects would like the enums in JS to be the integer value, not a string value. Is there reason you want to the string form instead of an int?

On another topic, one downside I see to not using TS enums is you can compare different enums and it will NOT be an error. e.g.

```ts
export type Dog = 'a'|'b';
export type Cat = 'a'|'b';

interface EmbindModule {
  Dog: {a: 'a', b: 'b'};
  Cat: {a: 'a', b: 'b'};
};

let module = {} as EmbindModule;

let myDog : Dog = module.Dog.a;

if (myDog == module.Cat.a) { // <--- this is not a compiler error
}
```
vs

```ts
declare enum Dog {
    a = 'a',
    b = 'b',
}

declare enum Cat {
    a = 'a',
    b = 'b',
}

let myDog : Dog = Dog.a;
if (myDog == Cat.a) { < --- compiler error
}
```

FelixNumworks:
## Why I'm not using TS enums

> On another topic, one downside I see to not using TS enums is you can compare different enums and it will NOT be an error. e.g.

I didn't find a clean way to bind the enum to a real TS enum.

If I understand well, you suggest doing:

```ts
// module.d.ts
declare enum Animal {
    Dog = 1;
    Cat = 2;
}
```

But since this enum is only declared in a `.d.ts` file, it won't be accessible at run time. It only exists as a TS type.

The only way to then make this enum declaration usable, is to add this at the **root** of `module.js`:

```js
// module.js, outside the module code

const Animal = {
  Dog: 1,
  Cat: 2,
};

//  Add reverse mapping like real TS enums
Animal[1] = 'Dog';
Animal[2] = 'Cat';

export Animal;
```

This means that: 
- The enum logic is fully reimplemented in plain js
- The enum values must be declared outside of the module object (how ?)

I didn't like the idea of *manually* reimplementing what TS usually does under the hood. And even if I wanted to, I didn't know where to do this implementation. Thus, I went for a plain TS type. 

I also think that TS types have the benefit over TS enums of being easier to use and to overload. I prefer using them over TS enums in general (but this is a personal preference. TS enums are a bit clunky imo)

I don't think the comparison problem you raised is that much of an issue. In the end, `Dog.a` and `Cat.a` are indeed equal ...

## Why I'm using strings

> Is there reason you want to the string form instead of an int

We saw above that I couldn't easily implement real TS enums, so I went for:

```ts
// module.d.ts
export type Animal = 'Dog' | 'Cat';

interface EmbindModule {
  Animal: { Dog: 'Dog', Cat: 'Cat' },
}
```

Would you suggest replacing it with the following implementation ? I think it's a bit strange to declare a union type of ints like this..

```ts
// module.d.ts
export type Animal = 1 | 2;

interface EmbindModule {
    Animal: { Dog: 1, Cat: 2 },
}
```

Also this is less close to real TS enums.
Indeed, in TS, you have two ways enums behave:

- Strings: the enum is "one way"

```ts
const enum Animal { Dog: "Dog", Cat, "Cat" };
console.log(Object.values(Animal)); // ["Dog", "Cat"]
```
 
 - Numerics: the enum also carries the reverse mapping of values:
 
 ```ts
const enum Animal { Dog: 1, Cat, 2 };
console.log(Object.values(Animal)); // ["Dog", "Cat", 1, 2]
```
 
 See https://www.typescriptlang.org/docs/handbook/enums.html for details
 
 Since my implementation only used strings, it was closer to a real TS string enum behaviour.
 
 Finally, using plain strings allows me to use the enum values without even needing the module:
 
 ```ts
 const a: Animal = "Dog"; // << doesn't need the module
 // vs
 const a: Animal = myModule.Animal.Dog: // << needs an instanciated module
```

With number I would need to do:

 ```ts
 const a: Animal = 1; // << What is 1 ? Dog, Cat ? 
 // vs
 const a: Animal = myModule.Animal.Dog; // << needs an instanciated module to be readable
```
 ## Possible solution
 
If people need the int values, I think we can add a parameter in the enum binding (and merge all implementations inside `enum_` for clarity)

#### Int enums:

```cpp
enum_<Animal>("Animal", enum_value_type::integer)
 .value("Dog", Animal::Dog)
 .value("Cat", Animal::Cat)
```

emits:

```ts
// module.d.ts
export type Animal = 1 | 2;

interface EmbindModule {
    Animal: { Dog: 1, Cat: 2 },
}
```

#### String enums:

```cpp
enum_<Animal>("Animal", enum_value_type::string)
 .value("Dog", Animal::Dog)
 .value("Cat", Animal::Cat)
```

emits:

```ts
// module.d.ts
export type Animal = "Dog" | "Cat";

interface EmbindModule {
    Animal: { Dog: "Dog", Cat: "Cat" },
}
```

#### Legacy enums (default):

```cpp
enum_<Animal>("Animal", enum_value_type::legacy)
 .value("Dog", Animal::Dog)
 .value("Cat", Animal::Cat)
```

emits:

```ts
// module.d.ts
export type AnimalValue<T extends number> {
  value: T
}

export type Animal = AnimalValue<1> | AnimalValue<2>;

interface EmbindModule {
  Animal: { Dog: Animal<1>, Cat: Animal<2> };
}
```

This handles the various cases. It's not as close to TS enums, but again I don't think it's that much of a problem.

FelixNumworks:
Hi :) Any new on this @brendandahl ? Do you want me to develop the change I suggested above ?

brendandahl:
I also can't see a way to make the TS enums work well in a definition file, so the suggested change sounds good. For `enum_value_type::legacy`, I'd lean towards calling that `enum_value_type::default` or `enum_value_type::value`.


> I don't think the comparison problem you raised is that much of an issue. In the end, `Dog.a` and `Cat.a` are indeed equal ...

FWIW, that misses the point of strong type safety and why the `enum class` feature was added to `c++`. Accidently comparing the wrong enums can lead to unexpected behavior.

> Finally, using plain strings allows me to use the enum values without even needing the module:

I'd also suggest against doing this. Another value of enums is you can change the underlying values and not have to update your code every place that you hardcoded the value.

test/embind/embind.test.js
```diff
@@ -2031,6 +2031,26 @@ module({
         });
     });
 
+    BaseFixture.extend("enums with string values", function() {
```
brendandahl, line 2055: Can you add a test for integers enums as well?

test/other/embind_tsgen.d.ts
```diff
@@ -36,6 +36,8 @@ export interface BarValue<T extends number> {
 }
 export type Bar = BarValue<0>|BarValue<1>|BarValue<2>;
 
+export type Baz = 'valueA'|'valueB'|'valueC';
```
brendandahl: Looks like the definition files need to be updated.

FelixNumworks:
Sorry I previously wasn't done with my changes but I pushed early that's why there were some obvious flaws ^^"

I implemented the changes you requested and squashed all the commits together for a cleaner history. (Nothing changed but the tests since last review)

FelixNumworks:
It's still not working. I'll spend more time on this later this week. I'll re-request review when ready

brendandahl:
Will you have time to revisit this?

FelixNumworks:
Sorry I didn't have time recently. I'll work on this now and come back today with fixes

FelixNumworks:
The bugs should be fixed. I also enhanced the docs to make it clearer.

FelixNumworks:
I edited last commit because some tests were failing after merge with main

brendandahl:
Thanks!

PR #25258, opened 2025-09-11
Fix test browser configuring.

juj:
Fix two problems with test browser configuring refactor:

1. Each run of non-browser suites would print `"No EMTEST_BROWSER set. Defaulting to google-chrome"`. Simplest way is to remove this print. There is already a print `logger.info('Launching browser: %s', str(browser_args))` later down the line, so it should be fairly explicit what browser cmdline is being run.
2. if user has no EMTEST_BROWSER set, then on Windows line 2479 would throw on the `'"' not in EMTEST_BROWSER` check, when `EMTEST_BROWSER` was still `None`.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -2476,16 +2476,15 @@ def configure_test_browser():
   if not has_browser():
     return
 
+  if not EMTEST_BROWSER:
+    EMTEST_BROWSER = 'google-chrome'
+
   if WINDOWS and '"' not in EMTEST_BROWSER and "'" not in EMTEST_BROWSER:
     # On Windows env. vars canonically use backslashes as directory delimiters, e.g.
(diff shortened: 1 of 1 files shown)


test/common.py
```diff
@@ -2476,17 +2476,16 @@ def configure_test_browser():
   if not has_browser():
     return
 
+  if not EMTEST_BROWSER:
```
sbc100, line 2479: Seems odd that this code even runs at all then a non-browser test suite runs

sbc100:
I wonder why that windows CI bot doesn't failing on this?

juj:
Maybe it unconditionally sets EMTEST_BROWSER? That was at least the reason my CI didn't fail on this - I set EMTEST_BROWSER on all runs.

PR #25260, opened 2025-09-11
Fix for errors when using `OFFSCREENCANVAS_SUPPORT`

wenhanchong:
This PR addresses an error encountered while using `OFFSCREENCANVAS_SUPPORT` for pthreads.
Unable to find canvas in `findCanvasEventTarget` when canvas is `specialHTMLTargets`.
```
Uncaught TypeError: document.querySelector is not a function
    at findCanvasEventTarget
    at _wgpuInstanceCreateSurface
```

### DIFF
--- a/src/lib/libbrowser.js
+++ b/src/lib/libbrowser.js
@@ -712,9 +712,7 @@ var LibraryBrowser = {
     styleSheet.insertRule('canvas.emscripten { border: 1px solid black; cursor: none; }', 0);
   },
 
-#if !OFFSCREENCANVAS_SUPPORT
   emscripten_set_canvas_size__proxy: 'sync',
-#endif
   emscripten_set_canvas_size: (width, height) => Browser.setCanvasSize(width, height),
 
   emscripten_get_canvas_size__proxy: 'sync',
--- a/src/lib/libbrowser.js
+++ b/src/lib/libbrowser.js
@@ -712,7 +712,9 @@ var LibraryBrowser = {
     styleSheet.insertRule('canvas.emscripten { border: 1px solid black; cursor: none; }', 0);
   },
 
+#if !OFFSCREENCANVAS_SUPPORT
   emscripten_set_canvas_size__proxy: 'sync',
+#endif
   emscripten_set_canvas_size: (width, height) => Browser.setCanvasSize(width, height),
 
   emscripten_get_canvas_size__proxy: 'sync',
--- a/src/lib/libhtml5.js
+++ b/src/lib/libhtml5.js
@@ -360,8 +360,9 @@ var LibraryHTML5 = {
     // it will move in the GL.offscreenCanvases array between threads. Hence GL.offscreenCanvases
     // represents the set of OffscreenCanvases owned by the current calling thread.
 
+    return specialHTMLTargets[target]
     // First check out the list of OffscreenCanvases by CSS selector ID ('#myCanvasID')
-    return GL.offscreenCanvases[target.slice(1)] // Remove '#' prefix
+     || GL.offscreenCanvases[target.slice(1)] // Remove '#' prefix
     // If not found, if one is querying by using DOM tag name selector 'canvas', grab the first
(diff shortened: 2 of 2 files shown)


src/lib/libbrowser.js
```diff
@@ -712,7 +712,9 @@ var LibraryBrowser = {
     styleSheet.insertRule('canvas.emscripten { border: 1px solid black; cursor: none; }', 0);
   },
 
+#if !OFFSCREENCANVAS_SUPPORT
   emscripten_set_canvas_size__proxy: 'sync',
+#endif
   emscripten_set_canvas_size: (width, height) => Browser.setCanvasSize(width, height),
```
juj: The `emscripten_set_canvas_size()` and `emscripten_get_canvas_size()` APIs are deprecated, and used in the old GLUT/GLEW/EGL/SDL1.4 era of Emscripten code.

There is a deprecation node in the header file emscripten/emscripten.h, but it is a bit tricky to find.

This function should not be used in applications that target OffscreenCanvas or WebGPU. I am not sure if adding a `#if !OFFSCREENCANVAS_SUPPORT` here will have intended consequences. (At least they are not tested)

Check out emscripten/html5.h API for the modern/stable API to adjust canvas size, i.e.

```
EMSCRIPTEN_RESULT emscripten_set_canvas_element_size(const char *target __attribute__((nonnull)), int width, int height);

// Returns the canvas.width & canvas.height properties.
EMSCRIPTEN_RESULT emscripten_get_canvas_element_size(const char *target __attribute__((nonnull)), int *width, int *height);

EMSCRIPTEN_RESULT emscripten_set_element_css_size(const char *target __attribute__((nonnull)), double width, double height);
EMSCRIPTEN_RESULT emscripten_get_element_css_size(const char *target __attribute__((nonnull)), double *width, double *height);
```

That access `canvas.{width,height}` and `canvas.style.{width,height}` properties.

src/lib/libhtml5.js
```diff
@@ -360,8 +360,9 @@ var LibraryHTML5 = {
     // it will move in the GL.offscreenCanvases array between threads. Hence GL.offscreenCanvases
     // represents the set of OffscreenCanvases owned by the current calling thread.
 
+    return specialHTMLTargets[target]
     // First check out the list of OffscreenCanvases by CSS selector ID ('#myCanvasID')
-    return GL.offscreenCanvases[target.slice(1)] // Remove '#' prefix
+     || GL.offscreenCanvases[target.slice(1)] // Remove '#' prefix
```
juj: This change looks good, though I think the `specialHTMLTargets[target]` check should be after line 368 below that has

```
|| (target == 'canvas' && Object.keys(GL.offscreenCanvases)[0])
```
i.e. the intent of `findCanvasEventTarget()` is to look for offscreenCanvases first, and take that if found, and if not, then revert to original `findEventTarget()` behavior.

src/lib/libbrowser.js
wenhanchong: Noted, will remove this change and use `emscripten_set_canvas_element_size`

src/lib/libhtml5.js
wenhanchong: Good point, switched the priority around as requested.

src/lib/libhtml5.js
```diff
@@ -365,6 +365,8 @@ var LibraryHTML5 = {
     // If not found, if one is querying by using DOM tag name selector 'canvas', grab the first
     // OffscreenCanvas that we can find.
      || (target == 'canvas' && Object.keys(GL.offscreenCanvases)[0])
+    // If not found, check specialHTMLTargets
```
juj, line 368: Looks like this also needs changing

```
$findCanvasEventTarget__deps: ['$GL', '$maybeCStringToJsString'],
```
above to
```
$findCanvasEventTarget__deps: ['$GL', '$maybeCStringToJsString', '$specialHTMLTargets'],
```

so that the variable will be pulled in to code.

PR #25263, opened 2025-09-11
Run `@is_slow_test` test first in `parallel_testsuite.py`

sbc100:
Previously we were relying special naming of tests (e.g. `test_xxx_foo`) and using reverse alphabetic sort to run them first, but we dropped the `test_xxx` convention in favor of the `@is_slow_test` decorator.

Ultimately it would be nice to merge this default sort with the `--failing-fast-and-slow` flag but for now they are still separate.

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -114,10 +114,7 @@ class ParallelTestSuite(unittest.BaseTestSuite):
     # Remove any old stale list of flaky tests before starting the run
     utils.delete_file(common.flaky_tests_log_filename)
 
-    # If we are running with --failing-and-slow-first, then the test list has been
-    # pre-sorted based on previous test run results. Otherwise run the tests in
-    # reverse alphabetical order.
-    tests = list(self if self.failing_and_slow_first else self.reversed_tests())
+    tests = self.get_sorted_tests()
     contains_browser_test = any(test.is_browser_test() for test in tests)
     use_cores = cap_max_workers_in_pool(min(self.max_cores, len(tests), num_cores()), contains_browser_test)
     errlog(f'Using {use_cores} parallel test processes')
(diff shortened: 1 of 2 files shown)


sbc100:
I don't actually seem to see much/any improvement when running this locally,  but it seems to match the intended ordering which was always to have slow tests run first.

juj:
I'll defer this to @kripken: how important is it anymore to be able to run fast tests first/defer slow tests last? I.e. is the `test_zzz_` manual ordering still important for you?

Ordering slow tests first does help CI in general.

`browser.test_zzz_source_map` should continue to be last, since it leaves behind a stray browser instance.

test/parallel_testsuite.py
```diff
@@ -179,17 +176,22 @@ def update_test_results_to(test_name):
...
+  def get_sorted_tests(self):
+    """A list of this suite's tests, sorted with the @is_slow_test tests first.
 
-    Many of the tests in test_core are intentionally named so that long tests
-    fall toward the end of the alphabet (e.g. test_the_bullet). Tests are
-    loaded in alphabetical order, so here we reverse that in order to start
-    running longer tasks earlier, which should lead to better core utilization.
-
-    Future work: measure slowness of tests and sort accordingly.
+    Future work: measure and store the speed of tests each test sort more accurately.
```
juj, line 182: I think this Future work is probably now resolved

sbc100: I would like it to work on the emscripten CI which currently uses a fresh checkout each time, so we would need to either (1) stash the last run results in storage / cache somewhere CI can restore it from.  or (2) check in a list of test timing, which we periodically update.

sbc100:
> I'll defer this to @kripken: how important is it anymore to be able to run fast tests first/defer slow tests last? I.e. is the `test_zzz_` manual ordering still important for you?
> 
> Ordering slow tests first does help CI in general.
> 
> `browser.test_zzz_source_map` should continue to be last, since it leaves behind a stray browser instance.

Actually its the other way round, `test_zzz_source_map` comes last in the serial runner, but comes *first* in the parallel runner.  This change doesn't change that, it only *also* takes into account the `@is_slow_test` decorator.

Before this change:  Tests were sorted in reverse alphabetical order so `test_zzz` naming was the only way to make it run first.

After this change: Tests are still sorted in reverse alphabetical order, but *also* `@is_slow_test` come first.

juj:
Ah right, that makes sense. LGTM.

PR #25267, opened 2025-09-11
Skip dynamic linking tests in GLOBAL_BASE overridden browser test modes.

juj:
These tests cannot be run e.g. in `browser64_2gb` modes.

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -2285,6 +2285,8 @@ void *getBindBuffer() {
     self.btest_exit('openal/test_openal_buffers.c', cflags=['--preload-file', test_file('sounds/the_entertainer.wav') + '@/'])
 
   def test_runtimelink(self):
+    if self.get_setting('GLOBAL_BASE'):
+      self.skipTest('GLOBAL_BASE is not compatible with SIDE_MODULE')
     create_file('header.h', r'''
       struct point {
         int x, y;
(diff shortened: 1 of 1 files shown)


test/test_browser.py
```diff
@@ -2285,6 +2285,8 @@ def test_openal_buffers(self):
     self.btest_exit('openal/test_openal_buffers.c', cflags=['--preload-file', test_file('sounds/the_entertainer.wav') + '@/'])
 
   def test_runtimelink(self):
+    if self.get_setting('GLOBAL_BASE'):
+      self.skipTest('GLOBAL_BASE is not compatible with SIDE_MODULE')
```
sbc100, line 2289: In other places we do `if not self.is_2gb() and not self.is_4gb(): ` for this.

sbc100, line 2289: I'm a little confused here since we do run the whole `browser_2gb` test suite in CI

sbc100: Indeed, this test seems to pass for me.  `./test/runner browser_2gb.test_runtimelink` and `browser64_4gb.test_runtimelink` both pass.

juj: Here's what I got:

http://clbri.com:8010/api/v2/logs/60196/raw_inline
http://clbri.com:8010/api/v2/logs/60149/raw_inline

sbc100: Oh I see the issue here.  The `browser64_2gb` config is lacking the `-Wno-unused-command-line-argument` flag that the other configs add.

Can you revert this change and instead add `-Wno-unused-command-line-argument` to `browser64_2gb`?

Issue #25270, opened 2025-09-11
browser.test_wasm_worker_lock_async_acquire is flaky


juj:
The test seems to be racy to occassionally postMessage() to a Wasm Worker after the test has finished, resulting in

```
======================================================================
FAIL: test_wasm_worker_lock_async_acquire (test_browser.browser_2gb.test_wasm_worker_lock_async_acquire)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/unittest/case.py", line 651, in run
    self._callTestMethod(testMethod)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/unittest/case.py", line 606, in _callTestMethod
    if method() is not None:
       ~~~~~~^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 1049, in resulting_test
    return func(self, *args)
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 668, in metafunc
    f(self, *args, **kwargs)
    ~^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/test_browser.py", line 5253, in test_wasm_worker_lock_async_acquire
    self.btest_exit('wasm_worker/lock_async_acquire.c', cflags=['--closure=1', '-sWASM_WORKERS'])
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/common.py", line 2835, in btest_exit
    return self.btest(filename, *args, **kwargs)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #25271, opened 2025-09-11
Mark test_wasm_worker_lock_async_acquire flaky.

juj:
https://github.com/emscripten-core/emscripten/issues/25270

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -5203,6 +5203,7 @@ Module["preRun"] = () => {
 
   # Tests emscripten_lock_async_acquire() function.
   @also_with_minimal_runtime
+  @flaky('https://github.com/emscripten-core/emscripten/issues/25270')
   def test_wasm_worker_lock_async_acquire(self):
     self.btest_exit('wasm_worker/lock_async_acquire.c', cflags=['--closure=1', '-sWASM_WORKERS'])
 


.circleci/config.yml
```diff
@@ -315,6 +315,7 @@ commands:
           environment:
             EMTEST_DETECT_TEMPFILE_LEAKS: "0"
             EMTEST_HEADLESS: "1"
+            EMTEST_RETRY_FLAKY: "3"
```
kripken: What are these changes for?

juj: When I added the `@flaky` annotation in this PR, CircleCI would just complain about the flaky test that it is failing.

The CircleCI bots don't seem to be restarting flaky tests in browser suite. Not sure if that happened after PR https://github.com/emscripten-core/emscripten/pull/25246 , or some other refactor.

So I added the `EMTEST_RETRY_FLAKY` directives in this PR so that CircleCI will retry the flaky browser tests.

kripken: I see. I haven't followed all the recent work on flakiness, so sgtm but deferring to @brendandahl @sbc100

.circleci/config.yml
```diff
@@ -995,6 +1002,8 @@ jobs:
             "
   test-browser-firefox-wasm64:
     executor: focal
+    environment:
+      EMTEST_RETRY_FLAKY: "3"
```
sbc100: Can you revert the changes to this file, and we can discuss this is a seperate PR?

There is an existing mechanism in `.circleci/config.yml` in the form of `set-retry-flaky-tests`.   TBH I can't remember why I wanted it only for PRs.. but any refactoring can/should be separate from marking test_wasm_worker_lock_async_acquire itself as flaky.

lgtm without the changes in this file.

juj: > Can you revert the changes to this file, and we can discuss this is a seperate PR?

That would mean landing this in a way that breaks CircleCI. I.e. before I did this change, all the CI bots started showing red, as they started failing on this test.

juj: Reverted the EMTEST_RETRY_FLAKY now, but it didn't pass in CircleCI, as adding the `@flaky` stopped the flaky test from being re-run.

sbc100: Yes, sorry, I guess I should have said that we need some further discussion/thought about how/when/where to enable `EMTEST_RETRY_FLAKY` more widely.  

lgtm to this change now even if the CI is still red, or if you like we can have the above discussion here before landing.  either way sgtm.

PR #25273, opened 2025-09-12
Update SDL3 port to version 3.2.22

chkuendig:
This fixed some issues with polling SDL_EVENT_KEY_DOWN events (I got multiple events for a single JS event), but generally probably a good idea to keep this up to date.


chkuendig:
Looks like merging #25274 caused a conflict. I rebased to resolve that.

PR #25274, opened 2025-09-12
Update Vorbis and Ogg ports

chkuendig:
The current ogg and vorbis ports are ancient (2015) and pull in repos from the old _emscripten-ports_ github org. From what I can tell, they do not have any patches and both pre-date Xiph.org's migration to Github, so presumably are just old mirrors:
- https://github.com/emscripten-ports/Vorbis/releases/tag/version_1 is the same as https://github.com/xiph/vorbis/commit/7187e7a48f0c3ba32cc080f6bc3d921fe4ec6cc2
- https://github.com/emscripten-ports/Ogg/releases/tag/version_1 is the same as xiph/ogg/commit@9b2ba419aecb4a1c97114545d57174593dc13111 

This updates both to the most recent update.(ogg one before the most recent as I haven't yet tested the most recent update)

### DIFF
--- a/tools/ports/ogg.py
+++ b/tools/ports/ogg.py
@@ -6,8 +6,8 @@
 import os
 import shutil
 
-TAG = 'version_1'
-HASH = '929e8d6003c06ae09593021b83323c8f1f54532b67b8ba189f4aedce52c25dc182bac474de5392c46ad5b0dea5a24928e4ede1492d52f4dd5cd58eea9be4dba7'
+VERSION = '1.3.5'
+HASH = 'e7b5f5d469090b66dbb33634591ae6de41af3a5644c10488e59ba7428ed78912e208a2c0fbb5003ec5b7eb2a0102a2f85cecb21fa9512d790139ecc45b6d03f4'
 
 
 def needed(settings):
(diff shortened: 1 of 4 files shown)


tools/ports/ogg.py
```diff
@@ -6,23 +6,23 @@
...
+HASH = 'e7b5f5d469090b66dbb33634591ae6de41af3a5644c10488e59ba7428ed78912e208a2c0fbb5003ec5b7eb2a0102a2f85cecb21fa9512d790139ecc45b6d03f4'
 
 
 def needed(settings):
   return settings.USE_OGG
 
 
 def get(ports, settings, shared):
-  ports.fetch_project('ogg', f'https://github.com/emscripten-ports/ogg/archive/{TAG}.zip', sha512hash=HASH)
+  ports.fetch_project('ogg', f'https://github.com/xiph/ogg/releases/download/v{VERSION}/libogg-{VERSION}.zip', sha512hash=HASH)
```
sbc100, line 18: We should look into that kind of patches (if any) were applied in the https://github.com/emscripten-ports/ogg/ repo.   Where there any changes from upstream?

sbc100 (approved):
Sorry I didn't see that you already verified no downstream changes.

lgtm if the tests pass!

Thanks for looking into this.

Once this change lands I will archive the emscripten-ports repos.

sbc100:
Can you add an entry to the ChangeLog?

chkuendig:
> Can you add an entry to the ChangeLog?

Added.

chkuendig:
Are these test failures related to this change? They seem to be happening in completely unrelated parts (similar to #25273 which encounters the same problems)

PR #25275, opened 2025-09-14
Fix parallel browser test harness to also work with Firefox on Windows.

juj:
This is a cleaned up version 2 of #25237.

-------

Two problems that I run into with the parallel browser harness on Windows with Firefox:

1. On Windows, all the browser windows open up in the exact same X,Y coordinate. Firefox has "is browser in the background" detection, which causes all but one test foreground harness to hang when running any `requestAnimationFrame()` based test.

2. Killing a Firefox browser process with

```py
proc = subprocess.Popen(['firefox'])
proc.kill()
```

does not work. This is because Firefox is a multiprocess browser, and does not have a process hierarchy where the first launched process would be some kind of a master process. (this is an old known issue, that can be seen also in existing emrun.py code) There seems to be about 12 processes that come alive when running `subprocess.Popen(['firefox'])`.

--------------

To fix these issues:

1. ask the Windows Win32 API to move each spawned process to its own X,Y location. Use a file-based multiprocessing lock + counter file to get unique counter to each process.

2. Snapshot alive processes before spawning Firefox, and after. Then acquire the delta of these two to find which process IDs correspond to the launched Firefox.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1133,8 +1133,12 @@ jobs:
       # Run browser tests as well.
       - run-tests-firefox-windows:
           title: "browser on firefox on windows"
+          # skip browser.test_glbook, as it requires UNIX tool 'make'
+          # skip browser.test_sdl2_mixer_wav_dash_l, fails to build on Windows
           test_targets: "
             browser
+            skip:browser.test_glbook
+            skip:browser.test_sdl2_mixer_wav_dash_l
             "
       - upload-test-results
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1130,8 +1130,10 @@ jobs:
       # Run browser tests as well.
       - run-tests-firefox-windows:
           title: "browser on firefox on windows"
-          # skip browser.test_glbook, as it requires UNIX tool 'make'
+          # skip browser.test_glbook, as it requires mingw32-make, which is not
+          # installed on CircleCI.
           # skip browser.test_sdl2_mixer_wav_dash_l, fails to build on Windows
+          # on CircleCI (works locally)
           test_targets: "
             browser
             skip:browser.test_glbook
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -456,9 +456,6 @@ commands:
             echo "Running browser tests"
             echo "-----"
             test/runner << parameters.test_targets >>
-            # posix and emrun suites are disabled because firefox errors on
-            #  "Firefox is already running, but is not responding."
-            # TODO: find out a way to shut down and restart firefox
       - upload-test-results
   test-sockets-chrome:
     description: "Runs emscripten sockets tests under chrome"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -220,6 +220,13 @@ commands:
       - emsdk-env
       - bootstrap
       - pip-install
+      - when:
+          # We only set EMTEST_RETRY_FLAKY on pull requests.  When we run
+          # normal CI jobs on branches like main we still want to be able to
+          # detect flakyness.
+          condition: ${CIRCLE_PULL_REQUEST}
+          steps:
+            - set-retry-flaky-tests
   remove-linux-binaries:
     description: "Remove linux binaries from workspace"
     steps:
@@ -246,13 +253,6 @@ commands:
         type: string
         default: ""
     steps:
-      - when:
-          # We only set EMTEST_RETRY_FLAKY on pull requests.  When we run
-          # normal CI jobs on branches like main we still want to be able to
-          # detect flakyness.
-          condition: ${CIRCLE_PULL_REQUEST}
-          steps:
-            - set-retry-flaky-tests
       - run:
           name: run tests (<< parameters.title >>)
           environment:
@@ -387,6 +387,71 @@ commands:
             # that are not included by "./embuilder build ALL".  For example the
             # PIC version of libSDL which is used by test_sdl2_misc_main_module
             export EM_FROZEN_CACHE=""
+            echo "-----"
+            echo "Running browser tests"
+            echo "-----"
+            test/runner << parameters.test_targets >>
+            # posix and emrun suites are disabled because firefox errors on
+            #  "Firefox is already running, but is not responding."
+            # TODO: find out a way to shut down and restart firefox
+      - upload-test-results
+  run-tests-firefox-windows:
+    description: "Runs emscripten tests under firefox on Windows"
+    parameters:
+      test_targets:
+        description: "Test suites to run"
+        type: string
+      title:
+        description: "Name of given test suite"
+        type: string
+        default: ""
+    steps:
+      - run:
+          name: download firefox
+          shell: powershell.exe -ExecutionPolicy Bypass
+          command: |
+            # To download Firefox, we must first figure out what the latest Firefox version name is.
+            # This is because there does not exist a stable/static URL to download latest Firefox from.
+            $html = Invoke-WebRequest "https://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-central/"
+            $zipLink = $html.Links |
+              Where-Object { $_.href -match "firefox-.*\.en-US\.win64\.zip$" } |
+              Select-Object -Last 1 -ExpandProperty href
+
+            # Download Win64 Firefox.
+            $nightlyUrl = "https://archive.mozilla.org$zipLink"
+            $outZip = "$env:TEMP\firefox_nightly.zip"
+
+            Write-Host "Downloading latest Firefox Nightly: $nightlyUrl"
+            Invoke-WebRequest -Uri $nightlyUrl -OutFile $outZip
+
+            # Extract to user home directory
+            $installDir = Join-Path $env:USERPROFILE "firefox"
+            if (Test-Path $installDir) { Remove-Item -Recurse -Force $installDir }
+            Expand-Archive -LiteralPath $outZip -DestinationPath $installDir
+
+            # Set environment variable for tests
+            $ffExe = Join-Path $installDir "firefox.exe"
+            [System.Environment]::SetEnvironmentVariable("EMTEST_BROWSER", $ffExe, "User")
+            $env:EMTEST_BROWSER = $ffExe
+            Write-Host "EMTEST_BROWSER set to $env:EMTEST_BROWSER"
+      - run:
+          name: run tests (<< parameters.title >>)
+          environment:
+            EMTEST_LACKS_GRAPHICS_HARDWARE: "1"
+            # TODO(https://github.com/emscripten-core/emscripten/issues/24205)
+            EMTEST_LACKS_SOUND_HARDWARE: "1"
+            EMTEST_LACKS_WEBGPU: "1"
+            # OffscreenCanvas support is not yet done in Firefox.
+            EMTEST_LACKS_OFFSCREEN_CANVAS: "1"
+            EMTEST_DETECT_TEMPFILE_LEAKS: "0"
+            EMTEST_HEADLESS: "1"
+            EMTEST_CORES: "2"
+            DISPLAY: ":0"
+          command: |
+            # There are tests in the browser test suite that using libraries
+            # that are not included by "./embuilder build ALL".  For example the
+            # PIC version of libSDL which is used by test_sdl2_misc_main_module
+            set EM_FROZEN_CACHE=
             echo "-----"
             echo "Running browser tests"
             echo "-----"
@@ -1064,6 +1129,13 @@ jobs:
       - run-tests:
           title: "sockets.test_nodejs_sockets_echo*"
           test_targets: "sockets.test_nodejs_sockets_echo*"
+      - upload-test-results
+      # Run browser tests as well.
+      - run-tests-firefox-windows:
+          title: "browser on firefox on windows"
+          test_targets: "
+            browser
+            "
       - upload-test-results
 
   test-mac-arm64:
--- a/test/common.py
+++ b/test/common.py
@@ -944,8 +944,7 @@ def force_delete_dir(dirname):
   """Deletes a directory. Returns whether deletion succeeded."""
   if not os.path.exists(dirname):
     return True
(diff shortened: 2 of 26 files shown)


test/common.py
```diff
@@ -938,8 +941,26 @@ def make_dir_writeable(dirname):
 
 
 def force_delete_dir(dirname):
-  make_dir_writeable(dirname)
-  utils.delete_dir(dirname)
+  """Deletes a directory. Returns whether deletion succeeded."""
+  if not os.path.exists(dirname):
+    return True
+  if os.path.isfile(dirname):
```
sbc100: How about `assert not os.path.isfile(dirname)` since it seems like a logical error to call this on a non-directory.

juj:
This is the last PR remaining that is blocking me from testing my CI directly on `emscripten/main` rather than my own fork. Would be good to get this in so that I can migrate to target my CI on `emscripten/main`.

test/common.py
```diff
@@ -2559,7 +2655,31 @@ def browser_open(cls, url):
 
     browser_args = shlex.split(browser_args)
     logger.info('Launching browser: %s', str(browser_args))
-    cls.browser_proc = subprocess.Popen(browser_args + [url])
+
+    with FileLock(browser_spawn_lock_filename) as count:
+      # Firefox is a multiprocess browser. Killing the spawned process will not
+      # bring down the whole browser, but only one browser tab. So take a delta
+      # snapshot before->after spawning the browser to find which subprocesses
+      # we launched.
```
sbc100: Does this comment only apply to windows?

test/common.py
```diff
@@ -2542,9 +2632,15 @@ def browser_open(cls, url):
       if worker_id is not None:
         # Running in parallel mode, give each browser its own profile dir.
         browser_data_dir += '-' + str(worker_id)
-      if os.path.exists(browser_data_dir):
-        utils.delete_dir(browser_data_dir)
+
+      # Delete old browser data directory. If we cannot (the data dir is in use on Windows),
+      # switch to another dir.
+      while not force_delete_dir(browser_data_dir):
+        browser_data_dir += '-another'
```
sbc100: This seems like it could generated `-another-another-another-another`?    Should we use a counter and do `_1`, `_2` etc?

Also, like the the rest of this PR I'd would love to not run this loop except on windows.

sbc100 (commented):
I'd like to find a way keep this code isolated to only running on windows?    We don't need any of that `FileLock` stuff except on windows right?

Perhaps it could even move into its own file?    `test/browser_launcher.py`?

sbc100:
I'd also love to add some windows + firefox testing on the emscripten CI as part of this so that we can feel free to refactor without breaking stuff.

juj:
> I'd like to find a way keep this code isolated to only running on windows? We don't need any of that `FileLock` stuff except on windows right?

I was thinking that the FileLock could serve as a mechanism to solve https://github.com/emscripten-core/emscripten/pull/25069#discussion_r2334580939 later as well.

The overhead of running the FileLock mechanism on all platforms should be trivial and negligible. I can certainly refactor it out, though I figured your review would nit on two separate/duplicate implementations to launch the browser if I did. To me the logic reads cleaner with one common launcher impl.

test/common.py

test/common.py
juj: Yes, it seems to. Updated the comment.

test/common.py
juj: Indeed it can, and does. I wanted to keep the logic simple. Updated to use `_number`

juj:
> I'd also love to add some windows + firefox testing on the emscripten CI as part of this so that we can feel free to refactor without breaking stuff.

That is a good idea. I think that is good to develop in a separate PR.

juj:
> > I'd also love to add some windows + firefox testing on the emscripten CI as part of this so that we can feel free to refactor without breaking stuff.
> 
> That is a good idea. I think that is good to develop in a separate PR.

Ok, updated this PR to include Firefox on Windows testing.

juj:
Looks like CircleCI now also passes here on the new Firefox on Windows tests.

test/common.py
```diff
@@ -2561,7 +2669,31 @@ def browser_open(cls, url):
...
+      # browser might decide that the tab is backgrounded, and not load a test,
+      # or it might not tick rAF()s forward, causing tests to hang.
+      if worker_id is not None and WINDOWS and is_firefox():
+        # On Firefox on Windows we needs to track subprocesses that got created
+        # by Firefox. Other setups can use 'browser_proc' directly to terminate
+        # the browser.
+        cls.browser_procs = list(set(procs_after).difference(set(procs_before)))
+        # Wrap window positions on a Full HD desktop area modulo primes.
+        for proc in cls.browser_procs:
+          move_browser_window(proc.pid, (300 + count * 47) % 1901, (10 + count * 37) % 997)
```
sbc100, line 2706: Instead of putting `if WINDOWS` in through this block can we make the entire thing windows-only:

```
if not WINDOWS:
  cls.browser_procs = [subprocess.Popen(browser_args + [url])]
else:
  ....
```

Perhaps append `url` to browser_args first?

test/runner.py
```diff
@@ -564,6 +564,8 @@ def set_env(name, option_value):
 
   # Remove any old test files before starting the run
   utils.delete_file(common.flaky_tests_log_filename)
+  utils.delete_file(common.browser_spawn_lock_filename)
+  utils.delete_file(f'{common.browser_spawn_lock_filename}_counter')
```
sbc100, line 568: Is it not possible to store the count in the lock file itself?  (i.e. can we avoid separate counter file?)

sbc100 (commented):
Wow, that so great having windows browser tests running!   Thanks for taking care of that.

I think maybe we should split them into their own running, but I'm happy to do that as a followup.

Regarding the process tracking issue ("Firefox is a multiprocess browser, and does not have a process hierarchy where the first launched process would be some kind of a master process."), do you know why it only effect firefox and not chrome (which is also multi-process?   I'm just curious if we could maybe get this fixed and remove all this extra code one day?

test/common.py
```diff
@@ -2561,7 +2669,31 @@ def browser_open(cls, url):
...
+      # Make sure that each browser window is visible on the desktop. Otherwise
+      # browser might decide that the tab is backgrounded, and not load a test,
+      # or it might not tick rAF()s forward, causing tests to hang.
+      if worker_id is not None and WINDOWS and is_firefox():
+        # On Firefox on Windows we needs to track subprocesses that got created
+        # by Firefox. Other setups can use 'browser_proc' directly to terminate
+        # the browser.
+        cls.browser_procs = list(set(procs_after).difference(set(procs_before)))
+        # Wrap window positions on a Full HD desktop area modulo primes.
+        for proc in cls.browser_procs:
```
brendandahl, line 2705: This shouldn't be needed in headless mode. Headless should bypass the window active tracking stuff.

.circleci/config.yml
```diff
@@ -395,6 +395,68 @@ commands:
...
+        type: string
+        default: ""
+    steps:
+      - run:
+          name: download firefox
+          shell: powershell.exe -ExecutionPolicy Bypass
+          command: |
+            # To download Firefox, we must first figure out what the latest Firefox version name is.
+            # This is because there does not exist a stable/static URL to download latest Firefox from.
+            $html = Invoke-WebRequest "https://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-central/"
```
brendandahl, line 415: I lean towards beta, so we're not living so near the bleeding edge.

test/runner.py
juj: I did that first, but got odd behavior with O_EXCL when O_CREAT was not always unconditionally passed. So reverted to this simpler two-file form

.circleci/config.yml
sbc100: This is pre-existing.  We already use firefox-nightly elsewhere in this file.    We can consider changing separately.

test/common.py
juj: ok I moved to a separate function.

test/common.py
juj: Good point, updated.

.circleci/config.yml
juj: Linux Firefox runs also download Nightly, so I did not want to diverge there. Maybe if we want to test beta, we'd switch both Linux and Windows Firefoxes at the same time, as a separate PR?

juj:
> do you know why it only effect firefox and not chrome (which is also multi-process?

I haven't stress tested browser tests on Chrome on Windows yet, so not quite sure.

> I'm just curious if we could maybe get this fixed and remove all this extra code one day?

Maybe. But one has to be careful.

The need to move windows is only present on Windows. On Linux Mint 22 and Raspberry Pi 5 OS at least, the Linux windowing system cascades the windows automatically for Firefox, so they won't be interpreted to be in background.

On macOS, all Firefox windows do stack up in identical locations, but curiously this is not a problem there: Firefox still thinks all those windows are visible. Maybe this is because of the active program vs active window distinction that is present in the macOS windowing system, or maybe Firefox just doesn't implement the is-visible logic there. Not sure.

So the active process tracking is needed for the purposes of moving the Firefox browser windows only on Windows OS.

On Windows, in addition to being able to move browser windows, Firefox processes need to be tracked for clean termination. Refactoring this code requires very special care, because when all tests are green in the harness, i.e. "the happy path", then the browser harness does not have a need to ever force-terminate and restart any browser instance.

So even if this process tracking was plain removed, it might not show up immediately as a problem, if every test in the browser harness is green. I.e. in green state, the `browser_restart()` function will never need to be run.

When refactoring or stress-testing, it is important to manually place some browser tests in a hanging state to ensure that `browser_restart()` is actually getting called during the test run.

sbc100:
> > do you know why it only effect firefox and not chrome (which is also multi-process?
> 
> I haven't stress tested browser tests on Chrome on Windows yet, so not quite sure.
> 
> > I'm just curious if we could maybe get this fixed and remove all this extra code one day?
> 
> Maybe. But one has to be careful.
> 
> The need to move windows is only present on Windows. On Linux Mint 22 and Raspberry Pi 5 OS at least, the Linux windowing system cascades the windows automatically for Firefox, so they won't be interpreted to be in background.
> 
> On macOS, all Firefox windows do stack up in identical locations, but curiously this is not a problem there: Firefox still thinks all those windows are visible. Maybe this is because of the active program vs active window distinction that is present in the macOS windowing system, or maybe Firefox just doesn't implement the is-visible logic there. Not sure.
> 
> So the active process tracking is needed for the purposes of moving the Firefox browser windows only on Windows OS.
> 
> On Windows, in addition to being able to move browser windows, Firefox processes need to be tracked for clean termination. Refactoring this code requires very special care, because when all tests are green in the harness, i.e. "the happy path", then the browser harness does not have a need to ever force-terminate and restart any browser instance.
> 
> So even if this process tracking was plain removed, it might not show up immediately as a problem, if every test in the browser harness is green. I.e. in green state, the `browser_restart()` function will never need to be run.
> 
> When refactoring or stress-testing, it is important to manually place some browser tests in a hanging state to ensure that `browser_restart()` is actually getting called during the test run.

I see, so perhaps we should have some kind of stress test most the restarts the browser for each test.  

Even if we did that, how would we know for sure that we didn't still have the old browsers lying around?    Especially in headless mode.  For that matter, how did you notice this issue in headless mode?  Or are you running locally in headfull mode?

sbc100:
Just to be clear,  if we are able to shut down the FF instance fully, would we still also need the window moving logic?

juj:
> Even if we did that, how would we know for sure that we didn't still have the old browsers lying around? Especially in headless mode.

Checking with `len(list_processes_by_name('firefox.exe')) > 0` will accurately tell if there are old browsers lying around, as long as we make an assumption that user does not have unrelated Firefox browser instances open in the background. (which CI won't have)

This will work also in headless mode.

> For that matter, how did you notice this issue in headless mode? Or are you running locally in headfull mode?

I currently only run browser tests in headful mode.

> Just to be clear, if we are able to shut down the FF instance fully, would we still also need the window moving logic?

Yes, these two needs (tracking processes to be able to move windows) and (tracking processes to be able to shut down FF instances fully) are two orthogonal needs to do process tracking.

sbc100:
> Yes, these two needs (tracking processes to be able to move windows) and (tracking processes to be able to shut down FF instances fully) are two orthogonal needs to do process tracking.

But do we still need to move the windows if we can shutdown the old FF windows?   Or is the moving only needed because the old window lingers?

juj:
> But do we still need to move the windows if we can shutdown the old FF windows? Or is the moving only needed because the old window lingers?

The moving is needed because if the harness launches, say, 8 browser instances, then on Windows on Firefox, all those browser instances launch in the exact same (x, y, width, height) window rectangle. Seven of those Firefoxes will decide that the browser tab is not visible, so Firefox will not bother to even load up the Emscripten test page contents. Only the topmost Firefox window will load up an Emscripten test.

So the seven background Firefoxes will time out in a hang after five minutes, while only the topmost window makes progress.

End result is that most of the tests do pass in the suite, save for a random selection that got allocated into a background browser window in the first place.

sbc100:
> > But do we still need to move the windows if we can shutdown the old FF windows? Or is the moving only needed because the old window lingers?
> 
> The moving is needed because if the harness launches, say, 8 browser instances, then on Windows on Firefox, all those browser instances launch in the exact same (x, y, width, height) window rectangle. Seven of those Firefoxes will decide that the browser tab is not visible, so Firefox will not bother to even load up the Emscripten test page contents. Only the topmost Firefox window will load up an Emscripten test.
> 
> So the seven background Firefoxes will time out in a hang after five minutes, while only the topmost window makes progress.
> 
> End result is that most of the tests do pass in the suite, save for a random selection that got allocated into a background browser window in the first place.

Ah I see so it relates the parallel runner.  Got it.  Should not be needed with `-j1`?

juj:
Yes, indeed, if running in singlethreaded mode, moving the Firefox windows is not needed. This is realized with the `if worker_id is not None:` checks: if running the browser harness in singlethreaded mode, `worker_id = None`, and so the window moving logic is not applied.

PR #25283, opened 2025-09-15
Fix get_native_triple() on Windows-on-ARM devices.

juj:
Fixes

```
other.test_bad_triple
other.test_dot_a_all_contents_invalid
other.test_native_link_error_message
```
on Windows-on-ARM.

### DIFF
--- a/test/clang_native.py
+++ b/test/clang_native.py
@@ -25,7 +25,8 @@ def get_native_triple():
 
   arch = {
       'aarch64': 'arm64',
-      'arm64': 'arm64',
+      'arm64': 'arm64', # Python on Apple Silicon ARM64 reports lowercase arm64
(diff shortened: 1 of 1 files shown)


test/clang_native.py
```diff
@@ -25,7 +25,8 @@ def get_native_triple():
 
   arch = {
       'aarch64': 'arm64',
-      'arm64': 'arm64',
+      'arm64': 'arm64', # Python on Apple Silicon ARM64 reports lowercase arm64
+      'ARM64': 'arm64', # Python on Windows-on-ARM reports uppercase ARM64
       'x86_64': 'x86_64',
       'AMD64': 'x86_64',
   }[platform.machine()]
```
sbc100, line 32: How about `platform.machine().lower()` here?

juj: Considered it, but I'd rather see more cases of this from Python before going there.

PR #25290, opened 2025-09-17
Make sockets suite more graceful to fail with user facing instructions

juj:
Make sockets suite more graceful to fail with user facing instructions, and add env. var. EMTEST_LACKS_PYTHON_WEBSOCKIFY=1 to allow testing in shipping configuration.

### DIFF
--- a/test/test_sockets.py
+++ b/test/test_sockets.py
@@ -24,6 +24,20 @@ from tools.shared import EMCC, path_from_root, run_process, CLANG_CC
 
 npm_checked = False
 
+EMTEST_LACKS_PYTHON_WEBSOCKIFY = int(os.getenv('EMTEST_LACKS_PYTHON_WEBSOCKIFY', '0'))
+
+
+def requires_python_websockify(func):
+  assert callable(func)
+
+  @common.wraps(func)
+  def decorated(self, *args, **kwargs):
+    if EMTEST_LACKS_PYTHON_WEBSOCKIFY:
+      return self.skipTest('python websockify based tests are disabled')
+    return func(self, *args, **kwargs)
+
+  return decorated
+
 
 def clean_processes(processes):
   for p in processes:
@@ -61,7 +75,10 @@ class WebsockifyServerHarness:
       process = Popen([os.path.abspath('server')])
       self.processes.append(process)
 
-    import websockify  # type: ignore
+    try:
+      import websockify  # type: ignore
+    except ModuleNotFoundError:
+      raise Exception('Unable to import module websockify. Run "python3 -m pip install websockify" or set environment variable EMTEST_LACKS_PYTHON_WEBSOCKIFY=1 to skip this test.')
 
     # start the websocket proxy
     print('running websockify on %d, forward to tcp %d' % (self.listen_port, self.target_port), file=sys.stderr)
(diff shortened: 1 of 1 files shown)


test/test_sockets.py
```diff
@@ -24,6 +24,20 @@
 
 npm_checked = False
 
+EMTEST_LACKS_PYTHON_WEBSOCKIFY = int(os.getenv('EMTEST_LACKS_PYTHON_WEBSOCKIFY', '0'))
```
sbc100: IIRC we decided its fine to the test suite to require `pip install -r requirements-dev.txt`?

In that case, I don't think we need this opt out?  (same for psutil change you are working on, right?)

test/test_sockets.py
```diff
@@ -61,7 +75,10 @@ def __enter__(self):
       process = Popen([os.path.abspath('server')])
       self.processes.append(process)
 
-    import websockify  # type: ignore
+    try:
+      import websockify  # type: ignore
+    except ModuleNotFoundError:
+      raise Exception('Unable to import module websockify. Run "python3 -m pip install websockify" or set environment variable EMTEST_LACKS_PYTHON_WEBSOCKIFY=1 to skip this test.') from None
```
sbc100: TIL about `from None` does that make the error easier to read?

Would `exit_with_error()` make it even more easy to ready in that case?

test/test_sockets.py
juj: > (same for psutil change you are working on, right?)

psutil is currently installed for all users as a runtime dependency, and not just as a dev dependency.

websockify is not being shipped as a end user dependency.

I am testing the suites in end user configuration to ensure that nothing slips through the cracks that would require a dev dependency. So I want to skip all these tests (even though it does reduce coverage - that is ok for me)

test/test_sockets.py
juj: Yes, the `from None` does make the error easier to read.

`exit_with_error()` would stop the whole harness from running.

test/test_sockets.py
sbc100: Ok, in that case perhaps we can re-use the existing `requires_dev_dependency` decorator and rename the corresponding env var from `EMTEST_SKIP_NODE_DEV_PACKAGES` to `EMTEST_NO_DEV_PACKAGES`?

sbc100: In the long run I wonder if we can perhaps try to find a way to run these extra tests without accidentally allowing emscripten itself to use these packages?

Perhaps we could find a way to force emscripten itself to use a hermetic python and node that doesn't have visibility of the dev packages (we could ever assert in emscripten itself that these packages are not visible).  ?

juj: > Ok, in that case perhaps we can re-use the existing `requires_dev_dependency` decorator and rename the corresponding env var from `EMTEST_SKIP_NODE_DEV_PACKAGES` to `EMTEST_NO_DEV_PACKAGES`?

Hmm, I'm not feeling so hot about that, I find this more documentative - distinguishing between node vs python packages is useful to find "the lay of the land", and also because the steps to install the dependencies are different (invoke `npm install` vs invoke `pip install`)

sbc100: Fair enough, but maybe at least use a single one for all python dev packages (like we do for node packages today).

juj: Hmm ok, maybe `EMTEST_SKIP_PYTHON_DEV_PACKAGES` then? Updated.

sbc100 (approved):

PR #25296, opened 2025-09-17
[ci] Split windows browser testing into its own runner. NFC

sbc100:


sbc100:
As a followup I hope to make this run faster by making it depend on the `build-linux` libraries like the other browser testers.

juj (approved):

sbc100:
Hmm.. I'm seeing `skipped 'too many unresponsive tests, skipping remaining tests'` very consistently.. which make me think that #25275 did not actually enable any actual browser testing.  I think maybe it was running with `EMTEST_BROWSER=0` by mistake.

sbc100:
See #25311

sbc100:
I'm seeing a log of failures and browser restarts now.

Rather then try to debug and fix all the current failures, I think we should land this change with a subset of working tests, since anything is an improvement over the current situation of having no browsers tests running at all.

sbc100:
Ok, I'm going to land this with only 2 tests enabled.  We can move forward from there by enabling more.

juj:
sgtm

PR #25298, opened 2025-09-17
Use standard symbol export mechanism for memory and table exports. NFC

sbc100:


src/jsifier.mjs
```diff
@@ -734,7 +734,7 @@ function(${args}) {
         // but we declare it as an 'undefined'.  It then gets assigned manually
         // once the wasm module is available.
         // TODO(sbc): This is kind of hacky, we should come up with a better solution.
-        var isDirectWasmExport = WASM_ESM_INTEGRATION && mangled == 'wasmTable';
+        var isDirectWasmExport = mangled == 'wasmTable';
```
kripken, line 739: Why did this change? I can't figure it out from other parts of the PR.

sbc100: Because this hack, which used to be ESM_INTEGRATION only, is now needed in the more general case.

e.g. `__indirect_function_table` is now exported as `wasmTable` using the normal wasm export mechanism, which means the declaration of `wasmTable: undefined` in libcore.js` would now be a duplicate declaration.

This hack here suppresses the declaration of `wasmTable: undefined` from `libcore.js`

sbc100: I added more context to the comment.  Note that there is a TODO here to clean this up.

src/jsifier.mjs
```diff
@@ -731,10 +731,12 @@ function(${args}) {
       } else if (typeof snippet == 'undefined') {
         // wasmTable is kind of special.  In the normal configuration we export
         // it from the wasm module under the name `__indirect_function_table`
-        // but we declare it as an 'undefined'.  It then gets assigned manually
-        // once the wasm module is available.
+        // but we declare it as an 'undefined' in `libcore.js`.
+        // Since the normal export mechanism will declar this variable we don't
```
kripken:
```suggestion
        // Since the normal export mechanism will declare this variable we don't
```

src/jsifier.mjs
```diff
@@ -780,7 +782,7 @@ function(${args}) {
       }
 
       let docs = LibraryManager.library[symbol + '__docs'];
-      if (docs) {
+      if (docs && contentText != '') {
```
kripken, line 787:
```suggestion
      // Add the docs if they exist and if we are emitting something here.
      if (docs && contentText != '') {
```

test/code_size/audio_worklet_wasm.js
```diff
@@ -235,7 +235,7 @@ function y() {
         C = a.m;
         B = a.n;
         Y = a.o;
-        A = a.k;
+        A = A = a.k;
```
kripken, line 238: Odd closure doesn't optimize this better...

kripken (approved):

juj:
It looks like this regressed `omitexports0.test_asyncify_longjmp` among other tests in the `omitexports0` suite.

juj:
Other failures in `strict` and `strict_js` suites: http://clbri.com:8010/#/builders/11/builds/828

sbc100:
Thanks, I'll take a look at those tomorrow, and add the specific failing tests from those suites to emscripten CI

Issue #25137, opened 2025-09-02
Separate code size tests to its own dedicated suite?


juj:
Something that came to mind when staring failures like this:

<img width="874" height="630" alt="Image" src="https://github.com/user-attachments/assets/9dce160f-6e8d-4f56-8470-853898e65fa1" />

It might be a good idea to separate all code size tests to its own suite `test_codesize.py` and run in `codesize.test_xxx` namespace.

Then for CircleCI maybe the code size tests could be run in a task of its own. That way one would be able to see at a glance if it is a code size rebaselining problem that caused the code size regression?

sbc100:
Good idea. I thing @brendandahl suggested this too.

We could start by adding EMTEST_SKIP_CODESIZE which we could enable on all the builders except the code size one.  That could then also be used by the emscripten-releases builder which currently hardcode some patterns: https://chromium.googlesource.com/emscripten-releases/+/refs/heads/main/bots.json#136: 

```
--test-param=other,skip:other.*_codesize*,skip:other.*_code_size*,skip:other.test_cmake_find_pkg_config
```

Or we could just move them..either approach sounds reasonable to me.

PR #25305, opened 2025-09-18
Split codesize tests into separate test file. NFC

sbc100:
This makes it easier to run all the code size tests.

test_other.py is currently 16,000 lines so splitting this out is good to that problem too.

Fixes: #25137

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,8 +27,8 @@ jobs:
         name: emscripten-${{ env.VERSION }}
         path: emscripten-${{ env.VERSION }}.tar.bz2
 
-  check-expectations:
-    name: Check Expectations
+  codesize-checks:
+    name: Codesize Checks
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
     steps:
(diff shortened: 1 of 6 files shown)


juj (approved):
Nice stuff! I wonder if it would be feasible in the future to migrate code size tests to be their own item on the CircleCI test runs? Then the code size failures would be visible on GitHub page without having to click on any CircleCI links (which are really slow to load at least for me)

sbc100:
> Nice stuff! I wonder if it would be feasible in the future to migrate code size tests to be their own item on the CircleCI test runs? Then the code size failures would be visible on GitHub page without having to click on any CircleCI links (which are really slow to load at least for me)

Yes, I think that is already done as part of this change.

The codesize test are now only run in GitHub CI as part of the Check Expectations runner.  No other runner now runs these tests.

sbc100:
(This does mean we have lots a little coverage here since the `-mac` and `-windows` runners not longer run the subset of codesize tests that were marked as cross platform).

However we will be running the full set of codesize tests on all platforms on the emscripten-releases waterfall, so we the coverage will still be there, just a little down the line.

kripken:
Where can I see these tests running on CI?

sbc100:
Under "Check Expectations".   Should I rename that to "Codesize Checks", maybe?

juj:
> Under "Check Expectations". Should I rename that to "Codesize Checks", maybe?

I think that would work. Could always rename again if there's other checks to pile up later.

PR #9064, opened 2019-07-23
Use mainReadsParams metadata from wasm-emscripten-finalize

kripken:
This lets the wasm backend path not emit code to set up argc/argv if main doesn't need them. It also avoids that code bringing in some string and stack allocation support (metadce improvements are from the latter).

This reduces `--closure 1 -O3` size of hello world by almost 10%.

PR #25307, opened 2025-09-18
Implement support for reading argc and argv in MINIMAL_RUNTIME.

juj:
Implement support for reading argc and argv in MINIMAL_RUNTIME. With the help of the existing MAIN_READS_PARAMS machinery (thanks to https://github.com/emscripten-core/emscripten/pull/9064), this is only-pay-for-what-you-use.

Note that unlike the regular runtime that locates `argv[0]` to be the script name, MINIMAL_RUNTIME will place the shell interpreter name in `argv[0]` and the script name in `argv[1]`, to align with Node.js and Python `process.argv` and `sys.argv` behavior.

In a browser run, regular runtime has a dummy `"./this.program"` in `argv[0]`. Instead of a hardcoded mock value in `argv[0]`, the current page URL (e.g. `"https://localhost:5932/foo/bar/baz.html"`) is located in `argv[0]` instead, followed by the search query parameters in `argv[1]` onwards.


sbc100:
> Note that unlike the regular runtime that locates argv[0] to be the script name, MINIMAL_RUNTIME will place the shell interpreter name in argv[0] and the script name in argv[1], to align with Node.js and Python process.argv and sys.argv behavior.

I don't think we should do that since argv is a C/C++ concept where we we expect the program name to be `argv[0]` and we expect its arguments to be argv[1] and above.

Also, it seems like a needless and possibly annoy different / discrepancy between MINIMAL_RUNTIME and regular runtime.  (For example you wouldn't need to update so many tests in this PR).

src/postamble_minimal.js
```diff
@@ -26,28 +26,61 @@ function exitRuntime(ret) {
...
+  // and retain their last-stored values between program startup and program
+  // termination."
+  // -> in particular this means that the stackAlloc() that we do below shall
+  // never be undone, and ideally should no longer be considered to be part of
+  // the stack. Though currently it will be. (TODO: figure if this will ever be
+  // a problem)
+  var arg, argc = args.length, argv = stackAlloc(argc * {{{ POINTER_SIZE }}} + {{{ POINTER_SIZE }}}),
+    argvIndex = argv >> {{{ 31-Math.clz32(POINTER_SIZE) }}};
+
+  for(arg of args) {{{ HEAPptr }}}[argvIndex++] = {{{ to64('stringToUTF8OnStack(arg)') }}};
```
sbc100: Space after `for`

sbc100: Use `makeSetValue` here with `*` to avoid having to use `to64` explicitly?

src/postamble_minimal.js
```diff
@@ -26,28 +26,61 @@ function exitRuntime(ret) {
...
+
+  // C standard (C17 §5.1.2.2.1/5): "The parameters argc and argv and the
+  // strings pointed to by the argv array shall be modifiable by the program,
+  // and retain their last-stored values between program startup and program
+  // termination."
+  // -> in particular this means that the stackAlloc() that we do below shall
+  // never be undone, and ideally should no longer be considered to be part of
+  // the stack. Though currently it will be. (TODO: figure if this will ever be
+  // a problem)
+  var arg, argc = args.length, argv = stackAlloc(argc * {{{ POINTER_SIZE }}} + {{{ POINTER_SIZE }}}),
```
sbc100: Can we put these one on each line?  At least the ones that contains assignments?

src/postamble_minimal.js
```diff
@@ -26,28 +26,61 @@ function exitRuntime(ret) {
...
+    argvIndex = argv >> {{{ 31-Math.clz32(POINTER_SIZE) }}};
```
sbc100: Whats wrong with just doing `argv / {{{ POINTER_SIZE }}}` here?

juj:
> I don't think we should do that since argv is a C/C++ concept where we we expect the program name to be `argv[0]` and we expect its arguments to be argv[1] and above.

The C standard states

```
If the value of argc is greater than zero, the array members argv[0] through argv[argc-1]
inclusive shall contain pointers to strings, which are given implementation-defined values
by the host environment prior to program startup
```
and
```
If the value of argc is greater than zero, the string pointed to by argv[0] represents the
program name; argv[0][0] shall be the null character if the program name is not available
from the host environment. If the value of argc is greater than one, the strings pointed to by
argv[1] through argv[argc-1] represent the program parameters.
```
here one can well argue that the program name is `node.exe`, and the rest of the parameters are the command line passed to the program.

Getting both the interpreter and the script name here does have utility value: one can read from `argv[0]` what is the current interpreter, and from `argv[1]` the current script.

... though now that I look at the core and other suite tests, there are quite a few of these that depend on argc/argv semantics. So maybe it would be too much of a hassle.

I'll drop it, in the future I suppose we could do a setting `ARGV0_CONTAINS_INTERPRETER=0/1` to help get the native behavior.

src/postamble_minimal.js
juj: Err, that'll work as well.

src/postamble_minimal.js
juj: ok

sbc100:
> Getting both the interpreter and the script name here does have utility value: one can read from `argv[0]` what is the current interpreter, and from `argv[1]` the current script.

Sure, I guess I can imagine wanting to know that name of the node interpreter I'm running on .. maybe?  Actually I'm struggling to think of why that would ever be useful inside of the emscripten program.

If you care about that node executable what you probably care about more is the node version, which the exec name doesn't really give you... you would need to go digging in `process.versions` or `process.version` for that (which is easy with a little EM_ASM)

src/postamble_minimal.js
```diff
@@ -26,28 +26,64 @@ function exitRuntime(ret) {
 }
 #endif
 
+{{{ globalThis.argc_argv = function(condition) {
+    if (!MAIN_READS_PARAMS) return '';
+    return `argc, ${MEMORY64 ? 'BigInt(argv)' : 'argv'}`;
```
sbc100: Can you do `argc, ${to64('argv')}` here?

juj: check

test/code_size/test_minimal_runtime_code_size_math.json
```diff
@@ -1,10 +1,10 @@
...
-  "a.wasm": 2678,
-  "a.wasm.gz": 1648,
-  "total": 3340,
-  "total_gz": 2141
+  "a.js": 847,
+  "a.js.gz": 505,
+  "a.wasm": 2715,
+  "a.wasm.gz": 1674,
+  "total": 4114,
+  "total_gz": 2552
```
juj: This maybe dramatic looking code size increase is correct and expected. The code size test in question does

```c
int main(int argc, char **argv) {
  double num = (double)argc / (double)(long)argv[0];
  double num2 = (double)argc / (double)(long)argv[1];
  return pow(num, num2) + log(num);
}
```
i.e. it calculates on the argc and argv values.. so it would have been calculating on a bunch of zeroes before.

sbc100 (approved):

PR #25308, opened 2025-09-18
Add a helper cast utility to BigInt that saves code size.

juj:


juj:
In https://github.com/emscripten-core/emscripten/pull/25307 at least when it lands.

juj:
Maybe on this line as well: https://github.com/emscripten-core/emscripten/blob/5f66c8af88beb2a7d66186e76245e28956e9d64e/src/postamble.js#L71

sbc100 (approved):
lgtm

I assume closure can also do this optimization?

juj:
> I assume closure can also do this optimization?

Doesn't look like it does.

test/other/test_parseTools.js
```diff
@@ -91,15 +91,15 @@ addToLibrary({
...
     // ignore the upper bits), in the same way that `BigInt64Array[X] = Y` does.
     // (see splitI16 in parseTools.js)
     _clearI64(ptr);
-    {{{ makeSetValue('ptr', '0', 0x1122334455667788AA, 'i64') }}};
+    {{{ makeSetValue('ptr', '0', 0x1122334455667788AAn, 'i64') }}};
     _printI64(ptr);
 
     _clearI64(ptr);
-    {{{ makeSetValue('ptr', '0', -0x1122334455667788AA, 'i64') }}};
+    {{{ makeSetValue('ptr', '0', -0x1122334455667788AAn, 'i64') }}};
```
juj: The comment in this test said

*"The current behaviour truncate (i.e. ignore the upper bits)"*

However, there are two problems with this comment:

a) truncation is not the same as ignoring upper bits. Truncation rounds towards zero, i.e reducing information from the lowest bits.

b) the current behavior is not to truncate, but actually it is arbitrary, and can round up or down at a whim, due to the magic JavaScript "print nice floats only" semantics. In the case of this test, the test code actually rounded up, rather than truncating, as the comment suggested.

Fixed this test to use BigInt literals directly.

TLDR: Before this change, there is a bug of double rounding. After this change, one source of rounding is fixed, so the test must change.

------------------

To understand specifically what happened in the test before:

Before this change, the value `0x1122334455667788AA` would first be captured as a JavaScript `Number` into ParseTools.mjs `makeSetValue()` function.

The value `0x1122334455667788AA` would be equal to `316059037807746189482`, but this value cannot be exactly represented as a JS Number.

This value is sandwiched between the following two 64-bit doubles:

```
316059037807746220032  // next larger number that is exactly representable in Number
316059037807746189482  // the precise value of 0x1122334455667788AA (not representable as 64-bit double)
316059037807746154496  // previous smaller number that is exactly representable in Number
```
However, `makeSetValue()` did not get called with any of these, but rather it got called with another Number
```
316059037807746200000
```
What is this number? Well, it is "nicest" value between `...154496 - ...220032`, i.e. the one with most trailing zeroes in it.

That is also the result if you typed in `(0x1122334455667788AA).toString()`.

So, previously the code would compute `(0x1122334455667788AA).toString()` in `makeSetValue()`, and then emit the text string `"BigInt(316059037807746200000)"` into the generated source code.

However, calculating `BigInt(316059037807746200000)` does not result in the BigInt `316059037807746200000n`, but it results in `316059037807746220032n`. That is because `316059037807746200000` is not a 64-bit double, but `316059037807746220032` is. Note that this is the very original value rounded up, rather than down. (unlike what the comment suggested)

After the PR, instead of emitting `"BigInt(316059037807746200000)"`, the code in makeSetValue() would then emit `"316059037807746200000n"`, which is more precise, not the same as `BigInt(316059037807746200000)`.

I.e.
```js
BigInt(316059037807746200000) == 316059037807746200000n 
```
returns false.

The left side rounds, whereas the right side is exact.

sbc100: Thanks for the detailed explanation!

Of of interest, do these test changes work without the `parseTools.js` change?  i.e. could we fix the test first before landing the parseTools.js update if we wanted to?

src/parseTools.mjs
```diff
@@ -262,14 +262,32 @@ function makeInlineCalculation(expression, value, tempVar) {
...
+    var maybeBigIntLiteral = Number(x.slice(0, -1));
+    if (Number.isInteger(maybeBigIntLiteral) && isFinite(maybeBigIntLiteral)) return x;
+  }
+  // Or if already a BigInt, then output as BigInt literal.
+  if (typeof x === 'bigint') return `${x}n`;
+  // Micro-size-optimization: if x is an integer literal, then we can append
+  // the suffix 'n' instead of casting to BigInt(), to get smaller code size.
+  var n = Number(x);
+  if (Number.isInteger(n) && isFinite(n)) return `${x}n`;
+  return `BigInt(${x})`;
```
sbc100, line 279: Is this complexity really worth it for just one or two locations?

src/parseTools.mjs
```diff
@@ -262,14 +262,32 @@ function makeInlineCalculation(expression, value, tempVar) {
...
 // XXX Make all i64 parts signed
 
+function castToBigInt(x) {
+  // If given string is already a "BigInt(something)", then no-op.
+  if (typeof x === 'string' && x.startsWith('BigInt(')) return x;
+  // Or if it is already a BigInt literal string, e.g. "42n".
+  if (typeof x === 'string' && x.endsWith('n')) {
+    var maybeBigIntLiteral = Number(x.slice(0, -1));
+    if (Number.isInteger(maybeBigIntLiteral) && isFinite(maybeBigIntLiteral)) return x;
+  }
```
sbc100: Do these cases actually happen in the codebase?   should we just assert that they cannot happen?   I worry that if they are not happening in our codebase then these codepaths are untested/unused.

src/parseTools.mjs
juj: Well, given that I wanted to change makeSetValue() to be able to write a

`{{{ makeSetValue('ptr', '0', 0x1122334455667788n, 'i64') }}};

precisely, then the above code lines are necessary, or the bits would not be preserved.

test/other/test_parseTools.js
juj: Well, looks like that is what I'm up against now, since we are running the test in no_bigint mode as well, which is now failing. Although I don't yet know how to fix this..

src/parseTools.mjs
juj: In this PR I changed one test to do

```
{{{ makeSetValue('ptr', '0', 0x1122334455667788AAn, 'i64') }}};
```
and I figured that it is best to be consistent and not leave it to a half-way house (or if I wasn't, I'd get reviewed with a bunch of "what if user passes BigInt(42) instead of 42n" or similar)

juj: Ok, I'll remove these.

juj:
I restricted the PR to apply to safe integers only, and removed the ability to specify bigint integer literals.

test/other/test_parseTools.js
```diff
@@ -91,9 +91,9 @@ addToLibrary({
     {{{ makeSetValue('ptr', '0', 0x12345678AB, 'i64') }}};
     _printI64(ptr);
 
-    // This value doesn't fit into i64.  The current behaviour truncate (i.e.
-    // ignore the upper bits), in the same way that `BigInt64Array[X] = Y` does.
-    // (see splitI16 in parseTools.js)
+    // This value doesn't fit into i64.  The current behaviour is
+    // in unspecified, subject to a double rounding problem. See
+    // note in castToBigInt() in parseTools.mjs.
```
sbc100, line 96: Should we add a FIXME here?

sbc100 (approved):

sbc100:
Maybe we should land #25307 first so that this change can show a codesize win?

juj:
> Maybe we should land #25307 first so that this change can show a codesize win?

Looks like we probably don't have any code size tests that both use argc/argv and run in MEMORY64 mode.

PR #25303, opened 2025-09-18
Fix "EMTEST_BENCHMARKERS=node-64 test/runner benchmark" run.

juj:
Fix `EMTEST_BENCHMARKERS=node-64 test/runner benchmark` run. Improve debug printing in emmake.py.

Fixes

```
benchmark.test_zzz_bullet
benchmark.test_zzz_lua_binarytrees
benchmark.test_zzz_lua_scimark
benchmark.test_zzz_lzma
````

for the suite to pass in node-64 benchmarker mode.

PR #25309, opened 2025-09-18
Fix f-string usage in emmake.py

sbc100:
Ruff was reporting:

```
emmake.py:60:19: SyntaxError: Cannot reuse outer quote character in f-strings on Python 3.8 (syntax was added in Python 3.12)
```

Instead use shlex.join like the other scripts do.


sbc100:
Somehow this slipped by in #25303

sbc100:
Landing TBR to green the tree since ruff and mypy tests have already passed.

juj:
Ops.. sorry, and thanks!

PR #25310, opened 2025-09-18
Move codesize test files. NFC

sbc100:
As a followup to #25305 this change moves all the codesize files into `test/codesize/...`.

This matches `test/other/...` and `test/core/...`.

This change basically moves all the files in `test/other/codesize` and `test/code_size` into a single `test/codesize` directory.


.github/workflows/ci.yml
```diff
@@ -45,16 +45,15 @@ jobs:
...
-          echo $EM_CONFIG
+          echo "emscripten config file: $EM_CONFIG"
           echo "EM_CONFIG=$EM_CONFIG" >> $GITHUB_ENV
           curl -# -L -o ~/emsdk-main.tar.gz https://github.com/emscripten-core/emsdk/archive/main.tar.gz
           tar -C ~ -xf ~/emsdk-main.tar.gz
           mv ~/emsdk-main ~/emsdk
           cd ~/emsdk
           ./emsdk install tot
           ./emsdk activate tot
-          echo "JS_ENGINES = [NODE_JS]" >> $EM_CONFIG
```
sbc100, line 56: This line was not needed (because its is the default)

PR #25312, opened 2025-09-18
[audioworklet] Fix audioworklet test shutdown hang.

brendandahl:
The audio worklet was still running after emscripten had exited. This caused the browser to hang intermittently.

### DIFF
--- a/test/webaudio/audioworklet_post_function.c
+++ b/test/webaudio/audioworklet_post_function.c
@@ -10,9 +10,12 @@
 //   2. post invocation from audio worklet thread to main thread
 _Atomic uint32_t callbackCount = 0;
 
+EMSCRIPTEN_WEBAUDIO_T context;
+
 void do_exit() {
   emscripten_out("do_exit");
   assert(callbackCount == 16);
+  emscripten_destroy_audio_context(context);
   emscripten_force_exit(0);
 }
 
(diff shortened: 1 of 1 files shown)


sbc100 (commented):
lgtm.. but should this happen automatically somehow?

sbc100:
ruff checks are fixed in main already

cwoffenden:
Are you able to reproduce this locally or only on CI?

juj:
Should this be a documented requirement for Emscripten developers that utilize AudioWorklet? I.e. do their web pages risk hanging (in Chrome?) if they do not close the worklet before closing the page?

cwoffenden:
> Should this be a documented requirement for Emscripten developers that utilize AudioWorklet?

I wonder if it's specific to how the test works? We don't see this in production (with millions of Chromebook users on the worst school hardware).

juj:
That sounds great. (Chrome CI tests should run on Chromebooks then? =)

Would be good to be able to capture what it is with this specific test, though ok to land this as a fix.

brendandahl:
> Are you able to reproduce this locally or only on CI?

Locally with `EMTEST_BROWSER="taskset -c 0 google-chrome-unstable"`. I can also reproduce a hang in`test_audio_worklet_*` and a similar fix there seems to work. As far as I can tell, there's no extra wasm or audio worklet code running after force_exit, so it seems like this may be a bug in chrome. I'm hoping we can maybe put together a smaller reproducer and file a bug.

brendandahl:
> lgtm.. but should this happen automatically somehow?

Do we handle cleanup on `emscripten_force_exit` automatically in other cases? It seems any `emscripten_create_audio_context` should be paired with a `emscripten_destroy_audio_context` by the user.

sbc100:
> > lgtm.. but should this happen automatically somehow?
> 
> Do we handle cleanup on `emscripten_force_exit` automatically in other cases? It seems any `emscripten_create_audio_context` should be paired with a `emscripten_destroy_audio_context` by the user.

We do shutdown pthreads automatically on exit, so it seems reasonable that we should shut down all the audio worklets too.

PR #25313, opened 2025-09-18
Fix --post-js scripts to execute after main() in MODULARIZE mode in MINIMAL_RUNTIME

juj:
Fix --post-js scripts to execute after main() in MODULARIZE mode. Fixes test `minimal0.test_unicode_js_library` after this change: https://github.com/emscripten-core/emscripten/pull/23157/files#r1889243405

### DIFF
--- a/src/postamble_minimal.js
+++ b/src/postamble_minimal.js
@@ -312,14 +312,17 @@ WebAssembly.instantiate(Module['wasm'], imports).then(/** @suppress {missingProp
 #if PTHREADS || WASM_WORKERS
 }
 
-// When running in a background thread we delay module loading until we have
-{{{ runIfMainThread('loadModule();') }}}
-#endif
-
-#if MODULARIZE
 // The semantics of MODULARIZE and --post-js foo.js scripts require that main()
 // should run before any of the --post-js scripts. Therefore await instantiation
 // here before reaching execution to the --post-js scripts to produce the
 // expected order.
+#if MODULARIZE
+{{{ runIfMainThread('loadModule(); await instantiatePromise;') }}}
+#else
+{{{ runIfMainThread('loadModule();') }}}
+#endif
+
+#elif MODULARIZE
 await instantiatePromise;
+
 #endif
--- a/src/postamble_minimal.js
+++ b/src/postamble_minimal.js
@@ -136,10 +136,19 @@ Module['wasm'] = base64Decode('<<< WASM_BINARY_DATA >>>');
 var wasmExports;
 #endif
 
+#if AUDIO_WORKLET || MODULARIZE
+var instantiatePromise;
+#endif
+
 #if PTHREADS || WASM_WORKERS
 var wasmModule;
 
+#if MODULARIZE
+async function loadModule() {
+#else
 function loadModule() {
+#endif
+
   assignWasmImports();
 #endif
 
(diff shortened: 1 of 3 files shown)


site/source/docs/tools_reference/emcc.rst
```diff
@@ -305,9 +305,9 @@ Options that are modified or new in *emcc* are listed below:
 
 ``--post-js <file>``
   [link]
-  Like ``--pre-js``, but emits a file *after* the emitted code. In addition to
-  this code being located after the emitted JS library code, this code will also
-  be executed after main() has run.
+  Like "--pre-js", but emits a file *after* the emitted code. In addition to
+  being located after the emitted JS library code, --post-js code will also
+  execute only after main() has run.
```
sbc100: Is this actually the behavior we want though? 

Is this true today?  I would have though that `WASM_ASYNC_COMPILATION=1` (which is the default) means that the code runs before we run main (while the wasm file is loading).

Which behaviour do we think is more useful / expected?

juj: In https://github.com/emscripten-core/emscripten/pull/23157/files#r1889243405 you had commented

```
This means that postjs code, which comes after main will now run after main like it would do with sync instantiation.

I think this is a pretty reasonable change.
```
so I wanted to highlight that as it is the behavior that the test suite depends upon.

juj: I would also be ok to leaving this unspecified (it is a code size increase in MINIMAL_RUNTIME after all, and not all users might need that guarantee). Although I wonder if that fights against the ChangeLog note in that PR

```
One side effect of this is that code in --post-js files will now be delayed until after module creation and after main runs. This matches the existing behaviour when using sync instantation (-sWASM_ASYNC_COMPILATION=0) but is an observable difference. (#23157)
 ```
 
 that mentions this to be the new behavior.

Comment (sbc100) [reply site/source/docs/tools_reference/emcc.rst]: I guess maybe we should be consistent, but for now it seems like `-sMODULARIZE` has already been updated to have the well-defined behaviour of running post-js after main.

For code built without `-sMODULARIZE` it currently depends if you build with async compilation or not, so its not as simple as saying that post-js code always runs after main.

So, perhaps revert the documentation change here until and we can separately consider how to document the current differences between modes (or remove them and then document the consistent new behaviour)?

Documenting this as always being one way would currently be inaccurate I think?

**Review by sbc100 — COMMENTED** [review]
lgtm % docs question

**Comment by juj** [comment]
> Do we perhaps have an existing test that verified this behaviour that we can now run in minimal runtime mode?

Yes, the test `test_unicode_js_library` which changed semantics in https://github.com/emscripten-core/emscripten/pull/23157/files#r1889243405 can be used to verify this, e.g. `minimal0.test_unicode_js_library`.

If this is the only test in the suite that depends on this behavior, then I think another way to resolve this test failure is to change the test or verify a `--pre-js` instead of a `--post-js`. The test intent in `test_unicode_js_library` is only to verify that inputting UTF8-encoded files in `--post-js` works. So one could reasonably assume that if UTF8 works in `--pre-js`, then it'll work in `--post-js` as well; and change the test to verify `--pre-js` instead.

**Comment by sbc100** [comment]
> minimal0.test_unicode_js_library

Can you add this test at line 708 of `.circleci/config.yml` to confirm the fix then?

I agree the intent of this test does not seem to be explicitly checking/verifying this ordering, but I'd be happy to land this with that tests as confirmation.   Alternatively you write a specialized new test for just this, but we can also do that as a followup when/if we decide to lock down the beaviour in all configurations.

File: src/postamble_minimal.js [hunk 2]
```diff
@@ -310,8 +319,19 @@ WebAssembly.instantiate(Module['wasm'], imports).then(/** @suppress {missingProp
 );
 
 #if PTHREADS || WASM_WORKERS
+
+#if MODULARIZE
+  await instantiatePromise;
```
sbc100, line 324: Maybe one day we can find a way to clean this up and just use `async/await` instead of have this global `instantiatePromise`.

lgtm for now though.

test/test_core.py
```diff
@@ -6409,6 +6409,9 @@ def test_unicode_js_library(self):
       self.assertContained('UnicodeDecodeError', err)
 
     self.cflags += ['-sMODULARIZE', '--js-library', test_file('unicode_library.js'), '--extern-post-js', test_file('modularize_post_js.js'), '--post-js', test_file('unicode_postjs.js')]
+    # In addition to verifying that --post-js files can contain UTF8 characters,
+    # this test verifies a specific semantic that in -sMODULARIZE mode, the
+    # content in --post-js files should be executed only after main() runs.
```
sbc100, line 6414: Can you reference the bug I just created here: https://github.com/emscripten-core/emscripten/issues/25346

sbc100:
Do you want to land this or wait for us to resolve #25346 (since it may change the behaviour here)

juj:
Probably better to shoot towards the final result. I'll close this for now.

PR #25318, opened 2025-09-19
Update posixtestsuite to deal with OOM issue on fast computers

juj:
A couple of tests in the posixtest suite hammer `pthread_kill()` and `pthread_cancel()` in an uncontrollably fast loop.

Since the event queue mailbox refactor, proxying `pthread_kill()` and `pthread_cancel()` allocate memory.

This causes fast CPUs like Ryzen 9800X3D 8c/16t and Apple M1 ARM to run out of memory while running the test:

http://clbri.com:8010/api/v2/logs/64925/raw_inline
http://clbri.com:8010/api/v2/logs/66008/raw_inline

Update the posixtest submodule to latest, which throttles these loops to sleep 1ms between signals.

### DIFF
--- a/test/test_posixtest.py
+++ b/test/test_posixtest.py
@@ -166,6 +166,7 @@ def make_test(name, testfile, browser):
             '-Wno-int-conversion',
             '-Wno-format',
             '-pthread',
+            '--profiling-funcs',
             '-sEXIT_RUNTIME',
             '-sTOTAL_MEMORY=512mb',
             '-sPTHREAD_POOL_SIZE=40']
--- a/test/test_posixtest.py
+++ b/test/test_posixtest.py
@@ -168,7 +168,7 @@ def make_test(name, testfile, browser):
             '-pthread',
             '--profiling-funcs',
             '-sEXIT_RUNTIME',
-            '-sTOTAL_MEMORY=512mb',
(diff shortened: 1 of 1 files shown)


sbc100:
Could you update the title to decribe that is changed?   How about "Update posixtestsuite to deal with OOM issue on fast computers"

test/test_posixtest.py
```diff
@@ -166,6 +166,9 @@ def f(self):
             '-Wno-int-conversion',
             '-Wno-format',
             '-pthread',
+            # Make sure all tests have callstacks to improve debuggability
+            # of log messages on CI runs.
+            '--profiling-funcs',
```
sbc100, line 171: I've always thought it might be good to add this flag to more of out test suite (for example the core0 suite) but @kripken was worried about the testing then not matching to defaults used by out users.

juj: I did have that in mind, and I agree. That is why I specifically added `--profiling-funcs` and not `-g` here.

Though in this case the posixtest suite is more of a satellite suite, and there will still be `core0` suite that verifies builds without debug symbols. (which includes multithreading tests). It is very unlikely that a test in posixtest suite would mask a failure by having `--profiling-funcs`, that the `core0` suite wouldn't also have.

PR #25320, opened 2025-09-19
Split some larger `other` tests into separate files. NFC

sbc100:


sbc100:
Maybe `test_switch64phi` should instead just be deleted?

PR #25321, opened 2025-09-19
Fix -sEXIT_RUNTIME in MINIMAL_RUNTIME build mode.

juj:
Fix -sEXIT_RUNTIME in MINIMAL_RUNTIME build mode. Fixes 13 tests in minimal0 suite, e.g. minimal0.test_static_variable.


src/postamble_minimal.js
```diff
@@ -80,7 +80,18 @@ function run() {
...
   // In regular exitRuntime mode, exit with the given return code from main().
-  exitRuntime(_main({{{ argc_argv() }}}));
+  try {
+    exitRuntime(_main({{{ argc_argv() }}}));
+  } catch(e) {
+    if (!e.match(/^exit\(\d+\)$/)) {
+      throw e;
+    }
+#if RUNTIME_DEBUG
+    else {
```
sbc100: No need for `else` block here I think.

juj:
> The `throw `exit(${code})`code is not dependent on`EXIT_RUNTIME` so maybe the catch here should not be either?

Then it would increase code size for no benefit.

If user is building with `-sEXIT_RUNTIME=0`, then they would likely not call `exit(0)` anywhere, hence emitting code to catch `exit()`s would be redundant.

src/postamble_minimal.js
```diff
@@ -87,9 +87,7 @@ function run() {
       throw e;
     }
 #if RUNTIME_DEBUG
-    else {
-      out(`main() called ${e}.`); // e.g. "main() called exit(0)."
-    }
+    out(`main() called ${e}.`); // e.g. "main() called exit(0)."
```
sbc100: Use `dbg` rather than `out` here.

src/postamble_minimal.js
```diff
@@ -83,12 +83,19 @@ function run() {
...
   } catch(e) {
-    if (!e.match(/^exit\(\d+\)$/)) {
+    var exitCode = e.match(/^exit\(\d+\)$/);
+    if (exitCode) {
       throw e;
     }
 #if RUNTIME_DEBUG
     dbg(`main() called ${e}.`); // e.g. "main() called exit(0)."
 #endif
+    // Report to Module that the program exited. TODO: Find a way to not emit
```
sbc100: Wrap this in `#if expectToReceiveOnModule('onExit')`.

juj: That's perfect - thanks.

PR #25322, opened 2025-09-19
Revise stack overflow checks in the core suite

juj:
Revise stack overflow and abort checks in the core suite to allow testing stack overflow tests also in minimal0 suite.

Checking for the presence of the `"Aborted("` prefix does not seem essential to any of these tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -706,6 +706,7 @@ jobs:
             wasmfs.test_fs_llseek_rawfs
             wasmfs.test_freetype
             minimal0.test_utf
+            minimal0.test_stack_overflow
             omitexports0.test_asyncify_longjmp
             strict.test_no_declare_asm_module_exports
             "
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -2815,7 +2815,7 @@ The current type of b is: 9
 
   def test_stack_overflow(self):
     self.set_setting('ASSERTIONS', 2)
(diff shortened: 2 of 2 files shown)


sbc100 (commented):
While not essential I think check for the `Aborted(` prefix does allow for fewer false positives.  

What about just adding the `Aborted(` prefix to `MINIMAL_RUNTIME`?

juj:
> While not essential I think check for the `Aborted(` prefix does allow for fewer false positives.

I don't think any of the prints here would have a big risk for a false positive.

> What about just adding the `Aborted(` prefix to `MINIMAL_RUNTIME`?

That is the kind of cruft that `MINIMAL_RUNTIME` wanted to avoid.. there is no essential reason for that except for debugging purposes, and those things pile up, like the initial cleanup from regular runtime -> minimal runtime showed.

sbc100:
> > While not essential I think check for the `Aborted(` prefix does allow for fewer false positives.
> 
> I don't think any of the prints here would have a big risk for a false positive.

Sure, I guess.  I could imagine we might have test that did something like `printf("Running stack overflow test")`.. but I guess the risk is fairly low.

It also makes the tests a little less clear in what they are testing.  Without this change I can clearly see that I'm expecting the test for fail with an `abort` call.. after this it less obvious the reader what is being tested for.

We could do something gross like `abort_message('stack_overflow)` which would conditionally add the prefix.. or maybe we could have the test output checker strip the `Aborted(` prefix when `MINIMAL_RUNTIME` is set.  

Neither of those options seem that great either though :)

juj:
Would this be ok to land?

PR #25319, opened 2025-09-19
Fix !DECLARE_ASM_MODULE_EXPORTS tests after #25298

sbc100:
In `!DECLARE_ASM_MODULE_EXPORTS` mode, all wasm symbols are programmatically exported onto the `Module` object.   After #25298, that includes the `wasmMemory` symbol.

This happens even for symbols that were not included in `EXPORTED_FUNCTIONS` or `EXPORTED_RUNTIME_SYMBOLS`.  In many of these tests `wasmMemory` was not being exported and so that `unexportedRuntimeSymbol` helper was installing a property accessor on the `Module` object, which did not allow for the assignment within `exportWasmSymbols`.

An alternative here would be to limit the list of symbols that `exportWasmSymbols` assigns to the `Module` object but that would only increase code size for very little benefit.

PR #25325, opened 2025-09-19
Fix omitexports0.test_emscripten_api after #25319

sbc100:


juj (approved):
Maybe there will be a nice clean way in the future, good to carry the FIXME.

PR #25326, opened 2025-09-19
[audioworklet] Fix test_audio_worklet shutdown hang.

brendandahl:


brendandahl:
Bummer, looks like there was a hang with on the last run in `test_audio_worklet_strict`.

brendandahl:
Maybe the hang is from and interaction with the previous test `test_audio_worklet_worker` that I haven't fixed. I'll land both the fixes and see if we get this anymore.

PR #25332, opened 2025-09-20
Fix sockets.test_nodejs_sockets_echo_connect_failure

juj:
On 3/4 of my systems (Windows, Linux and a macOS), I get `"Host is unreachable"`. On 1/4 of my systems, (Windows 10), I get a `"Connection refused"`.

### DIFF
--- a/test/test_sockets.py
+++ b/test/test_sockets.py
@@ -337,7 +337,7 @@ class sockets(BrowserCore):
       self.do_runf('sockets/test_sockets_echo_client.c', expected, cflags=['-DSOCKK=%d' % harness.listen_port] + args)
 
   def test_nodejs_sockets_connect_failure(self):
-    self.do_runf('sockets/test_sockets_echo_client.c', 'connect failed: Connection refused', cflags=['-DSOCKK=666'], assert_returncode=NON_ZERO)
+    self.do_runf('sockets/test_sockets_echo_client.c', r'connect failed: (Connection refused|Host is unreachable)', regex=True, cflags=['-DSOCKK=666'], assert_returncode=NON_ZERO)
 
   @requires_native_clang
   @requires_python_dev_packages


juj:
Landing to make use of CI time over the weekend.

test/test_sockets.py
```diff
@@ -337,7 +337,7 @@ def test_nodejs_sockets_echo(self, harness_class, port, args):
       self.do_runf('sockets/test_sockets_echo_client.c', expected, cflags=['-DSOCKK=%d' % harness.listen_port] + args)
 
   def test_nodejs_sockets_connect_failure(self):
-    self.do_runf('sockets/test_sockets_echo_client.c', 'connect failed: Connection refused', cflags=['-DSOCKK=666'], assert_returncode=NON_ZERO)
```
sbc100, line 340: Is the `r` prefix needed here?

sbc100:
Can you confirm that your linux system really reports "Host is unreachable".. I certainly see "Connection refused".  I wonder what the difference might be?

```
$ ./test/runner sockets.test_nodejs_sockets_connect_failure --browser=0 -v
...
-- begin program output --
got (expected) socket: 0.0.0.0:0 (0.0.0.0:0), size 9 (9)
got (expected) socket: 127.0.0.1:666 (127.0.0.1:666), size 13 (13)
connect failed: Connection refused
finish: 1
program exited (with status: 1), but keepRuntimeAlive() is set (counter=0) due to an async operation, so halting execution but not exiting the runtime or preventing further async execution (you can use emscripten_force_exit, if you want to force a true shutdown)
-- end program output --
ok

Ran 1 test in 0.814s

OK
```

juj:
> Can you confirm that your linux system really reports "Host is unreachable"..

Yeah, like I mentioned in the PR description, I do get Host is unreachable on three of my systems (Windows, Linux and Mac). On another Windows system I got a Connection refused.

sbc100:
Oh, I bet it depends on the local firewall rules or IP configuration, or something like that.

PR #25334, opened 2025-09-20
Fix emscripten_set_socket_*_callback() functions to work in Wasm64 mode.

juj:
Fix emscripten_set_socket_*_callback() functions to work in Wasm64 mode. Fixes test sockets64.test_sockets_async_bad_port.


juj:
Landing to make use of CI time over the weekend.

sbc100:
Do we not have any tests in other/core/browser that use these callbacks?

PR #25347, opened 2025-09-22
Building with -sTEXTDECODER=2 when targeting -sAUDIO_WORKLET is not catastrophic

juj:
Building with -sTEXTDECODER=2 when targeting -sAUDIO_WORKLET is not catastrophic since the user might very likely never attempt to interop any strings between JS and Wasm from the Audio Worklet thread. So leaving UTF8Decoder to be null there will be fine.

### DIFF
--- a/src/lib/libstrings.js
+++ b/src/lib/libstrings.js
@@ -9,9 +9,13 @@
 #endif
 
 addToLibrary({
+  // In -sAUDIO_WORKLET builds, TextDecoder will not exist in AudioWorkletGlobalScope,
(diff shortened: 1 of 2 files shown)


sbc100 (commented):
I suppose we should give the same treatment to `ENVIRONMENT_MAY_BE_SHELL`?   (This is other place we force the fallback IIRC).

juj:
> I suppose we should give the same treatment to `ENVIRONMENT_MAY_BE_SHELL`? (This is other place we force the fallback IIRC).

Hmm that could be a separate PR. I am not completely sure about that.

PR #25351, opened 2025-09-22
Optimize ENVIRONMENT_IS_AUDIO_WORKLET checks

juj:
Optimize ENVIRONMENT_IS_AUDIO_WORKLET checks and use it in more places to optimize code size.

### DIFF
--- a/src/modularize.js
+++ b/src/modularize.js
@@ -98,7 +98,7 @@ var isWW = {{{ nodeWWDetection() }}};
 #endif
 
 #if AUDIO_WORKLET
-isWW ||= typeof AudioWorkletGlobalScope !== 'undefined';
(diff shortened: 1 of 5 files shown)


src/modularize.js
```diff
@@ -98,7 +98,7 @@ var isWW = {{{ nodeWWDetection() }}};
 #endif
 
 #if AUDIO_WORKLET
-isWW ||= typeof AudioWorkletGlobalScope !== 'undefined';
```
sbc100, line 101: This one might not work since I believe its outside of the module scope where `ENVIRONMENT_IS_AUDIO_WORKLET` is defined.

juj:
> No codesize test cover this?

It does, just hadn't updated the PR. Wins a couple of bytes.

sbc100 (approved):

PR #25352, opened 2025-09-22
Add function emscripten_audio_context_sample_rate()

juj:
Given that we have a function `emscripten_audio_context_quantum_size(EMSCRIPTEN_WEBAUDIO_T audioContext)`, also add a function `emscripten_audio_context_sample_rate(EMSCRIPTEN_WEBAUDIO_T audioContext)` to easily get the sampling rate of the context graph.

Will need to add some tests etc. though putting this up already to organize PRs.


src/lib/libwebaudio.js
```diff
@@ -301,6 +301,14 @@ var LibraryWebAudio = {
     return emscriptenGetContextQuantumSize(contextHandle);
   },
 
+  emscripten_audio_context_sample_rate: (contextHandle) => {
+#if ASSERTIONS
+    assert(EmAudio[contextHandle], `Called emscripten_audio_context_sample_rate() with an invalid Web Audio Context handle ${contextHandle}`);
+    assert(EmAudio[contextHandle] instanceof (window.AudioContext || window.webkitAudioContext), `Called emscripten_audio_context_sample_rate() on handle ${contextHandle} that is not an AudioContext, but of type ${EmAudio[contextHandle]}`);
+#endif
```
sbc100, line 323: How about adding a `getAudioContext(handle)` helper that can take care of adding these assertions to all APIs that need to lookup a handler like this?

Then this would just be `emscripten_audio_context_sample_rate: (handle) => getAudioContext(handle)['sampleRate'],`

cwoffenden:
There are a bunch of properties like sample rate that may be worth adding to a struct passed to the callback function. Knowing the latency is useful, for example, when doing lip-synch or other timing based features. By passing these in a struct, the API could add them as they become either exposed or more interesting, and as non-breaking changes.

Most properties are static, so cheap to do, and dynamic things like time could be opt-in.

src/lib/libwebaudio.js
juj: > How about adding a `getAudioContext(handle)` helper that can take care of adding these assertions to all APIs that need to lookup a handler like this?

That would add to code size in non-ASSERTIONS builds. I think it's nice to see the flat assertions like this, since it helps with not having to read deep into nested sub-functions.

juj:
> There are a bunch of properties like sample rate that may be worth adding to a struct passed to the callback function.

By having separate functions, the code will only be executed when a user calls for it. They can then cache the value in a Wasm side variable.

This way users who do not need to access the sampling rate will not emit code for that.

If the sampling rate was always unconditionally populated, then there would be no way to opt out, and everyone's code size would be pessimized by code that executes unconditionally.

PR #25353, opened 2025-09-22
Optimize code size in MINIMAL_RUNTIME: not all users want to override INITIAL_MEMORY via Module.

juj:
Optimize code size in MINIMAL_RUNTIME: not all users want to override INITIAL_MEMORY via Module.

### DIFF
--- a/src/runtime_init_memory.js
+++ b/src/runtime_init_memory.js
@@ -30,7 +30,11 @@ function initMemory() {
   } else
 #endif
   {
+#if MINIMAL_RUNTIME
+    var INITIAL_MEMORY = {{{ INITIAL_MEMORY }}};
+#else
     var INITIAL_MEMORY = {{{ makeModuleReceiveExpr('INITIAL_MEMORY', INITIAL_MEMORY) }}};
+#endif
 
 #if ASSERTIONS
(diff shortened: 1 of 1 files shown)


src/runtime_init_memory.js
```diff
@@ -30,7 +30,11 @@ function initMemory() {
   } else
 #endif
   {
+#if MINIMAL_RUNTIME
+    var INITIAL_MEMORY = {{{ INITIAL_MEMORY }}};
+#else
     var INITIAL_MEMORY = {{{ makeModuleReceiveExpr('INITIAL_MEMORY', INITIAL_MEMORY) }}};
```
sbc100, line 36: The `makeModuleReceiveExpr` macro already takes care of this I think.  

IIRC, `MINIMAL_RUNTIME` sets `INCOMING_MODULE_JS_API` to empty by default which should make this expand to just `INITIAL_MEMORY`

juj: It looks like this does not happen by default, but testing this now, indeed if I manually set `-sINCOMING_MODULE_JS_API=[]`, then this does happen. So that's good solution for this, closing this PR.

sbc100:
Should we default `INCOMING_MODULE_JS_API` to empty under `MINIMAL_RUNTIME` like we do already in `STRICT` mode?   See `limit_incoming_module_api` in `link.py`.

juj:
Yeah, I think that would probably work out.

PR #25354, opened 2025-09-22
Avoid the Web Audio Worklet bootstrap processor node on browsers that support AudioWorkletScope port

juj:
Remove the redundant Web Audio Worklet bootstrap processor node after https://github.com/WebAudio/web-audio-api/issues/2456 has been fixed.

### DIFF
--- a/test/codesize/audio_worklet_wasm.expected.js
+++ b/test/codesize/audio_worklet_wasm.expected.js
@@ -7,7 +7,7 @@ function v(a) {
     m ||= {};
     m.wasm = a.G;
     y();
-    a.G = a.N = 0;
+    a.G = a.M = 0;
 }
 
 t && !r && (onmessage = a => {
@@ -191,7 +191,7 @@ var K = [], L = a => {
     e.addModule(m.js).then((() => {
         e.port || (e.port = {
             postMessage: p => {
-                p.M ? (e.D = new AudioWorkletNode(g, "em-bootstrap", {
+                p._boot ? (e.D = new AudioWorkletNode(g, "em-bootstrap", {
                     processorOptions: p
                 }), e.D.port.onmessage = h => {
                     e.port.onmessage(h);
@@ -200,7 +200,7 @@ var K = [], L = a => {
         });
         e.port.postMessage({
             _boot: 1,
-            O: ba++,
+            N: ba++,
             G: m.wasm,
             L: w,
             J: c,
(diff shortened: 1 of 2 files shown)


src/closure-externs/closure-externs.js
```diff
@@ -229,6 +229,7 @@ var registerProcessor = function(name, obj) {};
 var currentFrame;
 var currentTime;
 var sampleRate;
+var port;
```
sbc100: This seems like very generic name.  I wonder if we move these to `audio-worklet-externs.js`?

src/lib/libwebaudio.js
```diff
@@ -371,7 +371,7 @@ var LibraryWebAudio = {
     assert(UTF8ToString(sigPtr)[0] != 'v', 'Do NOT specify the return argument in the signature string for a call to emscripten_audio_worklet_post_function_sig(), just pass the function arguments.');
     assert(varargs);
 #endif
-    (audioContext ? EmAudio[audioContext].audioWorklet.bootstrapMessage.port : messagePort).postMessage({'_wsc': funcPtr, args: readEmAsmArgs(sigPtr, varargs) });
+    (audioContext ? EmAudio[audioContext].audioWorklet.port : messagePort).postMessage({'_wsc': funcPtr, args: readEmAsmArgs(sigPtr, varargs) });
```
sbc100: The `var messagePort;` was deleted from `audio_worklet.js` but I don't see it being added elsewhere.  Should it still be referenced here, or should it be just `port` (i.e. `globalThis.port`)?

juj:
> Do we not need to wait until we drop support for older browser? Or are audio worklets so new that we just assume evergreen browsers?

Originally I thought that maybe we could get by with assuming that it was an old issue that implementations have long had it resolved since, but unfortunately with closer testing I see that is not the case: https://issues.chromium.org/issues/446920095

I wrote a MOD player that does not like to launch in Chrome with the `.port` access, https://clbri.com/modplayer/ , but it only works in Firefox.

juj:
Also reported https://bugs.webkit.org/show_bug.cgi?id=299386

sbc100:
So I guess we can't land this without break chrome support?

juj:
I updated this to use a polyfilling code structure, it does work on Chrome and Safari now. (test failures were Closure errors)

sbc100 (approved):

PR #25358, opened 2025-09-22
Remove now unused Audio Worklet Module exports.

juj:
Earlier refactor that avoided the separate .js file also made these redundant; `src/audio_worklet.js` no longer accesses anything via `Module`.


sbc100 (approved):
Nice!

I though we had a codesize test that covers audio worklet?

juj:
Yeah, I just didn't bother updating those when splitting these changes to separate PRs, since they would all be out of date - I have to rebase these to land one by one to keep the code size tests intact.

PR #25359, opened 2025-09-22
Remove more dead python code found using `deadcode` module. NFC

sbc100:
The remaining dead code being reported is false positives I believe.

### DIFF
--- a/tools/feature_matrix.py
+++ b/tools/feature_matrix.py
@@ -42,7 +42,6 @@ class Feature(IntEnum):
   MEMORY64 = auto()
 
 
-default_features = {Feature.SIGN_EXT, Feature.MUTABLE_GLOBALS}
 disable_override_features = set()
 enable_override_features = set()
 
--- a/tools/maint/create_dom_pk_codes.py
+++ b/tools/maint/create_dom_pk_codes.py
@@ -237,29 +237,26 @@ def hash_all(k1, k2):
     else:
       hashes[h] = s[1]
       str_to_hash[s[1]] = h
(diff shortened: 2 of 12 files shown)


tools/feature_matrix.py
```diff
@@ -42,7 +42,6 @@ class Feature(IntEnum):
   MEMORY64 = auto()
 
 
-default_features = {Feature.SIGN_EXT, Feature.MUTABLE_GLOBALS}
```
kripken, line 45: Then where do we set the default features?

tools/maint/create_dom_pk_codes.py
```diff
@@ -237,29 +237,26 @@ def hash_all(k1, k2):
...
 
 # Last used perfect hash constants.  Stored here so that this script will
 # produce the same output it did when the current output was generated.
 k1 = 0x7E057D79
 k2 = 3
-perfect_hash_table = hash_all(k1, k2)
+str_to_hash = hash_all(k1, k2)
 
-while not perfect_hash_table:
+while not str_to_hash:
```
kripken, line 254: Wait, was this doing `not` on a tuple before, when `hash_all` returned a tuple?

sbc100: This was detecting the `return None` for when the function fails.

I verified the that script works and produces no changes to the files it generates.

tools/feature_matrix.py
sbc100: I don't think this variable was ever user or needed, even in the PR that added it: #17689

PR #25360, opened 2025-09-22
[embind] Re-allow references in val code.

brendandahl:
The static assertion here broke some existing code. In the future, I'd like to change this to support a pointer policy to allow references, but for now I need to get this working again for users.


system/include/emscripten/val.h
```diff
@@ -576,9 +576,6 @@ class EMBIND_VISIBILITY_DEFAULT val {
 
   template<internal::EM_INVOKER_KIND Kind, typename Policy, typename Ret, typename... Args>
   static Ret internalCall(EM_VAL handle, const char *methodName, Args&&... args) {
-    static_assert(!std::is_lvalue_reference<Ret>::value,
```
brendandahl, line 595: @RReverser another option is to not allow lvalue references to `std::is_arithmetic` types.

RReverser: In the issue you said we can allow it only with `allow_raw_pointers` - can we still do that?

brendandahl: Yeah, I hope to do that in a future PR after https://github.com/emscripten-core/emscripten/pull/24832 lands.

PR #16261, opened 2022-02-11
Consistent use of `typeof` operator in JS library code. NFC

sbc100:
I noticed that in some places we were using equality checks
(==/!=) and in others we were using identity checks (===/!==).

In general identity checks are safer, but since typeof is known
to always returns a string and the RHS of these checks is also
always a string we can choose to always prefer `!=` and `==`
over `!==` and `===` for the sake of code size (sad but true,
we care that much about each byte of output).

Perhaps a better solution would be to teach closure compiler
how to remove the extra `=` when its know that both sides of
the operator are the same type already.. but that would be a
much larger change.

PR #25364, opened 2025-09-23
Optimize ENVIRONMENT_IS_WASM_WORKER check when building with Audio Worklets enabled

juj:
Optimize ENVIRONMENT_IS_WASM_WORKER check when building with Audio Worklets enabled, for a code size saving of 7 bytes.

Before, Closure would emit:

```js
m = "em-ww" == globalThis.name,
n = !!globalThis.AudioWorkletGlobalScope,
n && (m = !0);
```

Afterwards, it emits
```js
n = !!globalThis.AudioWorkletGlobalScope,
m = "em-ww" == globalThis.name || n,
```


src/shell_minimal.js
```diff
@@ -63,22 +63,23 @@ if (ENVIRONMENT_IS_NODE) {
 }
 #endif
 
-#if WASM_WORKERS
+#if AUDIO_WORKLET
+var ENVIRONMENT_IS_AUDIO_WORKLET = typeof AudioWorkletGlobalScope !== 'undefined';
```
sbc100: You can just use `!=` here to save another byte.

We explicitly use `==` and `!=` when the type of both the LHS and the RHS can only be strings (as in this case).   See #16261 for more on this.

juj: The PR https://github.com/emscripten-core/emscripten/pull/25351 will cover that part. I'll land 25351 first, and then update this PR on top.

sbc100 (approved):

PR #25368, opened 2025-09-23
Fix possible ODR violation

j4niwzis:
Fix possible ODR violation (this is what clang says when building with C++20 modules and doesn't allow compilation to proceed further)


sbc100:
Nice, looks simpler too.

I'll let @brendandahl and @RReverser review since they know more C++ template stuff than I.

brendandahl (approved):

PR #25369, opened 2025-09-23
Another round of `deadcode` feedback and error suppression. NFC

sbc100:
There are still quite a few false positives after this but they are fall into basically 2 classes:

- Subclasses of TestCase, which are explicitly used
- Subclasses of Library, which are no explicitly user
- Settings, which are only ever read in the JS compiler.

I think we should probably wait until the `deadcode` tool at least supports returning error codes before we try to integrate into CI.  See https://github.com/albertas/deadcode/pull/36

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ lint.select = [
   "W",
   "YTT",
 ]
-
+lint.external = [ "D" ]
 lint.ignore = [
   "B011", # See https://github.com/PyCQA/flake8-bugbear/issues/66
   "B023",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -31,7 +31,7 @@ lint.select = [
   "W",
   "YTT",
 ]
-
+lint.external = [ "D" ]
```
kripken, line 34: What is this?

sbc100: This is the prefix used for `noqa` decorators.   They are mnemonics I guess so any error code starting in `D` relates to dead code.

PR #25370, opened 2025-09-23
Fix test_dlfcn_exceptions. NFC

sbc100:
This test was completely disabled.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -3707,12 +3707,9 @@ pre 9
 out!
 ''')
 
-  # TODO: make this work. need to forward tempRet0 across modules
-  # TODO Enable @with_all_eh_sjlj (the test is not working now)
   @needs_dylink
-  def zzztest_dlfcn_exceptions(self):
-    self.set_setting('DISABLE_EXCEPTION_CATCHING', 0)
-
+  @with_all_eh_sjlj
+  def test_dlfcn_exceptions(self):
     create_file('liblib.cpp', r'''
       extern "C" {
       int ok() {
@@ -3726,7 +3723,7 @@ out!
     self.build_dlfcn_lib('liblib.cpp')
 
     self.prep_dlfcn_main()
-    src = r'''
+    create_file('main.cpp', r'''
       #include <assert.h>
       #include <stdio.h>
       #include <dlfcn.h>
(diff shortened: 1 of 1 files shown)


test/test_core.py
```diff
@@ -3748,27 +3745,30 @@ def zzztest_dlfcn_exceptions(self):
...
-          printf("int %d\n", x);
+          printf("caught int: %d\n", x);
         }
 
+        /*
         try {
           printf("fail: %d\n", faill());
         } catch(double x) {
-          printf("caught %f\n", x);
+          printf("caught double: %f\n", x);
```
sbc100: What was the intent of this block here @kripken .. this catch block will never fire because it catching the wrong type.

test/test_core.py
```diff
@@ -3694,7 +3694,7 @@ def test_dlfcn_longjmp(self):
         return 0;
       }
       ''')
-    self.do_runf('main.c', '''go!
+    self.do_runf('main.cpp', '''go!
```
kripken: Don't you prefer C over C++ for such testcases?

test/test_core.py
kripken: I'm guessing the goal was to verify we don't catch the wrong type? That is the catch should not be reached.

test/test_core.py
sbc100: In this case are testing exceptions which are a C++-only thing

test/test_core.py
sbc100: If I uncomment this the exception will travel all the way to the top level and cause a crash/failure.

I guess I'll leave commented out for now?

test/test_core.py
kripken: Isn't this testcase testing longjmp? The one after it uses exceptions, but I don't see exceptions in this one.

kripken: (this is `test_dlfcn_longjmp`)

test/test_core.py
kripken: Maybe make the test not assume a returncode of 0, and verify that it did halt there? But if we have other tests doing that, maybe not important.

sbc100: I ended up implementing this using two levels of try/catch.  Seems to work fine.

test/test_core.py
sbc100: Damn.. no idea how that happened.

PR #25371, opened 2025-09-23
[empath-split] Fix windows' wasm-split path

aheejin:
We need the suffix `.exe` for Windows. This will fix the current waterfall failure:
https://ci.chromium.org/ui/p/emscripten-releases/builders/try/win/b8702928348541404113/overview


sbc100:
Maybe mark the test for as `@crossplatform` so it will run on windows in circleCI?

PR #25372, opened 2025-09-23
Import/export Wasm memory as `wasmMemory`. NFC

sbc100:
This makes the name used on the wasm side match the name used on the JS side, which simplifies things somewhat.

Oddly its a 2 byte codesize regression in some cases, but I think thats just minification noise.


sbc100:
Actually this approach might not work since standalone wasm requires the export to be called "memory".

Issue #25323, opened 2025-09-19
wasm-sourcemap.py invocation change drops PYTHONPATH, breaks Bazel builds


stephenduong1004:
The recent sourcemap improvements in commit #23741 introduced a change in how wasm-sourcemap.py is invoked, now using python exec-ing python. This causes environment variables like `PYTHONPATH` to be dropped.

This creates issues when building in isolated environments like Bazel, where third-party dependencies are not in the standard paths and rely on `PYTHONPATH`.

sbc100:
I don't see anything in #23741 that changes the way `wasm-sourcemap.py` is run.   

It looks like we have been doing `PYTHON, "-E", path_from_root('tools/wasm-sourcemap.py'` since at least back in #16736

sbc100:
I uploaded #25373 which should fix the issue. 

Although I still don't know what change broke you here since AFAICT we have always run `wasm-sourcemap.py` with `-E` (ignoring `PYTHONPATH`).

PR #25373, opened 2025-09-23
Convert wasm-sourcemap call from subprocess to library call. NFC

sbc100:
Fixes: #25323

### DIFF
--- a/tools/building.py
+++ b/tools/building.py
@@ -5,6 +5,7 @@
 
 from .toolchain_profiler import ToolchainProfiler
 
+import importlib
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


tools/building.py
```diff
@@ -1145,8 +1146,11 @@ def emit_wasm_source_map(wasm_file, map_file, final_wasm):
   # source file paths must be relative to the location of the map (which is
   # emitted alongside the wasm)
   base_path = os.path.dirname(os.path.abspath(final_wasm))
-  sourcemap_cmd = [sys.executable, '-E', path_from_root('tools/wasm-sourcemap.py'),
-                   wasm_file,
+
+  # TODO(sbc): Rename wasm-sourcemap so it can be imported directly without
+  # importlib.
```
aheejin, line 1151: Why can't we directly import it unless we don't rename it?

sbc100: Becuase of the hyphen in the name, you can't write `import foo-bar` directly in python since it would interpret the `-` as a minus sign.

aheejin (approved):

PR #25374, opened 2025-09-23
[testing] Add option to keep browser open after running a test.

brendandahl:
Handy option to inspect the browser after the test has run.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -62,6 +62,7 @@ logger = logging.getLogger('common')
 EMTEST_BROWSER = None
 EMTEST_BROWSER_AUTO_CONFIG = None
 EMTEST_HEADLESS = None
+EMTEST_KEEP_OPEN = None
 EMTEST_DETECT_TEMPFILE_LEAKS = None
 EMTEST_SAVE_DIR = None
 # generally js engines are equivalent, testing 1 is enough. set this
@@ -2587,6 +2588,10 @@ class BrowserCore(RunnerCore):
 
   @classmethod
   def browser_terminate(cls):
+    if EMTEST_KEEP_OPEN:
+      assert len(cls.browser_procs) == 1, '--keep-open only supports one browser process'
+      cls.browser_procs[0].wait()
+      return
     for proc in cls.browser_procs:
       try:
         proc.terminate()
--- a/test/runner.py
+++ b/test/runner.py
@@ -479,6 +479,8 @@ def parse_args():
                       help='Run browser tests in headless mode.', default=None)
   parser.add_argument('--browser-auto-config', type=bool, default=True,
                       help='Use the default CI browser configuration.')
+  parser.add_argument('--keep-open', action='store_true',
+                      help='Keep the browser open after tests have run.', default=None)
   parser.add_argument('tests', nargs='*')
   parser.add_argument('--failfast', action='store_true')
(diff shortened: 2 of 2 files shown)


test/common.py
```diff
@@ -2587,6 +2588,10 @@ def __init__(self, *args, **kwargs):
 
   @classmethod
   def browser_terminate(cls):
+    if EMTEST_KEEP_OPEN:
+      assert len(cls.browser_procs) == 1, '--keep-open only supports one browser process'
+      cls.browser_procs[0].wait()
```
sbc100, line 2593: But will this not also keep the test running open ? Why not just exit?

brendandahl: At least for chrome the browser will be killed when python exits.

sbc100: Oh I see, maybe add a comment there, something like "# Keep the python process alive until the browser is closed.  This also means that http server will keep runnning"

Alternatively we could exit the python process after orphaning the browser, but then I guess that server would also go down.

test/common.py
```diff
@@ -2587,6 +2588,10 @@ def __init__(self, *args, **kwargs):
 
   @classmethod
   def browser_terminate(cls):
+    if EMTEST_KEEP_OPEN:
+      assert len(cls.browser_procs) == 1, '--keep-open only supports one browser process'
```
juj, line 2592: In the Windows on Firefox case, `len(cls.browser_procs) > 1` even if there is only one test running.

Also, this failure seems like a configuration/setup failure, rather than an after-the fact failure?

I.e. if I run `EMTEST_CORES=128 test/runner browser --keep-open`, then this will still launch all the 128 browsers, and only when they are all being shut down at the very end of the test (or if one test fails), then that Worker will assert fail about this `--keep-open` not being compatible with multithreaded runs?

So maybe a better check might be to move this assert earlier when the test suite is starting, e.g. after the `parallel_testsuite.NUM_CORES` variable has been set, and verify that NUM_CORES==1?

test/common.py
juj, line 2593: how about

```py
for p in cls.browser_procs:
  p.wait()
```
?

test/runner.py
```diff
@@ -479,6 +479,8 @@ def parse_args():
                       help='Run browser tests in headless mode.', default=None)
   parser.add_argument('--browser-auto-config', type=bool, default=True,
                       help='Use the default CI browser configuration.')
+  parser.add_argument('--keep-open', action='store_true',
+                      help='Keep the browser open after tests have run.', default=None)
```
juj, line 483: In the current form, this might not be accurate?

Is the idea to be able to run only a single fixed test with this?

I.e. `test/runner browser.test_my_failing_test --keep-open` ?

Or to be able to run multiple tests, and to be able to fail on first failing one?

Currently `browser.terminate()` will be called if a test times out in the suite? Is that intended to interact with this `--keep-open`?

test/common.py
juj: > At least for chrome the browser will be killed when python exits.

Firefox browser windows stay open after python exits, both on Windows and Linux.

test/common.py
brendandahl: This is kind of a half baked feature. I really just use it for running one test and debugging. I can move the asserts around to be earlier and only support single tester mode.

PR #25375, opened 2025-09-24
Bump min node/firefox/chrome version to ensure globalThis

sbc100:
This change bumps the minimum versions of these engines so that we no longer need to support engines without `globalThis`.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,11 @@ See docs/process.md for more on how version tagging works.
 
 4.0.16 (in development)
 -----------------------
+- The minimum supported versions of node, chrome and firefox were bumped in
+  order remove the globalThis polyfill: (#25375)
+  - node: v10.19.0 -> v12.22.9
+  - chrome: v70 -> v71
+  - firefox: v55 -> v65
 
 4.0.15 - 09/17/25
 -----------------
--- a/src/settings.js
+++ b/src/settings.js
@@ -1873,7 +1873,7 @@ var AUTO_NATIVE_LIBRARIES = true;
 // for Firefox versions older than < majorVersion.
 // Firefox 79 was released on 2020-07-28.
 // MAX_INT (0x7FFFFFFF, or -1) specifies that target is not supported.
-// Minimum supported value is 55 which was released on 2017-08-08 (see
+// Minimum supported value is 65 which was released on 2019-01-29 (see
 // feature_matrix.py)
 // [link]
 var MIN_FIREFOX_VERSION = 79;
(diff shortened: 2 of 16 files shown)


ChangeLog.md
```diff
@@ -20,6 +20,11 @@ See docs/process.md for more on how version tagging works.
 
 4.0.16 (in development)
 -----------------------
+- The minimum supported versions of node, chrome and firefox were bumped in
+  order remove the globalThis polyfill:
+  - node: v10.19.0 -> v12.22.9
+  - chrome: v79 -> v71
```
```suggestion
  - chrome: v70 -> v71
```
(from the rest of the code)

src/settings.js
```diff
@@ -1908,10 +1908,9 @@ var MIN_CHROME_VERSION = 85;
 
 // Specifies minimum node version to target for the generated code.  This is
 // distinct from the minimum version required run the emscripten compiler.
-// This version aligns with the current Ubuuntu TLS 20.04 (Focal).
 // Version is encoded in MMmmVV, e.g. 181401 denotes Node 18.14.01.
-// Minimum supported value is 101900, which was released 2020-02-05 (see
-// feature_matrix.py).
+// Minimum supported value is 122209, which was released 2022-01-11 (see
+// feature_matrix.py). This version aligns with the Ubuntu TLS 22.04 (Jammy).
 var MIN_NODE_VERSION = 160000;
```
kripken, line 1914: The browsers also appear in this file and should be updated.

src/settings.js
```diff
@@ -1908,10 +1908,9 @@ var MIN_CHROME_VERSION = 85;
 
 // Specifies minimum node version to target for the generated code.  This is
 // distinct from the minimum version required run the emscripten compiler.
-// This version aligns with the current Ubuuntu TLS 20.04 (Focal).
 // Version is encoded in MMmmVV, e.g. 181401 denotes Node 18.14.01.
-// Minimum supported value is 101900, which was released 2020-02-05 (see
-// feature_matrix.py).
+// Minimum supported value is 122209, which was released 2022-01-11 (see
+// feature_matrix.py). This version aligns with the Ubuntu TLS 22.04 (Jammy).
```
kripken, line 1913: I guess the previous LTS is EOL, so this seems reasonable.

juj:
I vividly remember authoring a PR that unconditionally used `globalThis`, but then had to revert that due to some problem. I don't remember what it was though.

The caniuse numbers make it seem like it would be ubiquitous, and I do agree that bumping the min browser versions is ok to make this happen; I suppose our test suite should coax out what the issue was before, if one exists.

LGTM.

juj (approved):

sbc100:
> I vividly remember authoring a PR that unconditionally used `globalThis`, but then had to revert that due to some problem. I don't remember what it was though.

Prior to this change we were testing on a version of node that didn't include `globalThis`.. do maybe it was that?  Or maybe there were really old browser that you used to support.  Hopefully no currently targeted browser has this issue.  We can revert if needed.

sbc100:
Maybe as a followup we can see of converting `typeof foo == 'undefined'` everywhere to `!!globalThis.foo` saves on code size.

PR #25383, opened 2025-09-24
[fetch] Support streaming fetch requests.

brendandahl:
Add a new `FETCH_BACKEND` setting that can either be 'xhr' or 'fetch' to use the corresponding DOM API. To keep the emscripten Fetch.js code the same I've implemented a polyfill of XMLHttpRequest using Fetch. To support streaming I wired up the fetch code to use the old onprogress code that old versions of Firefox supported.

Most of the current API is supported with some notable exceptions:
 - synchronous requests
 - overriding the mime type

I also changed a few of the tests to support both sync and async so I could test them with the fetch backend.


inolen:
Just adding a comment here to say I'd appreciate this change.

I had just hacked up Fetch.js to also do this, and then found this PR while searching for issues (and this looks much better than my hack and slash).

RE: overrideMimeType, it's effectively supported in that it doesn't matter - you don't need the XHR to reinterpret whatever mime type the server specified, you treat it as a big binary buffer regardless thanks to the readable stream.

brendandahl:
I've been meaning to come back to this. I was thinking of changing it so that the DOM fetch is only used when streaming is requested. That way I don't have to try and replicate all the old XHR behavior when streaming isn't enabled.

brendandahl:
I've changed how it works from original PR, now when FETCH_STREAMING is enabled we'll only use the DOM fetch API when a streaming data is requested. This way we can worry less about fully polyfilling XHR and hopefully have less breakage.

src/settings.js
```diff
@@ -1836,6 +1836,21 @@ var FETCH_DEBUG = false;
 // [link]
 var FETCH = false;
 
+// Enables streaming fetched data when the fetch attribute
+// EMSCRIPTEN_FETCH_STREAM_DATA is used. For streaming requests, the DOM fetch
```
sbc100: Should "Fetch API" use a capital F?   as in https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch?

src/Fetch.js
```diff
@@ -276,8 +287,8 @@ function fetchXHR(fetch, onsuccess, onerror, onprogress, onreadystatechange) {
   xhr.open(requestMethod, url_, !fetchAttrSynchronous, userNameStr, passwordStr);
   if (!fetchAttrSynchronous) xhr.timeout = timeoutMsecs; // XHR timeout field is only accessible in async XHRs, and must be set after .open() but before .send().
   xhr.url_ = url_; // Save the url for debugging purposes (and for comparing to the responseURL that server side advertised)
-#if ASSERTIONS
-  assert(!fetchAttrStreamData, 'streaming uses moz-chunked-arraybuffer which is no longer supported; TODO: rewrite using fetch()');
+#if ASSERTIONS && !FETCH_STREAMING
+  assert(!fetchAttrStreamData, 'streaming is only supported when using the fetch backend');
```
sbc100: How about `streaming is only supported when FETCH_STREAMING is enabled`?  (i.e. mention the name of the setting in the error message)

src/lib/libfetch.js
```diff
@@ -38,7 +38,245 @@ var LibraryFetch = {
     '$fetchLoadCachedData',
     '$fetchDeleteCachedData',
 #endif
-  ]
+#if FETCH_STREAMING
+    '$FetchXHR',
+#endif
+  ],
```
sbc100, line 47: Can you wrap this whole definition in `if FETCH_STREAMING`?

src/lib/libfetch.js
```diff
@@ -38,7 +38,245 @@ var LibraryFetch = {
...
+    /**
+     * Aborts the request if it has already been sent.
+     */
+    abort() {
+      this._aborted = true;
+      this.status = 0;
+      this._changeReadyState(4); // 4: DONE
+      this._abortController?.abort();
+    }
+  }
```
sbc100: What do you think about putting this whole class in `Fetch.js` (you can still wrap it in `#if FETCH_STREAMING`).   I think the idea with this `libfetch.js` seems to be that it just exposes the implementation which lives in `Fetch.js`.

I don't feel strongly about this though.. what do you think?

test/decorators.py
```diff
@@ -404,6 +404,20 @@ def metafunc(self, with_wasm64, *args, **kwargs):
   return metafunc
 
 
+def also_with_fetch_backend(f):
```
sbc100: How about `also_with_fetch_streaming` ?  Or `also_with_streaming_fetch`?

test/decorators.py
```diff
@@ -404,6 +404,20 @@ def metafunc(self, with_wasm64, *args, **kwargs):
...
+  assert callable(f)
+
+  @wraps(f)
+  def metafunc(self, with_fetch, *args, **kwargs):
+    if with_fetch:
+      self.set_setting('FETCH_STREAMING', '2')
```
sbc100, line 413: Could you do `self.cflags.append('SKIP_SYNC_TESTS')` here maybe and save some complexity elsewhere?   I'm not sure.. maybe it not a shared pattern?

sbc100 (commented):
Very nice to see this working out!

src/Fetch.js
```diff
@@ -4,6 +4,244 @@
  * SPDX-License-Identifier: MIT
  */
 
+#if FETCH_STREAMING
+/**
+ * A class that mimics the XMLHttpRequest API using the modern Fetch API.
+ * This implementation is specifically tailored to only handle 'arraybuffer'
+ * responses.
+ */
+var FetchXHR = class {
```
sbc100, line 14: Why not just `class FetchXHR {` ?  Does it not work with the `#include` mechanism perhaps?  Maybe worth a comment since it looks a little odd

src/Fetch.js
```diff
@@ -4,6 +4,244 @@
...
+    this.statusText = '';
+
+    // --- Internal Properties ---
+    this._method = '';
+    this._url = '';
+    this._headers = {};
+    this._abortController = null;
+    this._aborted = false;
+    this._responseHeaders = null;
+  }
```
sbc100: Can you just make these fields of the class and remove the ctor completely?  (or is there some reason we can't do that?.. again maybe worth a comment.

src/Fetch.js
brendandahl: Yeah, it's broken. It gets turned into `class FetchXHR FetchXHR {`. I can add comment.

sbc100: Can you open a bug to get that fixed and link to it here?

src/Fetch.js
brendandahl: It was a while ago, but I thought I was running into an issue when using a class fields in a library, but it seems to work fine now.

inolen:
@brendandahl quick question regarding this after just rebasing and testing with it.

When using EMSCRIPTEN_FETCH_STREAM_DATA, shouldn't we be avoiding concatenating the chunks? As is, you do get the onprogress events, but you still end up concatenating each chunk and making a heap allocation for the entire file (which at least in my case, is what i'm trying to avoid).

It feels appropriate to provide each individual chunk to onprogress, and then set data / numBytes / byteOffset all to 0 for the final onload. This is also what the original moz-chunked-arraybuffer support did -

> moz-chunked-arraybuffer

> A Firefox-only value which instructs XMLHttpRequest to deliver [ArrayBuffer](https://udn.realityripple.com/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) objects containing chunks of the incoming data. Accessing response during a progress event returns the data received so far. Outside the progress event handler, the value of response is always null. You shouldn't use this non-standard (and, as of Firefox 68, entirely removed) API; instead, consider using [the Fetch API with readable streams](https://udn.realityripple.com/docs/Web/API/Streams_API/Using_readable_streams#Consuming_a_fetch_as_a_stream), which offers a standard alternative to accessing the response in a streaming fashion.

Edit: Sorry, looks like I was a bit off base here. The onload doesn't actually end up making the final allocation, because the last condition fails in `if (xhr.response && fetchAttrLoadToMemory && {{{ makeGetValue('fetch', C_STRUCTS.emscripten_fetch_t.data, '*') }}} === 0) {` (ptr is non-NULL when using onprogress). However, this looks problematic -
1. the pointer allocated by onprogress is lost here as `{{{ makeSetValue('fetch', C_STRUCTS.emscripten_fetch_t.data, 'ptr', '*') }}}` isn't inside the above conditional.
2. chunks are still concatenated into a new Uint8Array unnecessarily

PR #25384, opened 2025-09-24
[testing] Fix disable browser auto config.

brendandahl:
Can now use `--no-browser-auto-config` or the env variable.

### DIFF
--- a/test/runner.py
+++ b/test/runner.py
@@ -495,7 +495,7 @@ def parse_args():
                       help='Command to launch web browser in which to run browser tests.')
   parser.add_argument('--headless', action='store_true',
                       help='Run browser tests in headless mode.', default=None)
-  parser.add_argument('--browser-auto-config', type=bool, default=True,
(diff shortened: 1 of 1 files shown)


test/runner.py
```diff
@@ -477,7 +477,7 @@ def parse_args():
                       help='Command to launch web browser in which to run browser tests.')
   parser.add_argument('--headless', action='store_true',
                       help='Run browser tests in headless mode.', default=None)
-  parser.add_argument('--browser-auto-config', type=bool, default=True,
+  parser.add_argument('--browser-auto-config', action=argparse.BooleanOptionalAction, default=None,
```
sbc100, line 498: Hmm.. i've never seen that before.   Does it magically provide the `--no-` variant?

sbc100 (approved):

test/runner.py
brendandahl: Yeah, just learned about it the other day.

brendandahl: Err...I guess it's too new. Not available until 3.9.

PR #25386, opened 2025-09-25
[empath-split] Add --print-sources option

aheejin:
The 'paths' file should contain paths that match those of `sources` field in the source map, which can be hard to figure out because they can be relative to some build directory.

This adds `--print-sources` option, which prints the contents of the `sources` field. You can just manually open source map files to obtain the same information, but they are usually hard to read without any newlines. You also can use a general JSON prettyprinting tool, but I think it wouldn't hurt to provide the same info handy here.


tools/empath-split.py
```diff
@@ -59,23 +74,35 @@ def parse_args():
 enabling/disabling options. Run 'wasm-split -h' for the list of options. But you
 should NOT add --manifest, because this will be generated from this script.
 """)
-  parser.add_argument('wasm', help='Path to the input wasm file')
-  parser.add_argument('paths_file', help='Path to the input file containing paths')
+  parser.add_argument('wasm', nargs='?', help='Path to the input wasm file')
+  parser.add_argument('paths_file', nargs='?', help='Path to the input file containing paths')
```
aheejin, line 78: These became optional because when you use `--print-sources` these are not mandatory. (We manually check for their existence below when `--print-sources` is not given)

aheejin:
> Makes sense. maybe another useful diagnostic functionality would be when actually doing a split, to print how many functions (and bytes of functions?) matched each split-module in the input patterns, which might tell you at a glance whether things worked approximately as expected.

When `-v` is given we already print the {path: files} mapping. I've added the # of functions too: https://github.com/emscripten-core/emscripten/pull/25386/commits/61eeb49b512ceba817ed3072eae5c69689e7ba6f

PR #25388, opened 2025-09-25
Speed up test_windows_long_link_response_file. NFC

sbc100:
Reducing the number of files from 1000 to 300 makes the test run in 20 seconds rather than 60 seconds on my machine.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -12973,7 +12973,7 @@ exec "$@"
     self.do_runf('main.c', 'Top level code\nhello from pre-run\njs_func called\n')
 
   # On Windows maximum command line length is 32767 characters. Create such a long build line by linking together
-  # several .o files to test that emcc internally uses response files properly when calling llvm-nm and wasm-ld.
+  # several .o files to test that emcc internally uses response files properly when calling wasm-ld.
   @is_slow_test
   def test_windows_long_link_response_file(self):
     decls = ''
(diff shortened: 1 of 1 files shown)


test/test_other.py
```diff
@@ -12988,16 +12988,17 @@ def create_o(name, i):
       decls += 'int %s();' % name
       calls += 'value += %s();' % name
 
-    count = 1000
+    count = 300
     for i in range(count):
       name = 'a' + str(i)
-      for _ in range(5):
-        name += name
+      name = name * 32
```
juj, line 12994: Would using `count = 150`, `name * 64` reduce the test time even further?

sbc100: I suppose so, that internal names used for object files do happen to always be at least as long as the source file names, so the wasm-ld command line will currently be bounded by the sum of the lengths of the names.

However, I'm not sure I want to mess with the lengths of individual filenames in case we start running into filename length limits. 

This PR as it stands does change the lengths of individual filenames.

kripken:
lgtm, but why is it 20 seconds even after? isn't it just linking a bunch of small files?

sbc100:
> lgtm, but why is it 20 seconds even after? isn't it just linking a bunch of small files?

It just takes 20 seconds to run clang 300 times (once for each source file).  When the clang driver (or the emcc driver) runs clang internally it builds each source file in serial, not parallel.   e.g. `clang 1.c 2.c 3c` will run clang internally 3 times in sequence.   Its not a common pattern though and probably not something that we (or clang want to parallelize) since the convention is normally to do the parallelim in the build system (e.g. make/ninja) instead.

sbc100:
```"
$ time for i in `seq 1 300`; do ../llvm-build/bin/clang --target=wasm32-emscripten -c test/hello_world.c --sysroot=cache/sysroot; done
real	0m21.201s
```

sbc100:
Interestingly by local build of clang is way worse then system clang here:

```
$ time for i in `seq 1 300`; do ../llvm-build/bin/clang -c test/hello_world.c; done

real	0m22.072s
user	0m11.340s
sys	0m10.694s
$ time for i in `seq 1 300`; do clang -c test/hello_world.c; done

real	0m11.242s
user	0m4.394s
sys	0m6.877s
```

I guess because of the flags I'm using to build clang? Maybe `-DLLVM_ENABLE_ASSERTIONS=ON`?

kripken:
Interesting.. slower than I'd thought, but I guess it is parsing lots of headers and stuff...

kripken:
I also see ~10 seconds locally. Assertions make sense as a big slowdown.

sbc100:
OK to land this, even though we could make go further?

juj:
Yes definitely.

Issue #25387, opened 2025-09-25
emcc: recursive response files are ignored


ktock:
Emcc fails to compile a program if a respose file contains a reference to another respose file.

### Version of emscripten/emsdk:
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.15 (09f52557f0d48b65b8c724853ed8f4e8bf80e669)
clang version 22.0.0git (https:/github.com/llvm/llvm-project 3388d40684742e950b3c5d1d2dafe5a40695cfc1)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: /emsdk/upstream/bin
```

### Failing command line in full:

hello.c

```c
#include <stdio.h>

int main() {
  printf("hello\n");
  return 0;
}
```

file1 (response file)

```
@file2
```

file2 (response file)

```
-o hello hello.c
```

`file1` points to another response file `file2`.

#### Expected behaviour

Both of `emcc @file1` and `emcc @file2` succeeds.

#### Actual behaviour

- `emcc @file1` : failed with an error `emcc: error: @file2: No such file or directory ("@file2" was expected to be an input file, based on the commandline arguments provided)`
- `emcc @file2` : succeeded

### Full link command and output with `-v` appended:

N/A

### Additional information

GCC supports recursive response files, according to [the man page](https://man7.org/linux/man-pages/man1/gcc.1.html).

```
The file may itself contain additional @file options; any such options will be processed recursively.
```

sbc100:
Yes, it looks like clang also supports this.

PR #25390, opened 2025-09-25
Add support for recursive response files

sbc100:
Fixes: #25387

### DIFF
--- a/tools/response_file.py
+++ b/tools/response_file.py
@@ -13,14 +13,9 @@ from .utils import WINDOWS
 DEBUG = int(os.environ.get('EMCC_DEBUG', '0'))
 
 
-def create_response_file(args, directory):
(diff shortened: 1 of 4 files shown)


tools/response_file.py
```diff
@@ -63,12 +63,19 @@ def read_response_file(response_filename):
...
-  if not os.path.exists(response_filename):
-    raise OSError("response file not found: %s" % response_filename)
+  if arg.startswith('@'):
+    response_filename = arg[1:]
+  elif arg.startswith('-Wl,@'):
+    response_filename = arg[5:]
+  else:
+    response_filename = None
+
+  # Is the argument is not a response file, or of the file does not exist
```
kripken:
```suggestion
  # Is the argument is not a response file, or if the file does not exist
```

PR #25391, opened 2025-09-25
[CI] Remove `--experimental-wasm-stack-switching` from chrome flags.

sbc100:
It looks like this flag was removed, causing the flags after it to be ignored:

```
Error: unrecognized flag --experimental-wasm-stack-switching
The remaining arguments were ignored: --experimental-wasm-type-reflection --experimental-wasm-rab-integration
```


sbc100:
@fgm

sbc100:
@fgmccabe FYI

fgm:
@sbc100 Probably not me you meant to tag, I'm afraid.

Issue #25343, opened 2025-09-22
test_emdawnwebgpu_link_test is flaky


sbc100:
This error showed up on the emscripten-releases bots recently:

```
system_libs:INFO: compiled 1 inputs in 0.54s
cache:INFO:  - ok
cache:INFO: deleting cached file: /Volumes/Work/s/w/ir/x/w/install/emscripten/cache/sysroot/lib/wasm32-emscripten/libemdawnwebgpu-00c5e8e9-O0.a
ports:INFO: retrieving port: mpg123 from https://www.mpg123.de/download/mpg123-1.26.2.tar.bz2
test_embind_tsgen_val (test_other.other) ... ok (7.87s)
wasm-ld: error: cannot open /Volumes/Work/s/w/ir/x/w/install/emscripten/cache/sysroot/lib/wasm32-emscripten/libemdawnwebgpu-00c5e8e9-O0.a: No such file or directory
[...]

PR #25397, opened 2025-09-26
Roll Emdawnwebgpu and repoint USE_WEBGPU tests at Emdawnwebgpu

kainino0x:
In preparation to remove USE_WEBGPU entirely (#24265). Ideally these tests would live in Dawn, but we don't have a way to automate tests in the browser yet, so for the moment we'll keep them in Emscripten.

DO NOT SUBMIT: Needs Emdawnwebgpu roll in order to pass


kainino0x:
@lokokung PTAL

test/webgpu_required_limits.c
```diff
@@ -3,12 +3,10 @@
 #include <stdio.h>
 #include <webgpu/webgpu.h>
```
lokokung, line 4: nit. Why not use the C++ bindings in this test file as well?

kainino0x: IDK, I didn't write it, but I figure it doesn't hurt to have a C test.

BTW, added one commit to improve the test a bit since the async stuff is hairy.

test/test_browser.py
```diff
@@ -4450,24 +4450,16 @@ def test_webgl_simple_extensions(self, webgl_version, simple_enable_extensions):
...
-  @requires_webgpu
-  def test_webgpu_basic_rendering_pthreads(self):
-    self.btest_exit('webgpu_basic_rendering.cpp', cflags=['-Wno-error=deprecated', '-sUSE_WEBGPU', '-pthread', '-sOFFSCREENCANVAS_SUPPORT'])
-
-  def test_webgpu_get_device(self):
-    self.btest_exit('webgpu_get_device.cpp', cflags=['-Wno-error=deprecated', '-sUSE_WEBGPU', '-sASSERTIONS', '--closure=1'])
-
-  def test_webgpu_get_device_pthreads(self):
-    self.btest_exit('webgpu_get_device.cpp', cflags=['-Wno-error=deprecated', '-sUSE_WEBGPU', '-pthread'])
+    self.set_setting('NO_DEFAULT_TO_CXX', 0)  # emdawnwebgpu uses C++ internally
```
sbc100: What to do about this long term?

Is this some fundamental difference between the old and new implementation?     The only one manged to avoid depending on libc++ library functions but the new one did not?    Are folks who use this library simply bound to link with C++?

tools/ports/emdawnwebgpu.py
```diff
@@ -132,17 +144,37 @@
...
-URL = 'https://dawn.googlesource.com/dawn/+/80062b708e44aa4d8c48e555ed0cc801396069f6/src/emdawnwebgpu/pkg/README.md'
+URL = 'https://dawn.googlesource.com/dawn/+/faa7054b5b65c3ce3774151952a68aa7668aa20b/src/emdawnwebgpu/pkg/README.md'
+
+
+# Emscripten <4.0.10 won't notice EXTERNAL_PORT and will try to use this.
+def get(ports, settings, shared):
+    raise Exception('Remote ports require Emscripten 4.0.10+.')
+
+
+def clear(ports, settings, shared):
```
sbc100, line 180: Can / should we update the port as a separate PR that we can land first?

test/test_browser.py
kainino0x: Great question. I'm pretty sure your guess is correct - the old one only used `#include <array>` which probably didn't require any linking, but now we use a [whole bunch of stuff](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/third_party/emdawnwebgpu/pkg/webgpu/src/webgpu.cpp), in particular `<atomic>` and `<mutex>`, at least one of which I believe requires linking.

Technically we could remove those dependencies in single-threaded builds, but it's much easier for us to just rely on the compiler to handle those optimizations. (Though, mutexes might actually not be required until we have native multithreading in the browser's WebGPU but I'm not sure off hand. Atomics are used for refcounts which are needed even today.)

sbc100: Do `<atomic>` and `<mutex>` end up requiring library symbols at link time?   If not maybe it could be a goal to avoid link-time libc++ symbols?   Might be worth a try to see how hard it would be.

tools/ports/emdawnwebgpu.py
kainino0x: I'm fine either way, just did it in one PR because I needed the CI checks to run. Opened #25425

test/test_browser.py
kainino0x: Oh right, I suppose I can check. I totally forgot what the original error message was that led me to add this.

`wasm-ld: error: /Users/kainino/src/emscripten/cache/sysroot/lib/wasm32-emscripten/libemdawnwebgpu-ac372172-O0.a(webgpu.cpp.o): undefined symbol: operator new(unsigned long)`
`wasm-ld: error: /Users/kainino/src/emscripten/cache/sysroot/lib/wasm32-emscripten/libemdawnwebgpu-ac372172-O0.a(webgpu.cpp.o): undefined symbol: operator delete(void*, unsigned long)`
... many copies of those lines

Those we certainly can't easily get rid of...

sbc100: I think maybe you can?   Maybe either by defining your own local versions of new and delete, or switching the placement new/delete?

kainino0x: That's probably possible, I (mostly) figured out how to overload the new/delete operators - but I read the output properly and realized it said I should pass `-error-limit=0`, so figured out how to do that (`-Wl,--error-limit=0`) and here's the full (deduplicated) list:
```
undefined symbol: __cxa_pure_virtual
undefined symbol: operator delete(void*, unsigned long, std::align_val_t)
undefined symbol: operator delete(void*, unsigned long)
undefined symbol: operator new(unsigned long, std::align_val_t)
undefined symbol: operator new(unsigned long)
undefined symbol: std::__2::__libcpp_verbose_abort(char const*, ...)
undefined symbol: std::__2::__next_prime(unsigned long)
undefined symbol: std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char>>::__assign_external(char const*)
undefined symbol: std::__2::basic_string<char, std::__2::char_traits<char>, std::__2::allocator<char>>::~basic_string()
undefined symbol: std::__2::mutex::~mutex()
undefined symbol: std::__2::mutex::lock()
undefined symbol: std::__2::mutex::unlock()
undefined symbol: vtable for __cxxabiv1::__class_type_info
undefined symbol: vtable for __cxxabiv1::__si_class_type_info
undefined symbol: vtable for __cxxabiv1::__vmi_class_type_info
```

(TBH not sure where some of these things came from (like `__class_type_info`), there must be something in the STL.) Regardless I don't think it is really practical to avoid using libc++.

sbc100: I see, yes if you are using STL types then you will need libc++ linked in.

The question is, is it intended to be able to use this port from C programs?  I suppose the answer is yes, in which case we need to find a way to add `-lc++` when using the port.  I believe this is the first port with this requirement.  

I imagine the right place to perform this injection would be `process_dependencies` but I'm not sure we have a mechanism to inject link flags like this. 

Oh way.. I think you can do `settings.LINK_AS_CXX = true` in the function!  Can you try that?

kainino0x: Oh great idea, that works! It overrides `NO_DEFAULT_TO_CXX` for the rest of the program... but certainly easier to use. Will send a Dawn CL.

kainino0x: Resolved in https://dawn-review.googlesource.com/c/dawn/+/264434 with the next Emdawnwebgpu->Emscripten roll (upcoming).

test/test_other.py
```diff
@@ -9652,13 +9652,15 @@ def test_closure_full_js_library(self, args):
 
   @also_with_wasm64
```
sbc100, line 9658: Could you maybe mark this tests `@crossplatform`?  Might be a good idea anyway, but also will mean it gets run in our CI (the linux bot uses a frozen cache) and mac and windows only run the crossplatform tests.

kainino0x: Done

test/test_browser.py
```diff
@@ -4458,26 +4458,16 @@ def test_webgl_simple_extensions(self, webgl_version, simple_enable_extensions):
...
-
-  @requires_webgpu
-  def test_webgpu_get_device(self):
-    self.btest_exit('webgpu_get_device.cpp', cflags=['-Wno-error=deprecated', '-sUSE_WEBGPU', '-sASSERTIONS', '--closure=1'])
-
-  @requires_webgpu
-  def test_webgpu_get_device_pthreads(self):
-    self.btest_exit('webgpu_get_device.cpp', cflags=['-Wno-error=deprecated', '-sUSE_WEBGPU', '-pthread'])
+    self.set_setting('DEFAULT_TO_CXX')  # emdawnwebgpu uses C++ internally
+    self.btest_exit('webgpu_required_limits.c', cflags=['--use-port=emdawnwebgpu', '-sEXIT_RUNTIME'])
```
sbc100: Why is EXIT_RUNTIME now needed here?

test/test_other.py
```diff
@@ -2588,7 +2588,7 @@ def test_contrib_ports(self):
   def test_remote_ports(self):
     self.emcc(test_file('hello_world.c'), ['--use-port=emdawnwebgpu'])
 
-  @crossplatform
+  @crossplatform  # Note this only tests on platforms where FROZEN_CACHE isn't set.
```
sbc100: I don't think this comment is needed since you can see `test doesn't work with frozen cache` just below.

test/test_other.py
```diff
@@ -9650,15 +9650,18 @@ def test_closure_full_js_library(self, args):
...
+      # TODO(crbug.com/446944885): Make Emdawnwebgpu work with FROZEN_CACHE if possible.
+      self.skipTest("test doesn't work with frozen cache")
+    self.set_setting('DEFAULT_TO_CXX')  # emdawnwebgpu uses C++ internally
     self.build('hello_world.c', cflags=[
       '--closure=1',
       '-Werror=closure',
-      '-Wno-error=deprecated',
       '-sINCLUDE_FULL_LIBRARY',
-      '-sUSE_WEBGPU',
+      '--use-port=emdawnwebgpu',
```
sbc100, line 9668: Do we have any remaining tests for `-sUSE_WEBGPU`?

test/test_other.py
```diff
@@ -12298,24 +12301,19 @@ def test_standalone_syscalls(self):
...
       self.assertContained(expected, self.run_js('test.wasm', engine))
 
-  @parameterized({
-    '': ([],),
-    'assertions': (['-sASSERTIONS'],),
-    'closure': (['-sASSERTIONS', '--closure=1'],),
-    'dylink': (['-sMAIN_MODULE'],),
-  })
-  def test_webgpu_compiletest(self, args):
-    self.run_process([EMXX, test_file('webgpu_jsvalstore.cpp'), '-Wno-error=deprecated', '-sUSE_WEBGPU', '-sASYNCIFY'] + args)
```
sbc100, line 12303: Why is this test no longer useful?

test/test_other.py
```diff
@@ -12298,24 +12301,19 @@ def test_standalone_syscalls(self):
...
   @parameterized({
     '': ([],),
     'closure': (['--closure=1', '-Werror=closure'],),
     'closure_assertions': (['--closure=1', '-Werror=closure', '-sASSERTIONS'],),
   })
   def test_emdawnwebgpu_link_test(self, args):
-    self.run_process([EMXX, test_file('test_emdawnwebgpu_link_test.cpp'), '--use-port=emdawnwebgpu', '-sASYNCIFY'] + args)
+    if config.FROZEN_CACHE:
+      # TODO(crbug.com/446944885): Make Emdawnwebgpu work with FROZEN_CACHE if possible.
+      self.skipTest("test doesn't work with frozen cache")
```
sbc100, line 12310: Why is this `skipTest` needed now but not before this PR?

test/webgpu_required_limits.c
```diff
@@ -46,43 +43,50 @@ void assertLimitsCompatible(WGPULimits required_limits,
...
 
-  WGPUDeviceDescriptor device_desc = {0,};
-  device_desc.requiredFeatureCount = 0;
-  device_desc.requiredLimits = &device_required_limits;
-  wgpuAdapterRequestDevice(adapter, &device_desc, on_device_request_ended, NULL);
+  wgpuAdapterRelease(adapter);
 }
 
-int main() {
+int main(void) {
```
sbc100: Are these test changes coming from upstream or is emscripten (here) the source of truth for them?

(Wondering where this 'void' comes from because I don't think it should be needed by the compiler)

sbc100 (commented):
I'm not really qualified to read the `.c`/`.cpp` changes to the test files, so I'll have to take your word for it mostly that they good.

@juj might have some opinions?

test/test_browser.py
kainino0x: You're right, it's not, because I have an `exit(0)` which I was originally planning to take out. (With `EXIT_RUNTIME`, I could take it out, but `webgpu_basic_rendering` now has lots of testing for runtime keepalive.)

test/webgpu_required_limits.c
juj: `int main(void) {}` in a C function definition does nothing, it is the same as `int main() {}`.

Only in C declarations does `int main();` mean `int main(...);`, but not in definitions.

test/test_other.py
kainino0x: No, but I'm planning to remove it immediately after this (#25398).

test/test_other.py
kainino0x: I think it's because previously the cache was pre-populated with `-sUSE_WEBGPU` but we don't prepopulate it with `--use-port=emdawnwebgpu` (which would have to be done with several different flags too).

I can try to pre-populate the cache with the appropriate Emdawnwebgpu objects if you think that's appropriate.

test/webgpu_required_limits.c
kainino0x: Ah, I enabled `-Weverything` at some point because I was trying to track down a bug in some of these tests. No longer needed.

test/test_other.py
kainino0x: The whole jsvalstore thing has been removed in Emdawnwebgpu. We have another mechanism but it would need entirely new tests.

test/webgpu_basic_rendering.cpp
```diff
@@ -16,7 +18,34 @@
...
+};
+
+void RegisterCheckScopesAtExit() {
+    atexit([](){
+        // Check we don't exit before the tests are done.
+        // (Make sure there's a keepalive for everything the test has scopes for.)
+        // Build with -sEXIT_RUNTIME to trace keepalives.
+        assert(sScopeCount == 0);
+        // Overwrite the old return code and exit now that everything is done.
+        exit(0);
```
juj: I think it's undefined behavior to call `exit()` from inside an `atexit()` handler. (ISO/IEC 9899:n2347.pdf, §7.22.4.4/2: "If a program calls the exit function more than once, the behavior is undefined.")

(Calling exit() triggers the call of atexit() handlers)

test/test_other.py
kainino0x: https://github.com/emscripten-core/emscripten/blob/3b22cd69b6df702096361f8512d1286cb27b6f6d/embuilder.py#L103-L104
I think it comes from this? I don't think it'll be that easy to add Emdawnwebgpu here though.

kainino0x: Oh, it would be added via this https://github.com/emscripten-core/emscripten/blob/3b22cd69b6df702096361f8512d1286cb27b6f6d/embuilder.py#L139

juj:
> @juj might have some opinions?

LGTM with the test changes - looks like they're effectively refactoring callbacks to lambdas.

kainino0x:
> LGTM with the test changes - looks like they're effectively refactoring callbacks to lambdas.

Mainly yes. But also taking advantage of keepalives to let a bunch of async stuff finish before the test completes (without using asyncify).

test/test_other.py
sbc100: But the old code and the new code here both use `--use-port=emdawnwebgpu`.

kainino0x: Oh, sorry, I missed which test this was. I actually have no idea. If I remove emdawnwebgpu from my local cache and run the old test (on `main`), it still fails with `FROZEN_CACHE is set, but cache file is missing`.

Could the `@flaky` be suppressing it?

kainino0x: Oh, CI has an explicit separate Embuilder step. Let me figure out how to test that locally and then see where I need to skip tests.

kainino0x: Okay, I figured out which configurations don't get precached and skipped only those.

kainino0x: ... Although I did not figure out exactly why it was passing before and isn't now. `@also_with_wasm64` isn't newly-added.

test/test_browser.py
```diff
@@ -4454,30 +4455,24 @@ def test_webgl_simple_extensions(self, webgl_version, simple_enable_extensions):
...
+    'main_module': (False, ['-sMAIN_MODULE=1']),
+    # Not precached with SHARED_MEMORY
+    'pthreads': (False, ['-pthread', '-sOFFSCREENCANVAS_SUPPORT']),
   })
   @requires_webgpu
-  def test_webgpu_basic_rendering(self, args):
-    self.btest_exit('webgpu_basic_rendering.cpp', cflags=['-Wno-error=deprecated', '-sUSE_WEBGPU'] + args)
+  def test_webgpu_basic_rendering(self, assume_precached, args):
+    if config.FROZEN_CACHE and not assume_precached:
+      self.skipTest("test doesn't work with frozen cache")
```
sbc100: This is not needed because we don't currently run the browser tests with a frozen cache.

See `.circleci` where we set `EM_FROZEN_CACHE=""` for the browser tests.

test/test_other.py
```diff
@@ -2586,6 +2586,7 @@ def test_contrib_ports(self):
 
   @requires_network
   def test_remote_ports(self):
+    self.set_setting('DEFAULT_TO_CXX')  # emdawnwebgpu uses C++ internally
     self.emcc(test_file('hello_world.c'), ['--use-port=emdawnwebgpu'])
```
sbc100: How about just using `hello_world.cpp` here instead?  (with a comment as to why).

test/test_other.py
```diff
@@ -9652,13 +9653,15 @@ def test_closure_full_js_library(self, args):
 
   @also_with_wasm64
   def test_closure_webgpu(self):
-    # This test can be removed if USE_WEBGPU is later included in INCLUDE_FULL_LIBRARY.
+    if config.FROZEN_CACHE and self.get_setting('MEMORY64'):
+      # CI configuration doesn't run `embuilder` with wasm64 on ports
+      self.skipTest("test doesn't work with frozen cache")
+    self.set_setting('DEFAULT_TO_CXX')  # emdawnwebgpu uses C++ internally
     self.build('hello_world.c', cflags=[
```
sbc100: Same here regarding `hello_world.cpp`

kainino0x:
@sbc100 bump!

sbc100 (approved):

kainino0x:
I guess I should have rebased and rerun CI before landing... something broke on main. Seems to be cache related again. Looking into it.

sbc100:
Yes, it looks kind of odd yes.  I see `cache:INFO: deleting cached file: C:\Users\circleci\cache\sysroot\lib\wasm32-emscripten\libemdawnwebgpu-50506ac3-O0.a` after it was built...

kainino0x:
Good catch. Unfortunately no luck so far reproducing this locally on Windows. I'm not entirely clear from the logs if there's an interaction between different tests or not.

kainino0x:
Ah, just realized this is probably a flake. I think the first failure I saw on https://github.com/emscripten-core/emscripten/pull/25398 was the same, but after rerunning it, I got something else:
```
FAIL [14.583s]: test_embuilder_wildcards (test_sanity.sanity)
  File "/root/project/test/test_sanity.py", line 766, in test_embuilder_wildcards
    self.assertEqual(len(glob.glob(glob_match)), 1)
AssertionError: 30 != 1
```
and this test passed.

sbc100:
That assertions looks very odd.  Line 766 of test_sanity.py reads `self.assertGreater(len(glob.glob(glob_match)), 3)`, not `self.assertEqual(len(glob.glob(glob_match)), 1)`.

kainino0x:
Ohh wait, this flake was the existing one from #25343. I forgot about that and didn't read it closely so I didn't realize it was doing the same thing.

kainino0x:
The `test_embuilder_wildcards` flake I don't have any clue about.

kainino0x:
The next commit on main (0b127fffef51ed368d983b801b436f9134fd30ad) passed CI. (I also retriggered CI on my commit so it might show as passing too once it's done)

PR #25398, opened 2025-09-26
Remove -sUSE_WEBGPU in favor of --use-port=emdawnwebgpu

### LINKED ISSUE #24265: Deprecate and remove -sUSE_WEBGPU

kainino0x:
DO NOT SUBMIT: Should probably land after #25397

Fixes #24265


tools/settings.py
```diff
@@ -122,13 +122,16 @@
...
     'LEGALIZE_JS_FFI': 'to disable JS type legalization use `-sWASM_BIGINT` or `-sSTANDALONE_WASM`',
     'ASYNCIFY_EXPORTS': 'please use JSPI_EXPORTS instead',
-    'USE_WEBGPU': 'please try migrating to --use-port=emdawnwebgpu, which implements a newer, incompatible version of webgpu.h (see tools/ports/emdawnwebgpu.py for more info)',
     'LINKABLE': 'under consideration for removal (https://github.com/emscripten-core/emscripten/issues/25262)',
     'RELOCATABLE': ' under consideration for removal (https://github.com/emscripten-core/emscripten/issues/25262)',
     'PROXY_TO_WORKER': 'under consideration for removal (See https://github.com/emscripten-core/emscripten/issues/25440)',
     'PROXY_TO_WORKER_FILENAME': 'under consideration for removal (See https://github.com/emscripten-core/emscripten/issues/25440)',
 }
 
+REMOVED_SETTINGS = {
```
kainino0x: I couldn't find if you already had a mechanism for this, so just introduced something simple.

sbc100: You can add it to `LEGACY_SETTINGS` in `settings.js` I think,   If you have 0 as the only valid value then anyone setting it to 1 will see the error.

kainino0x: Nice, thanks!

sbc100 (approved):
Nice!

I'm currently trying to do an emscripten release so its likely this changelog entry will need to be moved if I succeed, but that is easy enough.

kainino0x:
Should I hold on merging this then?

sbc100:
Maybe we can wait an hour or so until the new release goes out yes

PR #25406, opened 2025-09-28
Fix test browser.test_preload_file to pass in old browsers

juj:
Fix test browser.test_preload_file to pass in old browsers that do not support IndexedDB 3.0 API.


juj:
This is another PR that will help me get a Firefox test run green.

PR #25407, opened 2025-09-28
Fix WasmFS Fetch backend

juj:
Fix WasmFS Fetch backend to work in Firefox < 128, Chrome < 132, Safari < 18 and Node.js < 22.3

### DIFF
--- a/src/lib/libwasmfs_fetch.js
+++ b/src/lib/libwasmfs_fetch.js
@@ -95,7 +95,13 @@ addToLibrary({
       if (!response.ok) {
         throw response;
       }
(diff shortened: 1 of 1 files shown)


juj:
Ping - maybe this could be easy to land, it would allow me to continue running my test harness.

sbc100 (commented):
lgtm.

(I wonder if we can can use normal member access for these?.. I guess it depends in closure knows about these methods?)

sbc100 (approved):

juj:
yeah if Closure is up to date to understand bytes() and arrayBuffer(), then it'd work.

PR #25408, opened 2025-09-28
-sEXPORT_ES6 and multithreading too old browsers.

juj:
~~Abort build if attempting to target -sEXPORT_ES6 and multithreading with too old target browser versions.~~

Add ES6 Module Workers as feature in the feature matrix.

### DIFF
--- a/tools/feature_matrix.py
+++ b/tools/feature_matrix.py
@@ -93,7 +93,7 @@ min_browser_versions = {
     'chrome': 80,
     'firefox': 114,
     'safari': 150000,
-    'node': UNSUPPORTED,
+    'node': 0, # This is a browser only feature, no requirements on Node.js
   },
 }
 


sbc100 (approved):

juj:
Naively adding `#error` messages resulted in all of our test suites failing, since we do currently have too low default MIN_*_VERSION requirements to enable EXPORT_ES6.

So changed this PR to establish ES6 Module Workers as a feature instead.

src/lib/libpthread.js
```diff
@@ -42,6 +42,15 @@ const pthreadWorkerScript = TARGET_JS_NAME;
...
 #if EXPORT_ES6
+#if MIN_FIREFOX_VERSION < 114
+#error new Worker() supports ECMAScript module only starting from Firefox 114. Pass -sMIN_FIREFOX_VERSION=114 to target -sEXPORT_ES6 with -pthread. See https://caniuse.com/mdn-api_worker_worker_ecmascript_modules
+#endif
+#if MIN_CHROME_VERSION < 80
+#error new Worker() supports ECMAScript module only starting from Chrome 80. Pass -sMIN_CHROME_VERSION=80 to target -sEXPORT_ES6 with -pthread. See https://caniuse.com/mdn-api_worker_worker_ecmascript_modules
+#endif
+#if MIN_SAFARI_VERSION < 150000
+#error new Worker() supports ECMAScript module only starting from Safari 15. Pass -sMIN_SAFARI_VERSION=150000 to target -sEXPORT_ES6 with -pthread. See https://caniuse.com/mdn-api_worker_worker_ecmascript_modules
+#endif
```
sbc100, line 53: It is even possible to hit these error now that the feature is added?   i.e. `-sEXPORT_ES6` + `-pthreads` will now force the correct versions or fail, right?

sbc100 (approved):

src/lib/libpthread.js
juj: I was pondering the same, so wanted to add these in to verify that won't be the case. Doesn't hurt to have them?

Or I could remove them - but seems like they're safe to have, they could trigger on possible refactors down the road.

PR #16245, opened 2022-02-10
Attempt to work around Firefox CI issue. See #16242

kripken:
I can't reproduce the issue locally, but this change got the test to pass on CI,
so I suspect the theory in #16242 is right: Firefox Dev now creates invalid
OffscreenCanvas elements in some cases. That happens on our CI, but not
locally even with the same flags, so it might be dependent on the OS or the
graphics backend or otherwise affect how Firefox creates the context. The
effect is that an OffscreenCanvas is created, but it lacks methods on the
object. To work around it, check for a valid method appearing on the object,
and if it's not there, use a normal (non-offscreen) Canvas.

Also add an assertion, which is mostly not needed assuming the workaround
is valid, but may help in the future.

Also remove `-O2` in the test, which was not necessary for any reason (and it
happens to disable assertions).

Issue #16242, opened 2022-02-10
Firefox dev edition browser test errors


kripken:
In the past few days `browser.test_alloctext` and others fail on `tempCtx.measureText is not a function / undefined`. That seems to not be a flake, but I can't reproduce it locally even with the same browser and flags. Perhaps it depends on OS support somehow and/or how it renders there.

My best guess is that this is related to offscreen canvas perhaps, which is still experimental, and which that test can use (`TTF_Init` will try to use one).

PR #25409, opened 2025-09-28
Fix SDL_TTF to work with older browser versions.

juj:


src/lib/libsdl.js
```diff
@@ -3122,7 +3122,11 @@ var LibrarySDL = {
       // valid. Check that explicitly and fall back to a plain Canvas if we need
       // to. See https://github.com/emscripten-core/emscripten/issues/16242
       if (typeof SDL.ttfContext.measureText != 'function') {
-        abort('bad context');
+        // According to https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvasRenderingContext2D
+        // OffscreenCanvasRenderingContext2D.measureText() appeared in
+        // Chrome 69, Firefox 105 and Safari 16.4. Fall back to using regular
+        // Canvas2D if OffscreenCanvas2D exists, but does not look workable.
```
sbc100: Is this comment just adding more detail the comment above?  If so, maybe merge them into a single comment?

Ideally we could wrap this whole block in some `MIN_XX_VERSION` checks, so that we clean it up one day, but that doesn't need to part of this PR I guess.

src/lib/libsdl.js
```diff
@@ -3122,7 +3122,11 @@ var LibrarySDL = {
       // valid. Check that explicitly and fall back to a plain Canvas if we need
       // to. See https://github.com/emscripten-core/emscripten/issues/16242
       if (typeof SDL.ttfContext.measureText != 'function') {
```
sbc100: Probably enough to just do `if (!SDL.ttfContext.measureText)`.. but again doesn't have to part of this PR.

src/lib/libsdl.js
sbc100: These version numbers (Chrome 69, Firefox 105 and Safari 16.4) looks like the same as the ones that support  `OffscreenCanvas` at all, but the comment seems to relate to `measureText`.   ?

src/lib/libsdl.js
```diff
@@ -3122,7 +3122,11 @@ var LibrarySDL = {
...
-        abort('bad context');
+        // According to https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvasRenderingContext2D
+        // OffscreenCanvasRenderingContext2D.measureText() appeared in
+        // Chrome 69, Firefox 105 and Safari 16.4. Fall back to using regular
+        // Canvas2D if OffscreenCanvas2D exists, but does not look workable.
+        throw 'no OffscreenCanvasRenderingContext2D.measureText';
```
sbc100: I wonder if we should just delete this block?  It seems like https://github.com/emscripten-core/emscripten/pull/16245 was a workaround for some kind of bug with firefox on tip-of-tree,   

Did you see this line be hit in practice anywhere?

juj: Yes, I was hitting this when testing on Firefox 91 ESR. If we deleted this block, the test would then fail there.

src/lib/libsdl.js
juj: Removed the old comment.

> These version numbers (Chrome 69, Firefox 105 and Safari 16.4) looks like the same as the ones that support `OffscreenCanvas` at all

Maybe those versions are aggregations of "these versions have the full OffscreenCanvas spec." Firefox 91 ESR definitely does have `new OffscreenCanvas`, but no `.measureText()` here.

src/lib/libsdl.js
```diff
@@ -3117,12 +3117,12 @@ var LibrarySDL = {
...
-      // Firefox support for OffscreenCanvas is still experimental, and it seems
-      // like CI might be creating a context here but one that is not entirely
-      // valid. Check that explicitly and fall back to a plain Canvas if we need
-      // to. See https://github.com/emscripten-core/emscripten/issues/16242
-      if (typeof SDL.ttfContext.measureText != 'function') {
-        abort('bad context');
+      // According to https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvasRenderingContext2D
+      // OffscreenCanvasRenderingContext2D.measureText() appeared in
+      // Chrome 69, Firefox 105 and Safari 16.4. Fall back to using regular
+      // Canvas2D if OffscreenCanvas2D exists, but does not look workable.
```
sbc100, line 3124: Can you keep the link to the original issue: https://github.com/emscripten-core/emscripten/issues/16242

src/lib/libsdl.js
```diff
@@ -3117,12 +3117,12 @@ var LibrarySDL = {
...
+      if (!SDL.ttfContext.measureText) {
+        throw 1; // no OffscreenCanvasRenderingContext2D.measureText
       }
```
sbc100, line 3128: I wish we could wrap these 3 lines in `#if MIN_FIREFOX_VERSION  < XX`, but I guess we don't know exactly when they added `measureText`?

juj: Doesn't have to be precise. I added check for < Firefox 128 conservatively. (Firefox 128 does have it)

sbc100 (approved):

PR #25410, opened 2025-09-28
Lift the previous 100 max workers to limit

juj:
A long time ago, Firefox had a default limit of `dom.workers.maxPerDomain == 20`, and a hardcoded cap of `navigator.hardwareConcurrency` reporting 16 even on systems that would have more hardware threads.

To work around those limits for Emscripten tests that hammered lots of Workers, we had bumped the default `dom.workers.maxPerDomain` limit from 20 up to 100.

In [August 2016](https://github.com/mozilla-firefox/firefox/commit/f6650e86149143e8ccebfac01fb8a68ed0365655#diff-8c2c5665d52eb6026c4017acc0c311c6d9cc6b2b8719f50c1539ee836f2ea361R97-R114), Firefox itself bumped the same default limit of  `dom.workers.maxPerDomain` from 20 up to 512.

Then in April of this year, Firefox [lifted the hardcoded cap](https://github.com/mozilla-firefox/firefox/commit/af05ed3758dcea82b568e3b2daf0c3f2171bab14) of `navigator.hardwareConcurrency` from 16 up to 128.

This means that the limit `user_pref("dom.workers.maxPerDomain", 100);` we now have is too small, and on a 256-thread workstation, caused attempting to spawn >100 Workers in the Pthread pool to hang, as 128 concurrent Workers would not be allowed.

So remove  the previous 100 max workers limit. Explicitly bump it up to the same default limit 512 that Firefox has to make the limit go away in testing older browsers as well.

Fixes e.g. `browser.test_pthread_hardware_concurrency` on >128-thread workstations.

### DIFF
--- a/test/firefox_user.js
+++ b/test/firefox_user.js
@@ -1,5 +1,6 @@
-// Lift the default max 20 workers limit to something higher to avoid hangs when page needs to spawn a lot of threads.
-user_pref("dom.workers.maxPerDomain", 100);
+// Old Firefox browsers have a maxPerDomain limit of 20. Newer Firefox browsers default to 512. Match the new
+// default here to help test spawning a lot of threads also on older Firefox versions.
+user_pref("dom.workers.maxPerDomain", 512);
 // Always allow opening popups
 user_pref("browser.popups.showPopupBlocker", false);
 user_pref("dom.disable_open_during_load", false);
--- a/emrun.py
+++ b/emrun.py
@@ -229,8 +229,9 @@ def create_emrun_safe_firefox_profile():
   temp_firefox_profile_dir = tempfile.mkdtemp(prefix='temp_emrun_firefox_profile_')
   with open(os.path.join(temp_firefox_profile_dir, 'prefs.js'), 'w') as f:
     f.write('''
-// Lift the default max 20 workers limit to something higher to avoid hangs when page needs to spawn a lot of threads.
-user_pref("dom.workers.maxPerDomain", 100);
+// Old Firefox browsers have a maxPerDomain limit of 20. Newer Firefox browsers default to 512. Match the new
+// default here to help test spawning a lot of threads also on older Firefox versions.
+user_pref("dom.workers.maxPerDomain", 512);
 // Always allow opening popups
 user_pref("browser.popups.showPopupBlocker", false);
 user_pref("dom.disable_open_during_load", false);


test/firefox_user.js
```diff
@@ -1,5 +1,6 @@
-// Lift the default max 20 workers limit to something higher to avoid hangs when page needs to spawn a lot of threads.
-user_pref("dom.workers.maxPerDomain", 100);
+// Old Firefox browsers have a maxPerDomain limit of 20. Newer Firefox browsers default to 512. Match the new
+// default here to help test spawning a lot of threads also on older Firefox versions.
+user_pref("dom.workers.maxPerDomain", 512);
```
sbc100, line 3: Can we just delete this instead, to keep the config minimal?

juj: No, since that would prevent testing in older browsers that still have the maxPerDomain limit.

sbc100: It wouldn't prevent testing would it, it would just mean that the max workers using in the test was whatever firefox intended in that particular version.

I think honoring the browser-default `maxPerDomain` makes sense, since then we are testing what most users of firefox will have.

Also, for your CI your will be tests with recent firefox right?  So it won't slow you down to just delete this line will it?

sbc100: If the specific `test_pthread_hardware_concurrency` fails on older firefox then perhaps it should just `skip` in that case, since the test is testing something that the browser doesn't support (without a config mod anyway).

juj: > I think honoring the browser-default `maxPerDomain` makes sense

That is unfortunately not how maxPerDomain limit works. Honoring the old limit of 20 is not particularly useful for unit test suites. (that is what led to needing to bump it in the first place)

If you have `navigator.hardwareConcurrency` say 32, and `maxPerDomain` of 20, then the web page will silently block any `new Worker()` calls from ever starting a Worker, until the previous Workers finish.

So if we have tests that want to launch e.g. a pool of #hardwareConcurrency of pthreads, then those pthreads will never launch, leading to the test failing.

In other words, if we honored the default, and the default was 20, and my system had 256 threads, then all the tests spawning a pool of threads would just hang and time out.

sbc100: This sounds like maybe a real bug in older versions of FF then?   What happens to users of older FF who run real wasm programs with thread pools?   Is there something we can do to prevent the hanging in the wild?

sbc100: i.e. it sounds like we need to somehow expose maxPerDomain to the user so that they avoid spawning more threads than that?  But I guess its not visible to JS?

juj: > This sounds like maybe a real bug in older versions of FF then? What happens to users of older FF who run real wasm programs with thread pools? Is there something we can do to prevent the hanging in the wild?

Yes, it was. They would experience a hang too.

But we don't have a good way to know if `new Worker()` will never actually proceed. I had a spec bug about that somewhere to get a proper `new Worker()` or throw on ENORESOURCES, but it never went forward.

End users will need to manage their own pools so that they don't attempt to overspawn Workers. Fortunately end users rarely have to run posixtest suite for example to hammer a lot of thread creation.

juj: > i.e. it sounds like we need to somehow expose maxPerDomain to the user so that they avoid spawning more threads than that? But I guess its not visible to JS?

Unfortunately not. :/

sbc100 (approved):
lgtm

(I wonder if we should go through all these settings and see which ones are really still needed)

PR #25411, opened 2025-09-29
Use createNamedFunction in the dynamic linker

sbc100:
This helps with debugging since the runtime-generated functions show up in backtraces with meaningful names.

### DIFF
--- a/src/lib/libcore.js
+++ b/src/lib/libcore.js
@@ -41,6 +41,10 @@ addToLibrary({
   setTempRet0: '$setTempRet0',
   getTempRet0: '$getTempRet0',
 
+  // Assign a name to a given function. This is mostly useful for debugging
+  // purposes in cases where new functions are created at runtime.
+  $createNamedFunction: (name, func) => Object.defineProperty(func, 'name', { value: name }),
+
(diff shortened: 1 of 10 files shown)


src/lib/libcore.js
```diff
@@ -41,6 +41,10 @@ addToLibrary({
   setTempRet0: '$setTempRet0',
   getTempRet0: '$getTempRet0',
 
+  // Assign a name of given function. This is mostly useful for debugging
```
kripken:
```suggestion
  // Assign a name to a given function. This is mostly useful for debugging
```

src/lib/libcore.js
```diff
@@ -41,6 +41,10 @@ addToLibrary({
...
+  // purposes in cases where new functions are created at runtime.
+  $createNamedFunction: (name, func) => Object.defineProperty(func, 'name', { value: name }),
```
kripken, line 46: maybe this could be `nameFunction` as it doesn't actually create a function?

sbc100: I think I'll just leave it for now, since the name of the function itself is not completely internal.

PR #25414, opened 2025-09-29
Generalize existing Node.js development-time version check

juj:
Generalize existing Node.js development-time version check into common file that also checks for browser versions.

### DIFF
--- a/src/minimum_runtime_check.js
+++ b/src/minimum_runtime_check.js
@@ -1,3 +1,9 @@
+/**
+ * @license
+ * Copyright 2024 The Emscripten Authors
+ * SPDX-License-Identifier: MIT
+ */
+
 #if ASSERTIONS
 
 // "30.0.0" -> 300000
@@ -14,13 +20,13 @@ var packedVersionToHumanReadable = n => [n / 10000 | 0, (n / 100 | 0) % 100, n %
 var currentNodeVersion = typeof process !== 'undefined' && process?.versions?.node ? humanReadableVersionToPacked(process.versions.node) : Infinity;
 if (currentNodeVersion < {{{ MIN_NODE_VERSION }}}) throw new Error(`This emscripten-generated code requires node v${ packedVersionToHumanReadable({{{ MIN_NODE_VERSION }}}) } (detected v${packedVersionToHumanReadable(currentNodeVersion)})`);
 
-var currentSafariVersion = navigator?.userAgent?.includes("Safari/") && navigator.userAgent.match(/Version\/(\d+\.?\d*\.?\d*)/) ? humanReadableVersionToPacked(navigator.userAgent.match(/Version\/(\d+\.?\d*\.?\d*)/)[1]) : Infinity;
+var currentSafariVersion = typeof navigator !== 'undefined' && navigator?.userAgent?.includes("Safari/") && navigator.userAgent.match(/Version\/(\d+\.?\d*\.?\d*)/) ? humanReadableVersionToPacked(navigator.userAgent.match(/Version\/(\d+\.?\d*\.?\d*)/)[1]) : Infinity;
 if (currentSafariVersion < {{{ MIN_SAFARI_VERSION }}}) throw new Error(`This emscripten-generated code requires Safari v${ packedVersionToHumanReadable({{{ MIN_SAFARI_VERSION }}}) } (detected v${currentSafariVersion})`);
 
-var currentFirefoxVersion = navigator?.userAgent?.match(/Firefox\/(\d+(?:\.\d+)?)/) ? parseFloat(navigator.userAgent.match(/Firefox\/(\d+(?:\.\d+)?)/)[1]) : Infinity;
+var currentFirefoxVersion = typeof navigator !== 'undefined' && navigator?.userAgent?.match(/Firefox\/(\d+(?:\.\d+)?)/) ? parseFloat(navigator.userAgent.match(/Firefox\/(\d+(?:\.\d+)?)/)[1]) : Infinity;
 if (currentFirefoxVersion < {{{ MIN_FIREFOX_VERSION }}}) throw new Error(`This emscripten-generated code requires Firefox v{{{ MIN_FIREFOX_VERSION }}} (detected v${currentFirefoxVersion})`);
 
-var currentChromeVersion = navigator?.userAgent?.match(/Chrome\/(\d+(?:\.\d+)?)/) ? parseFloat(navigator.userAgent.match(/Chrome\/(\d+(?:\.\d+)?)/)[1]) : Infinity;
+var currentChromeVersion = typeof navigator !== 'undefined' && navigator?.userAgent?.match(/Chrome\/(\d+(?:\.\d+)?)/) ? parseFloat(navigator.userAgent.match(/Chrome\/(\d+(?:\.\d+)?)/)[1]) : Infinity;
 if (currentChromeVersion < {{{ MIN_CHROME_VERSION }}}) throw new Error(`This emscripten-generated code requires Chrome v{{{ MIN_CHROME_VERSION }}} (detected v${currentChromeVersion})`);
 
 #endif
--- a/src/minimum_runtime_check.js
+++ b/src/minimum_runtime_check.js
@@ -6,30 +6,51 @@
 
 #if ASSERTIONS
 
-// "30.0.0" -> 300000
-function humanReadableVersionToPacked(str) {
-  str = str.split('-')[0]; // Remove any trailing part from e.g. "12.53.3-alpha"
-  var vers = str.split('.').slice(0, 3);
-  while(vers.length < 3) vers.push('00');
-  vers = vers.map((n, i, arr) => n.padStart(2, '0'));
-  return vers.join('');
-}
-// 300000 -> "30.0.0"
-var packedVersionToHumanReadable = n => [n / 10000 | 0, (n / 100 | 0) % 100, n % 100].join('.');
+(function() {
+  // "30.0.0" -> 300000
+  function humanReadableVersionToPacked(str) {
+    str = str.split('-')[0]; // Remove any trailing part from e.g. "12.53.3-alpha"
+    var vers = str.split('.').slice(0, 3);
+    while(vers.length < 3) vers.push('00');
+    vers = vers.map((n, i, arr) => n.padStart(2, '0'));
+    return vers.join('');
+  }
+  // 300000 -> "30.0.0"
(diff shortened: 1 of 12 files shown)


src/shell_minimal.js
```diff
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: MIT
  */
 
+#include "minimum_runtime_check.js"
```
sbc100, line 7: Maybe just add this to `runtime_common.js` which is already included in both runtime modes?

src/minimum_runtime_check.js
```diff
@@ -0,0 +1,35 @@
...
+if (currentNodeVersion < {{{ MIN_NODE_VERSION }}}) throw new Error(`This emscripten-generated code requires node v${ packedVersionToHumanReadable({{{ MIN_NODE_VERSION }}}) } (detected v${packedVersionToHumanReadable(currentNodeVersion)})`);
+
+var currentSafariVersion = typeof navigator !== 'undefined' && navigator?.userAgent?.includes("Safari/") && navigator.userAgent.match(/Version\/(\d+\.?\d*\.?\d*)/) ? humanReadableVersionToPacked(navigator.userAgent.match(/Version\/(\d+\.?\d*\.?\d*)/)[1]) : TARGET_NOT_SUPPORTED;
+if (currentSafariVersion < {{{ MIN_SAFARI_VERSION }}}) throw new Error(`This emscripten-generated code requires Safari v${ packedVersionToHumanReadable({{{ MIN_SAFARI_VERSION }}}) } (detected v${currentSafariVersion})`);
+
+var currentFirefoxVersion = typeof navigator !== 'undefined' && navigator?.userAgent?.match(/Firefox\/(\d+(?:\.\d+)?)/) ? parseFloat(navigator.userAgent.match(/Firefox\/(\d+(?:\.\d+)?)/)[1]) : TARGET_NOT_SUPPORTED;
```
sbc100: These days you can do `globalThis.navigator?.userAgent?.match` avoiding the need for the `typeof` check.  It makes it a little smaller and easier to read.

See #25375 and #25381.

src/minimum_runtime_check.js
```diff
@@ -0,0 +1,35 @@
...
+  str = str.split('-')[0]; // Remove any trailing part from e.g. "12.53.3-alpha"
+  var vers = str.split('.').slice(0, 3);
+  while(vers.length < 3) vers.push('00');
+  vers = vers.map((n, i, arr) => n.padStart(2, '0'));
+  return vers.join('');
+}
+// 300000 -> "30.0.0"
+var packedVersionToHumanReadable = n => [n / 10000 | 0, (n / 100 | 0) % 100, n % 100].join('.');
+
+var TARGET_NOT_SUPPORTED = 2147483647;
```
sbc100: Can you do `var TARGET_NOT_SUPPORTED = {{{ TARGET_NOT_SUPPORTED }}};` here?

src/minimum_runtime_check.js
```diff
@@ -0,0 +1,35 @@
...
+if (currentSafariVersion < {{{ MIN_SAFARI_VERSION }}}) throw new Error(`This emscripten-generated code requires Safari v${ packedVersionToHumanReadable({{{ MIN_SAFARI_VERSION }}}) } (detected v${currentSafariVersion})`);
+
+var currentFirefoxVersion = typeof navigator !== 'undefined' && navigator?.userAgent?.match(/Firefox\/(\d+(?:\.\d+)?)/) ? parseFloat(navigator.userAgent.match(/Firefox\/(\d+(?:\.\d+)?)/)[1]) : TARGET_NOT_SUPPORTED;
+if (currentFirefoxVersion < {{{ MIN_FIREFOX_VERSION }}}) throw new Error(`This emscripten-generated code requires Firefox v{{{ MIN_FIREFOX_VERSION }}} (detected v${currentFirefoxVersion})`);
+
+var currentChromeVersion = typeof navigator !== 'undefined' && navigator?.userAgent?.match(/Chrome\/(\d+(?:\.\d+)?)/) ? parseFloat(navigator.userAgent.match(/Chrome\/(\d+(?:\.\d+)?)/)[1]) : TARGET_NOT_SUPPORTED;
+if (currentChromeVersion < {{{ MIN_CHROME_VERSION }}}) throw new Error(`This emscripten-generated code requires Chrome v{{{ MIN_CHROME_VERSION }}} (detected v${currentChromeVersion})`);
```
sbc100: Maybe wrap this whole file in an IFFE to avoid creating all these global vars?

sbc100: Can you put the `throw` on its own line and using open and closing curly braces.  These long lines I find hard to parse.

sbc100 (commented):
Nice! 

I was thinking of doing the same thing myself.

I was also thinking that the code for setting `ENVIRONMENT_IS_NODE` etc could be shared, but that could be followup.  I guess we could add that code to this file and call it `runtime_environemnt_detect.js` or something like that.

src/shell_minimal.js
juj: The reason I made a separate file is that this code would be run as the very first thing in the generated .js.

If we put the code somewhere in the middle, we risk already attempting to execute some code that might require a new feature (e.g. ES6 Modules or some language syntax keyword, etc)

src/minimum_runtime_check.js
juj: That would require globalThis, which might not exist in an old browser. This is an ASSERTIONS check anyways, so code size is not that critical.

src/minimum_runtime_check.js
juj: Yeh, will do in a moment.

src/minimum_runtime_check.js
sbc100: Oh I see, we know the globalThis exists on supported browser, but here we are trying to explicitly detect unsupported ones.  That makes sense. 

I guess you should do check for `typeof globalThis` first before the other checks?  lgtm either way though

src/shell_minimal.js
sbc100: Ah yes that makes sense.

test/codesize/test_codesize_minimal_O0.expected.js
```diff
@@ -1,6 +1,55 @@
...
+  var currentSafariVersion = typeof navigator !== 'undefined' && navigator?.userAgent?.includes("Safari/") && navigator.userAgent.match(/Version\/(\d+\.?\d*\.?\d*)/) ? humanReadableVersionToPacked(navigator.userAgent.match(/Version\/(\d+\.?\d*\.?\d*)/)[1]) : TARGET_NOT_SUPPORTED;
+  if (currentSafariVersion < TARGET_NOT_SUPPORTED && 150000 == TARGET_NOT_SUPPORTED) {
+    throw new Error(`This page was compiled without support for Safari browser. Pass -sMIN_SAFARI_VERSION=${currentSafariVersion} or lower to enable support for this browser.`);
+  }
+  if (currentSafariVersion < 150000) {
+    throw new Error(`This emscripten-generated code requires Safari v${ packedVersionToHumanReadable(150000) } (detected v${currentSafariVersion})`);
+  }
+
+  var currentFirefoxVersion = typeof navigator !== 'undefined' && navigator?.userAgent?.match(/Firefox\/(\d+(?:\.\d+)?)/) ? parseFloat(navigator.userAgent.match(/Firefox\/(\d+(?:\.\d+)?)/)[1]) : TARGET_NOT_SUPPORTED;
+  if (currentFirefoxVersion < TARGET_NOT_SUPPORTED && 79 == TARGET_NOT_SUPPORTED) {
```
sbc100: `79 == TARGET_NOT_SUPPORTED` looks like typo.. is it not?

src/minimum_runtime_check.js
sbc100: Now that this is in an IFFY (so these are just local vars) could we shorten all the `currentChromeVersion` to just `chromeVersion` for readability?

test/codesize/test_codesize_minimal_O0.expected.js
juj: That was just runtime test on what could have been a compile time check. Optimized it to take place at compile time.

tools/link.py
```diff
@@ -172,7 +172,7 @@ def setup_environment_settings():
   if any(x for x in settings.ENVIRONMENT if x not in VALID_ENVIRONMENTS):
     exit_with_error(f'Invalid environment specified in "ENVIRONMENT": {settings.ENVIRONMENT}. Should be one of: {",".join(VALID_ENVIRONMENTS)}')
 
-  settings.ENVIRONMENT_MAY_BE_WEB = not settings.ENVIRONMENT or 'web' in settings.ENVIRONMENT
+  settings.ENVIRONMENT_MAY_BE_WEB = not settings.ENVIRONMENT or 'web' in settings.ENVIRONMENT or 'worker' in settings.ENVIRONMENT
```
kleisauke, line 175: Was this change intentional? It causes web-specific code to be included in the final JS output when building with `-sENVIRONMENT=node` + `-pthread`, see:
https://github.com/emscripten-core/emscripten/blob/7378d2f64c50b036f01d4c817568324501182359/tools/link.py#L165-L169
https://github.com/emscripten-core/emscripten/blob/7378d2f64c50b036f01d4c817568324501182359/emcc.py#L430-L432

juj: Hey, thanks for pinging.

The change was intentional, because I saw test code that was setting `-sENVIRONMENT=worker`, and then running the resulting code in a web browser inside a Web Worker.

That looks like we have an overloaded meaning then, i.e. in your case, this ends up passing `-sENVIRONMENT=node,worker` I presume?

And in the test code case, it was passing just `-sENVIRONMENT=worker` to mean `-sENVIRONMENT=web,worker`.

I wonder if a good resolution here would then be to disallow and abort a build if one passes `-sENVIRONMENT=worker`, and always require user to either pass `-sENVIRONMENT=node,worker` or `-sENVIRONMENT=web,worker` to disambiguate.

(Naturally `-pthread` would then also imply adding the `,worker` part into -sENVIRONMENT)

Would that work?

kleisauke: Indeed, this effectively ends up passing `-sENVIRONMENT=node,worker`. For full context, I noticed this on wasm-vips' test suite while trying to upgrade to Emscripten 4.0.16, where full concurrency isn't leveraged due to:
https://github.com/kleisauke/wasm-vips/blob/v0.0.15/src/vips-library.js#L14-L22

I think it makes sense to just bail out if someone uses `-sENVIRONMENT=worker` (and reverting this part of the code). 👍

juj: Great, posted https://github.com/emscripten-core/emscripten/pull/25514.

PR #25417, opened 2025-09-29
More use of globalThis to save on code size. NFC

sbc100:
Followup to #25381

### DIFF
--- a/src/lib/libopenal.js
+++ b/src/lib/libopenal.js
@@ -2064,7 +2064,7 @@ var LibraryOpenAL = {
       }
     }
 
-    if (typeof AudioContext != 'undefined' || typeof webkitAudioContext != 'undefined') {
+    if (globalThis.AudioContext || globalThis.webkitAudioContext) {
       var deviceId = AL.newId();
       AL.deviceRefCounts[deviceId] = 0;
       return deviceId;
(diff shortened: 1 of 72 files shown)


juj (approved):
Very nice 👍

src/lib/libopenal.js
```diff
@@ -2398,16 +2398,14 @@ var LibraryOpenAL = {
...
+      if (globalThis.AudioContext || globalThis.webkitAudioContext) {
         ret = AL.DEVICE_NAME;
       } else {
         return 0;
       }
       break;
     case 0x1005 /* ALC_DEVICE_SPECIFIER */:
-      if (typeof AudioContext != 'undefined' ||
-          typeof webkitAudioContext != 'undefined') {
+      if (globalThis.AudioContext || globalThis.webkitAudioContext) {
```
kripken, line 2408: I think we can probably remove webkitAudioContext?

sbc100: Will do that as a followup.

sbc100: Actually it looks like the unprefixed version is needed until Safari 14.5

kripken: Weird...wasn't Safari/Webkit the first to implement WebAudio back in the day?

But I see MDN says

kripken: `Safari 14.1 (Release date: ⁨2021-04-26⁩) `

sbc100: I think Safari iOS its 14.5

juj: Safari < 14.5 still needs the webkit prefix. I authored this on another call site recently: https://github.com/emscripten-core/emscripten/pull/25356

> Weird...wasn't Safari/Webkit the first to implement WebAudio back in the day?

I think what happened is that they went ahead to implement a very first version of the draft/spec, and then the spec got matured, and they took a long time to come back and unprefix their implementation. The `webkitAudioContext()` and the `.noteOn()` quirks stayed for a long time.

PR #25419, opened 2025-09-29
Add env. var. EMTEST_S_SETTINGS to allow overriding -s variables for a test run.

juj:
Adjust the testing base class to read extra -s settings from environment variable EMTEST_S_SETTINGS.

This way I can have my CI perform custom test runs with `EMTEST_S_SETTINGS=MIN_FIREFOX_VERSION=xyz` to verify support for a specific Firefox version.

For example, I can then run through all ESR versions, and pass `-s MIN_FIREFOX_VERSION=xyz` to match that specific ESR version:

image

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -167,7 +167,7 @@ def errlog(*args):
 
 def parse_emtest_s_settings():
   s_settings = {}
-  for setting in shlex.split(os.getenv('EMTEST_S_SETTINGS')):
+  for setting in shlex.split(os.getenv('EMTEST_S_SETTINGS', '')):
     if '=' in setting:
       key, value = setting.split('=')
     else:
--- a/test/common.py
+++ b/test/common.py
@@ -165,6 +165,20 @@ def errlog(*args):
   print(*args, file=sys.stderr)
 
 
+def parse_emtest_s_settings():
+  s_settings = {}
+  for setting in shlex.split(os.getenv('EMTEST_S_SETTINGS')):
+    if '=' in setting:
+      key, value = setting.split('=')
+    else:
+      key = setting
+      value = '1'
+
+    logger.debug(f'Applying testing override -s{key}={value}')
+    s_settings[key] = value
+  return s_settings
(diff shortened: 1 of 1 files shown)


sbc100:
When I test locally I normally use `EMCC_CFLAGS=-sMIN_FIREFOX_VERSION=xx` for this kind of thing.  Does that not work in this case?

juj:
That unfortunately doesn't: the test runners fail against -Werror when -sMIN_FIREFOX_VERSION goes unused.

sbc100:
> That unfortunately doesn't: the test runners fail against -Werror when -sMIN_FIREFOX_VERSION goes unused.

You mean because its setting compile time flags too?    Can you also add `-Wno-unused-command-line-argument`?

test/common.py
```diff
@@ -1422,6 +1436,10 @@ def setUp(self):
       utils.write_file(LAST_TEST, self.id() + '\n')
     os.chdir(self.working_dir)
 
+    # Apply overrides to -s settings from EMTEST_S_SETTINGS env. variable:
+    for k, v in parse_emtest_s_settings().items():
+      self.set_setting(k, v)
```
sbc100, line 1441: Why not do this in `RunnerCore::setUp` which would make it universally useful?

Also, why not support all different flags (not just -s flags), e.g. use `EMTEST_EXTRA_LDFLAGS=xxx`?  Then add to `self.ldflags`?

juj: Good points. I'll rerun the CI harness by attempting `EMCC_CFLAGS=-sMIN_FIREFOX_VERSION=65 -Wno-unused-command-line-argument` if that will be good enough, and if not, look at a `EMTEST_EXTRA_LDFLAGS` then.

juj:
Closing, since it looks like `EMCC_CFLAGS=-sMIN_FIREFOX_VERSION=65 -Wno-unused-command-line-argument` is enough to achieve the same.

PR #25421, opened 2025-09-29
OffscreenCanvas support in feature matrix

juj:
Add item for `OffscreenCanvas` support in feature matrix, and update missing `@requires_offscreen_canvas` in browser test.

This will automatically set min browser versions when `-sOFFSCREEN_CANVAS` is used, and error our if the users explicitly requests a too-old browser version

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -2643,6 +2643,7 @@ Module["preRun"] = () => {
     self.btest_exit('webgl_create_context2.c')
 
   @requires_graphics_hardware
+  @requires_offscreen_canvas
   # Verify bug https://github.com/emscripten-core/emscripten/issues/22943: creating a WebGL context with explicit swap control and offscreenCanvas
   @parameterized({
     'offscreencanvas': (['-sOFFSCREENCANVAS_SUPPORT'],),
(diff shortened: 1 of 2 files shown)


test/test_browser.py
```diff
@@ -2643,6 +2643,7 @@ def test_html5_webgl_create_context2(self):
     self.btest_exit('webgl_create_context2.c')
 
   @requires_graphics_hardware
+  @requires_offscreen_canvas
```
sbc100, line 2646: Doesn't only one of the variants require it?

sbc100 (approved):

sbc100:
Can you mention in the PR description what the effect this is, since I don't think its obvious. 

Correct me if I'm wrong but something like:  "This will automatically set min browser versions when `-sOFFSCREEN_CANVAS` is used, and error our if the users explictly requests a too-old browser version".

test/test_browser.py
juj: Yeah, that's right.

I thought to keep this change simple and over-conservative, to avoid too much churn from testing old browsers.

juj:
> Correct me if I'm wrong but something like: "This will automatically set min browser versions when `-sOFFSCREEN_CANVAS` is used, and error our if the users explictly requests a too-old browser version".

That's good. Added, thanks.

sbc100 (approved):

Issue #25365, opened 2025-09-23
safari 26.0 can not handle wasm-exceptions anymore


mithrendal:
in case anyone wonders and looks into here for help because their emsdk built web apps are not working anymore with apples latest macOS, iPadOS and iPhoneOS...

my two web projects were since years compiled and linked to use the faster and more efficient native wasm-exceptions

### My findings: 
+ compiled and linked with `-fwasm-exceptions` will crash the web app on startup on apples latest safari 26 while the previous safari versions and all other web browsers still work fine.

I had to rebuild with `-fexceptions` to make them run also under apples latest safari 26.0 

for reference the issue on my project:
https://github.com/vc64web/virtualc64web/issues/285

mithrendal:
to test a build with the faster native wasm exceptions run https://vamigaweb.github.io/uat/

the build with the about 13% slower javascript-based exceptions you can start with https://vamigaweb.github.io

because vAmigaWeb uses intensively local browser cache  to make the app offline capable the best is to open it  in a private browser tab to get the current version.

aheejin:
Can you paste the error message? I don't have Safari 26 available on my Mac yet.

I installed the latest JavaScriptCore (v300391) using jsvu and ran it with a simple program compiled with `-fwasm-exceptions`, and it ran successfully. I don't think jsvu supports installing older versions (https://webkitgtk.org/jsc-built-products/x86_64/release/ contains only recent versions). So either the problem was fixed at some point or the problem only manifests in the browser setting or is specific to application characteristics.

mithrendal:
@aheejin 

<img width="716" height="154" alt="Image" src="https://github.com/user-attachments/assets/b46d54d2-0209-4f59-a2ec-a03a9100b81b" />

it is here at the green line where an exception is thrown 

<img width="1122" height="451" alt="Image" src="https://github.com/user-attachments/assets/03d873d5-a770-44c3-b19b-58c90741ee95" />

I will try to step a bit into the entryFunction which is set to in the aboves code _main

UPDATE:

Safari Technology Preview Release 228 (WebKit 21623.1.7.19.1) 

also crashes... hope that safari 26.1 fixes it again

mithrendal:
I opened a bug report at the webkit bug tracker 

https://bugs.webkit.org/show_bug.cgi?id=299629

mithrendal:
Ok I tracked the problem down to this


[...]

PR #25423, opened 2025-09-29
Add feature matrix item for Legacy Wasm EH and Wasm Exnref EH, and separate test skipping to two environment variables

juj:
Add feature for Wasm Exnref support and separate test skips for legacy vs Exnref EH skipping.

This will automatically set min browser versions when either Wasm legacy or Exnref exception handling is used, and error our if the users explictly requests a too-old browser version.


tools/feature_matrix.py
```diff
@@ -155,7 +173,7 @@ def enable_feature(feature, reason, override=False):
         diagnostics.warning(
             'compatibility',
             f'{name}={user_settings[name]} is not compatible with {reason} '
-            f'({min_version} or above required)')
+            f'({name}={min_version} or above required)')
```
juj, line 190: A drive-by improvement to help error message

```
MIN_FIREFOX_VERSION=91 is not compatible with -sWASM_LEGACY_EXCEPTIONS=0 (131 or above required)
```

which one might at first glance read as needing to pass `-sWASM_LEGACY_EXCEPTIONS=131` to fix.

test/common.py
```diff
@@ -1207,18 +1207,21 @@ def require_wasm_legacy_eh(self):
...
   def require_wasm_eh(self):
     self.set_setting('WASM_LEGACY_EXCEPTIONS', 0)
-    if self.try_require_node_version(24):
-      self.node_args.append('--experimental-wasm-exnref')
-      return
 
     if self.is_browser_test():
+      if 'EMTEST_SKIP_WASM_EXNREF_EH' in os.environ:
+        self.skipTest('test requires a browser with new Exnref Wasm exceptions support (and EMTEST_SKIP_WASM_EXNREF_EH is set)')
+      return
```
aheejin: This looks like a repetition of line 1234-1235. Can't we check this only once at the top like we used to do?

tools/feature_matrix.py
```diff
@@ -105,6 +107,22 @@ class Feature(IntEnum):
     'safari': 170000,
     'node': 0, # This is a browser only feature, no requirements on Node.js
   },
+  # Legacy Wasm exceptions was the first (now legacy) format for native exception
+  # handling in WebAssembly.
```
aheejin:
```suggestion
  # Legacy Wasm exceptions was the first (now legacy) format for native
  # exception handling in WebAssembly.
```
Nit: Even though we don't have a rule yet, other comments in this file look like they are wrapped at 80 cols, so..

tools/feature_matrix.py
```diff
@@ -199,3 +217,7 @@ def apply_min_browser_versions():
     enable_feature(Feature.WORKER_ES6_MODULES, 'EXPORT_ES6 with -sWASM_WORKERS')
   if settings.OFFSCREENCANVAS_SUPPORT:
     enable_feature(Feature.OFFSCREENCANVAS_SUPPORT, 'OFFSCREENCANVAS_SUPPORT')
+  if settings.WASM_EXCEPTIONS and settings.WASM_LEGACY_EXCEPTIONS:
+    enable_feature(Feature.WASM_LEGACY_EXCEPTIONS, '-fwasm-exceptions')
+  if not settings.WASM_LEGACY_EXCEPTIONS:
+    enable_feature(Feature.WASM_EXNREF_EXCEPTIONS, '-sWASM_LEGACY_EXCEPTIONS=0')
```
aheejin: This is not nested in any other `if`s. Doesn't this mean we always enable exnref when `WASM_LEGACY_EXCEPTIONS` is not set, even if we are not using Wasm EH at all?

test/codesize/test_codesize_cxx_except_wasm.json
```diff
@@ -1,10 +1,10 @@
...
-  "a.out.js": 19608,
-  "a.out.js.gz": 8109,
+  "a.out.js": 19539,
+  "a.out.js.gz": 8086,
   "a.out.nodebug.wasm": 144630,
   "a.out.nodebug.wasm.gz": 54894,
-  "total": 164238,
-  "total_gz": 63003,
+  "total": 164169,
+  "total_gz": 62980,
```
aheejin: Why would this patch change the code size?

aheejin (commented):
I have been mostly trying to avoid using the term "exnref" in the code, especially in possibly user-facing parts. Others may disagree, but I think, that we implement EH using exnref is an implementation detail that does not necessarily need to be exposed, but was exposed anyway due to the unfortunate churns in the proposal development process. So I've been trying to call each of them "Wasm EH" and "Wasm legacy EH" respectively. I think it's fine (or even clarifying in some cases) to say this is the exnref version in the description or comments though.

I think having a setting specific to the legacy EH is good, but can we keep the name `EMTEST_SKIP_EH` and others? I think it's fine they mean the newest (exnref) EH. The same for the feature names. How about `WASM_LEGACY_EXCEPTIONS` and `WASM_EXCEPTIONS`?

juj:
> I have been mostly trying to avoid using the term "exnref" in the code, especially in possibly user-facing parts. Others may disagree, but I think, that we implement EH using exnref is an implementation detail that does not necessarily need to be exposed

I brought that name up because that is the user-facing name also advertised in [WebAssembly Roadmap](https://webassembly.org/features/), which developers publicly use.

> I've been trying to call each of them "Wasm EH" and "Wasm legacy EH" respectively

I find the name "Wasm EH" very ambiguous. It reads like it is what should be getting enabled if one passes `-fwasm-exceptions` (which is not the case). I found using the explicit name "Wasm Exnref exceptions" to be a very specific/explicit name to assign to this, that couldn't be confused with the earlier one.

Even if it is an internal implementation detail, that detail cannot change without requiring a full new spec? So it would be safe to use as a technical name for the new Wasm EH?

test/codesize/test_codesize_cxx_except_wasm.json
juj: That is because enabling Wasm EH via `-sWASM_LEGACY_EXCEPTIONS=0` will now result in the `MIN_*_VERSIONS` being automatically bumped upwards to the minimum required versions that have this feature.

So as corollary, other places of code can assume those versions, and they will drop old code targeting older browser versions. This is good, and as expected.

(On the other hand, if the user explicitly tried to set, say `-sWASM_LEGACY_EXCEPTIONS=0 -sMIN_FIREFOX_VERSION=100`, then the compilation would explicitly fail with emcc complaining that this version of Firefox cannot target Exnref exceptions)

test/common.py
juj: I wanted to keep the original error message for node v24 descriptive, rather than give a generic error mentioning all browser environments. That is because it might not be obvious to user that node 24 can be used to target exnref exceptions with the flag.

tools/feature_matrix.py
juj: Yes that is correct. I was a bit surprised about that at first as well, but that is the behavior as it currently exists. E.g.

```py
  def require_wasm_eh(self):
    self.set_setting('WASM_LEGACY_EXCEPTIONS', 0)
```
only reverts legacy exceptions off, and does not pass `-fwasm-exceptions`. (The passing of `-fwasm-exceptions` seems to be occurring somewhere else)

That may be behavior to change in the future, at which point this test matrix setup should also then change.

juj:
`can we keep the name EMTEST_SKIP_EH?`

Separating the environment variables to skip tests that verify legacy Wasm EH versus Exnref Wasm EH are important, since those were introduced to browsers at very different times. That allows testing legacy Wasm EH in browsers that do not have Exnref Wasm EH support.

test/codesize/test_codesize_cxx_except_wasm.json
aheejin: Do `MIN_*_VERSIONS` options change code size?

juj: Yeah, for example in https://github.com/emscripten-core/emscripten/pull/25407/files , if user is targeting older browsers, a more verbose code pattern is used. But if all target browser versions are new enough, a shorter code pattern is used.

aheejin:
> I brought that name up because that is the user-facing name also advertised in [WebAssembly Roadmap](https://webassembly.org/features/), which developers publicly use.

So the plan has been that we deprecate the legacy EH in near future. I'm not sure if that's a realistic short-term goal at this point, but we'd like to at least make the legacy EH a rarely used old special case we keep only for the compatibility's sake, even if we are not successful in completely deprecating it from the web. That's the reason I wanted to just call the new EH without any qualifiers, while calling the old EH "legacy EH", like here and other places:
https://github.com/emscripten-core/emscripten/blob/834eb888e3c841723bb1248f6a13b9fa66aff177/test/common.py#L815-L817

> I find the name "Wasm EH" very ambiguous. It reads like it is what should be getting enabled if one passes `-fwasm-exceptions` (which is not the case).

Yes that's not the case because `WASM_LEGACY_EXCEPTIONS` still defaults to 1 now, because not all browsers supported it until very recently (and Node still needs a flag). My hope is to make it default to 0 in near future, which is orthogonal to whether we actually succeed in the deprecation of the legacy EH.

> Even if it is an internal implementation detail, that detail cannot change without requiring a full new spec? So it would be safe to use as a technical name for the new Wasm EH?

I don't mean I want to change that detail with a new spec, no. In near future, I just want people to use the new (exnref) Wasm EH via `-fwasm-exceptions` and call it just "Wasm EH", and refer to the old one by "Wasm legacy EH".

aheejin:
> `can we keep the name EMTEST_SKIP_EH?`
> 
> Separating the environment variables to skip tests that verify legacy Wasm EH versus Exnref Wasm EH are important, since those were introduced to browsers at very different times. That allows testing legacy Wasm EH in browsers that do not have Exnref Wasm EH support.

What I meant was to use `EMTEST_SKIP_EH` to mean skipping the exnref EH, in the same vein that I wanted to use just "Wasm EH" to refer to the exnref EH. But yeah, I agree this can be confusing and this changes what it used to mean, and given that this is not a user-facing option, I guess using exnref here can be fine.

tools/feature_matrix.py
aheejin: Not sure if I understand. That code just means in Wasm EH (meaning exnref EH here), we turn off legacy EH. And that `require_` function runs only to check whether the current environment meets the Wasm EH requirements in the tests. But it looks this `apply_min_browser_versions` is used in `phase_linker_setup` the linking process. What this will do is, even when Wasm EH is not being used and there's no `-fwasm-exceptions`, this will set the minimum browser versions to that of the Wasm (exnref) EH, no?

hoodmane:
Since Node 24 doesn't support the new eh without an experimental flag and Node 26 isn't out yet, it seems to me that the legacy eh should be the default at least until Node 26 is out and probably a bit longer after that.

tools/feature_matrix.py
juj: I do see Wasm Exnref exceptions being enabled in a build that does not pass `-fwasm-exceptions`, which is what is confusing me.

But that looks like some kind of a bug instead in SDL2_image. Posted https://github.com/emscripten-core/emscripten/issues/25430 about that.

I understand that the intent is that `-fwasm-exceptions -sWASM_LEGACY_EXCEPTIONS=1` should target Wasm Legacy EH, and `-fwasm-exceptions -sWASM_LEGACY_EXCEPTIONS=0` should target Wasm Exnref EH. Updated the code to match that.

juj:
> we'd like to at least make the legacy EH a rarely used old special case we keep only for the compatibility's sake, even if we are not successful in completely deprecating it from the web.

I hope that web will not remove Wasm Legacy EH, because all Unity output targets it today. There is a lot of content with Wasm Legacy EH in the wild out there.

I was excited about roadmapping the work of migrating Unity to use Wasm Exnref EH, but after seeing https://github.com/emscripten-core/emscripten/issues/25365 , my excitement fell considerably. It looks like Unity will be on the Wasm Legacy EH for quite a while from now.

juj:
> I wanted to just call the new EH without any qualifiers, while calling the old EH "legacy EH"

> I just want people to use the new (exnref) Wasm EH via `-fwasm-exceptions` and call it just "Wasm EH", and refer to the old one by "Wasm legacy EH".

> What I meant was to use `EMTEST_SKIP_EH` to mean skipping the exnref EH

Ok, updated this PR to not use the term exnref, but just refer to Wasm exception handling.

Should I open a ticket over to WebAssembly/website to adjust the [WebAssembly Roadmap](https://webassembly.org/features/) to change "Exception Handling with exnref" to "Exception Handling"?

test/common.py
```diff
@@ -1197,6 +1197,9 @@ def require_simd(self):
       self.fail('either d8 or node >= 16 required to run wasm64 tests.  Use EMTEST_SKIP_SIMD to skip')
 
   def require_wasm_legacy_eh(self):
+    if 'EMTEST_SKIP_WASM_LEGACY_EH' in os.environ:
+      self.skipTest('test requires node >= 17 or d8 (and EMTEST_SKIP_WASM_LEGACY_EH is set)')
```
juj, line 1134: @aheejin now I realize my root source of the confusion. What was happening is that the semantics of `EMTEST_SKIP_EH=1` was to not actually skip the test, if the node.js shell was detected to support the target EH.

I would like to change the semantics to mean that if user passes `EMTEST_SKIP_EH=1` (or `EMTEST_SKIP_WASM_LEGACY_EH=1`), then do skip the tests, independent of Node.js version. (the tests are running in the browser, so Node.js version does not matter)

So what was happening all the time on my CI tests was that I was passing `EMTEST_SKIP_EH=1` and `EMTEST_SKIP_WASM_LEGACY_EH=1` and running in Firefox 91, which doesn't support either; but the tests would not be skipped because I am running on Node.js 24, and they would then proceed to fail in the old browser.

tools/feature_matrix.py
aheejin: For the tests, `-fwasm-exceptions` is enabled here: https://github.com/emscripten-core/emscripten/blob/0d529eee690168f00a7fe826f8eebacb5f124048/test/common.py#L800

test/common.py
aheejin: I agree that's confusing. I actually wasn't aware that we don't skip the tests even if we have `EMTEST_SKIP_***` set if the node version is high enough. But it looks this has been this way for years and it doesn't only apply to Wasm EH. Other methods like `require_simd` and `require_wasm64` have the same problem. I think it would be better to change it in a separate PR for all `require_***`s if we are to change this. @sbc100 What do you think about changing the meaning of `EMTEST_SKIP_***`? Is there a reason that we don't honor these skip variables when the node version is high?

aheejin:
> Should I open a ticket over to WebAssembly/website to adjust the [WebAssembly Roadmap](https://webassembly.org/features/) to change "Exception Handling with exnref" to "Exception Handling"?

Hmm, not sure. I don't mind either way, but people in CG might think that's confusing. (Which you might think applies to the users here with the same rationale, which I don't disagree completely, but I somehow feel I just don't want to expose "exnref" to day-to-day users.)

test/common.py
sbc100: Seems reasonable to always skip yes.

I guess the original meaning was more like "EMTEST_ALLOW_SKIPPING_OF_XXX_IF_NEEDED", but don't see why it shouldn't be "EMTEST_ALWAYS_SKIP_XX".

sbc100: I agree we that if we make this change it should be in its own PR.

juj: https://github.com/emscripten-core/emscripten/pull/25445

.circleci/config.yml
```diff
@@ -1226,6 +1227,7 @@ jobs:
       # We don't install d8 or modern node on the mac runner so we skip any
       # tests that depend on those.
       EMTEST_SKIP_V8: "1"
+      EMTEST_SKIP_WASM_LEGACY_EH: "1"
       EMTEST_SKIP_EH: "1"
```
aheejin: Sorry for the confusion, but if we want to skip Wasm EH (exnref), maybe this should be `EMTEST_SKIP_WASM_EH` (And this has been used to only skip Wasm EH so far)

juj:
Updated this PR on top of main.

test/common.py
```diff
@@ -1192,7 +1194,7 @@ def require_jspi(self):
       self.v8_args += exp_args
       return
 
-    self.fail('either d8 or node v24 required to run JSPI tests.  Use EMTEST_SKIP_JSPI to skip')
+      self.fail('either d8 or node v24 required to run JSPI tests.  Use EMTEST_SKIP_JSPI to skip')
```
aheejin: Was this meant to be included?

tools/feature_matrix.py
```diff
@@ -199,3 +221,8 @@ def apply_min_browser_versions():
     enable_feature(Feature.WORKER_ES6_MODULES, 'EXPORT_ES6 with -sWASM_WORKERS')
   if settings.OFFSCREENCANVAS_SUPPORT:
     enable_feature(Feature.OFFSCREENCANVAS_SUPPORT, 'OFFSCREENCANVAS_SUPPORT')
+  if settings.WASM_EXCEPTIONS or settings.SUPPORT_LONGJMP == 'wasm': # Wasm longjmp support will lean on Wasm (Legacy) EH
```
aheejin, line 224: The code looks fine, but Wasm longjmp support does not particularly require the legacy EH. It should work fine with both legacy and non-legacy EH.

.circleci/config.yml
juj: > Sorry for the confusion, but if we want to skip Wasm EH (exnref), maybe this should be `EMTEST_SKIP_WASM_EH`

Oh, I renamed it back to `EMTEST_SKIP_EH` because [you previously reviewed](https://github.com/emscripten-core/emscripten/pull/25423#issuecomment-3349551171) that you wanted to use `EMTEST_SKIP_EH=1` to do this: *"What I meant was to use EMTEST_SKIP_EH to mean skipping the exnref EH"*

> (And this has been used to only skip Wasm EH so far)

I don't think that is the case. `EMTEST_SKIP_EH` would skip both Legacy Wasm EH and Exnref Wasm EH tests the same before this PR. This PR is separating these two with two different skip environment variables.

test/common.py
juj: No, that was a bad git automerge. Cleaned up.

tools/feature_matrix.py
juj: Hmm, how do you mean? Is the comment wrong, or? Should I update to something else?

.circleci/config.yml
aheejin: > > Sorry for the confusion, but if we want to skip Wasm EH (exnref), maybe this should be `EMTEST_SKIP_WASM_EH`
> 
> Oh, I renamed it back to `EMTEST_SKIP_EH` because [you previously reviewed](https://github.com/emscripten-core/emscripten/pull/25423#issuecomment-3349551171) that you wanted to use `EMTEST_SKIP_EH=1` to do this: _"What I meant was to use EMTEST_SKIP_EH to mean skipping the exnref EH"_

Sorry I should've asked to rename it to `EMTEST_SKIP_WASM_EH`. What I was saying was mostly that I didn't want the name 'exnref' in there. Given that legacy / exnref differs only in whether it's legacy or not, I think `WASM` should be in both of them.

> > (And this has been used to only skip Wasm EH so far)
> 
> I don't think that is the case. `EMTEST_SKIP_EH` would skip both Legacy Wasm EH and Exnref Wasm EH tests the same before this PR. This PR is separating these two with two different skip environment variables.

Sorry for the confusion again. What I meant by "Wasm EH" was not exnref but just the Wasm proposal-using EH in general, i.e., not "Emscripten EH" that's simulated by JS. And currently `EMTEST_SKIP_EH` is used to skip both kinds of Wasm EH (legacy and exnref), but not Emscripten EH:
https://github.com/emscripten-core/emscripten/blob/7b36fc888a4bc8edf8c4abf783b14b45ec46ffab/test/common.py#L1132-L1165

juj: Ok, I renamed `EMTEST_SKIP_EH` to `EMTEST_SKIP_WASM_EH`. How does it look now?

tools/feature_matrix.py
aheejin: Oh sorry, I missed the parentheses, and read that sentence as "Wasm longjmp support will lean on Wasm legacy EH". So I said "No Wasm longjmp supports works fine with exnref too". Now I understand what you meant.

aheejin (approved):
Thanks!

Issue #24850, opened 2025-08-01
Proposal: Require C++17 for embind


brendandahl:
I would like to propose that we consider moving the minimum required C++ standard for the embind library to C++17.

Embind's implementation makes heavy use of C++ templates. While it currently supports C++11, there have been significant improvements in C++17 that are particularly beneficial for template metaprogramming. I think these new features could simplify the existing embind codebase, making it more readable, easier to maintain, and potentially simpler for new contributors to understand.

I wanted to start a discussion to see if there is anyone in the community who still requires C++11 support for their projects using embind and would be adversely affected by this change.

Looking forward to your feedback.

sbc100:
Perhaps we could start by injecting a preprocessor `#warning` into `embind.h` that warns if C++ < 17 is uses (with a link to this bug).

sbc100:
Nobody on the mailing list had concerns so I suggest we move forward with this!

PR #25424, opened 2025-09-29
Add warning about usage of embind without C++17

sbc100:
See #24850


brendandahl:
I think we need val.h updated too. Alternatively, add the asserts in wire.h which is included in both.

sbc100:
> I think we need val.h updated too. Alternatively, add the asserts in wire.h which is included in both.

Done!

PR #25426, opened 2025-09-29
Skip whole browser test, not just the execution part

juj:
I am running browser test suites several times, with

```
set EMCC_CFLAGS=-sMIN_FIREFOX_VERSION=xyz -Wno-unused-command-line-argument
```
but even when I also set the environment variables
```
set EMTEST_LACKS_OFFSCREEN_CANVAS=1
set EMTEST_LACKS_ES6_WORKERS=1
```
to skip the relevant ES6 or OffscreenCanvas tests that the Firefox browser under test does not support, the test will actually not skip, but still produce a test failure.

This is because the `skipExecIf()` function used to build the `@requires_offscreen_canvas` and `@requires_es6_workers` decorators will still attempt to compile the test, and only skip running it in the browser.

But when I have set e.g. `EMCC_CFLAGS=-sMIN_FIREFOX_VERSION=95`, then that Firefox does not support OffscreenCanvas, and it does not help even if I also set the env. var. `EMTEST_LACKS_OFFSCREEN_CANVAS=1`. The failing compilation still happens, and the test fails.

This PR fixes that issue by making EMTEST_LACKS_OFFSCREEN_CANVAS and EMTEST_LACKS_ES6_WORKERS skip the whole test, and not just the browser execution part.

Though this does have the drawback that CircleCI coverage might be reduced, e.g. for the OffscreenCanvas tests parts, if it was seen valuable to still compile those on targets that don't support OffscreenCanvas. But is that important anymore?

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -190,21 +190,6 @@ def skipExecIf(cond, message):
   return decorator
 
 
-def skipTestIf(cond, message):
-  def decorator(f):
-    assert callable(f)
-
-    @wraps(f)
-    def decorated(self, *args, **kwargs):
-      if cond:
-        self.skipTest(message)
-      f(self, *args, **kwargs)
-
-    return decorated
-
-  return decorator
-
-
 def webgl2_disabled():
   return os.getenv('EMTEST_LACKS_WEBGL2') or os.getenv('EMTEST_LACKS_GRAPHICS_HARDWARE')
 
@@ -217,8 +202,8 @@ requires_graphics_hardware = skipExecIf(os.getenv('EMTEST_LACKS_GRAPHICS_HARDWAR
 requires_webgl2 = unittest.skipIf(webgl2_disabled(), "This test requires WebGL2 to be available")
 requires_webgpu = unittest.skipIf(webgpu_disabled(), "This test requires WebGPU to be available")
 requires_sound_hardware = skipExecIf(os.getenv('EMTEST_LACKS_SOUND_HARDWARE'), 'This test requires sound hardware')
-requires_offscreen_canvas = skipTestIf(os.getenv('EMTEST_LACKS_OFFSCREEN_CANVAS'), 'This test requires a browser with OffscreenCanvas')
-requires_es6_workers = skipTestIf(os.getenv('EMTEST_LACKS_ES6_WORKERS'), 'This test requires a browser with ES6 Module Workers support')
+requires_offscreen_canvas = unittest.skipIf(os.getenv('EMTEST_LACKS_OFFSCREEN_CANVAS'), 'This test requires a browser with OffscreenCanvas')
+requires_es6_workers = unittest.skipIf(os.getenv('EMTEST_LACKS_ES6_WORKERS'), 'This test requires a browser with ES6 Module Workers support')
 
 
 class browser(BrowserCore):


sbc100:
Yes, the idea was that because we don't have the ability the run browser offscreen canvas tests in headless mode (i.e. in CI) then we should at least test that they compile.

juj:
Another way to fix this might be to use a separate flag `EMTEST_SKIP_BUILDING_UNSUPPORTED_TESTS=1` or similar, to turn the `skipTestIf()` decorator to skipping immediately, rather than waiting for test run time?

sbc100:
Actually I think maybe we do have compile coverage for the test because its only firefox where we set `EMTEST_LACKS_OFFSCREEN_CANVAS`?   At least thats what it looks like to me?

BTW do you know if firefox still doesn't support this?    The comment says "# OffscreenCanvas support is not yet done in Firefox.".

sbc100:
Lets see if it works: https://github.com/emscripten-core/emscripten/pull/25427

juj:
> BTW do you know if firefox still doesn't support this? The comment says "# OffscreenCanvas support is not yet done in Firefox.".

There is one recurring Firefox error that it prints in some (but not all) OffscreenCanvas tests:

```
NotSupportedError: Cannot transfer OffscreenCanvas bound to element using captureStream.
```
this appears in the following tests, which I currently skip on my runs:

```
test_webgl_offscreen_canvas_in_pthread
test_webgl_offscreen_canvas_in_pthread_chained
test_webgl_resize_offscreencanvas_from_main_thread
test_webgl_resize_offscreencanvas_from_main_thread_proxy
test_webgl_resize_offscreencanvas_from_main_thread_offscreen_proxy
test_webgl_resize_offscreencanvas_from_main_thread_offscreen_blocking_proxy
```
Haven't had a moment to triage this further yet.

juj:
Besides that quirk, I run without `EMTEST_LACKS_OFFSCREEN_CANVAS` on Firefox >= 105, and they do pass the OffscreenCanvas tests:

image

On Firefox < 105 I want to skip OffscreenCanvas tests, but that doesn't currently work (hence posting this PR)

juj:
Another way to resolve this would be to do

```diff
diff --git a/test/test_browser.py b/test/test_browser.py
index 914fc9b05..5264bc12a 100644
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -183,6 +183,10 @@ def skipExecIf(cond, message):
     def decorated(self, *args, **kwargs):
       if cond:
         self.skip_exec = message
+        # If user does not want to attempt building an unsupported test, then
+        # skip the test immediately.
+        if os.getenv('EMTEST_BUILD_UNSUPPORTED_TESTS', '1') == '0':
+          self.skipTest(message)
       f(self, *args, **kwargs)

     return decorated
```
and then I can set env. var. `EMTEST_BUILD_UNSUPPORTED_TESTS=0` in my CI runs on old browsers?

juj:
Or one more way (maybe more fixed to this specific use case) to resolve the issue might be to do

```diff
diff --git a/test/test_browser.py b/test/test_browser.py
@@ -183,6 +184,10 @@ def skipExecIf(cond, message):
     def decorated(self, *args, **kwargs):
       if cond:
         self.skip_exec = message
+        # If user is explicitly targeting an old browser, then skip the test
+        # here (to avoid emcc compilation from possibly failing to compile)
+        if bool(re.search(r"MIN_.*?_VERSION", os.getenv('EMCC_CFLAGS', ''))):
+          self.skipTest(message)
       f(self, *args, **kwargs)

     return decorated
```

juj:
LMK if one of these schemes would be preferable?

PR #25427, opened 2025-09-29
[CI] Remove EMTEST_LACKS_OFFSCREEN_CANVAS from firefox config. NFC

sbc100:
Apparently firefox has supported this since v105.


brendandahl (approved):

sbc100:
I guess because we also set `EMTEST_LACKS_GRAPHICS_HARDWARE` on firefox that this probably doesn't actually do anything?

PR #25428, opened 2025-09-30
Add @requires_offscreen_canvas annotation

juj:
Add @requires_offscreen_canvas annotation to browser.test_emscripten_animate_canvas_element_size_offscreen. This is over-conservative, but keeping test code compact for now.


sbc100 (commented):
I suppose this went unnoticed because we run firefox tests with `EMTEST_LACKS_GRAPHICS_HARDWARE` in CI (because headless mode I believe)

PR #25432, opened 2025-09-30
Fix wasm table access when accessing functions in splitted module

Mintyboi:
When accessing functions from the splitted module, there's an error because the wasmtable is using table64. So we would have to access it with a BigInt. This has been updated in other part of the codebase, but we missed this file.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -12849,6 +12849,8 @@ exec "$@"
                       '--export-prefix=%', 'test_split_module.wasm.orig', '-o1', 'primary.wasm', '-o2', 'secondary.wasm', '--profile=profile.data']
     if jspi:
       wasm_split_run += ['--jspi', '--enable-reference-types']
+    if self.get_setting('MEMORY64'):
+      wasm_split_run += ['--enable-memory64']
     self.run_process(wasm_split_run)
 
     os.remove('test_split_module.wasm')


src/preamble.js
```diff
@@ -503,9 +503,9 @@ var splitModuleProxyHandler = {
       // When the table is dynamically laid out, the placeholder functions names
       // are offsets from the table base. In the main module, the table base is
       // always 1.
-      return wasmTable.get(1 + parseInt(prop))(...args);
+      return wasmTable.get(toIndexType(1 + parseInt(prop)))(...args);
 #else
-      return wasmTable.get(prop)(...args);
+      return wasmTable.get(toIndexType(prop))(...args);
```
sbc100: This should be in `{{{` `}}}.  e.g. `wasmTable.get({{{ toIndexType('prop') }}})`.

Can you refactor so that there is s single `return wasmTable.get` statement?

i.e.:

```
#if RELOCATABLE
   prop = parseInt(prop) + 1
#endif
   return wasmTable.get(...)
```

src/preamble.js
```diff
@@ -503,10 +503,9 @@ var splitModuleProxyHandler = {
       // When the table is dynamically laid out, the placeholder functions names
       // are offsets from the table base. In the main module, the table base is
       // always 1.
-      return wasmTable.get(toIndexType(1 + parseInt(prop)))(...args);
-#else
-      return wasmTable.get(toIndexType(prop))(...args);
+      prop = 1 + parseInt(prop)
```
sbc100: nit: add a semicolon at end of line.

Mintyboi:
>I guess this means we don't have any test coverage for wasm64 + module splitting?

This seems to be the case.
Else it would have failed without this fix as well https://github.com/WebAssembly/binaryen/pull/7904

sbc100:
lgtm with the test update

sbc100:
Can you update the test ?

Mintyboi:
> Can you update the test ?

Sure, do you mean increasing the coverage of the `test_split_module` test to include wasm64 scenarios?

Mintyboi:
While working on split_module + mem64 in my application, I notice more changes required on top of this PR. I'm working on resolving those issues as well

sbc100:
Yes, I was hoping you can add `@also_with_wasm64` to `test_split_module`

Mintyboi:
@sbc100 Do you have suggestions on how I can tweak [test_split_module.post.js](https://github.com/Mintyboi/emscripten/blob/fix-split-module-load/test/other/test_split_module.post.js#L4) such that it'll work for both wasm32 and wasm64.
e.g.
- For wasm32, we use `__write_profile(0, 0);`
- For wasm64, we use `__write_profile(0n, 0);`

Could I perhaps create another file? i.e. `test_split_module_wasm64.post.js`

sbc100:
> @sbc100 Do you have suggestions on how I can tweak [test_split_module.post.js](https://github.com/Mintyboi/emscripten/blob/fix-split-module-load/test/other/test_split_module.post.js#L4) such that it'll work for both wasm32 and wasm64. e.g.
> 
> * For wasm32, we use `__write_profile(0, 0);`
> * For wasm64, we use `__write_profile(0n, 0);`
> 
> Could I perhaps create another file? i.e. `test_split_module_wasm64.post.js`

You can use the `toIndexType` macro.   e.g `__write_profile({{{ toIndexType(0) }}}, 0);`

sbc100:
Actually maybe `to64` is more appropriate this this case.

Mintyboi:
> __write_profile({{{ toIndexType(0) }}}, 0);

I've tried this actually, but the macro block does not expanded in the final output. Does this file go through the JS preprocessor?
I also tried using only the function as well, but `toIndexType` is not defined in the output `out\test\test_split_module.js` file

Mintyboi:
Update: I noticed that you have made a commit to handle this.
Thanks!

One more thing, I looked at the CI logs and it seems like we are not running wasm64 tests
`test_split_module_wasm64 (test_other.other) ... skipped 'test requires node >= 24 or d8 (and EMTEST_SKIP_WASM64 is set)'`

sbc100:
Yup, see https://github.com/emscripten-core/emscripten/pull/25531, which also includes a fix to run the wasm64 other tests

PR #25435, opened 2025-09-30
Add test for bat file quoting issue. NFC

sbc100:
See https://github.com/microsoft/terminal/issues/15212

This test exposes a bug, which is fixed in #25416

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -15079,9 +15079,32 @@ addToLibrary({
 
   @crossplatform
   def test_no_extra_output(self):
+    # Run this build command first to warm the cache (since caching can produce stdout).
     self.run_process([EMCC, '-c', test_file('hello_world.c')])
     output = self.run_process([EMCC, '-c', test_file('hello_world.c')], stdout=PIPE, stderr=STDOUT).stdout
     self.assertEqual(output, '')
+
+  @crossplatform
+  def test_shell_cmd_with_quotes(self):
+    file = test_file('hello_world.c')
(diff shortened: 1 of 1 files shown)


test/test_other.py
```diff
@@ -15079,10 +15079,33 @@ def test_embind_negative_enum_values(self):
...
     self.assertEqual(output, '')
 
+  @crossplatform
+  def test_shell_cmd_with_quotes(self):
+    file = test_file('hello_world.c')
+    # Run this build command first to warm the cache (since caching can produce stdout).
+    self.run_process([EMCC, '-c', file])
+
+    # Verify that "emcc" is found in the PATH correctly in shell scripts (shell=True).
+    # Specificaly when quotes are around the "emcc" command itself.
```
```suggestion
    # (specifically when quotes are around the "emcc" command itself)
```

test/test_other.py
```diff
@@ -15079,10 +15079,33 @@ def test_embind_negative_enum_values(self):
...
+    # See https://github.com/microsoft/terminal/issues/15212
+    old_path = os.environ['PATH']
+    new_path = old_path + os.pathsep + os.path.dirname(EMCC)
+    cmd = f'"emcc" -c "{file}"'
+    print('running cmd:', cmd)
+    with env_modify({'PATH': new_path}):
+      proc = subprocess.run(cmd, capture_output=True, text=True, shell=True, check=True)
+      # There is currently a bug in the windows .bar files that leads to stdout
```
```suggestion
      # There is currently a bug in the windows .bat files that leads to stdout
```

PR #25439, opened 2025-09-30
Fix browser.test_sdl2_canvas_proxy on Firefox

juj:
SDL2 renderer init attempts to compile a shader with GLES external image content. This causes Firefox to report a shader compilation error, which would cause WebGL proxying to assert().

Demote the assert() into a warning print to let Firefox pass through.

Also because async proxied code cannot get the shader info log, print the info log to the browser console when compile status is queried. This way developers will see all shader compile errors automatically in proxied GL mode.

### DIFF
--- a/src/webGLClient.js
+++ b/src/webGLClient.js
@@ -229,7 +229,12 @@ function WebGLClient() {
     40: { name: 'texImage2D', func: func9 },
     41: { name: 'compressedTexImage2D', func: func7 },
     42: { name: 'activeTexture', func: activeTexture },
-    43: { name: 'getShaderParameter', func: () => assert(ctx.getShaderParameter(objects[buffer[i++]], buffer[i++]), 'we cannot handle errors, we are async proxied WebGL') },
+    43: { name: 'getShaderParameter', func: () => {
+      // Log shader compilation errors as warnings in the console log, since we cannot get the compilation logs back to the proxied Worker in any other way.
+      var object = objects[buffer[i++]];
+      var param = buffer[i++];
+      if (!ctx.getShaderParameter(object, param) && param == ctx.COMPILE_STATUS) console.warn(`Shader compilation failed: ${ctx.getShaderInfoLog(object)}`);
+    } },
     44: { name: 'clearDepth', func: func1 },
     45: { name: 'depthFunc', func: func1 },
     46: { name: 'frontFace', func: func1 },


src/webGLClient.js
```diff
@@ -229,7 +229,12 @@ function WebGLClient() {
     40: { name: 'texImage2D', func: func9 },
     41: { name: 'compressedTexImage2D', func: func7 },
     42: { name: 'activeTexture', func: activeTexture },
-    43: { name: 'getShaderParameter', func: () => assert(ctx.getShaderParameter(objects[buffer[i++]], buffer[i++]), 'we cannot handle errors, we are async proxied WebGL') },
+    43: { name: 'getShaderParameter', func: () => {
+      // Log shader compilation errors as warnings in the console log, since we cannot get the compilation logs back to the proxied Worker in any other way.
+      var object = objects[buffer[i++]];
+      var param = buffer[i++];
+      if (!ctx.getShaderParameter(object, param) && param == ctx.COMPILE_STATUS) console.warn(`Shader compilation failed: ${ctx.getShaderInfoLog(object)}`);
```
sbc100, line 236: Should we keep the "we cannot handle errors, we are async proxied", or something like it?

juj: Maybe the message `// since we cannot get the compilation logs back to the proxied Worker in any other way.` covers it?

sbc100: sgtm

juj:
I wouldn't oppose removing `--proxy-to-worker`. It is a very old implementation by now, and quite WebGL-centric, and monolithic framework/runtime-esque thing.

I think if there are users of the `--proxy-to-worker` philosophy (i.e. don't assume SharedArrayBuffer, and only async postMessage() proxy what you can), then it would benefit from being re-cast into a build mode where the individual JS library functions themselves would manage the proxying in a DCEable way, using a smaller set of common runtime bits.

juj:
I wonder if @kripken knows of any users of `--proxy-to-worker` anymore? Iirc. it was originally written to satisfy the cube renderer way way back.

kripken:
I'm not sure how many people use `--proxy-to-worker` atm. I remember it was nice for e.g. making Doom on the Web much more responsive, with just a flag flip.

Github finds various uses, but not sure how many are active:

https://github.com/search?q=%22--proxy-to-worker%22&type=code

sbc100:
Lets continue that discussion on #25440.

Yes, obviously if there are users out there relying on this then we don't need to go through with the removal.  This goes for all the flags / features that we hope to remove.  It about removing **unused** stuff.

PR #19046, opened 2023-03-23
jsifier: Make JS libraries aliases work even when native export exists

sbc100:
Followup to #19033.

Updating the `test_closure_full_js_library` to include webgl2 caused a lot of `JSC_REFERENCE_BEFORE_DECLARE` closure errors.

This is because we had JS aliases who's target was defined both in JS and in native code.  In this case the jsifier was choosing not to include the JS version, which in the case of the GL symbols not what we want.  I imagine this is why, prior to #19033, we were duplicating the library entries.

This this change JS symbol aliases now force the alias target to be included from the JS library, even in the case that the native code also emits its own version of a symbol.

This happens in the case of pthreads + MAIN_MODULE + OFFSCREEN_FRAMEBUFFER. In this mode webgl2.c defines (and also exports) the entire set of GL library symbols.  In this case we still want the `emscripten_XX` aliases to point to the original JS versions, not the exported native version.

Sadly it looks like we have at testing gap for that particular combination.

PR #25441, opened 2025-09-30
[js-compiler] Remove special handling for JS library aliases. NFC

sbc100:
This change remove some special casing in `addFromLibrary` which
simplifies the code, and also lays the groundwork for #25441 which
explicitly allows JS symbol to alias native ones.

This special handling was added in #19046 in order to deal with case
where `glXXX` symbols were simultaneously aliased by `emscripten_glXX`
and exported (in the case of MAIN_MODULE=1).

However, this work is no longer needed since in #21785 we stopped
stopped included symbol exported due to `MAIN_MODULE=1` as part of
`WASM_EXPORTS` (the symbols are only exported to side modules, not to
JS).

In original problem was that `emscripten_glXX` should always points to
the original JS implementation, even if a native version is exported.
Just in case somebody is explicitly exporting these symbols this change
also flips the aliases such that the `glXX` functions are now aliases of
`emscripten_glXX` and not the other way around.  This means that
exporting `glXXX` will replace/override the alias, but will not effect
`emscripten_glXX`.

### DIFF
--- a/src/lib/libwebgl.js
+++ b/src/lib/libwebgl.js
@@ -4346,17 +4346,37 @@ createGLPassthroughFunctions(LibraryGL, glPassthroughFuncs);
 
 autoAddDeps(LibraryGL, '$GL');
 
+function renameSymbol(lib, oldName, newName) {
+  lib[newName] = lib[oldName];
+  delete lib[oldName];
+  for (const suffix of decoratorSuffixes) {
+    const oldDecorator = oldName + suffix;
+    if (lib.hasOwnProperty(oldDecorator)) {
+      const newDecorator = newName + suffix;
+      lib[newDecorator] = lib[oldDecorator];
+      delete lib[oldDecorator];
+    }
+  }
+}
+
 function recordGLProcAddressGet(lib) {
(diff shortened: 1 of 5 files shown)


sbc100:
Sorry for the long explanation, its quite a simple fix for a rather complex problem.

src/lib/libwebgl.js
```diff
@@ -4346,17 +4346,37 @@ createGLPassthroughFunctions(LibraryGL, glPassthroughFuncs);
...
+}
+
 function recordGLProcAddressGet(lib) {
   // GL proc address retrieval - allow access through glX and emscripten_glX, to
   // allow name collisions with user-implemented things having the same name
   // (see gl.c)
+  //
+  // We do this by renaming `glX` symbols to `emscripten_glX` and then setting
+  // `glX` as an alias of `emscripten_glX`.  The reason for this renaming is to
+  // ensure that `emscripten_glX` is always available, even in cases were native
```
kripken:
```suggestion
  // ensure that `emscripten_glX` is always available, even in cases where native
```

kripken (approved):

Issue #25440, opened 2025-09-30
Can we deprecate `--proxy-to-worker` / `-sPROXY_TO_WORKER`?


sbc100:
I don't think they are going to be many folks using this.

IIUC it was mostly and experiment from the early days of emscripten to try to get synchronous applications running in a worker. However, since it always takes over the whole browser window I can't imagine many folks actually using it.

The process for deprecating and removing features from emscripten is laid out out in https://github.com/emscripten-core/emscripten/blob/main/docs/process.md#deprecating-settings-and-features).

At any point in the process we may decide to delay or abort the deprecation and close this issue.

Deprecation checklist:

- [x] Open issue (#25440)
- [x] Perform global search for usage in public repos
- [x] Send message to emscripten-discuss (https://groups.google.com/g/emscripten-discuss/c/oFCpwLbDLvU)
- [x] Add "deprecation" warning to emcc (#25442)
- [x] Wait at least 4 releases or 2 months (deprecated in 4.0.16 on October 1st 2025)
- [ ] Remove feature

sbc100:
See https://github.com/emscripten-core/emscripten/pull/25439

juj:
The use case of proxying to Worker itself is sound, because just enabling multithreaded programs require nonstandard HTTP server configuration.

So applications that are based on async message passing multiprocessing-style of communication could definitely utilize proxying to Worker.

It is just that the current implementation of `--proxy-to-worker` is monolithic, which makes it a bit dusty to keep up with the times.

sbc100:
I think we should only consider keeping it around if there are active users of the feature.    If there are none, I don't see any point in maintaining it and continuing to fix bugs in it.

sbc100:
I performed the following global github search: https://github.com/search?q=%2F%28%3F-i%29%5CbsPROXY_TO_WORKER%5Cb%2F+-org%3Aemscripten-core+-path%3Aemcc.*+-path%3Asettings.*+-path%3Asettings_reference.*+-path%3Acommon.py+-path%3Ashell.js+-path%3A*wasm_worker*&type=code&p=5

None of the result, as far as I can tell, were actual live usages of the setting.

sbc100:
[...]

PR #25442, opened 2025-10-01
Mark PROXY_TO_WORKER as "under consideration for removal"

sbc100:
See #25440

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -3559,8 +3559,8 @@ More info: https://emscripten.org
     self.cflags += ['-lembind', '--emit-tsd', 'embind_tsgen.d.ts']
     # Passing -sWASM_WORKERS or -sPROXY_TO_WORKER requires the 'worker' environment
     # at link time. Verify that TS binding generation still works in this case.
-    for flag in ('-sWASM_WORKERS', '-sPROXY_TO_WORKER'):
-      self.emcc(test_file('other/embind_tsgen.cpp'), [flag])
+    for flags in (['-sWASM_WORKERS'], ['-sPROXY_TO_WORKER', '-Wno-deprecated']):
+      self.emcc(test_file('other/embind_tsgen.cpp'), flags)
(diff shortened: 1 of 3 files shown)


kripken:
Maybe let's wait on the mailing list discussion? A github search did show uses of the API.

sbc100:
> Maybe let's wait on the mailing list discussion? A github search did show uses of the API.

This PR just generates a warning so that users can report their usage.  It doesn't actually stop anyone using it.

kripken:
Can the warning be more conditional, then? Rather than `removal pending`, the text could be `ongoing discussion regarding removal` - ? I'm just trying to avoid this looking like the decision is already set in stone.

sbc100:
> Can the warning be more conditional, then? Rather than `removal pending`, the text could be `ongoing discussion regarding removal` - ? I'm just trying to avoid this looking like the decision is already set in stone.

Yes, I already tried to make to more like that with the change from "scheduled" to "pending".  I guess the better word would be "proposed", or "nominated" for removal?

sbc100:
I rephrased as "under consideration for removal"

test/test_other.py
```diff
@@ -3559,8 +3559,8 @@ def test_embind_tsgen_worker_env(self):
     self.cflags += ['-lembind', '--emit-tsd', 'embind_tsgen.d.ts']
     # Passing -sWASM_WORKERS or -sPROXY_TO_WORKER requires the 'worker' environment
     # at link time. Verify that TS binding generation still works in this case.
-    for flag in ('-sWASM_WORKERS', '-sPROXY_TO_WORKER'):
-      self.emcc(test_file('other/embind_tsgen.cpp'), [flag])
+    for flags in (['-sWASM_WORKERS'], ['-sPROXY_TO_WORKER', '-Wno-deprecated']):
```
kripken, line 3562: Why is there a mix of experiment/deprecated in the flags in this PR?

kripken (approved):
Text looks good, thanks.

lgtm % question

test/test_other.py
sbc100: Doh..!  my mistake.

PR #25447, opened 2025-10-01
Cleanup helper functions in libwebgl.js. NFC

sbc100:


src/lib/libwebgl.js
```diff
@@ -4352,15 +4350,16 @@ function recordGLProcAddressGet(lib) {
   // GL proc address retrieval - allow access through glX and emscripten_glX, to
   // allow name collisions with user-implemented things having the same name
   // (see gl.c)
-  Object.keys(lib).forEach((x) => {
-    if (x.startsWith('gl') && !isDecorator(x)) {
-      lib['emscripten_' + x] = x;
-      var sig = LibraryManager.library[x + '__sig'];
+  for (const sym of Object.keys(lib)) {
+    if (sym.startsWith('gl') && !isDecorator(sym)) {
+      const alias = `emscripten_${sym}`
```
kripken:
```suggestion
      const alias = 'emscripten_' + sym;
```
This seems simpler, and is more consistent with the code below that does ` + `.

sbc100: Fair enough.

PR #25449, opened 2025-10-01
Add some test coverage for threads+OFFSCREEN_FRAMEBUFFER+MAIN_MODULE=1. NFC

sbc100:
See https://github.com/emscripten-core/emscripten/pull/19046

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -1936,6 +1936,7 @@ simulateKeyUp(100, undefined, 'Numpad4');
   @parameterized({
     '': ([],),
     'pthreads': (['-pthread', '-sPROXY_TO_PTHREAD', '-sOFFSCREEN_FRAMEBUFFER'],),
+    'pthreads_main_module': (['-pthread', '-sPROXY_TO_PTHREAD', '-sOFFSCREEN_FRAMEBUFFER', '-sMAIN_MODULE', '-Wno-experimental'],),
   })
   @requires_graphics_hardware
   def test_gl_textures(self, args):
--- a/tools/link.py
+++ b/tools/link.py
@@ -3090,7 +3090,14 @@ def run(options, linker_args):
     logger.debug('stopping after linking to object file')
     return 0
 
-  linker_args += phase_calculate_system_libraries(options)
+  system_libs = phase_calculate_system_libraries(options)
+  # Only add system libraries that have not already been specified.
+  # This avoids issues where the user explictly includes, for example, `-lGL`.
+  # This is not normally a problem except in the case of -sMAIN_MODULE=1 where
+  # the duplicate library would result in duplicate symbols.
+  for s in system_libs:
+    if s not in linker_args:
+      linker_args.append(s)
 
   js_syms = {}
   if (not settings.SIDE_MODULE or settings.ASYNCIFY) and not shared.SKIP_SUBPROCS:


kripken (approved):

juj:
Remarkably, the new test combination added in this PR fails consistently in Firefox, except for Firefox Nightly:

image

The reason for this is that `Atomics.waitAsync()` was not implemented in Firefox, until very recently, [15 days ago](https://bugzilla.mozilla.org/show_bug.cgi?id=1884148).

We have a polyfill for Atomics.waitAsync in https://github.com/emscripten-core/emscripten/blob/7125c45fabecd64741af308a1d1059b99971a2bc/src/lib/libatomic.js#L10-L70 . The fact that the new added test combo fails in Firefox, seems to suggest that something is not up-to-par in our polyfill.

sbc100:
Oh man, that could be a pain to debug.

I guess the polyfill must be in use a in a lot of places in Firefox already though right?  Basically all pthread tests depend on this right?   So it must be some strange edge case that is being hit here, in code that works most of the time?

juj:
It will have to be a very specific condition when the Atomics.waitAsync() wait comes into play. It is not used in general: I never added an use of it into our system libraries that I could remember. (because I considered it to be just broken and misdesigned at the time) I vaguely recall reading through @tlively 's mailbox/proxying queue impl. that maybe it might have had an use, but can't find one now either.

sbc100:
> It will have to be a very specific condition when the Atomics.waitAsync() wait comes into play. It is not used in general: I never added an use of it into our system libraries that I could remember. (because I considered it to be just broken and misdesigned at the time) I vaguely recall reading through @tlively 's mailbox/proxying queue impl. that maybe it might have had an use, but can't find one now either.

Yes, its used in `_emscripten_thread_mailbox_await` which is used in basically all pthread-based programs to wake up threads when they have work to do.  IIUC.

sbc100:
I guess it only happens when pthreads exit to their event loop, which might be be too common?

tlively:
We only use `Atomics.waitAsync` when it is [available](https://github.com/emscripten-core/emscripten/blob/bd4676aa00785a6a1fad4a54a168a7b594540d10/src/lib/libpthread.js#L1246-L1258) and otherwise [fall back](https://github.com/emscripten-core/emscripten/blob/bd4676aa00785a6a1fad4a54a168a7b594540d10/system/lib/pthread/thread_mailbox.c#L108-L112) to using postMessage. I suppose that code probably tries to use the polyfill, although it might be better if it didn't.

sbc100:
> We only use `Atomics.waitAsync` when it is [available](https://github.com/emscripten-core/emscripten/blob/bd4676aa00785a6a1fad4a54a168a7b594540d10/src/lib/libpthread.js#L1246-L1258) and otherwise [fall back](https://github.com/emscripten-core/emscripten/blob/bd4676aa00785a6a1fad4a54a168a7b594540d10/system/lib/pthread/thread_mailbox.c#L108-L112) to using postMessage. I suppose that code probably tries to use the polyfill, although it might be better if it didn't.

Ah I see, so maybe the polyfill is only being injected under certain circumstances?

juj:
The polyfill is always being injected if the target browser does not support Atomics.waitAsync().

The issue is that in order to be 100% correct, it is not possible to polyfill Atomics.waitAsync() without also monkey-patching Atomics.notify(), to provide a separate channel for notifying the emulated waitAsync() waiters.

To avoid that issue, I implemented the polyfill waitAsync() as a setTimeout() loop that polls the waitAsync() target memory location, and notifies waitAsync() when the memory location changes value.

This matches the idea of futex wait "sleep this thread as long as this memory location has its current value X." And it works as long as the notifier will change the value of the memory location before notifying. That is the pattern that is used when implementing the common mutexes/semaphores/barries etc.

Reading the way that atomics.wait/.notify is being used by the mailbox implementation, it seems that it uses futex sleep/wake as a signal. When the wakeup notification is done, the memory location value is not changed, but the location is just notified. The polyfill waitAsync() cannot detect this as being a notification, so it never wakes up the waitAsync() waiter.

In general notifying without first changing the value is racy when locks/semaphores/barriers are concerned - changing the value helps implement the atomicity needed in the sleep+wakeup., But this is probably not an issue in the mailbox implementation.

To resolve this problem, three possible ways: remove the waitAsync polyfill, or restrict the polyfill to not cover the mailbox (essentially mailbox has its own polyfill), or change the mailbox impl. to also change the memory location value before waking up the waiter.

Issue #25415, opened 2025-09-29
Babel does not trigger for --proxy-to-worker


juj:
```
emcc.bat test\hello_world.c -sMIN_FIREFOX_VERSION=65 --proxy-to-worker -o test.html
```

and run in Firefox 65, results in

<img width="905" height="164" alt="Image" src="https://github.com/user-attachments/assets/ad83c64a-d9d6-4f11-a113-bb5f57256884" />

The `?.` constructs have not been babeled:

<img width="1098" height="150" alt="Image" src="https://github.com/user-attachments/assets/34803412-dca0-4b98-aebb-6080e255aae1" />

Remove `--proxy-to-worker` and the issue vanishes.

sbc100:
Can we remove the `--proxy-to-worker` maybe?  My understanding is that it has very limited applications since it takes over the whole page, right?

juj:
Do you mean we'd remove the whole `--proxy-to-worker` feature?

I am under the impression that there would be active users of that feature still.

To me this issue seems to be some intermediate .js file plumbing or something like that, so maybe not fundamental enough to require disabling `--proxy-to-worker` altogether?

sbc100:
> Do you mean we'd remove the whole `--proxy-to-worker` feature?
> 
> I am under the impression that there would be active users of that feature still.

Yes, I'm basically asking if there are any known active users?   Are you using it anywhere?

juj:
I am not using `--proxy-to-worker` anywhere.

sbc100:
Looks like we can remove the whole feature: https://github.com/emscripten-core/emscripten/issues/25440.  No comments yet on this.

PR #25451, opened 2025-10-01
Avoid ES6 syntax in places that are not Babeled. #25415

juj:


src/proxyClient.js
```diff
@@ -132,8 +132,7 @@ var frameId = 0;
 
 // Worker
 
-var filename;
-filename ||= '<<< filename >>>';
```
juj, line 136: Was this old code depending on `filename` being possibly populated beforehand, i.e. having two definitions of `var filename;`?

I tried searching the codebase for where the prior definition could come from - but if we did, then Closure would complain of two declarations of `var filename;` ?

... so in the absence of seeing any use, I opted to simplify. Maybe test suite will show a use of this.

PR #25459, opened 2025-10-02
Separate out multithreading tests in browser harness

juj:
Annotate browser tests that require SAB by attribute, or renaming test to have word 'thread' in it.

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -191,6 +191,10 @@ requires_webgpu = unittest.skipIf(webgpu_disabled(), "This test requires WebGPU 
 requires_sound_hardware = skipExecIf(os.getenv('EMTEST_LACKS_SOUND_HARDWARE'), 'This test requires sound hardware')
 requires_offscreen_canvas = unittest.skipIf(os.getenv('EMTEST_LACKS_OFFSCREEN_CANVAS'), 'This test requires a browser with OffscreenCanvas')
 requires_es6_workers = unittest.skipIf(os.getenv('EMTEST_LACKS_ES6_WORKERS'), 'This test requires a browser with ES6 Module Workers support')
+# N.b. not all SharedArrayBuffer requiring tests are annotated with this decorator, since at this point there are so many of such tests.
+# As a middle ground, if a test has a name 'thread' or 'wasm_worker' in it, then it does not need decorating. To run all single-threaded tests in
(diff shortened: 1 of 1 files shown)


test/codesize/test_codesize_file_preload.expected.js
```diff
@@ -32,7 +32,7 @@ var ENVIRONMENT_IS_NODE = globalThis.process?.versions?.node && globalThis.proce
 // --pre-jses are emitted after the Module integration code, so that they can
 // refer to Module (if they choose; they can also define Module)
 // include: <FILENAME REPLACED>
-Module["expectedDataFileDownloads"] ??= 0;
+if (!Module["expectedDataFileDownloads"]) Module["expectedDataFileDownloads"] = 0;
```
juj: If we don't want to remove these `??` expressions, we can alternatively bump minimum Firefox version up to Firefox 79 (July 28, 2020). That would be ok to me.

juj: Actually, let me do that instead.

juj:
Closing in favor of https://github.com/emscripten-core/emscripten/pull/25469

juj:
Reopening given the feedback in https://github.com/emscripten-core/emscripten/pull/25467#issuecomment-3362073560

sbc100 (commented):
The actually non-test change here seems specific to file_packager.  Perhaps you could mention that in the PR title?   Its only not specific to FF, right?  How about something like:  "Avoid ES6 constructs in file_packager output"?

Its kind of a shame the file packager output often *is* covered by babel since its embedded in the generated code, but we also have the standalone mode.

Can you added a standalone file packager test to the `test-node-compat` and confirm that it fails before and after this PR with the oldest support node version (12.22.9).

juj:
> Can you added a standalone file packager test to the `test-node-compat` and confirm that it fails before and after this PR with the oldest support node version (12.22.9).

I don't have Node.js 12, and working on ensuring the browser suite works on Node.js 12 seems like a separate pile of work. I'm verifying these PRs on Firefox atm, and that is already enough to get my table full.

sbc100:
> > Can you added a standalone file packager test to the `test-node-compat` and confirm that it fails before and after this PR with the oldest support node version (12.22.9).
> 
> I don't have Node.js 12, and working on ensuring the browser suite works on Node.js 12 seems like a separate pile of work. I'm verifying these PRs on Firefox atm, and that is already enough to get my table full.

Ok, I'll split out of the file_packager stuff maybe, and you can land the FF browser test stuff here.

juj:
Ok, updated with the browser suite reorganization bits.

sbc100 (approved):

PR #25461, opened 2025-10-02
[testing] Lower FPS assertion in hello_world_gles.c

brendandahl:
This test seems to flake a lot from this assertion. The fps on my machine often is only 20 fps, so I think when there are multiple browsers running the fps will be pretty low on CI.


sbc100 (approved):
Maybe add a comment here about why we are being so conservative (i.e. mutliple browsers).

PR #25467, opened 2025-10-02
Bump the minimum Firefox version to 79.

juj:
Bump the minimum Firefox version to 79. Emscripten unconditionally now depends on some ES6 syntax availability, so Emscripten output no longer runs in older Firefoxes.

I did go through the exercise of fixing Emscripten to be able to support Firefox 68ESR, which looks like this: https://github.com/emscripten-core/emscripten/compare/main...juj:emscripten:firefox_68esr_support

I can alternatively post that as a PR, and work to fix Firefox support down to 65. But I get an impression that bumping the version might be preferred.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -27,6 +27,8 @@ See docs/process.md for more on how version tagging works.
   - Node: v10.19.0 -> v12.22.9
   - Chrome: v70 -> v74
   - Firefox: v55 -> v65
+- Minimum Firefox version was further bumped up to Firefox 79, to reflect that
+  Emscripten now requires ES6 constructs (#25467)
 - The Embind `val` functions `call`, `operator()`, and `new_` now support
   passing `pointer`s by using the `allow_raw_pointers()` argument. This feature
   is only enabled with C++17 and newer. Older versions will allow pointers by
--- a/site/source/docs/tools_reference/settings_reference.rst
+++ b/site/source/docs/tools_reference/settings_reference.rst
@@ -2871,7 +2871,7 @@ are desired to work. Pass -sMIN_FIREFOX_VERSION=majorVersion to drop support
 for Firefox versions older than < majorVersion.
 Firefox 79 was released on 2020-07-28.
 MAX_INT (0x7FFFFFFF, or -1) specifies that target is not supported.
-Minimum supported value is 65 which was released on 2019-01-29 (see
+Minimum supported value is 79 which was released on 2020-07-28 (see
 feature_matrix.py)
 
 Default value: 79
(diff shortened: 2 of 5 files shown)


sbc100:
Do we not have other targets that don't have ES6 support?  What about our min node version:  12.22.9?

juj:
> We have support for transpiling ES6 features away, right? If we bump this version do we no longer need the trasnpilation stuff? Can we just remove it? Was it broken for some reason?

We have, but it is unfortunately spotty with its coverage. E.g. file packager and .html shells do not get it.

My expectation is that ES6 will not be the only time in the timeline of the web where new language features will be introduced. So having Babel around might be useful as a tool to opt-in to new features in the future?

Though when such a future happens, we will have to be more careful about extending Babel coverage to file packager and other parts as well.

sbc100:
It seems like if we really want to move to ES6 we would need to bump a lot more versions:

```
  if settings.POLYFILL:                                                                                   
    # Emscripten requires certain ES6+ constructs by default in library code                              
    # - (various ES6 operators available in all browsers listed below)                                    
    # - https://caniuse.com/mdn-javascript_operators_nullish_coalescing:                                  
    #                                          FF:72 CHROME:80 SAFARI:13.1 NODE:14                        
    # - https://caniuse.com/mdn-javascript_operators_optional_chaining:                                   
    #                                          FF:74 CHROME:80 SAFARI:13.1 NODE:14                        
    # - https://caniuse.com/mdn-javascript_operators_logical_or_assignment:                               
    #                                          FF:79 CHROME:85 SAFARI:14 NODE:16                          
    # Taking the highest requirements gives is our minimum:                                               
    #                             Max Version: FF:79 CHROME:85 SAFARI:14 NODE:16                          
    # TODO: replace this with feature matrix in the future.                                               
    settings.TRANSPILE = (settings.MIN_FIREFOX_VERSION < 79 or                                            
                          settings.MIN_CHROME_VERSION < 85 or                                             
                          settings.MIN_SAFARI_VERSION < 140000 or                                         
                          settings.MIN_NODE_VERSION < 160000)     
```

Basically every single one of those would need updating.

Given that, I don't think bumping the FF version alone here makes sense and I think removing ES6 constructs from the non-babel places is probably the best short term solution (along with better test coverage).  

I think `test-node-compat` in `.circleci/config.yml` is the best way to ensure we have coverage here.  There is a list of tests there that we run under the oldest supported node version.

juj:
> Given that, I don't think bumping the FF version alone here makes sense and I think removing ES6 constructs from the non-babel places is probably the best short term solution (along with better test coverage).

Ok, I will remove the non-Babel constructs from the code and update with Firefox 65 coverage.

juj:
> Do we not have other targets that don't have ES6 support? What about our min node version: 12.22.9?

I don't know, I have had cycles to investigate Firefox only so far. Google suggests `"Node.js 12 doesn't understand ES6 syntax like optional chaining, export & import."`

PR #25469, opened 2025-10-02
[test] Add annotations to help run in Firefox 79. NFC

juj:


test/test_browser.py
```diff
@@ -5517,7 +5520,7 @@ def test_audio_worklet_emscripten_locks(self):
   @requires_sound_hardware
   @also_with_minimal_runtime
   def test_web_audio_context_sample_rate(self):
-    self.btest_exit('webaudio/audio_context_sample_rate.c', cflags=['-sAUDIO_WORKLET', '-sWASM_WORKERS'])
+    self.btest_exit('webaudio/audio_context_sample_rate.c', cflags=['-lwebaudio.js'])
```
juj, line 5523: This test only creates a Web Audio context, so does not depend on Audio Worklets.

PR #25471, opened 2025-10-02
Fix -sMIN_NODE_VERSION=-1 -sENVIRONMENT=web to build without errors.

juj:
Users might "double-disable" Node.js support by setting both `-sENVIRONMENT=web` (no Node.js), and `-sMIN_NODE_VERSION=-1` (set Node to unsupported).

But Emscripten complained about this not being allowed. Let this combination pass.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -179,7 +179,7 @@ def setup_environment_settings():
   settings.ENVIRONMENT_MAY_BE_WORKER = not settings.ENVIRONMENT or 'worker' in settings.ENVIRONMENT
 
   if not settings.ENVIRONMENT_MAY_BE_NODE:
-    if 'MIN_NODE_VERSION' in user_settings:
+    if 'MIN_NODE_VERSION' in user_settings and user_settings['MIN_NODE_VERSION'] != '0x7FFFFFFF':
       diagnostics.warning('unused-command-line-argument', 'ignoring MIN_NODE_VERSION because `node` environment is not enabled')
     settings.MIN_NODE_VERSION = feature_matrix.UNSUPPORTED
 


sbc100 (commented):
Specifying `MIN_NODE_VERSION` when node is not even supported, is a kind of configuration error of sorts I think. 

As a user I think I would like to know that I can remove the `-sMIN_NODE_VERSION` flag completely in this case.   

Also, this is just a warning that is easy enough to suppress, right?

tools/link.py
```diff
@@ -179,7 +179,7 @@ def setup_environment_settings():
   settings.ENVIRONMENT_MAY_BE_WORKER = not settings.ENVIRONMENT or 'worker' in settings.ENVIRONMENT
 
   if not settings.ENVIRONMENT_MAY_BE_NODE:
-    if 'MIN_NODE_VERSION' in user_settings:
+    if 'MIN_NODE_VERSION' in user_settings and user_settings['MIN_NODE_VERSION'] != '0x7FFFFFFF':
```
sbc100: I think this can be `settings.MIN_NODE_VERSION != feature_matrix.UNSUPPORTED`

sbc100 (approved):
Presumably we should do that same for the web environment down below on line 190?

tools/link.py
juj: check

juj:
Yeah, makes sense. Updated and added a test.

sbc100:
Personally I would like to still see a warning here, since "double doing" something seems redundant is something that, as a user, I might want to fix in my build system.

But if you disagree, I don't feel too strongly.

juj:
I think this is very benign, and it does allow improving coverage of the test suite.

sbc100:
> I think this is very benign, and it does allow improving coverage of the test suite.

You do also have the option to add `-Wno-unused-command-line-argument` to your bot config.

sbc100:
> > I think this is very benign, and it does allow improving coverage of the test suite.
> 
> You do also have the option to add `-Wno-unused-command-line-argument` to your bot config.

Then injecting flags using `EMCC_CFLAGS` is seems like a good idea anyway because you don't know the context into which you are injecting your flags.

PR #25465, opened 2025-10-02
BigInt was introduced in Firefox 78, not Firefox 68.

juj:
Firefox 77 verified to fail e.g. `browser.test_utf8_textdecoder` with

image

```
======================================================================
FAIL: test_utf8_textdecoder (test_browser.browser.test_utf8_textdecoder)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\emsdk\python\3.13.3_64bit\Lib\unittest\case.py", line 58, in testPartExecutor
    yield
  File "C:\emsdk\python\3.13.3_64bit\Lib\unittest\case.py", line 651, in run
    self._callTestMethod(testMethod)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "C:\emsdk\python\3.13.3_64bit\Lib\unittest\case.py", line 606, in _callTestMethod
    if method() is not None:
       ~~~~~~^^
  File "C:\emsdk\emscripten\main\test\common.py", line 960, in resulting_test
    return func(self, *args)
  File "C:\emsdk\emscripten\main\test\test_browser.py", line 157, in decorated
    f(self, *args, **kwargs)
    ~^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\emsdk\emscripten\main\test\test_browser.py", line 4231, in test_utf8_textdecoder
    self.btest_exit('benchmark/benchmark_utf8.c', 0, cflags=['--embed-file', test_file('utf8_corpus.txt') + '@/utf8_corpus.txt'])
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\emsdk\emscripten\main\test\common.py", line 2769, in btest_exit
    return self.btest(filename, *args, **kwargs)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\emsdk\emscripten\main\test\common.py", line 2798, in btest
    self.run_browser(outfile, expected=['/report_result?' + e for e in expected], timeout=timeout)
[...]

PR #5011, opened 2017-03-09
Fix wasm loading in workers, and refactor async promise code

kripken:
Fixes #5005. Specifically, this

 * Adds readBinary support in web workers, as they can do sync binary xhrs after all, so that fixes #5005.
 * Refactor a getBinaryPromise method that gets a promise, no matter which platform we are on. In particular, on a worker (which lacks the Fetch API), it can use that readBinary method.

PR #25473, opened 2025-10-02
Remove Node.js self.run_js() test in browser suite.

juj:
Remove Node.js self.run_js() test in browser suite. The browser suite should test browser behavior only.

I am running the browser suite with

```
EMCC_CFLAGS= -sMIN_CHROME_VERSION=-1 -sMIN_SAFARI_VERSION=-1 -sMIN_NODE_VERSION=-1
```

to verify that targeting Firefox only should correctly result in a build that works in Firefox.

The test `test_hello_world_worker` is the only one that failed. Which happens because that test also verifies behavior in node.js.

Reading the test logic in that test, it seems to verify the file packager and `emscripten_run_script()` logic. That logic should be well covered in the `core` test suites and the `other` test suite already, so this seems redundant.

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -1605,12 +1605,6 @@ simulateKeyUp(100, undefined, 'Numpad4');
     self.assertExists('worker.js')
     self.run_browser('main.html', '/report_result?hello from worker, and :' + ('data for w' if file_data else '') + ':')
 
-    # code should run standalone too
-    # To great memories >4gb we need the canary version of node
-    if self.is_4gb():
-      self.require_node_canary()
-    self.assertContained('you should not see this text when in a worker!', self.run_js('worker.js'))
-
   @no_wasmfs('https://github.com/emscripten-core/emscripten/issues/19608')
   def test_mmap_lazyfile(self):
     create_file('lazydata.dat', 'hello world')


sbc100:
Is there some particular reason you want to remove this?    Did it causes issues somehow?

juj:
The rationale is that it prevents testing the said configuration. And given it looks redundant, I thought rather simpler to remove it, than to have a manual test skip. Is there anything relevant that this code tests that isn't tested elsewhere?

sbc100:
> The rationale is that it prevents testing the said configuration

Can you explain this?

juj:
I am setting the environment variable

```
EMCC_CFLAGS= -sMIN_CHROME_VERSION=-1 -sMIN_SAFARI_VERSION=-1 -sMIN_NODE_VERSION=-1 -sMIN_FIREFOX_VERSION=<firefox_version_under_test>
```
before running `test/runner browser`

so that all the tests would focus on verifying that MIN_FIREFOX_VERSION is accurate to the target Firefox version.

But because the one test `browser.test_hello_world_worker` also tries verifying behavior in Node.js, then adding those CFLAGS results in generating a build that doesn't run in Node, but aborts because that was not in the target environment list.

That one line is the only outlier in the whole `browser` test harness that attempts to test Node.js behavior, rest of the suite passes.

sbc100:
Oh I think I see.  We set `EMTEST_SKIP_NODE_CANARY: "1"` when we run the 4gb browser tests to avoid this.  See `test-browser-chrome-wasm64-4gb` in `.circleci/config.yml`.

Maybe you could do that same in your CI?  Or if not then maybe also remove the `EMTEST_SKIP_NODE_CANARY: "1"` from `config.yml` as part of this PR?

juj:
I get this issue in all the `browser` suites, not just the 4GB suite.

juj:
I do already pass `EMTEST_SKIP_NODE_CANARY=1` in the affected browser runs. (Maybe there will be more of these if that is not enabled?)

sbc100:
> -sMIN_NODE_VERSION=-1

Do you need to inject this flag?  It seems like that would mean you are testing a slightly different configuration than the one the test suite it setup for, no?    Why not just leave `MIN_NODE_VERSION` to its default value?

juj:
The idea is to improve test coverage by setting flags that only target Firefox. It should be a valid combination of configuration for end users to pass, so "why are you testing that. Don't do it" doesn't seem a good resolution.

sbc100:
Is specifying `-sMIN_NODE_VERSION=-1` the same as specifying a `-sENVIRONMENT` that doesn't contain node at all?  IIUC it is?    That seems like you would be testing a configuration that is more then just "support a specific FF version", its also testing the that whole browser test suite can be built with `-sENVIRONMENT=web` only, which is maybe useful but maybe separate to the support for older FF versions.

juj:
> Is specifying `-sMIN_NODE_VERSION=-1` the same as specifying a `-sENVIRONMENT` that doesn't contain node at all?

No, because there is also `-sENVIRONMENT=shell`. (or I guess maybe)

The reason I am only enabling Firefox in the test run is to make sure that supporting some other environment does not accidentally mask and enable a feature, that would hide a bug with the minimum version levels for Firefox (like e.g. in #25465)

sbc100 (commented):
Removing this seems reasonable on brief inspection yes

I didn't write this test though so I'm not sure exactly what its trying to verify.  It looks like it goes back to #5011, or even before.

@kripken WDTY?

sbc100:
> > Is specifying `-sMIN_NODE_VERSION=-1` the same as specifying a `-sENVIRONMENT` that doesn't contain node at all?
> 
> No, because there is also `-sENVIRONMENT=shell`. (or I guess maybe)

I'm not sure I understand.  `-sENVIRONMENT=shell` does not support node, right?  I think `shell` only refers to things like v8/spidermonkey/jsc.

juj:
> `-sENVIRONMENT=shell` does not support node, right?

Yeah that's what I was confused about.. I was not sure if it includes node or not.. but probably not.

juj:
@kripken would this be ok to land? The deleted block of code is an oddity in the browser suite: it verifies behavior in the Node.js shell rather than the browser.

To my eyes, the deleted code amounts to verifying that

```
[EMCC, test_file('hello_world_worker.c'), '-o', 'worker.js', '--preload-file', 'file.dat'] + self.get_cflags()
self.run_process(cmd)
self.assertExists('worker.js')
self.assertContained('you should not see this text when in a worker!', self.run_js('worker.js'))
```
works. That doesn't seem like anything special/relevant? We do have a lot of tests for --preload-file in the `other` suite, for example `other.test_include_file`?

By deleting that code, I can run the `browser` suite in a special mode with

```
EMCC_CFLAGS= -sMIN_FIREFOX_VERSION=143 -sMIN_CHROME_VERSION=-1 -sMIN_SAFARI_VERSION=-1 -sMIN_NODE_VERSION=-1
```

sbc100 (commented):
Can you also remove the corresponding  `EMTEST_SKIP_NODE_CANARY` entries from the `test-browser-*` tasks in `.circleci/config.yml`?

juj:
Hmm how do you mean? I don't see a reference to `browser.test_hello_world_worker` in config.yml?

juj:
I'm not sure if this PR will necessarily allow removing `EMTEST_SKIP_NODE_CANARY`, that seems unrelated?

sbc100:
> I'm not sure if this PR will necessarily allow removing `EMTEST_SKIP_NODE_CANARY`, that seems unrelated?

We currently set `EMTEST_SKIP_NODE_CANARY` in `test-browser-chrome-wasm64-4gb` specifically because if we don't this test would fail since it has `require_node_canary` in it.

juj:
Oh I see, that was the only test remaining in browser suite that had the canary part.

sbc100 (approved):

PR #25475, opened 2025-10-02
Remove unused internal setting: BULK_MEMORY. NFC

sbc100:
It looks like the last usage of this setting was removed in #22873.

### DIFF
--- a/emcc.py
+++ b/emcc.py
@@ -431,9 +431,6 @@ def phase_setup(options, state):
   if settings.PTHREADS or settings.WASM_WORKERS:
     settings.SHARED_MEMORY = 1
 
-  if settings.SHARED_MEMORY:
-    settings.BULK_MEMORY = 1
-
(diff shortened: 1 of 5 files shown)


juj:
This probably would prevent https://github.com/emscripten-core/emscripten/pull/25466 from landing?

sbc100:
> This probably would prevent #25466 from landing?

I just noticed that setting was redundant and unused.

I think you can still write what you want with `if not feature_matrix.caniuse(feature_matrix.Feature.BULK_MEMORY):`?

juj:
> I think you can still write what you want with `if not feature_matrix.caniuse(feature_matrix.Feature.BULK_MEMORY):`?

Thanks, that worked. Updated the PR to use that, and added a test.

sbc100:
OK to land this now?

juj (approved):
LGTM

PR #25476, opened 2025-10-02
Remove ES6 constructs from file_packager output

sbc100:
This allows the generated code to run on older engines.

Unlike the majority of the emscripten-generated code the file packager output does not always get run through babel when targeting older engines.

Specifically tests the standalone file packages code on the oldest
support node version.   This also required removing the use of the
`fs/Promises` packages which is not present in older versions of node.

Split out from #25459

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -935,6 +935,7 @@ jobs:
             other.test_native_call_before_init
             other.test_js_optimizer_verbose
             other.test_node_unhandled_rejection
+            other.test_file_packager_separate_metadata
             other.test_full_js_library*
             core2.test_hello_world
             core2.test_fcntl_open_nodefs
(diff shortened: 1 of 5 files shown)


juj (approved):

kripken:
What about running babel on this code? That would let us write nicer code, maybe worth the cost?

sbc100:
> What about running babel on this code? That would let us write nicer code, maybe worth the cost?

That is certainly another option?  Would would need some kind of new flag to `file_packager` to tell to run babel.  e.g. `file_packager --legacy-browser-support ...` or `file_packager --please-run-babel-for-me`, which is certainly possible. 

Remember, this only applies to the standalone version of file_packager so we don't have any of the normal flags that emcc has .

sbc100:
Honestly, if you are passing `--please-run-babel-for-me` it seems easy enough to just do it yourself.

We could just say in the docs for file packager something like this:

```
File packager produces mordern ES6 javascript.  If you are targeting older/legacy browsers you will need to run a transpiler such as babel to lower it to ES5 before deploying it.
```

WDTY?  We could then just update the handfull of tests we have the use file_packager directly to run babel too as needed?

kripken:
Yeah, maybe it's not worth it. I don't feel strongly.

juj:
As a short term solution it is nice to just remove ES6, that way I am now unblocked to dive through other test failures in older Firefoxes.

As a long term solution, we can imagine that the JavaScript standard has not stopped adding new language features into it, so future ECMAScript features will produce new needs for transpiling. So from that perspective, at some point adding a built-in transpiling support to file packager output could be nice.

PR #25480, opened 2025-10-03
[jspi] Require async js functions when used with __async decorator.

### LINKED ISSUE #25468: Embind + MEMORY64: Cannot convert Promise to a BigInt
Embind Promise in combination with MEMORY64 fails to correctly bind Promise types.
```txt
Uncaught (in promise) SyntaxError: Cannot convert [object Promise] to a BigInt
```

emscripten 4.0.15
Chrome Canary Version 143.0.7447.0

main.cpp
```c++
// cat main.cpp
#include <iostream>
#include <functional>
#include <emscripten/bind.h>

class ITestAsync {
public:
    virtual ~ITestAsync() = default;
    virtual void PrintValue() = 0;
};

class ITestAsyncWrapper : public emscripten::wrapper<ITestAsync> {
public:
    EMSCRIPTEN_WRAPPER(ITestAsyncWrapper);
    ~ITestAsyncWrapper() override = default;
    void PrintValue() override {
        std::cout << "PrintValue resolving" << std::endl;
        call<emscripten::val>("PrintValue").await().as<void>();
        std::cout << "PrintValue resolved" << std::endl;
    }
};

class TestAsyncFactory {
public:
    void SetFactory(emscripten::val factory) {
        mFactory = [factory]() -> std::unique_ptr<ITestAsync> { return factory().as<std::unique_ptr<ITestAsync>>(); };
    };

    void Execute() {
        if (mFactory)
            mFactory()->PrintValue();
    };

private:
    std::function<std::unique_ptr<ITestAsync>()> mFactory;
};

EMSCRIPTEN_BINDINGS(AsyncWrapper) {
    using namespace emscripten;
    class_<ITestAsync>("ITestAsync")
        .function("PrintValue", &ITestAsync::PrintValue, pure_virtual(), async())
        .allow_subclass<ITestAsyncWrapper>("ITestAsyncWrapper");

    class_<TestAsyncFactory>("TestAsyncFactory")
        .constructor()
        .function("SetFactory", &TestAsyncFactory::SetFactory)
        .function("Execute", &TestAsyncFactory::Execute, async());
}
```

main.html
```HTML
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Minimal HTML</title>
</head>
<body>
<script lang="ts" src="AsyncFactory.js"></script>
<script>
    async function main() {
        Module = await Module();
        console.log("AsyncFactory loaded",Module)

        const TestAsyncClass = Module.ITestAsync.extend("ITestAsync", {
            PrintValue() {
                return new Promise((resolve) => {
                    setTimeout(() => {
                        console.log('TestAsyncClass PrintValue executed');
                        resolve();
                    }, 10);
                });
            },

        });

        const factory = new Module.TestAsyncFactory();
        factory.SetFactory(() => new TestAsyncClass());
        await factory.Execute();
        factory.delete();


    }
    main();

</script>
</body>
</html>
```

Build command:
```bash
 emcc main.cpp -o AsyncFactory.js -fPIC -sMODULARIZE=1 -lembind -sJSPI -sMEMORY64=1
```

Image

If `-sMEMORY64=1` flag is omitted, the above example works just fine.

```bash
 emcc main.cpp -o AsyncFactory.js -fPIC -sMODULARIZE=1 -lembind -sJSPI
```
Image

brendandahl:
The `_emval_await` library function is marked `_emval_await__async: true`, but the js function is not async. With memory64 enabled we auto convert to bigint and look for the async keyword (which is missing) to apply the await before creating the BigInt. With my changes __async will require an async js function, which signals the function is used with JSPI and the appropriate awaits are then inserted.

Fixes #25468

### DIFF
--- a/src/jsifier.mjs
+++ b/src/jsifier.mjs
@@ -9,6 +9,7 @@
 
 import assert from 'node:assert';
 import * as fs from 'node:fs/promises';
+import { isAsyncFunction } from 'node:util/types';
 import {
   ATMODULES,
   ATEXITS,
@@ -536,13 +537,13 @@ function(${args}) {
         deps.push('setTempRet0');
       }
 
-      let isAsyncFunction = false;
+      let hasAsyncDecorator = false;
       if (ASYNCIFY) {
         const original = LibraryManager.library[symbol];
         if (typeof original == 'function') {
-          isAsyncFunction = LibraryManager.library[symbol + '__async'];
+          hasAsyncDecorator = LibraryManager.library[symbol + '__async'];
         }
-        if (isAsyncFunction) {
+        if (hasAsyncDecorator) {
           asyncFuncs.push(symbol);
         }
       }
@@ -669,6 +670,10 @@ function(${args}) {
         addImplicitDeps(snippet, deps);
       } else if (typeof snippet == 'function') {
         isFunction = true;
+        if (ASYNCIFY == 2 && hasAsyncDecorator && !isAsyncFunction(snippet)) {
+          error(`'${symbol}' is marked with the __async decorator but is not an async JS function.`);
+        }
+
         snippet = processLibraryFunction(snippet, symbol, mangled, deps, isStub);
         addImplicitDeps(snippet, deps);
         if (CHECK_DEPS && !isUserSymbol) {
@@ -766,7 +771,7 @@ function(${args}) {
         }
         contentText += `\n${mangled}.sig = '${sig}';`;
       }
-      if (ASYNCIFY && isAsyncFunction) {
+      if (ASYNCIFY && hasAsyncDecorator) {
         contentText += `\n${mangled}.isAsync = true;`;
       }
       if (isStub) {
--- a/src/lib/libasync.js
+++ b/src/lib/libasync.js
@@ -476,11 +476,11 @@ addToLibrary({
 
   emscripten_sleep__deps: ['$safeSetTimeout'],
   emscripten_sleep__async: true,
-  emscripten_sleep: (ms) => Asyncify.handleSleep((wakeUp) => safeSetTimeout(wakeUp, ms)),
+  emscripten_sleep: async (ms) => Asyncify.handleSleep((wakeUp) => safeSetTimeout(wakeUp, ms)),
 
   emscripten_wget_data__deps: ['$asyncLoad', 'malloc'],
   emscripten_wget_data__async: true,
-  emscripten_wget_data: (url, pbuffer, pnum, perror) => Asyncify.handleAsync(async () => {
+  emscripten_wget_data: async (url, pbuffer, pnum, perror) => Asyncify.handleAsync(async () => {
     /* no need for run dependency, this is async but will not do any prepare etc. step */
     try {
       const byteArray = await asyncLoad(UTF8ToString(url));
@@ -497,7 +497,7 @@ addToLibrary({
 
   emscripten_scan_registers__deps: ['$safeSetTimeout'],
   emscripten_scan_registers__async: true,
-  emscripten_scan_registers: (func) => {
+  emscripten_scan_registers: async (func) => {
     return Asyncify.handleSleep((wakeUp) => {
       // We must first unwind, so things are spilled to the stack. Then while
       // we are pausing we do the actual scan. After that we can resume. Note
@@ -585,7 +585,7 @@ addToLibrary({
 
   emscripten_fiber_swap__deps: ["$Asyncify", "$Fibers", '$stackSave'],
   emscripten_fiber_swap__async: true,
-  emscripten_fiber_swap: (oldFiber, newFiber) => {
+  emscripten_fiber_swap: async (oldFiber, newFiber) => {
     if (ABORT) return;
 #if ASYNCIFY_DEBUG
     dbg('ASYNCIFY/FIBER: swap', oldFiber, '->', newFiber, 'state:', Asyncify.state);
--- a/src/lib/libemval.js
+++ b/src/lib/libemval.js
@@ -402,12 +402,20 @@ ${functionBody}
 #if ASYNCIFY
   _emval_await__deps: ['$Emval', '$Asyncify'],
   _emval_await__async: true,
+#if ASYNCIFY == 1
   _emval_await: (promise) => {
     return Asyncify.handleAsync(async () => {
       var value = await Emval.toValue(promise);
       return Emval.toHandle(value);
     });
   },
+#endif
+#if ASYNCIFY == 2
+  _emval_await: async (promise) => {
+    var value = await Emval.toValue(promise);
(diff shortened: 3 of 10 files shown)


src/jsifier.mjs
```diff
@@ -9,6 +9,7 @@
 
 import assert from 'node:assert';
 import * as fs from 'node:fs/promises';
+import { isAsyncFunction } from 'node:util/types';
```
sbc100, line 12: Oh.. interesting.  TIL about this.

src/lib/libemval.js
```diff
@@ -402,12 +402,20 @@ ${functionBody}
...
   _emval_await__async: true,
+#if ASYNCIFY == 1
   _emval_await: (promise) => {
     return Asyncify.handleAsync(async () => {
       var value = await Emval.toValue(promise);
       return Emval.toHandle(value);
     });
   },
+#endif
+#if ASYNCIFY == 2
```
sbc100: Can this just be `#else`  Since it can only be either 1 or 2 right?  Or `elif` would work too.

src/lib/libemval.js
```diff
@@ -402,12 +402,20 @@ ${functionBody}
...
+  _emval_await: async (promise) => {
+    var value = await Emval.toValue(promise);
+    return Emval.toHandle(value);
+  },
```
sbc100, line 415: I don't really get why this functions needs to looks so different in ASYNCIFY 1 vs 2.

In other places in this PR we have `async (...) => Asyncify.handleSleep(...`.. but not here?

src/lib/libsdl.js
```diff
@@ -1753,7 +1753,7 @@ var LibrarySDL = {
 #else
   SDL_Delay__deps: ['emscripten_sleep'],
   SDL_Delay__async: true,
-  SDL_Delay: (delay) => _emscripten_sleep(delay),
+  SDL_Delay: async (delay) => _emscripten_sleep(delay),
```
sbc100: I kind of which `#if ASYNCIFY` case came first here, but that unrelated to this PR

test/test_other.py
```diff
@@ -3246,6 +3246,7 @@ def test_embind_asyncify(self):
     '': [['-sDYNAMIC_EXECUTION=1']],
     'no_dynamic': [['-sDYNAMIC_EXECUTION=0']],
     'dyncall': [['-sALLOW_MEMORY_GROWTH', '-sMAXIMUM_MEMORY=4GB']],
+    'wasm64': (['-sMEMORY64'],),
   })
   @requires_jspi
```
sbc100, line 3301: How about `also_with_wasm64` so that we get all 4 cases in wasm64 mode?

test/test_other.py
```diff
@@ -3484,6 +3485,24 @@ def test_jspi_async_function(self):
...
+    ''')
+    create_file('main.c', r'''
+      #include <emscripten.h>
+      extern void foo();
+      EMSCRIPTEN_KEEPALIVE void test() {
+        foo();
+      }
+    ''')
+    err = self.expect_fail([EMCC, 'main.c', '-o', 'out.js', '-sJSPI', '--js-library=lib.js', '-Wno-experimental',])
+    self.assertContained('error: foo is marked with the __async decorator but is not an async JS function.', err)
```
sbc100: Maybe move this function alongside all the other `test_jslib_xx` functions?   Or maybe its more suited to being grouped with JSPI?  Not sure.

src/lib/libasync.js
```diff
@@ -476,11 +476,11 @@ addToLibrary({
 
   emscripten_sleep__deps: ['$safeSetTimeout'],
   emscripten_sleep__async: true,
-  emscripten_sleep: (ms) => Asyncify.handleSleep((wakeUp) => safeSetTimeout(wakeUp, ms)),
+  emscripten_sleep: async (ms) => Asyncify.handleSleep((wakeUp) => safeSetTimeout(wakeUp, ms)),
```
juj, line 485: What does flagging this `async` gain here? (it costs code size)

I understand the rationale for `async` is that one can `await` for the return value, but for functions that return void, isn't the `async` keyword dead code?

Though Asyncify and JSPI build modes itself do not need the functions to be flagged async do they? (in [WebGPU JSPI support I went with this](https://github.com/juj/wasm_webgpu/blob/3ed529956d1ee3f14f64c07769385e1929908f68/lib/lib_webgpu.js#L779-L806))

src/lib/libemval.js
brendandahl: Unfortunately, `handleAsync` currently works slightly differently on old asyncify and JSPI. On JSPI, it returns the promise and on asyncify it pipes the return value magically through the wakeup.  Adding the `async` keyword here for asyncify=1 causes the test to fail. 

I'd really like to unify all this code and have it so if you mark a library function with `__async` you don't have to use handleSleep or handleAsync, but that's a bigger change.

sbc100: SGTM

src/lib/libasync.js
brendandahl: It really doesn't add anything here, but consistency. We have a few spots where we auto modify the library js (e.g. memory 64). It's much easier in those wrapper functions to assume we're in a `async` function and emit `await` code (also the `await` code is usually shorter).

poretga99:
Can this PR proceed? We'd be more than happy if this landed in the main branch. We're currently patching released Emscripten manually.

curiousdannii:
It's perfectly legitimate to return a promise instead, and sometimes that would result in cleaner code. This should be at most a warning, not a requirement.

sbc100:
> It's perfectly legitimate to return a promise instead, and sometimes that would result in cleaner code. This should be at most a warning, not a requirement.

Right, but there is no harm in doing both, right?  `async` functions can do `return somepromise` just fine right?

curiousdannii:
I guess.. though I think it would be slightly lower performance (probably not noticeable unless it's a really hot function.)

Hmm. If you did require async functions to be AsyncFunctions, then could you do away with the __async decorator?

sbc100:
> I guess.. though I think it would be slightly lower performance (probably not noticeable unless it's a really hot function.)

According to AI there is not any extra overhead: Returning a promise from an async function in JavaScript does not create a second promise in the sense of a new, distinct promise object being generated on top of the one you explicitly return. Instead, the async function's inherent promise-returning nature integrates with the promise you return.

Take that with as many grains of salt as you like ..

sbc100:
Can we land the fix for `_emval_await` withing requiring that all `__async` functions are marked with the async keyword? 

That later sounds a little more controversial, and its also reasonable to write old school async functions that return a promise without the keyword right?  (we have some of those in the codebase already)

brendandahl:
Looks like this was fixed in the __async: auto pr.

Issue #25463, opened 2025-10-02
Hitting into compilation issue in windows when emsdk and emcache are in different drives.


shinji-TH:
Hi,
May I check if there are any specific reason for using the resolve function in https://github.com/emscripten-core/emscripten/blob/9c04c5379445de9de768f6e20ae92f21fa05fd79/tools/cache.py#L197

Reason for ask is I'm hitting into the follow error while compiling my project 
> 2025-10-02 02:07:54 ValueError: path is on mount 'U:', start on mount 'D:'

This was due to my emsdk resides under U:/ and em_cache folder(actual) reside under D:/
However I have set shortcut (symbolic links) under U:/refs/em_cache/ via the `EM_CACHE` environment. 
However it seems the resolve function return actual path instead of symbolic links path instead. 
Hence wanted to check if it is okay for us to set the cache via (symbolic links)

sbc100:
I don't remember why `.resolve()` was used there.  

It looks like it was first added back in #17102 as a replacement for `os.path.normpath`.   

Seems strange to me that `.resolve()` cannot resolve symlinks across drives?

sbc100:
It looks like one of the uses of `resolve()` that was added in #17102 was remove in #17353.  I think we can maybe remove this one too.  Still strange to me that windows would crash on symlink resolution across drives.

sbc100:
What version of python are you using?

shinji-TH:
@sbc100  Thanks for resolving the error!

Mintyboi:
>Still strange to me that windows would crash on symlink resolution across drives.

@sbc100 Just to clarify. Resolve will follow the symbolic / mapped paths to the original target. This part is ok.

The issue occurs when we attempt to find the relative path across 2 different drives in https://github.com/emscripten-core/emscripten/blob/89cf437529b0bfba8655bb9fa8628222097b1648/tools/system_libs.py#L528
which is not support according to the [documentation ](https://docs.python.org/3/library/os.path.html#os.path.relpath).

sbc100:
@Mintyboi are you running into this issue?

It looks like this is only when do compile the system libraries with batch inputs, so I guess we could just skipping the batching or use absolute paths in the case the we are on separate drives?

Mintyboi:
Yep, I'm in the same team as @shinji-TH and this is one of an issue we were facing in our builds whilst porting to windows.
I was attempting to clarify your earlier statement on symlink resolution.

[...]

PR #25484, opened 2025-10-03
Update cache to use absPath instead of Path.resolve

shinji-TH:
Encounter issue compiling Emscripten in windows when EM_CACHE was provided a shortcut path (symbolic link).

The original code `Path(config.CACHE).resolve()` returns the original cache path which is in a different drive from where the emscripten source is in. 

Hence in this PR, the `cachedir` was change to use `os.path.abspath(config.CACHE)` to allow the build process to access the cache via the symbolic link 

https://github.com/emscripten-core/emscripten/issues/25463


shinji-TH:
#25463 issue resolve with #25488

PR #25486, opened 2025-10-03
Add comment about minimum Firefox version required to run parallel test harness.

juj:
The new parallel Firefox test harness requires Firefox 67 or newer to run.

Firefox 66 and earlier suffer from two bugs:

1) When launching multiple browsers with

```
[C:\\emsdk\\firefox\\65.0_64bit\\firefox.exe, '-new-instance', '-profile', 'C:\\emsdk\\emscripten\\main\\out\\browser-profile-1', 'http://localhost:5555'
```

then if a previous instance of Firefox is running, Firefox will interpret the `-profile` argument as a URL to visit. (It will take that argument as profile path as well). This results in two tabs being opened in each browser instance (beyond the first):

image

but unfortunately given that the actual test URL is in the background, the test never progresses.

(This issue could be fixed by passing `-url` parameter, but given this is an ancient browser, and bug 2 also exists, not worth it to complicate code)

2) If a test fails in Firefox 66 or older, and a browser instance is being terminated, it looks like all browser instances get torn down, and not just the one that was hosting the failing test. So this would result in a lot of false positives in test failures.

Bisected that the fix to the above issues occurred somewhere between Firefox 66 and 67, in this bisection range: https://hg-edge.mozilla.org/mozilla-central/pushloghtml?fromchange=4ab143dde4dc3424cfedc74b3648fbf2e47fb7bf&tochange=af29567ecdba5441117b3f01ac2559a3473c3d6a

This PR adds a comment suggesting to set `EMTEST_CORES=1` if testing older Firefox versions.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -116,6 +116,9 @@ class ChromeConfig:
     """Chrome has no special configuration step."""
 
 
+# N.b. The earliest Firefox version that can run multithreaded browser harness
+# is Firefox 67. If you need to test earlier Firefox versions, use EMTEST_CORES=1
+# environment variable to revert to the single-threaded browser harness.
 class FirefoxConfig:
   data_dir_flag = '-profile '
   default_flags = ('-new-instance',)


sbc100 (commented):
Should we bump the `OLDEST_SUPPORTED_FIREFOX` from 65 to 67 to sidestep the issue?   Since its only 2 versions it seems unlikely anyone would care.

sbc100:
It looks like doing so would also allow us to remove a `#if MIN_FIREFOX_VERSION < 67 ` hack in `libwebgl.js`

sbc100 (approved):

juj:
> Should we bump the `OLDEST_SUPPORTED_FIREFOX` from 65 to 67 to sidestep the issue? Since its only 2 versions it seems unlikely anyone would care.

We could probably then bump to Firefox 68 since that is an ESR release. I have coverage set up for the ESR releases.

sbc100:
> > Should we bump the `OLDEST_SUPPORTED_FIREFOX` from 65 to 67 to sidestep the issue? Since its only 2 versions it seems unlikely anyone would care.
> 
> We could probably then bump to Firefox 68 since that is an ESR release. I have coverage set up for the ESR releases.

Yes!   Sounds good.  How do we find out which are ESR?  Maybe we could add a comment alongside `OLDEST_SUPPORTED_FIREFOX` for how to do that.

juj:
> How do we find out which are ESR?

Searching for `"esr/"` on this page https://archive.mozilla.org/pub/firefox/releases/ finds ESR releases.

PR #25487, opened 2025-10-03
Flip some negated `#if` checks. NFC

sbc100:
This helps with readability IMHO since you don't need to do a double negative when considering the `#else` block.


src/jsifier.mjs
```diff
@@ -601,15 +601,7 @@ function(${args}) {
             warn('To build in STANDALONE_WASM mode without a main(), use emcc --no-entry');
           }
         }
-        if (!RELOCATABLE) {
```
kripken, line 604: OTOH, "not relocatable == normal", and it's nice to have the normal path first? Then reading the more complex relocatable one after is easier.

src/lib/libemval.js
```diff
@@ -101,22 +101,7 @@ var LibraryEmVal = {
...
-    return Emval.toHandle(a);
-  },
-  _emval_array_to_memory_view__deps: ['$Emval'],
-  _emval_array_to_memory_view: (dst, src) => {
-    dst = Emval.toValue(dst);
-    src = Emval.toValue(src);
-    dst.set(src);
-  },
-#else
+#if SUPPORT_BIG_ENDIAN
```
kripken: Same here, and some others below - wdyt?

src/jsifier.mjs
sbc100: This one has annoyed me for a while.  Every time I read I feel like have to do mental work for some reason.

src/lib/libemval.js
sbc100: Sure I'm OK reverting this one.

src/jsifier.mjs
```diff
@@ -626,6 +618,14 @@ function(${args}) {
...
           LibraryManager.library[symbol] = new Function('...args', functionBody);
           isStub = true;
+        } else {
+          // emit a stub that will fail at runtime
+          LibraryManager.library[symbol] = new Function(`abort('missing function: ${symbol}');`);
+          // We have already warned/errored about this function, so for the purposes of Closure use, mute all type checks
+          // regarding this function, marking ot a variadic function that can take in anything and return anything.
+          // (not useful to warn/error multiple times)
+          LibraryManager.library[symbol + '__docs'] = '/** @type {function(...*):?} */';
+          isStub = true;
```
sbc100, line 628: I just noticed the `isStub = true;` is repeated in both condition here too

src/jsifier.mjs
kripken: If the negation here is annoying, how about `var ABSOLUTE = !RELOCATABLE` or some better name for "simple, absolute-addressed single file with no linking involved" - ?

(Perhaps "linked file", "relocated file", "flat file" - maybe there is a good term?)

That would avoid the negation, and also let us keep having the simple code first in each such if-else.

sbc100: Actually in this case I like the `if RELOCATABLE`.

The code here is part of the undefined symbol handling.   Its within a block and its already an error in most cases.

It reads like this:

```
if (is_undefined) {
   maybe_report_an_error();
   // Here we are continuing with the undefined symbol
   if (RELOCATABLE) {  // AKA dynamic linking
      create_a_dylink_stub();
   } else {
      create_an_abort_stub();
   } 
}
```

I think this is more logical that the other way around:

```
if (is_undefined) {
   maybe_report_an_error();
   // Here we are continuing with the undefined symbol
   if (!RELOCATABLE) {
      create_an_abort_stub();
   } else {
      create_a_dylink_stub();
   } 
}
```

sbc100:
Ok to land?

src/jsifier.mjs
kripken: Fair enough, I can live with this one. Reading the others again now.

kripken (approved):

PR #25494, opened 2025-10-04
Delete the Atomics.waitAsync() polyfill

juj:
Delete the Atomics.waitAsync() polyfill, which is not good enough to polyfill over the usage pattern in Emscripten mailbox implementation. See https://github.com/emscripten-core/emscripten/pull/25449#issuecomment-3358594240 . The mailbox implementation has its own postMessage() fallback, so does not need the polyfill. Fixes test browser.test_gl_textures_pthreads_main_module in Firefox."

This will be a breaking change to any users who successfully depended on the polyfill (i.e. only needed the partial support it brought), though it is uncertain if there are any.

Fixes test `browser.test_gl_textures_pthreads_main_module` on Firefox.

### DIFF
--- a/src/lib/libatomic.js
+++ b/src/lib/libatomic.js
@@ -7,68 +7,6 @@
 assert(SHARED_MEMORY);
 
 addToLibrary({
-// Chrome 87 shipped Atomics.waitAsync:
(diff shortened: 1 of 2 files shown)


tlively (approved):
Oh yeah, wow, that's totally broken.

PR #25495, opened 2025-10-04
Add colored print to test harness output

juj:
Another take at adding colors to test harness:

image

Tested to work on:
 - Windows 10 x64
 - Windows 11 ARM64
 - MacOS 15.5 Sequoia Apple M1
 - Linux Debian 13 Trixie ARM64

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -16,6 +16,7 @@ from tools import utils
 import common
 from common import errlog
 
+from tools.diagnostics import with_color, CYAN, GREEN, RED
 from tools.utils import WINDOWS
 
 
@@ -304,35 +305,41 @@ class BufferedParallelTestResult:
     with self.lock:
       val = f'[{int(self.progress_counter.value * 100 / self.num_tests)}%] '
       self.progress_counter.value += 1
-    return val
+    return with_color(CYAN, val)
 
   def addSuccess(self, test):
-    errlog(f'{self.compute_progress()}{test} ... ok ({self.calculateElapsed():.2f}s)')
+    msg = f'ok ({self.calculateElapsed():.2f}s)'
+    errlog(f'{self.compute_progress()}{test} ... {with_color(GREEN, msg)}')
     self.buffered_result = BufferedTestSuccess(test)
     self.test_result = 'success'
 
   def addExpectedFailure(self, test, err):
-    errlog(f'{self.compute_progress()}{test} ... expected failure ({self.calculateElapsed():.2f}s)')
+    msg = f'expected failure ({self.calculateElapsed():.2f}s)'
+    errlog(f'{self.compute_progress()}{test} ... {with_color(RED, msg)}')
     self.buffered_result = BufferedTestExpectedFailure(test, err)
     self.test_result = 'expected failure'
 
   def addUnexpectedSuccess(self, test):
-    errlog(f'{self.compute_progress()}{test} ... unexpected success ({self.calculateElapsed():.2f}s)')
+    msg = f'unexpected success ({self.calculateElapsed():.2f}s)'
+    errlog(f'{self.compute_progress()}{test} ... {with_color(RED, msg)}')
     self.buffered_result = BufferedTestUnexpectedSuccess(test)
     self.test_result = 'unexpected success'
 
   def addSkip(self, test, reason):
-    errlog(f"{self.compute_progress()}{test} ... skipped '{reason}'")
+    msg = f"skipped '{reason}'"
+    errlog(f"{self.compute_progress()}{test} ... {with_color(CYAN, msg)}")
     self.buffered_result = BufferedTestSkip(test, reason)
     self.test_result = 'skipped'
 
   def addFailure(self, test, err):
-    errlog(f'{self.compute_progress()}{test} ... FAIL')
+    msg = f'{test} ... FAIL'
+    errlog(f'{self.compute_progress()}{with_color(RED, msg)}')
     self.buffered_result = BufferedTestFailure(test, err)
     self.test_result = 'failed'
 
   def addError(self, test, err):
-    errlog(f'{self.compute_progress()}{test} ... ERROR')
+    msg = f'{test} ... ERROR'
+    errlog(f'{self.compute_progress()}{with_color(RED, msg)}')
     self.buffered_result = BufferedTestError(test, err)
     self.test_result = 'errored'
 
--- a/tools/diagnostics.py
+++ b/tools/diagnostics.py
@@ -58,7 +58,7 @@ def reset_color():
 
 
 def with_color(color, text):
-  return f'\033[9{color}m{text}\033[0m'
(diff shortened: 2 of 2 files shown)


juj:
The screenshot shows a glitch in the color output, on line with `test_float_builtins` + `test_float_literals`, but that is an unrelated long standing existing bug on Windows. E.g. on current `main` before this PR, there are some stray escapes getting printed in the console:

image

tools/diagnostics.py
```diff
@@ -134,6 +134,10 @@ def reset_color():
   return '\033[0m'
 
 
+def with_color(color, text):
+  return f'\033[9{color}m{text}\033[0m'
```
sbc100: Why wouldn't this be `output_color() + text + reset_color()`?

Why is ok to use ansi here, even on windows, when we use different mechanism for windows above?

test/parallel_testsuite.py
```diff
@@ -304,35 +305,41 @@ def compute_progress(self):
...
+    errlog(f'{self.compute_progress()}{with_color(RED, msg)}')
     self.buffered_result = BufferedTestFailure(test, err)
     self.test_result = 'failed'
 
   def addError(self, test, err):
-    errlog(f'{self.compute_progress()}{test} ... ERROR')
+    msg = f'{test} ... ERROR'
+    errlog(f'{self.compute_progress()}{with_color(RED, msg)}')
     self.buffered_result = BufferedTestError(test, err)
     self.test_result = 'errored'
```
sbc100, line 344: Can we make this work for `-j1` too?  i.e. the non-parallel runner?  Maybe as a followup?  Maybe the default python runner already has some color support?

tools/diagnostics.py
juj: output_color() and reset_color() adjust the current terminal state already before printing. That works for full lines to be printed, e.g.

```
output_color(...)
print('asdf')
reset_color()
```
but not for adjusting color on an individual line.

> Why is ok to use ansi here

That I actually don't know. In general Python it is not ok on Windows to do that (i.e. if you write a standalone python file with a `print('\033[92m HELLO \033[0m')`, it will not work on Windows.

But we have run some code somewhere else to enable the current terminal to use ANSI escape codes on Windows. Modules like colorama seem to do that, but I don't know what it is that enables that in Emscripten's case for Windows.

test/parallel_testsuite.py
juj: That would be nice (also to add the [35%] progress bar), though looks more difficult, since in non-parallel runner the printing is handled by the built-in [TextTestResult](https://github.com/python/cpython/blob/0b2168275e8ec491fe7ea6f8c662e804437dfdab/Lib/unittest/runner.py#L100) class, which I don't know how to replace.

Another thing I'd want to do is remove the long wall of test results dump at the very end of the test run, which repeats the same results from the test run a second time.. that is a bit spammy.

tools/diagnostics.py
sbc100: But `output_color` and `reset_color` just return strings don't they?  it should be possible to construct your string here by using those function.

juj: `output_color()` calls to `output_color_windows()`, which calls to [SetConsoleTextAttribute](https://github.com/emscripten-core/emscripten/blob/9cc9962662edb359f7b52ad56eb73d6e51c1b886/tools/diagnostics.py#L75), which is a global state for the current terminal.

sbc100: OK, but given that `with_color` in this PR doesn't work on windows anyway why not just do:

```
def with_color(color, text): 
   if WINDOWS:
     ruturn text
   return output_color() + text + reset_color()
```

juj: `with_color()` does work on both Windows 10 and 11.. the screenshot I posted in the PR description is a screenshot from Windows cmd.exe.

sbc100: Ok, in that case I think we should probably just delete the current `output_color_windows` code.

I'm assuming more users are on windows 10 or 11 these days?  Users of older versions of windows just wouldn't get color? 

i.e. if we are going to assume ANSI support on windows lets just do that everywhere maybe?

sbc100: It looks like we can detect it using `ENABLE_VIRTUAL_TERMINAL_PROCESSING`... (I will take a stab at it unless you are feeling like it?)

juj: That sounds like a good idea, though I'd rather leave that for a future refactor PR, since I don't yet know the extent here.

In general for Windows, I think we can call all pre-Windows 10 versions unsupported. (i.e. XP, Vista, 7, 8, 8.1 all are obsolete). On Windows 10, we could take e.g. Win10 21H1 (build 19043, May 18, 2021) as the minimum version (not that it would matter for Emscripten I believe)

sbc100: This can now be `output_color() + text + reset_color()` right?  Now that #25502 has landed?

juj: hmm,

```py
def output_color(color):
  assert color_enabled
  return '\033[3%sm' % color
  ```
  
this generates a different code than `\033[9{color}m` that I was using. I wonder where that `\033[3` comes from. Gives blue instead of cyan.. but green and red are intact.
  
But maybe blue looks nicer than cyan, I'll let that be like that.

Comment (sbc100) [reply tools/diagnostics.py]: Yeah I was wondering that myself, certainly worth not have two ways to output a given color (at least not without documenting).

**Review by sbc100 — APPROVED** [review]

**Comment by juj** [comment]
CircleCI fails on

```
  File "C:\Users\circleci\path with spaces ()\test\parallel_testsuite.py", line 308, in compute_progress
    return with_color(CYAN, val)
  File "C:\Users\circleci\path with spaces ()\tools\diagnostics.py", line 61, in with_color
    return output_color(color) + text + reset_color()
           ~~~~~~~~~~~~^^^^^^^
  File "C:\Users\circleci\path with spaces ()\tools\diagnostics.py", line 46, in output_color
    assert color_enabled
           ^^^^^^^^^^^^^
AssertionError
"""
```

although locally for me on Windows 10, test run is good with colors.

juj:
Demoted the code to output non-colors instead of asserting, if colors not enabled.. Guess that should be per design?

sbc100:
> CircleCI fails on
> 
> ```
>   File "C:\Users\circleci\path with spaces ()\test\parallel_testsuite.py", line 308, in compute_progress
>     return with_color(CYAN, val)
>   File "C:\Users\circleci\path with spaces ()\tools\diagnostics.py", line 61, in with_color
>     return output_color(color) + text + reset_color()
>            ~~~~~~~~~~~~^^^^^^^
>   File "C:\Users\circleci\path with spaces ()\tools\diagnostics.py", line 46, in output_color
>     assert color_enabled
>            ^^^^^^^^^^^^^
> AssertionError
> """
> ```
> 
> although locally for me on Windows 10, test run is good with colors.

Ah I guess we should do `if not color_enabled: return text` maybe?

sbc100 (approved):

Issue #6897, opened 2018-07-24
browser.test_webgl_offscreen_canvas_in_pthread regression after PR #5878


hujiajie:
When `TEST_CHAINED_WEBGL_CONTEXT_PASSING` is defined, the test will fail to create new threads in `PollThreadExit` because the thread exits immediately after `mymain` returns.

`Module['noExitRuntime']` was once honored since #4412 so that async callbacks can continue after the entry function of the thread returns, but the change was reverted in https://github.com/kripken/emscripten/pull/5878/commits/f04b0cdf4956ebe57b852fcd3da95c1b5c6ef5ef. It remains to be discussed which behavior is more correct though.

While debugging this, I also noticed that the offscreen canvas was transferred back to the parent thread at thread shutdown, without checking the aliveness of the parent thread.

PR #25496, opened 2025-10-04
Update test skips and comments related to OffscreenCanvas tests in Firefox

juj:
Update test skips and comments related to OffscreenCanvas tests in Firefox.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1113,16 +1113,10 @@ jobs:
           # initial position of the mouse pointer relative to the canvas.
           # browser.test_html5_webgl_create_context is skipped because
           # anti-aliasing is not well supported.
-          # browser.test_webgl_offscreen_canvas_in_pthread and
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1113,16 +1113,10 @@ jobs:
           # initial position of the mouse pointer relative to the canvas.
           # browser.test_html5_webgl_create_context is skipped because
           # anti-aliasing is not well supported.
-          # browser.test_webgl_offscreen_canvas_in_pthread and
-          # browser.test_webgl_offscreen_canvas_in_mainthread_after_pthread
-          # are crashing Firefox (bugzil.la/1281796). The former case is
-          # further blocked by issue #6897.
```
sbc100, line 1119: What happened to the reference to #6897?  No longer needed?

juj: There is a `emscripten_exit_with_live_runtime()` call in the test which should make the test live past falling out from main(). I don't see that being relevant anymore.

Issue #23982, opened 2025-03-25
EmscriptenAudioWorkletNodeCreateOptions missing couple of options


RReverser:
`EmscriptenAudioWorkletNodeCreateOptions` has bindings for options of `AudioWorkletNode` itself, but `AudioWorkletNode` inherits from `AudioNode` which [also accepts](https://webaudio.github.io/web-audio-api/#audionode-constructor-init):

- `channelCount`
- `channelCountMode`
- `channelInterpretation`

Being able to configure those is useful when you e.g. care about only one input audio channel and don't want the Web Audio system to produce garbage `Float32Array`s for other channels that you're going to ignore anyway.

cc @juj

juj:
Indeed, these are fields that I missed in the initial implementation. Fortunately the options are passed as a struct, so it should be possible to expand to add these. I can't make a promise that I would be able to get to adding these in a moment, as I am currently looking to fix other things, so if you want to take a stab at an implementation meanwhile, that would be cool.

nikitha-m:
i can work on this issue, any guidelines or should i just start figuring it out

RReverser:
It should be fairly straightforward, but do give us a shout if you get stuck.

PR #25497, opened 2025-10-05
Add missing options to EmscriptenAudioWorkletNodeCreateOptions

nikitha-m:
## Problem Description
The EmscriptenAudioWorkletNodeCreateOptions struct currently only exposes options specific to AudioWorkletNode itself (numberOfInputs, numberOfOutputs, outputChannelCounts), but AudioWorkletNode inherits from AudioNode which has additional properties:

- channelCount 
- channelCountMode 
- channelInterpretation 

## Fix 

- Extend the C struct in `system/include/emscripten/webaudio.h`.
- Update the JavaScript implementation in `src/lib/libwebaudio.js` to pass these options to the `AudioWorkletNode` constructor.

Fixes: #23982


system/include/emscripten/webaudio.h
```diff
@@ -134,6 +134,11 @@ typedef struct EmscriptenAudioWorkletNodeCreateOptions
 	int numberOfOutputs;
 	// For each output, specifies the number of audio channels (1=mono/2=stereo/etc.) for that output. Default=an array of ones for each output channel.
 	int *outputChannelCounts;
+    // Extended options from AudioWorkletNode
+	int channelCount;
+	int channelCountMode;
+	int channelInterpretation;
```
juj: [channelCountMode](https://webaudio.github.io/web-audio-api/#enumdef-channelcountmode) and [channelInterpretation](https://webaudio.github.io/web-audio-api/#enumdef-channelinterpretation) are strings, so an integer will not work.

Create an enum for those, and in the JS code, index the enum into an array of strings.

See e.g. https://github.com/juj/wasm_webgpu/blob/3ed529956d1ee3f14f64c07769385e1929908f68/lib/lib_webgpu.js#L615 and https://github.com/juj/wasm_webgpu/blob/3ed529956d1ee3f14f64c07769385e1929908f68/lib/lib_webgpu.js#L729 in another repo for examples.

system/include/emscripten/webaudio.h
```diff
@@ -134,6 +134,11 @@ typedef struct EmscriptenAudioWorkletNodeCreateOptions
 	int numberOfOutputs;
 	// For each output, specifies the number of audio channels (1=mono/2=stereo/etc.) for that output. Default=an array of ones for each output channel.
 	int *outputChannelCounts;
+    // Extended options from AudioWorkletNode
```
juj: This comment documents the history of evolution of the codebase (i.e. "before this PR, we had a 'basic' set, after this PR, we now have 'extended' set"), but in the Web Audio specification, there is no distinction between basic and extended options - there are just options.

So instead of documenting history of evolution, maybe add short comments that describe the new parameters?

nikitha-m:
@juj how do i run test cases locally? I followed the docs but I'm facing some issues while running these commands
```
emsdk install tot
emsdk activate tot
```

juj:
Yeah, that is how you install a pre-compiled Emscripten. If installing `tot` doesn't work, you can also try `emsdk install latest` followed by `emsdk activate latest`.

To run browser audio-related tests, you can run `test/runner browser.test_*audio*` and `test/runner interactive.test_*audio*`. The first suite is automated, but the second suite requires an interactive user to prod the tests forward.

If installing `tot` or `latest` doesn't work out, you can also try these steps to build all from source: https://bugzilla.mozilla.org/show_bug.cgi?id=1992558#c2

nikitha-m:
@juj i think it's done. Is this fine?

src/lib/libwebaudio.js
```diff
@@ -318,6 +318,9 @@ var LibraryWebAudio = {
       numberOfInputs: {{{ makeGetValue('options', C_STRUCTS.EmscriptenAudioWorkletNodeCreateOptions.numberOfInputs, 'i32') }}},
       numberOfOutputs: optionsOutputs,
       outputChannelCount: readChannelCountArray({{{ makeGetValue('options', C_STRUCTS.EmscriptenAudioWorkletNodeCreateOptions.outputChannelCounts, 'i32*') }}}, optionsOutputs),
+      channelCount: (function(){ var v = {{{ makeGetValue('options', C_STRUCTS.EmscriptenAudioWorkletNodeCreateOptions.channelCount, 'u32') }}}; return v > 0 ? v : undefined; })(),
+      channelCountMode: (function(){ var v = {{{ makeGetValue('options', C_STRUCTS.EmscriptenAudioWorkletNodeCreateOptions.channelCountMode, 'i32') }}}; var arr = ['max','clamped-max','explicit']; return arr[v] || undefined; })(),
+      channelInterpretation: (function(){ var v = {{{ makeGetValue('options', C_STRUCTS.EmscriptenAudioWorkletNodeCreateOptions.channelInterpretation, 'i32') }}}; var arr = ['speakers','discrete']; return arr[v] || undefined; })(),
```
juj: The anonymous functions and `|| undefined` should be unnecessary here, they take up code size for no benefit.

E.g. `channelCountMode: ['max','clamped-max','explicit'][{{{ makeGetValue(...) }}}],` looks like would be enough here?

One thing to verify here is how do older browsers behave (or Safari) that do not yet support these options? If you run in an older browser that does not support any of this, would that cause an error?

It would be good to have a short test of non-default `channelCountMode` and non-default `channelInterpretation` to ensure that it works.

system/include/emscripten/webaudio.h
```diff
@@ -134,6 +145,10 @@ typedef struct EmscriptenAudioWorkletNodeCreateOptions
 	int numberOfOutputs;
 	// For each output, specifies the number of audio channels (1=mono/2=stereo/etc.) for that output. Default=an array of ones for each output channel.
 	int *outputChannelCounts;
+	unsigned long channelCount;
+	WEBAUDIO_CHANNEL_COUNT_MODE channelCountMode;
+	WEBAUDIO_CHANNEL_INTERPRETATION channelInterpretation;
```
juj, line 153: Can you add here a short comment describing the default values of each?

Hmm it seems that we get the struct initialization regression problem here.. that is, after this addition, all users must explicitly initialize these values to valid numbers; or their code will stop working? E.g. is existing users had code, say, 

```c
EmscriptenAudioWorkletNodeCreateOptions options = {
  .numberOfInputs = 2,
  .numberOfOutputs = 2,
};
```

then when they update Emscripten to recompile their code, they will get implicit
```
.channelCount = 0,
.channelCountMode = WEBAUDIO_CHANNEL_COUNT_MODE_MAX,
.channelInterpretation = WEBAUDIO_CHANNEL_INTERPRETATION_SPEAKERS
```
and the [spec says](https://www.w3.org/TR/webaudio-1.1/#dom-audionode-channelcount), `"If this value is set to zero [...] the implementation MUST throw a [NotSupportedError](https://webidl.spec.whatwg.org/#notsupportederror) exception."`

So in this form, this update would cause all Web Audio utilizing code to break from running, unless they are updated.

That is a bit annoying, since there is necessarily not a good way for developers to figure that out, and they might raise a bug report "I updated Emscripten and now I am getting an NotSupportedError exception. It used to work in previous version of Emscripten, so what did you regress?"

juj: Oh now I see in the init code there was a `v > 0 ? v : undefined` added to work around this problem.

Hmm, I suppose that's ok. It would be good to have a comment, e.g.

```
// the number of channels used when up-mixing and down-mixing connections to any inputs to the node.
// Pass 0 to omit specifying this parameter altogether during context creation.
unsigned long channelCount;
```

juj: @cwoffenden what do you think about the struct init problem for the future?

Should we try to model the struct so that all-zero-init should be preserved as a valid struct init with defaults? E.g.

```c
EmscriptenAudioWorkletNodeCreateOptions options = {};
```
would provide all zeros, and at JS side, we would always shoehorn the init to map to default values of an Audio Context?

In WebGPU I have been explicitly using global constant data default-initialized struct constants, e.g. here: https://github.com/juj/wasm_webgpu/blob/3ed529956d1ee3f14f64c07769385e1929908f68/lib/lib_webgpu_cpp20.cpp#L34-L45

which would mean we would enforce users to do
```c
EmscriptenAudioWorkletNodeCreateOptions options = AUDIO_WORKLET_NODE_CREATE_DEFAULT_INITIALIZER;
```
if they want to ensure forward compatibility with default values into the future.

The default-zero-init for all fields would be preferred for code size and simplicity, though only if we can expect that any upcoming fields in the Web Audio spec would reasonably map from zero to the default. WDYT?

cwoffenden: I’m out for a week or so but jotting some quick thoughts down, I personally prefer (and use) the zero initialiser, with, for example, an enum `UNDEFINED` mapped to zero, and other special treatment of zeros. I find it cleaner than building a struct with all the defaults, and the constructor passed the struct is probably going to have all the sanity checks anyway, so using that to create sane defaults is a nice byproduct.

juj: > I personally prefer [...] the zero initialiser

Yeah, of course using zeroes is nice, though this is not a question of (just) programming style or preference. Rather the risk there is whether we will always be technically able to do this.

For example, in this specific PR case we were just barely saved by a lucky happenstance that the new added param `channelCount` has zero be an invalid value in the upstream Web Audio spec.

If that were not the case, then the above `v > 0 ? v : undefined` workaround would not be possible, but user breakage would have been inevitable, unless we also added some extra `bool channelCountSpecified;` variable.

Though I suppose the resolution here is clear that we don't want to go the route of initializer structs (or functions, like `pthread_attr_init()` for example has), and deal with the problematic case when/if one arises in the future.

nikitha-m: Are we making the design changes and then merging this or is this good enough?

nikitha-m: > Are we making the design changes and then merging this or is this good enough?

@juj

nikitha-m:
@juj @cwoffenden could you please lmk if any changes are required other than adding comments?

src/lib/libwebaudio.js
juj: Did you have a chance to look at the above review comment?

How about this code:

```js
      channelCount: {{{ makeGetValue('options', C_STRUCTS.EmscriptenAudioWorkletNodeCreateOptions.channelCount, 'u32') }}} || undefined,
      channelCountMode: [/*'max'*/,'clamped-max','explicit'][{{{ makeGetValue('options', C_STRUCTS.EmscriptenAudioWorkletNodeCreateOptions.channelCountMode, 'i32') }}}],
      channelInterpretation: [/*'speakers'*/,'discrete'][{{{ makeGetValue('options', C_STRUCTS.EmscriptenAudioWorkletNodeCreateOptions.channelInterpretation, 'i32') }}}],
```

It is shorter for the same effect? ('max' and 'speakers' are the default in the spec if I understand correctly)

system/include/emscripten/webaudio.h
juj: No design changes, let's go with the approach to keep zero-initialized struct as mapping to the defaults.

nikitha-m:
@juj This test case that failed doesn't look like its related to the webaudio related tests can you help me with this?

sbc100:
@nikitha-m can you try merging/rebasing with main?

src/struct_info.json
```diff
@@ -1260,7 +1260,7 @@
             ],
             "EmscriptenWebSocketCreateAttributes": [
                 "protocols"
-            ]
```
cwoffenden, line 1263: Minor: formatting got lost.

nikitha-m:
@juj @cwoffenden its done

cwoffenden:
> @juj @cwoffenden its done

I can neither approve nor merge, I can just say that it looks correct and doesn't break anything. I've not personally used these audio node properties for mixing so I can't comment further (we perform all the mixing ourselves and to a regular stereo output).

juj:
Hmm looks like CircleCI is stuck here on code size checks.

I'll land this and re-roll the code size update action, I think those will change here.

Thanks for the contribution @nikitha-m .

PR #25499, opened 2025-10-05
Enable support for running the test harness in Safari.

juj:
Add support for running the `browser` harness with `EMTEST_BROWSER=/Applications/Safari.App/Contents/MacOS/Safari` on a macOS system.

This fixes both the singlethreaded and multithreaded harness runs.


test/common.py
```diff
@@ -2567,52 +2609,65 @@ def browser_open(cls, url):
...
+      if parallel_harness or is_safari():
         time.sleep(2 + count * 0.3)
         procs_after = list_processes_by_name(config.executable_name)
+
+        # Take a snapshot again to find which processes exist after launching
+        # the browser. Then the newly launched browser processes are determined
+        # by the delta before->after.
+        cls.browser_procs = list(set(procs_after).difference(set(procs_before)))
+        if len(cls.browser_procs) == 0:
+          logger.warning('Could not detect the launched browser subprocesses. The test harness may not be able to close browser windows if a test hangs, and at harness exit.')
```
sbc100: Can this be an error instead?

test/common.py
```diff
@@ -2567,52 +2609,65 @@ def browser_open(cls, url):
...
       browser_args += f' {config.data_dir_flag}"{browser_data_dir}"'
 
     browser_args = shlex.split(browser_args)
+    if is_safari():
+      # For the macOS 'open' command, pass
+      #   --new: to make a new Safari app be launched, rather than add a tab to an existing Safari process/window
+      #   --fresh: do not restore old tabs (e.g. if user had old navigated windows open)
+      #   --background: Open the new Safari window behind the current Terminal window, to make following the test run more pleasing (this is for convenience only)
+      #   -a <exe_name>: The path to the executable to open, in this case Safari
+      browser_args = ['open', '--new', '--fresh', '--background', '-a'] + browser_args
```
sbc100: Shouldn't these args go in `SafariConfig`?

sbc100: Or maybe these should be used also for Chrome and FF on macOS?    i.e. is `open` the way we should be launching all browsers on macOS?

test/common.py
```diff
@@ -2567,52 +2609,65 @@ def browser_open(cls, url):
...
       os.mkdir(browser_data_dir)
 
-      if is_chrome():
-        config = ChromeConfig()
-      elif is_firefox():
-        config = FirefoxConfig()
-      else:
-        exit_with_error(f'EMTEST_BROWSER_AUTO_CONFIG only currently works with firefox or chrome. EMTEST_BROWSER was "{EMTEST_BROWSER}"')
+      if not config:
+        exit_with_error(f'EMTEST_BROWSER_AUTO_CONFIG only currently works with firefox, chrome and safari. EMTEST_BROWSER was "{EMTEST_BROWSER}"')
```
sbc100: This error seems like its unreachable since its inside the `if EMTEST_BROWSER_AUTO_CONFIG and config` block?

test/common.py
juj: I can make it an error, but it is only so when running with EMTEST_BROWSER_AUTO_CONFIG enabled. (I think btw disabling EMTEST_BROWSER_AUTO_CONFIG might currently be broken and atm not possible)

If user adds a custom browser string, and EMTEST_BROWSER_AUTO_CONFIG is not enabled, then the browser might run in a tab on an existing browser, in which case this detection will not work.

test/common.py
juj: It'll need a new construct in the configs, there is no current way to model a prefix of args in the existing set of config fields. I can do that in a moment.

Using `open` is not needed for Firefox at least. Not sure about Chrome, but I presume not.

test/common.py
juj: Thanks, good catch.

juj:
This would be great to land, so I can get on to running tests on the Safari browser next.

sbc100 (commented):
Added @brendandahl who has also been working on this stuff recently.

lgtm, even though all this extra complexity is making me sad.

juj:
You know what they say, *better correct and sad, rather than end users mad.*

test/common.py
```diff
@@ -205,11 +226,28 @@ def is_firefox():
   return EMTEST_BROWSER and 'firefox' in EMTEST_BROWSER.lower()
 
 
+def is_safari():
+  return EMTEST_BROWSER and 'safari' in EMTEST_BROWSER.lower()
+
+
+def get_safari_version():
```
sbc100: How about `get_version_from_macos_app_dir` (or something like that?)

sbc100: Oh wait.. this function looks like it not actually used?

juj: Err, I was hoping you wouldn't notice :)

I used this function in the initial version of this PR. Initially I couldn't figure out why old Safari was failing some random tests. And it amounted to being that missing HTTP caching header `Vary: *` that was causing the failures.

So I could then remove the use of the version check.

But I have plans to use that Safari version check soon more once I get to test more coverage of the tests in different Safari versions. Instead of a `@no_safari('reason')` annotation, I would like to annotate tests with `@no_safari_older_than(XXYYZZ, 'reason')` so that the test skips are not that blunt and oblivious to the version. That way they can naturally evolve out when minimum supported Safari version evolves.

If you'd like, I can also reintroduce that function in a later PR that actually adds uses of such pattern. (I am planning to introduce a symmetric `get_firefox_version()` function as well, and instead of `@no_firefox`, to have `@no_firefox_older_than` checks)

sbc100: Can maybe run something like `browser-binary --version` in all cases?    Then the same code work everywhere?

Lets remove this PR since it also introduces that new plist import thing.

sbc100: Its a shame the new `vulture` check isn't able to be sure about this code being dead so it doesn't fail in the CI step :(

juj: ok, removed

brendandahl (approved):

PR #25502, opened 2025-10-06
Remove windows-specific console coloring

sbc100:
Windows 10 and above support ANSI coloring and there should be very few emscripten users on older version of windows at this point (if any?).

emscripten already does not support windows 7 IIUC, since its dependencies (both python 3 and node 18) do not support windows 7

This means that only windows 8 users should be effected by this change.  The upside is that we can cleanup and refactor this code now (see #25495).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,8 @@ See docs/process.md for more on how version tagging works.
 
 4.0.16 (in development)
 -----------------------
+- For windows users, colored console output for error messages and logging now
+  requires Windows 10 or above. (#25502)
 - A warning was added about usage of embind without C++17 or above. (#25424)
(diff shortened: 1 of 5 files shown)


juj:
I merged this branch against my PR branch https://github.com/emscripten-core/emscripten/pull/25495 and verified that this will not disrupt the coloring added in that branch on Windows 10. LGTM.

juj (approved):

juj:
Unfortunately I see that this PR did regress colors on Windows:

image

When I wrote `"verified that this will not disrupt the coloring added in that branch"` above, I had only tested the colors I had added to the parallel test harness. Before and after this PR #25502, the colors in the parallel test harness do work; but it looks like the other colors, such as the red `error:` got removed by this commit.

Fix in https://github.com/emscripten-core/emscripten/pull/25525

PR #25514, opened 2025-10-07
Remove the 'ENVIRONMENT=worker implies ENVIRONMENT=web' assumption

juj:
Remove the 'ENVIRONMENT=worker implies ENVIRONMENT=web' assumption. Fixes https://github.com/emscripten-core/emscripten/pull/25414#pullrequestreview-3309445257. Add test to verify the expectation mentioned in that comment.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -25,8 +25,8 @@ See docs/process.md for more on how version tagging works.
   - Firefox: v65 -> v68
 - For windows users, colored console output for error messages and logging now
   requires Windows 10 or above. (#25502)
-- Fixed an issue from previous release where "-sENVIRONMENT=worker" was made
-  erroneously to imply "-sENVIRONMENT=web,worker" (#25514)
+- Fixed an issue from previous release 4.0.16 where "-sENVIRONMENT=worker" was
+  erroneously made to imply "-sENVIRONMENT=web,worker" (#25514)
 - Passing '-sENVIRONMENT=worker' is now disallowed due to being ambiguous in
   its meaning. Instead, use '-sENVIRONMENT=web,worker' or
   '-sENVIRONMENT=node,worker' to refer to either Web or Node.js multithreading.


sbc100:
I suppose there is not much benefit in a program that *only* runs in webworker and therefore can elide any code related to the browser main thread?

My previous assumption is that that is what this test was testing.

juj:
> a program that _only_ runs in webworker and therefore can elide any code related to the browser main thread?

Ohhh. hmmm that is an interesting idea. I suppose there could be such an optimization opportunity, although we would then have to have a `worker-web` vs `worker-node` distinction to separate between browser and Node Workers? (so passing `-sENVIRONMENT=worker` is still ambiguous whether it means web vs Node worker?)

tools/link.py
```diff
@@ -172,7 +172,10 @@ def setup_environment_settings():
   if any(x for x in settings.ENVIRONMENT if x not in VALID_ENVIRONMENTS):
     exit_with_error(f'Invalid environment specified in "ENVIRONMENT": {settings.ENVIRONMENT}. Should be one of: {",".join(VALID_ENVIRONMENTS)}')
 
-  settings.ENVIRONMENT_MAY_BE_WEB = not settings.ENVIRONMENT or 'web' in settings.ENVIRONMENT or 'worker' in settings.ENVIRONMENT
+  if settings.ENVIRONMENT and settings.ENVIRONMENT == ['worker']:
```
sbc100: I guess we would want something more like 

`"worker" in settings.ENVIRONMENT and not any([e in settings.ENVIRONMENT for e in ['web', 'node'])`

And then the error message should explicitly say something like "specifying `worker` requires wither `node` or `web`"

sbc100 (approved):

sbc100 (commented):
Maybe add a test for the new error message?

kripken:
Please add a changelog entry for this breaking change.

juj:
Hmm, actually taking a step back.. would such a `-sENVIRONMENT=node-worker` ever be a use case?

That is, would there ever be such a thing as "My output targets Node.js, but only inside a pthread/worker in Node"?

Because if not (and it does seem to me that would never be a thing?), then we would not need to disambiguate `worker-web` vs `worker-node`, but we could take the convention that `-sENVIRONMENT=worker` means exactly what Sam mentioned above: the code is only ever to run inside a web browser, and there only in a Worker, but never on the main browser thread?

Maybe that was the original intent of `-sENVIRONMENT=worker`? (I can't quite recall.. I've been familiar only with `-sENVIRONMENT=web`, `-sENVIRONMENT=web,worker`, and `-sENVIRONMENT=node` use in general)

juj:
Actually yeah, that looks like what the original test intent of `browser.test_pthreads_started_in_worker_limited_env` was, now that I read it closer.

Adjusted the PR here to do a better fix for the problem. How does this look now?

sbc100 (approved):

juj:
Hmm I don't think this should break any previous use case, but fix the previous regression? Which breakage are you referring to?

kripken:
Will this not start to error on people passing `ENVIRONMENT=worker` and that were depending on that enabling "web" as well?

juj:
> Will this not start to error on people passing `ENVIRONMENT=worker` and that were depending on that enabling "web" as well?

The understanding here was that passing `-sENVIRONMENT=worker` has never intended to imply to enable `web` as well, but it was just my PR #25414 from last week that accidentally made it so (due to my misunderstanding of the situation), resulting in an accidental regression.

At least the test `browser.test_pthreads_started_in_worker_limited_env` seems to behave like `-sENVIRONMENT=worker` should only mean running in a web browser Worker, but not on the main thread.

kripken:
Yeah, since it's just from last week maybe it's not worth a mention. But it did make it into a release, so it is noticeable, that's what worries me. But I don't feel strongly.

juj:
> But it did make it into a release

Oh I see.. yeah, I'll add a ChangeLog entry. How does that look?

ChangeLog.md
```diff
@@ -25,6 +25,11 @@ See docs/process.md for more on how version tagging works.
   - Firefox: v65 -> v68
 - For windows users, colored console output for error messages and logging now
   requires Windows 10 or above. (#25502)
+- Fixed an issue from previous release where "-sENVIRONMENT=worker" was made
+  erroneously to imply "-sENVIRONMENT=web,worker" (#25514)
```
kripken: Please add the version (4.0.16?)

kripken (approved):
Thanks, lgtm

juj:
ok

kripken (approved):
Thanks!

PR #25519, opened 2025-10-07
[jsifier] Simplify stub function handling

sbc100:


src/jsifier.mjs
```diff
@@ -615,18 +618,10 @@ function(${args}) {
...
-          const functionBody = assertion + `return ${target}(...args);`;
-          LibraryManager.library[symbol] = new Function('...args', functionBody);
-          isStub = true;
-        } else {
-          // emit a stub that will fail at runtime
-          LibraryManager.library[symbol] = new Function(`abort('missing function: ${symbol}');`);
-          // We have already warned/errored about this function, so for the purposes of Closure use, mute all type checks
-          // regarding this function, marking ot a variadic function that can take in anything and return anything.
-          // (not useful to warn/error multiple times)
-          LibraryManager.library[symbol + '__docs'] = '/** @type {function(...*):?} */';
```
kripken, line 627: Is this not needed for some reason?

sbc100: Seems like its no longer needed, since closure can see the `...` in the function body I believe.

sbc100: Since we are not always creating the stubs with ` new Function('...args')` closure can see that the function variadic.

kripken: Interesting, thanks.

Issue #12682, opened 2020-11-02
Stop building MAIN_MODULE with RELOCATABLE (-fPIC)


sbc100:
There is no need to the main module itself to be relocatable.  We should be able to get some code size and performance wins by making it static.

sbc100:
@awtcode @tlively

sbc100:
I spend a while looking into this yesterday and hit a bit of a road block.

The goal here is to build the main module without `-fPIC` at compile time and without `-pie` link time.

The problem is that as of today when you build without `-fPIC` the codegen assume that all global addresses will be known at static link time (so it can always use `i32.const <RELOC>`).  Its only when we enable `-fPIC` that we use the more conservative sequence `global.get <GOT.RELOC>`.     As it turns out what we want in this case non-position-independent executable that still uses the GOT for accessing non-local symbol addresses.  

Basically we have conflated to concpets of PIC code and dynamically linked code (GOT usage for external symbols).

I am now leaning towards using GOT relocations conservatively in more cases (not just `-fPIC`) ... maybe all cases?  ... I think this is what llvm internally calls `-m relocation-model=dynamic-no-pic`.  I wonder if we should just make this the default?  The downside is that it generates a lot of extra wasm globals in the static linking case, but these end up being const and can be completely eliminated by wasm-opt.     One upside is that anyone who build with `--allow-undefined` will now be able to supply values for all undefined functions and global addresses at runtime.    As of today `--allow-undefined` will turn undefined functions into imports but undefined data into zero addresses.

Another minor positive is that since we can now name globals in the name section it should make it disassembing unoptimized executables nicer.

tlively:
Interesting! How does native code handle this situation? Can the relocation model be controlled by command line? Perhaps Emscripten could then opt into the more conservative mode rather than making it the LLVM default. Or we could make that conservative mode the LLVM default and folks who want the old behavior would be able to opt out.

kripken:
Interesting...

How big is the extra overhead here, and how important is it to reduce?

sbc100:
Are you talking about the extra overhead introduced by building the MAIN_MODULE as RELOCTABLE today?   Or the extra overhead of my proposal to enable `-m relocation-model=dynamic-no-pic` by default?

I believe all the overhead introduced by `-m relocation-model=dynamic-no-pic` can effectively be removed by `wasm-opt`. 

[...]

PR #25522, opened 2025-10-08
Don't build MAIN_MODULE as RELOCATABLE

sbc100:
This means we no longer link main modules with `-pie`.  The main
advantage here is that main modules no longer require
runtime relocation code for symbols defined locally.

To show the benefits of this approach I build binaryan_wasm
with `-sMAIN_MODULE=1` both before and after this change.

Before:

Wasm Size: 22.6M
__wasm_apply_data_relocs size: 123k

After:

Wasm Size: 16.6M
__wasm_apply_data_relocs size: 0 (no longer exists)

Since this is quite a major change, I've kept the `-sRELOCATABLE`
option available so folks that use this to get the old behaviour when
linking their main module.  In order to tests this path too now run
all the `@needs_dylink` tests in test_core.py with and without
`-sRELOCTABLE`.  Hopefully we can remove this configuration
completely at some point in the future.

Fixes: #12682


sbc100:
I still need to get some stats on the benefits of this, but that tests are all now passing so I think this should be good to review now.

An open question is: What do we do about `-sRELOCTABLE`.   Once this change lands we will have almost not testing on this flag.  We do have plan to remove it: #25262.   One option is to remove `-sRELOCATABLE` before this change lands?

kripken:
Yeah, I see what you mean, removing RELOCATABLE first would simplify this a little. I don't have a preference though.

I do see some tests still fail here though?

tools/compile.py
```diff
@@ -50,10 +50,10 @@ def get_clang_flags(user_args):
     if '-mbulk-memory' not in user_args:
       flags.append('-mbulk-memory')
 
-  if settings.RELOCATABLE and '-fPIC' not in user_args:
+  if (settings.MAIN_MODULE or settings.RELOCATABLE) and '-fPIC' not in user_args:
```
kleisauke, line 53: Is this correct? From my understanding of issue #12682, we want to _avoid_ compiling the main module with `-fPIC` and compile only the objects included in the side module as relocatable.

Note that there are still some other assumptions related to "compile with `-fPIC` when using the `-sMAIN_MODULE` setting". I attempted to revert these with commit https://github.com/kleisauke/emscripten/commit/cc6ae3e41408545e35652e08059479c2d134034, which appears to work well in a complex project that uses dynamic linking, see e.g. commit:
https://github.com/kleisauke/wasm-vips/commit/b5d804001040cb3d24e467dd903a7f0933145e31

sbc100: Yes we still need the inputs to be compiled with `-fPIC` in case there reference non-DSO local symbols.

I agree its a little confusing and perhaps we should have the compiler always use GOT references for symbols that it cannot determine are DSO-local? 

The main difference with this change does not occur at compile time by at link time where the linker creates a normal executable at a fixes location, but with the possibility of external DSO references.  This is kind of like the normal `-Bdynamic` mode linux, as opposes to the full `-pie` mode.

kleisauke: Ah, that makes sense. I'd forgotten that `-sRELOCATABLE` not only implies `-fPIC` during compilation but also `-pie` during linking. This PR ensures the main module is still compiled with `-fPIC`, but not linked with `-pie`.

ChangeLog.md
```diff
@@ -20,6 +20,9 @@ See docs/process.md for more on how version tagging works.
 
 4.0.19 (in development)
 -----------------------
+- When using dynamic linking the main module is no longer built as a relocatable
+  binaryen.  This should reduce the overhead of dynamic linking for the main
```
kleisauke:
```suggestion
  binary.  This should reduce the overhead of dynamic linking for the main
```

sbc100:
Ok, aside from the problem of what to do with `-sRELOCATABLE` setting, this change is not ready to go I think.

The stats look really promising.  See the new PR description for the great improvements measured when building a dylink version of binaryen.

src/lib/libdylink.js
```diff
@@ -315,13 +322,17 @@ var LibraryDylink = {
     dbg('reportUndefinedSymbols');
 #endif
     for (var [symName, entry] of Object.entries(GOT)) {
-      if (entry.value == 0) {
+      if (entry.value == {{{ to64(-1) }}}) {
```
kripken: Maybe comment what this means? (tombstone?)

src/lib/libdylink.js
```diff
@@ -409,7 +420,14 @@ var LibraryDylink = {
...
     ___heap_base = end;
-    GOT['__heap_base'].value = {{{ to64('end') }}};
+#if PTHREADS
+    if (!ENVIRONMENT_IS_PTHREAD) {
+#endif
+      var sbrk_ptr = _emscripten_get_sbrk_ptr();
+      {{{ makeSetValue('sbrk_ptr', 0, 'end', '*') }}}
+#if PTHREADS
+    }
+#endif
```
kripken, line 447: Why is this needed?

test/common.py
```diff
@@ -1439,24 +1439,6 @@ def ccshared(src, linkto=None):
...
-      }
-      #endif
-
-      int test_main() {
-        bfunc();
-        cfunc();
-        return 0;
-      }
-      ''',
-           'a: loaded\na: b (prev: (null))\na: c (prev: b)\n', cflags=extra_args)
```
kripken, line 1448: Why remove this test?

src/lib/libdylink.js
sbc100: In the old world `__heap_base` is defined by the dynamic linker and then imported by all the relocatable modules, including the main module.

In the new world the main module statically defines `__heap_base` and then exports it as a fixed address.

The old code basically treated `__heap_base` as non-constant and bumped it when loading dynamic libraries (prior to runtime initialization).

This requires a corresponding hack in the implementation of `emscripten_get_sbrk_ptr`.. I will see if I can split this change out and land it first.

sbc100: Actually this particular path is different with and without `-sRELOCATABLE` so I've included both paths now.

sbc100:
I solves the `-sRELOCTABLE` issue by keeping that available as an option and test that configuration too.

src/lib/libdylink.js
```diff
@@ -409,7 +420,25 @@ var LibraryDylink = {
     assert(end <= HEAP8.length, 'failure to getMemory - memory growth etc. is not supported there, call malloc/sbrk directly or increase INITIAL_MEMORY');
 #endif
     ___heap_base = end;
+
+    // After allocating the memory from the start the heap we need to ensure
+    // that once the program starts it doesn't use this region.  In relocatable
+    // mode we can just updat ethe __heap_base symbol that we are exporting to
```
kripken:
```suggestion
    // mode we can just update the __heap_base symbol that we are exporting to
```

src/lib/libdylink.js
```diff
@@ -409,7 +420,25 @@ var LibraryDylink = {
     assert(end <= HEAP8.length, 'failure to getMemory - memory growth etc. is not supported there, call malloc/sbrk directly or increase INITIAL_MEMORY');
 #endif
     ___heap_base = end;
+
+    // After allocating the memory from the start the heap we need to ensure
```
kripken:
```suggestion
    // After allocating the memory from the start of the heap we need to ensure
```

src/lib/libdylink.js
```diff
@@ -409,7 +420,25 @@ var LibraryDylink = {
...
+    // When not relocatable `__heap_base` is fixed and exported by the main
+    // module, but we can update the `sbrk_ptr` value instead.
+#if RELOCATABLE
     GOT['__heap_base'].value = {{{ to64('end') }}};
+#else
+#if PTHREADS
+    if (!ENVIRONMENT_IS_PTHREAD) {
+#endif
+      var sbrk_ptr = _emscripten_get_sbrk_ptr();
+      {{{ makeSetValue('sbrk_ptr', 0, 'end', '*') }}}
```
kripken, line 443: hmm, why not call `sbrk()` directly?

sbc100: The runtime has not been initialized so we have to be very careful about what we call here.   We certainly can't call arbitrary code.

Note that if the runtime has been initialized we just call `calloc` above.

kripken: Right, thanks, that's what I was missing.

sbc100:
I think these change should be more of less ready to land now.

The old `-sRELOCTABLE` mode is still available for now.

Once I land this I will likely send a message to mailing list too since its a fairly large change for users of dynamic linking.

src/lib/libdylink.js
```diff
@@ -409,7 +424,25 @@ var LibraryDylink = {
     assert(end <= HEAP8.length, 'failure to getMemory - memory growth etc. is not supported there, call malloc/sbrk directly or increase INITIAL_MEMORY');
 #endif
     ___heap_base = end;
+
+    // After allocating the memory from the start of the heap we need to ensure
+    // that once the program starts it doesn't use this region.  In relocatable
+    // mode we can just update the __heap_base symbol that we are exporting to
+    // the main module.
+    // When not relocatable `__heap_base` is fixed and exported by the main
+    // module, but we can update the `sbrk_ptr` value instead.
```
kripken:
```suggestion
    // module, but we can update the `sbrk_ptr` value instead (that is safe to
    // call, even if `sbrk` itself isn't yet).
```

test/test_core.py
```diff
@@ -191,10 +191,19 @@ def needs_dylink(func):
   assert callable(func)
 
   @wraps(func)
-  def decorated(self, *args, **kwargs):
+  def decorated(self, relocatable, *args, **kwargs):
     self.check_dylink()
+    if relocatable:
+      # Since `-sMAIN_MODULE` no longer implies `-sRELOCATABLE` but we want
+      # to keep that cominbation working we run all the `@needs_dylink` tests
+      # both with and without the explicit `-sRELOCATABLE`
```
kripken, line 199: This is quite a lot of slow tests - is it worth it?

test/test_other.py
```diff
@@ -5157,7 +5157,7 @@ def test_no_dynamic_execution(self):
     delete_file('a.out.js')
 
     # Test that -sDYNAMIC_EXECUTION=0 and -sMAIN_MODULE are allowed together.
-    self.do_runf('hello_world.c', cflags=['-O1', '-sDYNAMIC_EXECUTION=0', '-sMAIN_MODULE'])
+    self.do_runf('hello_world.c', cflags=['-O1', '-sDYNAMIC_EXECUTION=0', '-sMAIN_MODULE=2'])
```
kripken, line 5160: Why change this?

test/test_core.py
sbc100: I think this is important as long as we want to offer people a fallback using `-sRELOCATABLE` for some period of time, which I do think is good idea.

Once we remove the fallback option we can then remove all these tests.

tools/emscripten.py
```diff
@@ -902,7 +902,7 @@ def install_debug_wrapper(sym):
     return False
   # Likewise `__trap` can occur before the runtime is initialized since it is used in
   # abort.
-  return sym != '__trap'
+  return sym not in ['__trap', 'emscripten_get_sbrk_ptr']
```
kripken, line 906:
```suggestion
  # emscripten_get_sbrk_ptr can be called before initialization from dylink code.
  return sym not in ['__trap', 'emscripten_get_sbrk_ptr']
```

test/test_core.py
kripken: How much time does it take? I just vaguely worry dylink tests are most of our `core` time, and doubling them would almost double the total...

test/test_other.py
sbc100: This is just a drive by change which makes the test slightly faster.

test/test_core.py
sbc100: These two tests total 61 out of 1176 total core2 tests.

In terms of runtime of the core2 test suite they comprise 9 seconds of the full runtime of 63 seconds (I skipped the slow tests there too).

kripken: Ok, sg, thanks for checking.

PR #4935, opened 2017-02-11
Integrate the pipe() syscall implementation originally written by @cynecx

atrosinenko:
This pull request is based on the PR #4378 by @cynecx. I am not a JavaScript developer and I am not familiar with the Emscripten FS layer, though, the majority of PIPEFS code was already written by @cynecx and even reviewed there to some extent. The problem with that PR was the lack of tests. I have added some tests and the tests revealed some problems with the original implementation that I have fixed.

There still may be some performance issue with the (common?) use case with writing a few bytes, and then reading these bytes when implementing events, since every `read()` call seems to pop an almost empty 8kb buffer, and `write()` will allocate a new one, so It may generate some excessive garbage.

Another peculiarity of this code is that the read end of the pipe always behaves as if it is non-blocking -- is it currently possible to block the read operation at all?

PR #4378, opened 2016-06-01
Add support for pipe syscall (PIPEFS)

cynecx:
Addressing https://github.com/kripken/emscripten/issues/2883.
- [ ] Add test for pipefs
- [ ] Add more tests for pipefs

PR #25523, opened 2025-10-08
Fix `select` didn't work with pipes and the timeout argument

ktock:
This PR fixes an issue that the timeout argument for the select syscall wasn't applied to pipes and it returned immedeately instead of waiting for events.

Although the select syscall passes the timeout value to the `poll` method of the stream implementations, this can't handle multiple fds because a fd can't notify readiness while another is blocking in `poll`. As a result, using the select syscall with a combination of pipes and other streams (e.g. PTY) can be problematic.

To address this, this PR implements a callback-based event notification. Each stream implementation's `poll` invocation receives a callback, allowing it asynchronously notify the select syscall when it becomes ready. The select syscall blocks until one of these callbacks is triggered or the timeout expires. This behviour is enabled only when PROXY_TO_PTHREAD is enabled to avoid blocking the main worker.

To maintain compatibility with non-pipefs streams, the select syscall allows stream implementations to ignore the callback and synchronously return the event status instead. In such cases, the select syscall still updates the flags accordingly.


sbc100:
Thanks for the PR @ktock.  This looks like a fairly large and complex PR, and I'm out sick right now I'm afraid so I'm not sure when I will have time to go through it.

CC'ing @atrosinenko who wrote the original pipefs back in #4935. 

Also @kripken and @tlively who might be interesting the multithreaded event stuff.

ktock:
Rebased. Could we move this forward?

kripken:
I am unfortunately not familiar with this code (or with native pipes either). It looks like the PIPEFS code began in https://github.com/emscripten-core/emscripten/pull/4378 / #4935 by @cynecx and @atrosinenko - would one of you be interested to review this perhaps?

ktock:
@sbc100 @kripken

> This looks like a fairly large and complex PR

I've just simplified the patch by removing changes in libpthread.js. Instead, this patch uses emscripten_proxy_sync_with_ctx to synchronize the thread worker and the main worker. This patch allows pipefs and setTimeout to unblock the worker by calling emscripten_proxy_finish when an event occurs. I've added some wrapper functions for the proxying-related APIs so that the JS implementation of newselect can use them.

src/lib/libpipefs.js
```diff
@@ -21,6 +21,19 @@ addToLibrary({
         // able to read from the read end after write end is closed.
         refcnt : 2,
         timestamp: new Date(),
+        readableHandlers: [],
+        registerReadableHanlders: (notifyCallback) => {
```
sbc100: `registerReadableHanlder` (singular)?

sbc100: How about just `callback` rather than `notifyCallback` here?

src/lib/libpipefs.js
```diff
@@ -21,6 +21,19 @@ addToLibrary({
         // able to read from the read end after write end is closed.
         refcnt : 2,
         timestamp: new Date(),
+        readableHandlers: [],
```
sbc100, line 25: Do we need to support more than one handler at a time?   When/how could that happen?

src/lib/libpipefs.js
```diff
@@ -92,6 +105,7 @@ addToLibrary({
           }
         }
 
+        pipe.registerReadableHanlders(notifyCallback);
```
sbc100: How about putting the `if (notifyCallback)` check here rather than in `registerReadableHanlder`?

src/lib/libsyscall.js
```diff
@@ -542,31 +607,34 @@ var SyscallsLibrary = {
     FS.chdir(stream.path);
     return 0;
   },
+#if PROXY_TO_PTHREAD
```
sbc100: Isn't it the case that any pthread should be able block, not just the main / proxied thread?

i.e. should we not allow this blocking whenever we are not on the main browser thread?

In other words should set `__proxy: 'none'` whenever PTHREADS is set?

src/lib/libpipefs.js
ktock: Thank you for the feedback. Fixed this.

ktock: Fixed.

src/lib/libpipefs.js
ktock: > Do we need to support more than one handler at a time? When/how could that happen?

This can happen when polling one fd from multiple threads. I've added a test that reproduces this situation (`test_select_in_threads`).

src/lib/libpipefs.js
ktock: Fixed.

src/lib/libsyscall.js
ktock: > Isn't it the case that any pthread should be able block, not just the main / proxied thread?
> 
> i.e. should we not allow this blocking whenever we are not on the main browser thread?

Any thread can block using select's event and timeout args. I've added this test to `test_select_in_threads`.

> In other words should set __proxy: 'none' whenever PTHREADS is set?

I think PROXY_TO_PTHREAD is mandatory for now because if we don't have it we need to change the behaviour of `select` on the main thread (on the main browser worker where emscripten_proxy_sync_with_ctx can't be used) and other threads. Please let me know if I'm missing something.

sbc100: Exactly.   I think the behaviour should depend on `ENVIRONMENT_IS_PTHREAD` and should not require `PROXY_TO_PTHREAD`.   All `PROXY_TO_PTHREAD` does is create a pthread to run the main function on.

Even in `PROXY_TO_PTHREAD` mode its still possible for call select from the main browser thead.

For example could have an html `onClick` hander that calls `Module.doSelect()`.   

So in either case you need to handle the case when the call is made form the main browser thread and in that case blocking is not possible I guess.

src/lib/libsyscall.js
```diff
@@ -577,6 +645,37 @@ var SyscallsLibrary = {
...
+      }
+      cb.registerCleanupFunc = (f) => {
+        if (f != null) cleanupFuncs.push(f);
+      }
+      return cb;
+    }
+    if (timeoutInMillis > 0) {
+      setTimeout(() => makeNotifyCallback(-1)(0), timeoutInMillis);
+    }
+#endif
```
sbc100, line 683: On other words, can this block instead be written as:

```
#if PTHREADS
  if (ENVIRONMENT_IS_PTHREAD) {
     ...
  }
#endif
```

system/lib/pthread/proxying.c
```diff
@@ -658,3 +658,36 @@ double _emscripten_run_js_on_main_thread(int func_index,
...
+
+int _emscripten_proxy_newselect(int n, void *rfds, void *wfds, void *efds, void *tv) {
+  em_proxying_queue* q = emscripten_proxy_get_system_queue();
+  pthread_t target = emscripten_main_runtime_thread_id();
+  proxied_select_t t = {.n = n, .rfds = rfds, .wfds = wfds, .efds = efds, .tv = tv};
+  if (!emscripten_proxy_sync_with_ctx(q, target, call_newselect, &t)) {
+    assert(false && "emscripten_proxy_sync failed");
+    return -1;
+  }
+  return t.result;
```
sbc100: This code which is specific to the select system call probably doesn't block in this file (which is the general-purpose proxying system)

test/core/test_pthread_select_timeout.c
```diff
@@ -0,0 +1,174 @@
...
+
+  sleep(2);
+  write(pipe_shared[1], t, strlen(t));
+
+  return NULL;
+}
+
+// Check if select can unblock on an event
+void test_unblock_select()
+{
```
sbc100: Open braces go on the same line as function name.

src/lib/libsyscall.js
ktock: Thanks for the suggestion. Fixed the patch not to rely on the PROXY_TO_PTHREAD macro. Now `__emscripten_proxy_newselect` is used only when `ENVIRONMENT_IS_PTHREAD` is set and `_newselect_js` enables the event notification only when the call is made and proxied from a worker (i.e. `ctx` is not null). I've not used `ENVIRONMENT_IS_PTHREAD` in `_newselect_js` because it doesn't seem to be set in the proxied call in the main browser worker.

system/lib/pthread/proxying.c
ktock: Moved this to `/system/lib/libc/proxying_select.c`.

test/core/test_pthread_select_timeout.c
ktock: Fixed.

ktock:
The failure in `ci/circleci: test-mac-arm64` looks like unrelated to this PR:

> This job was rejected because the resource class is unavailable

sbc100:
Circle CI should be fixed in https://github.com/emscripten-core/emscripten/pull/25955

src/lib/libpipefs.js
```diff
@@ -21,6 +21,21 @@ addToLibrary({
         // able to read from the read end after write end is closed.
         refcnt : 2,
         timestamp: new Date(),
+        readableHandlers: [],
+        registerReadableHanlder: (callback) => {
```
sbc100: Hanlder -> Handler

src/lib/libsyscall.js
```diff
@@ -102,6 +102,73 @@ var SyscallsLibrary = {
...
+    getTimeoutInMillis(timeout) {
+      // select(2) is declared to accept "struct timeval { time_t tv_sec; suseconds_t tv_usec; }".
+      // However, musl passes the two values to the syscall as an array of long values.
+      // Note that sizeof(time_t) != sizeof(long) in wasm32. The former is 8, while the latter is 4.
+      // This means using "C_STRUCTS.timeval\.tv_usec" leads to a wrong offset.
+      // So, instead, we use POINTER_SIZE.
+      var tv_sec = ({{{ makeGetValue('timeout', 0, 'i32') }}}),
+          tv_usec = ({{{ makeGetValue('timeout', POINTER_SIZE, 'i32') }}});
+      return (tv_sec + tv_usec / 1000000) * 1000;
+    },
```
sbc100: Can we avoid this by passing tv_sec and tv_usec as two separate arguments?

You can use `uint64_t` for `tv_sec`  and then flag that function as `__i53abi` to have to comethrough as a JS number.

sbc100: Oh wait, I see this is preexisting... we can leave this for now and do it as a followup maybe

src/lib/libpipefs.js
```diff
@@ -21,6 +21,21 @@ addToLibrary({
...
+          });
+          pipe.readableHandlers.push(callback);
+        },
+        notifyReadableHanders: () => {
+          while (pipe.readableHandlers.length > 0) {
+            const cb = pipe.readableHandlers.shift();
+            if (cb) cb({{{ cDefs.POLLRDNORM }}} | {{{ cDefs.POLLIN }}});
+          }
+          pipe.readableHandlers = [];
+        }
```
sbc100, line 39: Can all the new code here be behind `if PTHREADS` ?   i.e. in single threaded builds non of this is possible, right?

src/lib/libsyscall.js
```diff
@@ -102,6 +102,73 @@ var SyscallsLibrary = {
...
+            {{{ makeSetValue('writefds', '0', 'dstWriteLow', 'i32') }}};
+            {{{ makeSetValue('writefds', '4', 'dstWriteHigh', 'i32') }}};
+          }
+          if (exceptfds) {
+            {{{ makeSetValue('exceptfds', '0', 'dstExceptLow', 'i32') }}};
+            {{{ makeSetValue('exceptfds', '4', 'dstExceptHigh', 'i32') }}};
+          }
+        }
+      };
+    },
```
sbc100: Can you make these free floating functions rather than having them be part of `SYSCALLS`.. I believe this makes it much easier to eliminate them when they are not used.

i.e. declare them as `$parseSelectFDSet`, mark them as `__internal: true` and add them to the `__deps` of the functions that depend on them?

src/lib/libpipefs.js
ktock: Fixed.

src/lib/libpipefs.js
ktock: Moved them to behind `if PTHREADS`.

src/lib/libsyscall.js
ktock: Fixed.

ktock:
> Can you run the code size tests (./test/runner codesize)?
>
> I think a good goal here would be to try to avoid code size regressions (within a few bytes) except for pthread-enabled programs.


I've run the codesize test and the following `FAIL`s. 

### `test_codesize_hello_dylink_all (test_codesize.codesize)`

```
Running codesize test: hello_world.c: ['-O3', '-sMAIN_MODULE'] True False
a.out.js: size=245936, expected=245673, delta=263 (+0.11%)
a.out.nodebug.wasm: size=574302, expected=573748, delta=554 (+0.10%)
total: size=820238, expected=819421, delta=817 (+0.10%)
Oops, overall generated code size regressed by 817 bytes!
If this is expected, rerun the test with --rebaseline to update the expected sizes
```

### `test_codesize_files_wasmfs (test_codesize.codesize)`

This seems to be an improvement according to the log.

```
Running codesize test: files.cpp: ['-O3', '-sWASMFS'] True False
a.out.js: size=5465, expected=5465
a.out.js.gz: size=2576, expected=2576
a.out.nodebug.wasm: size=50212, expected=50590, delta=-378 (-0.75%)
a.out.nodebug.wasm.gz: size=18067, expected=18089, delta=-22 (-0.12%)
total: size=55677, expected=56055, delta=-378 (-0.67%)
total_gz: size=20643, expected=20665, delta=-22 (-0.11%)
Hey amazing, overall generated code size was improved by 378 bytes!
If this is expected, rerun the test with --rebaseline to update the expected sizes
```

### `test_codesize_cxx_wasmfs (test_codesize.codesize)`

This seems to be an improvement according to the log.

```
Running codesize test: hello_libcxx.cpp: ['-O2', '-sWASMFS'] False False
a.out.js: size=7043, expected=7043
a.out.js.gz: size=3322, expected=3322
a.out.nodebug.wasm: size=169671, expected=170056, delta=-385 (-0.23%)
a.out.nodebug.wasm.gz: size=62949, expected=62932, delta=17 (+0.03%)
total: size=176714, expected=177099, delta=-385 (-0.22%)
total_gz: size=66271, expected=66254, delta=17 (+0.03%)
Hey amazing, overall generated code size was improved by 385 bytes!
If this is expected, rerun the test with --rebaseline to update the expected sizes

```

sbc100:
Can you run with `--rebaseline` and upload the results?  (Remember to use `emsdk install tot` to make sure you have the latest llvm and binaryen installed first).

system/lib/libc/proxying_select.c
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright 2021 The Emscripten Authors.  All rights reserved.
```
sbc100: 2025

test/test_core.py
```diff
@@ -9665,6 +9665,13 @@ def test_wasm_global(self, dynlink):
   def test_syscall_intercept(self):
     self.do_core_test('test_syscall_intercept.c')
 
+  def test_pthread_select_timeout(self):
+    self.do_runf('core/test_pthread_select_timeout.c', cflags=['-pthread', '-sPROXY_TO_PTHREAD=1', '-sEXIT_RUNTIME=1', '-Wno-pthreads-mem-growth'])
+
+  def test_select(self):
+    self.do_runf('core/test_select.c')
+    self.do_runf('core/test_select.c', cflags=['-pthread'])
```
sbc100: You use parameterize to do run this two different ways:

```
  @parameterized({                                                                                         
    '': ([],),                                                                                             
    'pthread': (['-pthread'],),                                    
  })    
  ```

File: test/core/test_pipe_select.c [hunk 14]
```diff
@@ -0,0 +1,24 @@
+#include <sys/select.h>
```
sbc100, line 1: Should we call this `test_pipe_select` maybe?

Or should we add this to the existing `test/unistd/pipe.c`?

test/core/test_select_blocking.c
```diff
@@ -0,0 +1,166 @@
+#include <sys/select.h>
```
sbc100, line 1: How about calling this test `test_select_blocking`?   Or maybe `test_pipe_select_blocking`?

system/lib/wasmfs/syscalls.cpp
```diff
@@ -1780,4 +1780,8 @@ int __syscall__newselect(int nfds,
   return -ENOMEM;
 }
 
+int _newselect_js(void* ctx, void* arg, int n, void *rfds, void *wfds, void *efds, void *tv) {
+  return -ENOSYS;
+}
```
sbc100, line 1853: Do we need this in addition to the dummy __syscall__newselect above?

sbc100:
I'm adding some basic testing of select (with no timeout) to the existing pipe test here: https://github.com/emscripten-core/emscripten/pull/25965

sbc100:
> I'm adding some basic testing of select (with no timeout) to the existing pipe test here: #25965

While working on #25965 I ended up implementing select() for WasmFS and did it in terms of `poll()`.. which makes me wonder if we should be doing that for the normal FS too?

When this change would actually be a change to poll rather than select?   Does that sound reasonable?

sbc100:
If you think that is too much to change at this point, we could land this change first and then consider a refactor later.

src/lib/libsyscall.js
```diff
@@ -104,6 +104,75 @@ var SyscallsLibrary = {
...
+  $getTimeoutInMillis: (timeout) => {
+    // select(2) is declared to accept "struct timeval { time_t tv_sec; suseconds_t tv_usec; }".
+    // However, musl passes the two values to the syscall as an array of long values.
+    // Note that sizeof(time_t) != sizeof(long) in wasm32. The former is 8, while the latter is 4.
+    // This means using "C_STRUCTS.timeval\.tv_usec" leads to a wrong offset.
+    // So, instead, we use POINTER_SIZE.
+    var tv_sec = ({{{ makeGetValue('timeout', 0, 'i32') }}}),
+        tv_usec = ({{{ makeGetValue('timeout', POINTER_SIZE, 'i32') }}});
+    return (tv_sec + tv_usec / 1000000) * 1000;
+  },
```
sbc100: This function can be completely removed once https://github.com/emscripten-core/emscripten/pull/25968 lands.

system/lib/libc/proxying_select.c
ktock: Fixed.

test/test_core.py
ktock: Fixed.

test/core/test_pipe_select.c
ktock: Renamed this to test_pipe_select.

test/core/test_select_blocking.c
ktock: Renamed this to test_select_blocking.

system/lib/wasmfs/syscalls.cpp
ktock: Yes. Otherwise we get the following test failure in `other.test_closure_full_js_library_wasmfs`:

> error: undefined symbol: _newselect_js (referenced by root reference (e.g. compiled C/C++ code))

ktock:
> Can you run with --rebaseline and upload the results? (Remember to use emsdk install tot to make sure you have the latest llvm and binaryen installed first).

Updated the JSON file using --rebaseline.

src/lib/libsyscall.js
ktock: Thanks for working on the PR. Removed this function.

ktock:
> While working on #25965 I ended up implementing select() for WasmFS and did it in terms of poll().. which makes me wonder if we should be doing that for the normal FS too?
> 
> When this change would actually be a change to poll rather than select? Does that sound reasonable?
>
> If you think that is too much to change at this point, we could land this change first and then consider a refactor later.

Thanks for working on the PR. Currently, `__syscall_poll()` doesn't implement the timeout support so we first need to fix this. I think this exceeds the scope of this PR and should be done in a separated refactoring PR.

src/lib/libsyscall.js
```diff
@@ -543,37 +600,84 @@ var SyscallsLibrary = {
...
+#endif
+    return doNewselect(null, null, nfds, readfds, writefds, exceptfds, timeoutInMillis);
+  },
+  _newselect_js__i53abi: true,
+  _newselect_js__proxy: 'none',
+  _newselect_js__deps: ['$doNewselect'],
+  _newselect_js: (ctx, arg, nfds, readfds, writefds, exceptfds, timeoutInMillis) => {
+    return doNewselect(ctx, arg, nfds, readfds, writefds, exceptfds, timeoutInMillis);
+  },
+  $doNewselect__proxy: 'none',
```
sbc100: This looks like its just a forwarding function.   Can we merge `doNewselect` and `_newselect_js`?   i.e.  can we remove doNewselect and just call it `_newselect_js`?

src/lib/libsyscall.js
```diff
@@ -543,37 +600,84 @@ var SyscallsLibrary = {
...
+  ],
+  $doNewselect: (ctx, arg, nfds, readfds, writefds, exceptfds, timeoutInMillis) => {
     // readfds are supported,
     // writefds checks socket open status
     // exceptfds are supported, although on web, such exceptional conditions never arise in web sockets
     //                          and so the exceptfds list will always return empty.
-    // timeout is supported, although on SOCKFS and PIPEFS these are ignored and always treated as 0 - fully async
+    // timeout is supported, although on SOCKFS these are ignored and always treated as 0 - fully async
+    // and PIPEFS supports timeout only when the select is called from a worker.
 #if ASSERTIONS
```
sbc100, line 641: Can we asssert that we are not running on a thread here?  i.e. `doNewselect` must be called on the main thread only right?

system/lib/libc/proxying_select.c
```diff
@@ -0,0 +1,43 @@
...
+int _emscripten_proxy_newselect(int n, void *rfds, void *wfds, void *efds, int64_t timeout) {
+  em_proxying_queue* q = emscripten_proxy_get_system_queue();
+  pthread_t target = emscripten_main_runtime_thread_id();
+  proxied_select_t t = {.n = n, .rfds = rfds, .wfds = wfds, .efds = efds, .timeout = timeout};
+  if (!emscripten_proxy_sync_with_ctx(q, target, call_newselect, &t)) {
+    assert(false && "emscripten_proxy_sync failed");
+    return -1;
+  }
+  return t.result;
+}
```
sbc100, line 43: So it seems like the whole point of this custom proxying is to allow the main thread to delay the calling of `_emscripten_proxy_newselect_finish` until later?  i.e. after `doNewselect` returns.

I'm happy to see this change land but I wonder if we could do this without that need for custom proxying?  I.e. I wonder if we could return a promise from `doNewselect` and have the automatic proxying system just work in that case?   Allowing all async proxyied functions to work.

test/sockets/test_sockets_echo_client.c
```diff
@@ -67,13 +67,16 @@ void main_loop() {
   fd_set fdr;
   fd_set fdw;
   int res;
+  struct timeval tv;
```
sbc100: I think you can do `struct timeval tv = {0}` here.   Maybe also add a comment "// Timeout of zero means don't block.  Emscripten doesn't support blocking select in the general case"

test/test_core.py
```diff
@@ -9665,6 +9665,16 @@ def test_wasm_global(self, dynlink):
   def test_syscall_intercept(self):
     self.do_core_test('test_syscall_intercept.c')
 
+  def test_select_blocking(self):
+    self.do_runf('core/test_select_blocking.c', cflags=['-pthread', '-sPROXY_TO_PTHREAD=1', '-sEXIT_RUNTIME=1', '-Wno-pthreads-mem-growth'])
```
sbc100: I don't think you should need `-Wno-pthreads-mem-growth` since memory growth is not enabled here, right?

test/core/test_select_blocking.c
```diff
@@ -0,0 +1,166 @@
...
+  assert(select(pipe_a[0] + 1, &readfds, NULL, NULL, &tv) == 0);
+  gettimeofday(&end, NULL);
+  assert((end.tv_sec - begin.tv_sec) * 1000000 + end.tv_usec - begin.tv_usec >= 1000000);
+
+  close(pipe_a[0]); close(pipe_a[1]);
+}
+
+int pipe_shared[2];
+
+void *wakeup_after_2s(void * arg) {
```
sbc100: Maybe called this `write_after_2s`

test/core/test_select_blocking.c
```diff
@@ -0,0 +1,166 @@
...
+}
+
+int main() {
+  test_select_in_threads();
+  test_timeout_without_fds();
+  test_timeout_with_fds_without_events();
+  test_unblock_select();
+  test_ready_fds();
+  return 0;
+}
```
sbc100, line 166: Out of interest have you tried running this test code on a normal UNIX system?   Does that test pass on linux, for example?

sbc100 (approved):

sbc100:
BTW, can I ask what you use case is?  Are you stuff with a PTY?

test/core/test_select_blocking.c
ktock: Yes, it passes on Linux.

test/core/test_select_blocking.c
ktock: Fixed.

ktock:
> BTW, can I ask what you use case is? Are you stuff with a PTY?

I've been working on porting QEMU to Wasm. It relies on the select and pipes for managing events and timers.

https://wiki.qemu.org/ChangeLog/10.1

> Experimental support for compiling to WASM using Emscripten.

Status: https://github.com/ktock/qemu-wasm?tab=readme-ov-file#status-of-upstreaming

sbc100:
> > BTW, can I ask what you use case is? Are you stuff with a PTY?
> 
> I've been working on porting QEMU to Wasm. It relies on the select and pipes for managing events and timers.
> 
> https://wiki.qemu.org/ChangeLog/10.1
> 
> > Experimental support for compiling to WASM using Emscripten.
> 
> Status: https://github.com/ktock/qemu-wasm?tab=readme-ov-file#status-of-upstreaming

Thats awesome!   If there are any other emscripten improvements that would help you please let us know.  

I'm excited to set better support for these things.   Once this lands I think I'd like to get poll working too, and convert select to be sit on top of poll.  After that we should get this stuff working with WASMFS (that next generation FS).

sbc100:
There seem to be a couple of issues with the tests here.

Firstly I've seen some new flakes which I guess are timing related?

Secondly they seems to be failing under lsan/asan:https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket/8695064874929935185/+/u/Emscripten_testsuite__ASan_/stdout

```
emcc: error: -pthread + ALLOW_MEMORY_GROWTH may run non-wasm code slowly, see https://github.com/WebAssembly/design/issues/1271 [-Wpthreads-mem-growth] [-Werror]
[240/1180] None
None
test_select_blocking (test_core.asan.test_select_blocking) ... FAIL
```

maybe an easy fix?

ktock:
> There seem to be a couple of issues with the tests here.
> 
> Firstly I've seen some new flakes which I guess are timing related?
> 
> Secondly they seems to be failing under lsan/asan:https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket/8695064874929935185/+/u/Emscripten_testsuite__ASan_/stdout
> 
> emcc: error: -pthread + ALLOW_MEMORY_GROWTH may run non-wasm code slowly, see https://github.com/WebAssembly/design/issues/1271 [-Wpthreads-mem-growth] [-Werror]
> [240/1180] None
> None
> test_select_blocking (test_core.asan.test_select_blocking) ... FAIL
> 
> maybe an easy fix?

Sorry for the slow reply, and thank you for the follow-up PRs. I've also seen the ongoing PR #25990 and added comments there as well.  Please let me know if there is something I can do for these features.

PR #25525, opened 2025-10-08
Fix emcc color output regression from #25502 on Windows.

juj:
This was lost in refactor of #25502 .

The reason I did not see the regression in my comment https://github.com/emscripten-core/emscripten/pull/25502#issuecomment-3373780715 is that calling `subprocess.run(['C:/emsdk/node/22.16.0_64bit/bin/node.exe', 'C:\\emsdk\\emscripten\\main\\test\\hello_world.js'])` has a side effect of also leaking/enabling `ENABLE_VIRTUAL_TERMINAL_PROCESSING` for the current terminal, so by the time that the parallel test harness was running, `node` subprocess had been executed and vterm processing was activated.

### DIFF
--- a/tools/colored_logger.py
+++ b/tools/colored_logger.py
@@ -13,6 +13,7 @@ import sys
 import logging
 from functools import wraps
 
+
 # ANSI colors
 RED = 1
 GREEN = 2


tools/colored_logger.py
```diff
@@ -101,6 +101,14 @@ def new(*args):
 
 def enable(force=False):
   global color_enabled
+
+  if sys.platform.startswith('win'):
+    kernel32 = ctypes.windll.kernel32
+    handle = kernel32.GetStdHandle(-11)  # STD_OUTPUT_HANDLE = -11
+    mode = ctypes.c_uint32()
+    kernel32.GetConsoleMode(handle, ctypes.byref(mode))
+    kernel32.SetConsoleMode(handle, mode.value | 0x0004)  # ENABLE_VIRTUAL_TERMINAL_PROCESSING
```
sbc100: if that parent process doesn't already have color enabled don't we want to honor that?

`ansi_color_available` checks if `ENABLE_VIRTUAL_TERMINAL_PROCESSING` set and honors the current setting. 

Under what circumstances would `ENABLE_VIRTUAL_TERMINAL_PROCESSING` not already be set?   Do we want to always set it?

tools/colored_logger.py
```diff
@@ -101,6 +101,14 @@ def new(*args):
 
 def enable(force=False):
   global color_enabled
+
+  if sys.platform.startswith('win'):
+    kernel32 = ctypes.windll.kernel32
+    handle = kernel32.GetStdHandle(-11)  # STD_OUTPUT_HANDLE = -11
```
sbc100: `STD_OUTPUT_HANDLE` and `ENABLE_VIRTUAL_TERMINAL_PROCESSING` are already defined in `ansi_color_available`, maybe use shared definitions?

tools/colored_logger.py
sbc100: What happens here if stdout is not a tty?  Will this call fail?

juj: > if that parent process doesn't already have color enabled don't we want to honor that?

Launching any `subprocess.run('node')` call from Python already has the effect of "dishonoring" color setting on Windows, but it just happens too late, so first prints don't get color.

I presume cmd.exe terminals don't by default have color enabled, so if we did honor it, then we wouldn't have colors on Windows anymore.

tools/colored_logger.py
juj: ok

tools/colored_logger.py
sbc100: I see, if we going to always try to modify the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` setting then I think we can/should just do that in the existing `ansi_color_available` function.   

It should call `SetConsoleMode` if `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is not already part of the mode?

sbc100: We should also flip the condition so that it reads `if ansi_color_available() or force:` .. since we want `ansi_color_available` to always run even in `force` mode (since its now has side effects).

juj: > What happens here if stdout is not a tty? Will this call fail?

I don't know. I verified that file redirection at least does not get ANSI codes:

image

If this is problematic via some other use case, then we should probably look at using the earlier code.

juj: The previous code has this block:

```
  if not kernel32.GetConsoleMode(stdout_handle, ctypes.byref(console_mode)):
    # Handle error if GetConsoleMode fails
    return False
```
when would GetConsoleMode be expected to fail?

sbc100: I assume when stdout is not a console at all but a file?

sbc100:
I think them main thing I overlooked was thinking that `GetConsoleMode` would be enough, but we actually also need to try `SetConsoleMode`,  but I think this can all happen in the same function/block.

sbc100:
Looks like llvm just does the "Set" part:  https://github.com/llvm/llvm-project/blob/774ffe5cce7392e6b4e29c83e725337727c2c994/llvm/lib/Support/Windows/Process.inc#L315-L326

But I think we also want to know if the "Set" succeeded, because if it didn't we don't want to output the ANSI codes (not without `force` anyway).

tools/colored_logger.py
```diff
@@ -101,9 +80,23 @@ def new(*args):
...
+
+  if sys.platform.startswith('win'):
+    # Constants from the Windows API
+    STD_OUTPUT_HANDLE = -11
+    ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004
+
+    kernel32 = ctypes.windll.kernel32
+    handle = kernel32.GetStdHandle(STD_OUTPUT_HANDLE)
+    mode = ctypes.c_uint32()
+    if not kernel32.GetConsoleMode(handle, ctypes.byref(mode)):
```
sbc100, line 92: We still want to support `force` here too I think rather than returning early.

Can we keep this as a separate function maybe?   Then we can check for availability separately from the `force` handling which happens here in this function?

I know `ansi_color_available` might not be a great name of side-effecting function.. but I still kind of like it being separate.   Maybe it can be called something like `try_enable_ansi_colors`?

juj: Hmm I am not 100% sure how you mean. Could you propose a patch on top of this PR, or as a separate PR?

sbc100: `force` here means "write ansi codes even if the terminal doesn't support them"

tools/colored_logger.py
```diff
@@ -101,9 +80,23 @@ def new(*args):
...
+      return
+    kernel32.SetConsoleMode(handle, mode.value | ENABLE_VIRTUAL_TERMINAL_PROCESSING)
```
sbc100, line 94: Don't we want to check if this calls fails?  i.e. isn't that how to detect if the termimal actually supports color codes?   IIRC this is the line that will fail prior to windows 10 where ANSI started being supported.

sbc100:
OK, I do have access to a windows machine, its just tricky to get to it.   I can take a look at his there to make sure my ideas hold true.

juj:
Closing for a later fix.

PR #25528, opened 2025-10-08
Safari test skips

juj:
Improve Safari test harness skips to be aware of Safari version.

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -14,7 +14,7 @@ import subprocess
 import time
 import unittest
 import zlib
-from functools import wraps, lru_cache
+from functools import wraps
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from pathlib import Path
 from urllib.request import urlopen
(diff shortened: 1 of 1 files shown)


test/test_browser.py
```diff
@@ -141,6 +144,19 @@ def is_swiftshader(_):
   return is_chrome() and '--use-gl=swiftshader' in common.EMTEST_BROWSER
 
 
+def get_safari_version():
+  plist_path = os.path.join(common.EMTEST_BROWSER.strip(), 'Contents', 'version.plist')
+  if not os.path.isfile(plist_path):
+    return UNSUPPORTED
```
sbc100, line 150: How about `if !is_safari(): return UNSUPPORTED` on line 0 of this function.

Then you don't need this change since `version.plist` should always exist when the current browser is safari, right?  Or at least if `is_safari()` is true this function should just crash if the plist is missing since it means something is wrong.

juj: ok

juj: Yeah at least all the versions I am running on do have the version.plist.

sbc100: Right, if we find other types of safari we probably want crash here in investigate.

test/test_browser.py
```diff
@@ -141,6 +144,20 @@ def is_swiftshader(_):
   return is_chrome() and '--use-gl=swiftshader' in common.EMTEST_BROWSER
 
 
+@lru_cache(maxsize=1) # Memoize this function to only be called once during runtime
```
sbc100: You can use `@memoize` for this `(from tools.utils import memoize)`

juj:
Any more review, or good to go?

PR #25529, opened 2025-10-08
Fix enabling of windows color terminal processing

sbc100:
This was broken in #25502 and we only have limited testing of this features.  Specifically we only test that the ANSI codes are generated, not that they are successfully processed.

The change here is not only to check for
ENABLE_VIRTUAL_TERMINAL_PROCESSING mode but to actually set this mode.

Replaces #25525

### DIFF
--- a/tools/colored_logger.py
+++ b/tools/colored_logger.py
@@ -67,14 +67,16 @@ def ansi_color_available():
   kernel32 = ctypes.windll.kernel32
   stdout_handle = kernel32.GetStdHandle(STD_OUTPUT_HANDLE)
 
-  # Get the current console mode
   console_mode = ctypes.c_uint()
-  if not kernel32.GetConsoleMode(stdout_handle, ctypes.byref(console_mode)):
-    # Handle error if GetConsoleMode fails
+  # Attempt to enable ANSI color processing (ENABLE_VIRTUAL_TERMINAL_PROCESSING).
+  # Assume that failure of either GetConsoleMode or SetConsoleMode means that stdout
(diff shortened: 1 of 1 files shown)


tools/colored_logger.py
```diff
@@ -67,14 +67,16 @@ def ansi_color_available():
...
-  # Get the current console mode
   console_mode = ctypes.c_uint()
-  if not kernel32.GetConsoleMode(stdout_handle, ctypes.byref(console_mode)):
-    # Handle error if GetConsoleMode fails
+  # Attempt to enable ANSI color processing (ENABLE_VIRTUAL_TERMINAL_PROCESSING).
+  # Assume that failure of either GetConsoleMode or SetConsoleMode means that stdout
+  # is not attached to a terminal or that the terminal does not support this mode.
+  if kernel32.GetConsoleMode(stdout_handle, ctypes.byref(console_mode)) == 0:
+    return False
+  if kernel32.SetConsoleMode(stdout_handle, console_mode.value & ENABLE_VIRTUAL_TERMINAL_PROCESSING) == 0:
```
juj: Looks like this should be `console_mode.value | ENABLE_VIRTUAL_TERMINAL_PROCESSING` instead of `&`?

sbc100: Wow.. thats true, but i did specifically test this code on my windows machine, something strange must be going on.

sbc100: Ah I see, this state is global, so when you test it you need to continually open a new cmd.exe window.

sbc100:
Tested on a fresh cmd.exe window on windows.

PR #25532, opened 2025-10-08
Add `MUTABLE_GLOBALS` to feature_matrix

sbc100:
This is needed by dynamic linking since GOT entries are all imported as mutable globals.

The net effect of this is that you cannot actually target the oldest version of safari that we support (12) and with dynamically linked programs.  Specifying `-sLEGACY_VM_SUPPORT` will give you safari 13.1 instead of 12 if you also using `-sMAIN_MODULE`.

### DIFF
--- a/tools/feature_matrix.py
+++ b/tools/feature_matrix.py
@@ -31,6 +31,7 @@ OLDEST_SUPPORTED_NODE = 122209
 
 
 class Feature(IntEnum):
+  MUTABLE_GLOBALS = auto()
   NON_TRAPPING_FPTOINT = auto()
   SIGN_EXT = auto()
   BULK_MEMORY = auto()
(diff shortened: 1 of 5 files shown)


tools/feature_matrix.py
```diff
@@ -213,6 +220,8 @@ def apply_min_browser_versions():
     enable_feature(Feature.BULK_MEMORY, 'pthreads')
   elif settings.WASM_WORKERS or settings.SHARED_MEMORY:
     enable_feature(Feature.BULK_MEMORY, 'shared-mem')
+  if settings.RELOCATABLE:
```
kripken, line 223: Were you not hoping to remove this setting?

sbc100: Yes I will replace this with MAIN_MODULE as part of that change.

Issue #20983, opened 2023-12-26
Add a way to remove a single event listener


ypujante:
As far as I understand the current API in regards to events, you can add as many listeners as you want, but there is no way to remove just one.

Here is an example:

If I call:

```cpp
emscripten_set_resize_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW, window, false, callback1);
emscripten_set_resize_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW, window, false, callback2);
```

in order to remove the callback, you need to call

```cpp
emscripten_set_resize_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW, window, false, nullptr);
```

but then it will remove both callbacks as the [code](https://github.com/emscripten-core/emscripten/blob/85c099ab519d574c3ce1f6d3603eb74aabbdf604/src/library_html5.js#L191) shows:

```javascript
        if (eventHandler.callbackfunc) {
          eventHandler.eventListenerFunc = jsEventHandler;
          eventHandler.target.addEventListener(eventHandler.eventTypeString, jsEventHandler, eventHandler.useCapture);
          JSEvents.eventHandlers.push(eventHandler);
          JSEvents.registerRemoveEventListeners();
        } else {
          for (var i = 0; i < JSEvents.eventHandlers.length; ++i) {
            if (JSEvents.eventHandlers[i].target == eventHandler.target
             && JSEvents.eventHandlers[i].eventTypeString == eventHandler.eventTypeString) {
               JSEvents._removeHandler(i--);
             }
          }
        }
```

The `else` block is executed which indiscriminately remove all listeners that have the same target and event type.

The issue I am bumping into with this is that the glfw/emscripten/cpp port I am currently implementing, is setting a bunch of listeners for its own purpose. But the client code can also set its own listeners, the typical use case is the one I showed above to react on window resize to call `glfwSetSize` when the canvas size changes. But if the client then decides to remove its own listener, then it will remove the one from the library as well, thus crippling the proper functioning of the library.

I do realize that the API is named `emscripten_set_resize_callback` which implies that you are setting (vs adding) a callback, but the underlying code is adding a listener not replacing a current one.

I realize that changing the current APIs may break existing code, but maybe we could add a new set of APIs (and maybe deprecate the current one if that makes sense)... something like:

```cpp
int emscripten_add_XXXXX_callback(....);
void emscripten_remove_callback(id);
```

* instead of `set` it is `add`
[...]

PR #25535, opened 2025-10-09
Add a way to remove a single event listener

ypujante:
I know it has been quite a while since I opened #20983 but I finally got the time to work on it. I followed the discussion we had at the time and did the following:

- added `emscripten_remove_callback`
  - iterate over all handlers and only remove the one matching the arguments provided
- changed all event setters to pass `userData` and `eventTypeId` which are required by this API
- added a test that shows that it works

This first pass is to make sure that this would be an ok change to Emscripten and that the changes work (no failures introduced by the changes). I can obviously change whatever you want or drop it entirely if this is not desired. And can work on documentation/examples if that gets approval.

I believe it is a good addition because it allows to remove a previously set listener, which is not possible at the moment, as the APIs provided will remove all of the ones set (see discussion in the #20983 ticket).

Fixes: #20983


ypujante:
@sbc100 The only failing test is the one about code size, which I can address if you are satisfied with the direction of this PR.

In the grand scheme of things it is a pretty low risk PR since the only changes to existing code (17 functions in total) is passing an additional 2 fields in the EventHandler map that are not used prior to this PR. Of course it makes the resulting code bigger.

I added a test which does a basic test of the functionality by counting the number of event handlers registered. That being said, during development of such test, I was running it manually in the browser and making sure that the result of removing a handler was actually having an action on the underlying JavaScript event handler being present or not.

Let me know your thoughts.

Thanks

src/lib/libhtml5.js
```diff
@@ -721,6 +760,8 @@ var LibraryHTML5 = {
     var eventHandler = {
       target: findEventTarget(target),
       eventTypeString,
+      eventTypeId,
+      userData,
```
sbc100, line 739: I don't really love all the repetition here, or the fact that we need to store two extra fields just for this new API, but I guess it makes sense, and the repetitive nature of this file is kind of a pre-existing condition :)

sbc100:
@juj what do you think about adding this new API?

src/lib/libhtml5.js
ypujante: I agree. I don't love it either, but you are right, this code is replicated over and over (17 times!). I feel like the API could have been simplified in the first place (having just 1 function call for all the calls using the same callback type and providing the eventTypeId... which would have reduced, for example, 9 mouse related calls into just 1...). It is what is now unfortunately. So that is the only solution I can think of to be able to implement this new API.

ypujante: Let me correct what I just said in regards to the count. There are indeed 17 places where I had to change the code, but there are way many more "emscripten_set_xxx_callback". There are 17 types of events supported, so we could not really simplyif this since they all have different callback (function pointer) types.

The example I gave for mouse counts as just 1 on these 17 places because there are 9 "emscripten_set_mousexxx_callback" functions which all funnel into 1 `registerMouseEventCallback` function call. I feel like this API is way too verbose and having just one `emscripten_set_mouse_callback` with the eventTypeId as a parameter (which is exactly what the internal `registerMouseEventCallback` is) would have been plenty in the first place...

src/lib/libhtml5.js
```diff
@@ -412,6 +433,18 @@ var LibraryHTML5 = {
   },
 #endif
 
+  emscripten_remove_callback__proxy: 'sync',
+  emscripten_remove_callback__deps: ['$JSEvents', '$findEventTarget'],
+  emscripten_remove_callback: (target, userData, eventTypeId, callback) => {
```
juj: Hmm, should we key the `userData` pointer into the removal? I.e. it would be better for this API to be

```
emscripten_remove_callback: (target, eventTypeId, callback)
```

The rationale is that the `userData` pointer is generally just data to the function call, and not identifying. I.e. it is more of the 'value' part in a key->value association.

It might be some function local value in the caller (e.g. a boolean for some behavior), and might result in user having to store the `userData` field in some global to be able to unregister. So requiring users to know to match the `userData` value could be a bit tedious.

(If users did specifically want to differentiate unregistrations based on what was seemingly a different userData value, they could do that by separating to a different function with a trampoline)

Also, this function would be good to follow the same `_on_thread` model, so that it can be called from pthreads symmetric to how registering on pthreads works.

The semantics of function `emscripten_remove_callback`  are to remove all copies of callback registrations of the given function. I wonder if the function naming should somehow reflect that?

src/lib/libhtml5.js
juj: Indeed the original API shape is a bit more verbose than what I would like. Pardon, this was my very first API design into Emscripten way back in 2014.

I am not much worried about the repetition though, since it will closure minify to just a condensed `,a` in code size for each parameter.

One could create a map of ID->string to avoid carrying both ID and string here, but storing that map would take up more bytes than the added `,a` for each function. (and the map would not DCE individual fields out)

juj:
> @juj what do you think about adding this new API?

Looks good to me. I would consider removing the `userData` part from the unregistration, and also this would be good to support operation from pthreads to be symmetric/complete.

Apart from that, looks like a great addition.

ypujante:
@juj Thank you for the feedback

> Looks good to me. I would consider removing the `userData` part from the unregistration, 

Personally I do not have a strong opinion either way. What matters is that if my library (ex: glfw) sets a listener, I am able to remove it without removing other listeners set by the client code on the same target. The eventTypeId/callback combination should be enough to achieve this result (although it might make the code a bit more tedious when the caller wants to include userData as a differentiator). It will also simplify a bit the code and make it smaller since we don't need to store userData anymore. So if @sbc100 is fine with it I will remove it.

> and also this would be good to support operation from pthreads to be symmetric/complete.

At first I tried to implement support for pthread but it doesn't seem to make sense from my understanding of the code. pthread is only used to specify how to invoke the callback once it has been registered. Registering the callback itself (which is done by calling `JSEvents.registerOrRemoveHandler(eventHandler);`) has nothing to do with pthread. So I don't see why we need pthread to remove/unregister the callback. Am I missing something?

> The semantics of function emscripten_remove_callback are to remove all copies of callback registrations of the given function. I wonder if the function naming should somehow reflect that?

I suppose we could call it `emscripten_remove_callbacks` instead? Either way, I will also add this function to the documentation once approved to move forward to describe what it does.

From a schedule point of view, I am traveling so I won't be able to work on it for a couple weeks. I don't think there is urgency anyway ;)

ypujante:
I am done with traveling so I can work on this again. And I would like to wrap this up if at all possible. 

@sbc100 do you have any opinion on @juj feedback? Do you have an opinion on removing the `userData` parameter?

@juj Can you provide feedback on my comment about pthread?

Thanks

src/lib/libhtml5.js
sbc100: I think this makes sense to me, but in that case I think it would also be good it if we make it impossible to register the same callback function  for the same event with different userData.  

```
emscripten_set_click_callback(window, data1, 0, mycallback);   // success
emscripten_set_click_callback(window, data2, 0, mycallback);   // error `mycallback` is already registered as a click callback
```

If we didn't error in the second case, and allowed the same callback to be registred twice, there would be no way to unregister just one of them.

ypujante: * if we want to error when registering the same callback with a different user data, then it means we need to store the user data (which we don't prior to this changes)
* it also means that code that is currently working could fail after this change

I think what @juj suggested was to indeed un-register all callbacks that were registered with different data but with the same target/event_id/callback combination. That would have eliminated having to store the user data.

It sounds like you don't like this because then "there would be no way to unregister just one of them".

I am definitely not in favor of erroring when registering 2 identical callbacks with different user data, because from a developer point of view, this is absolutely not an error and I can imagine scenarios where that could be useful. 

So I guess the solution is to leave the API and changes as implemented in this PR. That way we can un-register a single callback (which IS the point of this PR). So I am personally in favor of this.

I will now work on updating the documentation and fixing the "Code Size" failure.

sbc100: > I am definitely not in favor of erroring when registering 2 identical callbacks with different user data, because from a developer point of view, this is absolutely not an error and I can imagine scenarios where that could be useful.

Can you give an example of when this might be useful?

ypujante: This a theoretical example, because I am not implementing it this way at the moment, but imagine for GLFW, you have 2 windows (represented in C by GLFWwindow pointers and canvases in HTML) and you need to set mouse move callbacks to track the mouse movements.

In order to detect moving the mouse outside the canvas while the button is clicked (something my library supports where others don't) you cannot set the callback on each canvas, it has to be set on the window... so the callbacks would be:

```cpp
emscripten_set_mousemove_callback(WINDOW, glfwWindow1, true, handle_move_callback);
emscripten_set_mousemove_callback(WINDOW, glfwWindow2, true, handle_move_callback);
```

Again I am not doing it this way, but that could be a way to implement it...

sbc100: Fair enough.  If we do want to officially support that then we probably also want to officially support unregistering them individually (sadly)

ypujante:
I think I am done on my side:

Based on the discussion, I left the code as-is with the `userData` parameter included. If we exclude it, then we need to disallow registering multiple callbacks with the same `userData`. It is still doable if you prefer to go down this route, but it means that it will break code relying on the fact that it is allowed now.

* I have changed the code
* Updated the documentation
* Updated the Changelog
* Fixed the size (after many attempts...)
* All tests are green

Let me know if there is anything else or if this can be merged as-is.

ypujante:
@sbc100 No worries about the back and forth. I suppose that is the point of a PR.

So the "N" you are talking about is 35 in `html5.h` (and some additional in WebGL).

Honestly, I don't think it is very reasonable to add this kind of bloat to Emscripten to add this lacking feature (or fix a bug depending on how you look at it, since after all there is a way to remove callbacks right now, just that removing them, will remove more than you want, see [#20983](https://github.com/emscripten-core/emscripten/issues/20983) for details).

In addition, I feel like the design was way too granular. I understand the need/want to have a type safe API with the proper function pointers, but there could have been 1 "set_callback" API per type (ex: 1 for all mouse events, instead of 9 like there is...). And yes you could totally have a single generic set_callback api, you would just loose the type safety...

For these reasons, I feel like adding an additional 35+ calls just for this is somewhat not the right approach.

Having a single generic call to remove any callback is by far the least intrusive. Yes you loose on type safety, but in this instance it is really unnecessary to have it since the function pointer is never used as a type pointer to invoke, but as a comparison argument.

I agree with you that taken out of context the name `emscripten_remove_callback` is not a good name. I wished I had realized this sooner. I noticed that there is already a function called `emscripten_html5_remove_all_event_listeners` so maybe we could name this one `emscripten_html5_remove_event_listener` instead?

Finally I am wondering if the API should return an error in case the event listener being asked to be removed doesn't exist (right now it doesn't in this instance). That way it could somehow cover cases where the wrong callback/function pointer is being used accidentally (a poor type safety check in some way).

In any case, these are my opinion and observations. You have the final call, and I will be fine with any way you want to go. Even if you want to go down the 35+ way, it will be far more work, but I am fine doing it. There is also the other alternative to simply close this PR and forget about it. There is currently no way to remove a specific callback (which is what is being addressed by this PR), but beside me, it doesn't seem that it has bothered anybody else. So I suppose we could just not do it.

sbc100:
Agree that adding 35 new API calls also seems like a bad idea :-/

Maybe we should at at least add a new generic API for this?    i.e. if we also add a single API for registering callbacks then we could at least say that the old API is deprecated?     On the other hand we do have a general policy of avoiding adding API that do too much multi-plexing since they prevent DCE.   Perhaps that is good argument for allowing an asymmetry in this case.

ypujante:
I have no opinion on whether to add a generic API or not but I have no issue implementing it. From your last message it looks like the 35+ option is out of the debate.

So what do you want me to do:

1) Rename `emscripten_remove_callback` into `emscripten_html5_remove_event_listener` (or some other name / please specify)
2) Implement 1)  and add a generic API (like `emscripten_html5_set_event_listener`)
3) Close this PR
4) Something else? Please specify

Please let me know which option you want me to do.

sbc100:
I'm afraid I don't have an answer for you now.  I think some more thought/discussion needs to go into this.  Lets involve @juj and @kripken and see what they think.   

Am I right in thinking that you can use this downstream today (as an out-of-tree JS library function) until we resolve it here?

ypujante:
> Am I right in thinking that you can use this downstream today (as an out-of-tree JS library function) until we resolve it here?

Not really no, because it requires changing the code of 17+ functions to pass the information that is currently not stored and needed by `emscripten_html5_remove_event_listener`. So I cannot just add `emscripten_html5_remove_event_listener` as an out-of-tree JS library function.

But waiting and not having an answer right now until we get it right is fine. I am not blocked. This has been like this for my library since I released it: if a user register/unregister event(s) using any of these functions, it will break the functioning of the library.

Also note that adding this function will still not fix the issue that my library can call the new `emscripten_html5_remove_event_listener` (and do the right thing) while the user code calls `emscripten_set_resize_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW, window, false, nullptr);` to remove their listeners that would impact the ones set by my library because they would be removed as well (unless we deprecate and later forbid passing `nullptr` in these functions...)

In any case, I am good in waiting.

site/source/docs/api_reference/html5.h.rst
```diff
@@ -90,6 +90,50 @@ The ``useCapture`` parameter  maps to ``useCapture`` in `EventTarget.addEventLis
 
 Most functions return the result using the type :c:data:`EMSCRIPTEN_RESULT`. Zero and positive values denote success. Negative values signal failure. None of the functions fail or abort by throwing a JavaScript or C++ exception. If a particular browser does not support the given feature, the value :c:data:`EMSCRIPTEN_RESULT_NOT_SUPPORTED` will be returned at the time the callback is registered.
 
+Remove callback function
+------------------------
+
+In order to remove a callback, previously set via a ``emscripten_set_some_callback`` call, there is a dedicated and generic function for this purpose:
```
juj: Maybe slightly simpler phrasing:

```
In order to unregister a single event handler callback, call the following function:
```

site/source/docs/api_reference/html5.h.rst
```diff
@@ -90,6 +90,50 @@ The ``useCapture`` parameter  maps to ``useCapture`` in `EventTarget.addEventLis
...
+
+    EMSCRIPTEN_RESULT emscripten_remove_callback(
+      const char *target,   // ID of the target HTML element.
+      void *userData,       // User-defined data (passed to the callback).
+      int eventTypeId,      // The event type ID (EMSCRIPTEN_EVENT_XXX).
+      void *callback        // Callback function.
+    );
+
+
+The ``target``, ``userData`` and ``callback`` parameters are the same parameters provided in ``emscripten_set_some_callback`` with the only difference being that, since this function applies to all types of callbacks, the type of ``callback`` is ``void *``.
```
juj, line 108: Maybe after this paragraph, add something like

```
Note in particular that the value of `userData` will need to match with the call that was used to register the callback. If you are having trouble, double check the value of `userData`.
```

to try to reinforce that people will track the userData pointer as being relevant.

juj:
> Maybe we should at at least add a new generic API for this? i.e. if we also add a single API for registering callbacks then we could at least say that the old API is deprecated?

I think it would be good to avoid trying to eat too big pie here at once. @ypujante's use case is looking to be able to remove an individual callback, so it would be good to not snowball this into needing to redesign the whole API at once.

The name `emscripten_html5_remove_event_listener()` sounds good to me.

> > and also this would be good to support operation from pthreads to be symmetric/complete.
> 
> At first I tried to implement support for pthread but it doesn't seem to make sense from my understanding of the code. pthread is only used to specify how to invoke the callback once it has been registered. Registering the callback itself (which is done by calling `JSEvents.registerOrRemoveHandler(eventHandler);`) has nothing to do with pthread. So I don't see why we need pthread to remove/unregister the callback. Am I missing something?

Thanks for the note. I realize now that there is already a `emscripten_remove_callback__proxy: 'sync'` directive on the removal function handler, and that is indeed enough to make unregistering work from a pthread. Doesn't look like anything else is needed.

You had a good point at https://github.com/emscripten-core/emscripten/pull/25535/files#r2505004777 about why to keep the `userData` as part of the key to removing a callback. So it's a good idea to keep that.

So I would recommend:
- rename the function name to `emscripten_html5_remove_event_listener()`,
- keep `userData` as part of the key. Add a comment to the function docs to highlight this fact

That would work for me to land.

If we wanted to have a modern revision of the HTML5 event handler APIs that would have fewer registration functions, that can be a completely separate story from this one.

ypujante:
Thank you @juj for the review and comments. I am personally in full agreement with the approach. @sbc100 what do you think?

The only think I wanted to call out is the fact that this PR does not address the fact that a user can still call (for example) `emscripten_set_resize_callback` with `nullptr` for the function pointer to remove it. But as we know (that is the point of this PR), that it is wrong because it would remove **all** callbacks using the same target. I believe the solution is to not allow this anymore and instead use the new `emscripten_html5_remove_event_listener` function.

Obviously we could start by adding a deprecation warning when this scenario is detected. I could add this deprecation warning to this PR or open another one once this one is merged. Any preference?

sbc100:
> Thank you @juj for the review and comments. I am personally in full agreement with the approach. @sbc100 what do you think?
> 

Fair enough.

> The only think I wanted to call out is the fact that this PR does not address the fact that a user can still call (for example) `emscripten_set_resize_callback` with `nullptr` for the function pointer to remove it. But as we know (that is the point of this PR), that it is wrong because it would remove **all** callbacks using the same target. I believe the solution is to not allow this anymore and instead use the new `emscripten_html5_remove_event_listener` function.
> 
> Obviously we could start by adding a deprecation warning when this scenario is detected. I could add this deprecation warning to this PR or open another one once this one is merged. Any preference?

Lets consider that (breaking) change separately, as a potentially followup.

ypujante:
@sbc100 I have implemented the changes discussed.

I have reverted the code size changes because it was polluting this PR with 30+ file changes and the codesize checks were still failing (a bit of a mystery to me). I can see that the failing codesize checks states that it is a warning that can be ignored, so I have chosen to ignore it...

There is only one test failing during package installation as I [reported earlier](https://github.com/emscripten-core/emscripten/pull/25799#issuecomment-3533012835) and it is clearly not related to these changes.

The relevant test is passing:

```
# from test-browser-firefox
  test_html5_remove_event_listener (test_browser.browser) ... ok (0.000s)
```

Let me know if there is anything else you want me to do for this PR

sbc100:
To get the codesize expectation match can you try the following commands:

```
$ emsdk install tot
$ ./emcc --clear-cache
$ ./tools/maint/rebaseline_tests.py
```

If you run those command one `upstream/main` they should produce no changed.  If you run them on your branch it should produce a diff.

sbc100:
Once the codesize tests pass I think we can land this.

site/source/docs/api_reference/html5.h.rst
```diff
@@ -90,6 +90,52 @@ The ``useCapture`` parameter  maps to ``useCapture`` in `EventTarget.addEventLis
...
+
+      // 1. set callbacks for mouse down and mouse move
+      emscripten_set_mousedown_callback("#mydiv", 0, my_mouse_callback_1);
+      emscripten_set_mousedown_callback("#mydiv", (void *) 34, my_mouse_callback_2);
+      emscripten_set_mousemove_callback("#mydiv", 0, my_mouse_callback_1);
+
+      // 2. remove these callbacks
+      emscripten_html5_remove_event_listener("#mydiv", 0, EMSCRIPTEN_EVENT_MOUSEDOWN, my_mouse_callback_1);
+      emscripten_html5_remove_event_listener("#mydiv", (void *) 34, EMSCRIPTEN_EVENT_MOUSEDOWN, my_mouse_callback_2);
+      emscripten_html5_remove_event_listener("#mydiv", 0, EMSCRIPTEN_EVENT_MOUSEMOVE, my_mouse_callback_1);
```
sbc100, line 138: I still don't love the asymmetry to the API design, but I get that adding 30+ `emscripten_clear_xxx_callback` API calls probably isn't great either.

src/lib/libhtml5.js
```diff
@@ -201,6 +201,25 @@ var LibraryHTML5 = {
...
+        console.dir(eventHandler);
+#endif
+        return {{{ cDefs.EMSCRIPTEN_RESULT_UNKNOWN_TARGET }}};
+      }
+      for (var i = 0; i < JSEvents.eventHandlers.length; ++i) {
+        if (JSEvents.eventHandlers[i].target === eventHandler.target
+          && JSEvents.eventHandlers[i].eventTypeId === eventHandler.eventTypeId
+          && JSEvents.eventHandlers[i].callbackfunc === eventHandler.callbackfunc
+          && JSEvents.eventHandlers[i].userData === eventHandler.userData) {
+          JSEvents._removeHandler(i--);
```
sbc100: Can you just `return {{{ cDefs.EMSCRIPTEN_RESULT_SUCCESS }}};` here?

And should we `return {{{ cDefs.EMSCRIPTEN_RESULT_INVALID_PARAM }}};`  below if its not found?

src/lib/libhtml5.js
```diff
@@ -201,6 +201,25 @@ var LibraryHTML5 = {
...
 
+    removeSingleHandler(eventHandler) {
+      if (!eventHandler.target) {
+#if ASSERTIONS
+        err('removeSingleHandler: the target element for event handler registration does not exist, when processing the following event handler registration:');
+        console.dir(eventHandler);
+#endif
+        return {{{ cDefs.EMSCRIPTEN_RESULT_UNKNOWN_TARGET }}};
+      }
+      for (var i = 0; i < JSEvents.eventHandlers.length; ++i) {
```
sbc100: `for (var [i, handler] of JSEvents.eventHandlers.entries()) {` ..  would avoid repeating `JSEvents.eventHandlers[i]` in the loop?

src/lib/libhtml5.js
```diff
@@ -201,6 +201,25 @@ var LibraryHTML5 = {
       return {{{ cDefs.EMSCRIPTEN_RESULT_SUCCESS }}};
     },
 
+    removeSingleHandler(eventHandler) {
+      if (!eventHandler.target) {
+#if ASSERTIONS
+        err('removeSingleHandler: the target element for event handler registration does not exist, when processing the following event handler registration:');
```
sbc100: Maybe just remove this ASSERTIONS block?

Also, maybe EMSCRIPTEN_RESULT_INVALID_PARAM would be a better return code here?

test/test_html5_remove_event_listener.c
```diff
@@ -0,0 +1,136 @@
...
+  return 0;
+}
+
+bool mouse_callback_1(int eventType, const EmscriptenMouseEvent *e, void *userData) {
+  printf("mouse_callback_1: eventType=%d, userData=%s\n", eventType, userDataToString(userData));
+  return 0;
+}
+
+void checkCount(int count)
+{
```
sbc100: Open brace on the line above

test/test_html5_remove_event_listener.c
```diff
@@ -0,0 +1,136 @@
...
+  int eventHandlersCount = EM_ASM_INT({ return JSEvents.eventHandlers.length; });
+  printf("Detected [%d] handlers\n", eventHandlersCount);
+  assert(count == eventHandlersCount);
+}
+
+void test_done() {}
```
sbc100: unused?

test/test_html5_remove_event_listener.c
```diff
@@ -0,0 +1,136 @@
...
+
+// Report API failure
+#define TEST_RESULT(x) if (ret != EMSCRIPTEN_RESULT_SUCCESS) printf("%s returned %s.\n", #x, emscripten_result_to_string(ret));
+
+// Like above above but also assert API success
+#define ASSERT_RESULT(x) TEST_RESULT(x); assert(ret == EMSCRIPTEN_RESULT_SUCCESS);
+
+char const *userDataToString(void *userData) {
+  return userData ? (char const *) userData : "nullptr";
+}
```
sbc100: Can this function simply be removed?

I think `printf` can handler printf NULL pointers as `nullptr` already so it should not be needed?

I.e. can you can just cast to userData to `char*` in the call to printf ?

src/lib/libhtml5.js
ypujante: This code is 100% copied from `registerOrRemoveHandler`

```cpp
    registerOrRemoveHandler(eventHandler) {
      if (!eventHandler.target) {
#if ASSERTIONS
        err('registerOrRemoveHandler: the target element for event handler registration does not exist, when processing the following event handler registration:');
        console.dir(eventHandler);
#endif
        return {{{ cDefs.EMSCRIPTEN_RESULT_UNKNOWN_TARGET }}};
      }
```

I can do it differently but I was trying to be consistent...

src/lib/libhtml5.js
ypujante: Same comment as above: it is a straight copy from `registerOrRemoveHandler` but I can implement your suggestion if you wish

src/lib/libhtml5.js
ypujante: I like this suggestion a lot because then there is a way to know whether the handler was removed successfully or not. I will make the change

ypujante:
@sbc100 I made all the changes you asked for and tweaked the test since now we can test when parameters are invalid.

All the tests pass except the code size checks. I don't understand what I am doing wrong. I follow your instructions exactly and somehow the end result does not pass. If I rerun the command there is no change.

test/test_html5_remove_event_listener.c
ypujante: Removed

src/lib/libhtml5.js
ypujante: I ended up simplifying the code entirely and only returning `EMSCRIPTEN_RESULT_INVALID_PARAM` as you suggested

sbc100:
I can download the PR and upload the matching code size results for you.

ypujante:
> I can download the PR and upload the matching code size results for you.

That would be awesome. I wished I understood why it doesn't work on my machine as I am following the exact steps you are describing (including clearing the cache which I believe is very important). Does it matter if my regular install of emsdk is not the latest released one? (I cannot move to 4.0.18+ because you yanked support for `-sWEB_GPU` after only a few months of deprecating it and ImGui is currently not supporting it...)

sbc100:
> > I can download the PR and upload the matching code size results for you.
> 
> That would be awesome. I wished I understood why it doesn't work on my machine as I am following the exact steps you are describing (including clearing the cache which I believe is very important). Does it matter if my regular install of emsdk is not the latest released one? (I cannot move to 4.0.18+ because you yanked support for `-sWEB_GPU` after only a few months of deprecating it and ImGui is currently not supporting it...)

As long as your emscripten checkout is configured to point the `tot` release of emsdk it should work.   I normally do this by setting `EM_CONFIG=/path/to/emsdk/.emscripten` before running clear cache + rebase.

ypujante:
My setup is the following:

- emsdk is installed under `/usr/local/emsdk`
- as a result the "latest" version ends up being under `/usr/local/emsdk/upstream/emscripten`
- the branch/pr I am working on is under `/usr/local/emsdk/emscripten/main`

1) If I open a shell, go to my PR folder and type `./tools/maint/rebaseline_tests.py` I get this error:

```text
  File "/usr/local/emsdk/emscripten/main/test/runner.py", line 39, in <module>
    import browser_common
  File "/usr/local/emsdk/emscripten/main/test/browser_common.py", line 23, in <module>
    import psutil
ModuleNotFoundError: No module named 'psutil'
  File "/usr/local/emsdk/emscripten/main/./tools/maint/rebaseline_tests.py", line 136, in <module>
    sys.exit(main())
  File "/usr/local/emsdk/emscripten/main/./tools/maint/rebaseline_tests.py", line 82, in main
    subprocess.check_call([utils.bat_suffix(os.path.join('test', 'runner')), '--rebaseline', 'codesize'], cwd=root_dir)
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/subprocess.py", line 373, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['test/runner', '--rebaseline', 'codesize']' returned non-zero exit status 1.
```

2) If I try setting `EM_CONFIG` like you suggested (`export EM_CONFIG=/usr/local/emsdk/.emscripten`) I get the exact same error

3) So in order to be able to run this tool, the only thing that seems to work is to issue `source /usr/local/emsdk/emsdk_env.sh`. And I am wondering if that is what is causing the problem? Maybe?

sbc100:
`source /usr/local/emsdk/emsdk_env.sh` should work just fine.

If you want to be able to just set `EM_CONFIG` then you need to make sure your python3 install has psutuil installed.   You can do that is serveral ways but one way is to do `pip4 install -r requirements-dev.txt`.    On my system its actually already installed as part of the `python3-psutil` package.

sbc100:
> * as a result the "latest" version ends up being under `/usr/local/emsdk/upstream/emscripten`

Are you installing "latest" or "tot"  ?  For rebaslining tests you you really need "tot".   You should also do `emsdk activate tot` at least once

ypujante:
I am not sure I activated `tot`... (the instructions did not specify to activate it)... that might be my issue

sbc100:
> I am not sure I activated `tot`... (the instructions did not specify to activate it)... that might be my issue

Yeah, the activate part is often not needed because if you had already activated a previous prebuilt sdk then `emskd install tot` will just replace the existing activated version (and there is no need to activate again in that case).  So that probably wasn't the reason.

ypujante:
@sbc100 I saw that 4.0.20 was released. When I check the documentation (although the title states "html5.h — Emscripten 4.0.20-git (dev) documentation"), the documentation that I added is not present.

sbc100:
> @sbc100 I saw that 4.0.20 was released. When I check the documentation (although the title states "html5.h — Emscripten 4.0.20-git (dev) documentation"), the documentation that I added is not present.

The automatic updating to website is something that has never existed before but we are in the process of adding.  Looks like there was a bug that was preventing it from working as expected: https://github.com/emscripten-core/emscripten/pull/25830

ypujante:
I guess I found a bug :). Thanks for taking care of it.

sbc100:
> I guess I found a bug :). Thanks for taking care of it.

Fixed now: https://emscripten.org/docs/api_reference/html5.h.html#unregister-function

Thanks for catching that!

juj:
The new added test fails on my CI boxes:

image

The failure occurs on Windows, Linux and macOS, but oddly only on "firefox latest beta" version of Firefox.

When I locally install that Firefox version (which is Firefox 144.0b6) via emsdk, and try out the failure, it never happens.

But for some reason on the CI, the failure is deterministic on all OSes.

The error reads `Assertion failed: count == eventHandlersCount`, so one of the registered event handler counts checks is failing.

I can't quite figure out right now what is different from local tests vs on the CI, except that the CI runs multiple tests in parallel (and with high contention), but locally, there is no contention, and it is just one test.

Curious if anyone would be able to reproduce this failure?

sbc100:
I think the problem is `f'-DSAFARI_SUPPORT={int(not self.is_wasm64())}'` .. it doesn't take into account that you have extra browser settings in your `EMCC_CFLAGS` environment variable.

ypujante:
I assume if that is a problem for everybody, we can certainly revert https://github.com/emscripten-core/emscripten/pull/25858 which was just adding 1 test case for the 2 event handlers (and I assume this is what is causing problem with @juj setup, not the rest of the test)

juj:
Ohh, that must be it.

Posted PR https://github.com/emscripten-core/emscripten/pull/25876 to resolve this. I think it's not that critical to verify the two different cases separately, so a more relaxed test assert should be enough.

PR #25536, opened 2025-10-09
[ci] Split wasm64 misc tests into their own runner. NFC

sbc100:
Prior to this we were tacking on the extra misc tests on the end of wasm64-4gb runner.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -820,7 +820,31 @@ jobs:
           command: echo "NODE_JS = NODE_JS_TEST" >> ~/emsdk/.emscripten
       - run-tests:
           title: "wasm64_4gb"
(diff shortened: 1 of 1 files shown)


kripken:
What is the advantage to this split? Is it faster overall?

sbc100:
> What is the advantage to this split? Is it faster overall?

Its likely fast yes, but mostly its just organizational.  I don't like lumping in these misc tests with the runner called `_4gb` (which mostly does run just the 4gb test suite).

sbc100:
> Ok, lgtm, but maybe measure to make sure this isn't wasteful (maybe there are so few misc tests that the general job overhead is high compared to them).

This job takes 5 mins to run, and only a few seconds to setup so I think its worth it.

Issue #25537, opened 2025-10-09
Native aliases like `wasmTable` are no longer exported correctly after #25376


sbc100:
For example, this no longer correctly exports the table: `./emcc -sEXPORTED_RUNTIME_METHODS=wasmTable hello.c`

PR #25539, opened 2025-10-09
Fix exporting of native symbol aliases (e.g. wasmTable)

sbc100:
This was broken by #25376.

Fixes: #25537

### DIFF
--- a/src/modules.mjs
+++ b/src/modules.mjs
@@ -433,6 +433,11 @@ function exportSymbol(name) {
 function exportRuntimeSymbols() {
   // optionally export something.
   function shouldExport(name) {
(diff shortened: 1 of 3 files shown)


src/modules.mjs
```diff
@@ -433,6 +433,9 @@ function exportSymbol(name) {
 function exportRuntimeSymbols() {
   // optionally export something.
   function shouldExport(name) {
+    if (nativeAliases[name]) {
```
kripken, line 438: Maybe add a comment here? I'm not sure why being a native alias would mean we should not export.

sbc100: Done.

PR #25559, opened 2025-10-15
[SPLIT_MODULE] Make getWasmImports return Proxy

aheejin:
When `SPLIT_MODULE` is set, this makes the `imports` object returned by `getWasmImports` itself a Proxy, which redirects property requests starting with `placeholder` to an inner handler that processes the second module loading, and process all other property requests to the original `imports` object itself.

This is a prepration to add multi-split loading functionality to the JS runtime. In the multi-split mode, we don't have a single seconday module `[modulename].deferred.wasm`, but instead multiple secondary modules. We plan to have multiple placeholder namespace (e.g., `placeholder2`, `placeholder3`, ...) that match those multiple files, and this can't be hardcoded with one `placeholder` string as we currently do, so having the `imports` object itself as a `Proxy` provides us flexibility to handle multiple placeholder namespaces and load correct modules later.

Thanks @tlively for the idea!

### DIFF
--- a/test/codesize/test_unoptimized_code_size.json
+++ b/test/codesize/test_unoptimized_code_size.json
@@ -1,16 +1,16 @@
 {
-  "hello_world.js": 56566,
-  "hello_world.js.gz": 17601,
+  "hello_world.js": 56592,
+  "hello_world.js.gz": 17608,
   "hello_world.wasm": 15119,
   "hello_world.wasm.gz": 7444,
-  "no_asserts.js": 26608,
-  "no_asserts.js.gz": 8874,
+  "no_asserts.js": 26634,
+  "no_asserts.js.gz": 8883,
   "no_asserts.wasm": 12219,
   "no_asserts.wasm.gz": 6005,
-  "strict.js": 54581,
-  "strict.js.gz": 16947,
+  "strict.js": 54607,
+  "strict.js.gz": 16955,
   "strict.wasm": 15119,
   "strict.wasm.gz": 7442,
-  "total": 180212,
-  "total_gz": 64313
+  "total": 180290,
+  "total_gz": 64337
 }


aheejin:
"Hide whitespaces" will make the diff easier to view.

src/preamble.js
```diff
@@ -487,27 +487,37 @@ async function getWasmBinary(binaryFile) {
 
 #if SPLIT_MODULE
 {{{ makeModuleReceiveWithVar('loadSplitModule', undefined, 'instantiateSync') }}}
-var splitModuleProxyHandler = {
+var wasmImportsProxyHandler = {
   get(target, prop, receiver) {
-    return (...args) => {
+    if (prop.startsWith('placeholder')) {
+      const moduleName = prop;
```
tlively: Maybe simpler to just give the outer property and inner property parameters different names.
```suggestion
get(target, moduleName, receiver) {
    if (moduleName.startsWith('placeholder')) {
```

tlively (approved):

src/preamble.js
aheejin: Manually applied. Somehow this couldn't be not automatically applied. Also changed `moduleName` to `env`, given that `placeholder[something]` is not a module name after all.

aheejin: Also renamed the inner `prop` to `base`, to make it clearer

aheejin:
Not sure why the codesize test is failing... I think my local machine's `rebaseline_tests.py` result is different from the CI's. Any ideas?

kripken:
Usually updating the emsdk (for LLVM and Binaryen) and merging in latest main here fixes those code size errors, for me.

src/preamble.js
```diff
@@ -487,27 +487,36 @@ async function getWasmBinary(binaryFile) {
 
 #if SPLIT_MODULE
 {{{ makeModuleReceiveWithVar('loadSplitModule', undefined, 'instantiateSync') }}}
-var splitModuleProxyHandler = {
-  get(target, prop, receiver) {
-    return (...args) => {
+var wasmImportsProxyHandler = {
```
sbc100: Can we keep the splitModule prefix (or at least include it somewhere in the name) so its clear when this needed/used?

src/preamble.js
sbc100: I don't understand why `env` is better then moduleName here?   Surely moduleName is the name of the module in the Wasm sense, no?    What does `env` mean here?

src/preamble.js
```diff
@@ -487,27 +487,36 @@ async function getWasmBinary(binaryFile) {
...
+            throw new Error('Placeholder function "' + base + '" should not be called when using JSPI.');
 #else
-      err(`placeholder function called: ${prop}`);
-      var imports = {'primary': wasmExports};
-      // Replace '.wasm' suffix with '.deferred.wasm'.
-      var deferred = wasmBinaryFile.slice(0, -5) + '.deferred.wasm'
-      loadSplitModule(deferred, imports, prop);
-      err('instantiated deferred module, continuing');
+            // TODO: Implement multi-split module loading
+            err(`placeholder function called: ${base}`);
```
sbc100: Pre-existing, but should this be `dbg` and wrapped in `ASSERTIONS` or somesuch?

src/preamble.js
aheejin: Renamed it to `splitImportsProxyHandler`: https://github.com/emscripten-core/emscripten/pull/25559/commits/e918916ecd5a4705c6c98bf21f694ad2b2983864

src/preamble.js
aheejin: So I thought... `placeholder` or `env` or `a` (in case of minification) were kind of "environments" as opposed to "modules" that mean wasm modules. But yeah, in many places we call this 'module' as well, so.. I changed it back to `moduleName`: https://github.com/emscripten-core/emscripten/pull/25559/commits/6cb0c42754724b6682640750a628aba986c8b329

src/preamble.js
aheejin: Used `dbg` in two places and wrapped them in `RUNTIME_DEBUG`, given that the places where `dbg()` is used in this file are all seem to be wrapped in it: https://github.com/emscripten-core/emscripten/pull/25559/commits/4f2f3b394450db70eec4f97945ec5edce5049bbc

aheejin:
> Usually updating the emsdk (for LLVM and Binaryen) and merging in latest main here fixes those code size errors, for me.

I updated both but still somehow does not match the CI's results....

aheejin:
Is there a way to log in to the CI via ssh and run things? I used to an option that's called like "run via SSH" or something but I don't see it anymore.

Also the message very confusing. The CI failure message is like
```console
The following (5) test expectation files were updated by

codesize/test_codesize_cxx_wasmfs.json: 176780 => 176780 [+0 bytes / +0.00%]
codesize/test_codesize_files_wasmfs.json: 55683 => 55683 [+0 bytes / +0.00%]
codesize/test_codesize_hello_O0.json: 39197 => 39189 [-8 bytes / -0.02%]
codesize/test_codesize_hello_dylink_all.json: 843553 => 843559 [+6 bytes / +0.00%]
codesize/test_unoptimized_code_size.json: 180236 => 180212 [-24 bytes / -0.01%]

Average change: -0.01% (-0.02% - +0.00%)
```
But I already updated those files in this PR. For example, you can see this PR updates `codesize/test_codesize_hello_O0.json`'s `total` line from 39197 to 39189. Why does this error out saying the size should be 39189?

sbc100:
The "Run via SSH" things only works for circleci.

Are you sure you did `emsdk install tot` and rebased your change onto main?

src/preamble.js
```diff
@@ -653,21 +673,22 @@ function getWasmImports() {
...
-    'placeholder': new Proxy({}, splitModuleProxyHandler),
-#endif
 #if RELOCATABLE
     'GOT.mem': new Proxy(wasmImports, GOTHandler),
     'GOT.func': new Proxy(wasmImports, GOTHandler),
 #endif
-  }
+  };
+#if SPLIT_MODULE
+  imports = new Proxy(imports, splitImportsProxyHandler);
```
sbc100: How about `splitModuleProxyHandler` for the name?

src/preamble.js
```diff
@@ -487,27 +487,46 @@ async function getWasmBinary(binaryFile) {
...
+      } else {
+        let moduleID = moduleName.split('.')[1];
+        secondaryFile = wasmBinaryFile.slice(0, -5) + '.' + moduleID + '.wasm';
+      }
+      var splitModuleProxyHandler = {
+        get(target, base, receiver) {
+          return (...args) => {
 #if ASYNCIFY == 2
-      throw new Error('Placeholder function "' + prop + '" should not be called when using JSPI.');
+            throw new Error('Placeholder function "' + base + '" should not be called when using JSPI.');
```
sbc100, line 497: (As an aside I wonder if this line should be an assert).

sbc100 (approved):

aheejin:
Sorry, reverted an accidental change that was not meant to be in this PR: https://github.com/emscripten-core/emscripten/pull/25559/commits/68c330878fcc4e2049c7e5a0788b4fd2149838ef

src/preamble.js
aheejin: That's the name of the inner handler... Do you want it to change to something else?

src/preamble.js
aheejin: Not sure... Can't this be triggered by user errors? If so an `Error` might be fine?

src/preamble.js
```diff
@@ -487,27 +487,39 @@ async function getWasmBinary(binaryFile) {
 
 #if SPLIT_MODULE
 {{{ makeModuleReceiveWithVar('loadSplitModule', undefined, 'instantiateSync') }}}
-var splitModuleProxyHandler = {
-  get(target, prop, receiver) {
-    return (...args) => {
+var splitImportsProxyHandler = {
+  get(target, moduleName, receiver) {
+    if (moduleName.startsWith('placeholder')) {
+      var splitModuleProxyHandler = {
```
sbc100: Just call this one `innerHandler` then maybe?

sbc100: This inner name is less important since its not in the global namespace.  It could even just be inlined (anonymous) bat that could make harder to grok maybe?

aheejin: Renamed the outer handler to `splitModuleProxyHandler` and the inner one to `innerHandler`: https://github.com/emscripten-core/emscripten/pull/25559/commits/eaea168cee1d1c7a423b2abed387f07d7e6aab62

The inner handler name is used here, so now sure how to anonymize it.
```js
      return new Proxy({}, innerHandler); 
```

sbc100: I mean you could just write `return new Proxy({},  { ... inner handler code }`

aheejin:
I removed `(NFC)` from the title because it changes some behaviors now (like, wrapping debug messages in `RUNTIME_DEBUG`)

src/preamble.js
aheejin: Done: https://github.com/emscripten-core/emscripten/pull/25559/commits/67e5380b3b9fc1f3e77f2b619c7cc6b12cc3a9d0

sbc100:
I just landed https://github.com/emscripten-core/emscripten/pull/25566 which should help with the codesize issues.

aheejin:
Thanks, but it still fails with the codesize checks... I think the wasm sizes are different and have no clue why. I also tried `emsdk tot` but it failed too. Maybe I should ask someone tomorrow to run the codesize script with my PR and compare that with my results.

sbc100:
> Thanks, but it still fails with the codesize checks... I think the wasm sizes are different and have no clue why. I also tried `emsdk tot` but it failed too. Maybe I should ask someone tomorrow to run the codesize script with my PR and compare that with my results.

The first thing to do is make sure you can run `tools/maint/rebaseline_tests.py` locally on upstream/main and get no chnages.

If you have changes in this case then either:

1. main it self is out-of-date
2. You need to install the latest emsdk
3. You need to run `--clear-cache`

Maybe you are getting bitten by (3)

aheejin:
I thought I did all three, but now I do it again, somehow I again see changes... Let's see if this time it works 🙏

aheejin:
Codesize checks succeeded! Thanks!

PR #25565, opened 2025-10-16
Add EMTEST_RETRY_COUNT to force retrying failed tests

juj:
The way this differs from the existing EMTEST_RETRY_FLAKY is that this retries any failed test, whereas EMTEST_RETRY_FLAKY only retries those tests that had explicitly been deemed to be flaky beforehand.

The rationale for adding this feature is twofold:

1. There is currently so much flakiness in tests in the current test suites, that I would be flagging flaky tests for many months to come in my own CI. It is unclear if some of that flakiness is a harness problem or a systemic problem rather than an individual test problem, so I could end up flagging a majority of all tests in the suites as flaky.

2. Whenever a test fails in my CI, the very first thing I need to check is whether the failure was just a one-off, or whether the failure was a deterministic failure. So being able to run with `EMTEST_RETRY_COUNT=5` will automate such testing for me and immediately give me feedback whether any test failure was deterministic or intermittent.

### DIFF
--- a/test/retryable_unittest.py
+++ b/test/retryable_unittest.py
@@ -6,10 +6,10 @@ EMTEST_RETRY_COUNT = int(os.getenv('EMTEST_RETRY_COUNT', '0'))
 
 
 class RetryableTestCase(unittest.TestCase):
-  ''' This class patches in to the Python unittest TestCase object to incorporate
+  '''This class patches in to the Python unittest TestCase object to incorporate
   support for an environment variable EMTEST_RETRY_COUNT=x, which enables a
   failed test to be automatically re-run to test if the failure might have been
-  due to an instability. '''
+  due to an instability.'''
 
   def run(self, result=None):
     retries_left = EMTEST_RETRY_COUNT


test/retryable_unit_test.py
```diff
@@ -0,0 +1,34 @@
...
+import unittest
+
+# This class patches in to the Python unittest TestCase object to incorporate
+# support for an environment variable EMTEST_RETRY_COUNT=x, which enables a
+# failed test to be automatically re-run to test if the failure might have been
+# due to an instability.
+class RetryableTestCase(unittest.TestCase):
+  def run(self, result=None):
+    self.origTestMethodName = self._testMethodName
+    test_retry_count = int(os.getenv('EMTEST_RETRY_COUNT', '0'))
```
sbc100: This line can maybe be at the top level of the file and the name match the environment variable?  (Like the other env vars we check)

test/retryable_unit_test.py
```diff
@@ -0,0 +1,34 @@
+import common
+import os
+import unittest
+
+# This class patches in to the Python unittest TestCase object to incorporate
+# support for an environment variable EMTEST_RETRY_COUNT=x, which enables a
+# failed test to be automatically re-run to test if the failure might have been
+# due to an instability.
```
sbc100: Maybe make this a docstring `"""` n the line after the class?

test/retryable_unit_test.py
```diff
@@ -0,0 +1,34 @@
...
+class RetryableTestCase(unittest.TestCase):
+  def run(self, result=None):
+    self.origTestMethodName = self._testMethodName
```
sbc100: Is this used anywhere?

test/retryable_unittest.py
```diff
@@ -0,0 +1,34 @@
...
+# failed test to be automatically re-run to test if the failure might have been
+# due to an instability.
+class RetryableTestCase(unittest.TestCase):
+  def run(self, result=None):
+    self.origTestMethodName = self._testMethodName
+    test_retry_count = int(os.getenv('EMTEST_RETRY_COUNT', '0'))
+    retries_left = test_retry_count
+
+    num_fails = len(result.failures)
+    num_errors = len(result.errors)
```
sbc100, line 18: Are these part of the non-parallel test runner too?

test/common.py
```diff
@@ -10,6 +10,7 @@
 from typing import Dict, Tuple
 from urllib.parse import unquote, unquote_plus, urlparse, parse_qs
 from http.server import ThreadingHTTPServer, SimpleHTTPRequestHandler
+from retryable_unit_test import RetryableTestCase
```
sbc100: Maybe just `unitest` (without the undercore) to match the python naming?

test/common.py
```diff
@@ -1032,7 +1033,7 @@ def __new__(mcs, name, bases, attrs):
     return type.__new__(mcs, name, bases, new_attrs)
 
 
-class RunnerCore(unittest.TestCase, metaclass=RunnerMeta):
+class RunnerCore(RetryableTestCase, metaclass=RunnerMeta):
```
sbc100, line 1036: If we are going to inject this into all our test cases, maybe we can unify with the way the current flaky decorator does retying.

I'm fine with doing that as followup also.

sbc100 (approved):

test/retryable_unit_test.py
juj: Thanks, that was leftover from testing. Removed.

test/retryable_unittest.py
juj: Yes, these are part of the default unittest.py `unittest.TestCase` implementation. Had to add them to the parallel test runner to match the shape of the upstream Python implementation. Verified that this works in single-threaded and parallel test runner.

test/common.py
juj: Yeah, folding the current flaky decorator into this class sounds natural. I looked into it a bit, but couldn't immediately see how to read the decorator in this class. That's a good followup.

test/retryable_unittest.py
```diff
@@ -0,0 +1,36 @@
+import common
+import os
+import unittest
+
+EMTEST_RETRY_COUNT = int(os.getenv('EMTEST_RETRY_COUNT', '0'))
+
+
+class RetryableTestCase(unittest.TestCase):
+  ''' This class patches in to the Python unittest TestCase object to incorporate
```
sbc100: docstrings don't normally start and when with a space like this AFAIK.

PR #25569, opened 2025-10-16
Add --max-failures <n> option in test harness

juj:
Add --max-failures <n> option in test harness to generalize on the --failfast option (which is effectively a '--max-failures 0' mode).

This allows running the harness while allowing a couple of failures. I find that when using `--failfast`, aborting on the first error is sometimes a bit too eager, i.e. there might be a flaky test that triggers the abort, or there might be an error that causes more than one test to fail.

Examining test harness results, that interpreting a small handful of failures (1,2,3...5) is still doable one-by-one, but when the number of failures grows to, say, > 5, the typically I want to deal with the first couple of errors only, and re-run the harness after that, assuming the rest of the errors would be from the same root cause.

So this enables running with, say, `--max-failures 5`, and get info on the first handful of errors, and ignore if any more errors occur.

This helps retain the failfast-like behavior while producing a bit more info, and being a bit more tolerant to instabilities.

### DIFF
--- a/test/runner.py
+++ b/test/runner.py
@@ -481,7 +481,7 @@ def parse_args():
                       help='Use the default CI browser configuration.')
   parser.add_argument('tests', nargs='*')
   parser.add_argument('--failfast', action='store_true', help='If true, test run will abort on first failed test.')
-  parser.add_argument('--max-failures', type=int, default=2**31 - 1, help='Specifies the maximum number of allowed failures before aborting the test run.')
+  parser.add_argument('--max-failures', type=int, default=2**31 - 1, help='If specified, test run will abort after N failed tests.')
   parser.add_argument('--failing-and-slow-first', action='store_true', help='Run failing tests first, then sorted by slowest first. Combine with --failfast for fast fail-early CI runs.')
   parser.add_argument('--start-at', metavar='NAME', help='Skip all tests up until <NAME>')
   parser.add_argument('--continue', dest='_continue', action='store_true',


test/runner.py
```diff
@@ -480,7 +480,8 @@ def parse_args():
   parser.add_argument('--browser-auto-config', type=bool, default=True,
                       help='Use the default CI browser configuration.')
   parser.add_argument('tests', nargs='*')
-  parser.add_argument('--failfast', action='store_true')
+  parser.add_argument('--failfast', action='store_true', help='If true, test run will abort on first failed test.')
+  parser.add_argument('--max-failures', type=int, default=2**31 - 1, help='Specifies the maximum number of allowed failures before aborting the test run.')
```
sbc100: How about using the same phrasing as in the line above: "If true, test run will abort after N failed tests"?

test/runner.py
```diff
@@ -492,7 +493,15 @@ def parse_args():
...
                       help='Repeat each test N times (default: 1).')
   parser.add_argument('--bell', action='store_true', help='Play a sound after the test suite finishes.')
-  return parser.parse_args()
+
+  options = parser.parse_args()
+
+  if options.failfast:
+    if options.max_failures != 0:
+      utils.exit_with_error('--failfast and --max-failures are mutually exclusive!')
+    options.max_failures = 0
```
sbc100, line 502: Shouldn't this be `= 1`?

test/runner.py
```diff
@@ -492,7 +493,15 @@ def parse_args():
   parser.add_argument('--repeat', type=int, default=1,
                       help='Repeat each test N times (default: 1).')
   parser.add_argument('--bell', action='store_true', help='Play a sound after the test suite finishes.')
-  return parser.parse_args()
+
+  options = parser.parse_args()
+
+  if options.failfast:
+    if options.max_failures != 0:
+      utils.exit_with_error('--failfast and --max-failures are mutually exclusive!')
```
sbc100, line 501: Looks like argparse supports `parser.add_mutually_exclusive_group()` .. but I think it might be more complicated that this simple check.

test/runner.py
juj: ok

test/runner.py
juj: I took a convention of max_failures to mean "maximum allowed failures before aborting the test run".

So max_failures = 0 means that zero failures are allowed, after that test run is aborted. (first failure will abort)
max_failures = 1 means that one failure is allowed, and the second one after that aborts the run.

test/runner.py
```diff
@@ -480,7 +480,8 @@ def parse_args():
   parser.add_argument('--browser-auto-config', type=bool, default=True,
                       help='Use the default CI browser configuration.')
   parser.add_argument('tests', nargs='*')
-  parser.add_argument('--failfast', action='store_true')
+  parser.add_argument('--failfast', action='store_true', help='If true, test run will abort on first failed test.')
```
sbc100, line 483: Actually I wonder if you can you do something `action=store_constant` + `target=max_failures` here?  Maybe too clever.

lgtm either way.

juj: Hmm there are still a couple of uses of the `options.failfast` variable that this PR does not change, so keeping that around for now helps.

juj:
Anything more here?

sbc100 (approved):

PR #25571, opened 2025-10-16
[SPLIT_MODULE] Support multi-split loading

aheejin:
Currently, if the primary module name is `test.wasm` and when a placeholder function is called, the emscripten JS runtime loads `test.deferred.wasm`, which is hardcoded. But as we want emscripten to support multi-split, the runtime has to know which secondary module to load.

Binaryen's https://github.com/WebAssembly/binaryen/pull/7975 changes the placeholder import module naming scheme. Currently placeholder's import modules are all `placeholder`: https://github.com/WebAssembly/binaryen/blob/dcc704ce20e4723ae42012039bdb3b84ec2035f9/test/lit/wasm-split/multi-split.wast#L291-L295

But the Binaryen's companion PR will change this to the format of `placeholder.[secondaryName]`. Then Emscripten runtime will parse this module name to learn which secondary file to load. For example, if the import module is `placeholder.2` and the import base is `3` and the primary wasm file is `test.wasm`,
```wast
(import "placeholder.2" "3" (func $placeholder_3 (result i32)))
```
when `placeholder_3` function is loaded, the runtime will parse `placeholder.2` and correctly figure out that it should load `test.2.wasm`.

This PR still supports the old `placeholder` namespace, because this PR has to land first for the Binaryen companion PR to land. After it lands, we can remove the `placeholder` handling part.

Companion PR: https://github.com/WebAssembly/binaryen/pull/7975

### DIFF
--- a/src/preamble.js
+++ b/src/preamble.js
@@ -490,20 +490,25 @@ async function getWasmBinary(binaryFile) {
 var splitModuleProxyHandler = {
   get(target, moduleName, receiver) {
     if (moduleName.startsWith('placeholder')) {
+      let secondaryFile;
+      if (moduleName == 'placeholder') { // old format
+        secondaryFile = wasmBinaryFile.slice(0, -5) + '.deferred.wasm';
+      } else { // new format
+        let moduleID = moduleName.split('.')[1];
+        secondaryFile = wasmBinaryFile.slice(0, -5) + '.' + moduleID + '.wasm';
+      }
       return new Proxy({}, {
         get(target, base, receiver) {
           return (...args) => {
 #if ASYNCIFY == 2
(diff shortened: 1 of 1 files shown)


src/preamble.js
```diff
@@ -490,20 +490,25 @@ async function getWasmBinary(binaryFile) {
 var splitModuleProxyHandler = {
   get(target, moduleName, receiver) {
     if (moduleName.startsWith('placeholder')) {
+      let secondaryFile;
+      if (moduleName == 'placeholder') { // old format
+        secondaryFile = wasmBinaryFile.slice(0, -5) + '.deferred.wasm';
```
aheejin, line 495: In case I wasn't clear, the plan is to remove this special old format handling after https://github.com/WebAssembly/binaryen/pull/7975 lands.

PR #25574, opened 2025-10-17
Add a clarifying note on value_object docs

aestuans:
The sentence "The JavaScript code does not need to worry about lifetime management" could be misleading on its own.


aestuans:
@brendandahl just a ping, in case this got lost in your notifications.

PR #25575, opened 2025-10-17
Fix tools/install.py script to use Emsdk Node.js and NPM

juj:
Fix tools/install.py script to use Emsdk Node.js and NPM. Fixes other.test_install on macOS if system macOS Node.js version is out of date.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -15687,7 +15687,7 @@ addToLibrary({
   @crossplatform
   @requires_git_checkout
   def test_install(self):
-    self.run_process([PYTHON, path_from_root('tools/install.py'), 'newdir'])
+    self.run_process([PYTHON, path_from_root('tools/install.py'), 'newdir'], env=shared.env_with_node_in_path())
     self.assertExists('newdir/emcc')
     # Some files, such as as maintenance tools should not be part of the
     # install.


tools/install.py
```diff
@@ -85,8 +85,23 @@ def copy_emscripten(target):
...
+  if not node:
+    return shutil.which('npm')
+
+  directory = os.path.dirname(os.path.abspath(node))
+  npm = 'npm.cmd' if os.name == 'nt' else 'npm'
+  return os.path.join(directory, npm)
+
+
 def npm_install(target):
-  subprocess.check_call([shutil.which('npm'), 'ci', '--omit=dev'], cwd=target)
```
sbc100, line 89: Could we instead try using the existing `env_with_node_in_path` helper from shared.py?

tools/install.py
```diff
@@ -85,8 +85,23 @@ def copy_emscripten(target):
       shutil.copy2(full, os.path.join(target, root, f), follow_symlinks=False)
 
 
+def get_npm():
+  node = os.getenv('EMSDK_NODE')
```
sbc100: Can we use `shared.NODE_JS` instead of `EMSDK_NODE` here?   I'd rather not use these emsdk-specific things here if we can avoid it.

`shared.NODE_JS` is guaranteed to exist.

Maybe install.py isn't setup to import shared.py yet?

sbc100: Hmm.. I guess `bootstrap.py` has the same issue?  Is this script also not working your CI machine?

sbc100:
> macOS if system macOS Node.js version is out of date.

What does the failure look like in this case?

juj:
The failure looks like 

image

tools/install.py
juj: check

tools/install.py
juj: Updated to use `config.NODE_JS`

juj: bootstrap.py does work. It is only the test `other.test_install` that is failing.

juj:
Hmm, the changes are resulting in these errors: https://github.com/emscripten-core/emscripten/actions/runs/18601976754/job/53042367932

I am a bit stumped as to what could cause this.. it is as if .emscripten no longer exists, but the changes here shouldn't affect that?

juj:
The version that relied on `EMSDK_NODE` env. var. above did pass CI tests: https://github.com/emscripten-core/emscripten/actions/runs/18589062792/job/52999639477

sbc100:
I think the `archive` step is assuming the `make install` (and by extension) `install.py` don't require a fulling configured version of emscripten to work.

i.e we are assuming the install.py is completely standalone... but it seems like maybe it cannot actually be standalone.

tools/install.py
sbc100: That seems odd because `bootstrap.py` also runs `npm` from the PATH.

sbc100:
Could you maybe update the `test_install` test to use `env_with_node_in_path` when it runs `install.py`?

juj:
Oh right, it is the mere act of attempting to import shared that requires .emscripten config to be present.

Updated to change the test instead.

sbc100 (approved):

PR #25576, opened 2025-10-17
Only export selected data addresses on the Module object

sbc100:
Just like for other types of symbols we don't want to export symbols on the `Module` object itself if they user requested this.

Also, fix the memory profiler to not depend the external export of `__heap_base` but instead the internal version (like it does for `_sbrk` and other dependencies.

### DIFF
--- a/src/memoryprofiler.js
+++ b/src/memoryprofiler.js
@@ -502,7 +502,7 @@ var emscriptenMemoryProfiler = {
     html += '. STACK_MAX: ' + toHex(stackMax, width) + '.';
     html += '<br />STACK memory area used now (should be zero): ' + self.formatBytes(stackBase - stackCurrent) + '.' + colorBar('#FFFF00') + ' STACK watermark highest seen usage (approximate lower-bound!): ' + self.formatBytes(stackBase - self.stackTopWatermark);
 
-    var heap_base = Module['___heap_base'];
+    var heap_base = ___heap_base;
     var heap_end = _sbrk({{{ to64('0') }}});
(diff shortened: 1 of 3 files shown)


dschuff (commented):
Maybe this should have a changelog entry too, since it might be user visible?

sbc100:
> Maybe this should have a changelog entry too, since it might be user visible?

Im not actually sure this will be user visible... or at least I'm having trouble articulating exactly how it would be.

The thing is that the users visible controls are not effected e.g EMSCRIPTEN_KEEPALIVE and EXPORTED_FUNCTIONS still continue to work as expected and export the symbols. 

Its native symbols that were added to `settings.REQUIRED_EXPORTS` internally that are no longer exported, and this is not user-visible setting.

sbc100:
I could say something like:

"Data symbols that were exported from the wasm module for internal use are no longer exports on the Module object."? 

I don't know how common this is outside of this specific `--memory-profiler` usage though.

juj:
> Im not actually sure this will be user visible...

The output on the `Module` object will be user visible. Though whether anyone depended on data in there, it is harder to say.

LGTM, a note in ChangeLog might be good here.

sbc100:
> > Im not actually sure this will be user visible...
> 
> The output on the `Module` object will be user visible. Though whether anyone depended on data in there, it is harder to say.
> 
> LGTM, a note in ChangeLog might be good here.

Note the requests to export data symbols like via EMSCRIPTEN_KEEPALIVE and/or EXPORTED_FUNCTIONS are not effect by this change.    Those once we still continue to export.

PR #25577, opened 2025-10-17
[empath-split] Support multi-paths modules

aheejin:
This changes the format of the user input "paths" file so that multiple paths can be split into a single module. The new paths file structure is now very similar to wasm-split's manifest file, with functions replaced with paths. For example, the format will be like
```
module1
path/to/a
path/to/b

module2
path/to/c
```
Where `module1` and `module2` are module names.


tools/empath-split.py
```diff
@@ -252,31 +320,40 @@ def main():
...
   # Compute {path: list of functions} map
-  path_to_funcs = get_path_to_functions_map(args.wasm, sourcemap, paths)
+  all_paths = []
+  for paths in module_to_paths.values():
+    all_paths.extend(paths)
+  path_to_funcs = get_path_to_functions_map(args.wasm, sourcemap, all_paths)
 
   # Write .manifest file
-  with tempfile.NamedTemporaryFile(suffix=".manifest", mode='w+', delete=args.preserve_manifest) as f:
+  with tempfile.NamedTemporaryFile(suffix=".manifest", mode='w+', delete=not args.preserve_manifest) as f:
```
aheejin: Drive-by fix. `--preserve-manifest` was doing the opposite of what it should do...

test/test_other.py
```diff
@@ -15760,9 +15760,14 @@ def test_empath_split(self):
       void foo() { std::cout << "foo" << std::endl; }
     ''')
     create_file('path_list', r'''
```
sbc100: Should we call this `path_list.txt` ?

tools/empath-split.py
```diff
@@ -252,31 +309,40 @@ def main():
...
+            print('    ' + func)
         print()
-      for func in path_to_funcs[path]:
+
+      f.write(f'{module}\n')
+      for func in funcs:
         f.write(func + '\n')
-      if i < len(paths) - 1:
+      if i < len(module_to_paths) - 1:
         f.write('\n')
```
sbc100: Maybe instead do this at the top with:

```
if i != 0:  // Unless we are the first entry add a newline separator
   f.write('\n')
```

tools/empath-split.py
```diff
@@ -252,31 +309,40 @@ def main():
...
+        for path in paths:
+          if path in path_to_funcs:
+            print(f'  {path}: {len(path_to_funcs[path])} functions')
+          for func in path_to_funcs[path]:
+            print('    ' + func)
         print()
-      for func in path_to_funcs[path]:
+
+      f.write(f'{module}\n')
+      for func in funcs:
```
sbc100, line 340: Annoyingly If you want this to be deterministic then I think you need to make this a dict rather than set.

The `dict` in python became deterministic in 3.7 but the `set` remains unordered.    Annoying..

test/test_other.py
aheejin: Done: https://github.com/emscripten-core/emscripten/pull/25577/commits/ac1d2454ad55e38375483705c7147fde5654bd2a

tools/empath-split.py
aheejin: Done: https://github.com/emscripten-core/emscripten/pull/25577/commits/97258c682d7491ec7bd8d2d2f02e318b9477a27f

tools/empath-split.py
aheejin: Changed it and also one more place from `set` to list: https://github.com/emscripten-core/emscripten/pull/25577/commits/7eedf6d01c9f36e7eda03da917799fb3d19c0851
Apparently they didn't even need to be sets after all...

tools/empath-split.py
aheejin: Apparently this drive-by fix is seems to be a reason that Windows CI is failing: https://app.circleci.com/pipelines/github/emscripten-core/emscripten/46843/workflows/78cfd5c9-b267-45b7-ac41-6669c0b2adb6/jobs/1058317

It looks in Windows it is not allowed to re-open an already opened file unless some conditions are met: https://docs.python.org/3/library/tempfile.html
> Opening the temporary file again by its name while it is still open works as follows:
> - On POSIX the file can always be opened again.
> - On Windows, make sure that at least one of the following conditions are fulfilled:
>   - `delete` is false
>   - additional open shares delete access (e.g. by calling [os.open()](https://docs.python.org/3/library/os.html#os.open) with the flag `O_TEMPORARY`)
>   - `delete` is true but `delete_on_close` is false. Note, that in this case the additional opens that do not share delete access (e.g. created via builtin [open()](https://docs.python.org/3/library/functions.html#open)) must be closed before exiting the context manager, else the [os.unlink()](https://docs.python.org/3/library/os.html#os.unlink) call on context manager exit will fail with a [PermissionError](https://docs.python.org/3/library/exceptions.html#PermissionError).

The Windows CI didn't error out before this PR because the first condition (`delete` is false) was accidentally satisfied due to this bug. To fix this I think we should use `try`-`finally`. Will do this in another PR, and will revert this drive-by fix here.

aheejin:
I don't understand why `test_sanity` fails.. https://app.circleci.com/pipelines/github/emscripten-core/emscripten/46879/workflows/0ae5c3bb-a690-4d5f-aeaf-a43705d8484e/jobs/1059459

@sbc100 Do you think other things look fine?

test/test_other.py
```diff
@@ -15169,16 +15169,21 @@ def test_empath_split(self):
       #include <iostream>
       void foo() { std::cout << "foo" << std::endl; }
     ''')
-    create_file('path_list', r'''
+    create_file('path_list.txt', r'''
+      myapp
```
sbc100, line 15173: nit: Can you end these lines with `:` maybe?   I feel like it would make them easier to read if I could clearly see the different between module names and filenames.

aheejin: I'd like to make the file format similar to that of wasm-split's multi split manifest, which does not have colons: https://github.com/WebAssembly/binaryen/blob/main/test/lit/wasm-split/multi-split.wast.manifest

sbc100: Fair enough, maybe we could update them both one day?

aheejin: I don't mind much, but wasm-split's manifest file already has some users. If we decide to do it, I think we can give some internal clients a heads-up though. WDYT? @tlively

aheejin: Will land this PR for now then.

tlively: > I don't mind much, but wasm-split's manifest file already has some users. If we decide to do it, I think we can give some internal clients a heads-up though. WDYT? @tlively

Making improvements to the manifest file format sounds good to me. I imagine we might want to make more changes in the future as well, for example to include information about dependencies between modules.

PR #25579, opened 2025-10-17
[ci] Enable console logging in chrome testing. NFC

sbc100:
The format looks like this:

```
[4082634:4082634:1017/153703.620972:INFO:CONSOLE:146] "another line.", source: http://localhost:8888/test.html (146)
```


sbc100:
This could make the CI logs very big.. I guess we will see if that is a problem.

juj:
I don't yet have Chrome testing on my CI bots to have an opinion here either way. LGTM.

PR #25589, opened 2025-10-19
Add EMTEST_RESTART_BROWSER_EVERY_N_TESTS feature

juj:
Add EMTEST_RESTART_BROWSER_EVERY_N_TESTS feature that allows restarting the browser after every N tests. Allows working around https://bugzilla.mozilla.org/show_bug.cgi?id=1992558

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -94,6 +94,7 @@ if 'EM_BUILD_VERBOSE' in os.environ:
 # 0: Disabled (reuse the browser instance to run all tests. Default)
 # 1: Restart a fresh browser instance for every browser test.
 # 2,3,...: Restart a fresh browser instance after given number of tests have been run in it.
+# Helps with e.g. https://bugzil.la/1992558
 EMTEST_RESTART_BROWSER_EVERY_N_TESTS = int(os.getenv('EMTEST_RESTART_BROWSER_EVERY_N_TESTS', '0'))
 
 # If we are drawing a parallel swimlane graph of test output, we need to use a temp


test/common.py
```diff
@@ -90,6 +90,12 @@
 if 'EM_BUILD_VERBOSE' in os.environ:
   exit_with_error('EM_BUILD_VERBOSE has been renamed to EMTEST_BUILD_VERBOSE')
 
+# Triggers the browser to restart after every given number of tests.
+# 0: Disabled (reuse the browser instance to run all tests. Default)
+# 1: Restart a fresh browser instance for every browser test.
+# 2,3,...: Restart a fresh browser instance after given number of tests have been run in it.
```
sbc100, line 96: Can you link to the firefox bug here?

test/common.py
```diff
@@ -2585,6 +2593,7 @@ def browser_restart(cls):
     logger.info('Restarting browser process')
     cls.browser_terminate()
     cls.browser_open(cls.HARNESS_URL)
+    BrowserCore.num_tests_ran = 0
```
sbc100, line 2596: This counter is process-local so I think in the parallel runner this will effect each of the N browsers that get started right?   Seems like the right thing to do though

test/common.py
```diff
@@ -2652,7 +2661,7 @@ def launch_browser_harness_with_proc_snapshot_workaround(cls, parallel_harness,
       # Give the browser time to spawn its subprocesses. Use an increasing
       # timeout as a crude way to account for system load.
       if parallel_harness or is_safari():
-        time.sleep(2 + count * 0.3)
+        time.sleep(min(2 + count * 0.3, 10))
```
sbc100, line 2664: Unrelated, but this seems like a quite a lot of sleeping ... do we need to do this even on linux?

Oh wait this function is never run on linux...   maybe we could add `assert not LINUX` or update the doc string saying if/when this hack is needed?

test/common.py
sbc100: Why not just use `cls` here like the lines above?

sbc100 (approved):

test/common.py
juj: `cls` here points to the instance, not the class.

test/common.py
juj: It is.. and maybe there is a better way, but before changing this too much, I just want to settle to getting passing runs first.

test/common.py
sbc100: This is a `classmethod` and `cls` points to the class

sbc100: I suppose it points to the subclass .... it just seems rather odd to use `cls` and `BrowserCore` in the same method here.   

If we made `num_tests_ran` into a boxed value we could update it using `cls` or `self` I guess, but that sounds annoying too.   

I guess this solution seems reasonable.

Issue #25570, opened 2025-10-16
em++.ps1 passes arguments differently than em++.bat


jrxpana:
**Version of emscripten/emsdk:**
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.15 (09f52557f0d48b65b8c724853ed8f4e8bf80e669)
clang version 22.0.0git (https:/github.com/llvm/llvm-project 3388d40684742e950b3c5d1d2dafe5a40695cfc1)
Target: wasm32-unknown-emscripten
Thread model: posix

**Problem Description**
In previous versions of emscripten, calling em++ with a specific set of command line arguments succeeded. After we updated to 4.0.15 it suddenly failed. I traced it back to the fact that the previous versions we used had 3 versions of the file/command:

- em++
- em++.bat
- em++.py

and 4.0.15 now also has:

- em++.ps1

When our build script calls em++ it now ends up calling em++.ps1 instead of em++.bat, which somehow passes the arguments to python differently.

Minimal repro:

**hi.cpp**
```cpp
#include <iostream>

int main()
{
	std::cout << "hi" << std::endl;
	return 0;
}
```

**build.ps1**
```powershell
$emsdk_path = "$env:LOCALAPPDATA\emsdk"

& "$emsdk_path\emsdk.ps1" construct_env

Write-Host "build using em++.bat" -fore green
& em++.bat -pthread -s ALLOW_MEMORY_GROWTH=1 hi.cpp 2>&1 | % ToString | Tee-Object -Variable build_result

Write-Host "build using em++.ps1" -fore green
& em++.ps1 -pthread -s ALLOW_MEMORY_GROWTH=1 hi.cpp 2>&1 | % ToString | Tee-Object -Variable build_result
```

With both of these files in the same folder, when I open a Powershell window and run .\build.ps1, I get the following output:

> Setting up EMSDK environment (suppress these messages with EMSDK_QUIET=1)
> Setting environment variables:
> 
> build using em++.bat
> em++: warning: -pthread + ALLOW_MEMORY_GROWTH may run non-wasm code slowly, see https://github.com/WebAssembly/design/issues/1271 [-Wpthreads-mem-growth]
> 
> build using em++.ps1
> python.exe : em++: warning: -pthread + ALLOW_MEMORY_GROWTH may run non-wasm code slowly, see
https://github.com/WebAssembly/design/issues/1271 [-Wpthreads-mem-growth]
[...]

PR #25591, opened 2025-10-19
Fix PowerShell argument handling for em++.ps1

devalgupta404:
## Root Cause

1. **PowerShell stderr handling**: PowerShell treats stderr output as `RemoteException` objects instead of normal text
2. **Argument passing**: `$MyInvocation.UnboundArguments` doesn't work reliably with piped operations  
3. **Python detection**: No fallback when `py.exe` is not available

## Solution

- **Use `ProcessStartInfo`**: Manually capture stdout/stderr to avoid PowerShell's exception handling
- **Improve argument handling**: Use `$args` with fallback to `$MyInvocation.UnboundArguments`
- **Better Python detection**: Add fallback chain: `py.exe` → `python` → `'python'`
- **Preserve behavior**: Maintain identical exit codes and output as batch files

## Changes

- Modified `tools/maint/run_python.ps1` template
- Added robust argument collection and processing
- Implemented proper stdout/stderr handling via `ProcessStartInfo`
- Enhanced Python launcher detection with error handling

**Before**: `em++.ps1` fails with `RemoteException` when piped  
**After**: `em++.ps1` works identically to `em++.bat` in all scenarios

Fixes: #25570

### DIFF
--- a/tools/maint/run_python.ps1
+++ b/tools/maint/run_python.ps1
@@ -15,10 +15,16 @@ $ErrorActionPreference = 'Stop'
 $launcher = $env:EMSDK_PYTHON
 $launcherArgs = @()
 if (!$launcher) {
-  # Use the global Python launcher that manages all Python installations.
-  $launcher = Get-Command py.exe
-  # ...but ask for Python 3
-  $launcherArgs += '-3'
+  try {
+    $launcher = Get-Command py.exe -ErrorAction Stop
+    $launcherArgs += '-3'
+  } catch {
+    try {
+      $launcher = Get-Command python -ErrorAction Stop
+    } catch {
+      $launcher = 'python'
+    }
+  }
 }
 
 # Ignore PYTHON* environment variables.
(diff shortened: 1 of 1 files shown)


tools/maint/run_python.ps1
```diff
@@ -29,4 +35,28 @@ Remove-Item -ErrorAction Ignore Env:\_PYTHON_SYSCONFIGDATA_NAME
...
+$processInfo.UseShellExecute = $false
+$processInfo.CreateNoWindow = $true
+$process = New-Object System.Diagnostics.Process
+$process.StartInfo = $processInfo
+$process.Start() | Out-Null
+$stdout = $process.StandardOutput.ReadToEnd()
+$stderr = $process.StandardError.ReadToEnd()
+$process.WaitForExit()
+if ($stdout) { Write-Output $stdout }
+if ($stderr) { Write-Output $stderr }
```
sbc100, line 59: This seems broken because it means that all of the stdout is printed before any of the stderr.

In the default case we want these to be interleved.

Also, shouldn't stderr be printed to a different place that stderr?

tools/maint/run_python.ps1
```diff
@@ -15,10 +15,16 @@ $ErrorActionPreference = 'Stop'
...
+  try {
+    $launcher = Get-Command py.exe -ErrorAction Stop
+    $launcherArgs += '-3'
+  } catch {
+    try {
+      $launcher = Get-Command python -ErrorAction Stop
+    } catch {
+      $launcher = 'python'
+    }
+  }
```
sbc100, line 27: Is this a fix for a separate issue?

sbc100:
@RReverser this issue seems kind of gnarly.  Can/should we instead just remove the .ps1 launchers?   What would be the downside of doing that?

RReverser:
They do work a lot better with Ctrl+C behaviour (which for .bat you can't override). I wouldn't want to lose that because of one edge case.

This PR looks like straight up generated by LLM though (at least the description is fairly typical), I'm not sure it's the right fix for the issue.

PR #25592, opened 2025-10-19
Merge assignWasmExports and assignWasmSymbols. NFC

sbc100:
These two functions serve exactly the same purpose so just make them into a single function.

There are still two different methods by which this gets defined but from the POV of the caller they are the same thing.

### DIFF
--- a/src/lib/libcore.js
+++ b/src/lib/libcore.js
@@ -1615,14 +1615,16 @@ addToLibrary({
 #endif
 
 #if !DECLARE_ASM_MODULE_EXPORTS
-  // When DECLARE_ASM_MODULE_EXPORTS is not set we export native symbols
-  // at runtime rather than statically in JS code.
-  $exportWasmSymbols__deps: ['$asmjsMangle',
+  // When DECLARE_ASM_MODULE_EXPORTS is set, this function is programatically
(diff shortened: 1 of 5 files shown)


src/lib/libcore.js
```diff
@@ -1617,12 +1617,12 @@ addToLibrary({
...
   // When DECLARE_ASM_MODULE_EXPORTS is not set we export native symbols
   // at runtime rather than statically in JS code.
-  $exportWasmSymbols__deps: ['$asmjsMangle',
+  $assignWasmExports__deps: ['$asmjsMangle',
 #if DYNCALLS || !WASM_BIGINT
     , '$dynCalls'
 #endif
   ],
-  $exportWasmSymbols: (wasmExports) => {
+  $assignWasmExports: (wasmExports) => {
```
juj, line 1627: Can you add a comment before this function that this is library-defined in !DECLARE_ASM_MODULE_EXPORTS, but programmatically generated in DECLARE_ASM_MODULE_EXPORTS.. otherwise it could be highly confusing to see a library function that doesn't match anything in the output.

sbc100: Added comments.

PR #25593, opened 2025-10-19
Simplify makeRetainedCompilerSettings. NFC

sbc100:
In strict mode we don't define any of the legacy settings of the `ignore` mechanism here was doing nothing.


sbc100:
(Landing TBR since this is fairly trivial, and I have a larger change on top of it).

PR #25594, opened 2025-10-19
Enforce python import ordering rules. NFC

sbc100:
This change was automatically generated using `ruff check --fix`

### DIFF
--- a/emrun.py
+++ b/emrun.py
@@ -38,10 +38,11 @@ import time
 from operator import itemgetter
 
 if sys.version_info.major == 2:
+  from urllib import unquote
+
   import SocketServer as socketserver
   from BaseHTTPServer import HTTPServer
   from SimpleHTTPServer import SimpleHTTPRequestHandler
-  from urllib import unquote
   from urlparse import urlsplit
 else:
   import socketserver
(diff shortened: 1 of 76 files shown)


emrun.py
```diff
@@ -38,10 +38,11 @@
 from operator import itemgetter
 
 if sys.version_info.major == 2:
+  from urllib import unquote
+
   import SocketServer as socketserver
   from BaseHTTPServer import HTTPServer
   from SimpleHTTPServer import SimpleHTTPRequestHandler
-  from urllib import unquote
   from urlparse import urlsplit
```
kripken, line 46: what is the rule here, do you know? the new version seems... arbitrary at first glance.

sbc100: I don't know about this one.. the rule is generally that things should be alphabetical and that system libraries always come first.

My guess is that because `urlparse` was removed in python3 and this code is supposed run on python2 that maybe the formatter is confused about local vs system modules here.  I think I can just revert this file and mark it with an exception.

sbc100: Also, we could just remove python2 support from this file at this point I think.

kripken:
Oh, I missed the auto-merge... anyhow, it's ok as-is.

emrun.py
sbc100: Reverted this file.

PR #21478, opened 2024-03-06
Embed wasm by encoding binary data as UTF-8 code points.

juj:
Quick and dirty test to see what the CI says.

Possible alternative to #21426.

PR #21426, opened 2024-02-26
Add new option -sGZIP_EMBEDDINGS that gzip the wasm output

msqr1:
Doesn't work with WASM_ASYNC_COMPILATION off, this new setting gzip the wasm binary before embedding, which can reduce code size. It will decompress via JS DecompressionStream. Can increase startup time, but the code size benefit can be massive for a big project. For mine: -O0, 5.3MB --> 1.7MB, -O3, 3.7MB --> 1.4MB. Implements #21383. Edit any 

Fixes: #21383

PR #25599, opened 2025-10-20
Implement -s SINGLE_FILE_BINARY_ENCODE=1 option

juj:
Implement -s SINGLE_FILE_BINARY_ENCODE=1 option to embed Wasm binary as binary-encoded form instead of base64 form in SINGLE_FILE mode. Continuation of #21478.

For comparison of code size, see https://github.com/emscripten-core/emscripten/pull/21426#issuecomment-1979536160.

### DIFF
--- a/src/binaryDecode.js
+++ b/src/binaryDecode.js
@@ -1,0 +1,10 @@
+// Prevent Closure from minifying the binaryDecode() function, or otherwise
+// Closure may analyze through the WASM_BINARY_DATA placeholder string into this
+// function, leading into incorrect results.
+/** @noinline */
+function binaryDecode(bin) {
+  for (var i = 0, l = bin.length, o = new Uint8Array(l); i < l; ++i) {
+    o[i] = bin.charCodeAt(i) - 1;
+  }
+  return o;
+}
--- a/tools/link.py
+++ b/tools/link.py
@@ -2986,7 +2986,7 @@ def binary_encode(filename):
 def get_subresource_location(path, mimetype='application/octet-stream'):
   if settings.SINGLE_FILE:
     if settings.SINGLE_FILE_BINARY_ENCODE:
-      return binary_encode(utils.read_binary(path))
+      return binary_encode(path)
     return f'data:{mimetype};base64,{base64_encode(path)}'
   else:
     return os.path.basename(path)
--- a/src/proxyClient.js
+++ b/src/proxyClient.js
@@ -132,7 +132,18 @@ var frameId = 0;
 
 // Worker
 
-var filename = '<<< filename >>>';
+var filename = "<<< filename >>>";
+
+#if SINGLE_FILE && SINGLE_FILE_BINARY_ENCODE
+#include "binaryDecode.js"
+
+#if ENVIRONMENT_MAY_BE_NODE
+if (ENVIRONMENT_IS_NODE) filename = "data:text/javascript;base64," + Buffer.from(binaryDecode(filename)).toString('base64');
+else
+#endif
+  filename = URL.createObjectURL(new Blob([binaryDecode(filename)], {type: 'application/javascript'}));
+
+#endif
 
 var worker = new Worker(filename);
 
--- a/src/runtime_common.js
+++ b/src/runtime_common.js
@@ -28,16 +28,7 @@ function growMemViews() {
 #endif
 
 #if SINGLE_FILE && SINGLE_FILE_BINARY_ENCODE && !WASM2JS
-// Prevent Closure from minifying the binaryDecode() function, or otherwise
-// Closure may analyze through the WASM_BINARY_DATA placeholder string into this
-// function, leading into incorrect results.
-/** @noinline */
-function binaryDecode(bin) {
-  for (var i = 0, l = bin.length, o = new Uint8Array(l); i < l; ++i) {
-    o[i] = bin.charCodeAt(i) - 1;
-  }
-  return o;
-}
+#include "binaryDecode.js"
 #endif
 
 #if MODULARIZE


src/preamble.js
```diff
@@ -410,6 +410,10 @@ function instrumentWasmTableWithAbort() {
 }
 #endif
 
+#if SINGLE_FILE && SINGLE_FILE_BINARY_ENCODE && !WASM2JS
+#include "binaryDecode.js"
```
sbc100: Did you man to add this file to source control?

test/test_other.py
```diff
@@ -9475,8 +9475,9 @@ def test_standalone_system_headers(self, prefix):
     'closure': (False, True),
   })
   @parameterized({
-    '': (True,),
-    'disabled': (False,),
+    '': (1,),
+    'disabled': (0,),
+    'binary_encode': (2,),
```
sbc100: Can we just add another boolean param here rather than a magic 2 value?

juj:
> Do you have some rough number you could put in the PR description?

Added.

> How about adding a codesize check that asserts the overall result is smaller?

Added.

> Assuming no unforseen issues do you expect we could just make this setting always-enabled one day?

Maybe one day. For now I think it's good to have the opt-out to give people the fallback if issues come up.

src/preamble.js
```diff
@@ -423,27 +427,35 @@ function getWasmBinary(file) {}
 #else
 
 function findWasmBinary() {
-#if SINGLE_FILE
+#if SINGLE_FILE && SINGLE_FILE_BINARY_ENCODE && !WASM2JS
+  return binaryDecode('<<< WASM_BINARY_DATA >>>');
+#elif SINGLE_FILE
   return base64Decode('<<< WASM_BINARY_DATA >>>');
+#elif AUDIO_WORKLET || !EXPORT_ES6
+  // For an Audio Worklet, we cannot use `new URL()`.
```
sbc100: Its hard to see why the locateFile path is only used in EXPORT_ES6 mode... do you know why EXPORT_ES6 is here?

src/binaryDecode.js
```diff
@@ -0,0 +1,10 @@
+// Prevent Closure from minifying the binaryDecode() function, or otherwise
+// Closure may analyze through the WASM_BINARY_DATA placeholder string into this
+// function, leading into incorrect results.
+/** @noinline */
+function binaryDecode(bin) {
+  for(var i = 0, l = bin.length, o = new Uint8Array(l); i < l; ++i) {
```
sbc100: Space after `for`

src/preamble.js
sbc100: Do we not want to support this in MINIMAL_RUNTIME too?   If so maybe this could go in `runtime_common.js` instead?

sbc100:
Are there any cases where `SINGLE_FILE_BINARY_ENCODE` could make the total payload larger?

src/preamble.js
juj: I don't know. It is indeed a little bit odd code flow, which is why I cleaned the ifdefs a bit here as a drive-by. Something to ponder further in a separate PR.

src/preamble.js
juj: Thanks, we do. I lost this when resurrecting the code. Added back, and updated test to cover minimal runtime as well.

tools/link.py
```diff
@@ -163,6 +163,10 @@ def base64_encode(filename):
   return b64.decode('ascii')
 
 
+def base64_or_binary_encode(b):
+  return binary_encode(b) if settings.SINGLE_FILE and settings.SINGLE_FILE_BINARY_ENCODE else base64_encode(b)
```
sbc100: I think you can remove `settings.SINGLE_FILE` here (make it an assert if you like)

src/binaryDecode.js
```diff
@@ -0,0 +1,10 @@
+// Prevent Closure from minifying the binaryDecode() function, or otherwise
+// Closure may analyze through the WASM_BINARY_DATA placeholder string into this
+// function, leading into incorrect results.
+/** @noinline */
+function binaryDecode(bin) {
+  for (var i = 0, l = bin.length, o = new Uint8Array(l); i < l; ++i) {
+    o[i] = bin.charCodeAt(i) - 1;
+  }
+  return o;
+}
```
sbc100: Since this is only 5 lines i wonder if its worth creating a completely new flie?  Maybe just inline into `runtime_common.js`?

sbc100 (approved):

juj:
> Are there any cases where `SINGLE_FILE_BINARY_ENCODE` could make the total payload larger?

base64 encoding encodes each byte uniformly with +33% overhead.

Binary encoding encodes compatible bytes in the original byte (+0% overhead), but incompatible bytes (09h, 0Ch, 21h, 5Bh, 80h-FFh) will be encoded as two bytes, so with +100% overhead.

If the distribution of input bytes were uniform, this comes out to a +49.61% overhead. So a larger size in uncompressed form.

But the trick here is that

a) since all those +100% code points are byte-delimited, they compress efficiently with gzip and brotli compression, and
b) in practice those incompatible bytes are more rare (MSB is more common to be unset than set), so the end result when encoding a .wasm code file ends up being a major win in uncompressed form as well.

src/binaryDecode.js
juj: check

test/codesize/test_codesize_minimal_O0.expected.js
```diff
@@ -585,7 +585,8 @@ function createExportWrapper(name, nargs) {
 var wasmBinaryFile;
 
 function findWasmBinary() {
-    return locateFile('a.out.wasm');
+  // For an Audio Worklet, we cannot use `new URL()`.
```
sbc100: Maybe the new location of this comment is misplaced?

juj: Hmm the code size before was relying on the comment being swallowed by virtue of being at the end of an #ifdef.. moved it back up there.

kripken:
@juj @sbc100 This caused a regression I bisected to here, reported on the mailing list:

https://groups.google.com/g/emscripten-discuss/c/E_HmYqXGjN8

sbc100:
@juj can you take a look at this early next week?   If you don't have time, or there is no easy fix I suggest we just flip the default for `SINGLE_FILE_BINARY_ENCODE` back for false until we have a fix.

juj:
Thanks for the ping - posted PR https://github.com/emscripten-core/emscripten/pull/25704 to address this.

PR #25603, opened 2025-10-20
Split js library tests out of test_other.py into their own file.  NFC

sbc100:
The `test_other.py` is over 15,000 lines long so its to to carve out larger subsections where possible.

The tests themselves are all unmodified aside from one that was referring to filenames that got moved.


kripken:
I am for smaller files, but a downside though is now running `other` runs less. One needs to remember to run more test suites locally to get the same coverage as before. Can we separate out tests from the same suite?

sbc100:
> I am for smaller files, but a downside though is now running `other` runs less. One needs to remember to run more test suites locally to get the same coverage as before. Can we separate out tests from the same suite?

Not really.   We could make an alias for `other` that runs all these though.

I think we can figure that out after this PR though.

sbc100:
> > I am for smaller files, but a downside though is now running `other` runs less. One needs to remember to run more test suites locally to get the same coverage as before. Can we separate out tests from the same suite?
> 
> Not really. We could make an alias for `other` that runs all these though.
> 
> I think we can figure that out after this PR though.

As long as the CI runs this suite at least I don't think its a big deal.  We can come up with some kind of local alias for devs who don't want to type/remember the different suite names as a follow up think.

PR #25607, opened 2025-10-21
Remove references to `RELOCATABLE`/`LINKABLE` from test code. NFC

sbc100:
These settings are due for deprecation/removal, plus none of these tests were really depending on them specifically (as opposed to dynamic linking in general).

Replace this usage with three specific simple tests that we can remove when we remove these settings.

See https://github.com/emscripten-core/emscripten/issues/25262

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -1580,12 +1580,11 @@ int f() {
     self.assertContained(['unknown file type', "libfoo.a: archive member 'native.o' is neither Wasm object file nor LLVM bitcode"], stderr)
 
   def test_export_all(self):
-    lib = r'''
+    create_file('main.c', r'''
       #include <stdio.h>
       void libf1() { printf("libf1\n"); }
       void libf2() { printf("libf2\n"); }
-    '''
-    create_file('lib.c', lib)
+    ''')
 
     create_file('pre.js', '''
       Module.onRuntimeInitialized = () => {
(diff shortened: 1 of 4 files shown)


test/test_other.py
```diff
@@ -1596,8 +1595,10 @@ def test_export_all(self):
 
     # Explicitly test with -Oz to ensure libc_optz is included alongside
     # libc when `--whole-archive` is used.
-    self.emcc('lib.c', ['-Oz', '-sEXPORT_ALL', '-sLINKABLE', '-Wno-deprecated', '--pre-js', 'pre.js'], output_filename='a.out.js')
-    self.assertContained('libf1\nlibf2\n', self.run_js('a.out.js'))
+    self.do_runf('main.c', 'libf1\nlibf2\n', cflags=['-Oz', '-sEXPORT_ALL', '-sMAIN_MODULE', '--pre-js', 'pre.js'])
+
+    # Without the `-sEXPORT_ALL` these symbols will not be visible from JS
+    self.do_runf('main.c', '_libf1 is not defined', assert_returncode=NON_ZERO, cflags=['-Oz', '-sMAIN_MODULE', '--pre-js', 'pre.js'])
```
kripken, line 1601: isn't this tested elsewhere? it seems new here.

sbc100: I just thought it might be good to test the negative case too when testing the `-sEXPORT_ALL` flag so I added this to the test.

kripken: Fair enough, this is `test_export_all` I guess. sg.

PR #25608, opened 2025-10-21
Split test decorators into their own file.  NFC

sbc100:


test/decorators.py
```diff
@@ -0,0 +1,631 @@
+# Copyright 2025 The Emscripten Authors.  All rights reserved.
+# Emscripten is available under two separate licenses, the MIT license and the
+# University of Illinois/NCSA Open Source License.  Both these licenses can be
+# found in the LICENSE file.
+
+"""Common decorators shared between the various suites.
+
+When decorators are only using a single file they can be defined locally.  If they
```
kripken:
```suggestion
When decorators are only used in a single file they can be defined locally.  If they
```

test/decorators.py
```diff
@@ -0,0 +1,631 @@
...
+are using from multiple locations they should be defined here.
```
kripken:
```suggestion
are used from multiple locations they should be defined here.
```

PR #25609, opened 2025-10-21
Fix WebAssembly global exports with DECLARE_ASM_MODULE_EXPORTS=0(issue #25556)

devalgupta404:
## Root Cause
The `exportWasmSymbols` function had conditional logic that prevented proper global export:

```javascript
// before :
if (typeof exportedSymbol.value === 'undefined') {
  // Export functions 
} 
```

## Solution
Removed the if condition and simplified to export all symbols unconditionally:

```javascript
// After (fixed):
// Export all symbols (both functions and globals) to the global scope
// when DECLARE_ASM_MODULE_EXPORTS=0
globalThis[name] = exportedSymbol;
```

## Changes
- **Modified**: `src/lib/libcore.js` - Simplified `exportWasmSymbols` function
- **Enabled**: `test/test_core.py` - Removed `@no_omit_asm_module_exports` decorator

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -9711,7 +9711,6 @@ NODEFS is no longer included by default; build with -lnodefs.js
     'dylink': [True],
   })
   @no_esm_integration('https://github.com/emscripten-core/emscripten/issues/25543')
-  @no_omit_asm_module_exports('https://github.com/emscripten-core/emscripten/issues/25556')
   def test_wasm_global(self, dynlink):
     if '-flto' in self.cflags or '-flto=thin' in self.cflags:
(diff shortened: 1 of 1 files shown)


devalgupta404:
@sbc100 please review it

kleisauke:
This is expected to be addressed by PR #25568.

sbc100:
@devalgupta404 are you using DECLARE_ASM_MODULE_EXPORTS=0 yourself?   Can I ask why you are using it?

sbc100:
This a fix for https://github.com/emscripten-core/emscripten/issues/25556 right?

I don't think this is the right fix though since it would change the way immutable globls are currently exported.. right now they are plain numbers, after this change they would be Wasm globals.

PR #25613, opened 2025-10-21
Optimize binaryEncode() string quoting

juj:
One of the optimizations of Closure compiler is to choose the string quote character that results in fewer escaping in the string itself.

I.e. Closure compiler will turn

```js
var x = '\'\'\'\'\'';
var y = "\"\"\"\"\"";
```
into
```js
var x = "'''''";
var y = '"""""';
```
by selecting the string quote char that allows the contents to have fewer escapse.

However, in the SINGLE_FILE mode, we emit the Wasm code inside the .js file after Closure has run. So all Closure sees is

```js
  return binaryDecode("<<< WASM_BINARY_DATA >>>");
```
when optimizing.

This PR implements the same smart string quote selection optimization directly into the `binaryEncode()` function, by checking if there are fewer `'`s or `"`s in the binary content that is to be encoded.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -2668,12 +2668,12 @@ def generate_worker_js(target, options, js_target, target_basename):
   utils.write_file(target, target_contents, options.output_eol)
 
 
-def worker_js_script(proxy_worker_filename):
+def worker_js_script(proxy_worker_filename_enclosed_in_quotes):
   web_gl_client_src = read_file(utils.path_from_root('src/webGLClient.js'))
   proxy_client_src = building.read_and_preprocess(utils.path_from_root('src/proxyClient.js'), expand_macros=True)
-  if not settings.SINGLE_FILE and not os.path.dirname(proxy_worker_filename[1:-1]):
-    proxy_worker_filename = f'"./{proxy_worker_filename[1:-1]}"'
-  proxy_client_src = do_replace(proxy_client_src, '"<<< filename >>>"', proxy_worker_filename)
+  if not settings.SINGLE_FILE and not os.path.dirname(proxy_worker_filename_enclosed_in_quotes[1:-1]):
+    proxy_worker_filename_enclosed_in_quotes = f'"./{proxy_worker_filename_enclosed_in_quotes[1:-1]}"'
+  proxy_client_src = do_replace(proxy_client_src, '"<<< filename >>>"', proxy_worker_filename_enclosed_in_quotes)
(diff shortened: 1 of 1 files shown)


tools/link.py
```diff
@@ -2441,7 +2437,11 @@ def phase_binaryen(target, options, wasm_target):
   if final_js and settings.SINGLE_FILE and not settings.WASM2JS:
     js = read_file(final_js)
 
-    js = do_replace(js, '<<< WASM_BINARY_DATA >>>', base64_or_binary_encode(wasm_target))
+    if settings.SINGLE_FILE_BINARY_ENCODE:
+      js = do_replace(js, '"<<< WASM_BINARY_DATA >>>"', binary_encode(wasm_target))
+    else:
+      js = do_replace(js, '<<< WASM_BINARY_DATA >>>', base64_encode(wasm_target))
+    assert '<<< WASM_BINARY_DATA >>>' not in js
```
sbc100: The do_replace helper already asserts that the search pattern is present ... is this also needed for some reason in addition to that?

juj: Yeah, I added this here to guarantee that e.g. someone refactoring `"<<< WASM_BINARY_DATA >>>"` into a `'<<< WASM_BINARY_DATA >>>'` would trigger an error.

juj: oh right, now I see what you mean.. yeah, that does make the assert redundant.

juj:
Ping, any thoughts here?

juj:
> I wonder if its worth the extra complexity here in the code generatation.
> 
> Is it worth those 7 bytes?

The 7 bytes is only in a test case that itself is only 5400 bytes in size.

Running the encoder on one of the smallest .wasm files from Unity output:
- old encoding: 1,924,603 bytes
- new encoding: 1,896,794 bytes (-27,809 bytes)

So the wins do scale with the size of the .wasm file, so not just a constant 7 bytes.

> BTW, are you making use of this in Unity? Do you, or you users, use the `-sSINGLE_FILE` setting a lot? (And why?)

Yes. The SINGLE_FILE output from Unity is for use in the [MRAID standard](https://www.iab.com/wp-content/uploads/2015/08/IAB_MRAID_v2_FINAL.pdf). What they do is standardize on the CDN storage and lifecycle of an interactive (playable) ad.

The restrictions that the CDN people state is that the ads need to be a single .html file containing everything embedded in it (no extra XHRs or Fetches allowed), and the size limit is <= 5MB **un**compressed.

PR #25614, opened 2025-10-21
Have test harness delete all old Emscripten files in the temp directory

juj:
When starting a test run, delete all old Emscripten files in the temp directory to help avoid runaway temp file leaks filling up a CI system hard drive.

Do it before starting a run, so that developers can still access test files in debug logs.

### DIFF
--- a/test/runner.py
+++ b/test/runner.py
@@ -530,8 +530,10 @@ def cleanup_emscripten_temp():
   that look like they might have been created by Emscripten."""
   for entry in os.listdir(shared.TEMP_DIR):
     if entry.startswith(('emtest_', 'emscripten_')):
+      entry = os.path.join(shared.TEMP_DIR, entry)
       try:
-        utils.delete_dir(os.path.join(shared.TEMP_DIR, entry))
+        if os.path.isdir(entry):
+          utils.delete_dir(entry)
       except Exception:
         pass
 


sbc100:
If this specific to the parallel test runner or should it happen on startup in general?

We do have `EMTEST_DETECT_TEMPFILE_LEAKS` which has a mechansim to detect leaks, so I would hope they do not happen much in practice.

It looks like we enable `EMTEST_DETECT_TEMPFILE_LEAKS` as the main CI, but not for browser tests.

juj:
> If this specific to the parallel test runner or should it happen on startup in general?

This should run in both singlethreaded and parallel runner startup.

> We do have `EMTEST_DETECT_TEMPFILE_LEAKS` which has a mechansim to detect leaks, so I would hope they do not happen much in practice.

Yeah, I remember creating that a long time ago to combat temp file leaks. Added that env. var. now to my CI bots, but wanted to ensure that even if these leaks do happen, that they won't be able to take my CI bots down.

I see on all my CI bots that there are a bunch of `emtest_` directories that have crept up over time. Some of those are definitely caused by test harness runs aborting mid-flight (e.g. due to CI disconnect, which aborts in buildbot).

sbc100:
> > If this specific to the parallel test runner or should it happen on startup in general?
> 
> This should run in both singlethreaded and parallel runner startup.
> 
> > We do have `EMTEST_DETECT_TEMPFILE_LEAKS` which has a mechansim to detect leaks, so I would hope they do not happen much in practice.
> 
> Yeah, I remember creating that a long time ago to combat temp file leaks. Added that env. var. now to my CI bots, but wanted to ensure that even if these leaks do happen, that they won't be able to take my CI bots down.
> 
> I see on all my CI bots that there are a bunch of `emtest_` directories that have crept up over time. Some of those are definitely caused by test harness runs aborting mid-flight (e.g. due to CI disconnect, which aborts in buildbot).

Yes, I think we explictly do not clean stuff up when the run is interrupted, since this is good for debugging.

sbc100:
Looking into this I noticed we were creating a lot more temp directories than we needed to: https://github.com/emscripten-core/emscripten/pull/25615

juj:
> Yes, I think we explictly do not clean stuff up when the run is interrupted, since this is good for debugging.

Agreed. That is why the idea here was to do the cleanup at startup, and not at end (or at `atexit`)

sbc100:
> > Yes, I think we explictly do not clean stuff up when the run is interrupted, since this is good for debugging.
> 
> Agreed. That is why the idea here was to do the cleanup at startup, and not at end (or at `atexit`)

Sounds good to me.

I would feel more comfortable if it was just the `emtest_*` that we cleanup, since that is obviously stuff created by the test harness itself.   Are you times when the compiler itself is not cleaning up its stuff too?

juj:
> Are you times when the compiler itself is not cleaning up its stuff too?

The majority of the items are from `emtest_`, but I did see some `emscripten_` there as well, which is why I added that pattern there as well.

I think combining this with running with `EMTEST_DETECT_TEMPFILE_LEAKS` should ensure that there won't be leaks in that area on `emscripten_` either.

sbc100:
I found some more usages of tempfile that we can completely remove : https://github.com/emscripten-core/emscripten/pull/25617

sbc100 (approved):

test/runner.py
```diff
@@ -525,6 +525,17 @@ def configure():
...
 
+def cleanup_emscripten_temp():
+  """Deletes all files and directories under Emscripten
+  that look like they might have been created by Emscripten."""
+  for entry in os.listdir(shared.TEMP_DIR):
+    if entry.startswith(('emtest_', 'emscripten_')):
+      try:
+        utils.delete_dir(os.path.join(shared.TEMP_DIR, entry))
+      except Exception:
+        pass
```
sbc100, line 538: How about using `force_delete_dir` here... then I think you don't need the try/catch

test/runner.py
```diff
@@ -525,6 +525,17 @@ def configure():
   common.configure_test_browser()
 
 
+def cleanup_emscripten_temp():
+  """Deletes all files and directories under Emscripten
+  that look like they might have been created by Emscripten."""
+  for entry in os.listdir(shared.TEMP_DIR):
+    if entry.startswith(('emtest_', 'emscripten_')):
```
sbc100, line 532: If `and os.path.isdir()`?

test/runner.py
juj: `force_delete_dir()` raises an exception if the deletion failed. The thinking here was that it wouldn't be catastrophic if that happens.

sbc100: Fair enough

PR #18214, opened 2022-11-16
parallel_testsuite.py: Use multiprocessing pool

sbc100:
Use the `Pool` seems to address the longstanding issue we have had with interrupting the test suite using `crtl-C`.

Its also a lot less code.

Also, limit the pools size to the number of tests being run.  No need to start 56 processes if we are just running 2 tests.

I measured the performance of the running the core2 test suite both before and after this change and it was not effected.

See: #18213
Supersedes: #11299

PR #6150, opened 2018-01-28
Parallel other tests

saschanaz:
Fixes #5817, fixes #6130 

Tried implementing what @juj [suggested](https://github.com/kripken/emscripten/issues/5817#issuecomment-360941775):

>I think the solution forward here might be to assign unique/dedicated temp directories to each parallel test runner process in `core` and `other`? That is, for each parallel test runner thread, append a unique subdirectory path to `EMCC_TEMP_DIR` for each of them, so that they all operate in unique directories? Then the leak checker could also be enabled for the `core` suite.

**Edit:** `other` now only takes 16 minutes on CircleCI (while ~40 minutes on Travis)!

PR #25615, opened 2025-10-21
Avoid using a new temp directory for each parallel test. NFC

sbc100:
It should be fine for all parallel tests to using the same TEMP_DIR.

The only limitation is on tests that use the canonical temp directory (e.g. EMCC_DEBUG).

It looks like we used to use a unique temp directory per process (See #6150). However that was changed to one-directory-per-test in #18214, although I don't remember why.

With this change we don't create a new temp directory except for tests that are marked as `uses_canonical_tmp`

Note that each test already run in its own unique directory, which is a different concept the TEMP_DIR.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -21,6 +21,7 @@ import struct
 import subprocess
 import sys
 import tarfile
+import tempfile
 import time
 from datetime import datetime
 from functools import wraps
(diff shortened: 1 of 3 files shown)


test/test_other.py
```diff
@@ -153,6 +154,13 @@ def uses_canonical_tmp(func):
   """
   @wraps(func)
   def decorated(self, *args, **kwargs):
+    if self.runningInParallel():
+      # Because the canonical temp directory is a global/central location, when
+      # tests use it in parallel they need to temporarily override the temp directory
+      # to ovoid clobbering each each.
```
brendandahl, line 160:
```suggestion
      # to avoid clobbering each each.
```

Issue #25618, opened 2025-10-22
JS runtime exception with iOS 26.1 accessing t.alpha on null under some circumstances


andvgal:
It was noticed that some third party pre-built components fail in iOS 26.1 under some conditions while accessing t.alpha on null. Despite the exact details cannot be provided, below is the likely offending place.

https://github.com/emscripten-core/emscripten/blob/c81dbdd5f564665ef3124635af89cfbaed546698/src/lib/libhtml5_webgl.js#L311

According to MDN (https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getContextAttributes), the function `WebGLRenderingContext::getContextAttributes()` can return `null` under some circumstances.

Therefore, another safety check should solve the problem, e.g.:

```patch
diff --git a/src/lib/libhtml5_webgl.js b/src/lib/libhtml5_webgl.js
index 63f8d2ad3..fa298e3a9 100644
--- a/src/lib/libhtml5_webgl.js
+++ b/src/lib/libhtml5_webgl.js
@@ -309,6 +309,7 @@ var LibraryHtml5WebGL = {
     var t = c.GLctx;
     if (!t) return {{{ cDefs.EMSCRIPTEN_RESULT_INVALID_TARGET }}};
     t = t.getContextAttributes();
+    if (!t) return {{{ cDefs.EMSCRIPTEN_RESULT_INVALID_TARGET }}};
 
     {{{ makeSetValue('a', C_STRUCTS.EmscriptenWebGLContextAttributes.alpha, 't.alpha', 'i8') }}};
     {{{ makeSetValue('a', C_STRUCTS.EmscriptenWebGLContextAttributes.depth, 't.depth', 'i8') }}};
```

andvgal:
A blind fix: https://github.com/emscripten-core/emscripten/pull/25619

andvgal:
The PR is merged.

PR #25619, opened 2025-10-22
Fix JS runtime exception on WebGL getContextAttributes (#25618)

andvgal:
This is a blind fix for https://github.com/emscripten-core/emscripten/issues/25618

### DIFF
--- a/src/lib/libhtml5_webgl.js
+++ b/src/lib/libhtml5_webgl.js
@@ -306,9 +306,8 @@ var LibraryHtml5WebGL = {
     if (!a) return {{{ cDefs.EMSCRIPTEN_RESULT_INVALID_PARAM }}};
     c = GL.contexts[c];
     if (!c) return {{{ cDefs.EMSCRIPTEN_RESULT_INVALID_TARGET }}};
-    var t = c.GLctx;
+    var t = c.GLctx?.getContextAttributes();
     if (!t) return {{{ cDefs.EMSCRIPTEN_RESULT_INVALID_TARGET }}};
-    t = t.getContextAttributes();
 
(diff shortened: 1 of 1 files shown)


src/lib/libhtml5_webgl.js
```diff
@@ -309,6 +309,7 @@ var LibraryHtml5WebGL = {
     var t = c.GLctx;
     if (!t) return {{{ cDefs.EMSCRIPTEN_RESULT_INVALID_TARGET }}};
     t = t.getContextAttributes();
+    if (!t) return {{{ cDefs.EMSCRIPTEN_RESULT_INVALID_TARGET }}};
```
juj, line 310: How about

```js
var t = c.GLctx?.getContextAttributes();
if (!t) return {{{ cDefs.EMSCRIPTEN_RESULT_INVALID_TARGET }}};
```
?

Although do note that if the GL context is lost, then all subsequent GL calls will fail as well. So if you have a code path that can somehow recover (if it is an iOS 26.1 internal bug, one might not exist), you will need to call `emscripten_is_webgl_context_lost()` or register to `emscripten_set_webglcontextlost_callback()` to be notified of a context loss, to do something else.

I.e. the issue will likely not be isolated to this function, but all subsequent GL operations will fail as well.

andvgal: First of all, that is a blind fix for a runtime exception. The fact that the context is not usable any more is out of scope.

The refactoring suggestion imposes another minimal ECMAScript version. I am not that familiar with the project and its application, but the code around suggests ES6-level of coding about when WebAssembly got first introduced. Optional chaining is ES2020.

juj: Emscripten project uses Babel to transpile to older ES standards. Using `?.` is ok here.

andvgal: OK, then following your advice.

PR #25621, opened 2025-10-22
[SPLIT_MODULE] Fix imports into deferred module when loading

Mintyboi:
Fix an error when attempted to reserve a new memory space in the deferred module when building with MEMORY64 + SPLIT_MODULE.
```
TypeError: Cannot convert <func ptr> to a BigInt
```

In a mem64 build, we call applySignatureConversions on `malloc()` in the wasmExports. `applySignatureConversions()` wraps the exports and converts the return value to a javascript number. The wasmExports is passed into the deferred module during instantiation so when we call a malloc in the deferred module, we'd be calling the wrapped function, hitting into this error.

Furthermore, we are making a direct call when calling a malloc, so strictly speaking, we should not need to go through the JS trampoline.

### DIFF
--- a/test/other/test_split_module.c
+++ b/test/other/test_split_module.c
@@ -1,7 +1,5 @@
 #include <stdio.h>
 #include <emscripten.h>
-#include <stdlib.h> // For malloc and free
-#include <stdint.h> // For int64_t
 
 int foo() {
   return 42;


sbc100:
I only see a test change so I guess this PR is WIP?

Mintyboi:
> I only see a test change so I guess this PR is WIP?

Yes, I plan to have the tests fail first and add in the fix later, which should fix the tests as well 🙏

sbc100:
Sure, we can't actually land the failing test right?  So I guess I'll wait to review this?

sbc100:
(If you are just experimenting then maybe do that with the PR marked as "DRAFT" or you can also do that in your own fork too)

Mintyboi:
@sbc100 I have reproduced the issue I'm facing in the test suite and some attempted fix that works in resolving this issue.
However these changes does not sit too well with me, can I check if you have any suggestions / alternative approach to resolve the issue where we are calling a wrapper with incorrect types.

sbc100:
@aheejin @tlively who have been working on module splitting more than I have.

The solution does seem reasonable yes.  We so something similar in dynamic linking where we maintain `wasmImports` as the symbol table we feed to new wasm module  This gets populated with raw exports from each module as opposed to wrapped exports.

Can you please wrap all these changes in `#if SPLIT_MODULE` to avoid bloating the generated code for those not use module splitting?

test/other/test_split_module.c
```diff
@@ -1,10 +1,20 @@
...
 
 int foo() {
   return 42;
 }
 
 EMSCRIPTEN_KEEPALIVE void say_hello() {
   printf("Hello! answer: %d\n", foo());
+
+  int64_t* ptr = (int64_t*)malloc(sizeof(int64_t) * 10);
+  if (ptr == NULL) {
```
sbc100, line 12: I con't think you need to use `int64_t` here to repro this problem.   I think the fact that malloc and free take/return pointers should be enough.

Can you just do `void* ptr = malloc(10)` and repro the failure?

aheejin:
Sorry, I'm not sure if I understand the problem and the solution here.

Can you give the command line to run your test? It'd be good if we have a test in one of `test_***.py` files that actually runs the .c file you added.

Also... I guess I'm missing something, but the solution just looks like renaming `wasmExports` to `wasmRawExports`..? Why does this solve the problem?

sbc100:
> Sorry, I'm not sure if I understand the problem and the solution here.
> 
> Can you give the command line to run your test? It'd be good if we have a test in one of `test_***.py` files that actually runs the .c file you added.
> 
> Also... I guess I'm missing something, but the solution just looks like renaming `wasmExports` to `wasmRawExports`..? Why does this solve the problem?

I think the issue is that if we just use `wasmExports` this map start life as the raw wasm exports but in some configuration we add JS wrapper functions around these (e.g. in wasm64).    However with module splitting (just like dynamic linking) we never want to feed there wrapper function as imports into another wasm module, we want to use the original exports in this case. 

The issue only shows up course then wrapper functions are used.

Mintyboi:
> Sorry, I'm not sure if I understand the problem and the solution here.
> 
> Can you give the command line to run your test? It'd be good if we have a test in one of `test_***.py` files that actually runs the .c file you added.
> 
> Also... I guess I'm missing something, but the solution just looks like renaming `wasmExports` to `wasmRawExports`..? Why does this solve the problem?

Thanks for explaning the issue @sbc100
wasmRawExports is saved before we apply the wrappers in preamble.js‎L#738.
I have update the test_split_module tests that you can execute with `python runner.py other.test_split_module`. And to reproduce the issue, you can rename wasmRawExports back to wasmExports in `splitModuleProxyHandler()`

test/other/test_split_module.c
Mintyboi: You're right.
Simplified test case in 568bd8a0a9d8ee5fd81410ca7a6427eee63a0779

test/other/test_split_module.c
```diff
@@ -1,10 +1,20 @@
 #include <stdio.h>
 #include <emscripten.h>
+#include <stdlib.h> // For malloc and free
+#include <stdint.h> // For int64_t
```
sbc100: This line is no longer needed

aheejin:
> > Sorry, I'm not sure if I understand the problem and the solution here.
> > Can you give the command line to run your test? It'd be good if we have a test in one of `test_***.py` files that actually runs the .c file you added.
> > Also... I guess I'm missing something, but the solution just looks like renaming `wasmExports` to `wasmRawExports`..? Why does this solve the problem?
> 
> I think the issue is that if we just use `wasmExports` this map start life as the raw wasm exports but in some configuration we add JS wrapper functions around these (e.g. in wasm64). However with module splitting (just like dynamic linking) we never want to feed there wrapper function as imports into another wasm module, we want to use the original exports in this case.
> 
> The issue only shows up course then wrapper functions are used.

Does this mean `SPLIT_MODULE` is not going to work with wasm64?

sbc100:
> > > Sorry, I'm not sure if I understand the problem and the solution here.
> > > Can you give the command line to run your test? It'd be good if we have a test in one of `test_***.py` files that actually runs the .c file you added.
> > > Also... I guess I'm missing something, but the solution just looks like renaming `wasmExports` to `wasmRawExports`..? Why does this solve the problem?
> > 
> > 
> > I think the issue is that if we just use `wasmExports` this map start life as the raw wasm exports but in some configuration we add JS wrapper functions around these (e.g. in wasm64). However with module splitting (just like dynamic linking) we never want to feed there wrapper function as imports into another wasm module, we want to use the original exports in this case.
> > The issue only shows up course then wrapper functions are used.
> 
> Does this mean `SPLIT_MODULE` is not going to work with wasm64?

the wrappers we generate for wasm64 are for the benefit of JS.   They basically turn bigint values into numbers so that JS developers don't have to worry about the "pointers are bigints" problem (at least not in most cases).

The importing function *between* one wasm module and another we explictly do not want to use these wrappers, and that would actually break, which this PR fixes I believe.

Mintyboi:
@sbc100 test sanity has been failing but it seems to be unrelated to my changes.
Do you know how I can fix it?

sbc100:
Yup, i think a rebase would fix but I'm also happy to just land as-is. 

Thanks for the contribution!

inolen:
Quick question on this PR -

I ran into an issue where I do actually need the original exports from the WASM instance - untouched by any internal routines, just the original value from instance.exports. I'm dynamically compiling some WASM modules and want to pass some exports from the current module to my new modules, skipping any JS wrappers in between.

I see this PR added the wasmRawExports variable, but these aren't exactly the raw exports - they're the exports before applySignatureConversions is ran, but after relocateExports and Asyncify.instrumentWasmExports are ran.

Were you intending for relocateExports / Asyncify.instrumentWasmExports to have ran on the value that's being stored in wasmRawExports, and if not, is it possible to move this declaration up a bit and remove the #if SPLIT_MODULE wrapper?

sbc100:
It sounds like you are trying to implement you own dynamic linking based on some emscripten-internals.   While we can't stop you from doing this you should be aware that its not a supported thing to do and the emscripten-internals could change in future releases.   Are you sure you want to be doing this and not using the dynamic linking support that comes with emscripten?

sbc100:
You are correct though `wasmRawExports` should be declared higher up here.. lets at least fix that now.

sbc100:
In emscripten's own dyanmic linking implementation the raw wasm symbols all get stored in `wasmImports` (each module that gets loaded adds to this map, including the main module).

Perhaps you could use `wasmImports` in the same way (maybe build with `-sMAIN_MODULE=2` which has almost no overhead these days).

sbc100:
Essentially `wasmImports` is the combination of the exports of all previously loaded modules.

inolen:
> It sounds like you are trying to implement you own dynamic linking based on some emscripten-internals. While we can't stop you from doing this you should be aware that its not a supported thing to do and the emscripten-internals could change in future releases. Are you sure you want to be doing this and not using the dynamic linking support that comes with emscripten?

It's something like that - I'm dynamically compiling Quake3's game-related VMs (which is some stack based byte code generated by an old C compiler) to WASM and then executing them. It's possible to massage them into a module that could be dlopen'd, but that'd be another can of worms. The engine has code emitters for other architectures, so there's already a pattern there to follow.

sbc100:
> > It sounds like you are trying to implement you own dynamic linking based on some emscripten-internals. While we can't stop you from doing this you should be aware that its not a supported thing to do and the emscripten-internals could change in future releases. Are you sure you want to be doing this and not using the dynamic linking support that comes with emscripten?
> 
> It's something like that - I'm dynamically compiling Quake3's game-related VMs (which is some stack based byte code generated by an old C compiler) to WASM and then executing them. It's possible to massage them into a module that could be dlopen'd, but that'd be another can of worms. The engine has code emitters for other architectures, so there's already a pattern there to follow.

So in this case Quake3 itself is the emscripten-built program and its trying to load JIT'd extensions to itself?

When you say "The engine has code emitters for other architectures, so there's already a pattern there to follow.", how do those other platforms load, if not via `dlopen`?   Are they literally writing to executable memory in the process?

inolen:
> So in this case Quake3 itself is the emscripten-built program and its trying to load JIT'd extensions to itself?
> 
> When you say "The engine has code emitters for other architectures, so there's already a pattern there to follow.", how do those other platforms load, if not via `dlopen`? Are they literally writing to executable memory in the process?

Exactly that. Quake3 split the main game code out of the engine, and mod authors could either compile the game code to a native system library, or their own "quake vm" format, which used a stack-based byte code emitted by LCC, enabling them to be cross-platform by default. Most mods shipped these VMs.

The engine has its own arm, x86, etc. code emitters to speed up the execution of the mods (and a fallback interpreter). I wrote a JS-based code emitter a decade ago (aka emit a big string and `eval()`), and I just updated that to now emit WASM. While trying to fiddle with JSPI however, I noticed I was hitting some JS wrappers switching contexts, and wanted to get rid of them.

sbc100:
Nice.  Sounds interesting.  If would recommend simply using `dlopen` (creating a dylib is just a question of adding tiny custom section at the start of the wasm).  

 If you don't want to use `dlopen` for some reason then I suggest you build with `-sMAIN_MODULE=2` and then reference the `wasmImports` object for all you symbols.

Issue #25627, opened 2025-10-22
Node.js 22.16.0 on Windows11-on-ARM is broken with respect to Wasm heap growth


juj:
`a.c`

```c
#include <pthread.h>
#include <stdlib.h>
#include <assert.h>

#define FINAL_HEAP_SIZE (64 * 1024 * 1024)

const char *_Atomic buffer = NULL;

void *thread_start(void *arg) {
  char *buf = malloc(FINAL_HEAP_SIZE);
  buf += FINAL_HEAP_SIZE - 1;
  *buf = 42;
  buffer = buf;
  return NULL;
}

int main() {
  pthread_t thr;
  int res = pthread_create(&thr, NULL, thread_start, NULL);
  while (!buffer);
  assert(*buffer == 42);
  return 0;
}
```

yields

```
C:\buildbot\win11-arm64-32gb-snapdragon-x\emscripten_win11_arm64\build\emscripten\main>emcc a.c -pthread -sALLOW_MEMORY_GROWTH -sINITIAL_MEMORY=32MB -sMAXIMUM_MEMORY=256MB -sMINIMAL_RUNTIME -g -Oz -o a.js
emcc: warning: -pthread + ALLOW_MEMORY_GROWTH may run non-wasm code slowly, see https://github.com/WebAssembly/design/issues/1271 [-Wpthreads-mem-growth]
emcc: warning: running limited binaryen optimizations because DWARF info requested (or indirectly required) [-Wlimited-postlink-optimizations]

C:\buildbot\win11-arm64-32gb-snapdragon-x\emscripten_win11_arm64\build\emscripten\main>%EMSDK_NODE% a.js
wasm://wasm/a.wasm-000f691e:1


RuntimeError: memory access out of bounds
    at a.wasm.__original_main (wasm://wasm/a.wasm-000f691e:wasm-function[19]:0x661)
    at a.wasm.main (wasm://wasm/a.wasm-000f691e:wasm-function[20]:0x684)
    at run (C:\buildbot\win11-arm64-32gb-snapdragon-x\emscripten_win11_arm64\build\emscripten\main\a.js:1022:3)
    at ready (C:\buildbot\win11-arm64-32gb-snapdragon-x\emscripten_win11_arm64\build\emscripten\main\a.js:48:5)
    at C:\buildbot\win11-arm64-32gb-snapdragon-x\emscripten_win11_arm64\build\emscripten\main\a.js:1071:5

Node.js v22.16.0
```

[...]

PR #25628, opened 2025-10-22
Skip other.test_pthread_growth test on Windows on ARM

juj:
Skip other.test_pthread_growth test on Windows on ARM if Node.js is 22.16.0. See https://github.com/emscripten-core/emscripten/issues/25627

In the future, after https://github.com/emscripten-core/emscripten/pull/25533 lands, we can move towards a direction of auto-detecting and auto-detecting skipping per Node version.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -11,6 +11,7 @@ import itertools
 import json
 import locale
 import os
+import platform
 import random
 import re
 import select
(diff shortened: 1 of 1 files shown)


test/test_other.py
```diff
@@ -12987,6 +12988,11 @@ def test_growable_arraybuffers(self):
     'minimal': (['-sMINIMAL_RUNTIME', '-sMODULARIZE', '-sEXPORT_NAME=MyModule'],),
   })
   def test_pthread_growth(self, cflags, pthread_pool_size = 1):
+    if WINDOWS and platform.machine() == 'ARM64':
+      # https://github.com/emscripten-core/emscripten/issues/25627
+      # TODO: Switch this to a "require Node.js 24" check
+      self.require_node_canary()
```
sbc100, line 12994: Should this go in the `else` block below because the `if` branch already requires node canary.

juj: Hmm how do you mean the `if` branch already requires node canary?

sbc100: I mean line 13000 below.

Since the `GROWABLE_ARRAYBUFFERS` already requires canary, and the comment here belngs instead in the `else` branch below.

sbc100 (commented):
lgtm % comment

test/test_other.py
juj: Right.. I think it probably reads cleaner to not intertwine this test skip/workaround with the growable arraybuffer logic part.

juj:
Looks like

```
FAIL [0.082s]: test_firstrun (test_sanity.sanity)
```
has regressed. (ignored that to land)

sbc100:
> Looks like
> 
> ```
> ======================================================================
> FAIL [0.082s]: test_firstrun (test_sanity.sanity)
> ----------------------------------------------------------------------
> ```
> 
> has regressed. (ignored that to land)

Already reverted.

PR #22216, opened 2024-07-10
Update SIMDe to 0.8.2

mr-c:
https://github.com/simd-everywhere/simde/compare/314452b...71fd833

Follow on to https://github.com/emscripten-core/emscripten/pull/19304

PR #25635, opened 2025-10-23
[docs] Neon updates from 6223c43 / #22216

mr-c:
This table wasn't updated when we upgraded SIMDe in #22216


mr-c:
Ping @tlively @sbc100

PR #25639, opened 2025-10-23
Add assert_fail test helper. NFC

sbc100:
I noticed that we were almost always following `expect_fail` with `assertContains` so I created a helper that does both.

In many cases this collapses two lines of test code into a single one.

In other cases I decided to keep the expectation string on its own line to avoid over-long lines, but it still simplifies the code I believe.

I mostly used some crazy regex patterns to create this PR.

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -447,8 +447,7 @@ class other(RunnerCore):
     self.assertContained('export default Module;', src)
 
   def test_esm_requires_modularize(self):
-    err = self.expect_fail([EMCC, test_file('hello_world.c'), '-sEXPORT_ES6', '-sMODULARIZE=0'])
-    self.assertContained('EXPORT_ES6 requires MODULARIZE to be set', err)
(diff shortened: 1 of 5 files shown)


test/test_other.py
```diff
@@ -3761,9 +3742,8 @@ def test_emit_tsd_sync_compilation(self):
     shared.check_call(cmd)
 
   def test_emit_tsd_wasm_only(self):
-    err = self.expect_fail([EMCC, test_file('other/test_emit_tsd.c'),
-                            '--emit-tsd', 'test_emit_tsd_wasm_only.d.ts', '-o', 'out.wasm'])
-    self.assertContained('Wasm only output is not compatible --emit-tsd', err)
+    expected = 'Wasm only output is not compatible --emit-tsd'
```
kripken, line 3745:
```suggestion
    expected = 'Wasm only output is not compatible with --emit-tsd'
```
pre-existing typo, might be worth fixing separately

sbc100: I will do a followup

PR #25642, opened 2025-10-24
Add a check for process.versions in minimum_runtime_check.js

stephenduong1004:
I am getting the error `ERROR - [JSC_POSSIBLE_INEXISTENT_PROPERTY] Property versions never defined on process`;
adding a check if `process.versions` exists.


sbc100:
Looks like closure compiler error, but we are not seeing in our tests.  Can you share repro steps?   Presumably any prorgram that is built that that includes this code would fail with `--closure=1`?   But we are not seeing that on our end.

It also looks like of the accesses are guarded.

PR #25643, opened 2025-10-24
Remove redundant null chaining. NFC

sbc100:
This null chaining operator is not only redundant (because it happens only after a typeof check), but simply doesn't work.

One cannot do `foo?.bar` if `foo` does not exist.  One would need to instead do `globalThis?.foo?.bar`.

### DIFF
--- a/src/minimum_runtime_check.js
+++ b/src/minimum_runtime_check.js
@@ -20,7 +20,9 @@
 
   var TARGET_NOT_SUPPORTED = {{{ TARGET_NOT_SUPPORTED }}};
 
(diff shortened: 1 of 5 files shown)


sbc100:
Speaking of which, I assume there is a reason we use `typeof` here in this file instead of simply using an optional chain starting with `globalThis`?

juj:
> Speaking of which, I assume there is a reason we use `typeof` here in this file instead of simply using an optional chain starting with `globalThis`

This was one of the review comments: https://github.com/emscripten-core/emscripten/pull/25414#discussion_r2388566492 . The idea here was that even if this code is run on a very old browser, it would still work.

Though now I realize that I was using `?.` thinking that "it'll get babeled away", but naturally that won't be the case.

So if something, it would probably be good to remove uses of ES6 altogether in this code.

PR #25644, opened 2025-10-24
Improve Firefox browser process tracking

juj:
Improve Firefox browser process tracking by using the -wait-for-browser cmdline arg, and waiting longer for process startup.

Unfortunately `-wait-for-browser` is not enough to get rid of the process delta polling, since the process handle returned by `subprocess.Popen()` cannot be used to cascade the spawned Firefox windows.

Waiting for 2 seconds for Firefox startup was not enough, so make the wait time more conservative.

Add the detected processes to the list rather than replace, so that all process handles are retained.

It may be possible to remove the use of `launch_browser_harness_with_proc_snapshot_workaround()` for headless Firefox runs, but more testing will be needed to verify that.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -30,7 +30,6 @@ from pathlib import Path
 from subprocess import PIPE, STDOUT
 from typing import Dict, Tuple
 from urllib.parse import parse_qs, unquote, unquote_plus, urlparse
-from .utils import WINDOWS
 
 import clang_native
 import jsrun


sbc100:
So what is `-wait-for-browser` supposed to actually do then?

test/common.py
```diff
@@ -2081,13 +2081,13 @@ def launch_browser_harness_with_proc_snapshot_workaround(cls, parallel_harness,
...
       if parallel_harness or is_safari():
-        time.sleep(min(2 + count * 0.3, 10))
+        time.sleep(min(5 + count * 0.3, 10))
         procs_after = list_processes_by_name(config.executable_name)
 
         # Take a snapshot again to find which processes exist after launching
         # the browser. Then the newly launched browser processes are determined
         # by the delta before->after.
-        cls.browser_procs = list(set(procs_after).difference(set(procs_before)))
+        cls.browser_procs += list(set(procs_after).difference(set(procs_before)))
```
sbc100, line 2090: Shouldn't `cls.browser_procs` be empty before starting a new browser?

Maybe add an assert that its empty to the top of this method?

juj:
> So what is `-wait-for-browser` supposed to actually do then?

It keeps the spawned process alive until the browser closes.

```py
import subprocess, sys, time
proc = subprocess.Popen(['C:\\Program Files\\Mozilla Firefox\\firefox.exe', '-new-instance', '-wait-for-browser', 'https://wiki.mozilla.org/Firefox/CommandLineOptions'])

while True:
  if proc.poll() is not None:
    print('Browser process quit.')
    sys.exit(0)
  time.sleep(1)
```

With `-wait-for-browser`, the above test program prints `'Browser process quit.'` when user closes the browser from the X button.

Without `-wait-for-browser`, the above test program immediately prints `'Browser process quit.'` even when the browser window is still open.

test/common.py
juj: It is not empty, but it is initialized 11 lines above on line 2079 to get the return of `subprocess.Popen()` initially.

sbc100:
I see.   But it sounds like `-wait-for-browser` doesn't actually do what we need in that its still not enough just to kill the returned PID and wait for it to exit?  i.e. its still possible, even with this flag, to have orphans firefox processes?

Is that a bug in `-wait-for-browser`?

Given that we still cannot just wait on the single PID, is there any advantage to using `-wait-for-browser`?

sbc100:
I think i'm not understanding the "since the process handle returned by subprocess.Popen() cannot be used to cascade the spawned Firefox windows." part

juj:
> But it sounds like `-wait-for-browser` doesn't actually do what we need in that its still not enough just to kill the returned PID and wait for it to exit?

It does. Seemingly so at least, in the small python test case that I posted above. Although I see CircleCI is giving some error, so I am running this on my CI also to test.

> i.e. its still possible, even with this flag, to have orphans firefox processes?

I am hoping that would not be the case. That is what I referred with the comment

"It may be possible to remove the use of `launch_browser_harness_with_proc_snapshot_workaround()` for headless Firefox runs, but more testing will be needed to verify that."

> "since the process handle returned by subprocess.Popen() cannot be used to cascade the spawned Firefox windows." part

The Emscripten test harness needs the process handles for two purposes:
1. to be able to shut down the browser, and
2. to be able to re-size and re-position the browser windows so they are cascaded on the desktop. Each browser has unique visible position on the desktop:

image

The comment "since the process handle returned by subprocess.Popen() cannot be used to cascade the spawned Firefox windows." means that even with `-wait-for-browser`, while we are able to solve 1., it is not able to solve 2.

sbc100:
Ok great.  lgtm either way.

I guess we only need to keep the handle to the root processes then maybe we can do the cascading thing logically here in this function and avoid storing the extended process list in cls.browser_procs?   i.e. for the purposes of kill in the browser we only need a single process now?

sbc100 (approved):

test/common.py
```diff
@@ -137,7 +137,7 @@ def configure(data_dir):
 
 class FirefoxConfig:
   data_dir_flag = '-profile '
-  default_flags = ('-new-instance',)
+  default_flags = ('-new-instance', '-wait-for-browser') if WINDOWS else ('-new-instance',)
```
sbc100, line 140: Probably warrants a comment here.

sbc100 (approved):

test/common.py
juj: Yeah, I'll add one. I am still testing to understand why this is necessary.

PR #25655, opened 2025-10-26
[libwebaudio.js] Simplify debug checks and logging. NFC

sbc100:
Split out from #25649

### DIFF
--- a/src/lib/libwebaudio.js
+++ b/src/lib/libwebaudio.js
@@ -23,10 +23,48 @@ var LibraryWebAudio = {
 #endif
     EmAudio[++EmAudioCounter] = object;
 #if WEBAUDIO_DEBUG
-    console.log(`Registered new WebAudio object ${object} with ID ${EmAudioCounter}`);
+    dbg(`Registered new WebAudio object ${object} with ID ${EmAudioCounter}`);
 #endif
     return EmAudioCounter;
   },
+
+#if ASSERTIONS || WEBAUDIO_DEBUG
+  $_emAudioExpectHandle__internal: true,
+  $_emAudioExpectHandle: (handle, methodName) => {
+#if WEBAUDIO_DEBUG
+    dbg(`called ${methodName}() with ID ${handle}`);
+#endif
+#if ASSERTIONS
+    var obj = EmAudio[handle];
+    assert(obj, `Called ${methodName}() on a nonexisting handle ${handle}`);
+    return obj;
+#endif
+  },
+
+  $emAudioExpectContext__internal: true,
+  $emAudioExpectContext: (handle, methodName) => {
+    var obj = _emAudioExpectHandle(handle, methodName);
+#if ASSERTIONS
+    assert(obj instanceof (window.AudioContext || window.webkitAudioContext), `${methodName}() called with ${handle} that is not an AudioContext, but of type ${typeof obj}`);
+#endif
+  },
+
+  $emAudioExpectNode__internal: true,
+  $emAudioExpectNode: (handle, methodName) => {
+    var obj = _emAudioExpectHandle(handle, methodName);
+#if ASSERTIONS
+    assert(obj instanceof window.AudioNode, `${methodName}() called with a handle ${handle} that is not an AudioNode, but of type ${typeof obj}`);
+#endif
+  },
+
+  $emAudioExpectNodeOrContext_internal: true,
+  $emAudioExpectNodeOrContext: (handle, methodName) => {
+    var obj = _emAudioExpectHandle(handle, methodName);
+#if ASSERTIONS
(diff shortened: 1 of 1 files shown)


src/lib/libwebaudio.js
```diff
@@ -338,32 +340,28 @@ var LibraryWebAudio = {
...
+#if ASSERTIONS || WEBAUDIO_DEBUG
+    emAudioCheckHandle(contextHandle, 'emscripten_audio_context_sample_rate');
 #endif
     return EmAudio[contextHandle]['sampleRate'];
   },
 
   emscripten_audio_node_connect: (source, destination, outputIndex, inputIndex) => {
+#if ASSERTIONS || WEBAUDIO_DEBUG
+    emAudioCheckHandle(source, 'emscripten_audio_node_connect', 1);
+    emAudioCheckHandle(destination, 'emscripten_audio_node_connect', 2);
```
juj: The choice to use `dbg()` instead of `console.log()` is fine, though I find this kind of `emAudioCheckHandle()` function that is passed magic values, sometimes `true`, sometimes `1`, sometimes `2`, and the default is `false`, to be quite unreadable.

It forces the reader and maintainer to jump between the implementation and call site to verify the preconditions. And when the preconditions are tested somewhere else, it risks there being a disconnect, e.g. in a refactor, and something falling through.

And if a new precondition comes up, then it requires adding new integers to do other things. This kind of code resembles a structure of

```js
function doSomething(what) {
  switch(what) {
  case 1: /*this*/; break;
  case 2: /*that*/; break;
  default: /*something else*/; break;
  }
}
```

Maybe unpack to separate functions `assertIsAudioNode(srcNode);` and `assertIsAudioNodeOrContext(dstNode);` that would have a single fixed function that is more straightforward to infer from the function name?

sbc100: OK, that does make sense yes.

src/lib/libwebaudio.js
```diff
@@ -23,11 +23,49 @@ var LibraryWebAudio = {
...
+  $_emAudioCheckHandle__internal: true,
+  $_emAudioCheckHandle: (handle, methodName) => {
+#if WEBAUDIO_DEBUG
+    dbg(`called ${methodName}() with ID ${handle}`);
+#endif
+#if ASSERTIONS
+    var obj = EmAudio[handle];
+    assert(obj, `Called ${methodName}() on a nonexisting handle ${handle}`);
+    return obj;
+#endif
```
juj, line 41: The `return obj;` was probably intended outside the ASSERTIONS, or otherwise the function returns nothing?

Or hmm, I suppose maybe the callers expect that to be the case.

src/lib/libwebaudio.js
```diff
@@ -23,11 +23,49 @@ var LibraryWebAudio = {
...
+  },
+
+  $emAudioCheckContext__internal: true,
+  $emAudioCheckContext: (handle, methodName) => {
+    var obj = _emAudioCheckHandle(handle, methodName);
+#if ASSERTIONS
```
juj, line 47: Would the above `var obj = ...;` expression go under `#if ASSERTIONS` as well, if the function returns `undefined` in non-ASSERTIONS mode?

src/lib/libwebaudio.js
```diff
@@ -250,10 +273,8 @@ var LibraryWebAudio = {
   },
 
   emscripten_create_wasm_audio_worklet_processor_async: (contextHandle, options, callback, userData) => {
-#if ASSERTIONS
-    assert(contextHandle, `Called emscripten_create_wasm_audio_worklet_processor_async() with a null Web Audio Context handle!`);
-    assert(EmAudio[contextHandle], `Called emscripten_create_wasm_audio_worklet_processor_async() with a nonexisting/already freed Web Audio Context handle ${contextHandle}!`);
-    assert(EmAudio[contextHandle] instanceof (window.AudioContext || window.webkitAudioContext), `Called emscripten_create_wasm_audio_worklet_processor_async() on a context handle ${contextHandle} that is not an AudioContext, but of type ${typeof EmAudio[contextHandle]}`);
+#if ASSERTIONS || WEBAUDIO_DEBUG
+    emAudioCheckContext(contextHandle, 'emscripten_create_wasm_audio_worklet_processor_async');
```
juj: The function name `emAudioCheckContext()` is a bit vague as to what it is checking? Maybe `emAudioAssertContext` or `emAudioAssertIsContext`?

src/lib/libwebaudio.js
sbc100: It seem harmless enough since non of this code is included without `-sWEBAUDIO_DEBUG` or `-sASSERTIONS`.   Nobody is going to be shipping release builds with `-sWEBAUDIO_DEBUG`.

src/lib/libwebaudio.js
sbc100: I renamed them to `emAudioExpectXXX` since they perform not just assertions but also logging

juj: That sounds good.

juj (approved):

Issue #19469, opened 2023-05-30
EMSCRIPTEN_VERSION macro in JS libraries


juj:
When working on [wasm_webgpu bindings library](https://github.com/juj/wasm_webgpu), I realized that after #19097 which landed for Emscripten 3.1.35, I needed to add these directives

https://github.com/juj/wasm_webgpu/commit/a1cc82dc445512f3acc2bf044a699531b8e4d944

which broke the JS library from working on Emscripten < 3.1.35, so in order to make the JS library work for different versions, I had to a backwards compatibility item to the library, that looks like this:

https://github.com/juj/wasm_webgpu/commit/5460a0d8c71343b5a1984da11f46a606cec3c9e3

Maintaining that kind of backwards compatibility item is brittle and difficult, since if the upstream code changes in the future, the change won't appear for users who use the wasm_webgpu library.

So instead, what I would like to do is write something like

```
#if EMSCRIPTEN_VERSION < 3.1.35
wgpu_object_get_label__deps: ['$stringToUTF8'],
#endif
```
to make the needed deps appear only for old Emscripten versions, or simply
```
#if EMSCRIPTEN_VERSION < 3.1.35
#error "lib_webgpu.js requires Emscripten 3.1.35 or newer"
#endif
```
to require developers to update to a newer version.

However looks like we don't have a construct that would allow implementing a `EMSCRIPTEN_VERSION < 3.1.35` yet - seems like we should add one.

Any preferences on how the syntax would like to look?

Use the same kind of compacted version number like the MIN_xxx_VERSION fields?
```
#if EMSCRIPTEN_VERSION < 30135
```
or use the same names as C/C++ code gets as preprocessor:
```
#if __EMSCRIPTEN_major__*10000 + __EMSCRIPTEN_minor__*100 + __EMSCRIPTEN_tiny__ < 30135
```
or use function syntax?
```
#if EMSCRIPTEN_VERSION_LESS_THAN(3,1,35)
#if EMSCRIPTEN_VERSION_LESS_OR_EQUAL(3,1,35)
#if EMSCRIPTEN_VERSION_GREATER_THAN(3,1,35)
#if EMSCRIPTEN_VERSION_GREATER_OR_EQUAL(3,1,35)
```
or something else?

(I think I'd prefer the very first one)

sbc100:
It looks like `EMSCRIPTEN_VERSION` is already available to JS libraries, but in string form: 

[...]

PR #25661, opened 2025-10-27
Remove unused internal setting: EMSCRIPTEN_VERSION. NFC

sbc100:
I did have to update the test for `emscripten_get_compiler_setting` which was using this internal settings
to test for compiler settings that return strings.  I changed to a different (non-inernal) setting: `CLOSURE_WARNINGS`.

I think we should probably followup by making `emscripten_get_compiler_setting` *not* report internal setting since the
whole point of internal settings is that we should be free to remove/change them at any time without breaking any
user code.  See #25663


juj:
Hmm, does(n't) removing this prevent this user code from working?

https://github.com/juj/wasm_webgpu/blob/54c51d715d68f22a62e1f84eb072737e8a8270c0/lib/lib_webgpu.js#L1-L4

If not, then LGTM. Though I suppose that's what this was for.

(It would be nice to get a better construct for downstream users to #ifdef check the version. I think we had an open bug about that somewhere)

sbc100:
I see.  Its kind of shame that internal settings like this are exposed to user libraries.   I was kind of hoping that internal settings we internal in the sense that we were free to add/remove/change them with impunity. 

But I guess it turns out that they are visible in this way to use JS libraries.

sbc100:
I guess we should add test to `test_jslib.py` and document that this setting is used externally like this.

juj:
> Its kind of shame that internal settings like this are exposed to user libraries. I was kind of hoping that internal settings we internal in the sense that we were free to add/remove/change them with impunity.

I agree it would be good for settings_internal.js to be internal.

FWIW, back in 2023 it seems like you were agreeing with manually parsing EMSCRIPTEN_VERSION as a good solution, https://github.com/emscripten-core/emscripten/issues/19469#issuecomment-1568765836 to avoid depending on changes to Emscripten.

Not sure if EMSCRIPTEN_VERSION was back then located in settings.js though.

kripken:
That setting was intentionally added for this purpose, to be used as in that test.

Someone asked for a way to  get the emscripten version at runtime, and we had other requests for such things, so a general "get compiler setting" option seemed simplest, plus adding that option with the version in it.

sbc100:
> That setting was intentionally added for this purpose, to be used as in that test.
> 
> Someone asked for a way to get the emscripten version at runtime, and we had other requests for such things, so a general "get compiler setting" option seemed simplest, plus adding that option with the version in it.

But the runtime version of emscripten and build time version are always the same right?   By definition?   At the build time version is available much more simply via `__EMSCRIPTEN_XXX__` version macros.

I did a global github search and could only find one use of this.. which could easily be replaced by a build time check.

kripken:
Oh, are you saying there is an easy way to capture the build-time value and make it available at runtime? Yeah, I guess it could be saved as a string with a C preprocessor `#`?

kripken:
Seems reasonable. In that case I'm ok to remove this, but perhaps we could add to the changelog or docs an example of the other way.

sbc100:
I decided to abandon this for now.   Since its needed for @juj's external JS library use case.

Issue #25663, opened 2025-10-27
RETAIN_COMPILER_SETTINGS should not retain internal settings


sbc100:
As it stands, in combination with `emscripten_get_compiler_setting` this makes
internal settings visible to user code, which seems problematic.

See https://github.com/emscripten-core/emscripten/pull/25661

PR #25664, opened 2025-10-27
Update docs for internal EMSCRIPTEN_VERSION setting. NFC

sbc100:
I proposed removing this internal setting in #25661 but it turns out it was useful for externally managed JS library files.

However, I don't think it should be tested as part of `emscripten_get_compiler_setting.c` since C/C++ can already get the compiler version using `VERSION` macros. . There should be no need for the runtime checks.

Indeed I hope we can stop exposing any internal settings to emscripten_get_compiler_setting: #25663

Also, add test for @juj's specific use case for `EMSCRIPTEN_VERSION` in a JS library file.


src/settings_internal.js
```diff
@@ -107,7 +107,8 @@ var WASI_MODULE_NAME = "wasi_snapshot_preview1";
 var JS_LIBRARIES = [];
 
 // This will contain the emscripten version. This can be useful in combination
-// with RETAIN_COMPILER_SETTINGS
+// external JS library files that need to check the version of emscripten they
```
juj: in combination **with**

PR #25665, opened 2025-10-27
[test] Add lto test configs to passing_core_test_modes. NFC

sbc100:
Without this, one cannot do things like `./test/runner lto0.test_foo*`.


juj:
Oh so this is why running `somesuite.test_foo*` often fails. The suite must be in the list of `passing_core_test_modes` for the wildcard to work?

sbc100:
> Oh so this is why running `somesuite.test_foo*` often fails. The suite must be in the list of `passing_core_test_modes` for the wildcard to work?

Yes... probably we should fix that .. but that seems to be how it works today.

PR #25666, opened 2025-10-27
Extract `emAudioWorkletPostMessage` helper. NFC

sbc100:
This avoids duplication and also means we can add consistent checks on the audio context that is passed in.

Its also a code size win.  We don't have a specific code size test that references these functions but I ran
`browser.test_audio_worklet_post_function_closure` before and after this change and got these results:

Before:

13297 out/test/test.js

After:

12951 out/test/test.js

### DIFF
--- a/src/lib/libwebaudio.js
+++ b/src/lib/libwebaudio.js
@@ -383,14 +383,28 @@ var LibraryWebAudio = {
     srcNode.connect(dstNode.destination || dstNode, outputIndex, inputIndex);
   },
 
-  emscripten_current_thread_is_audio_worklet: () => ENVIRONMENT_IS_AUDIO_WORKLET,
-
-  emscripten_audio_worklet_post_function_v: (audioContext, funcPtr) => {
-    (audioContext ? EmAudio[audioContext].audioWorklet['port'] : port).postMessage({'_wsc': funcPtr, args: [] }); // "WaSm Call"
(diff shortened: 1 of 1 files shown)


src/lib/libwebaudio.js
```diff
@@ -429,16 +441,15 @@ var LibraryWebAudio = {
...
   emscripten_audio_worklet_post_function_sig: (audioContext, funcPtr, sigPtr, varargs) => {
 #if ASSERTIONS
-    assert(audioContext >= 0);
     assert(funcPtr);
     assert(sigPtr);
     assert(UTF8ToString(sigPtr)[0] != 'v', 'Do NOT specify the return argument in the signature string for a call to emscripten_audio_worklet_post_function_sig(), just pass the function arguments.');
     assert(varargs);
 #endif
-    (audioContext ? EmAudio[audioContext].audioWorklet['port'] : port).postMessage({'_wsc': funcPtr, args: readEmAsmArgs(sigPtr, varargs) });
+    emAudioWorkletPostMessage(audioContext, {'_wsc': funcPtr, args: readEmAsmArgs(sigPtr, varargs)});
```
juj: Nice outlining.

If we don't have other uses, could also inject the `{'_wsc': funcPtr, args: ... }` part into the function?

sbc100: Nice!  That saves even more code size and allows for another centralized assertion.

PR #25667, opened 2025-10-27
Don't include internal compiler settings with RETAIN_COMPILER_SETTINGS

sbc100:
This is technically a breaking change so I mentioned it in the ChangeLog.

Fixes: #25663

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,12 @@ See docs/process.md for more on how version tagging works.
 
 4.0.19 (in development)
 -----------------------
+- The `RETAIN_COMPILER_SETTINGS` setting and the corresponding
+  `emscripten_get_compiler_setting` API no longer store or report internal
+  compiler settings (those listed in `setttings_internal.js`).  We made an
+  exception here for `EMSCRIPTEN_VERSION` which is the only internal setting
+  where we could find usage of `emscripten_get_compiler_setting` (in a global
+  GitHub search). (#25667)
 
 4.0.18 - 10/24/25
 -----------------
--- a/src/parseTools.mjs
+++ b/src/parseTools.mjs
@@ -806,17 +806,8 @@ function addAtPostRun(code) {
 
 function makeRetainedCompilerSettings() {
   const ret = {};
-  for (const [name, value] of Object.entries(global)) {
-    if (name[0] !== '_' && name == name.toUpperCase()) {
-      if (
-        typeof value == 'number' ||
-        typeof value == 'boolean' ||
-        typeof value == 'string' ||
-        Array.isArray(name)
-      ) {
-        ret[name] = value;
-      }
-    }
+  for (const name of PUBLIC_SETTINGS) {
+    ret[name] = globalThis[name];
   }
   return ret;
 }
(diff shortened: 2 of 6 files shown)


juj:
I'll defer to @kripken on this - I do not recall what the context was around adding this feature, and what the usage were.

Might be worth mentioning specific examples in the ChangeLog of which settings in particular will no longer be accessible, that people might have been referring to.

sbc100:
OK, will wait for @kripken to chime in

juj:
> that people might have been referring to.

The fact that the test code had `OPT_LEVEL` and `DEBUG_LEVEL` makes me ponder if those were some of the settings that the users were specifically interested in having. (It is a common motivation to add a test to specifically cover the feature that a user is interested in)

sbc100:
> > that people might have been referring to.
> 
> The fact that the test code had `OPT_LEVEL` and `DEBUG_LEVEL` makes me ponder if those were some of the settings that the users were specifically interested in having. (It is a common motivation to add a test to specifically cover the feature that a user is interested in)

I don't actually know if this feature has any users at all.

It was adding way back in 2014 but without much rational: 153f2d93520f37525f30a3808b969c5533387a3c

I can't find any users with google.  There are some users on github .. I'll investigate more.

sbc100:
It looks like there are 14 users on github: https://github.com/search?q=emscripten_get_compiler_setting%28%22++NOT+path%3Aemscripten.h++NOT+path%3Aemscripten_get_compiler_setting.c&type=code

The settings used are `EMSCRIPTEN_VERSION`, `PTHREAD_POOL_SIZE` and `LEGACY_GL_EMULATION`.  

Of these only `EMSCRIPTEN_VERSION` is an internal settings, so those 2 users would be affected.   Maybe we should try to patch them to use runtime detection instead before we land this.

sbc100:
There are only two uses of `EMSCRPITEN_VERSION` in the above search and they are both actually same usage: `save-editor/source/main.cpp`.   It also looks like the could both be replaced with `__EMSCRIPTEN_xxx__` version macros at compile time.

So I think we should go ahead with this change.   The other usages are using non-internal settings.

sbc100:
OK, I made a carve out for EMSCRIPTEN_VERSION here... so there should be minimal risk of breakages.

ChangeLog.md
```diff
@@ -20,6 +20,12 @@ See docs/process.md for more on how version tagging works.
 
 4.0.19 (in development)
 -----------------------
+- The `RETAIN_COMPILER_SETTINGS` settings and the corresponding
+  `emscripten_get_compiler_setting` API not longer store or report internal
```
```suggestion
  `emscripten_get_compiler_setting` API no longer store or report internal
```

ChangeLog.md
```diff
@@ -20,6 +20,12 @@ See docs/process.md for more on how version tagging works.
...
+  compiler settings (this listed in `setttings_internal.js`).  We made an
```
```suggestion
  compiler settings (those listed in `setttings_internal.js`).  We made an
```

ChangeLog.md
```diff
@@ -20,6 +20,12 @@ See docs/process.md for more on how version tagging works.
...
+  excption here for `EMSCRIPTEN_VERSION` which is the only internal setting
```
```suggestion
  exception here for `EMSCRIPTEN_VERSION` which is the only internal setting
```

kripken (approved):

PR #20802, opened 2023-11-29
GL_ENABLE_GET_PROC_ADDRESS

juj:
Some renderers that are ports from native codebases have a problem that they use the `*glGetProcAddress()` functions to dynamically discover the function entry points for different GL extensions.

In native GL world, there is a tacit guarantee that if an extension `GL_FOO` includes an entrypoint `glFoo()`, then if an implementation advertise support for `GL_FOO` extension, then calling `glGetProcAddress("glFoo")` can be expected to succeed.

More robust user code would check not only the presence of the extension, but also that the function `glGetProcAddress("glFoo")` does not return null, but such robustness is not necessarily built in to user applications since in the native GL world, this expectation doesn't generally fail.

Since some [6 years apo](https://github.com/emscripten-core/emscripten/commit/1facf3e687e7ede1c814f025683acfb6311f6f15#diff-aaacabb86431f19f9cc0196f449f8bc34764a5ac84556c21a7ce4934dddd9926R163), our best practices document has advocated users against using `*glGetProcAddress()` functions when targeting the web: https://emscripten.org/docs/optimizing/Optimizing-WebGL.html#optimizing-load-times-and-other-best-practices , due to the massive code size bloat that it causes (but also due to the performance problems that using it brings). Users are advised to not use that function.

Nevertheless, the function exists to ease porting existing codebases and to maintain parity between platforms.

One additional identified drawback is that this `*glGetProcAddress()` scheme enables GL code that is compiled to the web the possibility to activate "dormant" features from native GL side that were not enabled at the time the site is built; but later when a browser exposes a new WebGL extension, it can suddenly cause the site to try to activate those GL features.

This by itself wouldn't be a problem, except that **WebAssembly & WebGL do not support a facility to enable a true enumerating `*glGetProcAddress()` implementation**, but we must emulate that via a big if-else string search table. And so this statically implemented if-else chain cannot automatically manage new extension entry points as they come online when browsers evolve.

To fix above issues, this PR moves the `*glGetProcAddress()` functions behind a dedicated linker flag `-sGL_ENABLE_GET_PROC_ADDRESS=1` that users must now pass if they want to use a `*glGetProcAddress()` function. This enables our implementation to explicitly restrict to only advertising WebGL extensions that exist at the time that the site is built, so that the limited implementation of `*glGetProcAddress()` won't be able to break the promise that the existence of `GL_FOO` extension should imply that calling `glGetProcAddress("glFoo")` would succeed.

Fixes #20798

PR #25672, opened 2025-10-28
Move settings of GL_ENABLE_GET_PROC_ADDRESS to sdl2.py. NFC

sbc100:
This also allows it to be removed from COMPILE_TIME_SETTINGS.

Followup to #20802


sbc100:
l actually failed to get any sdl2 tests to fail with this set.. so maybe its not needed at all anymore.. I'm not sure.

PR #25674, opened 2025-10-28
Remove LINKABLE from COMPILE_TIME_SETTINGS. NFC

sbc100:
Maybe one day in the past it needed to be here becuase SIDE_MODULE and MAIN_MODULE are both compile-time settings (since they imply PIC).  But the automatic setting of LINKABLE, based on MAIN_MODULE now happens only in link.py.


kripken (approved):

sbc100:
Hm... we can't do this actually since `-sLINKABLE` is used to change the default `-fvisibility` at compile time.

PR #25687, opened 2025-10-29
Sort members of EmccOptions. NFC

sbc100:
Also, use class variables instead of instance fields. This doesn't
change the semantics (at least not the way we use this class) and it
saves a little bit of visual noise.

### DIFF
--- a/tools/cmdline.py
+++ b/tools/cmdline.py
@@ -11,6 +11,7 @@ import shlex
 import sys
 from enum import Enum, auto, unique
 from subprocess import PIPE
+from typing import List, Set
 
 from tools import (
   cache,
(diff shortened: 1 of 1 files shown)


tools/cmdline.py
```diff
@@ -60,59 +61,58 @@ class OFormat(Enum):
...
+  save_temps = False
+  shared = False
+  shell_path = None
+  source_map_base = ''
+  syntax_only = False
+  target = ''
+  use_closure_compiler = None
+  use_preload_cache = False
+  use_preload_plugins = False
+  valid_abspaths: List[str] = []
```
kripken, line 115: Looks like this also added types? They seem right.

sbc100: Yeas, for some reason mypy requires them to be typed when they are of this form.  Not sure why types were not needed before.

PR #25689, opened 2025-10-29
Use `ArrayBuffer.isView` over `instanceof ArrayBuffer`. NFC

sbc100:


sbc100:
Reduced scope to just one instance

PR #25694, opened 2025-10-31
Fix nested dynamic library loading via RPATH

hoodmane:
There is a bug with dynamic library loading. Suppose liba.so is on LD_LIBRARY_PATH and it has an RPATH of `$ORIGIN/other_dir` and loads `libb.so` from other_dir. Then suppose `libb.so` has an RPATH of `$ORIGIN` and wants to load `libc.so` also from `other_dir`.

Before this PR this doesn't work. The problem is that `flags.rpath.parentLibPath` is set to `libb.so` and we replace $ORIGIN with `PATH.dirname(parentLibName)` which is `.`. So unless `other_dir` is on the `LD_LIBRARY_PATH` or is the current working directory, loading will fail.

The fix: if `findLibraryFS()` returns a value that is not `undefined`, replace `libName` with the returned value.

### DIFF
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -1051,7 +1051,7 @@ var LibraryDylink = {
 #endif
     }
     var rpathResolved = (rpath?.paths || []).map((p) => replaceORIGIN(rpath?.parentLibPath, p));
-    return withStackSave(() => {
+    var result = withStackSave(() => {
       // In dylink.c we use: `char buf[2*NAME_MAX+2];` and NAME_MAX is 255.
       // So we use the same size here.
       var bufSize = 2*255 + 2;
@@ -1061,6 +1061,7 @@ var LibraryDylink = {
       var resLibNameC = __emscripten_find_dylib(buf, rpathC, libNameC, bufSize);
       return resLibNameC ? UTF8ToString(resLibNameC) : undefined;
     });
+    return FS.lookupPath(result).path;
   },
 #endif // FILESYSTEM
 
@@ -1168,6 +1169,7 @@ var LibraryDylink = {
       dbg(`checking filesystem: ${libName}: ${f ? 'found' : 'not found'}`);
 #endif
       if (f) {
+        libName = f;
         var libData = FS.readFile(f, {encoding: 'binary'});
         return flags.loadAsync ? Promise.resolve(libData) : libData;
       }
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -2437,6 +2437,67 @@ Module['postRun'] = () => {
     self.assertEqual(get_runtime_paths('libside1.so'), ['$ORIGIN'])
     self.assertEqual(get_runtime_paths('a.out.wasm'), ['$ORIGIN'])
 
+  def test_dylink_dependencies_rpath_nested(self):
+    create_file('pre.js', r'''
+      Module.preRun.push(() => {
+        Module.ENV.LD_LIBRARY_PATH = "/lib1";
+      });
+    ''')
+    create_file('side1.c', r'''
+    #include <stdio.h>
+
+    void side2();
+
+    void side1() {
+        printf("side1\n");
+        side2();
+    }
+    ''')
+    create_file('side2.c', r'''
+    #include <stdio.h>
+    void side3();
+
+    void side2() {
+        printf("side2\n");
+        side3();
+    }
+    ''')
+    create_file('side3.c', r'''
+    #include <stdio.h>
+
+    void side3() {
+        printf("side3\n");
+    }
+    ''')
+    create_file('main.c', r'''
+    #include <dlfcn.h>
+    #include <stdio.h>
+
+    typedef void (*F)(void);
+
+    int main() {
+        void* handle = dlopen("libside1.so", RTLD_NOW);
+        F side1 = (F)dlsym(handle, "side1");
+
+        printf("main\n");
+        side1();
+        return 0;
+    }
+    ''')
+    os.mkdir('libs')
+    os.mkdir('lib1')
+    self.emcc('side3.c', ['-fPIC', '-sSIDE_MODULE', '-olibs/libside3.so'])
+    self.emcc('side2.c', ['-fPIC', '-sSIDE_MODULE', '-olibs/libside2.so', '-Wl,-rpath,$ORIGIN', 'libs/libside3.so'])
+    self.emcc('side1.c', ['-fPIC', '-sSIDE_MODULE', '-Wl,-rpath,$ORIGIN/../libs', '-olib1/libside1.so', 'libs/libside2.so'])
+    settings = ['-sMAIN_MODULE=2', '-sDYLINK_DEBUG', "-sEXPORTED_FUNCTIONS=[_printf,_main]", "-sEXPORTED_RUNTIME_METHODS=ENV"]
+    preloads = []
+    for file in ['lib1/libside1.so', 'libs/libside2.so', 'libs/libside3.so']:
+      preloads += ['--preload-file', file]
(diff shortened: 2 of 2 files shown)


ryanking13:
I think this is the same thing I wanted to fix in https://github.com/emscripten-core/emscripten/pull/24234 and which is already in [Pyodide patches](https://github.com/pyodide/pyodide/blob/main/emsdk/patches/0003-dylink-Fix-rpath-calculation-in-nested-dependencies.patch)

sbc100:
Apologies for now getting around to reviewing #24234 for such a long time.

@hoodmane does that look like it solving the same issue?   Do you have a preference which approach is better?

test/test_other.py
```diff
@@ -2437,6 +2437,67 @@ def get_runtime_paths(path):
...
+    self.emcc('side2.c', ['-fPIC', '-sSIDE_MODULE', '-olibs/libside2.so', '-Wl,-rpath,$ORIGIN', 'libs/libside3.so'])
+    self.emcc('side1.c', ['-fPIC', '-sSIDE_MODULE', '-Wl,-rpath,$ORIGIN/../libs', '-olib1/libside1.so', 'libs/libside2.so'])
+    settings = ['-sMAIN_MODULE=2', '-sDYLINK_DEBUG', "-sEXPORTED_FUNCTIONS=[_printf,_main]", "-sEXPORTED_RUNTIME_METHODS=ENV"]
+    preloads = []
+    for file in ['lib1/libside1.so', 'libs/libside2.so', 'libs/libside3.so']:
+      preloads += ['--preload-file', file]
+    cmd = [EMCC, 'main.c', '-fPIC', '--pre-js', 'pre.js'] + settings + preloads
+    self.run_process(cmd)
+
+    self.run_js('a.out.js')
```
sbc100, line 2499: I think these 4 lines can be replaced with:

```
cflags = ['-fPIC', '--pre-js', 'pre.js'] + settings + preloads
self.do_runf('main.c', cflags=cflags)
```

Also maybe just inline the settings into `cflags` rather than a separate var.

test/test_other.py
```diff
@@ -2437,6 +2437,67 @@ def get_runtime_paths(path):
...
+    ''')
+    os.mkdir('libs')
+    os.mkdir('lib1')
+    self.emcc('side3.c', ['-fPIC', '-sSIDE_MODULE', '-olibs/libside3.so'])
+    self.emcc('side2.c', ['-fPIC', '-sSIDE_MODULE', '-olibs/libside2.so', '-Wl,-rpath,$ORIGIN', 'libs/libside3.so'])
+    self.emcc('side1.c', ['-fPIC', '-sSIDE_MODULE', '-Wl,-rpath,$ORIGIN/../libs', '-olib1/libside1.so', 'libs/libside2.so'])
+    settings = ['-sMAIN_MODULE=2', '-sDYLINK_DEBUG', "-sEXPORTED_FUNCTIONS=[_printf,_main]", "-sEXPORTED_RUNTIME_METHODS=ENV"]
+    preloads = []
+    for file in ['lib1/libside1.so', 'libs/libside2.so', 'libs/libside3.so']:
+      preloads += ['--preload-file', file]
```
sbc100, line 2495: Does this test depend on file preloading?  Or can you just use `-sNODERAWFS` instead?

test/test_other.py
```diff
@@ -2437,6 +2437,67 @@ def get_runtime_paths(path):
...
+        F side1 = (F)dlsym(handle, "side1");
+
+        printf("main\n");
+        side1();
+        return 0;
+    }
+    ''')
+    os.mkdir('libs')
+    os.mkdir('lib1')
+    self.emcc('side3.c', ['-fPIC', '-sSIDE_MODULE', '-olibs/libside3.so'])
```
sbc100, line 2489: You don't need `-fPIC` in addtion to `MAIN_MODULE` and/or `SIDE_MODULE`.

test/test_other.py
```diff
@@ -2437,6 +2437,67 @@ def get_runtime_paths(path):
     self.assertEqual(get_runtime_paths('libside1.so'), ['$ORIGIN'])
     self.assertEqual(get_runtime_paths('a.out.wasm'), ['$ORIGIN'])
 
+  def test_dylink_dependencies_rpath_nested(self):
+    create_file('pre.js', r'''
+      Module.preRun.push(() => {
+        Module.ENV.LD_LIBRARY_PATH = "/lib1";
```
sbc100, line 2443: I think you can just do `ENV.LD_LIBRARY_PATH = "/lib1";` and you don't need to actually export `ENV`, just add it to `DEFAULT_LIBRARY_FUNCS_TO_INCLUDE`

test/test_other.py
sbc100: Maybe add some expected output too?

test/test_other.py
```diff
@@ -2437,6 +2437,67 @@ def get_runtime_paths(path):
...
+    }
+    ''')
+    create_file('main.c', r'''
+    #include <dlfcn.h>
+    #include <stdio.h>
+
+    typedef void (*F)(void);
+
+    int main() {
+        void* handle = dlopen("libside1.so", RTLD_NOW);
```
sbc100, line 2479: assert(handle);

and then 

assert(side1)

src/lib/libdylink.js
```diff
@@ -1168,6 +1169,7 @@ var LibraryDylink = {
       dbg(`checking filesystem: ${libName}: ${f ? 'found' : 'not found'}`);
 #endif
       if (f) {
+        libName = f;
```
sbc100, line 1172: Add a comment here maybe?  "Replace libName with its full path.  Important for resolving other libs via $ORIGIN rpath"

src/lib/libdylink.js
```diff
@@ -1061,6 +1061,7 @@ var LibraryDylink = {
       var resLibNameC = __emscripten_find_dylib(buf, rpathC, libNameC, bufSize);
       return resLibNameC ? UTF8ToString(resLibNameC) : undefined;
     });
+    return FS.lookupPath(result).path;
```
sbc100, line 1064: Maybe either put this line inside the `withStackSave`, or move the final line out (i.e. just return resLibNameC), since that one doesn't need to be inside there either.

sbc100: Should we have `__emscripten_find_dylib` return an abspath instead?  Or is that more complex that this solution?

test/test_other.py
```diff
@@ -2437,6 +2437,67 @@ def get_runtime_paths(path):
...
+
+        printf("main\n");
+        side1();
+        return 0;
+    }
+    ''')
+    os.mkdir('libs')
+    os.mkdir('lib1')
+    self.emcc('side3.c', ['-fPIC', '-sSIDE_MODULE', '-olibs/libside3.so'])
+    self.emcc('side2.c', ['-fPIC', '-sSIDE_MODULE', '-olibs/libside2.so', '-Wl,-rpath,$ORIGIN', 'libs/libside3.so'])
```
sbc100, line 2490: The `self.emcc` take an `output_filename` named param (instead of `-o` arg).

test/test_other.py
sbc100, line 2479: Does this test depend on dlopen?  Or can the bug be reproduced using normal runtime dynamic linking too?

test/test_other.py
sbc100: Actually scratch that.. I'm gonna remove that argument :)  https://github.com/emscripten-core/emscripten/pull/25699

sbc100:
Does this also fix https://github.com/emscripten-core/emscripten/issues/25814?

sbc100:
I think we can probably close this now that #24234 has landed.   If #24234 doesn't solve your issue feel free to re-open

PR #25698, opened 2025-11-01
Prefer `for..of` over `forEach` with lambdas. NFC

sbc100:
I imagine theses usages of `forEach` predated the use of `for..of` in our codes and the `for..in` loops were more clumsy.

Using simple `for` loops here is generally faster at runtime and simpler to reason about (since `return` is not needed in place of `continue` and the binding of `this` is not hard to reason about, its just the same as the surrounding code).

Its also a codesize win.


juj (approved):
Very nice, highly approve.

Issue #25654, opened 2025-10-26
emrun --kill_exit 4.0.16 regression


neondev9:
Hello,

I just updated EmSDK from 4.0.13 to 4.0.18 and noticed that CI tests for WASM in my project failed on timeout.
I have Catch2 based unittests, which passed successfully.
CI failed on timeout.

I launch unittest using following cmdline:
`emrun --browser chrome --hostname=localhost --kill_exit --browser_args="--enable-experimental-webassembly-features --headless --remote-debugging-port=0 --disable-gpu --disable-software-rasterizer" <pathtotest.html>`

They fail when test code is finished on exit.
If I add --verbose for emrun, then it says:
```
Web page has quit with a call to exit() with return code 0. Shutting down web server. Pass --serve-after-exit to keep serving even after the page terminates with exit().
Web server loop done.
```

When I try to interrupt it using Ctrl+C, then:
```
  File "/Users/dev/Documents/Project/ThirdParty/EmSDK/Source/upstream/emscripten/emrun.py", line 1896, in <module>
    sys.exit(main(sys.argv[1:]))
             ~~~~^^^^^^^^^^^^^^
  File "/Users/dev/Documents/Project/ThirdParty/EmSDK/Source/upstream/emscripten/emrun.py", line 1888, in main
    returncode = run(args)
  File "/Users/dev/Documents/Project/ThirdParty/EmSDK/Source/upstream/emscripten/emrun.py", line 1861, in run
    httpd.server_close()
    ~~~~~~~~~~~~~~~~~~^^
  File "/Users/dev/Documents/Project/ThirdParty/EmSDK/Source/python/3.13.3_64bit/lib/python3.13/socketserver.py", line 715, in server_close
    self._threads.join()
    ~~~~~~~~~~~~~~~~~~^^
  File "/Users/dev/Documents/Project/ThirdParty/EmSDK/Source/python/3.13.3_64bit/lib/python3.13/socketserver.py", line 661, in join
    thread.join()
    ~~~~~~~~~~~^^
  File "/Users/dev/Documents/Project/ThirdParty/EmSDK/Source/python/3.13.3_64bit/lib/python3.13/threading.py", line 1092, in join
    self._handle.join(timeout)

```

Note: I compared behaviour between different EmSDK versions and this problem didn't exist on 4.0.15 and **4.0.16 introduced it**.
I have the same issue across different desktop platforms: Windows 11, Linux (Ubuntu 24), MacOS 26.0.1.

[...]

PR #25703, opened 2025-11-02
Fix for emrun --kill-exit

sbc100:
This was broken by #25359.  I though I was just moving this assignment to the correct place, but it seems like it broken emrun because the 1.1 protocol was not actually enabled before.  Once enabled it seems like this was causing `httpd.server_close` to hang forever.

Fixes: #25654

### DIFF
--- a/emrun.py
+++ b/emrun.py
@@ -578,8 +578,6 @@ class HTTPWebServer(socketserver.ThreadingMixIn, HTTPServer):
 
 # Processes HTTP request back to the browser.
 class HTTPHandler(SimpleHTTPRequestHandler):
-  protocol_version = 'HTTP/1.1'  # noqa: DC01
-
   def send_head(self):
     global page_last_served_time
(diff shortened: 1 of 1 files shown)


juj:
Is `protocol_version` a upstream `SimpleHTTPRequestHandler` variable, or something used just by us?

juj:
Would be good to have a test of the failing cmdline if possible. I think we had emrun tests in the other suite?

sbc100:
I think that our emrun tests are currently disabled..  we should certainly re-enable.  When we do I will try to confirm if there is already a test for this, and if not I will add one.

sbc100:
> Is `protocol_version` a upstream `SimpleHTTPRequestHandler` variable, or something used just by us?

Its upstream.   However I guess that point at which is it read means that out previous usage of it was not effectual   When I I fixed that, the 1.1 starting having and effect which then broke emrun due to the open connection.

PR #25709, opened 2025-11-03
Fixes references to exit_with_error helper function. NFC

sbc100:
This function was moved to utils, but continued to be referenced via shared.

### DIFF
--- a/embuilder.py
+++ b/embuilder.py
@@ -216,10 +216,10 @@ def main():
   args = parser.parse_args()
 
   if args.operation != 'rebuild' and len(args.targets) == 0:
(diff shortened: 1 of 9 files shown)


tools/utils.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Emscripten Authors.  All rights reserved.
+# Copyrigh 2020 The Emscripten Authors.  All rights reserved.
```
kripken:
```suggestion
# Copyright 2020 The Emscripten Authors.  All rights reserved.
```
Unintended I assume

PR #25717, opened 2025-11-04
Explicitly pass `--no-stack-first` to the linker. NFC

sbc100:
This is in preparation for the default change on the llvm side.

See https://github.com/llvm/llvm-project/issues/151015

### DIFF
--- a/tools/building.py
+++ b/tools/building.py
@@ -266,8 +266,12 @@ def lld_flags_for_executable(external_symbols):
         # function like we do in STANDALONE_WASM mode.
         cmd += ['--no-entry']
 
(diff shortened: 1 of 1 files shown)


tools/building.py
```diff
@@ -263,8 +263,12 @@ def lld_flags_for_executable(external_symbols):
         # function like we do in STANDALONE_WASM mode.
         cmd += ['--no-entry']
 
+  # The default for `--stack-first` is transitioning from disabled to
+  # enabled.  So be explicitly in all cases for now.
```
kripken:
```suggestion
  # enabled.  So be explicit in all cases for now.
```

kripken (approved):

Issue #25479, opened 2025-10-02
Bulk memory is enabled in -O3 even when asked not to.


juj:
If one runs

```
emcc test\hello_world.c -o a.js -O3 -mno-bulk-memory -mno-bulk-memory-opt
```

the resulting code has a bulk memory Data Count section. See https://github.com/emscripten-core/emscripten/pull/25466

sbc100:
It seems like we are lowering the actual instructions away `--llvm-memory-copy-fill-lowering` but binaryen is still including the data count section.

kripken:
Binaryen will only emit it when the feature is enabled, so perhaps we are setting it for some reason?

https://github.com/WebAssembly/binaryen/blob/7ba340a934d17c3d8fc23ef9cd9f99e4e3ac4367/src/wasm/wasm-binary.cpp#L647-L654

kripken:
(Also, `--llvm-memory-copy-fill-lowering` disables it, https://github.com/WebAssembly/binaryen/blob/7ba340a934d17c3d8fc23ef9cd9f99e4e3ac4367/src/passes/LLVMMemoryCopyFillLowering.cpp#L75 and https://github.com/WebAssembly/binaryen/blob/7ba340a934d17c3d8fc23ef9cd9f99e4e3ac4367/src/passes/LLVMMemoryCopyFillLowering.cpp#L114 )

juj:
> Binaryen will only emit it when the feature is enabled, so perhaps we are setting it for some reason?
> 
> https://github.com/WebAssembly/binaryen/blob/7ba340a934d17c3d8fc23ef9cd9f99e4e3ac4367/src/wasm/wasm-binary.cpp#L647-L654

From what I am seeing, Binaryen is scanning the .wasm file for what feature it has, and propagate the same features into the output. It is seeing the input .wasm file to have bulk-memory enabled, so it passes `--enable-bulk-memory` into the output:

```
C:\emsdk\emscripten\main>emcc test\hello_world.c -o a.js -O3 -mno-bulk-memory -mno-bulk-memory-opt

[...]

PR #25719, opened 2025-11-05
Fix bulk memory feature removal in optimizing builds

sbc100:
The bug here was the BINARYEN_FEATURES was been stashed right away after linking and then used unmodified for future binaryen calls.

Fixes: #25479

### DIFF
--- a/test/test_other.py
+++ b/test/test_other.py
@@ -9974,6 +9974,23 @@ int main() {
     compile(['-sMIN_SAFARI_VERSION=140100', '-mbulk-memory'])
     verify_features_sec_linked('nontrapping-fptoint', False)
 
+  def test_no_bulk_memory(self):
+    # The test_wasm_features test (above) uses the feature section to confirm
+    # if a feature is present, but that doesn't work in optimizing builds
+    # since we strip the feature section in release builds.
+    # This test confirms that no DATACOUNT section is present in the final
+    # binary.
+
+    def has_data_count(filename):
+      with webassembly.Module(filename) as wasm:
+        return wasm.get_section(webassembly.SecType.DATACOUNT)
+
+    self.emcc('hello_world.c', ['-O3', '-o', 'bulk.js'])
+    self.assertTrue(has_data_count('bulk.wasm'))
+
+    self.emcc('hello_world.c', ['-O3', '-o', 'nobulk.js', '-mno-bulk-memory', '-mno-bulk-memory-opt'])
+    self.assertFalse(has_data_count('nobulk.wasm'))
+
   @crossplatform
   def test_html_preprocess(self):
     src_file = test_file('module/test_stdin.c')
--- a/tools/link.py
+++ b/tools/link.py
@@ -34,6 +34,7 @@ from . import (
   webassembly,
 )
 from .cmdline import OFormat
+from .feature_matrix import Feature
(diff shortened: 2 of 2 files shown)


sbc100:
Depends on https://github.com/emscripten-core/emscripten/pull/25718

test/test_other.py
```diff
@@ -9974,6 +9974,21 @@ def compile(flags):
     compile(['-sMIN_SAFARI_VERSION=140100', '-mbulk-memory'])
     verify_features_sec_linked('nontrapping-fptoint', False)
 
+  def test_wasm_features_opt(self):
+    # The test_wasm_features test uses the feature section to confirm
+    # if a feuture is present, but that doesn't work in optimizing builds
+    # since binaayen will strip the feature section.
```
juj: `binaayen` -> `binaryen`

Maybe also `since` -> `then`?

test/test_other.py
```diff
@@ -9974,6 +9974,21 @@ def compile(flags):
     compile(['-sMIN_SAFARI_VERSION=140100', '-mbulk-memory'])
     verify_features_sec_linked('nontrapping-fptoint', False)
 
+  def test_wasm_features_opt(self):
```
juj: Maybe the test would be better named more concretely, e.g. `test_no_data_count_without_bulkmem` or similar.

Now it reads like it is a general test that would test all possible ways that binaryen would remove an unsupported feature? (but if there are other ways, those would probably be better to be separate tests, to avoid "kitchen sink tests")

sbc100:
Rebased

test/test_other.py
```diff
@@ -9974,6 +9974,23 @@ def compile(flags):
     compile(['-sMIN_SAFARI_VERSION=140100', '-mbulk-memory'])
     verify_features_sec_linked('nontrapping-fptoint', False)
 
+  def test_no_bulk_memory(self):
+    # The test_wasm_features test (above) uses the feature section to confirm
+    # if a feuture is present, but that doesn't work in optimizing builds
```
juj: `feuture` -> `feature`

test/test_other.py
```diff
@@ -9974,6 +9974,23 @@ def compile(flags):
...
+    # binary.
+
+    def has_data_count(filename):
+      with webassembly.Module(filename) as wasm:
+        return wasm.get_section(webassembly.SecType.DATACOUNT)
+
+    self.emcc('hello_world.c', ['-O3', '-o', 'bulk.js'])
+    self.assertTrue(has_data_count('bulk.wasm'))
+
+    self.emcc('hello_world.c', ['-O3', '-o', 'nobulk.js', '-mno-bulk-memory', '-mno-bulk-memory-opt'])
```
juj, line 9991: I wonder if it might make sense to test the `'-mno-bulk-memory'` and `'-mno-bulk-memory-opt'` separately?

IIUC `'-mno-bulk-memory'` implies `'-mno-bulk-memory-opt'`?

So `self.emcc('hello_world.c', ['-O3', '-o', 'nobulk.js', '-mno-bulk-memory']` should not get bulk memory either?

sbc100: > IIUC `'-mno-bulk-memory'` implies `'-mno-bulk-memory-opt'`?

From the LLVM POV, I'm not sure it does sadly.

PR #25720, opened 2025-11-05
Move BINARYEN_PASSES out of settings_internal.js. NFC

sbc100:
I'm making an effort to remove stuff from `settings_internal.js` unless its specifically needed by the JS compiler.

We might want to invent a new place for such python global state but we I think we can add this organically, over time.

See https://github.com/emscripten-core/emscripten/issues/25721

### DIFF
--- a/tools/building.py
+++ b/tools/building.py
@@ -13,7 +13,7 @@ import shutil
 import subprocess
 import sys
 from subprocess import PIPE
-from typing import Dict, Set
+from typing import Dict, List, Set
 
 from . import (
   cache,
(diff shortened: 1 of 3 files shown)


tools/building.py
```diff
@@ -60,6 +60,10 @@
 _is_ar_cache: Dict[str, bool] = {}
 # the exports the user requested
 user_requested_exports: Set[str] = set()
+# A list of feature flags to pass to each binaryen invocation (like wasm-opt,
+# etc.). This is received by the first call binaryen (e.g. `wasm-emscripten-finalize`)
```
kripken:
```suggestion
# etc.). This is received by the first binaryen call (e.g. `wasm-emscripten-finalize`)
```

kripken (approved):

PR #25727, opened 2025-11-05
Use utils.py directly rather than imported via shared.py. NFC

sbc100:


tools/ports/__init__.py
```diff
@@ -275,7 +275,7 @@ def build_port(src_dir, output_path, port_name, includes=[], flags=[], cxxflags=
   @staticmethod
   def get_dir(*parts):
     dirname = os.path.join(config.PORTS, *parts)
-    shared.safe_ensure_dirs(dirname)
+    utils.safe_ensure_dirs(dirname)
```
kripken, line 278: Looks like `safe_ensure_dirs` is changed as well, not just `path_from_root` and `WINDOWS`?

sbc100: Updated the title

PR #25704, opened 2025-11-03
Improve Binary Encoding testing+docs. NFC

juj:
1. Add test for all byte pairs on binary encoding, and
2. Improve binary encoding settings docs to mention that UTF-8 encoding is now needed, and
3. Add a note to ChangeLog to highlight this requirement.

Addresses https://groups.google.com/g/emscripten-discuss/c/E_HmYqXGjN8

PR #25728, opened 2025-11-05
[test] Split browser-specific code into browser_common.py. NFC

sbc100:
common.py was getting pretty huge and about 900 lines of it was browser-test specific stuff.


juj:
Looks good, though please after https://github.com/emscripten-core/emscripten/pull/25704 lands?

PR #25737, opened 2025-11-06
Add `--buffer` option to test runner. NFC

sbc100:
This is standard option in the default python unintest runner.

It now works for both the normal serial runner as well as the parallel runner.    In the case of the parallel runner we not inherit from the standard `unittest.TestResult` so that we can inherit the the stdout/stderr buffering methods it implements.

In the future we may want to enable this by default.  

This could especially useful the parallel runner where we have issue of interleaving the stdout and stderr of different tests is confusing.

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -43,7 +43,7 @@ def cap_max_workers_in_pool(max_workers, is_browser):
   return max_workers
 
 
-def run_test(test, allowed_failures_counter, lock, progress_counter, num_tests):
+def run_test(test, allowed_failures_counter, lock, progress_counter, num_tests, buffer):
   # If we have exceeded the number of allowed failures during the test run,
   # abort executing further tests immediately.
   if allowed_failures_counter and allowed_failures_counter.value < 0:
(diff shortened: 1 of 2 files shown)


test/parallel_testsuite.py
```diff
@@ -295,13 +298,17 @@ def log_test_run_for_visualization(self):
...
     self.test_name = str(test)
     self.start_time = time.perf_counter()
 
   def stopTest(self, test):
+    super().stopTest(test)
     # TODO(sbc): figure out a way to display this duration information again when
     # these results get passed back to the TextTestRunner/TextTestResult.
     self.buffered_result.duration = self.test_duration
+    del self._original_stdout
+    del self._original_stderr
```
juj, line 339: Might warrant a comment as to what this does and why we delete these?

test/parallel_testsuite.py
```diff
@@ -312,42 +319,54 @@ def compute_progress(self):
...
     self.buffered_result = BufferedTestSkip(test, reason)
     self.test_result = 'skipped'
 
   def addFailure(self, test, err):
-    msg = f'{test} ... FAIL'
-    errlog(f'{self.compute_progress()}{with_color(RED, msg)}')
     self.buffered_result = BufferedTestFailure(test, err)
     self.test_result = 'failed'
     self.failures += [test]
+    self._mirrorOutput = True
```
juj: What is this about?

test/parallel_testsuite.py
```diff
@@ -312,42 +319,54 @@ def compute_progress(self):
...
-    errlog(f'{self.compute_progress()}{with_color(RED, msg)}')
     self.buffered_result = BufferedTestError(test, err)
     self.test_result = 'errored'
-    self.errors += [test]
+    self._mirrorOutput = True
+
+  def printResult(self):
+    if self.test_result == 'success':
+      msg = f'ok ({self.calculateElapsed():.2f}s)'
+      errlog(f'{self.compute_progress()}{self.test} ... {with_color(GREEN, msg)}')
```
juj: In `unittest` docs https://docs.python.org/3/library/unittest.html

```
-b, --buffer

    The standard output and standard error streams are buffered during the test run. Output during a passing test is discarded. Output is echoed normally on test fail or error and is added to the failure messages.
```

so if the goal is to be faithful to `--buffer`, the above line should not print anything if `--buffer` is active?

juj:
> This could b especially useful the parallel runner where we have issue of interleaving of different test output.

Do you have more details on this? Afaik printing the test results is intentionally done as one print to ensure it interleaves correctly? I have not seen interleaving issues on any of Win/Linux/macOS. (except for one ANSI color coding related issue on Windows, where occassionally the ANSI color code does not register correctly.. but that should be a color issue)

juj:
Looks good, though can you elaborate what the difference here is? I am not sure I follow.

The implementation here seems to refactor the test result printing out to a common function, but what else happens here?

Python's docs of --buffer are very odd to me:

```
The standard output and standard error streams are buffered during the test run.
```
This description seems to conflict with the meaning of `python -u` flag, and the `PYTHONUNBUFFERED=1` env. var. I.e. to my understanding stdout and stderr are already buffered by default in Python, and those two options can be used to unbuffer. Maybe there is some extra manual buffering here?

test/parallel_testsuite.py
sbc100: I took this test directly from `python3 -m unittest --help` which itself comes from `/usr/lib/python3.13/unittest/main.py` on my system:

```
           parser.add_argument('-b', '--buffer', dest='buffer',                    
                                action='store_true',                                
                                help='Buffer stdout and stderr during tests')    
```

sbc100:
> > This could b especially useful the parallel runner where we have issue of interleaving of different test output.
> 
> Do you have more details on this? Afaik printing the test results is intentionally done as one print to ensure it interleaves correctly? I have not seen interleaving issues on any of Win/Linux/macOS. (except for one ANSI color coding related issue on Windows, where occassionally the ANSI color code does not register correctly.. but that should be a color issue)

I'm talking about the stdout and stderr that is produced by the test itself, not the printing of the single result line.   That latter works fine, the former is interleaved.   Even with `--buffer` should could still have interleaving but since only failing tests output anything at all the chances are less.

As a followup we could send the stdout back the main thread for printing, which would solve all interleaving issues.

sbc100:
> Looks good, though can you elaborate what the difference here is? I am not sure I follow.
> 
> The implementation here seems to refactor the test result printing out to a common function, but what else happens here?

The hard work here is done by the adding the `TestResult` base class.   This has `_setupStdout` and `_restoreStdout` method internally which is used when the test starts and stops.

The reason I had to move the result printing was so that it happend only after `_restoreStdout`. 

I will add to the PR description here.

test/parallel_testsuite.py
sbc100: I removed this and replaced with a call to `super()` method.   Basically `_mirrorOutput` tells the result to print the buffered stdout and stderr once the test completes.. this is done only for failure cases (when buffering).

sbc100:
To be clear this change is not only about interleaving.. its about hiding all the of the stdout and stderr from non-failing tests.

So if we adopt this we no longer need to worry about leaving random `print` statements in our test code.. since you would only ever see them if/when the test fails.

sbc100:
Hopefully the comments and updated description here are enough, landing this now since I have a very nice followup ready

PR #25739, opened 2025-11-07
Remove a bunch of redundant code from parallel_testsuite.py

sbc100:
All this infrastructure with the BufferedTestBase was added to work around that fact that we were previously using using a fake TestResult class so that is could be serialized.

However in #25737 I converted to using TestResult as the superclass of BufferedParallelTestResult.  There were only two members that needed to me removed to make it work.   The `TestResult` class stores all exceptions as strings so there is also no need to the Fake exceptions stuff anymore.

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -263,7 +263,7 @@ class ParallelTestSuite(unittest.BaseTestSuite):
     # Extract only the test short names
     flaky_tests = [x.split('.')[-1] for x in flaky_tests]
 
-    # The next updateResult loop will print a *lot* of lines really fast. This
+    # The next integrateResult loop will print a *lot* of lines really fast. This
     # will cause a Python exception being thrown when attempting to print to
     # stderr, if stderr is in nonblocking mode, like it is on Buildbot CI:
     # See https://github.com/buildbot/buildbot/issues/8659
@@ -272,11 +272,9 @@ class ParallelTestSuite(unittest.BaseTestSuite):
       os.set_blocking(sys.stderr.fileno(), True)
 
     for r in results:
-      # Merge information of flaky tests into the test result
-      if r.test_result == 'success' and r.test_short_name() in flaky_tests:
-        r.test_result = 'warnings'
-      # And integrate the test result to the global test object
-      r.updateResult(result)
+      # Integrate the test result to the global test result object
+      r.integrateResult(result)
+      r.log_test_run_for_visualization(flaky_tests)
 
     # Generate the parallel test run visualization
     if EMTEST_VISUALIZE:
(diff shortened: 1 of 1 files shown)


sbc100:
PTAL, I think this is a really nice cleanup of unnecessary code.

test/parallel_testsuite.py
```diff
@@ -272,11 +272,9 @@ def combine_results(self, result, buffered_results):
       os.set_blocking(sys.stderr.fileno(), True)
 
     for r in results:
-      # Merge information of flaky tests into the test result
-      if r.test_result == 'success' and r.test_short_name() in flaky_tests:
-        r.test_result = 'warnings'
-      # And integrate the test result to the global test object
+      # Integrate the test result to the global test restul oject
```
juj: `result object`

test/parallel_testsuite.py
```diff
@@ -310,179 +304,89 @@ def test_short_name(self):
...
+    if self.test_result == 'success':
+      other.addSuccess(self.test)
+    elif self.test_result == 'failed':
+      other.addFailure(*self.failures[0])
+    elif self.test_result == 'errored':
+      other.addError(*self.errors[0])
+    elif self.test_result == 'skipped':
+      other.addSkip(*self.skipped[0])
+    elif self.test_result == 'unexpected success':
+      other.addUnexpectedSuccess(*self.unexpectedSuccessesed[0])
```
juj: `unexpectedSuccessesed` -> `unexpectedSuccesses`

test/parallel_testsuite.py
```diff
@@ -310,179 +304,89 @@ def test_short_name(self):
...
 
-  def updateResult(self, result):
-    result.startTest(self.test)
-    self.buffered_result.updateResult(result)
-    result.stopTest(self.test)
-    result.core_time += self.test_duration
-    self.log_test_run_for_visualization()
+  def updateResult(self, other):
+    """This method get called on the main thread once the buffered result
+    is recieved.  It add the buffered result to the overall result."""
```
juj: `recieved` -> `received`

test/parallel_testsuite.py
```diff
@@ -310,179 +304,89 @@ def test_short_name(self):
...
+    # The exception info objects that we are adding here have already
+    # been turned into strings so make _exc_info_to_string into a no-op.
+    other._exc_info_to_string = lambda x, _y: x
+    # No need to worry about stdou/stderr buffering since are a not
```
juj: `stdout`

test/parallel_testsuite.py
```diff
@@ -310,179 +304,89 @@ def test_short_name(self):
...
+    # actually runnin the test here, only setting the results.
```
juj: `running`

test/parallel_testsuite.py
juj: Here's a reference to "the buffered result" and "the overall result". It would be good to mention which of the `self` and `other` have which roles.

And maybe reflect with the variable name, like `def updateResult(self, overall_results):` ?

juj (approved):
Nice, looks like a great simplification.

test/parallel_testsuite.py
sbc100: Done.  I  also renamed this method to `integrateResult`.

PR #25740, opened 2025-11-07
[dylink] Fix addEmJs when parameter is indirect pointer

eagleoflqj:
I'm in the middle of porting [libskk](https://github.com/ueno/libskk) with glib, and encountered a `SyntaxError: Unexpected token '*'`.
Given the value of `cSig`, `jsArg` and `jsArgs` in debugger view, it seems straightforward.

### DIFF
--- a/.github/workflows/rebaseline-tests.yml
+++ b/.github/workflows/rebaseline-tests.yml
@@ -15,7 +15,7 @@ jobs:
     name: Rebaseline Tests
     runs-on: ubuntu-latest
     env:
-      GH_TOKEN: ${{ github.token }}
+      GH_TOKEN: ${{ secrets.EMSCRIPTEN_BOT_TOKEN }}
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
@@ -58,5 +58,5 @@ jobs:
             fi
           fi
           git push origin rebaseline_tests
-          gh pr create --fill --base ${{ github.ref_name }}
+          gh pr create --fill --base ${{ github.ref_name }} --reviewer sbc100,kripken
           gh pr merge --squash --auto
--- a/.github/workflows/update-website.yml
+++ b/.github/workflows/update-website.yml
@@ -1,0 +1,51 @@
+name: Update website
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ main ]
+
+jobs:
+  update-website:
+    name: Update website
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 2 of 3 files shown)


kripken (commented):
Thanks! This looks right.

I believe the test for it is `other.test_em_js_side_module`, runnable with
```
./test/runner other.test_em_js_side_module
```
Please add a `**` pointer there, to test this.

test/other/test_em_js_side.c
```diff
@@ -14,9 +14,14 @@ EM_JS(void, js_side_func_void, (), {
...
 
+EM_JS(void, js_side_func_indirect_ptr, (char **pptr), {
+  out(`indirect ptr value: ${pptr}`);
+});
+
 int test_side() {
   js_side_func(42, "hello");
   js_side_func2("hello again");
   js_side_func_void();
+  js_side_func_indirect_ptr(NULL);
```
sbc100: Can you pass `&"hello"` and then print the string js_side_func_indirect_ptr?

otherwise lgtm

kripken:
Also see the failing test, an expectation needs to be updated. See instructions for that in the CI error.

sbc100:
> Also see the failing test, an expectation needs to be updated. See instructions for that in the CI error.

I think just rebase should fix that

test/other/test_em_js_side.c
eagleoflqj: Code inside EM_JS is JavaScript function so I cannot dereference pointer in the C way. Is there a way to do it with JS?

sbc100: Yes you can use the makeGetValue macro.  See the libraries in src/lib for examples

eagleoflqj: I failed to make the macro work inside this file, but figured out it compiles to a library function, so opted to use that.

test/other/test_em_js_side.c
```diff
@@ -14,9 +14,15 @@ EM_JS(void, js_side_func_void, (), {
   out(`hello from void func`);
 });
 
+EM_JS(void, js_side_func_indirect_ptr, (const char **pptr), {
+  out(UTF8ToString(readI53FromU64(pptr)));
```
sbc100: I think you can do `getValue(pptr, '*')`

The you can add `EM_JS_DEPS(deps, "$getValue");` above.

eagleoflqj: Applied. It seems `getValue ` is already available like `UTF8ToString` so I don't need to explicitly import.

PR #25741, opened 2025-11-07
Add emscripten_queue_microtask() API

juj:
Add emscripten_queue_microtask() API to enable use case described in this thread: https://github.com/emscripten-core/emscripten/pull/25670#discussion_r2469827161.


sbc100:
I had a version of this API in https://github.com/emscripten-core/emscripten/pull/24481... but sure how it compares

test/test_browser.py
```diff
@@ -5016,6 +5016,13 @@ def test_emscripten_request_animation_frame_loop(self):
   def test_request_animation_frame(self):
     self.btest_exit('test_request_animation_frame.c')
 
+  def test_emscripten_queue_microtask(self):
+    self.btest_exit('emscripten_queue_microtask.c')
+
+  @requires_shared_array_buffer
+  def test_emscripten_queue_microtask_pthread(self):
```
sbc100: Other tests in this file do it this way:

```
  @parameterized({                                                                                 
    '': ([],),                                                                                     
    'proxy_to_pthread': (['-pthread', '-sPROXY_TO_PTHREAD'],),                                     
  })  
  ```

File: src/lib/libhtml5.js [hunk 2]
```diff
@@ -2374,6 +2374,12 @@ var LibraryHTML5 = {
     return requestAnimationFrame(tick);
   },
 
+  emscripten_queue_microtask: (cb, userData) => {
+    queueMicrotask(() => {
+      {{{ makeDynCall('vp', 'cb') }}}(userData);
```
sbc100: This probably needs to be wrapped in `callUserCallback`.

sbc100:
> I had a version of this API in #24481... but sure how it compares

Maybe just copy the version I had in that PR?  Should probably live in libeventloop.js too.

I was recently chatting to someone on chrome team about why we this API event exists given that its pretty much indistinguishable from `Promise.resolve().then(X)` which predates `queueMicrotask` and is available everywhere.

test/test_browser.py
juj: The reason I didn't, was to get the `@requires_shared_array_buffer` annotation be precise.

Though now I remember I have a `pthread` magic regex catch for the same purpose, so changed to this form.

src/lib/libhtml5.js
juj: Yeah.. Both the push and pop callback and callUserCallback are nice for some uses, but result in bloat :/

Added these, maybe callUserCallback() can be made zero-cost some day. (Or maybe it is so at least with Closure, not sure)

test/test_browser.py
sbc100: I think we should be consistent and use @parameterize everywhere.

There must be magic handling that injects `@requires_shared_array_buffer` I suppose?  If you don't like that we could make a new `@also_with_proxy_to_pthread` that takes care of adding the cflags and also adding `requires_shared_array_buffer`.

src/lib/libhtml5.js
sbc100: We do need `callUserCallback` whenever we enter user code from the event loop otherwise lots of things can break.   A classic one is calling `exit()` from within the callback.

IIRC the overhead of `callUserCallback` in MINIMAL_RUNTIME, where we don't have the same requirements is very close to zero.   If you want to make reduce it to absolutely zero that seems like a good thing, but unrelated to this specific change.

system/include/emscripten/html5.h
```diff
@@ -466,6 +466,8 @@ int emscripten_request_animation_frame(bool (*cb)(double time, void *userData),
 void emscripten_cancel_animation_frame(int requestAnimationFrameId);
 void emscripten_request_animation_frame_loop(bool (*cb)(double time, void *userData), void *userData);
 
+void emscripten_queue_microtask(void (*cb)(void *userData), void *userData);
```
sbc100: This should probably go in `emscripten/eventloop.h`.

test/test_browser.py
```diff
@@ -5016,6 +5016,13 @@ def test_emscripten_request_animation_frame_loop(self):
   def test_request_animation_frame(self):
     self.btest_exit('test_request_animation_frame.c')
 
+  @parameterized({
+    '': ([],),
+    'proxy_to_pthread': (['-pthread', '-sPROXY_TO_PTHREAD'],),
+  })
+  def test_emscripten_queue_microtask(self, flags):
+    self.btest_exit('emscripten_queue_microtask.c', cflags=cflags)
```
sbc100, line 5032: This test can run on node too so maybe can live in test_other.py?  But maybe we want to specifically test in the browser?

See https://nodejs.org/api/globals.html#queuemicrotaskcallback

juj: Hmm, did you have a mechanism for running browser suite with 'node' as the browser executable? I recall glancing over something along those lines.

Otherwise I presume we could duplicate the test in other + browser?

sbc100: You can do `--browser=node` as a quick hack, but the there is no president for doing that anywhere except on your local machine. 


The current method is to just duplicate the tests that it makes sense to duplicate.

src/lib/libhtml5.js
```diff
@@ -2335,6 +2335,15 @@ var LibraryHTML5 = {
...
   },
 
+  emscripten_queue_microtask__deps: ['$callUserCallback'],
+  emscripten_queue_microtask: (cb, userData) => {
+    {{{ runtimeKeepalivePush(); }}}
+    return queueMicrotask(() => {
+      {{{ runtimeKeepalivePop(); }}}
+      callUserCallback(() => {{{ makeDynCall('vp', 'cb') }}}(userData));
+    });
+  },
```
sbc100, line 2345: Move to `libeventloops.js`?

PR #25742, opened 2025-11-07
Fix references to browser_common in test runner

juj:


juj:
Landing to unblock CI.

PR #25743, opened 2025-11-07
Fix test start time tracking

juj:
Fix test start time tracking. Used when EMTEST_VISUALIZE=1 env. var. is set.


juj:
Landing to unblock CI.

sbc100:
Sorry about that, I guess we should add some testing for EMTEST_VISUALIZE

PR #25747, opened 2025-11-07
[AUDIO_WORKET] Add support for setting the render quantum size

cwoffenden:
The `renderSizeHint` is now available in Chrome Canary (`--enable-features=WebAudioConfigurableRenderQuantum`) so this can finally be tested. Most of the work was already done in preparation, so this required few changes.

Test with:
```
```
Which has been extended to request a large `2048` sample audio frame (which in an older browser will fallback to the default 128 samples).

Note for me: look at why and since how long we can bust out of the stack without asserting. The stackAlloc() no longer appears to trip when requesting more bytes than `wwParams.stackSize`.

### DIFF
--- a/src/lib/libwebaudio.js
+++ b/src/lib/libwebaudio.js
@@ -95,8 +95,8 @@ var LibraryWebAudio = {
 #endif
 #endif
 
-    // AUDIO_CONTEXT_RENDER_SIZE_DEFAULT and AUDIO_CONTEXT_RENDER_SIZE_HARDWARE
-    // into their AudioContextRenderSizeCategory enum, or take the positive int. 
+    // Converts AUDIO_CONTEXT_RENDER_SIZE_* into AudioContextRenderSizeCategory
+    // enums, otherwise returns a positive int value.
     function readRenderSizeHint(val) {
       return (val < 0) ? 'hardware' : (val || 'default');
     }


src/lib/libwebaudio.js
```diff
@@ -100,9 +95,15 @@ var LibraryWebAudio = {
 #endif
 #endif
 
+    // AUDIO_CONTEXT_RENDER_SIZE_DEFAULT and AUDIO_CONTEXT_RENDER_SIZE_HARDWARE
+    // into their AudioContextRenderSizeCategory enum, or take the positive int. 
+    function readRenderSizeHint(val) {
+        return (val == 0) ? "default" : ((val < 0) ? "hardware" : val);
```
lindell: I think it is better to be clearer and have a couple of extra lines here. Nested ternary expressions are hard to read.


```suggestion
        if (val == 0) return "default";
        if (val == -1) return "hardware";
        return val;
```

lindell (commented):
Thanks for a fast PR :)

src/lib/libwebaudio.js
cwoffenden: I agree, it's terrible to read, but @juj would've had something to say about the extra bytes!

src/lib/libwebaudio.js
```diff
@@ -77,12 +77,7 @@ var LibraryWebAudio = {
 
   // Performs the work of getting the AudioContext's render quantum size.
   $emscriptenGetContextQuantumSize: (contextHandle) => {
-    // TODO: in a future release this will be something like:
-    //   return EmAudio[contextHandle].renderQuantumSize || 128;
-    // It comes two caveats: it needs the hint when generating the context adding to
-    // emscripten_create_audio_context(), and altering the quantum requires a secure
-    // context and fallback implementing. Until then we simply use the 1.0 API value:
-    return 128;
+    return EmAudio[contextHandle].renderQuantumSize || 128;
```
juj: Closure will not understand `renderQuantumSize`, unless you update the externs (not sure if the externs of AudioWorklet were managed by us, or by upstream). So this would need to be `EmAudio[contextHandle]['renderQuantumSize']` to be Closure compatible.

I imagine in a distant future when all browsers implement `renderQuantumSize`, we could gate this with `#if MIN_x_VERSION < yyy` checks to phase out the `|| 128` part, and require `renderQuantumSize` to be present.

src/lib/libwebaudio.js
juj: Couldn't resist golfing here:

```
return val < 0 ? 'hardware' : val || 'default';
```

src/audio_worklet.js
```diff
@@ -154,6 +154,10 @@ function createWasmAudioWorkletProcessor() {
       var stackMemoryAligned = (stackMemoryStruct + stackMemoryData + 15) & ~15;
       var structPtr = stackAlloc(stackMemoryAligned);
       var dataPtr = structPtr + (stackMemoryAligned - stackMemoryData);
+#if ASSERTIONS
+      // TODO: look at why stackAlloc isn't tripping the assertions
+      console.assert(stackMemoryAligned <= wwParams.stackSize, `Not enough stack allocated to the AudioWorklet (need ${stackMemoryAligned}, got ${wwParams.stackSize})`);
```
juj: We haven't traditionally used `console.assert()`, but `assert(...)`

src/lib/libwebaudio.js
cwoffenden: Fixed by quoting.

src/lib/libwebaudio.js
cwoffenden: Haha!

src/audio_worklet.js
cwoffenden: I just followed my earlier code which is all `console.assert()` (I think because it goes back to when this was a standalone file).

src/lib/libwebaudio.js
cwoffenden: (Replaced with something less hideous)

src/lib/libwebaudio.js
```diff
@@ -100,9 +95,15 @@ var LibraryWebAudio = {
 #endif
 #endif
 
+    // AUDIO_CONTEXT_RENDER_SIZE_DEFAULT and AUDIO_CONTEXT_RENDER_SIZE_HARDWARE
+    // into their AudioContextRenderSizeCategory enum, or take the positive int. 
+    function readRenderSizeHint(val) {
+        return (val < 0) ? 'hardware' : (val || 'default');
```
sbc100: two space indent

src/struct_info.json
```diff
@@ -1269,7 +1269,8 @@
         "structs": {
             "EmscriptenWebAudioCreateAttributes": [
               "latencyHint",
-              "sampleRate"
+              "sampleRate",
+              "renderSizeHint"
```
sbc100, line 1273: Why not `quantumSizeHint`?

cwoffenden: > Why not `quantumSizeHint`?

I copied what the JS API now calls it:

https://github.com/emscripten-core/emscripten/blob/e455f226f73e161ecf20b7c2ed3590f970c0cdd9/src/lib/libwebaudio.js#L106

src/lib/libwebaudio.js
cwoffenden: For once I managed to not get tabs in there...

Fixed.

src/lib/libwebaudio.js
```diff
@@ -100,9 +95,15 @@ var LibraryWebAudio = {
 #endif
 #endif
 
+    // AUDIO_CONTEXT_RENDER_SIZE_DEFAULT and AUDIO_CONTEXT_RENDER_SIZE_HARDWARE
+    // into their AudioContextRenderSizeCategory enum, or take the positive int.
```
juj: Wonder if a verb got cut off from this comment, or maybe this was intentionally brief?

src/struct_info.json
juj: I wonder if there are any users of this API yet, could we rename `quantum` -> `render` as well? E.g. in

`$emscriptenGetContextQuantumSize` -> `$emscriptenGetContextRenderSize`
`emscripten_audio_context_quantum_size` -> `emscripten_audio_context_render_size`
?

system/include/emscripten/webaudio.h
```diff
@@ -19,10 +19,16 @@ extern "C" {
...
+// Default render size of 128 frames
+#define AUDIO_CONTEXT_RENDER_SIZE_DEFAULT 0
+// Let the hardware determine the best render size
+#define AUDIO_CONTEXT_RENDER_SIZE_HARDWARE -1
+
 typedef struct EmscriptenWebAudioCreateAttributes
 {
 	const char *latencyHint; // Specify one of "balanced", "interactive" or "playback"
 	uint32_t sampleRate; // E.g. 44100 or 48000
+	int32_t renderSizeHint; // AUDIO_CONTEXT_RENDER_SIZE_* or number of samples
```
juj, line 31: I presume since this is a hint, the context creation is allowed to not honor it?

juj (approved):
Looks great - I'll wait for that one comment before merging.

system/include/emscripten/webaudio.h
cwoffenden: > I presume since this is a hint, the context creation is allowed to not honor it?

From the spec:

https://webaudio.github.io/web-audio-api/#dom-audiocontextoptions-rendersizehint

_"It is a hint that might not be honored."_

src/struct_info.json
cwoffenden: The spec mixes the two, even for the `renderSizeHint` it says _"This allows users to ask for a particular render quantum size"._

Currently there's no reason to use this call, since it's always been 128. It's handy before the callback for calculating the stack size, and I should write an example of this.

src/lib/libwebaudio.js
cwoffenden: It made sense when I typed it... I'll clarify it.

Issue #25745, opened 2025-11-07
AttributeError: 'BufferedParallelTestResult' object has no attribute '_original_stdout'


juj:
After https://github.com/emscripten-core/emscripten/pull/25737/files#diff-fd5a5983b044aa841287bf197f2b5e46781162bf0e6a0ca8d1ccaddf1cc1c969R335-R339 I am getting

```
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/parallel_testsuite.py", line 101, in run_test
    test(result)
    ~~~~^^^^^^^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/unittest/case.py", line 707, in __call__
    return self.run(*args, **kwds)
           ~~~~~~~~^^^^^^^^^^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/retryable_unittest.py", line 22, in run
    super().run(result)
    ~~~~~~~~~~~^^^^^^^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/unittest/case.py", line 677, in run
    result.stopTest(self)
    ~~~~~~~~~~~~~~~^^^^^^
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/parallel_testsuite.py", line 339, in stopTest
    del self._original_stdout
        ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'BufferedParallelTestResult' object has no attribute '_original_stdout'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
                    ~~~~^^^^^^^^^^^^^^^
  File "/home/clb/.pyenv/versions/3.13.3/lib/python3.13/multiprocessing/pool.py", line 51, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
  File "/home/clb/buildbot/h12dsi-linux-mint22/emscripten_linux_x64/build/emscripten/main/test/parallel_testsuite.py", line 109, in run_test
    result.addError(test, e)
[...]

PR #25746, opened 2025-11-07
Delete _original_stdout and _original_stderr only if they exist. #25745

juj:
The first thing that came to mind to fix #25745. I didn't dig too much into the details here if there might be something better.

PR #25748, opened 2025-11-07
Fix parallel_testsuite in the case that tests are retried

sbc100:
Fixes: #25745

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -115,6 +115,13 @@ def run_test(test, allowed_failures_counter, lock, progress_counter, num_tests, 
   # working directory is not within it.
   os.chdir(olddir)
   common.force_delete_dir(temp_dir)
+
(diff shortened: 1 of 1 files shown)


test/parallel_testsuite.py
```diff
@@ -115,6 +115,13 @@ def printResult(res):
   # working directory is not within it.
   os.chdir(olddir)
   common.force_delete_dir(temp_dir)
+
+  # Since we are returning this result to the main thread we need to make sure
+  # that it is serializable/picklable. To do this, we delete any non-pickalble
```
kripken:
```suggestion
  # that it is serializable/picklable. To do this, we delete any non-picklable
```
(not that this is a real word, no idea how to spell it :smile: but this makes us more consistent internally)

kripken (approved):

sbc100:
Supersedes (I think) #25746

PR #25751, opened 2025-11-07
parallel_testsuite: Switch to pool.imap_unordered. NFC

sbc100:
This allows us the print that status line back on the main thread as the results arrive there.  It also avoids some of the shared state.

I have a followup change that uses just a single line for output line other modern test runners do.

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -43,7 +43,8 @@ def cap_max_workers_in_pool(max_workers, is_browser):
   return max_workers
 
 
-def run_test(test, allowed_failures_counter, lock, progress_counter, num_tests, buffer):
+def run_test(args):
+  test, allowed_failures_counter, lock, buffer = args
   # If we have exceeded the number of allowed failures during the test run,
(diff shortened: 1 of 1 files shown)


test/parallel_testsuite.py
```diff
@@ -177,15 +180,23 @@ def run(self, result):
...
         else:
           allowed_failures_counter = manager.Value('i', self.max_failures)
-          progress_counter = manager.Value('i', 0)
           lock = manager.Lock()
-        results = pool.starmap(run_test, ((t, allowed_failures_counter, lock, progress_counter, len(tests), result.buffer) for t in tests), chunksize=1)
+
+        results = []
+        args = ((t, allowed_failures_counter, lock, result.buffer) for t in tests)
+        for res in pool.imap_unordered(run_test, args, chunksize=1):
+          # results maybe be None if # of allowed errors was exceeded
```
kripken:
```suggestion
          # results may be be None if # of allowed errors was exceeded
```

PR #25752, opened 2025-11-08
[parallel_testsuite.py] Output testing results in a single line

sbc100:
Like many modern tests runners we can output all results on a single line a by re-writing the some line on the TTY.

We only do this under certain circumstances:

1. stderr is a TTY
2. The `--buffer` options us used (test output hidden unless fail)

This mode is still not the default yet since `--buffer` is not yet the default, but the output is so much nicer I think we should do that at some point.

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -3,13 +3,17 @@
 # University of Illinois/NCSA Open Source License.  Both these licenses can be
 # found in the LICENSE file.
 
+import io
 import json
+import logging
 import multiprocessing
 import os
+import shutil
 import sys
 import tempfile
 import time
 import unittest
+from types import MethodType
 
 import browser_common
 import common
@@ -107,6 +111,12 @@ def tear_down():
   return True
 
 
+def clearline():
+  assert sys.stderr.isatty()
+  sys.stdout.write('\r\033[K')
+  sys.stdout.flush()
+
+
 class ParallelTestSuite(unittest.BaseTestSuite):
   """Runs a suite of tests in parallel.
 
@@ -119,14 +129,16 @@ class ParallelTestSuite(unittest.BaseTestSuite):
     self.max_failures = options.max_failures
     self.failing_and_slow_first = options.failing_and_slow_first
     self.progress_counter = 0
+    self.use_single_line_output = options.buffer and sys.stderr.isatty()
+    if self.use_single_line_output:
+      self.terminal_width = shutil.get_terminal_size()[0]
 
   def addTest(self, test):
     super().addTest(test)
     test.is_parallel = True
 
   def printOneResult(self, res):
-    percent = int(self.progress_counter * 100 / self.num_tests)
-    progress = f'[{percent:2d}%] '
+    progress = f'[{self.progress_counter}/{self.num_tests}] '
     self.progress_counter += 1
 
     if res.test_result == 'success':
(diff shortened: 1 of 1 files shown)


sbc100:
Small things, but this change I think will be huge qualify of life improvement for those of us that run the test suite on the command line a lot!

sbc100:
shot

This is now what it looks like when I run the entire core test suite of 1176 tests!

test/parallel_testsuite.py
```diff
@@ -107,6 +111,12 @@ def tear_down():
   return True
 
 
+def clearline():
+  assert sys.stderr.isatty()
+  sys.stdout.write('\r\033[K')
```
juj, line 116: This might warrant a comment on what this code is?

test/parallel_testsuite.py
```diff
@@ -119,14 +129,16 @@ def __init__(self, max_cores, options):
     self.max_failures = options.max_failures
     self.failing_and_slow_first = options.failing_and_slow_first
     self.progress_counter = 0
+    self.use_single_line_output = options.buffer and sys.stderr.isatty()
+    if self.use_single_line_output:
```
juj, line 133: Maybe instead of `use_single_line_output`, a more precise name could be something like `collapse_test_success_prints`?

test/parallel_testsuite.py
```diff
@@ -119,14 +129,16 @@ def __init__(self, max_cores, options):
...
+      self.terminal_width = shutil.get_terminal_size()[0]
```
juj, line 134: How expensive is getting this size? Caching this size here sounds like things will break if one then resizes the window size mid-run?

It might be quite common for user to start a test run, then mid-run realize "the logs are not fitting on one line, let me grow the terminal window a bit."

test/parallel_testsuite.py
```diff
@@ -139,8 +151,11 @@ def printOneResult(self, res):
       msg = 'FAIL'
       color = RED
     elif res.test_result == 'skipped':
-      reason = res.skipped[0][1]
-      msg = f"skipped '{reason}'"
+      if self.use_single_line_output:
+        msg = 'skipped'
+      else:
+        reason = res.skipped[0][1]
+        msg = f"skipped '{reason}'"
```
juj, line 158: What is this block doing? In single line output mode, the skip reasons are not printed? Why does that matter if the skip reasons get cleared anyways?

test/parallel_testsuite.py
```diff
@@ -154,7 +169,23 @@ def printOneResult(self, res):
     if res.test_result != 'skipped':
       msg += f' ({res.elapsed:.2f}s)'
 
-    errlog(f'{with_color(CYAN, progress)}{res.test} ... {with_color(color, msg)}')
+    if self.use_single_line_output:
+      min_len = len(progress) + len(msg) + 5
+      max_name = self.terminal_width - min_len
+      test_name = str(res.test)[:max_name]
```
juj, line 175: What is happening here? Some kind of padding? Maybe a comment would help.

test/parallel_testsuite.py
```diff
@@ -229,7 +261,25 @@ def update_test_results_to(test_name):
 
       json.dump(previous_test_run_results, open(common.PREVIOUS_TEST_RUN_RESULTS_FILE, 'w'), indent=2)
 
-    return self.combine_results(result, results)
+    old_stream = result.stream
+    if self.use_single_line_output:
+      def writeln(x, arg=None):
+        if arg:
+            x.write(arg)
+        x.write('\n')
```
juj, line 269: `x` -> `stream`? `arg` -> `text`?

juj:
Looks good, though needs Windows and macOS testing before landing. We still have the previous Windows 10 TTY/coloring regression piled up, I hope this won't interact with that.

sbc100:
> Looks good, though needs Windows and macOS testing before landing. We still have the previous Windows 10 TTY/coloring regression piled up, I hope this won't interact with that.

Was that not fixed by #25529?   What is the regression you are referring to?

sbc100:
Closing in favor of better solution pending.

PR #25754, opened 2025-11-08
Fix test_bootstrap_without_em_config

sbc100:
This tests was broken on systems where the python executable lives in path that also contains clang executable.

We were remove PATH elements here, but we cannot remove the one that contains the python executable otherwise the bootstrap script cannot run at all.


sbc100:
This test was failing on my system, I'm not sure for how long.

PR #25755, opened 2025-11-08
Add SingleLineTestRunner/Result. NFC

sbc100:
This test runner does a few things differ the base TextTestRunner:

1. It improves the behviour of `--buffer` by also buffering/redirecting logging output that occurs during the test run.
2. It displays all results on a single line, each result erasing the contents of the line before re-drawing it.
3. It uses ANSI colors to the show the results.
4. It should the progress as each results is displayed so its easy to see how far you are through the test suite "[XX/YY]"

Followups that I'm a planning:

1. Enable `--buffer` by default which will then enable this runner by default
2. Integrate with parallel running (for now you need use `-j1` to see the effect of this runner)

I also updated parallel_testsuite.py use the same "XX/YY" progress rather than a percent.

See #25752, which implements similar thing in the parallel_runner.   Hopefully once this change lands I can simplify #25752.

### DIFF
--- a/test/color_runner.py
+++ b/test/color_runner.py
@@ -1,0 +1,72 @@
+# Copyright 2025 The Emscripten Authors.  All rights reserved.
+# Emscripten is available under two separate licenses, the MIT license and the
+# University of Illinois/NCSA Open Source License.  Both these licenses can be
+# found in the LICENSE file.
+
+import logging
+import unittest
+
+from tools.colored_logger import CYAN, GREEN, RED, with_color
+
+
+class BufferingMixin:
+  """This class takes care of redirecting `logging` output in `buffer=True` mode.
+
+  To use this class inherit from it along with a one of the standard unittest result
+  classes.
+  """
+  def _setupStdout(self):
+    super()._setupStdout()
+    # In addition to redirecting sys.stderr and sys.stdout, also update the python
+    # loggers which hold cached versions of these handles.
+    if self.buffer:
+      for handler in logging.root.handlers:
+        if handler.stream == self._original_stderr:
+          handler.stream = self._stderr_buffer
+
+  def _restoreStdout(self):
+    super()._restoreStdout()
+    if self.buffer:
+      for handler in logging.root.handlers:
+        if handler.stream == self._stderr_buffer:
+          handler.stream = self._original_stderr
+
+
+class ProgressMixin:
+  test_count = 0
+  progress_counter = 0
+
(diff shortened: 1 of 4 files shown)


test/color_runner.py
```diff
@@ -0,0 +1,72 @@
...
+# found in the LICENSE file.
+
+import logging
+import unittest
+
+from tools.colored_logger import CYAN, GREEN, RED, with_color
+
+
+class BufferingMixin:
+  """This class takes care of redirectting `logging` output in `buffer=True` mode.
```
kripken:
```suggestion
  """This class takes care of redirecting `logging` output in `buffer=True` mode.
```

test/color_runner.py
```diff
@@ -0,0 +1,72 @@
...
+
+  To use this class inherit from it along with a one of the standard unittest result
+  classe.
```
kripken:
```suggestion
  classes.
```

test/color_runner.py
```diff
@@ -0,0 +1,72 @@
...
+  """
+  def _setupStdout(self):
+    super()._setupStdout()
+    # In addition to redirecting sys.stderr and sys.stdout, also update the python
+    # loggers have cached versions of these.
```
kripken:
```suggestion
    # loggers to have cached versions of these.
```

test/runner.py
```diff
@@ -427,8 +429,12 @@ def run_tests(options, suites):
     testRunner = xmlrunner.XMLTestRunner(output=output, verbosity=2,
                                          failfast=options.failfast)
     print('Writing XML test output to ' + os.path.abspath(output.name))
+  elif options.buffer and options.ansi and not options.verbose:
+    # And buffering is enabled and ansi color output is available use our nice single-line
```
kripken:
```suggestion
    # When buffering is enabled and ansi color output is available use our nice single-line
```

test/single_line_runner.py
```diff
@@ -0,0 +1,95 @@
...
+    self.progress_counter = 0
+
+  def writeStatusLine(self, line):
+    clearline(self._original_stderr)
+    self._original_stderr.write(line)
+    self._original_stderr.flush()
+
+  def updateStatus(self, test, msg, color):
+    progress = f'[{self.progress_counter}/{self.test_count}] '
+    # Format the line so that it fix within the terminal width, unless its less then min_len
```
```suggestion
    # Format the line so that it fix within the terminal width, unless it's less then min_len
```

PR #25760, opened 2025-11-10
[AUDIO_WORKLET] clarify docs and use for renderSizeHint

cwoffenden:
Minor clarification of the docs and example.

### DIFF
--- a/site/source/docs/api_reference/wasm_audio_worklets.rst
+++ b/site/source/docs/api_reference/wasm_audio_worklets.rst
@@ -44,12 +44,11 @@ and then, these processors are instantiated one or more times in the audio
 processing graph as AudioWorkletNodes.
 
 Once a class type is instantiated on the Web Audio graph and the graph is
-running, a C/C++ function pointer callback will be invoked for each 128
-samples of the processed audio stream that flows through the node. Newer Web
-Audio API specs allow this to be changed, so for future compatibility use the
-``AudioSampleFrame``'s ``samplesPerChannel`` to get the value (and request the
-preferred number of samples with ``EmscriptenWebAudioCreateAttributes``'s
-``renderSizeHint`` option).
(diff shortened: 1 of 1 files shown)


site/source/docs/api_reference/wasm_audio_worklets.rst
```diff
@@ -47,7 +47,9 @@ Once a class type is instantiated on the Web Audio graph and the graph is
 running, a C/C++ function pointer callback will be invoked for each 128
 samples of the processed audio stream that flows through the node. Newer Web
 Audio API specs allow this to be changed, so for future compatibility use the
-``AudioSampleFrame``'s ``samplesPerChannel`` to get the value.
+``AudioSampleFrame``'s ``samplesPerChannel`` to get the value (and request the
+preferred number of samples with ``EmscriptenWebAudioCreateAttributes``'s
+``renderSizeHint`` option).
```
sbc100: Why not just remove the hardcoded 128 value here and say "callback will be invoked for each N
samples of the processed (where N is is the number of samples per channel)".

The naming here also seems strange.  Why is to called `samplesPerChannel` when its read but `renderSizeHint` when its hinted.  Why not `samplesPerChannelHint`?

cwoffenden: `samplesPerChannel` has been in the Emscripten API for about a year in prep for the web audio API changes. `renderSizeHint` is what the spec calls it and is now available in Chrome Canary.

I chose `samplesPerChannel` last year when extending the API, simply because it speaks the most to what it does and where it's accessed in the API (and having been doing this sort of thing since forever, I've never once heard anyone call it a render quantum).

I think it does make sense to change the 128, but I think it'll stay 128 for most people for a long time.

cwoffenden: Clarified the 128.

PR #25761, opened 2025-11-10
Improve the SingleLineTestResult and integrate with parallel_testsuite

sbc100:
Now the parallel test suite now longer needs its own reporting logic it can just share use the existing Runner/Result which now supports the color and single long output formats.

By default we now buffer test output and use the single line test runner.  Adding `-v` will disabled this and give multi-line non-buffered output.  Adding another `-v` will enable logging in the test framework (same as setting EMTEST_VERBOSE)

### DIFF
--- a/test/single_line_runner.py
+++ b/test/single_line_runner.py
@@ -6,7 +6,7 @@
 import shutil
 import unittest
 
-from color_runner import BufferingMixin, ColorTextRunner
+from color_runner import ColorTextResult, ColorTextRunner
 
 from tools.colored_logger import CYAN, GREEN, RED, with_color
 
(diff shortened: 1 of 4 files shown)


sbc100:
Replaced #25752

sbc100:
shot

test/single_line_runner.py
```diff
@@ -20,65 +20,47 @@ def term_width():
...
-  def writeStatusLine(self, line):
-    clearline(self._original_stderr)
-    self._original_stderr.write(line)
-    self._original_stderr.flush()
-
-  def updateStatus(self, test, msg, color):
-    progress = f'[{self.progress_counter}/{self.test_count}] '
-    # Format the line so that it fix within the terminal width, unless it's less then min_len
+  def writeStatus(self, test, msg, color, line_pos):
+    # Becuse the message can include the skip reason (which can be very long somtimes), truncate it
```
```suggestion
    # Because the message can include the skip reason (which can be very long sometimes), truncate it
```

test/single_line_runner.py
```diff
@@ -20,65 +20,47 @@ def term_width():
...
+    # to a reasonable length to avoid exceeding ling length.
```
```suggestion
    # to a reasonable length to avoid exceeding line length.
```

brendandahl:
Looks nice!
I don't think we need to fix this yet, but one thing I notice when running with `-j1` is it will show the last test that has passed e.g. `(test_core.core0.test_promise_await_jspi) ... ok` instead of the currently running test.

sbc100:
> Looks nice! I don't think we need to fix this yet, but one thing I notice when running with `-j1` is it will show the last test that has passed e.g. `(test_core.core0.test_promise_await_jspi) ... ok` instead of the currently running test.

Yes thats is how the parallel test runner works for sure.   We don't  overwrite the old line until we have new test result.

For the `-j1` runner I agree we could maybe show the current tests.. but then you would also never see the `.. ok` since that would be immediately replaced with the new test name following by `...`

brendandahl:
I'd be fine not seeing the 'ok', it's more helpful to me to see if a test is taking a long time or hung.

sbc100:
> Looks nice! I don't think we need to fix this yet, but one thing I notice when running with `-j1` is it will show the last test that has passed e.g. `(test_core.core0.test_promise_await_jspi) ... ok` instead of the currently running test.

OK, I've updated the code to have this behaviour now!

PR #25762, opened 2025-11-10
Add error handling for stderrno when errno is invalid

google-yfyang:
This is to match llvm-libc's behavior when an invalid errno is specified.

When errno is specified as -1, for instance, the current behavior prints "Success". With this change, any invalid errno emits out  "Unknown error %d".


system/lib/libc/musl/src/errno/strerror.c
```diff
@@ -34,15 +34,21 @@ char *__strerror_l(int e, locale_t loc)
...
 	else if (e==EDQUOT_ORIG) e=EDQUOT;
 #endif
-	if (e >= sizeof errmsgidx / sizeof *errmsgidx) e = 0;
-	s = (char *)&errmsgstr + errmsgidx[e];
 #ifdef __EMSCRIPTEN__
+	if (e < 0 || e >= sizeof errmsgidx / sizeof *errmsgidx || (e != 0 && !errmsgidx[e])) {
+		static char buf[32];
+		snprintf(buf, sizeof buf, "Unknown error %d", e);
+		return buf;
+	}
```
sbc100, line 40: Can we avoid dragging in snprintf here?  How about just returning "Unknown Error"?

system/lib/libc/musl/src/errno/strerror.c
```diff
@@ -34,15 +34,21 @@ char *__strerror_l(int e, locale_t loc)
 	if (e==EDQUOT) e=0;
 	else if (e==EDQUOT_ORIG) e=EDQUOT;
 #endif
-	if (e >= sizeof errmsgidx / sizeof *errmsgidx) e = 0;
-	s = (char *)&errmsgstr + errmsgidx[e];
 #ifdef __EMSCRIPTEN__
+	if (e < 0 || e >= sizeof errmsgidx / sizeof *errmsgidx || (e != 0 && !errmsgidx[e])) {
```
sbc100, line 38: I don't think you need the third clause here do you?   If `e` is in range then we know valid right?

system/lib/libc/musl/src/errno/strerror.c
google-yfyang: This is really just to match what llvm-libc is doing. https://github.com/llvm/llvm-project/blob/main/libc/src/__support/StringUtil/error_to_string.cpp#L62

system/lib/libc/musl/src/errno/strerror.c
google-yfyang: We technically don't need this atm. But llvm-libc's error handling does handle discontinuous errno.

sbc100: Oh i see .. the array is not necessarily continuous.

sbc100: That seems like separate unrelated issue though.

google-yfyang: I can remove this if you prefer.

system/lib/libc/musl/src/errno/strerror.c
google-yfyang: I see. Not all environments have snprintf here. The llvm-libc has access to some cpp constructs; so the only way I can think of is to implement a simple itoa for this purpose.

sbc100: No, sorry, we do always have snprintf.. I was just hoping we could make just avoid it so there is no dependency here?   i.e. if a program only calls strerror it doesn't pay the codesize cost of strnprintf.

But I guess most real world program will include snprintf so maybe its not a huge deal.

Can we just report "Unknown error" working the integer value?

google-yfyang: There were some lto tests that failed with snprintf. For instance, `test_longjmp_standalone_standalone (test_core.lto0.test_longjmp_standalone_standalone) ... wasm-ld: error: /usr/local/google/home/yfyang/emscripten/cache/sysroot/lib/wasm32-emscripten/lto/libc-debug.a(snprintf.o): attempt to add bitcode file after LTO (snprintf)`.

The llvm-libc counterpart makes use of some cpp constructs like bufferstream and span, which might not be super concise as well. But I get what you are saying. 

Returning with just "Unknown error" deviates from both musl libc and llvm-libc. If you find conforming to llvm-libc can have a negative impact on code size, I think we can just revert back to the behavior of musl and returns "No error information".

sbc100: Hmm.. we should try to fix the "attempt to add bitcode file after LTO" issue think, but it not easy than I think we just go with a "Unknown Error" and update the absl tests accordingly.

google-yfyang: Error code is now removed from the error message.

test/test_core.py
```diff
@@ -6169,6 +6169,10 @@ def test_unistd_curdir(self):
   def test_unistd_close(self):
     self.do_run_in_out_file_test('unistd/close.c')
 
+  @also_with_noderawfs
+  def test_unistd_errno(self):
+    self.do_run_in_out_file_test('unistd/errno.c')
```
sbc100: Can you move this to test_other (line 13126)

test/unistd/errno.c
```diff
@@ -0,0 +1,17 @@
...
+
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <unistd.h>
+#include <emscripten.h>
+
+int main() {
+    int error = -1;
+    printf("errno: %s\n", strerror(error));
```
sbc100: Can you call this `strerror.c`?  And test_unistd_strerror?

sbc100: Just pass -1 directly?

test/unistd/errno.c
```diff
@@ -0,0 +1,17 @@
...
+ * Emscripten is available under two separate licenses, the MIT license and the
+ * University of Illinois/NCSA Open Source License.  Both these licenses can be
+ * found in the LICENSE file.
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <unistd.h>
+#include <emscripten.h>
```
sbc100: No need for this include

test/unistd/errno.c
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2011 The Emscripten Authors.  All rights reserved.
```
sbc100: 2025

test/unistd/strerror.c
```diff
@@ -0,0 +1,13 @@
...
+ * University of Illinois/NCSA Open Source License.  Both these licenses can be
+ * found in the LICENSE file.
+ */
+
+#include <stdio.h>
+#include <string.h>
+
+int main() {
+    printf("errno: %s\n", strerror(-1));
+}
```
sbc100: Missing newline at EOF

sbc100 (approved):

sbc100:
It looks this does come at a codesize cost.   It is a very small one so maybe fine.  Can you run `./tools/maint/rebaseline_tests.py`

google-yfyang:
> It looks this does come at a codesize cost. It is a very small one so maybe fine. Can you run `./tools/maint/rebaseline_tests.py`

Done.

test/codesize/test_codesize_hello_dylink_all.json
```diff
@@ -1,7 +1,7 @@
 {
   "a.out.js": 245483,
-  "a.out.nodebug.wasm": 574056,
-  "total": 819539,
+  "a.out.nodebug.wasm": 573907,
```
sbc100, line 3: How did this one get smaller??  Strange.

google-yfyang: possibly heavy optimization takes the return "Unknown error" branch and avoids the more complicated 
`s = (char *)&errmsgstr + errmsgidx[e];`?

PR #25765, opened 2025-11-10
Fix ccall() returning a pointer in CAN_ADDRESS_2GB mode.

juj:
`ccall()` returned pointers as negative numbers in CAN_ADDRESS_2GB modes, breaking pointer `begin < end` comparisons.


juj:
Of course we could.

I wrote this in core0 to ensure that it also works in wasm64 mode, so it gets all the codegen variants tested.

sbc100:
> Of course we could.
> 
> I wrote this in core0 to ensure that it also works in wasm64 mode, so it gets all the codegen variants tested.

Ok, leave but maybe the test could stay in code but be as simple as the one I show?

sbc100 (commented):
The test just seems a lot more complex that it needs to be.   I can't just look at it and see what its trying to show.

sbc100 (commented):
If you are specifically concern about comparisons then something like this seems like it should work om EM_ASM:

```
var end = ccall("getDataStart"):
var start = call("getDataEnd");
assert(start < end);
```

juj:
Yeah, that works.. simplified the code but kept in core. Probably the begin < end invariant will be enough to test via the > 0 condition.

PR #25767, opened 2025-11-10
[parallel_testsuite.py] Simplify allowed_failures_counter

sbc100:
We do all the modification of allowed_failures_counter on the main thread without the need for a lock.

I believe this was not possible prior #25751.

### DIFF
--- a/test/parallel_testsuite.py
+++ b/test/parallel_testsuite.py
@@ -45,7 +45,7 @@ def cap_max_workers_in_pool(max_workers, is_browser):
 
 
 def run_test(args):
-  test, allowed_failures_counter, lock, buffer = args
+  test, allowed_failures_counter, buffer = args
   # If we have exceeded the number of allowed failures during the test run,
   # abort executing further tests immediately.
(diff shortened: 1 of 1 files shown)


test/parallel_testsuite.py
```diff
@@ -174,20 +168,24 @@ def run(self, result):
...
 
         results = []
-        args = ((t, allowed_failures_counter, lock, result.buffer) for t in tests)
+        args = ((t, allowed_failures_counter, result.buffer) for t in tests)
         for res in pool.imap_unordered(run_test, args, chunksize=1):
           # results may be be None if # of allowed errors was exceeded
           # and the harness aborted.
           if res:
+
+            # Alert all other multiprocess pool runners that they need to stop executing further tests.
```
kripken: Is this comment still accurate?

sbc100: Yes, its still basically correct.  Maybe I should remove "all other"..

sbc100: Updated

PR #25773, opened 2025-11-11
[embind] Require C++17 or newer.

brendandahl:
As announced in #24850 and on the [mailing list](https://groups.google.com/g/emscripten-discuss/c/BSLc5y2aKDY) we are going to require C++17 or newer for embind.


sbc100:
CC @RReverser

PR #25780, opened 2025-11-11
Disable assertions in tsgen

stephenduong1004:
When using `tsgen` with `-sASYNCIFY=1` and assertions enabled, an assertion failure occurs: `Assertion failed: missing Wasm export: asyncify_start_unwind`. This is due to https://github.com/emscripten-core/emscripten/pull/25541.

This PR disables assertions when running `tsgen` to resolve the issue.

### DIFF
--- a/tools/link.py
+++ b/tools/link.py
@@ -2056,10 +2056,10 @@ def run_embind_gen(options, wasm_target, js_syms, extra_settings):
     settings.MEMORY64 = 2
   # Source maps haven't been generated yet and aren't needed to run embind_gen.
   settings.LOAD_SOURCE_MAP = 0
+  settings.ASSERTIONS = 0
   outfile_js = in_temp('tsgen.js')
   # The Wasm outfile may be modified by emscripten.emscript, so use a temporary file.
   outfile_wasm = in_temp('tsgen.wasm')
-  settings.ASSERTIONS = 0
   emscripten.emscript(wasm_target, outfile_wasm, outfile_js, js_syms, finalize=False)
   # Build the flags needed by Node.js to properly run the output file.
   node_args = []


tools/link.py
```diff
@@ -2059,6 +2059,7 @@ def run_embind_gen(options, wasm_target, js_syms, extra_settings):
   outfile_js = in_temp('tsgen.js')
   # The Wasm outfile may be modified by emscripten.emscript, so use a temporary file.
   outfile_wasm = in_temp('tsgen.wasm')
+  settings.ASSERTIONS = 0
```
sbc100: Can you move this a few lines up to be with all the other settings mods

sbc100 (commented):
@brendandahl do you know why this is needed?

Can you add a tsgen test that covers this case?

tools/link.py
```diff
@@ -2056,10 +2056,10 @@ def run_embind_gen(options, wasm_target, js_syms, extra_settings):
     settings.MEMORY64 = 2
   # Source maps haven't been generated yet and aren't needed to run embind_gen.
   settings.LOAD_SOURCE_MAP = 0
+  settings.ASSERTIONS = 0
```
sbc100, line 2059: Maybe add a comment here as to why this is needed?

brendandahl:
> @brendandahl do you know why this is needed?

We run tsgen before the wasm-opt asyncify pass so the asyncify functions haven't been created yet.

kripken:
Yeah, basically the wasm isn't fully generated yet... disabling assertions fixes asserts that expect those contents, but I wonder if we might hit other issues later. If so, for Asyncify at least, we'd need to run it before tsgen somehow...

brendandahl:
Fixed slightly differently in #25899

PR #25784, opened 2025-11-12
[parallel_runner.py] Fix failure under XMLRunner

sbc100:
With #25739 I inadvertently broke the handling of errors with the XML test runner.


sbc100:
Landing TBR since this is causing tests failures to get misreported in CI

PR #25786, opened 2025-11-12
Remove global `currentModuleWeakSymbols` from dynamic linking code

sbc100:
Using a global has led to at least one bug: #25214

### DIFF
--- a/src/lib/libdylink.js
+++ b/src/lib/libdylink.js
@@ -164,14 +164,15 @@ var LibraryDylink = {
   },
 
   $GOT: {},
-  $currentModuleWeakSymbols: '=new Set({{{ JSON.stringify(Array.from(WEAK_IMPORTS)) }}})',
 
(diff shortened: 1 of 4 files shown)


src/lib/libdylink.js
```diff
@@ -164,22 +164,23 @@ var LibraryDylink = {
   },
 
   $GOT: {},
-  $currentModuleWeakSymbols: '=new Set({{{ JSON.stringify(Array.from(WEAK_IMPORTS)) }}})',
 
-  // Create globals to each imported symbol.  These are all initialized to zero
-  // and get assigned later in `updateGOT`
+  // Proxy handler used to for GOT.mem and GOT.func imports.  Each of these
```
kripken:
```suggestion
  // Proxy handler used for GOT.mem and GOT.func imports.  Each of these
```

PR #25797, opened 2025-11-14
fix: Big-endian support does not work with optimizations

slavek-kucera:
While migrating to 4.0.19 I discovered that enabling optimizations removes `ParenthesizedExpression` node from AST causing the `littleEndianHeap` to not perform the transform correctly.

### DIFF
--- a/src/lib/liblittle_endian_heap.js
+++ b/src/lib/liblittle_endian_heap.js
@@ -109,11 +109,11 @@ function LE_HEAP_UPDATE() {
     const res = order(Atomics.sub(heap, offset, order(value)));
     return heap.unsigned ? heap.unsigned(res) : res;
   },
-  $LE_ATOMICS_WAIT: (heap, offset, value, timeout) => {
+  $LE_ATOMICS_WAIT: (heap, offset, value, timeout = Infinity) => {
     const order = LE_ATOMICS_NATIVE_BYTE_ORDER[heap.BYTES_PER_ELEMENT - 1];
     return Atomics.wait(heap, offset, order(value), timeout);
   },
-  $LE_ATOMICS_WAITASYNC: (heap, offset, value, timeout) => {
+  $LE_ATOMICS_WAITASYNC: (heap, offset, value, timeout = Infinity) => {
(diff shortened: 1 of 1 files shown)


sbc100:
Sorry if I've asked you this already, but what is you use case for bigendian suport @slavek-kucera ?  Are you running in production on actual BE machines?

slavek-kucera:
Yes, the output targets s390x running z/os.

slavek-kucera:
As for testing, I guess we could add bigendian3 test suite (assuming @juj still runs it in the private pipeline).

juj:
> As for testing, I guess we could add bigendian3 test suite (assuming @juj still runs it in the private pipeline).

I totally do. (e.g. http://clbri.com:8010/#/builders/11/builds/1275 - ignore lots of red for other failures, but there's a passing `bigendian0` at the very bottom of the page)

Posted PR https://github.com/emscripten-core/emscripten/pull/25820 to complete the testing matrix. After that lands, I can add the remaining bigendian suites onto the test runner list.

sbc100:
> Yes, the output targets s390x running z/os.

What engine are you running the emscripten output on?  i.e. Are these servers running Nodejs I guess?

slavek-kucera:
Yes, the engine is nodejs.

juj:
@slavek-kucera My Linux box now completed run of all the bigendian suites. It looks like there are some failures in -O2 and -Os modes: http://clbri.com:8010/#/builders/11/builds/1278

slavek-kucera:
I think those are all false positive build failures due to closure compiler.
```
Function LE_ATOMICS_WAITASYNC: called with 3 argument(s). Function requires at least 4 argument(s) and no more than 4 argument(s)
```
Maybe the function needs an annotation  just like `Atomics.waitAsync`?

slavek-kucera:
@juj Could you re-run the CI on the newest version?

juj:
Now running at http://clbri.com:8010/#/builders/11/builds/1292

slavek-kucera:
It looks like all the big-endian variants are passing now.

juj:
Indeed, this looks great.

juj (approved):

PR #25803, opened 2025-11-15
Move emscripten lld flags to start start of cmd line

sbc100:
This allows the `-Bdynamic` flag to be moved there which is its logical place.  It also theoretically allows user flags to override any system flags.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,8 +20,11 @@ See docs/process.md for more on how version tagging works.
 
 4.0.20 (in development)
 -----------------------
-- Added `emscripten_html5_remove_event_listener` function in `html5.h` in order to be
-  able to remove a single callback. (#25535)
+- Linker flags specified on the command line are now passed to `wasm-ld` after
+  the internal emscripten linker flags.  This means that users can now override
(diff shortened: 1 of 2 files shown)


kripken:
>  It also theoretically allows user flags to override any system flags.

Is this perhaps going to break some users that depended on the current order?

sbc100:
> > It also theoretically allows user flags to override any system flags.
> 
> Is this perhaps going to break some users that depended on the current order?

The only users who might be effected are those that pass wasm-specific `-Wl,xxx` flags.   They might find that these flags are no longer ignore/overridden by emscripten, since we now place the emscripten defaults first.

IMHO this behaviour is more correct.  If a users specifies `-Wl,xxx` we should probably honor that.

Its a very unlikely use case though.. since how/why would such as flag ever have been added but such a user if was to be ignored anyway.  I will add a ChangeLog entry just in case anyway.

kripken (approved):
lgtm with a test - as you say the new behavior is more correct, let's make sure we continue to have that behavior

kripken:
Oops, I didn't notice automerge...

PR #25813, opened 2025-11-17
Refactor test_workerfs. NFC

sbc100:
- Split tests into separate functions.
- Rename test (its not just about reading).
- Remove use of "secret" terminology


test/test_browser.py
```diff
@@ -1433,8 +1433,8 @@ def test_fs_workerfs_read(self):
           files: [file],
         }, '/work');
       };
-    ''' % (secret, secret2))
-    self.btest_exit('fs/test_workerfs_read.c', cflags=['-lworkerfs.js', '--pre-js', 'pre.js', f'-DSECRET="{secret}"', f'-DSECRET2="{secret2}"', '--proxy-to-worker', '-Wno-deprecated', '-lworkerfs.js'])
+    ''' % (file1, file2))
+    self.btest_exit('fs/test_workerfs.c', cflags=['-lworkerfs.js', '--pre-js', 'pre.js', f'-DFILE1="{file1}"', f'-DFILE2="{file2}"', '--proxy-to-worker', '-Wno-deprecated', '-lworkerfs.js'])
```
kripken: "secret" implies it is data (the data is secret), but "file" could be a file's data or its name. How about `file => data`? Or "token/info/etc." ?

sbc100: But the data is is not secret in any way that I can determine.

I'm fine with data, or file_contents, sure,

sbc100: I worked around this by simply hardcoding the expectations in into the file.   Make the whole test easier to read IMHO.

test/fs/test_workerfs.c
```diff
@@ -15,46 +15,61 @@
...
-  int len2 = SECRET_LEN / 2;
-
-  rtn = stat("/work/notexist.txt", &st);
+  int rtn = stat("/work/notexist.txt", &st);
   assert(rtn == -1 && errno == ENOENT);
+}
 
-  rtn = stat("/work/blob.txt", &st);
+void test_blob_txt() {
+  // Check that /work/blob.txt contain the extected data
```
kripken:
```suggestion
  // Check that /work/blob.txt contains the expected data
```

test/fs/test_workerfs.c
```diff
@@ -15,46 +15,61 @@
...
-  assert(strncmp(buf, SECRET, SECRET_LEN) == 0);
+  ssize_t cnt = read(fd, buf, 1000);
+  printf("read blob.txt: %ld, expecting: %lu\n", cnt, file_len);
+  assert(cnt == file_len);
+  assert(strcmp(buf, content) == 0);
+}
+
+void test_file_txt() {
+  // Seek half way through /work/file.txt then verify that the data
+  // read from the that point on
```
kripken:
```suggestion
  // Seek half way through /work/file.txt then verify the data
  // read from that point on
```

Issue #25814, opened 2025-11-17
browser.test_dylink_dso_needed fails due to file-no-found


sbc100:
Then loading libb.so from the DT_NEEDED section of liba.so this test currently fails, even though libb.so exists in the filesystem (via --embed-file).

PR #25815, opened 2025-11-17
Fix test arguments in test_dylink_dso_needed

sbc100:
I fixed the arguments so the non-worker version of this test now runs, but sadly is currently failing so I opened #25814

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -3636,11 +3636,12 @@ Module["preRun"] = () => {
   # verify that dynamic linking works in all kinds of in-browser environments.
   # don't mix different kinds in a single test.
   @parameterized({
-    '': ([0],),
-    'inworker': ([1],),
+    '': (0,),
+    'inworker': (1,),
   })
   def test_dylink_dso_needed(self, inworker):
-    self.cflags += ['-O2']
+    if not inworker:
+      self.skipTest('https://github.com/emscripten-core/emscripten/issues/25814')
 
     def do_run(src, expected_output, cflags):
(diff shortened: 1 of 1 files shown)


test/test_browser.py
```diff
@@ -3636,11 +3636,12 @@ def test_dlopen_blocking(self):
   # verify that dynamic linking works in all kinds of in-browser environments.
   # don't mix different kinds in a single test.
   @parameterized({
-    '': ([0],),
-    'inworker': ([1],),
+    '': (0,),
+    'inworker': (1,),
   })
   def test_dylink_dso_needed(self, inworker):
-    self.cflags += ['-O2']
```
kripken, line 3643: We don't need the `-O2`?

lgtm otherwise

test/test_browser.py
```diff
@@ -3636,11 +3636,12 @@ def test_dlopen_blocking(self):
   # verify that dynamic linking works in all kinds of in-browser environments.
   # don't mix different kinds in a single test.
   @parameterized({
-    '': ([0],),
-    'inworker': ([1],),
```
kripken, line 3640: Maybe make these True/False? But I don't feel strongly.

sbc100: Switching to true/false would not have helped prevent this logic bug sadly

test/test_browser.py
sbc100: `-O2` was just making the errors harder to interpret there so I dropped it.  I don't think its there for a any good reason.

PR #25831, opened 2025-11-19
[CI] Remove auto submit from update-website.yml. NFC

sbc100:


sbc100:
Landing TBR to prevent runnaway auto-commits on the site repo

PR #25835, opened 2025-11-20
Fix Firefox 91 ESR and Firefox 102 ESR test runs

juj:
#25823 added a new ES6 Worker test. Gate it behind feature check to fix Firefox 91 ESR and Firefox 102 ESR test runs.

### DIFF
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -5671,9 +5671,11 @@ Module["preRun"] = () => {
     'es6': (['-sEXPORT_ES6', '--extern-post-js', test_file('modularize_post_js.js')], 9999),
   })
   def test_cross_origin(self, args, port):
-    # Verfies that the emscripten-generted JS and Wasm can be hosted on a different origin.
-    # This test create a second HTTP server running a different port that servers files from `subdir`.
-    # The main html is the servers from the normal 8888 server while the JS and Wasm are hosted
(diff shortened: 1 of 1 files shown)


test/test_browser.py
```diff
@@ -5671,9 +5671,11 @@ def test_rollup(self):
     'es6': (['-sEXPORT_ES6', '--extern-post-js', test_file('modularize_post_js.js')], 9999),
   })
   def test_cross_origin(self, args, port):
-    # Verfies that the emscripten-generted JS and Wasm can be hosted on a different origin.
-    # This test create a second HTTP server running a different port that servers files from `subdir`.
-    # The main html is the servers from the normal 8888 server while the JS and Wasm are hosted
+    if '-sEXPORT_ES6' in args and browser_should_skip_feature('EMTEST_LACKS_ES6_WORKERS', Feature.WORKER_ES6_MODULES):
+      self.skipTest('This test requires a browser with ES6 Module Workers support')
```
sbc100, line 5675: I guess we should have a method version of `@requires_es6_workers` decorator.  But this LGTM for now.

juj: Yeah, the awkwardness of decorators not combining well with these parameterized modes repeats in many tests. That would be a bit of a cleanup refactor to figure out how to create a simple pattern that would work well in both.

Issue #25836, opened 2025-11-20
regression: SingleLineTestResult cuts off test result printing


juj:
The PR that changed test runner to use the single line test result printer caused a problem that test output is cut off at the end.

Noticed this when adding #25835:

<img width="2104" height="933" alt="Image" src="https://github.com/user-attachments/assets/4448bf8e-9b62-4cc1-8912-659b0f120dd9" />

Above screenshot shows Command Prompt window on top a browser window.

the line should print `skipped 'This test requires a browser with ES6 Module Workers support'`

sbc100:
It should indeed be truncate, but it should happen at the terminal width.. not in the middle like that.  I'll take a look.

PR #25838, opened 2025-11-20
Fix status message output for single line test output. NFC

sbc100:
We now use the full terminal width and only truncate when we need to.

Fixes: #25836

### DIFF
--- a/test/single_line_runner.py
+++ b/test/single_line_runner.py
@@ -16,14 +16,17 @@ def clearline(stream):
 
 
 def term_width():
-  return shutil.get_terminal_size()[0]
+  return shutil.get_terminal_size().columns
 
 
 class SingleLineTestResult(ColorTextResult):
   """Similar to the standard TextTestResult but uses ANSI escape codes
   for color output and reusing a single line on the terminal.
   """
-  max_status_msg = 20
+
+  # Reserve at least 20 columns for the status message
+  min_status_msg = 20
+  status_limit = None
 
   def _write_status(self, _test, status):
     # Add some color to the status message
@@ -36,7 +39,7 @@ class SingleLineTestResult(ColorTextResult):
       status += '\n'
     else:
       color = CYAN
-    status = status[:self.max_status_msg]
+    status = status[:self.status_limit]
     line = f'{with_color(color, status)}'
     self.stream.write(line)
     self.stream.flush()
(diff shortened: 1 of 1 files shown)


juj (approved):
Nice, thanks for taking a quick peek. 👍

test/single_line_runner.py
```diff
@@ -47,12 +50,17 @@ def startTest(self, test):
...
     prefix_len = self.writeProgressPrefix()
-    max_desc = term_width() - prefix_len - len(' ... ') - self.max_status_msg
+    total_width = term_width()
+    max_desc = total_width - prefix_len - len(' ... ') - self.min_status_msg
     desc = str(test)[:max_desc]
     self.stream.write(desc)
     self.stream.write(' ... ')
     self.stream.flush()
 
+    # Calcualte the remaining terminal width available for the _write_status
```
kripken, line 60:
```suggestion
    # Calculate the remaining terminal width available for the _write_status
```

test/single_line_runner.py
```diff
@@ -47,12 +50,17 @@ def startTest(self, test):
...
+    # message This will never be lower that `self.max_status_msg`
```
```suggestion
    # message This will never be lower than `self.max_status_msg`
```

test/single_line_runner.py
```diff
@@ -47,12 +50,17 @@ def startTest(self, test):
...
-    max_desc = term_width() - prefix_len - len(' ... ') - self.max_status_msg
+    total_width = term_width()
+    max_desc = total_width - prefix_len - len(' ... ') - self.min_status_msg
     desc = str(test)[:max_desc]
     self.stream.write(desc)
     self.stream.write(' ... ')
     self.stream.flush()
 
+    # Calcualte the remaining terminal width available for the _write_status
+    # message. This will never be lower that `self.max_status_msg`
```
kripken, line 61:
```suggestion
    # message. This will never be lower than `self.max_status_msg`
```

test/single_line_runner.py
```diff
@@ -16,14 +16,17 @@ def clearline(stream):
...
 
 class SingleLineTestResult(ColorTextResult):
   """Similar to the standard TextTestResult but uses ANSI escape codes
   for color output and reusing a single line on the terminal.
   """
-  max_status_msg = 20
+
+  # Reserve at least 20 columns for the status message
+  min_status_msg = 20
+  status_limit = None
```
kripken, line 29: Why set this to none if it is set lower down?

kripken (approved):

kripken:
oops, I didn't notice the automerge, sorry... edit: oh wait, it was racing approvals

test/single_line_runner.py
sbc100: It just starts out as None, and gets set when each line is logged

sbc100:
Fixed typos in followup (directly pushed)

PR #25839, opened 2025-11-20
Fix preserving layout(std140) directives with -sGL_EXPLICIT_UNIFORM_BINDING

### LINKED ISSUE #25828: A bug of glShaderSource in libwebgl.js?
The latest main branch

https://github.com/emscripten-core/emscripten/blob/main/src/lib/libwebgl.js#L3273-L3278
Check the code here.
According the comment, the 1st replacement should not overrides the 2nd & the 3rd replacement.
The layout rules should be kept.

Actually, the 1st replacement make the 2nd & the 3rd replacement useless.
This makes some custom development WebGL2 implementation render incorrectly on some devices.
For example, the Kwai Minigame renders incorrectly on Vivo S18.

source = "layout(binding = 1, std140)";
Actual Result: source = ""
Expected Result: source = "layout(std140)"

source = "layout(std140, binding = 1)" ;
Actual Result: source = ""
Expected Result: source = "layout(std140)"

juj:
Fix preserving layout(std140) directives in WebGL shaders when using `-sGL_EXPLICIT_UNIFORM_BINDING` WebGL extension. Improve the existing test to catch the reported scenario. 

Fixes #25828.

### DIFF
--- a/src/lib/libwebgl.js
+++ b/src/lib/libwebgl.js
@@ -3273,7 +3273,7 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
     // Remove all the layout(binding = x) directives so that they do not make
     // their way to the actual WebGL shader compiler. These regexes get quite
     // hairy, check against https://regex101.com/ when working on these.
-    source = source.replace(/layout\s*\([^,]*?binding\s*=\s*([-\d]+)[^,]*?\)/g, ''); // "layout(binding = 3)" -> ""
+    source = source.replace(/layout\s*\(\s*binding\s*=\s*([-\d]+)\s*\)/g, ''); // "layout(binding = 3)" -> ""
     source = source.replace(/(layout\s*\((.*?)),\s*binding\s*=\s*([-\d]+)\)/g, '$1)'); // "layout(std140, binding = 1)" -> "layout(std140)"
     source = source.replace(/layout\s*\(\s*binding\s*=\s*([-\d]+)\s*,\s*(.*?)\)/g, 'layout($2)'); // "layout(binding = 1, std140)" -> "layout(std140)"
 


src/lib/libwebgl.js
```diff
@@ -3273,9 +3273,9 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
     // Remove all the layout(binding = x) directives so that they do not make
     // their way to the actual WebGL shader compiler. These regexes get quite
     // hairy, check against https://regex101.com/ when working on these.
-    source = source.replace(/layout\s*\(.*?binding\s*=\s*([-\d]+).*?\)/g, ''); // "layout(binding = 3)" -> ""
+    source = source.replace(/layout\s*\([^,]*?binding\s*=\s*([-\d]+)[^,]*?\)/g, ''); // "layout(binding = 3)" -> ""
     source = source.replace(/(layout\s*\((.*?)),\s*binding\s*=\s*([-\d]+)\)/g, '$1)'); // "layout(std140, binding = 1)" -> "layout(std140)"
-    source = source.replace(/layout\s*\(\s*binding\s*=\s*([-\d]+)\s*,(.*?)\)/g, 'layout($2)'); // "layout(binding = 1, std140)" -> "layout(std140)"
+    source = source.replace(/layout\s*\(\s*binding\s*=\s*([-\d]+)\s*,\s*(.*?)\)/g, 'layout($2)'); // "layout(binding = 1, std140)" -> "layout(std140)"
```
juj, line 3278: This `\s*` addition is not part of the fix itself, but just to improve cleanliness of the string replace for the assert() in the test below. Without this, `layout(binding = 1, std140)` was converted to form `layout( std140)`, so the `\s*` here swallows the stray space before `std140`.

src/lib/libwebgl.js
```diff
@@ -3273,9 +3273,9 @@ for (/**@suppress{duplicate}*/var i = 0; i <= {{{ GL_POOL_TEMP_BUFFERS_SIZE }}};
     // Remove all the layout(binding = x) directives so that they do not make
     // their way to the actual WebGL shader compiler. These regexes get quite
     // hairy, check against https://regex101.com/ when working on these.
-    source = source.replace(/layout\s*\(.*?binding\s*=\s*([-\d]+).*?\)/g, ''); // "layout(binding = 3)" -> ""
+    source = source.replace(/layout\s*\([^,]*?binding\s*=\s*([-\d]+)[^,]*?\)/g, ''); // "layout(binding = 3)" -> ""
```
juj: The actual fix here is to change the regex match to use `[^,]` (capture anything else than a `,` character), instead of a `.` (capture any character).

This way the first replace won't catch on `layout(binding = 3, ...)` or `layout(..., binding = 3)`

sbc100: Why not just use `\s*` here?  i.e. what chars are allowed between the opening paren and the word `binding`?

juj: It could be `layout(std140, binding=5)` for example.

sbc100: I you *don't* want to match those is that right?

Why doesn't `\s*` work then?     Otherwise won't you end up matching things like `layout(whatever binding = 0)`?

I guess its not valid but I don't see why you would want to match anything except space?

sbc100: Also, can you remove the `?` that follow the `*` operators?  Isn't it redundant to say `[^,]*?` since the `*` already includes matching zero things.

juj: Err, yeah, I somehow read your first question as "why not use `.*` here?".

Now I see what you meant. Definitely using `\s*` is simpler than `[^,]*?`. Updated to that form. Thanks!

kainino0x (approved):
FWIW @juj I discussed the issue more with @SetoKaiba on chat and determined that the root issue, that exposed this bug, was actually in the non-browser runtime they're using, which doesn't implement one of the WebGL special cases: in WebGL (unlike GLES), uniform buffers [always have std140 layout](https://registry.khronos.org/webgl/specs/latest/2.0/#5.25).

Because of that, for WebGL I don't think it really matters if the `layout(std140)` is preserved or not. I don't know if this changes what you want to do in Emscripten.

SetoKaiba:
> FWIW @juj I discussed the issue more with @SetoKaiba on chat and determined that the root issue, that exposed this bug, was actually in the non-browser runtime they're using, which doesn't implement one of the WebGL special cases: in WebGL (unlike GLES), uniform buffers [always have std140 layout]().
> 
> Because of that, for WebGL I don't think it really matters if the `layout(std140)` is preserved or not. I don't know if this changes what you want to do in Emscripten.

Yes.
After a discussion with @kainino0x, the issue is confirmed that the Kwai minigame runtime doesn't implement the spec strictly. https://registry.khronos.org/webgl/specs/latest/2.0/#5.25
I think there maybe two options here.
1. Keep the layout, current PR.
2. Keep only the first replacement. std140 is the only supported.

juj:
Yeah, that makes sense. It is still good to land this, given that it can otherwise produce red herrings to developers (like happened in this case)

PR #25841, opened 2025-11-20
Split test_nested_structs into its own file. NFC

sbc100:


sbc100:
Do you think this test is actually doing anything useful these days?  Seems like a maybe a fastcomp era test?

kripken:
Yeah, more fastcomp and so covered (I hope) by clang. Though if verifying clang has coverage is hard, we can just keep it (but maybe move to other so it runs less)

PR #25842, opened 2025-11-20
Enable test_varargs_byval. NFC

sbc100:
Seems to work now..


sbc100:
I'm not sure this is actually doing anything useful...

PR #25843, opened 2025-11-20
Add test runner option --log-test-environment

juj:
Add test runner option --log-test-environment, which will introspect and print detailed information about Emscripten tool setup: Emcc, Clang, Binaryen, Node, Python and LLVM. This helps CI maintainers emit info to test run logs of the active setup for reference.

`test/runner --log-test-environment <tests_to_run>` will then print a following log info dump before starting the tests:

```

C:\emsdk\emscripten\main>test\runner --log-test-environment
======================== Test Setup ========================
Test time: Thursday, November 20, 2025 22:28:18 UTC
Python: "C:\emsdk\python\3.13.3_64bit\python.exe". Version: 3.13.3 (tags/v3.13.3:6280bb5, Apr  8 2025, 14:47:33) [MSC v.1943 64 bit (AMD64)]
Emscripten test runner path: "C:\emsdk\emscripten\main\test\runner.py"

Emscripten repository: "C:\emsdk\emscripten\main"
emscripten-version.txt: 4.0.21-git

Emscripten commit cce3a701889bbc5b107056b0de3fab8d5ce6f885
Author: Jukka Jylänki <jujjyl@gmail.com>
Date:   Fri Nov 21 00:26:51 2025 +0200

    Add test runner option --log-test-environment, which will introspect and print detailed information about Emscripten tool setup: Emcc, Clang, Binaryen, Node, Python and LLVM. This helps CI maintainers emit info to test run logs of the active setup for reference.

EM_CONFIG: "C:\emsdk\.emscripten"

import os
emsdk_path = os.path.dirname(os.getenv('EM_CONFIG')).replace('\\', '/')
LLVM_ROOT = emsdk_path + '/llvm/git/build_main_vs2022_64/Release/bin'
NODE_JS = emsdk_path + '/node/22.16.0_64bit/bin/node.exe'
EMSDK_CMAKE = emsdk_path + '/cmake/4.2.0-rc3_64bit/bin/cmake.exe'
PYTHON = emsdk_path + '/python/3.13.3_64bit/python.exe'
EMSCRIPTEN_ROOT = emsdk_path + '/emscripten/main'
BINARYEN_ROOT = emsdk_path + '/binaryen/main_vs2022_64bit_binaryen'
MINGW_ROOT = emsdk_path + '/mingw/7.1.0_64bit'
NINJA = emsdk_path + '/ninja/git-release_64bit/bin'
EMSDK_ACTIVATED_TEST_BROWSER = emsdk_path + '/firefox/78.15.0esr_64bit/firefox.exe'

NODE_JS: ['C:/emsdk/node/22.16.0_64bit/bin/node.exe']. Version: v22.16.0
BINARYEN_ROOT: C:/emsdk/binaryen/main_vs2022_64bit_binaryen
wasm-opt version: wasm-opt version 124 (version_124-122-g8a66d2a7d)
Binaryen git directory: "C:\emsdk\binaryen\main"

Binaryen commit 8a66d2a7d48a0014dce5e22a2fa8eee12eb09b7c
Author: Alon Zakai <azakai@google.com>
Date:   Thu Nov 6 11:53:49 2025 -0800

    GUFA: Represent imported functions as globals (#8025)

    The meaning of Global in PossibleContents changes from "an actual wasm
    Global" to "an immutable global wasm thing, either a Global or a Function."
    Imported wasm Functions are effectively immutable values in the global
    scope - not concrete, specifically-known functions - so this is more correct.

    In particular, two imported Globals (of compatible types) might or might not
    be the same, and likewise with imported Functions. And that is not the case
    for defined Functions - they are definitely different.

    This does not fix the issues related to imported function type handling -
    #7993 does that - but this refactoring makes it possible to properly
    optimize imported functions afterward.

    This does make the PossibleContents object grow from 32 to 40 bytes,
    which might be why this adds 3-4% overhead to GUFA, but I don't see a
    good way to avoid that, unfortunately. Keeping our ability to optimize
    imported functions might be worth that 3-4% (we recently saw a few digits
    improvement due to properly optimizing imported externs, for example,
    from #8005).

LLVM_ROOT: C:/emsdk/llvm/git/build_main_vs2022_64/Release/bin
LLVM git directory: "C:/emsdk/llvm/git\src"

LLVM commit d65be16ab6adf00af21e75d29049ae5de0f3a38a
Author: Ryan Cowan <ryan.cowan@arm.com>
Date:   Mon Nov 17 15:55:40 2025 +0000

    [AArch64][GlobalISel] Add combine for build_vector(unmerge, unmerge, undef, undef) (#165539)

    This PR adds a new combine to the `post-legalizer-combiner` pass. The
    new combine checks for vectors being unmerged and subsequently padded
    with `G_IMPLICIT_DEF` values by building a new vector. If such a case is
    found, the vector being unmerged is instead just concatenated with a
    `G_IMPLICIT_DEF` that is as wide as the vector being unmerged.

    This removes unnecessary `mov` instructions in a few places.

Clang: "C:/emsdk/llvm/git/build_main_vs2022_64/Release/bin\clang.exe"
clang version 22.0.0git (https://github.com/llvm/llvm-project.git d65be16ab6adf00af21e75d29049ae5de0f3a38a)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\emsdk\llvm\git\build_main_vs2022_64\Release\bin

EMTEST_BROWSER: "C:\\emsdk\\firefox\\78.15.0esr_64bit\\firefox.exe" -new-instance -wait-for-browser
Firefox version: 78
Not detected as a Safari browser

EMSDK: "C:/emsdk"

Emsdk commit fe91d96cfd53f505a2ce561b01320124bff438a1
Author: Jukka Jylänki <jujjyl@gmail.com>
Date:   Mon Nov 17 18:41:28 2025 +0200

    Use exit_with_error

==================== End of Test Setup =====================
```

### DIFF
--- a/test/runner.py
+++ b/test/runner.py
@@ -568,7 +568,7 @@ def cleanup_emscripten_temp():
 
 
 def print_repository_info(directory, repository_name):
-  current_commit = utils.run_process(['git', 'show', '--no-patch'], cwd=directory, stdout=subprocess.PIPE).stdout.strip()
+  current_commit = utils.run_process(['git', 'log', '--abbrev-commit', '-n1', '--pretty=oneline'], cwd=directory, stdout=subprocess.PIPE).stdout.strip()
   print(f'\n{repository_name} {current_commit}\n')
   local_changes = utils.run_process(['git', 'diff'], cwd=directory, stdout=subprocess.PIPE).stdout.strip()
   if local_changes:


test/runner.py
```diff
@@ -563,6 +567,94 @@ def cleanup_emscripten_temp():
         pass
 
 
+def print_repository_info(directory, repository_name):
+  import subprocess
```
sbc100: No need for this import here

test/runner.py
```diff
@@ -563,6 +567,94 @@ def cleanup_emscripten_temp():
...
+def log_test_environment():
+  """Print detailed information about the current test environment. Useful for
+  logging test run configuration in a CI."""
+  print('======================== Test Setup ========================')
+  print(f'Test time: {datetime.datetime.now(datetime.timezone.utc).strftime("%A, %B %d, %Y %H:%M:%S %Z")}')
+  print(f'Python: "{sys.executable}". Version: {sys.version}')
+  print(f'Emscripten test runner path: "{os.path.realpath(__file__)}"')
+
+  if os.path.isdir(os.path.join(config.EMSCRIPTEN_ROOT, '.git')):
+    print(f'\nEmscripten repository: "{config.EMSCRIPTEN_ROOT}"')
```
sbc100: I'm trying to remove `EMSCRIPTEN_ROOT` as a thing.   You can just use `path_from_root()`  or `__rootpath__` in this file.

As of today we only support running tests againt the version of emscripten in which the test runner lives.

test/runner.py
```diff
@@ -563,6 +567,94 @@ def cleanup_emscripten_temp():
...
+
+  emscripten_version = os.path.realpath(os.path.join(config.EMSCRIPTEN_ROOT, 'emscripten-version.txt'))
```
sbc100: `path_from_root('emscripten-version.txt')`

test/runner.py
```diff
@@ -563,6 +567,94 @@ def cleanup_emscripten_temp():
...
+  if os.path.isfile(emscripten_version):
+    print(f'emscripten-version.txt: {utils.read_file(emscripten_version).strip()}')
```
sbc100: You can just use `utils.EMSCRIPTEN_VERSION` here (its is populated from this file)

test/runner.py
```diff
@@ -563,6 +567,94 @@ def cleanup_emscripten_temp():
...
+
+  if os.path.isdir(os.path.join(config.EMSCRIPTEN_ROOT, '.git')):
+    print_repository_info(config.EMSCRIPTEN_ROOT, 'Emscripten')
+
+  print(f'EM_CONFIG: "{config.EM_CONFIG}"')
+  if os.path.isfile(config.EM_CONFIG):
+    print(f'\n{utils.read_file(config.EM_CONFIG).strip()}\n')
+
+  node_js_version = subprocess.run(config.NODE_JS + ['--version'], stdout=subprocess.PIPE, text=True).stdout.strip()
+  print(f'NODE_JS: {config.NODE_JS}. Version: {node_js_version}')
```
sbc100, line 601: We already have this available via `utils.get_node_version()`

test/runner.py
```diff
@@ -563,6 +567,94 @@ def cleanup_emscripten_temp():
...
+
+  print(f'BINARYEN_ROOT: {config.BINARYEN_ROOT}')
+  wasm_opt_version = subprocess.run([os.path.join(config.BINARYEN_ROOT, 'bin', 'wasm-opt.exe'), '--version'], stdout=subprocess.PIPE, text=True).stdout.strip()
+  print(f'wasm-opt version: {wasm_opt_version}')
```
sbc100, line 605: We have `building.check_binaryen()` that does this so we could maybe refactor that to export `building.get_binaryen_version()`?

test/runner.py
```diff
@@ -563,6 +567,94 @@ def cleanup_emscripten_temp():
...
+        return None
+      dir = os.path.dirname(dir)
+
+  llvm_git_root = find_llvm_git_root(config.LLVM_ROOT)
+  if llvm_git_root:
+    print(f'LLVM git directory: "{llvm_git_root}"')
+    print_repository_info(llvm_git_root, 'LLVM')
+
+  clang_version = subprocess.run([shared.CLANG_CC, '--version'], stdout=subprocess.PIPE, text=True).stdout.strip()
+  print(f'Clang: "{shared.CLANG_CC}"\n{clang_version}\n')
```
sbc100, line 634: We have `shared.get_clang_version()` for this.

test/runner.py
```diff
@@ -563,6 +567,94 @@ def cleanup_emscripten_temp():
...
+
+  print(f'EMTEST_BROWSER: {browser_common.EMTEST_BROWSER}')
+  firefox_version = browser_common.get_firefox_version()
+  if firefox_version != 2**31 - 1:
```
sbc100: Maybe use `is_firefox` here ?

test/runner.py
```diff
@@ -563,6 +567,94 @@ def cleanup_emscripten_temp():
         pass
 
 
+def print_repository_info(directory, repository_name):
+  import subprocess
+  current_commit = subprocess.run(['git', 'log', '-n1'], cwd=directory, stdout=subprocess.PIPE, text=True).stdout.strip()
```
sbc100: Maybe `git show -s` ?

sbc100: Can you use `utils.run_process` instead of using `subprocess` directly (at least where possible).   I'm trying to keep all subprocess usage using those helpers.

test/runner.py
```diff
@@ -563,6 +567,94 @@ def cleanup_emscripten_temp():
...
+  print(f'\n{repository_name} {current_commit}\n')
+  local_changes = subprocess.run(['git', 'diff'], cwd=directory, stdout=subprocess.PIPE, text=True).stdout.strip()
```
sbc100: Maybe just remove the `stdout=subprocess.PIPE` so that the output goes directly to stdout?

sbc100 (commented):
Seems like useful thing!

test/runner.py
juj: There's `shared.get_node_version()`, but that parses the output. I think it's better in this case to output the raw version string as emitted by `NODE_JS --version`.

test/runner.py
juj: That too parses the output - it would be preferable to print the raw output here.

test/runner.py
juj: This doesn't work for Windows at least, but enters an interactive git diff prompt.

test/runner.py
```diff
@@ -631,20 +630,22 @@ def find_llvm_git_root(dir):
     print(f'LLVM git directory: "{llvm_git_root}"')
     print_repository_info(llvm_git_root, 'LLVM')
 
-  clang_version = subprocess.run([shared.CLANG_CC, '--version'], stdout=subprocess.PIPE, text=True).stdout.strip()
+  clang_version = utils.run_process([shared.CLANG_CC, '--version'], stdout=subprocess.PIPE, text=True).stdout.strip()
```
sbc100: You don't need `text=True` here, or above

test/runner.py
```diff
@@ -584,25 +583,25 @@ def log_test_environment():
   print(f'Python: "{sys.executable}". Version: {sys.version}')
   print(f'Emscripten test runner path: "{os.path.realpath(__file__)}"')
 
-  if os.path.isdir(os.path.join(config.EMSCRIPTEN_ROOT, '.git')):
-    print(f'\nEmscripten repository: "{config.EMSCRIPTEN_ROOT}"')
+  if os.path.isdir(os.path.join(__rootpath__, '.git')):
```
sbc100: `utils.path_from_root('.git')`

test/runner.py
```diff
@@ -606,6 +699,9 @@ def set_env(name, option_value):
 
   browser_common.init(options.force_browser_process_termination)
 
+  if options.log_test_environment:
```
sbc100: I guess we should probably print all this into in circle CI too.  The best way to do that is probably to add `or os.getenv('CI')`.

juj: Sounds good. Added os.getenv('CI').

sbc100:
Do we really need show the whole commit messsage from at HEAD for each repo?

Isn't the revision enough?   The commit message seems to add a log noise to me.

juj:
I thought it would be a good way to read out what the current tested commit was.

When I see a list of

```
<red>
<red>
<green>
<green>
<green>
```

test runs on my CI (runs in newest to oldest order), I could look at the first red run, and then read the commit message there - that would give a direct PR #number to find what regressed.

If I only printed a commit hash, I would then need to go and `git log` it manually. Of course doable, but I figured why not just provide the information automatically - that way the test logs would carry more information without having to translate manually.

sbc100:
> I thought it would be a good way to read out what the current tested commit was.
> 
> When I see a list of
> 
> ```
> <red>
> <red>
> <green>
> <green>
> <green>
> ```
> 
> test runs on my CI (runs in newest to oldest order), I could look at the first red run, and then read the commit message there - that would give a direct PR #number to find what regressed.
> 
> If I only printed a commit hash, I would then need to go and `git log` it manually. Of course doable, but I figured why not just provide the information automatically - that way the test logs would carry more information without having to translate manually.

I think the commit has is normally enough for me, but maybe can do `git log --abbrev-commit -n1 --pretty=oneline` as a compromise?

juj:
Ok, that works.

sbc100 (approved):

PR #25844, opened 2025-11-21
[wasm64] Fix pointer math in glTransformFeedbackVaryings() and glGetUniformIndices()

gkv311:
Extend patch from #21184 to two more functions.

### DIFF
--- a/src/lib/libwebgl2.js
+++ b/src/lib/libwebgl2.js
@@ -386,7 +386,7 @@ var LibraryWebGL2 = {
     program = GL.programs[program];
     var vars = [];
     for (var i = 0; i < count; i++)
-      vars.push(UTF8ToString({{{ makeGetValue('varyings', 'i*4', 'i32') }}}));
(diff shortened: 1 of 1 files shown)


src/lib/libwebgl2.js
```diff
@@ -386,7 +386,7 @@ var LibraryWebGL2 = {
     program = GL.programs[program];
     var vars = [];
     for (var i = 0; i < count; i++)
-      vars.push(UTF8ToString({{{ makeGetValue('varyings', 'i*4', 'i32') }}}));
+      vars.push(UTF8ToString({{{ makeGetValue('varyings', 'i*' + POINTER_SIZE, '*') }}}));
```
sbc100, line 389: Are there any tests that cover these functions?

gkv311: Not any inside `test` folder as far as I can see.

sbc100: Wow, thats no bad.  No testing here.  We should certainly add some, but I don't think we need to block this change on that.

Issue #25846, opened 2025-11-21
emscripten_html5_remove_event_listener does not clean fullscreenchange entirely


ypujante:
Due to this code in Emscripten (libhtml5.js)

```js
  emscripten_set_fullscreenchange_callback_on_thread: (target, userData, useCapture, callbackfunc, targetThread) => {
    if (!JSEvents.fullscreenEnabled()) return {{{ cDefs.EMSCRIPTEN_RESULT_NOT_SUPPORTED }}};
#if DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR
    target = findEventTarget(target);
#else
    target = target ? findEventTarget(target) : specialHTMLTargets[{{{ cDefs.EMSCRIPTEN_EVENT_TARGET_DOCUMENT }}}];
#endif
    if (!target) return {{{ cDefs.EMSCRIPTEN_RESULT_UNKNOWN_TARGET }}};

#if MIN_SAFARI_VERSION != TARGET_NOT_SUPPORTED
    // As of Safari 13.0.3 on macOS Catalina 10.15.1 still ships with prefixed webkitfullscreenchange. TODO: revisit this check once Safari ships unprefixed version.
    registerFullscreenChangeEventCallback(target, userData, useCapture, callbackfunc, {{{ cDefs.EMSCRIPTEN_EVENT_FULLSCREENCHANGE }}}, "webkitfullscreenchange", targetThread);
#endif

    return registerFullscreenChangeEventCallback(target, userData, useCapture, callbackfunc, {{{ cDefs.EMSCRIPTEN_EVENT_FULLSCREENCHANGE }}}, "fullscreenchange", targetThread);
  },
```

when calling `emscripten_set_fullscreenchange_callback_on_thread` 2 event listeners are actually added:

```text
2: Object { eventTypeString: "webkitfullscreenchange", eventTypeId: 19, userData: 143568, … }
​3: Object { eventTypeString: "fullscreenchange", eventTypeId: 19, userData: 143568, … }
​
```

They are identical besides the `eventTypeString`

As a result calling `emscripten_html5_remove_event_listener` removes only one of them:

```text
0: Object { eventTypeString: "fullscreenchange", eventTypeId: 19, userData: 143568, … }
```

ypujante:
I suppose we could do the following fix:

```js
    removeSingleHandler(eventHandler) {
      for (var [i, handler] of JSEvents.eventHandlers.entries()) {
[...]

PR #25847, opened 2025-11-21
Fixes emscripten_html5_remove_event_listener when multiple listeners

ypujante:
Fixes: #25846

Iterating and removing ALL instances of a given listener.

I wanted to add a test for this use case, but I am not sure how I would do that. Is there a way to have a test that targets Safari? And have the code of the test do something different in this case?

### DIFF
--- a/test/codesize/test_codesize_hello_dylink_all.json
+++ b/test/codesize/test_codesize_hello_dylink_all.json
@@ -1,7 +1,7 @@
 {
-  "a.out.js": 245720,
+  "a.out.js": 245732,
   "a.out.nodebug.wasm": 574007,
-  "total": 819727,
+  "total": 819739,
   "sent": [
     "IMG_Init",
     "IMG_Load",


ypujante:
I tried to fix the code size test failure, but once again I am unsuccessful. I reverted the (code size) change since it is wrong anyway.

ypujante:
I apologize for not being able to run the codesize myself. I try but it just doesn't give the right results and I just don't know why. In any case, thank you for doing it.

sbc100:
> I apologize for not being able to run the codesize myself. I try but it just doesn't give the right results and I just don't know why. In any case, thank you for doing it.

Its very strange.  Must be some kind of configuration issue but I can't guess what it might be.  I think we tried all the things I could think of. 

Oh, there was one other cause of discrpency that can stem from the version of zlib on you system.  Can you confirm, when you see discrepancies is it only the gz compressed sizes?

ypujante:
From this commit  https://github.com/emscripten-core/emscripten/commit/9265ba423ccba2650f24eb0e568af7e1bc516977 it doesn't look like it was just the zip files

